PR #53, opened 2025-01-13
Fsx develop

fsx950223:
Modified interface to adapt to sgl, added inplace output.
Fixed the bug causing naive multi-GPU CUDA graph to crash.


ater/ops/attention.py
```diff
@@ -53,6 +53,7 @@ def pa_fwd_asm(
     value_cache: torch.Tensor,
     block_tables: torch.Tensor,
     context_lens: torch.Tensor,
+    max_num_blocks: int,
```
valarLip, line 56: update to op test is alse needed, when we have changed interface. please, we always need make sure our op test runable

PR #64, opened 2025-01-19
fix quant test fail

carlushuang:
python3 `op_tests/test_quant.py`. with this PR, this test can pass.
1. change per-tensor torch use `240` as max
2. change the per-tensor test to calculate scale, instead of using scale passed in.
3. need use `torch.zeros` for ater pertensor quant, scale buffer creation
4. For fp8, we need to use `atol=1e-3, rtol=0.125`

### DIFF
--- a/ater/tuned_gemm.py
+++ b/ater/tuned_gemm.py
@@ -41,8 +41,8 @@ class TunedGemm:
                     lambda s: getHipblasltKernelName(s.solidx) if s.libtype == 'hipblaslt' else "", axis=1)
                 pd.set_option('display.max_colwidth', 100)
                 assert hipblasltKernelNames.equals(
-                    self.bestsols['kernelName'].fillna("")), "error: gradlib tune gemm not match the current environment, need re-tune!!!\n" + \
-                    f'differece:\n{pd.concat([self.bestsols[['solidx','kernelName']], hipblasltKernelNames], axis=1)[hipblasltKernelNames != self.bestsols["kernelName"].fillna("")]}'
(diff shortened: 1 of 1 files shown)


op_tests/test_quant.py
```diff
@@ -11,7 +11,7 @@
 def test_ater_perTensorQuantFp8(input, sacle=None):
```
valarLip: sacle->scale...

op_tests/test_quant.py
```diff
@@ -51,23 +51,23 @@ def test_quant(m, n, dtype=torch.bfloat16):
...
     # checkAllclose(a_sacle, b_sacle, msg='sacle')
     # print(f'{a_sacle=}, {b_sacle=}')
 
-    scale = a_sacle
-    (a, a_sacle), avg_a = test_torch_perTensorQuantFp8(input, scale)
-    (b, b_sacle), avg_b = test_ater_perTensorQuantFp8(input, scale)
+    # scale = a_sacle
+    (a, a_sacle), avg_a = test_torch_perTensorQuantFp8(input)
+    (b, b_sacle), avg_b = test_ater_perTensorQuantFp8(input)
     msg = f"[perf] static_perTensorQuantFp8 dim: {str(dim):<20} dtype: {dtype}, torch avg: {avg_a:<8.2f} us, ater avg: {avg_b:<8.2f} us, uplift: {avg_a/avg_b-1:<5.1%}"
```
valarLip, line 64: static_perTensorQuantFp8 .... remove or?

PR #111, opened 2025-02-10
Add Triton Layer Norm Kernel

lucas-santos-amd:
Add Triton Layer Norm Kernel

### DIFF
--- a/op_tests/triton/test_layernorm.py
+++ b/op_tests/triton/test_layernorm.py
@@ -1,0 +1,37 @@
+import triton
+import torch
+import triton.language as tl
+import pytest
+from aiter.ops.triton.norm import layer_norm
+
+
+# pytest
+@pytest.mark.parametrize(
+    "M, N",
+    [
+        (1823, 781),
+        (2, 128),
+        (1, 4),
+        (128, 2),
+        (1, 128),
+        (8192, 8192),
+        (4096, 8192),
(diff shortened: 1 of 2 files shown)


op_tests/triton/test_layernorm.py
```diff
@@ -0,0 +1,24 @@
...
+    x = torch.randn(M, N, device='cuda')
+    w_shape = (N, )
+    w = torch.rand(w_shape, device='cuda')
+    b = torch.rand(w_shape, device='cuda')
+
+    # forward pass
+    y_triton = layer_norm(x, w, b, eps)
+    y_ref = torch.nn.functional.layer_norm(x, w_shape, w, b, eps)
+
+    assert torch.allclose(y_triton, y_ref, rtol=1e-05, atol=1e-06)
```
vgokhale: Can you use `assert_close` instead of `allclose`? The former prints the largest values that mismatch each tolerance and their indices so it is easier to debug.

rahulbatra85:
@lucas-santos-amd  can you squash your commit history into just one commit. We don't need that to show in up in the main AITER branch. Thanks!

op_tests/triton/test_layernorm.py
lucas-santos-amd: Done

vgokhale (approved):

PR #124, opened 2025-02-18
[CK] [FP8] Add FP8 support to CK A8W8 GEMM

tjtanaa:
# Description
This PR refactors the CK A8W8 which initially only support INT8 A8W8 kernel to support FP8. 
It also includes updated tuning script to support tuning FP8 A8W8 Kernel.

# Caveat
The FP8 and INT8 kernel shares the same tuning filename at this moment.


op_tests/utils.py
valarLip, line 1: we have a test_common.py, move and merge this into test_common

vllmellm: How should we deal with the conflicting implementation of the `checkAllClose` method. we would like to keep the assertions inside there, but if we update the current implementation, the existing tests will throw assertion errors, instead of logging results.

Badr-Ba-Sowid: [b5477f9](https://github.com/ROCm/aiter/pull/124/commits/b5477f93ca4e00f98960cce212da873eed228a85) merged utilities to test_common.py.

vllmellm: @valarLip [b5477f9](https://github.com/ROCm/aiter/pull/124/commits/b5477f93ca4e00f98960cce212da873eed228a85) merged utilities to test_common.py

tjtanaa: @valarLip It is ready for review. Could you take a look at this PR? Thank you.

PR #129, opened 2025-02-19
Workaround for failing Fmoe-g1u1 fp8smoothquant test

jayfurmanek:
The MOE fp8 smoothquant test was failing. I traced to this tile size heuristic which returns SelectTile sizes that do not currently have an implementation.
The workaround here is to force 512 size for now but this heuristic will need to be reworked to be more flexible based on the currently available range of implementations.

### DIFF
--- a/csrc/py_itfs_cu/asm_fmoe.cpp
+++ b/csrc/py_itfs_cu/asm_fmoe.cpp
@@ -640,7 +640,8 @@ void fmoe_fp8_g1u1_a16(torch::Tensor &out,                    // [token_cnt, dim
     FMoeKernel *impl_ptr = nullptr;
     int inter_dim = down.size(2);
     int sub_X_cnt = sorted_expert_ids.size(0);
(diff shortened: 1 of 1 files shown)


csrc/py_itfs_cu/asm_fmoe.cpp
```diff
@@ -640,7 +640,8 @@ void fmoe_fp8_g1u1_a16(torch::Tensor &out,                    // [token_cnt, dim
     FMoeKernel *impl_ptr = nullptr;
     int inter_dim = down.size(2);
     int sub_X_cnt = sorted_expert_ids.size(0);
-    int selectedTile = get_heuristic_tile(inter_dim, sub_X_cnt); // todo,add tune interface here
+    //int selectedTile = get_heuristic_tile(inter_dim, sub_X_cnt); // todo,add tune interface here
+    int selectedTile = 512;
 
     if (selectedTile == 512)
```
valarLip: it's better to use ->>>> if (inter_dim %512==0)

valarLip (approved):

PR #152, opened 2025-02-25
Add Triton Quantization Kernels

rahulbatra85:


aiter/ops/triton/quant.py
```diff
@@ -0,0 +1,228 @@
+import triton
+import triton.language as tl
+import torch
+
+
+_QUANT_FLOAT_E4M3FNUZ_MAX = 240.0
```
vgokhale: Can this just be taken from finfo instead?

op_tests/triton/test_quant.py
```diff
@@ -0,0 +1,90 @@
...
+def torch_dynamic_per_tensor_fp8_quant(out, x, scale_out):
+    x_max = torch.max(torch.abs(x))
+    scale_out = x_max  / torch.finfo(torch.float8_e4m3fnuz).max
+
+    out = (x / scale_out).to(torch.float8_e4m3fnuz)    
+    
+    return out, torch.tensor([scale_out], dtype=x.dtype, device=x.device)
+
+@pytest.mark.parametrize('M, N', [(1,32), (32,32), (2,16), (10,128), (32, 8192), (93,75)])
+#@pytest.mark.parametrize('dtype', [torch.float16, torch.bfloat16, torch.float32])
```
vgokhale: May be add a TODO to track so this doesn't fall off the radar.

op_tests/triton/test_quant.py
```diff
@@ -0,0 +1,90 @@
...
+    scale = torch.randn(1, dtype=torch.float32, device='cuda')
+
+    torch_out = torch.zeros((M, N), dtype=torch.float8_e4m3fnuz, device='cuda')
+    torch_out = torch_static_per_tensor_fp8_quant(torch_out, x, scale)
+
+    triton_out = torch.empty_like(x, dtype=torch.float8_e4m3fnuz, device='cuda')
+    triton_out = static_per_tensor_fp8_quant(triton_out, x, scale)    
+
+    #Note: Torch doesn't support comparing fp8 type
+    torch.allclose(triton_out.to(dtype=torch.float16), torch_out.to(dtype=torch.float16), atol=1e-02, rtol=1e-02)
```
vgokhale: May be just compare in f32 instead? Also torch.testing.assert_close might be better.

op_tests/triton/test_quant.py
```diff
@@ -0,0 +1,90 @@
...
+
+@pytest.mark.parametrize('M, N', [(1,32), (32,32), (2,16), (10,128), (32, 4096), (1024,128), (2048,1024), (193,76), (256,13), (400,400)])
+#@pytest.mark.parametrize('dtype', [torch.float16, torch.bfloat16, torch.float32]) #TODO Fix accuracy issues with fp16 and bf16
+@pytest.mark.parametrize('dtype', [torch.float32])
+def test_dynamic_per_token_fp8_quant(M: int, N: int, dtype):
+    torch.manual_seed(20)
+    torch.set_printoptions(precision=7, threshold=4000)
+    x = torch.randn((M, N), dtype=dtype, device='cuda')
+
+    torch_scale_out = torch.zeros(M, dtype=torch.float32, device='cuda')
```
vgokhale: I don't think you need these initializations since the function returns these tensors.

op_tests/triton/test_quant.py
```diff
@@ -0,0 +1,90 @@
...
+    torch_out = torch.zeros((M, N), dtype=torch.float8_e4m3fnuz, device='cuda')
+    torch_out, torch_scale_out = torch_dynamic_per_token_fp8_quant(torch_out, x, torch_scale_out)
+
+    triton_scale_out = torch.zeros(M, dtype=torch.float32, device='cuda')
```
vgokhale, line 76: I think this can be an empty also?

op_tests/triton/test_quant.py
```diff
@@ -0,0 +1,90 @@
...
+    torch_out = torch.zeros((M, N), dtype=torch.float8_e4m3fnuz, device='cuda')
+    torch_out = torch_static_per_tensor_fp8_quant(torch_out, x, scale)
+
+    triton_out = torch.empty_like(x, dtype=torch.float8_e4m3fnuz, device='cuda')
+    triton_out = static_per_tensor_fp8_quant(triton_out, x, scale)    
+
+    #Note: Torch doesn't support comparing fp8 type
+    torch.allclose(triton_out.to(dtype=torch.float16), torch_out.to(dtype=torch.float16), atol=1e-02, rtol=1e-02) 
+
+def torch_dynamic_per_tensor_fp8_quant(out, x, scale_out):
```
valarLip: for torch ref, we need using same one to make sure aiter's function are aligned between hip/triton/asm,
for quant we have one under aiter/ops/quant.py, for test you can refer to op_tests/test_quant.py

rahulbatra85: @valarLip I agree with you that there should one common test for hip/triton/asm, but for Triton we use pytest tests. For HIP/ASM/CK the tests are using nested loops which is an inefficient way to write and execute tests. I have setup meeting with Carlus to go over this and few other things.

op_tests/triton/test_quant.py
rahulbatra85: Done

op_tests/triton/test_quant.py
rahulbatra85: Done

aiter/ops/triton/quant.py
rahulbatra85: Done

op_tests/triton/test_quant.py
rahulbatra85: For Triton, I just pass the tensors that the caller creates to Triton Kernels. Keeping it similar to how CK/ASM/HIP does quant kernels.

op_tests/triton/test_quant.py
rahulbatra85: I changed it, but I had them to make torch calls look similar to Triton

PR #154, opened 2025-02-26
split pybind module of mha

rocking5566:


aiter/jit/core.py
```diff
@@ -303,28 +303,31 @@ def convert(d_ops: dict):
 
 def compile_ops(ops_name: str, fc_name: Optional[str] = None):
     def decorator(func):
-        def wrapper(*args, **kwargs):
+        def wrapper(*args, custom_build_args=None, **kwargs):
             loadName = fc_name
             if fc_name is None:
                 loadName = func.__name__
 
+            d_args = get_args_of_build(ops_name)
```
valarLip: we must keep this in except, otherwise we will lose perf by load this everytime we run a op

rocking5566: 66fd4072122369edd527d435216c4e156ed436bb

valarLip (approved):

PR #155, opened 2025-02-26
Comment Triton RMSNorm unit tests causing the CI to fail

lucas-santos-amd:
All tests passed locally with this compiler version: triton-lang/triton@c4c8bac, but the CI is failing with the commented unit tests. So I'm putting it here until we find the reason.


vgokhale (approved):
Can you add a TODO to uncomment the test with a short note on why it was commented out?

vgokhale (approved):

PR #160, opened 2025-02-27
Triton moe kernel

rahulbatra85:


aiter/ops/triton/moe_op.py
```diff
@@ -114,13 +349,22 @@ def _fused_moe_kernel(
     offs_token = tl.load(sorted_token_ids_ptr + offs_token_id)
     token_mask = offs_token < num_valid_tokens
 
+    off_experts = tl.load(expert_ids_ptr + pid_m).to(tl.int64)
+    if off_experts == -1:
+        # -----------------------------------------------------------
+        # Write back zeros to the output when the expert is not
+        # in the current expert parallel rank.
+        _write_zeros_to_output(c_ptr, stride_cm, stride_cn, pid_n, N,
```
vgokhale, line 357: I'm not sure I understand this. Why is this needed?

rahulbatra85: This is the new code in vllm where they are sharding the experts and fill -1 during moe_align_block_size.
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/fused_moe.py#L591.

I will need to update the moe_align_block_size code as well.
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/fused_moe.py#L580

vgokhale (approved):

PR #176, opened 2025-03-05
Implement Variants of RoPE

ruanjm:
ATT. Includes:
* Rotate style: Supports both NEOX and GPT-J style. Backward is also supported by both styles.
* The size in hidden dim of `freqs`/`sin`/`cos` can be half of that of input/output tensor.
* Nope first: Rotate latter half of tensor.
* 2 channels: handle two inputs at once.
* Inplace mode: Input and output is the same. Some specific optimization is also designed for this mode.
* Get rotate angle or cos/sin via positions and offsets.

Compared with legacy implementation, the average latency in fp16 and bf16 is reduced to about 73.5% (from 37.1% in best cases to 142.5% in worst cases). Meanwhile, more functions are supported.


ruanjm:
I think there is nothing more need to be added. @carlushuang @valarLip Please review the PR.

valarLip (approved):
LGTM

PR #237, opened 2025-03-21
XCD remapping for moe

Chi-Chu319:
This pr remaps the XCD for moe, http://github.com/ROCm/triton/blob/main_perf/python/perf-kernels/gemm.py#L94

The logic:
The pid m and n are grouped to access the adjacent tiles on the same XCD so it promotes the L2 usage


rahulbatra85:
@Chi-Chu319  This looks good to me except can you clean up the commit history. I think the commit history in main would be hard to follow when your PR is merged because of the merges of main into your branch. May be they get resolved automatically. Anyways, I usually do a rebase of main onto a local branch when I need to sync up latest main. That way no issues pop up during PR merge.
Also, if it makes sense squash your two commits into one.

Thanks!

rahulbatra85 (approved):

PR #253, opened 2025-03-28
Add GEMM_A8W8_Blockscale Triton Kernel

lucas-santos-amd:
Add GEMM_A8W8_Blockscale Triton Kernel

### DIFF
--- a/aiter/ops/triton/gemm_a8w8_blockscale.py
+++ b/aiter/ops/triton/gemm_a8w8_blockscale.py
@@ -1,0 +1,270 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+
+import torch
+import triton
+import triton.language as tl
+from typing import Optional
+
+
+@triton.heuristics(
+    {
+        "EVEN_K": lambda args: args["K"] % args["BLOCK_SIZE_K"] == 0,
+        "GRID_MN": lambda args: triton.cdiv(args["M"], args["BLOCK_SIZE_M"])
+        * triton.cdiv(args["N"], args["BLOCK_SIZE_N"]),
+    }
+)
+@triton.jit
+def _gemm_a8w8_blockscale_kernel(
+    # Pointers to matrices
+    a_ptr,
+    b_ptr,
+    c_ptr,
+    a_scale_ptr,
+    b_scale_ptr,
+    # Matrix dimensions
+    M,
+    N,
+    K,
+    # The stride variables represent how much to increase the ptr by when
+    # moving by 1 element in a particular dimension. E.g. `stride_am` is
+    # how much to increase `a_ptr` by to get the element one row down
+    # (A has M rows).
+    stride_am,
+    stride_ak,
+    stride_bk,
+    stride_bn,
+    stride_cm,
+    stride_cn,
+    stride_ascale_m,
+    stride_ascale_k,
+    stride_bscale_k,
+    stride_bscale_n,
+    # Meta-parameters
+    GROUP_K: tl.constexpr,
+    GROUP_N: tl.constexpr,
+    BLOCK_SIZE_M: tl.constexpr,
+    BLOCK_SIZE_N: tl.constexpr,
+    BLOCK_SIZE_K: tl.constexpr,
+    GROUP_SIZE_M: tl.constexpr,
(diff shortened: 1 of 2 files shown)


aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,271 @@
...
+
+    NUM_XCDS: tl.constexpr = 8
+
+    tl.assume(stride_am > 0)
+    tl.assume(stride_ak > 0)
+    tl.assume(stride_bk > 0)
+    tl.assume(stride_bn > 0)
+    tl.assume(stride_cm > 0)
+    tl.assume(stride_cn > 0)
+    tl.assume(stride_cn > 0)
```
azaidy: Duplicate line.

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,271 @@
...
+            a = tl.load(a_ptrs, mask=offs_k[None, :] < K - k * BLOCK_SIZE_K, other=0.0)
+            b = tl.load(b_ptrs, mask=offs_k[:, None] < K - k * BLOCK_SIZE_K, other=0.0)
+
+        a_scale = tl.load(a_scale_ptrs)
+        b_scale = tl.load(b_scale_ptrs)
+
+        # Apply scale
+        a = a.to(a_scale_ptr.type.element_ty) * a_scale[:, None]
+        b = b.to(b_scale_ptr.type.element_ty) * b_scale[None, :]
+        accumulator += tl.dot(a, b, input_precision="ieee")
```
azaidy: I am not fully sure what code is generated but it seems like this matmul would happen at fp32 precision because both scales are fp32. This would reduce the performance gain from fp8 matmul.

azaidy: Can you perform the dot first before scaling?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,271 @@
...
+    offs_bn = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N
+    a_ptrs = a_ptr + (offs_am[:, None] * stride_am + offs_k[None, :] * stride_ak)
+    b_ptrs = b_ptr + (offs_k[:, None] * stride_bk + offs_bn[None, :] * stride_bn)
+
+    # Create pointers for the scales
+    k_start = 0
+    offs_ks = k_start // GROUP_K
+    a_scale_ptrs = a_scale_ptr + offs_am * stride_ascale_m + offs_ks * stride_ascale_k
+    offs_bsn = offs_bn // GROUP_N
+    b_scale_ptrs = b_scale_ptr + offs_ks * stride_bscale_k + offs_bsn * stride_bscale_n
```
azaidy, line 141: Are we leaving out mixed precision support for now or planning to support it in a separate kernel?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,271 @@
...
+
+    BLOCK_SIZE_M = 256
+    BLOCK_SIZE_N = 256
+    BLOCK_SIZE_K = 64
+    GROUP_SIZE_M = 4
+    waves_per_eu = 2
+    kpack = 1
+    matrix_instr_nonkdim = 16
+    num_warps = 8
+    num_stages = 2
```
azaidy, line 235: The optimal config for fp8 matmul with block scales is 128x128x128 for now. We will be tuning again in the future but might be good to set this to 128x128x128 to begin with.

azaidy (changes requested):
See comments

aiter/ops/triton/gemm_a8w8_blockscale.py
azaidy: Irrelevant. Has to be a separate kernel if at all.

aiter/ops/triton/gemm_a8w8_blockscale.py
lucas-santos-amd: You're totally right, I was focused on matching the torch reference and overlooked this.

azaidy (approved):
LGTM!

PR #262, opened 2025-04-02
[feat]: support elementwise (m,n,1) and (n,1) broadcast case

TennyWang1223:
case(m, n, 1), (n, 1), (m, 1, p, q) support, optimize naive broadcast case by using vectorization and unrolling 8


valarLip (approved):
LGTM

PR #280, opened 2025-04-09
refine aiter::mha & add benchmark

slippedJim:
1. merge bench_mha_fwd && bench_mha_fwd_splitkv into one
2. move benchmark code into op_test/cpp/mha
3. provide script to build bench_mha_fwd & bench_mha_bwd (op_test/cpp/mha/build_mha.sh)
4. rename some api 
5. add smoke test


aiter/jit/core.py
```diff
@@ -255,7 +255,7 @@ def exec_blob(blob_gen_cmd, op_dir, src_dir, sources):
                 is_python_module=False,
                 is_standalone=True,
             )
-            shutil.copy(f'{opbd_dir}/{md_name}', f'{get_user_jit_dir()}')
```
valarLip, line 258: we must use this in case aiter is installed by root, but used by none-root users

aiter/jit/core.py
```diff
@@ -241,7 +241,7 @@ def exec_blob(blob_gen_cmd, op_dir, src_dir, sources):
             f"{old_bd_include_dir}",
         ]
 
-        if md_name in ["module_bench_mha_fwd", "module_bench_mha_fwd_splitkv", "module_bench_mha_bwd"]:
+        if md_name in ["bench_mha_fwd", "bench_mha_bwd"]:
```
valarLip: add an attribute in config.json, instead of these way, so we can extend to others without chang core.py everytime

valarLip (approved):
LGTM

PR #281, opened 2025-04-09
Add Triton MoE fused with GeLU

rahulbatra85:


aiter/ops/triton/moe_op_gelu.py
```diff
@@ -0,0 +1,647 @@
...
+    tile_id2 = start_pid - NUM_SMS
+    num_valid_tiles = -1
+    while pid_m * BLOCK_SIZE_M < num_tokens_post_padded:
+        num_valid_tiles += 1
+        tile_id2 += NUM_SMS
+        group_id = tile_id2 // num_pid_in_group
+        first_pid_m = group_id * GROUP_SIZE_M
+        group_size_m = min(num_pid_m - first_pid_m, GROUP_SIZE_M)
+        pid_m = first_pid_m + ((tile_id2 % num_pid_in_group) % group_size_m)
+
```
Chi-Chu319: Would it be possible to adopt the logic I mentioned in https://github.com/ROCm/aiter/pull/284/files so we don't run into errors when group_size_m is not 1?
```
    num_tokens_post_padded = tl.load(num_tokens_post_padded_ptr)
    num_pid_m = tl.cdiv(num_tokens_post_padded, BLOCK_SIZE_M)
    num_pid_n = tl.cdiv(N, BLOCK_SIZE_N)
    tile_id = start_pid

    offs_k = tl.arange(0, BLOCK_SIZE_K)

    num_tiles = num_pid_m * num_pid_n
    # Compute how many tiles are outside the padding region
    num_valid_tiles = tl.cdiv((num_tiles - tile_id), NUM_SMS)
```

aiter/ops/triton/moe_op_gelu.py
```diff
@@ -0,0 +1,647 @@
...
+    for _ in range(0, num_valid_tiles):
+        if GROUP_SIZE_M == 1:
```
Chi-Chu319, line 391: You can add the pid remapping logic here to boost perf.

aiter/ops/triton/moe_op_gelu.py
```diff
@@ -0,0 +1,647 @@
...
+                    (offs_k[None, :] < K - k * BLOCK_SIZE_K),
+                    other=0.0)
+        b = tl.load(b_ptrs,
+                    mask=offs_k[:, None] < K - k * BLOCK_SIZE_K,
+                    other=0.0)
+        # We accumulate along the K dimension.
+        if use_int8_w8a16:
+            accumulator = tl.dot(a, b.to(compute_type), acc=accumulator)
+        elif use_fp8_w8a8:
+            if group_k > 0 and group_n > 0:
```
vgokhale: Can we make this a constexpr called blockscale or something like that?

aiter/ops/triton/moe_op_gelu.py
```diff
@@ -0,0 +1,647 @@
...
+
+    if MUL_ROUTED_WEIGHT:
+        moe_weight = tl.load(topk_weights_ptr + offs_token,
+                             mask=token_mask,
+                             other=0)
+        accumulator = accumulator * moe_weight[:, None]
+    if use_int8_w8a16:
+        accumulator = (accumulator * b_scale)
+    elif use_fp8_w8a8:
+        if group_k > 0 and group_n > 0:
```
vgokhale: Then this would go away...

aiter/ops/triton/moe_op_gelu.py
```diff
@@ -0,0 +1,647 @@
...
+            accumulator = accumulator
+        else:
+            accumulator = (accumulator * a_scale * b_scale)
+    else:
+        accumulator = accumulator
```
vgokhale: Does this need to be explicitly mentioned if using if / elif?

rahulbatra85: typo. Earlier conversion to compute type was done here. With GeLU is done in one spot.

aiter/ops/triton/moe_op_gelu.py
rahulbatra85: Done

aiter/ops/triton/moe_op_gelu.py
rahulbatra85: Done

aiter/ops/triton/moe_op_gelu.py
rahulbatra85: I pulled this in here, but didn't change the get_config in test_moe. That will still be done in your PR. If I do in this PR, then it will break other variants of moe kernel

aiter/ops/triton/moe_op_gelu.py
rahulbatra85: I will do this in another PR

vgokhale (approved):

Chi-Chu319 (approved):
LGTM!

PR #320, opened 2025-04-18
add pa qlen for mtp

junhaha666:


csrc/include/rocm_ops.hpp
```diff
@@ -43,17 +43,19 @@
...
+            py::arg("V"),                        \
+            py::arg("block_tables"),             \
+            py::arg("context_lens"),             \
+            py::arg("max_num_blocks"),           \
+            py::arg("max_qlen") = 1,             \
+            py::arg("K_QScale") = std::nullopt,  \
+            py::arg("V_QScale") = std::nullopt,  \
+            py::arg("out_") = std::nullopt,      \
+            py::arg("qo_indptr") = std::nullopt, \
             py::arg("high_precision") = 1);
```
sijiac: Thanks for adding the change. Is it possible to add a offset for each batch? it's used by local attention case, similar to leftpad_K for FA3

https://github.com/Dao-AILab/flash-attention/blob/a1be1cc38d18385fec82e2e1ee203d482c35c24c/hopper/flash.h#L78

valarLip (approved):

PR #323, opened 2025-04-19
add asser to check topk-weights dtype

sijiac:
topk-weights in dtype bf16 will cause numeric issues silently. This PR adds the check

cc @carlushuang


valarLip (approved):
LGTM

PR #325, opened 2025-04-21
[Triton] [mha] Remove if-else block in the causal loop

zhanglx13:
remove the old if condition in the causal masking and do the mask computation unconditionally. Though this will unconditionally compute mask_partial at runtime, the causal for loop does not have the if-else block any more, which helps instruction scheduling and register pressure.


vgokhale (approved):
LGTM. I verified these changes are perf neutral. Thanks!

PR #329, opened 2025-04-22
Resolve import conflicts with installed `triton` library by renaming `op_tests.triton` for modular imports

AdrianAbeyta:
Rename op_tests/triton/ --> op_tests/triton_tests/ to avoid conflicts with the installed Triton package.

Python's import system prioritizes local directories over site-packages, causing `import triton.language` to incorrectly resolve to the local op_tests/triton/ folder.

This change ensures imports correctly target the installed Triton library. No functional changes — only directory renaming and minor import path updates.


valarLip:
still some issue caused triton ut failed

AdrianAbeyta:
@valarLip updated imports in op_tests.triton_tests to use relative paths. Issue should be resolved now. Thanks for taking a look.

valarLip (approved):

PR #332, opened 2025-04-23
Asm extend chip arch

valarLip:


op_tests/test_moe_2stage.py
```diff
@@ -276,53 +295,53 @@ def test_fmoe(
...
-    )
+    # # ######################## fused 2 stage #########
+    # out2_ck, us = run_perftest(
+    #     ck_moe_2stages,
+    #     input,
+    #     w1_qt_aiter,
+    #     w2_qt_aiter,
+    #     topk_weights,
+    #     topk_ids,
+    #     quant_type=qType,
```
fxmarty-amd, line 333: Why disable? `ck_moe_2stages` is now tested nowhere and broken following https://github.com/ROCm/aiter/pull/773

PR #333, opened 2025-04-23
[Triton] [FA] Compute max after SCALE multiplication

zhanglx13:
We have the following simple flow of computation
```python
qk += tl.dot(q, k)
qk *= SCALE
m_ij = tl.maximum(m_i, tl.max(qk, 1))
p = exp2(qk - m_ij)
```
The max separates multiplication of SCALE and subtract of max apart, therefore we need both v_fma and v_sub instructions.

This PR changes the computation into
```python
qk += tl.dot(q, k)
m_ij = tl.maximum(m_i, tl.max(qk, 1))
q_shifted = qk * SCALE - (m_ij * SCALE)[:, None]
```
m_ij * SCALE is a single f_mul. Then q_shifted can be lowered to
```asm
v_fma v_qk, v_qk, s_scale, -m_ij*SCALE
```
v_sub is folded into v_fma

### DIFF
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -250,13 +250,13 @@ def _attn_fwd_inner(
             qk += (tl.dot(q, k) * descale_q * descale_k)
         else:
             qk += tl.dot(q, k)
-        qk_scaled =  qk * SM_SCALE
+
         if IS_CAUSAL:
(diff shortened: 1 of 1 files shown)


vgokhale (approved):
LGTM. lixun confirmed offline to be perf neutral.

PR #339, opened 2025-04-24
Exclude rocm 6.4+ for hipcc flag -amdgpu-coerce-illegal-types

jtang10:
Running Triton MoE code (not necessary limited to MoE) causes a compilation failure, throwing errror that `-amdgpu-coerce-illegal-types` flag for `hipcc` cannot be found.

CK uses this flag but it also checks if it is available in its own cmake file. AITER, however, only checks for `hipcc` version. Excluding version >= 6.4 fix the problem.

Using a docker from 6.5 does not have this problem at all and I don't know why it does not have issue at 6.5. 

If the reviewer knows the exact reason, I can update this PR to make it flexible

### DIFF
--- a/aiter/jit/core.py
+++ b/aiter/jit/core.py
@@ -246,7 +246,7 @@ def build_module(
                 "-mllvm",
                 "-amdgpu-function-calls=false",
             ]
-        if hip_version > Version("6.2.41133"):
+        if Version("6.4") >= hip_version > Version("6.2.41133"):
             flags_hip += ["-mllvm", "-amdgpu-coerce-illegal-types=1"]
 
         flags_cc += flags_extra_cc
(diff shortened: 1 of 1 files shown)


valarLip:
this should be a issued in rocm 6.4, without this flag we will have huge perf drop, even in rocm 6.5

jtang10:
Then is there any way we can check if the flag exists in the `hipcc` before passing it in? Because the container I used had no such flag and it simply couldn't compile.

I think CK cmake file has this check https://github.com/ROCm/composable_kernel/blob/develop/CMakeLists.txt#L258C1-L260C1 but we don't have it done in AITER.

valarLip:
we do need some check here... @amd-ruitang3 try add a flag checker here

amd-ruitang3:
hi @jtang10 , we have added checker. 
you can try [.](https://github.com/ROCm/aiter/pull/385) if work. we will merge into main and close this PR

PR #342, opened 2025-04-25
Fused multihead attention backward kernel

juuso-oskari:
This PR adds a fused version of the multihead attention backward kernel at aiter/ops/triton/mha.py. Previously we launched 2 kernels, _bwd_kernel_dkdv_causal and _bwd_kernel_dq_causal, to first compute the dk & dv and then the dq gradients (key, value and query tensor gradients). Now we fuse them all into a single kernel _bwd_kernel_dkdvdq_causal. The key idea is that we can use atomic adds to compute the dq during the inner loop of the _bwd_kernel_dkdv_causal. Because these kernels are compute-bound, the extra memory latency we get from the atomics is not that bad. And we remove extra compute of ds, which would have otherwise be necessary with the separate launch of  _bwd_kernel_dq_causal.

Some performance measures:

MI300X

image

MI350X

image

TODOs:

- ATTViewer shows unneccessary waits at the end of inner loop iterations, waiting for the atomic add writebacks:
image
This is a recurring problem with the atomics. Probably no easy fix in sight, but marking at least as a known issue that causes extra latency.
- BLK_SLICE_FACTOR=1 performs better, but produces mismatches on tests. It shouldn't, so go through the corresponding code.

### DIFF
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -1496,7 +1496,6 @@ def _bwd_kernel_dkdvdq_causal(
 
     start_m += num_steps * MASK_BLOCK_M
     num_steps = tl.cdiv(seqlen_q - start_m, BLOCK_M)
-    end_m = start_m + num_steps * BLOCK_M
 
     
 
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -1148,8 +1148,6 @@ def _bwd_dkdvdq_inner(
     curr_m = start_m
     step_m = BLOCK_M
     curr_philox_offset = batch_philox_offset
-    curr_dropout_offset = dropout_offset
-    RCP_LN2: tl.constexpr = 1.4426950408889634
 
     #Iterate over blocks(BLOCK_M size) of Q while calculating 
     #a fixed block(BLOCK_N) of dk and dv. Note, during backward
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -1377,7 +1377,6 @@ def _bwd_kernel_dkdvdq_causal(
     delta_qk = seqlen_q - seqlen_k
 
     # q > k: diretcly skip all the way until the start of causal block
-    start_delta_q_gt_k = delta_qk
 
     # q < k: some blocks will have no Masked block, other needs to re-calc
     # starting position
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -221,24 +221,17 @@ def _attn_fwd_inner(
         # We start from end of seqlen_k so only the first iteration would need
         # to be checked for padding if it is not a multiple of block_n
         # TODO: This can be optimized to only be true for the padded block.
-        mask = tl.full([BLOCK_M, BLOCK_N], True, dtype=tl.int1)
         if MASK_STEPS:
             # If this is the last block / iteration, we want to
             # mask if the sequence length is not a multiple of block size
             # a solution is to always do BLOCK_M // BLOCK_N + 1 steps if not is_modulo_mn.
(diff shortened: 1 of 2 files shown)


juuso-oskari:
No merging yet, because automatic merge misses some things from the main

3rdparty/composable_kernel
rahulbatra85, line 1: Can you make sure to update the submodule locally before pushing your change?

aiter/ops/triton/mha.py
```diff
@@ -1110,6 +1099,442 @@ def _bwd_dkdv_inner(
...
+    
+    qT_ptrs_start = Q + offs_m[None, :] * stride_q_m + offs_k[:, None] * stride_q_k #[BLOCK_D_MODEL_POW2, BLOCK_M]
+    dq_ptrs_start = DQ + offs_m[:, None] * stride_dq_m + offs_k[None,:] * stride_dq_k #[BLOCK_M, BLOCK_D_MODEL_POW2]
+    
+    do_ptrs_start = DO + offs_m[:, None] * stride_do_m + offs_k[None,: ] * stride_do_k
+    curr_m = start_m
+    step_m = BLOCK_M
+    curr_philox_offset = batch_philox_offset
+    curr_dropout_offset = dropout_offset
+    RCP_LN2: tl.constexpr = 1.4426950408889634
```
Chi-Chu319:
```suggestion
```

aiter/ops/triton/mha.py
```diff
@@ -1110,6 +1099,442 @@ def _bwd_dkdv_inner(
...
+    # Figure out causal starting block since we have seqlen_q >=< seqlen_k.
+    # Unlike forward pass where we tile on M dim and iterate on N dim, so that
+    # we can skip some M blocks, in backward pass, we tile on the N dim for kv
+    # and iterate over the M. In this way, we cannot skip N blocks, but only to
+    # determine the starting M blocks to skip some initial blocks masked by
+    # causal.
+    delta_qk = seqlen_q - seqlen_k
+
+    # q > k: diretcly skip all the way until the start of causal block
+    start_delta_q_gt_k = delta_qk
```
Chi-Chu319:
```suggestion
```

aiter/ops/triton/mha.py
```diff
@@ -1110,6 +1099,442 @@ def _bwd_dkdv_inner(
...
+        IS_FP8=IS_FP8,
+        FP8_MAX=FP8_MAX,
+        workgroup_id=seq_k_blk_idx,
+        num_atomics_concurrent=num_atomics_concurrent
+    )
+
+
+    start_m += num_steps * MASK_BLOCK_M
+    num_steps = tl.cdiv(seqlen_q - start_m, BLOCK_M)
+    end_m = start_m + num_steps * BLOCK_M
```
Chi-Chu319:
```suggestion
```

op_benchmarks/triton/bench_mha.py
```diff
@@ -8,15 +9,11 @@
...
 import argparse
 
-# Add two parent directories to sys.path
-current_dir = os.path.dirname(os.path.abspath(__file__))
-parent_dir = os.path.dirname(os.path.dirname(current_dir))
-sys.path.append(parent_dir)
-
-from aiter.ops.triton.mha import flash_attn_varlen_func, flash_attn_func, flash_attn_varlen_fp8_func, flash_attn_fp8_func
+from aiter.ops.triton.mha import flash_attn_func, flash_attn_fp8_func, flash_attn_varlen_func, flash_attn_varlen_fp8_func
+from aiter.test_mha_common import attention_ref, generate_random_padding_mask, generate_qkv, pad_rearrange_dropout_mask_hts_to_bhss
```
Chi-Chu319:
```suggestion
from aiter.test_mha_common import attention_ref, generate_random_padding_mask, generate_qkv
```

PR #347, opened 2025-04-27
[MoE] refine moe ep test

yyccli:
This test mainly simulates the ep test on a GPU with gpu rank `ep-1`, so the mask should set the **last ep size** experts to 1. In addition, when constructing the weight, each GPU should only need to allocate the weight of **ep size experts**.


valarLip (approved):
LGTM

PR #353, opened 2025-04-28
[TRITON][REFACTOR]: Refactored Triton MoE 

jtang10:
Refactored Triton MoE to enable extendibility.


Chi-Chu319 (dismissed):
Looks good. But please fix the checks

aiter/ops/triton/utils/quantization.py
rahulbatra85, line 1: All these quantize functions are torch implementation, but the code in aiter/ops/triton folder should be mostly Triton related. As I understand these are utility/helper functions for tests and benchmarking. Can we move this to op_benchmarks/triton/utils or op_tests/triton_tests/utils? 
The only exception should be anything required for moe config.

aiter/ops/triton/utils/moe_utils.py
rahulbatra85, line 1: These functions are again for testing and input generation and not associated with the Triton code ie the op user will call. Can we move this to op_benchmarks/triton/utils or op_tests/triton_tests/utils?

The only exception should be anything required for moe config.

rahulbatra85:
Completed as part of another PR.

PR #361, opened 2025-04-29
Add leanAttention op and test

valechen:
This is the first turn-in of leanAttention op and test. LeanAttention is a streamK style hardware-aware scalable attention mechanism. It's primarily for decode attention, although this implementation has extended it to prefill/causal=1. The core algorithm is there but expecting interface changes needed to meet AITER requirements.


rahulbatra85:
Duplicated in https://github.com/ROCm/aiter/pull/376

PR #367, opened 2025-05-01
Revert "Compute max after multiplying SCALE (#333)"

rahulbatra85:
This reverts commit 70ae241a4d58ce4c05ab6e9ba37838923623e2b5.

The computation in the commit for mi_j are not mathematically equivalent and causes some UTs to fail


vgokhale (approved):

zhanglx13:
@rahulbatra85 Do we know if SCALE is guaranteed to be positive? If so, why is it not mathematically equivalent?

PR #372, opened 2025-05-04
formating the code for aiter thorugh ruff

okakarpa:


rahulbatra85:
Duplicated here 
https://github.com/ROCm/aiter/pull/397

PR #374, opened 2025-05-05
Lean Attention

alexdutu:
Adding a Triton implementation of Lean Attention.


rahulbatra85:
duplicated in https://github.com/ROCm/aiter/pull/361

alexdutu:
I have updated PR 361, so this is not needed anymore.

https://github.com/ROCm/aiter/pull/361

PR #385, opened 2025-05-07
mdf_1

amd-ruitang3:


aiter/jit/core.py
```diff
@@ -134,6 +134,19 @@ def validate_and_update_archs():
     ), f"One of GPU archs of {archs} is invalid or not supported"
     return archs
 
+@functools.lru_cache()
+def hip_flag_checker(flag_hip : str):
+    with open(f"{bd_dir}/tmp.hip", 'w') as f:
```
valarLip: never R/W a file in critical path...

valarLip (changes requested):

aiter/jit/core.py
amd-ruitang3: Modified~. now using pipe and -fsyntax-only to avoid R/W files.

valarLip (approved):

PR #387, opened 2025-05-07
[TRITON] Causal Attention XCD balancing and MHA Scale

Chi-Chu319:
Authors: @juuso-oskari, @Chi-Chu319 

This PR is based on the following issue: https://github.com/issues/assigned?issue=ROCm%7Ctriton-internal%7C771. It fixes the grid traversal in order to have good workload balance between XCDs, and L2 caching for consecutive sequence blocks and grouped query heads.

The current grid traversal order (batch, head q, seqlen) essentially handles workload balancing between XCDs (round robin of assigning workgroups to XCDs increments batch id and head q id before seqlen id => assigned workgroups have most likely the same workload as workload only changes based on seqlend id). This brings benefit especially with the "bshd" layout, where we might otherwise have significantly imbalanced workloads between XCDs.

But this traversal order significantly reduces L2 caching (observed as lowered L2 cache hit rate) because consecutive seqlen blocks are unlikely to be assigned concurrently to the same XCD. This hurts the performance with the "thd" layout, where we don't even benefit from the load balancing, because the variable sequence lengths already take care of that.

L2 hit rate of the flipped grid
model | batch | (seqlen q, head q, batch) | (batch, head q, seqlen q)
-- | -- | -- | -- 
llama3-8B | 1 | 51.56% | 20.94% 
llama3-8B | 2 | 62.33% | 22.25%
llama3-8B | 4 | 64.96% | 46.07%
llama3-8B | 8 | 55.07% | 58.09%
llama3-8B | 16 | 53.72% | 58.84%

The fix has the following idea: Do the round robin of workgroup launches over heads q dim (so we achieve workload balance because the seqlen block id does not change), and then increase the seqlen block id for a next "round robin" round, which also achieves that the workgroups that get launched at, lets say XCD 0, process consecutive seqlen blocks.

We also promote grouped query heads (q heads that share the same k head) to be placed on the same XCD by remapping the computed q head indices.

The performance:
image
image


rahulbatra85:
@Chi-Chu319  can you please rebase to main branch and resolve conflicts

vgokhale (approved):

xinyazhang:
@vgokhale  @zhanglx13 The FMA optimization of MHA Scale has a known numerical issue: https://github.com/ROCm/aotriton/issues/54
https://github.com/ROCm/triton-internal/issues/325

PR #390, opened 2025-05-08
tuned gemm hotfix

junhaha666:


aiter/tuned_gemm.py
```diff
@@ -168,7 +168,7 @@ def apply_rocb_mm(
         scale_c=None,
     ):
         assert (
-            scale_a != None and scale_b != None and scale_c != None
+            scale_a == None and scale_b == None and scale_c == None
```
valarLip: use 'is' or 'is not', instead of '=='

valarLip (changes requested):

valarLip (approved):

PR #391, opened 2025-05-08
Torch implementation of silu and mul

Chi-Chu319:
Torch implementation of silu and mul.


rahulbatra85:
Merged into this 
https://github.com/ROCm/aiter/pull/398

PR #397, opened 2025-05-08
AITER Ruff Fixes

rahulbatra85:


valarLip (dismissed):

valarLip:
we failed a number of tests in Aiter CI (Standard-Tests)...

rahulbatra85:
@valarLip  Looks like running black on non-triton stuff causes tests to fail. As you can see there are 2 commits in this PR. One for Triton stuff and other non-triton. If I just run with Triton stuff locally, then triton tests and other tests run fine.

vgokhale (approved):

PR #409, opened 2025-05-12
Fix c++ api use case

slippedJim:
- Get AITER_ROOT_DIR while aiter is not installed
- skip import module while not is_python_module


valarLip (approved):
LGTM

PR #434, opened 2025-05-16
stride 64

micmelesse:


aiter/ops/triton/mha.py
```diff
@@ -1,9 +1,17 @@
...
 import triton.language as tl
 
 from typing import Optional, Tuple
 
+# env vars
+USE_INT64_STRIDES = os.environ.get("USE_INT64_STRIDES", "1").lower() in (
+    "1",
+    "true",
+    "yes",
+)
```
rahulbatra85: @micmelesse  We don't want to have env vars in AITER. Can you please set USE_INT64_STRIDES=0(default value) and then add a function to override/set it.  This way the user of the op(like VLLM framework, etc) can set it. If they want to have it controlled through env var, then they can add it on their end.

By the way, we use a similar approach for MoE op.

op_tests/triton_tests/test_mha.py
```diff
@@ -182,6 +182,100 @@ def test_mha(
...
+@pytest.mark.parametrize("BATCH", [1])
+@pytest.mark.parametrize(
+    "SEQLEN_Q, SEQLEN_K",
+    [(1, 1)],
+)
+@pytest.mark.parametrize("NUM_Q_HEADS, NUM_K_HEADS", [(128, 8)])
+@pytest.mark.parametrize("HEAD_SZ", [128])
+@pytest.mark.parametrize("CAUSAL", [True])
+@pytest.mark.parametrize("DROPOUT", [0.0])
+def test_mha_int64_strides(
```
rahulbatra85, line 196: Do we need a whole new test? Is it possible to add it one of the existing ones and add parameter to test int64 strides?
Also, please use the function that I mentioned in my above comment set the USE_INT64_STRIDES.

Thanks!

aiter/ops/triton/mha.py
micmelesse: I can remove the env variable

op_tests/triton_tests/test_mha.py
micmelesse: I think a seperate test make sense for this. Having a simple focused test seems better them than introducing a bunch of branching to already existing tests with multiple cases. This issue occurs in very specifc situations

rahulbatra85 (approved):

PR #448, opened 2025-05-20
[Fix] aiter installed/usable with gfx950

fxmarty-amd:
As per title.

Aiter should be installable/usable with `GPU_ARCHS="gfx950"`.


fxmarty-amd:
@valarLip done

valarLip (approved):

PR #451, opened 2025-05-20
fix typo in build script

dtrifiro:
`exclue` -> `exclude`


valarLip (approved):

dtrifiro:
@valarLip it seems that the vLLM benchmark is failing (?) but I have no access to Jenkins and I cannot debug it. I'd say that's unrelated though

PR #463, opened 2025-05-22
fix: add missing column `cu_num` for `gemm_a8w8_blockscale_tune.py`

gau-nernst:
Per title. When `tuned_gemm_file` does not exist, the returned dataframe is missing column `cu_num`, which is present in `a8w8_blockscale_tuned_gemm.csv`. This makes the script fails to run.

This PR fixes the issue.


valarLip:
thank you

valarLip (approved):

PR #466, opened 2025-05-23
Add aiter namespace to avoid symbol conflict with vllm

xw285cornell:
As title - those symbols are in vllm too which will cause symbol conflict if compile with vllm.


valarLip:
looks like CI stucked..

okakarpa:
> looks like CI stucked..

track this https://ml-ci-internal.amd.com/blue/organizations/jenkins/aiter%2Faiter-builds/detail/PR-466/6/pipeline/181

PR #493, opened 2025-05-29
Sync batch_prefill optimization for low CU utilization

poyenc:


HaiShaw:
@poyenc @valarLip can we have a look for merge?

PR #517, opened 2025-06-06
Batra/fused act mxfp4 quant

rahulbatra85:


rahulbatra85:
Merged in main through Cagri's PR

Issue #519, opened 2025-06-08
[Issue]: ambiguous block_allreduce in csrc/kernels/cache_kernels.cpp


xw285cornell:
### Problem Description

csrc/kernels/cache_kernels.hip:816:30: error: call to 'block_reduce' is ambiguous
  816 |     int numtokens_in_block = block_reduce(tokens_in_block, [](float a, float b) { return a + b; });
      |                              ^~~~~~~~~~~~
csrc/kernels/cache_kernels.hip:1453:13: note: in instantiation of function template specialization 'aiter::reshape_and_cache_with_block_quant_kernel_for_asmpa<float, _BitInt(8), float, true, 256>' requested here
 1453 |             CALL_RESHAPE_AND_CACHE_WITH_BLOCK_QUANT_FOR_ASMPA(
      |             ^
csrc/kernels/cache_kernels.hip:1204:36: note: expanded from macro 'CALL_RESHAPE_AND_CACHE_WITH_BLOCK_QUANT_FOR_ASMPA'
 1204 |        hipLaunchKernelGGL(( aiter::reshape_and_cache_with_block_quant_kernel_for_asmpa<KV_T,                           \
      |                                    ^
csrc/include/hip_reduce.h:105:24: note: candidate function [with T = int, F = (lambda at csrc/kernels/cache_kernels.hip:816:60), BlockSize = 256, waveBroadcast = true]
  105 | __device__ constexpr T block_reduce(T local, F reduce_op)
      |                        ^
csrc/kernels/cache_kernels.hip:26:24: note: candidate function [with T = int, F = (lambda at csrc/kernels/cache_kernels.hip:816:60)]
   26 | __device__ constexpr T block_reduce(T val, F reduce_f)
      |                        ^

### Operating System

centos

### CPU

AMD

### GPU

MI300X

### ROCm Version

Rocm 6.2

carlushuang:
https://github.com/ROCm/aiter/pull/523 <= should be fixed here

PR #520, opened 2025-06-08
Fixes on typename and ambiguous func

xw285cornell:
Fixing https://github.com/ROCm/aiter/issues/519


valarLip:
formal fix in https://github.com/ROCm/aiter/pull/523

PR #522, opened 2025-06-09
[fix]: replace static_cast<float>(half)

TennyWang1223:
convert dtype between half and float by using __half2float and __float2half


carlushuang (approved):
LGTM

PR #525, opened 2025-06-09
Add gfx950 support in asm_moe

sogalin:
Add gfx950 support for ASM MoE solution.


valarLip:
please stop update aiter/fused_moe_bf16_asm.py, and switch to use aiter/fused_moe.py

PR #549, opened 2025-06-16
Find executable in rocm home when not found in PATH

xli:
In my production environment, we don't have rocm/bin in PATH, but the rocm is installed at the place _find_rocm_home can find it.
So add executable_path function to lookup the executable in ROCM/bin dir when it is not found in the PATH

### DIFF
--- a/aiter/jit/utils/chip_info.py
+++ b/aiter/jit/utils/chip_info.py
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
-import os
 import functools
+import os
 import subprocess
 
-from aiter.jit.utils.cpp_extension import executable_path
+from cpp_extension import executable_path
 
 
 @functools.lru_cache(maxsize=1)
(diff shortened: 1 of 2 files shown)


aiter/jit/utils/chip_info.py
```diff
@@ -4,15 +4,16 @@
 import functools
 import subprocess
 
+from aiter.jit.utils.cpp_extension import executable_path
```
valarLip: could you please do anything like "from aiter.xxx import" in core.py path... this will break our PREBUILD path where AITER is not fully installed

xli: Don't understand the recommendation, do you mean we do in core.py:

      from cpp_extension import executable_path

then here we use: 

        from aiter.jit.core import executable_path

xli: Another option is, if you share with me how to test PREBUILD path, I can try it out to confirm it works. Is it "pip install ." with PREBUILD_KERNELS=1?

valarLip: > Another option is, if you share with me how to test PREBUILD path, I can try it out to confirm it works. Is it "pip install ." with PREBUILD_KERNELS=1?

yes, something like "PREBUILD_KERNELS=1 python setup.py develop"

xli: @valarLip  all tests passed, is it good enough for accepting this PR? I did update the import code to relative to jit/utils dir like cpp_extension.py did to import file_baton.

I'm still trying to run PREBUILD_KERNELS=1 python setup.py develop, but got different problems here and there, it will take me a while to fix my local env to run it. 
But I think the change should be good to go.

valarLip (approved):

PR #552, opened 2025-06-17
[update]: update all-reduce

TennyWang1223:
replace static_cast as safe function
add template param: vectorialize size, ngpus, dtype
add dispatch logic


valarLip:
https://github.com/ROCm/aiter/blob/72019bbb70662264c82649ef103cd5a6eaad7124/csrc/kernels/cache_kernels.cu#L242
let's transfer to ck_tile type_convert

csrc/include/custom_all_reduce.cuh
```diff
@@ -479,101 +533,152 @@ namespace vllm
...
+    return packDowncast<T, pack_size>(ret_val);
+  }
 
-      // load quant rslt
-      scale_factor = *(reinterpret_cast<half*>(&tmps[device_id][part_size / 8]) + index);
-      half8_reg = packDequant<hip_fp8, half, 8>(tmps[device_id][index], scale_factor);
+  // bf16 quant fp8 kernel function
+  // too slow need to be optimized
+  // fp16
+#define FP8_UPBOUND 240.0f
```
valarLip: don't hard code this one... do someting like this, we will have different fp8 with different max value
https://github.com/ROCm/aiter/blob/d765e80ddf65891e41bd608e01c3d4be3ee6cb83/csrc/kernels/quant_kernels.cu#L90

PR #573, opened 2025-06-23
[TRITON]: GEMM updates for DS

rahulbatra85:


k50112113:
So it seems that here is a file called `aiter/ops/triton/gemm_afp4wfp4_pre_quant_atomic.py` for DS but it's not here, are we not merging this one here?

rahulbatra85:
> So it seems that here is a file called `aiter/ops/triton/gemm_afp4wfp4_pre_quant_atomic.py` for DS but it's not here, are we not merging this one here?

@k50112113 There is no test case for that. I am working on adding that and will push once it's done

k50112113 (commented):
looks good to me, I think there are still some reviewdog failing, would be great if you can fix that, thx

k50112113 (approved):

PR #580, opened 2025-06-25
merge moe from 350 launch

lalala-sh:
1. for gfx950, enable new MFMA instructions, inlcude fp16/bf16/fp8/fp8/i4
2. switch to new moe api
3. add tile dispatch for gfx942


HaiShaw:
@valarLip 
Can we have some description on details of further MoE optimizations from this?
Like: ISA update, tile, more kernel selections?
Other - what about MI300?

lalala-sh:
> Can you add description/summary to PR, for major ones like this? Thanks.

Sure, I'll add summary here

op_tests/test_moe_2stage_ck_fp4.py
valarLip, line 1: why we have this one....

csrc/py_itfs_ck/moe_sorting_kernels.cu
```diff
@@ -46,6 +46,7 @@ void moe_sorting_fwd(torch::Tensor &topk_ids,          // [m, topk]
                 {topk_ids.data_ptr(),     // p_topk_ids
                  topk_weights.data_ptr(), // p_weights
                  local_expert_mask.has_value() ? local_expert_mask.value().data_ptr() : nullptr,
+                 nullptr,                      // p_local_tokens
```
valarLip: we have formal support for new api, don't do this

valarLip (changes requested):

op_tests/test_moe_2stage_ck_fp4.py
lalala-sh: asm needs weight_shuffle, but ck don't, so i create a ut file to test ck... Do you want to merge these two in the same ut?

csrc/py_itfs_ck/moe_sorting_kernels.cu
lalala-sh: what should i do here? seems the support don't merge into mainline.

op_tests/test_moe_2stage_ck_fp4.py
lalala-sh: removed

csrc/py_itfs_ck/moe_sorting_kernels.cu
lalala-sh: checkout into main branch update here.

PR #582, opened 2025-06-26
Fix rmsnorm & layernorm strides

MHYangAMD:
We found the strides are all 1 in CK. It should be **n** here.


csrc/py_itfs_ck/norm_kernels.cu
```diff
@@ -22,9 +22,9 @@ void layernorm2d(torch::Tensor &out,    // [m, n]
     std::string dtype_str = torchDTypeToStr(dtype);
     int n = input.size(-1);
     int m = input.numel() / n;
-    int stride = input.stride(0);
+    int stride = n;
```
valarLip, line 25: nerver do this.... our input might be none-contiguous

MHYangAMD:
It's my environment issue.

PR #588, opened 2025-06-26
[TRITON]: Quantization updates(add int8 and use common fp8 dtypes)

rahulbatra85:
1. Add int8 quantization ops
2. Add doc
3. Fix per_token_fp8 testcase

### DIFF
--- a/aiter/ops/triton/quant.py
+++ b/aiter/ops/triton/quant.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT
-# Copyright (c) 2024, Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (c) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
 
 import triton
 import triton.language as tl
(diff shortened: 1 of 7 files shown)


aiter/ops/triton/quant.py
```diff
@@ -75,49 +102,91 @@ def _dynamic_per_tensor_fp8_quant_kernel(
...
+    cols = x_in.shape[1]
+    NUM_COL_POW2 = triton.next_power_of_2(cols)
+    grid = lambda meta: (rows,)  # noqa: E731
+    _dynamic_per_tensor_quant8_kernel[grid](
+        x_in,
+        scale_out,
+        cols,
+        x_in.stride(0),
+        NUM_COL_POW2=NUM_COL_POW2,
+        DTYPE_MAX=torch.iinfo(qx.dtype).max,
```
valarLip: why we keep two api, if the only diff is DTYPE_MAX between dynamic_per_tensor_int8_quant/dynamic_per_tensor_fp8_quant

rahulbatra85: That's what I did first, but then didn't want to change the existing API for fp8. I guess the wrapper you have doesn't care, so I will make it one API.

valarLip (approved):

PR #595, opened 2025-06-30
[TRITON]: add json config and refactor

rahulbatra85:


aiter/ops/triton/mha_onekernel_bwd.py
```diff
@@ -1525,6 +1529,19 @@ def is_contiguous(x, name):
...
+def _get_config():
+    if not hasattr(_get_config, "_config_dict"):
+        dev = arch_info.get_device()
+        _get_config._config_dict = {}
+        fpath = f"{AITER_TRITON_CONFIGS_PATH}/{dev}-MHA-DEFAULT.json"
+        with open(fpath, "r") as file:
+            config = json.load(file)
+        _get_config._config_dict = config
+
+    return _get_config._config_dict["bkwd_onekernel"]
```
Chi-Chu319, line 1542: Would it be better to move this in some utils file? so it would be like

```
@functools.lru_cache(maxsize=1024)
def _get_config(config_suffix, kernel_name):
    if not hasattr(_get_config, "_config_dict"):
        dev = arch_info.get_device()
        _get_config._config_dict = {}
        fpath = f"{AITER_TRITON_CONFIGS_PATH}/{dev}-{config_suffix}"
        with open(fpath, "r") as file:
            config = json.load(file)
        _get_config._config_dict = config

    return _get_config._config_dict[kernel_name]
```

rahulbatra85: There is some common code among ops, but different ops have config files structured differently. For example, GEMMs have multiple files which are keyed by N. MHA have forward and backward kernel config in one single file.  Also, logic to select a config from the file is op dependent. For now, to keep things simple, I think it's best to keep a function like this in the op file itself and let the performance optimization/tuning not affect other ops. May be later on once all this is a bit more stable, we can look into refactoring

Chi-Chu319: Cool, that sounds good to me

aiter/ops/triton/extend_attention.py
```diff
@@ -341,8 +346,8 @@ def extend_attention_fwd(
         BLOCK_DPE = 0
     BLOCK_DV = triton.next_power_of_2(Lv)
 
-    BLOCK_M, BLOCK_N = (64, 64)
-    num_warps = 4
+    # BLOCK_M, BLOCK_N = (64, 64)
+    # num_warps = 4
```
Chi-Chu319, line 350: We could probably remove the commented out configs since they are not in used?

rahulbatra85: Good catch, I will clean this up in the next PR.

PR #597, opened 2025-07-01
[TRITON]: Standardize GEMM weight shape to (N, K) and TN memory layout (by default)

willzhou-amd:
Changes:
- [x] Fix .assume() bug that causes M=1 cases to fail in some kernels
- [x] Add test cases for M = 1 to all tests
- [x] Standardize weight shapes to (N, K)

### DIFF
--- a/aiter/ops/triton/mha.py
+++ b/aiter/ops/triton/mha.py
@@ -24,7 +24,7 @@ def mha_set_use_fused_bwd_kernel(value: bool):
     _USE_FUSED_BWD_KERNEL = value
 
 
-_USE_INT64_STRIDES = False
+_USE_INT64_STRIDES = True
 
 
 def mha_set_use_int64_strides(value: bool):


aiter/ops/triton/batched_gemm_a8w8.py
```diff
@@ -108,14 +108,11 @@ def _batched_gemm_a8w8_kernel(
...
         pid_n = (pid % num_pid_in_group) // group_size_m
 
-    tl.assume(pid_m > 0)
-    tl.assume(pid_n > 0)
+    tl.assume(pid_m >= 0)
+    tl.assume(pid_n >= 0)
 
     # Cast batch id and batch dimension strides to int64 to avoid int32 overflow during offset calculation
     batch_id = batch_id.to(tl.int64)
-    stride_ab = stride_ab.to(tl.int64)
```
rahulbatra85, line 116: Add this back

willzhou-amd:
There's a gnarly issue where Triton will implicitly promote small integers that it believes to be constants to `tl.constant()`. In some cases where the dimensions are small, typecasting arguments to your kernel will yield a cast error:
```shell
    batch_id = batch_id.to(tl.int64)
    stride_ab = stride_ab.to(tl.int64)
                ^
AttributeError("'constexpr' object has no attribute 'to'")
```
Replacing `.to()` with `tl.cast()` prevents this implicit promotion. I'm uncertain whether this happens on other versions of Triton, but this is reliably replicable on Triton 3.3.1.

tldr: If you're attempting to cast strides to int64 to prevent integer overflow, use `tl.cast`.

rahulbatra85 (approved):

PR #603, opened 2025-07-02
add ck gemm a4w4 blockscale with splitK support

ukannika-amd:


valarLip (dismissed):
Hi Uma, could  you please add the kernel instance format back as we talk offline, thanks

valarLip (approved):

PR #610, opened 2025-07-03
Moe tuning mi350

Chi-Chu319:
- Added basic config for mi350
- Fix config loading
- Added various support for benchmark for future tuning

### DIFF
--- a/aiter/ops/triton/configs/moe/MI350X-MOE-DEFAULT.json
+++ b/aiter/ops/triton/configs/moe/MI350X-MOE-DEFAULT.json
@@ -1,0 +1,35 @@
+{
+  "small_M": {
+    "BLOCK_SIZE_M": 64,
+    "BLOCK_SIZE_N": 64,
+    "BLOCK_SIZE_K": 64,
+    "GROUP_SIZE_M": 4,
+    "num_warps": 8,
+    "num_stages": 2,
+    "waves_per_eu": 0,
+    "matrix_instr_nonkdim": 16,
+    "kpack": 1
+  },
+  "medium_M": {
+    "BLOCK_SIZE_M": 128,
+    "BLOCK_SIZE_N": 128,
+    "BLOCK_SIZE_K": 128,
+    "GROUP_SIZE_M": 8,
+    "num_warps": 8,
+    "num_stages": 2,
+    "waves_per_eu": 0,
+    "matrix_instr_nonkdim": 16,
+    "kpack": 1
+  },
+  "large_M": {
+    "BLOCK_SIZE_M": 256,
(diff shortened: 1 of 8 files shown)


aiter/ops/triton/moe_op_gelu.py
```diff
@@ -43,12 +43,6 @@ def moe_set_quant_func(func):
...
 
 
-@triton.heuristics(
-    {
-        "GRID_MN": lambda args: triton.cdiv(args["EM"], args["BLOCK_SIZE_M"])
-        * triton.cdiv(args["N"], args["BLOCK_SIZE_N"])
-    }
-)
 @triton.jit
 def _fused_moe_kernel(
```
rahulbatra85: Does the EVEN_K optimization not help with this kernel?

aiter/ops/triton/moe_op.py
```diff
@@ -212,12 +208,17 @@ def _fused_moe_kernel_gptq_awq(
...
 
-        a = tl.load(
-            a_ptrs,
-            mask=token_mask[:, None] & (offs_k[None, :] < K - k * BLOCK_SIZE_K),
-            other=0.0,
-        )
-        b = tl.load(b_ptrs)
+            if EVEN_K:
+                a = tl.load(a_ptrs, mask=token_mask[:, None], other=0.0)
+                b = tl.load(b_ptrs)
```
rahulbatra85: load of b is not dependent on EVEN_K. May be we can move it outside the if-else?

aiter/ops/triton/moe_op.py
```diff
@@ -429,12 +434,17 @@ def _fused_moe_persistent_kernel_gptq_awq(
                 k_mask = None
                 k_other = None
 
-            a = tl.load(
-                a_ptrs,
-                mask=token_mask[:, None] & (offs_k[None, :] < K - k * BLOCK_SIZE_K),
-                other=0.0,
-            )
-            b = tl.load(b_ptrs)
+            if EVEN_K:
```
rahulbatra85: Same as above

aiter/ops/triton/moe_op_gelu.py
Chi-Chu319: Ah just missed adding it there

aiter/ops/triton/moe_op.py
Chi-Chu319: On it

Chi-Chu319: We should rather add the k mask for the else branch I think? I will do it that way

Chi-Chu319: Just realized we had block_k_diviable. I integrated with it

PR #623, opened 2025-07-07
Update dependencies and add backup for unknown hw

kunaltyagi:


aiter/utility/dtypes.py
```diff
@@ -9,15 +9,17 @@
     "gfx950": {"fp8": torch.float8_e4m3fn},
 }
 
+_f8_fallback = torch.uint8
```
valarLip: i would like name it _8bit_fallback as we also it for all fp4/fp8

requirements.txt
```diff
@@ -1,5 +1,7 @@
 pandas
 pytest
 psutil
+matplotlib
```
valarLip, line 4: where is this one used for

valarLip (changes requested):

requirements.txt
kunaltyagi: [op_tests/op_benchmarks/triton/bench_topk.py:11](https://github.com/ROCm/aiter/blob/f4d118998885dafea44f09f1425dceff9761f5ae/op_tests/op_benchmarks/triton/bench_topk.py#L11)

kunaltyagi:
can someone re-trigger the CI?

valarLip (approved):

PR #627, opened 2025-07-08
Fix Precision Issue in RoPE Tests

ruanjm:
Kernel is correct but there is precision issue in reference code:
1. All the computation in kernel is conducted with fp32 but with fp16 or bf16 in reference code. 
2. In the case of cached cases, the pre-computated cos/sin are truncated. Reference code need to simulate such case.

Besides, this PR also move the code about RoPE to a new place.


ruanjm:
Looks the failures come from format. How can I fix them?

csrc/kernels/rope/general_bwd_kernels.cu
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
 
-#include "rope_common.h"
+#include "common.h"
```
valarLip: let's keep use rope_common.h...

valarLip (changes requested):

csrc/kernels/rope/general_bwd_kernels.cu
ruanjm: ok

valarLip (dismissed):

valarLip (approved):
LGTM

PR #647, opened 2025-07-11
Add option to choose between CK RMSNorm pipelines

ClementLinCF:
We have integrated the new option of `rmsnorm2d_fwd_traits` from the CK  to support a new RMSNorm pipeline. This commit focuses on this integration, allowing users to select between the new T5-like RMSNorm implementation and the no specific version. This enhancement aims to improve model accuracy while providing flexibility in applying RMS normalization techniques.


valarLip (approved):
LGTM

PR #652, opened 2025-07-12
Optimize topksoftmax WARPS_PER_TB for higher occupancy and remove redundant precision conversion

CuiCu-618:
**Description:**
This PR improves the `topksoftmax` kernel by:

* Tuning `WARPS_PER_TB` to boost occupancy and throughput.
* Eliminating the explicit precision conversion step before launch, since the kernel now handles all necessary type casting internally (significant performance gains on bf16 and fp16).

*No functional changes.*

**Note:** The planned topk optimization was not included because it introduced precision issues that would require additional handling—and thus too much overhead. @valarLip @junhaha666

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -911,7 +911,7 @@ def fused_topk(
         topk_weights,
         topk_ids,
         token_expert_indicies,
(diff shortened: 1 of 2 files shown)


valarLip (approved):
LGTM

PR #703, opened 2025-07-23
mla asm inline & dynamic splitkv

Zzz9990:
python3 op_test/test_mla.py
batch_size=64 latency: 760.226us -> 409.361us


Zzz9990:
python3 op_test/test_mla.py
batch_size=64 latency: 760.226us -> 409.361us

PR #724, opened 2025-07-24
[TRITON]: LeanAttention implement loop unrolling to reduce VGPR usage

valechen:
Reduce VGPR usages, especially in the causal case to improve overall performance.


rahulbatra85 (dismissed):

valechen:
The PR has been integrated into PR744.

PR #736, opened 2025-07-28
[TRITON]: Add fused GEMMs to optimize FF block

willzhou-amd:
Adding fused GEMM + activation + gating options to reduce memory movement overhead in a FF block.

Changes:
- [x] Add `activation` string parameter support for `gemm_a16w16` to fuse an activation function with the GEMM.
- [x] Add `gemm_a16w16_gating.py` which uses the first half of the (M, N) output along the N dimension as the gating activations and the second half as the layer activations. Produces a tensor of shape (M, N//2). This also includes an `activation` parameter for the gating activation.
- [x] Write tests.
- [x] ~~Write a fully-fused E2E FF kernel.~~ Will go into next PR.

Benchmarks on 350x:
image

Post-tuning:
image

Benchmarks on 300x:
image

Multikernel-ff is where we're at right now (e.g launching two Aiter Triton GEMMs).

### DIFF
--- a/aiter/ops/triton/gemm_a16w16_gated.py
+++ b/aiter/ops/triton/gemm_a16w16_gated.py
@@ -12,6 +12,9 @@ from aiter.ops.triton.utils.pid_preprocessing import pid_grid, remap_xcd
 import aiter.ops.triton.utils.arch_info as arch_info
 from aiter.ops.triton.utils.core import AITER_TRITON_CONFIGS_PATH
 from aiter.ops.triton.activation import _get_activation_from_str
+from aiter.ops.triton.utils.logger import AiterTritonLogger
+
+_LOGGER = AiterTritonLogger()
 
 
 @triton.heuristics(
@@ -197,6 +200,7 @@ def gemm_a16w16_gated(
     Returns:
     - Y: The output matrix with shape (M, N//2).
     """
+    _LOGGER.info(f"GEMM_A16W16_GATED: x={tuple(x.shape)} w={tuple(w.shape)}")
 
     # Shape checks
     assert x.shape[1] == w.shape[1], "Incompatible matrix shapes."


aiter/ops/triton/configs/gemm/MI350X-GEMM-A16W16-gated.json
rahulbatra85, line 1: can you also add MI300 file as well?

aiter/ops/triton/gemm_a16w16_gated.py
rahulbatra85, line 1: Do you think this can be put in the gemm_a16w16.py?
We already have lot of gemm variants and will unify them at some point. For now, don't want keep adding new ones

willzhou-amd: Yeah, I was thinking about that - in fact, that was how I'd originally written it. The issue is that the gated GEMM kernel has a different output shape than the regular one because of the gating - e.g it produces an `(M, N//2)` tensor instead of an `(M, N)` tensor, and it also has slightly different optimal tuning params since it has two accumulators. When we unify our GEMM kernels, are we thinking of unifying them under one frontend API that switches into individual implementations, or are we thinking of merging them into one large file?

aiter/ops/triton/configs/gemm/MI350X-GEMM-A16W16-gated.json
willzhou-amd: Yep, added.

aiter/ops/triton/gemm_a16w16_gated.py
rahulbatra85: No, it will not be one API because on CK side it's split, but may be we unify variants of a16w16, a8w8, etc in their own files.

willzhou-amd: Gotcha, thanks. I think we can probably keep the two files separate for now since the parallelization scheme's a bit different (so code reuse isn't that high) and the tensor output shapes are different (halved along the N dimension). Let's keep it like this for now and reconsider when we're unifying variants?

aiter/ops/triton/gemm_a16w16_gated.py
```diff
@@ -0,0 +1,237 @@
...
+    - dtype: Optional parameter to specifcy bf16 or fp16 datatype. Default is bf16
+    - Y: Output Matrix Y with shape (M, N//2).
+    If this is none, then it's created by this API and returned as output.
+    - activation: Optional activation function to apply to the output. One of ("gelu", "gelu_tanh", "silu", "silu_exp2", "relu")
+
+    Returns:
+    - Y: The output matrix with shape (M, N//2).
+    """
+
+    # Shape checks
```
rahulbatra85, line 205: Can you please add logging here?
See gemm_a16w16 for example

willzhou-amd: On it, thanks!

PR #739, opened 2025-07-29
Try to get cu num from env first

slippedJim:
when build aiter with dockerfile, can not access rocminfo, try to acquire cu_num from env var.

### DIFF
--- a/aiter/jit/utils/chip_info.py
+++ b/aiter/jit/utils/chip_info.py
@@ -30,25 +30,28 @@ def get_gfx():
 
 @functools.lru_cache(maxsize=1)
 def get_cu_num():
-    try:
-        rocminfo = executable_path("rocminfo")
-        result = subprocess.run(
-            [rocminfo], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
-        )
-        output = result.stdout
-        devices = re.split(r"Agent\s*\d+", output)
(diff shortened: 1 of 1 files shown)


aiter/jit/utils/chip_info.py
```diff
@@ -30,25 +30,28 @@ def get_gfx():
...
-                if "Device Type" in line and line.find("GPU") != -1:
-                    match = re.search(r"Compute Unit\s*:\s*(\d+)", device)
-                    if match:
-                        gpu_compute_units.append(int(match.group(1)))
-                    break
-    except Exception as e:
-        raise RuntimeError(f"Get GPU Compute Unit from rocminfo failed {str(e)}")
-    assert len(set(gpu_compute_units)) == 1
-    return gpu_compute_units[0]
+    cu_num = os.getenv("CU_NUM", 0)
```
VeeraRajasekhar: I think this should be converted to int before assigning, or else we are getting an error in get_device_function(). Saw this error while building TE through dockerFile. Thanks.

SKPsanjeevi: yes, even for local build, it would throw an error. `cu_num = int(os.getenv("CU_NUM", 0))` does not throw an error.

shengnxu (approved):

PR #749, opened 2025-07-30
Fix scale preshuffling 

plognjen:
This PR changes preshuffling kernel to "unshuffle" scales which are preshuffled in global memory.
This way no compiler changes are needed. This PR also adds efficient shuffling for mfma32 instructions.


vgokhale (commented):
I would prefer to revert the MFMA32 changes. I don't think we want to expose that to the user. And MFMA32 is not used by other frameworks. Considering the shuffling layout is different, there will likely never be a use case for MFMA32 because we won't be able to coexist with them.

plognjen:
> I would prefer to revert the MFMA32 changes. I don't think we want to expose that to the user. And MFMA32 is not used by other frameworks. Considering the shuffling layout is different, there will likely never be a use case for MFMA32 because we won't be able to coexist with them.

Done!

PR #779, opened 2025-08-05
allow passing in AITER_CORE_DIR as env var

842974287:
Due to python package packing, we need to specify a different dir as the core dir in our prod env.

### DIFF
--- a/csrc/cpp_itfs/utils.py
+++ b/csrc/cpp_itfs/utils.py
@@ -16,6 +16,7 @@ import time
 logger = logging.getLogger("aiter")
 this_dir = os.path.dirname(os.path.abspath(__file__))
 AITER_CORE_DIR = os.path.abspath(f"{this_dir}/../../")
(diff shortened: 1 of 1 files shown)


fsx950223 (approved):

valarLip:
looks like you can reuse this one https://github.com/ROCm/aiter/blob/ee66417cd37ad7626d859cffc5943c921ee0274e/aiter/jit/core.py#L106

842974287:
@valarLip Do you mean switch to use `get_user_jit_dir()` in utils.py? It feels like the result from `get_user_jit_dir()` and `AITER_CORE_DIR` serves different purpose though. Like AITER_CORE_DIR is more for a place to look up aiter code/resource files while `get_user_jit_dir()` is a dir to use for JIT compilation.

valarLip:
could you please describ more about your use case? i don't get the point why we need specify the rscr dir

PR #791, opened 2025-08-08
Add receipt for pytorch

alugorey:
The current "combined" api receipt includes batch_prefill variant. Pytorch does not support this (yet). Creating a new receipt that removes the BATCH_PREFILL variant for a new receipt `6`. This change should not cause any regression as it's a path only hit when the receipt for fwd kernel generation is set to 6. Currently only used by PyTorch.


slippedJim (approved):
LGTM

PR #805, opened 2025-08-11
Reduce CPU overhead of aiter paged attention

mxz297:
## Motivation

Reduce CPU overhead of aiter paged attention, which is exposed during decode forward passes. When running with L4 maverick, we can see some gpu bubbles caused by high cpu overhead from aiter::paged_attention_v1

Screenshot 2025-08-11 at 1 01 15 PM

## Technical Details

Further analysis shows we can optimize CPU overhead from
1. calls to get torch.cuda.get_device_propery() and torch.cuda.get_current_stream(): the values of these calls should not change during inference so there is no need to keep calling them. So we can add new interfaces to allow user (in vllm) to set them before inference.
2. compile_template_op checks if we need to JIT compile by doing file system check (os.exists), which is also incurred for every layer. Always checking file system seems unnecessary and cause high overhead. For a model, we should only need to JIT compile at most once.

Screenshot 2025-08-11 at 1 03 57 PM

## Test Result
aiter::paged_attention_v1 cpu overhead reduced from 120us to 80us. E2E decode forward latency for L4 maverick also reduced by about 5% ~ 10% depends on decode batch size. Note that this improvement requires some changes on vllm side to use the new interface.


mxz297:
My thinking is that this change is to add interface to allow users to set streams at user specified time. I think it is user's decision of the new interface to choose always use the same stream or use a different stream; currently in vllm, we use a single stream and we change single vs multistream in vllm. And if we do not use this interface, the behavior should not change for AITER.

HAIAI:
> My thinking is that this change is to add interface to allow users to set streams at user specified time. I think it is user's decision of the new interface to choose always use the same stream or use a different stream; currently in vllm, we use a single stream and we change single vs multistream in vllm. And if we do not use this interface, the behavior should not change for AITER.

I think the concern (from code proposed) is that we now have an implication that aiter can only deal with one stream from the `global` introduced. Any thoughts?

HAIAI:
@valarLip @carlushuang Would you please have a look?

mxz297:
> I think the concern (from code proposed) is that we now have an implication that aiter can only deal with one stream from the `global` introduced. Any thoughts?

I see your concern but i don't think it actual exists. It is indeed the case that this PR introduced a single global variable holding a stream and AITER will launch kernels using that stream. However, it does not mean that users cannot change the global stream between different AITER kernel launches. 

Withou the change, AITER also only launch kernels to torch.cuda.currentStream, and we also require user intervention to change the current stream to be a different one to trigger multi-stream behaviors.

Therefore, I would argue that this PR does not change AITER's stream semantics.

csrc/cpp_itfs/torch_utils.py
```diff
@@ -44,6 +44,33 @@ def getTensorInfo(el):
 
 aiter_lib = Library("aiter", "FRAGMENT")
 
+aiter_user_set_stream = None
+
+
+def set_gpu_stream(stream: Optional[torch.cuda.Stream]) -> None:
+    global aiter_user_set_stream
```
valarLip, line 51: try something like this https://github.com/ROCm/aiter/blob/f9884d0b1b1353f8eddf67aff8a504b4593528a2/aiter/jit/utils/chip_info.py#L77

mxz297: @valarLip  sorry... i don't think i fully understand this yet...

So, if I try something recommended
```
@functools.lru_cache(maxsize=1)
def get_gfx():
    gfx_num = get_gfx_custom_op()
    return GFX_MAP.get(gfx_num, "unknown")
```

Then it would become
```
@functools.lru_cache(maxsize=1)
def get_gpu_stream()
    return torch.cuda.current_stream()
```

In this case, AITER will always launch kernels to the stream it first see because of the cache and AITER won't be able to change stream?

valarLip: for stream it can be get at c++ layer like this "const cudaStream_t stream = at::cuda::getCurrentCUDAStream();", 
lru_cache still can be used for get_warp_size_for_device

mxz297: > for stream it can be get at c++ layer like this "const cudaStream_t stream = at::cuda::getCurrentCUDAStream();",

I thought about this. But it would require have pytorch headers during JIT compile time. I may have missed something but i did not see pytorch headers used in your pa attention includes, so i am not sure if that's easy to do for JIT? 

I do agree that if we can do this in c++, then it would better. In fact, if we can include pytorch c++ headers, then we can even get rid of ctype bindings, which is currently also causing lots of CPU overhead. 

> lru_cache still can be used for get_warp_size_for_device

I see. I agree with this. And will revise accordingly.

mxz297: @valarLip can you elaborate more on using c++ torch layer in AITER JIT? Is it possible? I guess it would also mean aiter jitted library will be linked against libtorch?

dhonnappa-amd:
Jenkins CI skipped: Check lint is missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Check lint is missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Check lint is missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Check lint is missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Check lint is missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

csrc/cpp_itfs/torch_utils.py
mxz297: @HAIAI  @valarLip  can i get some feedback on this? it has been 3 weeks

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

dhonnappa-amd:
Jenkins CI skipped: Required check(s) 'lint, check-ck, ruff_black' are missing. Exiting the entire job...

PR #825, opened 2025-08-14
[TRITON]: CI Test Set & Benchmark Speedups

willzhou-amd:
This speeds up our benchmarking & test scripts, in some cases by an order of magnitude.
Thanks to @xiaohuguo2023 for sniffing out the benchmarking time discrepancy.

Changes:
- [x] Generate bench & test tensors on the GPU instead of the CPU.
- [x] Use torch.testing.assert_close() instead of triton.testing.assert_close(), as the latter uses a np implementation (on cpu).
- [x] Add readme.


gyohuangxin:
@willzhou-amd I'm glad to see the speedup, this will save almost an hour in the Triton tests. Thank you for the enhancement!
cc @okakarpa

PR #836, opened 2025-08-18
[fused moe][workload] add a test wrapper for fused moe kernel with the metadata from DeepSeek/Qwen

zejunchen-zejun:


wuhuikx (commented):
Please add a README

op_tests/module_test/test_fused_moe.py
```diff
@@ -0,0 +1,237 @@
...
+    topk: int
+
+    def __str__(self):
+        # Parse dim string to extract hidden_dim and moe_intermediate_size
+        hidden_dim, moe_intermediate_size = self.dim.split(',')
+        return f"model={self.model_name}.tp{self.tp_size}.hidden_dim={hidden_dim}.moe_intermediate_size={moe_intermediate_size}.expert={self.expert}.topk={self.topk}"
+
+TEST_CONFIGS = [
+    #          model_name,        tp_size, dim(hidden_dim,inter_dim), expert, topk
+    # DeepSeek-R1 -- TP8
```
wuhuikx: Can we add some arguments to let user select the model and TP size? If user does not set the value, then we run all test cases.

zejunchen-zejun:
```
  # Run all tests (default behavior)
  python test_fused_moe.py
  
  # Run tests for a specific model
  python test_fused_moe.py --model "DeepSeek-R1"
  
  # Run tests for a specific model and tensor parallel size
  python test_fused_moe.py --model "Qwen3-235B-A22B" --tp_size 4
  
  # List all available test configurations
  python test_fused_moe.py --list-configs
```

dhonnappa-amd:
Jenkins CI skipped: Check lint failed. Exiting the entire job...

PR #897, opened 2025-08-27
[Triton] add tunned config for fp8 gemm blockscale for DS-V3 shapes

k50112113:
This PR adds new configs for gemm_a8w8_blockscale for DS-V3 TP8 shapes

note that this tunning was done one MI355, and the MI300 config files are duplicates, which may be suboptimal.

further tunning may be required for MI300 configs


k50112113:
Moving this to 355_wip branch so closing this PR

PR #916, opened 2025-08-28
GPT OSS bug fixes

cagrikymk:
This PR reverts the changes to the attention.

Another related update is GEMM related but it is already committed to 355_wip earlier.


aiter/ops/triton/unified_attention.py
```diff
@@ -795,48 +795,36 @@ def unified_attention(
     total_num_q_blocks = q.shape[0] // BLOCK_Q + num_seqs
     target_num_prgms = get_num_sms() * 4
     num_2d_prgms = total_num_q_blocks * num_kv_heads
-    ALL_DECODE = q.shape[0] == num_seqs
```
maleksan85, line 798: when vLLM is building cudagraph, it preserves all kernels addresses and its parameters, the graph is pretty static and formed when different batches are run at start of vLLM model. Do you think that this condition would ever work?

cagrikymk: The updated condition is `ALL_DECODE = max_seqlen_q == 1`, basically the same idea as before. I am reverting these changes to minimize the friction for the new release. The PR contains the heuristics from the previous release.

I believe both conditions should work as any issue regarding these would even affect correctness since the grid is also dependent on these values.

What do you think?

PR #939, opened 2025-09-02
[TRITON]: Triton add jax triton support

rahulbatra85:
## Motivation

Restructure code to support usage of Triton kernels in JAX through JAX-Triton call

## Technical Details

Triton API Wrappers and triton jitted kernels are split into separate folders

## Test Plan

Run Triton unit tests

## Test Result

All should pass

## Submission Checklist


aiter/ops/triton/gemm_a16w16_agnostic.py
```diff
@@ -0,0 +1,174 @@
...
+
+@triton.heuristics(
+    {
+        "EVEN_K": lambda args: args["K"] % args["BLOCK_SIZE_K"] == 0,
+        "GRID_MN": lambda args: triton.cdiv(args["M"], args["BLOCK_SIZE_M"])
+        * triton.cdiv(args["N"], args["BLOCK_SIZE_N"]),
+    }
+)
+@triton.jit
+def _gemm_a16_w16_kernel(
```
willzhou-amd: This kernel might be a holdover (e.g we want to move it to _triton_kernels). The import's currently broken since the corresponding file in _triton_kernels doesn't exist yet.

aiter/ops/triton/gemm_a8w8.py
```diff
@@ -7,193 +7,17 @@
 import torch
 import triton
 import triton.language as tl
-import aiter.ops.triton.utils.arch_info as arch_info
+import aiter.ops.triton.utils._triton.arch_info as arch_info
```
willzhou-amd, line 8: There's some unused imports in the kernel caller functions (e.g functools, json, arch_info) since the kernels proper have been moved to a separate folder. Not sure if this has a performance impact, but we can definitely remove these. : )

aiter/ops/triton/gluon/gemm_a8w8_blockscale.py
```diff
@@ -7,8 +7,8 @@
 import os
 import torch
 import triton
-from aiter.ops.triton.utils.pid_preprocessing import pid_grid, remap_xcd
-import aiter.ops.triton.utils.arch_info as arch_info
+from aiter.ops.triton.utils._triton.pid_preprocessing import pid_grid, remap_xcd
+import aiter.ops.triton.utils._triton.arch_info as arch_info
 from aiter.ops.triton.utils.core import AITER_TRITON_CONFIGS_PATH
 from aiter.ops.triton.utils.logger import AiterTritonLogger
 from triton import language as tl
```
willzhou-amd, line 14: Thoughts on if we want to move the Gluon GEMM kernel into a subfolder inside _triton_kernels and reuse the Triton calling interface in `gemm_a8w8_blockscale.py`? I'd be happy to do this as well if we want to go ahead with it.

willzhou-amd (commented):
Looks great! Minor nit (not sure if it's relevant) is that we're using relative imports in the kernels in `_triton_kernels` - might make sense to use package imports instead if we're planning to restructure further in the future.

aiter/ops/triton/gemm_a8w8.py
rahulbatra85: Good catch. I thought black is supposed to catch things like these.

aiter/ops/triton/gemm_a16w16_agnostic.py
rahulbatra85: Good catch!

aiter/ops/triton/gluon/gemm_a8w8_blockscale.py
rahulbatra85: yes, we would want to that. I didn't touch gluon yet because you had one PR merged and another in review at the time. Also, others may have gluon related PRs in the works.

willzhou-amd:
> The reason I used relative import is because JAX team can simply copy over this folder in their repo and the relative paths will still work for them. To be honest, this is what worked. I am not sure if this is the best approach. Any ideas on this?

Yeah, that makes sense. Are they copying over the `aiter/ops/triton` folder as a standalone `triton` folder in their repo? If so, our caller functions have absolute imports and will also need to be changed. I think this depends on exactly what they want, but we can consider:
1. Just keeping absolute imports everywhere and letting the JAX team deal with the modifications once they slice our repo.
2. Maybe using a nested package? Could get complex and ugly.
3. I guess we can also make all of our internal imports that they use relative imports, though that would make our own repo less readable.

aiter/ops/triton/gemm_a8w8.py
willzhou-amd: Hmm, yeah, in the past I recall black CI catching unused imports. Could it be bugged after the directory change?

PR #956, opened 2025-09-04
v3_param_check func

minmengdie:


op_tests/cpp/mha/benchmark_mha_fwd.cpp
```diff
@@ -1054,15 +1057,39 @@ bool run(const ck_tile::ArgParser& arg_parser)
...
+                                        stream_config,
+                                        data_type,
+                                        mode == mode_enum::group,
+                                        mask.type,
+                                        bias.type,
+                                        lse,
+                                        fwd_v3,
+                                        v3_bf16_cvt);
+            if (check_fwd_v3 != fwd_v3) {
+                printf("\n[Error]:fwd_v3 param check is wrong!\n");
```
wangye805: This is a little bit confusing. Why not just print "v3 requested but not supported" if (fwd_v3 and !check_fwd_v3)

op_tests/cpp/mha/build_mha.sh
```diff
@@ -8,8 +8,8 @@ else
     FMA_API=""  # build all
 fi
 
-echo "######## building mha kernel $FMA_API"
-python3 compile.py --api=$FMA_API
+# echo "######## building mha kernel $FMA_API"
+# python3 compile.py --api=$FMA_API
```
wangye805: Remove those if not needed? We can always check the history via git

op_tests/cpp/mha/param_check/mha_v3_param_check_codegen.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="delete the line which has '<', and replace the  'return r' -> 'is_v3=true'"
+    )
+    parser.add_argument(
+        "--work_path",
+        help="the input aiter work path to be process",
+        default="/home/memin/code/aiter/hsa",
```
wangye805: This deafult work-path seems to be hardcoded to a local dir? Can we make it more general

op_tests/cpp/mha/param_check/mha_v3_param_check_codegen.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    output_file_name: str = (
+        "mha_bwd_param_v3_check.cpp"  # which file the generated function will be writed
+    )
+    encoding: str = "utf-8"  # the file encoding type
+
+
+codegen_info_list = [
+    CodegenInfo(
+        "gfx942", "fmha_v3_fwd", "fmha_fwd_v3(", 226, 394, "mha_fwd_param_v3_check.cpp"
```
wangye805: Let's avoid the hard cording the start and finish line of a source codes. What if it changes in the future

op_tests/cpp/mha/param_check/mha_fwd_param_v3_check.cpp
```diff
@@ -0,0 +1,246 @@
...
+            r = fmha_fwd_v3_gfx942_check(traits, args, stream_config);
+            break;
+        case arch_enum::gfx950:
+            r = fmha_fwd_v3_gfx950_check(traits, args, stream_config);
+            break;
+    }
+    return r != -1;
+}
+
+float fmha_fwd_v3_gfx942_check(mha_fwd_traits t, mha_fwd_args a, const ck_tile::stream_config& s, const void* seqstart_q_padding_ptr, const void* seqstart_k_padding_ptr) {
```
wangye805, line 47: I'm a little bit confused. Was this file generated by mha_v3_param_check_codegen.py or a human written source code? It looks the if-else statements are quite similar to those in hsa/<arch>/fmha_v3_fwd/codegen.py. 

If it's generated by scripts, we should not include those files in the commit. 

If it's hand written, we should avoid since it will be hard to maintain in the future

In fact, why not adding this v3 check apis into the hsa/<arch>/fmha_v3_fwd(bwd)/codegen.py's?

op_tests/cpp/mha/param_check/mha_v3_param_check_codegen.py
minmengdie: If the codegen.py is changed in the future, the param check code must be regenerated to match the new v3 condition. the start and finish line is a simplified processing method

op_tests/cpp/mha/param_check/mha_fwd_param_v3_check.cpp
minmengdie: you are right, it's generated by scripts. But for the time being, it is not the official external interface of the software, but merely an additional check interface provided for users to understand the v3 support conditions. I think we need provide the complete source code for users to use directly.

op_tests/cpp/mha/benchmark_mha_fwd.cpp
```diff
@@ -1054,15 +1057,39 @@ bool run(const ck_tile::ArgParser& arg_parser)
...
         return aiter::mha_fwd(fmha_args,
-                              stream_config,
-                              data_type,
-                              mode == mode_enum::group,
-                              mask.type,
-                              bias.type,
-                              lse,
-                              fwd_v3,
-                              v3_bf16_cvt);
+                        stream_config,
```
wangye805, line 1085: This section of code change seems to be only indentation. Why do we need to change it?

op_tests/cpp/mha/param_check/mha_fwd_param_v3_check.cpp
wangye805: Emm, but we need the official external interface for this v3 asm kernel check. 

Currently our Transformer Engine library generate aiter c++ apis/kernels through aiter/hsa/<gpu_arch>/<fmha_v3_fwd/fmha_v3_bwd>/codegen.py. It's hard for us to use source codes in op_tests/cpp/mha. 

It also add to your maintenance burden in the future.

PR #1036, opened 2025-09-19
[TRITON] update bf16 GEMM config path

cagrikymk:
This PR updates the BF16 GEMM config path.


dllehr-amd (approved):
Okay with the renaming here.  Approved

PR #1063, opened 2025-09-23
[TRITON] Benchmarking improvements

eky-amd:
## Changes
- [x] When running with a given model, all benchmark scripts now always output the metric evaluated. Previously, table columns would only show "fc1" and/or "fc2", which is not helpful when reading from the terminal.
- [x] Standardize table/graph labelling across Triton benchmarking scripts (standard formatting: `<Provider>_<Layer>_<Metric>_(<Unit>)`). Provider describes whether the bench was performed using the Triton kernel or Torch functions. Layer is either `fc1` or `fc2` for model benchmarks. Both are optional. Metric is either time (ms), throughput (TFLOPS), or bandwidth (GB/s).
- [x] Modify benchmark outputs to match table schema in `op_tests/op_benchmarks/triton/bench_schema.yaml`.
- [x] Implement comparisons with PyTorch for the Triton GEMM kernels, using the `-bench_torch` flag.
- [x] Resolve `arch_info` and `get_splitk` import issues in `triton/bench_batched_gemm_afp4wfp4.py`, `triton/bench_batched_gemm_afp4wfp4_pre_quant.py` and `triton/bench_gemm_a8wfp4.py`.
- [x] Resolve `STR_DTYPE_TO_TORCH_DTYPE` import issue in `triton/bench_pa_prefill.py`.
- [x] Fix default parameters in `op_tests/op_benchmarks/triton/utils/model_configs.json` (removing MoE parameters from llama models, changing top_k parameter from 4 to 8 in deepseek-v3).
- [x] Add gpt-oss model family to Triton benchmarking model config (parameters are determined from the [model card](https://arxiv.org/pdf/2508.10925)).

## Testing
Manuel tests were performed on both MI300 and MI350, across both shape and model benchmarks. Please see these [metrics](https://amd.atlassian.net/wiki/x/3Sr5Qw) for fixed shape parameters, comparing the Triton kernels and PyTorch. Below are some sample outputs, from running `triton/bench_gemm_a8w8_blockscale.py` with llama3-8B.

Before:
Before
After:
After

Triton vs Torch:
Bench Example

Below is a sample run of `op_tests/op_benchmarks/triton/bench_moe.py`, using the new gptoss model family.
image
Compared with deepseek-V3:
image

Attempting to benchmark MoE with the llama3 family of models now accurately throws an error.
image

### DIFF
--- a/op_tests/op_benchmarks/triton/bench_batch_prefill.py
+++ b/op_tests/op_benchmarks/triton/bench_batch_prefill.py
@@ -155,6 +155,14 @@ def run_benchmark(args):
     line_names = ["Time_(ms)", "TFLOPS", "Bandwidth_(GB/s)"]
     line_vals = ["time", "tflops", "bandwidth"]
 
+    # FIXME: The Triton benchmark infrastructure has been changed so that the provided
+    # 'ylabel' is appended to each of the column names in the output table.
+    # In this benchmark, the table would have columns like:
+    # - Time_(ms) (ms / TFLOPS / GB/s)
+    # - TFLOPS (ms / TFLOPS / GB/s)
+    # - Bandwidth_(GB/s) (ms / TFLOPS / GB/s)
+    # This is probably not the desired behavior, but Triton does not seem to support separate
+    # 'ylabel's for each of the provided line values.
     benchmark = triton.testing.Benchmark(
         x_names=x_names,
         x_vals=x_vals_list,
--- a/op_tests/op_benchmarks/triton/bench_extend_attention.py
+++ b/op_tests/op_benchmarks/triton/bench_extend_attention.py
@@ -3,7 +3,6 @@ from aiter.ops.triton import extend_attention, prefill_attention
 import triton
 
 from op_tests.op_benchmarks.triton.utils.benchmark_utils import (
-    get_evaluation_label,
     get_model_configs,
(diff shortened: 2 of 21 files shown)


eky-amd:
# Updates

- Triton recently introduced a change which ensured that the `ylabel` provided for each `triton.testing.Benchmark` object is printed alongside the `line_name` in the printed output table (see [here](https://github.com/triton-lang/triton/pull/8631)).
- This is unfortunate because it directly conflicts with the formatting for the performance CI, which are found in `op_tests/op_benchmarks/triton/bench_schema.yaml`.
- I had taken extra care to make sure my formatting improvements aligned with the schema. But since having the latest pulled Triton would break it anyway (on the current AITER main), I figured I would use a new standard formatting that looks better.
- The new standard formatting looks like for nearly every benchmark: `<Provider>_<Layer>_<Metric>`, with Triton providing the appended `(<Unit>)`. `<Layer>` is generally only available in the gemm benchmarks, when run with the `--model` flag. Similarly, `<Provider>` comes from running the gemm benchmarks with the `-bench_torch` flag.
- I am really not sure how to handle the discrepancy with the benchmark schema, hopefully someone else can provide guidance.

## Examples of new formating
image
image
image
image


## Before (with latest Triton changes on AITER main)
image
image

PR #1124, opened 2025-10-01
[aiter ci] vllm fp8 kv-cache tests | Deepseek & Mixtral

divakar-amd:
Add fp8 kv-cache tests for Deepseek & mixtral


dhonnappa-amd:
Changes LGTM. Merging with failed test as it's expected to catch kernel issue.
cc: @okakarpa

PR #1130, opened 2025-10-06
Bump composable_kernel for fmha 192x128 head dims commit

samremes:
## Motivation

Composable kernel PR https://github.com/ROCm/composable_kernel/pull/2980 changes the occupancy to 1 for 192x128 hdim_q / hdim_v to avoid spilling. Bump CK to that commit in aiter to test.

Also added support for fwd only mode in test_mha.py so we can run it with e.g.
`python3 op_tests/test_mha.py -d fp16 --causal --d_qk 192 --d_v 128 --seqlen_q 2048 --seqlen_k 2048 --nheads 128 --fwd_only --mha_type=mha`

## Submission Checklist


samremes:
Closing this as this was determined to be a compiler regression. Reducing the occupancy would reduce the performance when going back to older compiler, there is a proposed fix already for the compiler.

PR #1170, opened 2025-10-11
[MI35X] Fix meta mha 950 error to make guard return bool func itself

ZhangLirong-amd:
## Motivation
Make bool guard return func itself to avoid torch.compile creating wrong fx graph

## Submission Checklist


aiter/tuned_gemm.py
```diff
@@ -96,7 +96,14 @@ def query_sol_core(
     return solution_idx
 
 
-@torch_compile_guard()
+def query_sol_fake(
+    m: int, n: int, k: int, bias: bool, dtype: str, otype: str, scaleAB: bool = False
+) -> int:
+    solution_idx = 0
+    return solution_idx
```
valarLip, line 117: this will make we always using solution 0 with enable torch compile

PR #1184, opened 2025-10-13
[TRITON] Add Positional Encoding (PE) support to Triton MHA kernels

brunomazzottiamd:
## Motivation

DeepSeek models ([DeepSeek-V2](https://arxiv.org/abs/2405.04434) and [DeepSeek-V3](https://arxiv.org/abs/2412.19437)) rely on Multi-head Latent Attention (MLA). MLA splits $Q$ and $K$ tensors into non-positional (NOPE) and positional (ROPE) parts:

image

The head dimensions of $Q$ and $K$ tensors are 192, 128 elements for NOPE part plus 64 elements for ROPE part. In Triton, all block sizes must be a power of 2, so the next power of 2 after 192 is 256. This inherent design of Triton ends up with 192 / 256 = 75% use along head dimension, with 64 elements being zeroed or masked out of computations.

This PR decouples NOPE and ROPE parts of $Q$ and $K$ tensors in Triton MHA kernels, so we can achieve a 100% use given that both NOPE and ROPE parts are powers of 2. For instance, instead of doing just one matrix multiplication with head dimension equals to 256, we can do two equivalent matrix multiplications with head dimensions 128 and 64.

## Technical Details

* Added support for forward and backward passes. Kernel wrappers figure out if we're dealing with non-PE case (DQ = DK = DV) or PE case (DQ = DK, DK > DV).
* Triton MHA backward has two implementations: "fused" and "one kernel". PE support was added only to "one kernel" implementation since it's the default one and it provides the best performance most of the time.
* For PE case, only head dimensions that are powers of 2 are supported, there's no support for padded PE heads.
* PE lacks support for `fp8` data types, changes were tested with `bf16` and `fp16` data types.

## Test Plan

* Enabled `test_mha_backward` and `test_mha_backward_varlen` that were previously disabled due to accuracy issues related to Triton compiler. These tests are now passing with Triton compiler from October 11th.
* Added 4 new unit tests for PE case : `test_mha_with_pe`, `test_mha_varlen_with_pe`, `test_mha_backward_with_pe`, `test_mha_backward_varlen_with_pe`. They cover ~~2640~~ 768 cases. _(edited after merge, there was a code review suggestion about reducing the number of test cases to speedup CI)_
* Causal + dropout case isn't supported in backward with PE because this case isn't supported in non-PE backward. I think it's best to fix the base implementation before adding PE to the mix.

## Test Result

All tests from `op_tests/triton_tests/test_mha.py` are passing.

## Performance Assessment

Target case is:

* Causal attention
* Data type = `bf16`
* BATCH = 1
* HQ = HK = 128
* SQ = SK = 4096
* DQ = DK = 192 (128 NOPE + 64 ROPE)
* DV = 128

*Non-PE time* refers to the kernel execution time when DQ = DK = DV. I used `-d 192` in forward pass and `-d 256` in backward pass. It wasn't possible to use `-d 192` in backward, the backward kernel doesn't compile with a head dimension that isn't a power of 2.

*PE time* refers to the kernel execution time when DQ = DK and DK > DV, it's the positional encoding case. I used `-d 192 -dv 128` for both forward and backward passes.

### On MI300X

| Pass     | Layout | Non-PE time (ms) | PE time (ms) | Speedup |
|----------|--------|-----------------:|-------------:|--------:|
| Forward  | BSHD   |             8.88 |         2.20 |    4.04 |
| Forward  | THD    |             9.28 |         2.29 |    4.05 |
| Backward | BSHD   |            38.72 |        10.26 |    3.77 |
| Backward | THD    |            37.03 |        14.62 |    2.53 |

### On MI350X

| Pass     | Layout | Non-PE time (ms) | PE time (ms) | Speedup |
|----------|--------|-----------------:|-------------:|--------:|
| Forward  | BSHD   |             3.12 |         1.17 |    2.67 |
| Forward  | THD    |             3.50 |         1.25 |    2.80 |
| Backward | BSHD   |            23.96 |        10.94 |    2.19 |
| Backward | THD    |            24.70 |        11.59 |    2.13 |

## Submission Checklist

### DIFF
--- a/op_tests/triton_tests/test_mha.py
+++ b/op_tests/triton_tests/test_mha.py
@@ -792,12 +792,10 @@ def test_mha_backward_varlen(
 @pytest.mark.parametrize("BATCH", [1, 3])
 @pytest.mark.parametrize(
     "SEQLEN_Q, SEQLEN_K",
-    [(1, 1), (4, 4), (128, 128), (2, 1), (1, 2), (32, 16), (16, 48), (4096, 4096)],
+    [(128, 128), (32, 16), (16, 48), (4096, 4096)],
 )
-@pytest.mark.parametrize(
-    "NUM_Q_HEADS, NUM_K_HEADS", [(1, 1), (4, 4), (2, 1), (128, 128)]
-)
-@pytest.mark.parametrize("HEAD_SZ_QK, HEAD_SZ_V", [(48, 32), (128, 64), (192, 128)])
+@pytest.mark.parametrize("NUM_Q_HEADS, NUM_K_HEADS", [(1, 1), (2, 1), (128, 128)])
+@pytest.mark.parametrize("HEAD_SZ_QK, HEAD_SZ_V", [(128, 64), (192, 128)])
 @pytest.mark.parametrize("DROPOUT", [0.0, 0.25])
 @pytest.mark.parametrize("CAUSAL", [True, False])
 def test_mha_with_pe(
@@ -862,12 +860,10 @@ def test_mha_with_pe(
 @pytest.mark.parametrize("BATCH", [1, 3])
 @pytest.mark.parametrize(
     "SEQLEN_Q, SEQLEN_K",
-    [(1, 1), (2, 2), (4, 1), (1, 4), (16, 16), (32, 16), (64, 128), (4096, 4096)],
+    [(16, 16), (32, 16), (64, 128), (4096, 4096)],
 )
-@pytest.mark.parametrize(
-    "NUM_Q_HEADS, NUM_K_HEADS", [(1, 1), (4, 4), (16, 4), (128, 128)]
-)
-@pytest.mark.parametrize("HEAD_SZ_QK, HEAD_SZ_V", [(32, 16), (96, 64), (192, 128)])
+@pytest.mark.parametrize("NUM_Q_HEADS, NUM_K_HEADS", [(4, 4), (16, 4), (128, 128)])
+@pytest.mark.parametrize("HEAD_SZ_QK, HEAD_SZ_V", [(96, 64), (192, 128)])
 @pytest.mark.parametrize("DROPOUT", [0.0, 0.17])
 @pytest.mark.parametrize("CAUSAL", [True, False])
 def test_mha_varlen_with_pe(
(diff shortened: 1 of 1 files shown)


op_tests/op_benchmarks/triton/bench_mha.py
```diff
@@ -441,22 +510,25 @@ def fn():
...
 
-        input_bytes = q.element_size()
-        output_bytes = q.element_size()
         if varlen:
             total_num_tokens_q = cu_seqlens_q[-1].item()
             total_num_tokens_k = cu_seqlens_k[-1].item()
         else:
             total_num_tokens_q = BATCH * N_CTX_Q
             total_num_tokens_k = BATCH * N_CTX_K
+        # TODO: Is it right for backward? Backward reads do as well, and writes dq, dk and dv...
```
brunomazzottiamd: Resolve this TODO before merging.

tune_mha_bwd.sh
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
```
brunomazzottiamd: Remove this file before merging.

tune_mha_fwd.sh
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env bash
```
brunomazzottiamd: Remove this file before merging.

aiter/ops/triton/_triton_kernels/mha.py
```diff
@@ -296,6 +314,7 @@ def _attn_fwd(
     BLOCK_N: tl.constexpr,
     BLOCK_DMODEL: tl.constexpr,
     BLOCK_DMODEL_POW2: tl.constexpr,
+    BLOCK_DMODEL_PE: tl.constexpr,  # it's zero or a power of 2
```
cagrikymk, line 317: I couldnt add the comment on the respective line but this comment is for lines 308:309:

Wouldnt be better to remove tl.constexpr from max_q_len and max_kv_len arguments since they might change often in an inference run? Looks like removing const. expr. doesnt change perf. based on my limited tests.

Your PR doesn't modify them but might be nice to keep them dynamic for inference.
```
SEQLEN_Q: tl.constexpr,
SEQLEN_K: tl.constexpr,
```

brunomazzottiamd: @cagrikymk, I'll remove `tl.constexpr` as you suggested and benchmark my target shape again. I'll keep you posted on this. Thanks for the suggestion.

brunomazzottiamd: I removed `tl.constexpr` from `SEQLEN_Q` and `SEQLEN_K` arguments of forward kernel. I couldn't notice any performance regression.

aiter/ops/triton/_triton_kernels/mha.py
```diff
@@ -163,6 +177,8 @@ def _attn_fwd_inner(
             qk += tl.dot(q, k) * descale_q * descale_k
         else:
             qk += tl.dot(q, k)
+            if HAS_PE:
+                qk += tl.dot(q_pe, k_pe)
```
cagrikymk, line 181: This comment is for line 186 in the same file:

I get the following warning using the Triton from [56765e8c](https://github.com/ROCm/triton/commit/56765e8c):
```
UserWarning: Logical operators 'and' and 'or' are deprecated for non-scalar tensors; please use '&' or '|' instead
  mask = mask and causal_mask
```
Nothing important but it would be nicer to do "and" operation using "&".

brunomazzottiamd:
**Triton Test** CI job is failing. Unable to compile Triton compiler... I'm going to investigate this issue.

cagrikymk (approved):

aiter/ops/triton/_triton_kernels/mha.py
brunomazzottiamd: It's done. Thanks for noticing this.

PR #1216, opened 2025-10-17
add several f4 tuned config for fw shapes

hongxiayang:
## Motivation
add several f4 tuned config for fw shapes


dllehr-amd (approved):
Thanks @hongxiayang !  approved

PR #1260, opened 2025-10-24
Kernel naming: add reusable constexpr repr helper and testing it with gemm_a16w16

Boss2002n:
## Add config-aware naming to kernels

Introduces a repr mechanism for kernels that embeds compile-time configuration parameters directly into kernel names, making performance traces significantly more informative.

- Created make_kernel_repr() helper in "aiter/ops/triton/utils/_triton/kernel_repr.py"

Examples:
Before: _gemm_a16_w16_kernel

After: _gemm_a16_w16_kernel_BLOCK_SIZE_M_128_BLOCK_SIZE_N_256_BLOCK_SIZE_K_64_GROUP_SIZE_M_8_NUM_KSPLIT_1_SPLITK_BLOCK_SIZE_32_cache_modifier_CG_activation_NONE_use_activation_0

This will help us correlate performance with specific kernel configurations and enable config-level analysis in traces

How to use make_kernel_repr() helper:
_kernel_repr = make_kernel_repr("kernel_name", ["PARAM1", "PARAM2", ...])

@triton.jit(repr=_kernel_repr)
def my_kernel(...):
    ...


aiter/ops/triton/_triton_kernels/gemm_a16w16.py
```diff
@@ -6,6 +6,23 @@
...
+    [
+        "BLOCK_SIZE_M",
+        "BLOCK_SIZE_N",
+        "BLOCK_SIZE_K",
+        "GROUP_SIZE_M",
+        "NUM_KSPLIT",
+        "SPLITK_BLOCK_SIZE",
+        "cache_modifier",
+        "activation",
+        "use_activation",
```
vgokhale, line 25: I think you can skip use_Activation because the "activation" argument will specify what the activation will be, and None if there is nothing specified which makes this redundant.

Boss2002n: Okay I can remove it if you want me to.
My thought process was to include everything possible so that we would never have to debug what any argument was set to
for e.g. in the case where "use_activation" is set to 0 but we can still get visibility into if there's anything being passed to the "activation" argument

vgokhale: I see what you mean. Makes sense. I think for this specific case, may be as part of your clean up effort we can remove the use activation arg and infer from the provided activation. Not for this PR - see if it makes sense in the code you're developing for your most recent change.

PR #1331, opened 2025-11-04
[Triton] add gemm tune check utility function

k50112113:
this PR only adds a utility function to check if a triton gemm has a tunned config or not

### DIFF
--- a/aiter/ops/triton/_triton_kernels/batched_gemm_afp4wfp4.py
+++ b/aiter/ops/triton/_triton_kernels/batched_gemm_afp4wfp4.py
@@ -327,7 +327,7 @@ def _get_config(
     from ..utils.gemm_config_utils import get_gemm_config
 
     # Note: Config files use K=2*K in their naming
-    config = get_gemm_config("BATCHED_GEMM-AFP4WFP4", M, N, 2 * K)
+    config, is_tunned = get_gemm_config("BATCHED_GEMM-AFP4WFP4", M, N, 2 * K)
 
     # Apply custom split-K logic for AFP4WFP4
     if config["NUM_KSPLIT"] > 1:
@@ -345,4 +345,4 @@ def _get_config(
         config["BLOCK_SIZE_K"] = triton.next_power_of_2(2 * K)
         config["SPLITK_BLOCK_SIZE"] = 2 * K
 
-    return config
+    return config, is_tunned
--- a/aiter/ops/triton/_triton_kernels/gemm_a16w16.py
+++ b/aiter/ops/triton/_triton_kernels/gemm_a16w16.py
@@ -251,5 +251,5 @@ def _get_config(
     N: int,
     K: int,
 ):
-    config, _ = get_gemm_config("GEMM-A16W16", M, N, K)
-    return compute_splitk_params(config, K)
+    config, is_tunned = get_gemm_config("GEMM-A16W16", M, N, K)
+    return compute_splitk_params(config, K), is_tunned


aiter/ops/triton/_triton_kernels/batched_gemm_a8w8.py
```diff
@@ -196,7 +196,7 @@ def _get_config(
         with open(fpath, "r") as file:
             config = json.load(file)
         _get_config._config_dict = config
-
+    # TODO: add N_K key
```
vgokhale: Some more info in the TODO on what this is for? That way someone other than you can also kind of get an idea of what the TODO is talking about.

vgokhale (dismissed):

aiter/ops/triton/_triton_kernels/batched_gemm_a8w8.py
k50112113: added more info:
```
    # TODO: use /app/aiter/aiter/ops/triton/_triton_kernels/gemm_afp4wfp4.py::_get_config as an example,
    #       include an addtional layer of dict under _get_config._config_dict[key] where the key = f"{N}_{K}",
    #       so that aiter/ops/triton/utils/_triton/gemm_tune_check.py::gemm_tune_check is able to check if the GEMM is tunned for a particular N, K shape
```

aiter/ops/triton/utils/_triton/gemm_tune_check.py
```diff
@@ -0,0 +1,68 @@
+import importlib
+import inspect
+from typing import Optional
+from aiter.ops.triton.utils.logger import AiterTritonLogger
+
+_LOGGER = AiterTritonLogger()
+
+
+def gemm_tune_check(
+    func: callable, M: int, N: int, K: int, shuffle: Optional[bool] = None
```
maleksan85: wonder if `M` check is needed, at some places in vLLM kernel selection might happen without knowing `M`, thus would be nice to have `M` optional. Assuming that if NxK matches then all `M`s are supported.

k50112113: I understand this, but there are cases where only "any" keys exist in the config files, I wouldn't call that "tunned", I think it would be users' (e.g. vLLM dev) responsibility, to assume a set of shapes to examine

k50112113:
@maleksan85, 
since `_get_config` is no longer `lru_cache` decorated and it was moved to `get_gemm_config`, I cannot use the old way to examine if a shape is tunned or not. I make a change to `get_gemm_config` so that it now returns a tuple of `config` and `is_tunned`. In normal config fetching in APIs, `is_tunned` will just be ignored, but it will be used in `/root/aiter/aiter/ops/triton/utils/_triton/gemm_tune_check.py::gemm_tune_check`

the logic is as follows:
`is_tunned = False if ("any" config is used or config_dict_key is not "default")`

aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -51,7 +51,7 @@ def get_gemm_config(
     K: int | None = None,
     bounds: tuple[int, ...] | None = None,
     specialized_filename: str | None = None,
-) -> dict:
+) -> tuple[dict, bool]:
```
Boss2002n, line 54: If we are changing get_gemm_config function then we have to modify all the triton kernels that call this function as well. 
I only see those changes for aiter/ops/triton/_triton_kernels/gemm_a8wfp4.py

k50112113: those changes were made on the API side, other _get_configs do not require changes cuz it's just gonna return whatever get_gemm_config is returning

k50112113: but I think I missed the one in gemm_a16w16, just fix that

Boss2002n: We would also have to change
aiter/ops/triton/_triton_kernels/batched_gemm_afp4wfp4.py L330
    config = get_gemm_config("BATCHED_GEMM-AFP4WFP4", M, N, 2 * K)
to
  config, _ = get_gemm_config("BATCHED_GEMM-AFP4WFP4", M, N, 2 * K)

k50112113: resolved, also changed `aiter/ops/triton/_triton_kernels/gemm_a16w16.py` L254 ~ L255

from
```
    config, _ = get_gemm_config("GEMM-A16W16", M, N, K)
    return compute_splitk_params(config, K)
```
to
```
    config, is_tunned = get_gemm_config("GEMM-A16W16", M, N, K)
    return compute_splitk_params(config, K), is_tunned
```

aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -51,7 +51,7 @@ def get_gemm_config(
...
     """
     Load a GEMM configuration using the standardized M_LEQ_x/M_GEQ_y/any format.
```
Boss2002n, line 56: would be better to update the return parameters to show the current output params on Line 76

Returns:
        Dictionary with the config params

azaidy (approved):
LGTM!

PR #1365, opened 2025-11-07
mha fp8 support descale

rocking5566:
## Motivation

support descale for flash_attn_fp8_pertensor_func and flash_attn_varlen_fp8_pertensor_func

## Submission Checklist

AITER_LOG_MORE=1 pytest op_tests/test_mha_fp8.py
AITER_LOG_MORE=1 pytest op_tests/test_mha_varlen_fp8.py


aiter/ops/mha.py
```diff
@@ -151,6 +157,9 @@ def gen_mha_fwd_fake_tensors(
     is_causal: bool,
     window_size_left: int,
     window_size_right: int,
+    q_descale: float,
```
valarLip, line 160: it's better user a tensor instead of float, as scale maybe dynamic calculated which is not cudagraph supported

PR #1479, opened 2025-11-24
Jacchang/pa ragged experimental

Jacob0226:
## Motivation

The decode kernel `_paged_attention_kernel `in `aiter/csrc/cpp_itfs/pa/pa_kernels.cuh` shows an increasing runtime as the batch size scales from 8 to 512, rising from 4.8% to 11% of total runtime.
This PR currently supports only `head_size = 128` and `kv_dtype = bf16`.

## Technical Details

This work introduces the following changes:

1. Added double buffering for K-cache loading.
2. Used 64 threads to load the continuous K-cache into LDS and then distributed the data to thread registers to match the MFMA layout.
3. Turn on non-temporal loads for KV cache.

## Test Plan

- Unit test: `aiter/op_tests/test_pa_ragged_experimental.py`
This script differs from the original `test_pa_ragged.py`. The new test reproduces the same KV-cache layout used in SGLang, making it more realistic. It also performs numerical verification against the original kernel (referred to as GOLDEN) to ensure correctness. The new implementation is labeled EXPERIMENTAL.
```
# Docker: lmsysorg/sglang:v0.5.5.post3-rocm700-mi35x
export BS=512
export CL=2048 # context length (prefill length)
export PageSize=1
export OUT=UT
rocprof-compute profile -n $OUT -- python aiter/op_tests/test_pa_ragged_experimental.py -n 512 -c 2048 --warmup 3 --page-size 1
rocprof-compute analyze -p workloads/$OUT/MI355/ --list-stats  > tmp.log 
# Find the kernel paged_attention_ll4mi_QKV_mfma16_kernel dispatch ID for GOLDEN and EXPERIMENTAL
# The second-to-last matched dispatch  ID is GOLDEN
# The last one is for matched dispatch ID EXPERIMENTAL
```
- Accuracy test: 
`sglang/benchmark/gsm8k/bench_sglang.py --num-questions 200 --parallel 200 --num-shots 8`

## Test Result

Node: MI355X (smci355-ccs-aus-n10-09)
- Unit test: 
GOLDEN: 100 us
EXPERIMENTAL: 88 us. 
Speedup: 1.13x.
- Accuracy test: GSM8K doesn't drop by using this PR.

## Submission Checklist

- [V] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/.github/workflows/vllm_benchmark.yaml
+++ b/.github/workflows/vllm_benchmark.yaml
@@ -14,7 +14,7 @@ concurrency:
 env:
   VLLM_BRANCH: "main"
   VLLM_REPOSITORY_URL: "https://github.com/vllm-project/vllm"
-  BASE_IMAGE: rocm/vllm-dev:nightly@sha256:9ed6489aa34b70f8dbd1b2e93f2c8bf8c2f2806c2c28c1b89ed4754057213066
+  BASE_IMAGE: rocm/vllm-dev:nightly
   GITHUB_REPO_URL: ${{ github.event.pull_request.head.repo.clone_url || 'https://github.com/ROCm/aiter.git' }}
   GITHUB_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.event.head_commit.id }}
 
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -157,6 +157,10 @@ def fused_moe_fake(
     num_local_tokens: Optional[torch.Tensor] = None,
     moe_sorting_dispatch_policy: bool = 0,
     dtype: Optional[torch.dtype] = None,
+    hidden_pad: int = 0,
+    intermediate_pad: int = 0,
+    bias1: Optional[torch.Tensor] = None,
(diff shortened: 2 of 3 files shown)


fsx950223:
Format code.
https://github.com/ROCm/aiter/blob/1f5b378dcc9d9b0bcd9456c8c767b7424a5e8190/.githooks/pre-commit#L75

csrc/cpp_itfs/pa/pa_kernels.cuh
```diff
@@ -200,7 +200,7 @@ _paged_attention_kernel(const int* block_table_seq,
 
     // set to true to enable non temporal kv loads: has some benefit in very high
     // batch size cases
-    constexpr bool NT_KV_LOAD = false;
+    constexpr bool NT_KV_LOAD = true;
```
fsx950223: In some cases, false is faster than true.

Jacob0226:
@fsx950223 Ready for re-review

PR #1486, opened 2025-11-25
[PA](fix): fix format error in str

PerryZhang01:
image
SyntaxError in origin, this PR fixed it.


PerryZhang01:
> @sjfeng1999 note the f-string diff between different python version

we use Python 3.10.12

PR #1502, opened 2025-11-28
[fix] correct return type of fused arnorm fake impl

gbyu-amd:
## Motivation

Correct the return type of arnorm fake impl to a tuple, otherwise we will see dynamo error during torch.compile model using this custom fusion kernel:
```bash
torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function aiter.fused_allreduce_rmsnorm_(*(FakeTensor(..., device='cuda:4', size=(s72, 7168), dtype=torch.float8_e4m3fn), FakeTensor(..., device='cuda:4', size=(s72, 7168)), Parameter(FakeTensor(..., device='cuda:4', size=(7168,), requires_grad=True)), 1e-06), **{'group_name': 'tp:0'}): got TypeError("Object of type 'FakeTensor' is not an instance of 'sequence'")
```
image

## Test Plan
Test it with deepseek in ATOM.

## Test Result
It can be compiled without error and run normally.

## Submission Checklist

### DIFF
--- a/.github/workflows/vllm_benchmark.yaml
+++ b/.github/workflows/vllm_benchmark.yaml
@@ -94,7 +94,7 @@ jobs:
 
   vllm_benchmark:
     if: ${{ !github.event.pull_request.head.repo.fork }}
-    runs-on: aiter-8gpu-runner
+    runs-on: aiter-mi325-8gpu-bm
     needs: build_vllm_image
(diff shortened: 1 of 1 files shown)


valarLip (approved):
LGTM

PR #1507, opened 2025-11-28
[fix]: local_device_load_rmsnorm acc error

TennyWang1223:
## Motivation

fix local_device_load_rmsnorm kernel acc error

## Technical Details

keep middle result be fp32, instead of fp32->bf16->fp32, this will make accuracy drop

## Test Plan

use dtype fp16, bf16 has 5% err rate, which is confused,  test result should be pass without warning by using fp16

## Test Result

fp16 all pass

## Submission Checklist

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -998,7 +998,7 @@ namespace aiter
     for (int bid = blockIdx.x; bid < m; bid += gridDim.x)
     {
       float square_sum = 0.0f;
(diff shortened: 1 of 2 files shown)


gbyu-amd:
hi @TennyWang1223 , i think there was a small issue and just comment here: when fallback to split path, res_inp_ is not used? https://github.com/ROCm/aiter/blob/68f5a6b4fcff52cdc560ca88fb6a3c69a825de5a/aiter/dist/device_communicators/communicator_cuda.py#L183-L198

PR #1541, opened 2025-12-03
Add fused bias support for GMM and bias‑gradient/accumulate support for persistent TGMM

sudhu2k:
## Motivation

This PR extends the Triton GMM/TGMM kernels and their wrappers to better match common MLP and MoE use‑cases:
GMM: support an optional, per‑expert bias term in the forward pass so that GMM can directly model out = lhs @ rhs + bias without a separate kernel.
TGMM: expose fused bias gradient computation and an ACCUMULATE mode in the persistent TGMM kernel to make it usable for efficient backward passes (e.g. MoE expert bias gradients, multi‑step accumulation) while keeping numerical behavior well‑tested against PyTorch.
These changes reduce redundant launches, enable fused gradient computation, and bring the Triton operators closer to how PyTorch Linear/MoE layers are typically used in end‑to‑end training.

## Technical Details

* Added an optional `bias` argument to the `gmm` kernel, allowing each group to add a bias vector to its output. The kernel and its Python wrapper now validate the shape, dtype, and memory layout of `bias`. 
* Extended the persistent TGMM kernel (`tgmm_persistent_kernel`) to optionally compute and accumulate bias gradients (`bias_grad`) in float32, using atomic adds for correctness. The Python wrapper now exposes `bias_grad`, `compute_bias_grad`, and `accumulate` options and validates their usage. 

**Testing improvements:**

* Updated the PyTorch reference implementation and test cases to support the new `bias` argument for GMM, including parameterizing tests to run with and without bias, and adjusting tolerances for bias cases.
* Enhanced the TGMM test and reference implementation to verify bias gradient computation and accumulation, including shape, dtype, and memory layout checks.

## Test Plan
Unit tests:

1. Ran pytest op_tests/triton_tests/test_gmm.py -k "gmm" with:
2. use_bias toggled in test_gmm.
3. Multiple shapes from TEST_SHAPES (both test‑only and real model shapes).
4. All supported in_dtype/out_dtype combinations (fp16, bf16), TRANS_RHS variants, and RNG seeds from the existing parametrization.

Ran pytest op_tests/triton_tests/test_gmm.py -k "tgmm" to exercise:

1. test_tgmm for both persistent (ptgmm) and non‑persistent (nptgmm) variants.
2. test_tgmm_accumulate for the persistent TGMM accumulate path.

## Test Result

Forward outputs for both persistent and non‑persistent TGMM match the PyTorch reference within the configured tolerances across all shapes/dtypes/transpositions.
For persistent TGMM, bias gradients match the reference for small/medium shapes; per‑element bias‑grad comparisons are intentionally skipped for the largest shapes due to expected fp32 reduction noise.

## Submission Checklist


aiter/ops/triton/_triton_kernels/gmm.py
```diff
@@ -204,6 +206,17 @@ def gmm_kernel(
                 else:
                     rhs_ptrs += BLOCK_SIZE_K * N
 
+            # Add bias if enabled
+            if USE_BIAS:
```
brunomazzottiamd, line 213: **Question:** Did you test bias pre-loading for better performance? It may help for some shapes, we need to profile in a case by case basis.

By bias pre-loading I mean this:

```python
# Pre-load bias:
if USE_BIAS:
    offs_bias_n = tile_n.to(tl.int64) * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)
    bias_ptrs = bias_ptr + g.to(tl.int64) * N + offs_bias_n
    bias = tl.load(bias_ptrs, mask=offs_bias_n < N, other=0.0)

# K-loop:
for k in range(0, tl.cdiv(K, BLOCK_SIZE_K)):
    ...

# Add bias:
if USE_BIAS:
    bias = bias.to(tl.float32)
    acc += bias[None, :]
```

aiter/ops/triton/_triton_kernels/gmm.py
```diff
@@ -333,12 +349,19 @@ def tgmm_persistent_kernel(
                 loop_m -= 1
 
             acc = tl.zeros((BLOCK_SIZE_K, BLOCK_SIZE_N), dtype=tl.float32)
+            
+            # Initialize bias accumulator
```
brunomazzottiamd, line 358: I think you should initialize `bias_acc` conditionally:

```python
if COMPUTE_BIAS_GRAD:
    bias_acc = tl.zeros((BLOCK_SIZE_K,), dtype=tl.float32)
```

aiter/ops/triton/_triton_kernels/gmm.py
```diff
@@ -371,11 +398,21 @@ def tgmm_persistent_kernel(
...
 
-            tl.store(
-                out_ptrs,
-                acc,
-                mask=(offs_out_k[:, None] < K) & (offs_out_n[None, :] < N),
-            )
+            mask = (offs_out_k[:, None] < K) & (offs_out_n[None, :] < N)
+            if ACCUMULATE:
+                # Load existing values and add to them (like beta=1 in BLAS)
+                old_vals = tl.load(out_ptrs, mask=mask, other=0.0)
```
brunomazzottiamd, line 411: **Question:** Why are you using `store(load() + acc)` instead of `atomic_add(acc)` here? I have two hypothesis:
1. `bf16` doesn't support atomic ops
2. atomic add is bad for performance!

aiter/ops/triton/gmm.py
```diff
@@ -131,11 +136,19 @@ def gmm(
...
     - out must be row-major (out.stride() == (N, 1)).
+    - bias must be row-major (bias.stride() == (N, 1)) if provided.
     """
     check_input_device_dtype(lhs, rhs, group_sizes)
 
     M, K, N, G = get_gmm_shape(lhs, rhs, group_sizes)
 
+    use_bias = bias is not None
+    if use_bias:
+        assert bias.shape == (G, N), f"bias must have shape (G, N) = ({G}, {N}), got {bias.shape}."
```
brunomazzottiamd: Please double check `aiter/ops/triton/utils/gmm_common.py` source file. I think you can add device, type and shape checks to `check_input_device_dtype` and `get_gmm_shape` functions. I may be wrong, but as far as I can remember such helper functions are called in other places as well.

aiter/ops/triton/gmm.py
```diff
@@ -229,6 +243,9 @@ def ptgmm(
     preferred_element_type: torch.dtype = DTYPE,
     existing_out: Tensor | None = None,
     config: dict[str, int] | None = None,
+    bias_grad: Tensor | None = None,
+    compute_bias_grad: bool | None = None,
```
brunomazzottiamd: I think `compute_bias_grad` argument can be removed. You can deduce it from `bias_grad` as `compute_bias_grad = bias_grad is not None`. This will simplify the logic a bit, as well as the function interface.

aiter/ops/triton/gmm.py
```diff
@@ -327,6 +344,45 @@ def ptgmm(
...
+    # -----------------------
+    # By default, compute bias gradient iff a bias_grad buffer is provided.
+    if compute_bias_grad is None:
+        compute_bias_grad = bias_grad is not None
+
+    if compute_bias_grad:
+        assert (
+            bias_grad is not None
+        ), "bias_grad buffer must be provided when compute_bias_grad is True."
+        # Expected layout: (G, K), row-major, float32 (to match atomic_add).
```
brunomazzottiamd: Please double check `aiter/ops/triton/utils/gmm_common.py` source file. I think you can add device, type and shape checks to `check_input_device_dtype` and `get_tgmm_shape` functions. I may be wrong, but as far as I can remember such helper functions are called in other places as well.

aiter/ops/triton/gmm.py
```diff
@@ -327,6 +344,45 @@ def ptgmm(
...
+            1,
+        ), "bias_grad must be row-major with stride (K, 1)."
+
+        # If not accumulating into an existing buffer, zero it first.
+        if not accumulate:
+            bias_grad.zero_()
+
+        bias_grad_ptr = bias_grad
+    else:
+        # Dummy pointer; kernel won't touch it when COMPUTE_BIAS_GRAD is False.
```
brunomazzottiamd: Is this really required? Why can't you pass `bias_grad` being `None` to the kernel?

If this is required then would `bias_grad_ptr = torch.tensor([], device=lhs.device, dtype=torch.float32` work? This is indeed an empty tensor, it isn't a 1 x 1 tensor.

op_tests/triton_tests/test_gmm.py
```diff
@@ -165,9 +166,10 @@ def torch_gmm(
         start_idx = last_row
         end_idx = last_row + m
 
-        out[start_idx:end_idx, :] = (lhs[start_idx:end_idx, :] @ rhs[g]).to(
-            preferred_element_type
-        )
+        result = (lhs[start_idx:end_idx, :] @ rhs[g]).to(torch.float32)
+        if bias is not None:
+            result = result + bias[g].to(torch.float32)
```
brunomazzottiamd: **Small suggestion:** `result += bias[g].to(torch.float32)`

op_tests/triton_tests/test_gmm.py
```diff
@@ -208,13 +212,20 @@ def test_gmm(
     )
     out_triton = torch.empty_like(out_torch)
 
+    # Generate bias tensor if needed
+    bias = None
+    if use_bias:
```
brunomazzottiamd: Please check `gen_gmm_input` function in `gmm_common.py`. This input generation function is also called by the benchmark script. I'd add bias generation to `gen_gmm_input`, so the benchmark script would benefit from the shared code.

op_tests/triton_tests/test_gmm.py
```diff
@@ -257,6 +289,33 @@ def torch_tgmm(
         existing_out=existing_out,
     )
 
+    # Bias gradient handling (test/reference only).
+    # By default, compute bias gradient iff a bias_grad buffer is provided.
+    if compute_bias_grad is None:
```
brunomazzottiamd: I think you can deduce `compute_bias_grad` from `bias_grad`, as I said before.

op_tests/triton_tests/test_gmm.py
```diff
@@ -257,6 +289,33 @@ def torch_tgmm(
...
+        compute_bias_grad = bias_grad is not None
+
+    if compute_bias_grad:
+        assert (
+            bias_grad is not None
+        ), "bias_grad buffer must be provided when compute_bias_grad is True."
+        expected_shape = (G, K)
```
brunomazzottiamd: `torch_tgmm` is calling `check_input_device_dtype` and `get_tgmm_shape` functions from `gmm_commons.py`. You can add all new parameter checking logic to `gmm_commons.py` so the real kernel wrapper and the PyTorch reference implementation can reuse it.

op_tests/triton_tests/test_gmm.py
```diff
@@ -268,10 +327,14 @@ def torch_tgmm(
 
         start_idx = last_col
         end_idx = last_col + m
+        # Match Triton kernel: do matmul in float32, then cast to output dtype.
```
brunomazzottiamd: Why are you converting to `fp32` and then back to `preferred_element_type`? If `lhs` and `rhs` are `bf16` then `lhs[:, start_idx:end_idx] @ rhs[start_idx:end_idx, :]` is a `bf16` matmul. The comment from line 330 seems to be wrong. Please double check the comment and these type casts.

op_tests/triton_tests/test_gmm.py
```diff
@@ -327,26 +390,145 @@ def test_tgmm(
...
-        )
-
-        kernel_wrapper(
-            lhs,
-            rhs,
-            group_sizes,
-            preferred_element_type=out_dtype,
-            existing_out=out_triton,
-        )
+        if persistent:
```
brunomazzottiamd: You aren't testing TGMM kernel without bias gradient, you added a new feature but you have disabled the test of the old behavior. You should add a new Boolean `bias_grad` parameter to the test and skip non-persistent + bias gradient combination with a nice skip message.

op_tests/triton_tests/test_gmm.py
```diff
@@ -327,26 +390,145 @@ def test_tgmm(
...
+            # Allocate bias gradient buffers for reference and Triton.
+            bias_grad_torch = torch.empty(
```
brunomazzottiamd: Try to add `bias_grad` allocation to `gen_tgmm_output` and related functions from `gmm_commons.py` file. By this way, this logic can be reused.

aiter/ops/triton/_triton_kernels/gmm.py
sudhu2k: I couldn't initialize it conditionally because triton is throwing below error:
```
    return ast_to_ttir(self.fn, self, context=context, options=options, codegen_fns=codegen_fns,
triton.compiler.errors.CompilationError: NameError('bias_acc is not defined')
```

Seems like during the ast -> ttir phase, triton walks through all statments and demands every variable used in the kernel have a definition.

aiter/ops/triton/_triton_kernels/gmm.py
sudhu2k: Yeah both!
Also, atomics is unnecessary since each output element `(g, k, n)` is owned by exactly one program instance. There're no overlapping tiles on output tensor.

aiter/ops/triton/gmm.py
sudhu2k: Moved this check to `check_input_device_dtype` function in `gmm_common.py`

aiter/ops/triton/gmm.py
sudhu2k: Yeah, I thought we could create an empty space for `bias_grad` when `bias_grad` is `None `and `compute_bias_grad` is `True`, but I guess AITER shouldn't be responsible for that. We'll let the user create the space if they want it to populate them.

aiter/ops/triton/gmm.py
sudhu2k: Moved `dtype` and `device `check to `check_input_device_dtype` and created a `check_bias_shape_stride` function to check the shape of `bias `specifically.

aiter/ops/triton/gmm.py
sudhu2k: Ah, I did try return `torch.tensor([], device=device, dtype=torch.float32)` and it seems to be working. Yes, it seems to be necessary because triton is trying to validate the dtype of this variable since it's being used in atomic_add even though it might be in a dead branch.

op_tests/triton_tests/test_gmm.py
sudhu2k: Done!

op_tests/triton_tests/test_gmm.py
sudhu2k: Moved bias creation to gen_gmm_tensors in gmm_common.py

op_tests/triton_tests/test_gmm.py
sudhu2k: Done!

op_tests/triton_tests/test_gmm.py
sudhu2k: Done!

op_tests/triton_tests/test_gmm.py
sudhu2k: Reverted it. So If `lhs` and `rhs` are bf16 then `lhs[:, start_idx:end_idx] @ rhs[start_idx:end_idx, :]` will be a bf16 matmul and is then cast into output dtype.

op_tests/triton_tests/test_gmm.py
sudhu2k: Fixed it!

op_tests/triton_tests/test_gmm.py
sudhu2k: I've created a separate function `gen_tgmm_bias_grad` to generate bias_grad since it gave me more flexibility and I was able to use it in `gen_tgmm_tensors` and `get_tgmm_bias_grad` functions

aiter/ops/triton/_triton_kernels/gmm.py
sudhu2k: I did try that, but performance actually went down a bit for the shape I'm testing
```
2025-12-09 00:16:37,898 > Benchmarking Triton GMM kernel:
2025-12-09 00:16:37,898 >   input_type = bf16, output_type = bf16, rng_seed = 0, use_bias = True, accumulate = False
2025-12-09 00:16:37,898 >   trans_lhs = False, trans_rhs = True
2025-12-09 00:16:37,898 >   num_group_sizes = 64, unif_group_sizes = False
2025-12-09 00:16:37,898 >   metric = throughput (in tflops)
2025-12-09 00:16:38,292 >     (M, K, N, G) = (24576, 2048, 2816, 64)
2025-12-09 00:16:47,980 >       ms: p20 =  1.2681, p50 =  1.2859, p80 =  1.3094
2025-12-09 00:16:47,980 >       TFLOPS: p20 = 216.48, p50 = 220.44, p80 = 223.53
2025-12-09 00:16:47,980 >       GB/s: p20 = 746.35, p50 = 759.97, p80 = 770.63
triton_gmm_ibf16_obf16_ntn_tflops:
         M       K       N     G  triton_gmm_ibf16_obf16_ntn_tflops
0  24576.0  2048.0  2816.0  64.0                             220.44
```

previously:
```
2025-12-08 21:51:37,460 > Benchmarking Triton GMM kernel:
2025-12-08 21:51:37,460 >   input_type = bf16, output_type = bf16, rng_seed = 0, use_bias = True, accumulate = False
2025-12-08 21:51:37,460 >   trans_lhs = False, trans_rhs = True
2025-12-08 21:51:37,460 >   num_group_sizes = 64, unif_group_sizes = False
2025-12-08 21:51:37,460 >   metric = throughput (in tflops)
2025-12-08 21:51:37,866 >     (M, K, N, G) = (24576, 2048, 2816, 64)
2025-12-08 21:51:47,558 >       ms: p20 =  1.2396, p50 =  1.2591, p80 =  1.2821
2025-12-08 21:51:47,558 >       TFLOPS: p20 = 221.09, p50 = 225.14, p80 = 228.68
2025-12-08 21:51:47,558 >       GB/s: p20 = 762.23, p50 = 776.19, p80 = 788.37
triton_gmm_ibf16_obf16_ntn_tflops:
         M       K       N     G  triton_gmm_ibf16_obf16_ntn_tflops
0  24576.0  2048.0  2816.0  64.0                             225.14
```

brunomazzottiamd: OK, let's not mess with bias pre-loading.

aiter/ops/triton/_triton_kernels/gmm.py
brunomazzottiamd: So far so good, let's obey the compiler.

op_tests/triton_tests/test_gmm.py
brunomazzottiamd: It's good to know it's working this way!

aiter/ops/triton/utils/gmm_common.py
```diff
@@ -507,6 +528,22 @@ def gen_tgmm_output(
...
+def gen_tgmm_bias_grad(
+    K: int,
+    G: int,
+    device: torch.device | str = DEVICE,
+    with_bias_grad: bool = False,
+) -> Tensor:
+    if with_bias_grad:
+        assert K > 0, f"Number of bias_grad rows K must be positive (K = {K})."
+        assert G > 0, f"Number of groups G must be positive (G = {G})."
+        return torch.empty((G, K), device=device, dtype=torch.float32)
```
brunomazzottiamd, line 544: **Question:** It's fine to use `torch.empty` instead of `torch.zeros` here because we do `bias_acc = tl.zeros(...)` in the kernel. Am I right?

op_tests/triton_tests/test_gmm.py
```diff
@@ -297,13 +340,17 @@ def test_tgmm(
 ):
     assert persistent_str in {"p", "np"}
     persistent: bool = persistent_str == "p"
+    
+    # Skip test: non-persistent TGMM does not support bias gradient computation.
+    if not persistent and with_bias_grad:
+        pytest.skip("Non-persistent TGMM (nptgmm) does not support bias gradient computation.")
```
brunomazzottiamd: As far as I can see from the latest diff content, both TGMM variants (persistent and non-persistent) now support bias gradient computation. I think it's worth enabling this test case.

brunomazzottiamd (approved):
LGTM!

aiter/ops/triton/utils/gmm_common.py
sudhu2k: empty is fine because it's getting zeroed before kernel call
https://github.com/ROCm/aiter/blob/e12dfe439716e19fbd05429a50db73471ec6893f/aiter/ops/triton/utils/gmm_common.py#L701-L704

op_tests/op_benchmarks/triton/bench_gmm.py
```diff
@@ -176,14 +179,22 @@ def benchmark(M: int, K: int, N: int, G: int, provider: str):
...
+                "lhs": lhs,
+                "rhs": rhs,
+                "group_sizes": group_sizes,
+                "preferred_element_type": out_dtype,
+                "existing_out": out,
+            }
+
+            if gmm_type == "gmm":
+                kwargs["bias"] = bias
+            elif gmm_type == "ptgmm" or gmm_type == "nptgmm":
```
brunomazzottiamd, line 192: **Minor suggestion:** I think `"tgmm" in gmm_type` may work here.

PR #1563, opened 2025-12-04
fix(mla_reduce): fix dispatch

dbyoung18:
fix dispatch if-else and add case for head1


ruanjm:
This temp fix looks good to me. Thank you for the quick fix.

This fix is also included in https://github.com/ROCm/aiter/pull/1562.

dbyoung18:
closed, since https://github.com/ROCm/aiter/pull/1562 got merged

PR #1576, opened 2025-12-05
[TRITON] Add attention sink support to Triton MHA kernels

brunomazzottiamd:
## Motivation

In [gpt-oss](https://arxiv.org/pdf/2508.10925) attention implementation, each attention head has a learned bias in the denominator of the softmax. This is similar to [attention sink](https://arxiv.org/abs/2309.17453) and we can enable gpt-oss by adding attention sink support to our AITER MHA kernels (both forward and backward kernels). The target model is [gpt-oss20b](https://huggingface.co/openai/gpt-oss-20b/blob/main/config.json).

## Technical Details

* gpt-oss has one sink parameter per query head, so sink tensor and its gradient must be 1D ([source](https://github.com/openai/gpt-oss/blob/864020abceb92dc5354ebd0b0f51be43bedf65ed/gpt_oss/torch/model.py#L189)).
* Attention sink feature lacks support for `fp8` data types. The proposed changes were tested with `bf16` and `fp32` data types, but they should also work with `fp16` data type.
* Triton MHA backward has two implementations: "fused" and "one kernel". Sink support was added only to "one kernel" implementation since it's the default one and it provides the best performance most of the time.
* Changes in forward kernel:
  * Initializes the running maximum with the respective sink value instead of `-inf`.
* Changes in backward kernel:
  * Gradient of sink is a sum reduction. It should be computed and accumulated per query head.
  * Sink gradient computation was added to the outer loop that computes Q gradient. Partial accumulation is done once per query head and sequence block.
  * The sum reduction was implemented with atomic add because the kernel is parallelized across (KV heads, sequence length, batch). So multiple Triton programs would be computing partial sums of the same sink gradient element, given that we have one per query head.
  * Sink gradient type is `fp32` to enable atomics ops.

## Test Plan

* Added 2 new unit tests for the sink case to `op_tests/triton_tests/test_mha.py`: `test_mha_with_sink` and `test_mha_varlen_with_pe`. They cover 192 new cases and test both forward and backward passes.
* BSHD layout + Causal + Dropout case isn't supported in backward with sink because this case isn't supported in regular backward. THD layout + Dropout case isn't supported in backward with sink because this case isn't supported in regular backward. I think it's best to fix the base backward implementation before adding sink to the mix.

## Test Result

* All sink tests from `op_tests/triton_tests/test_mha.py` are passing on `gfx942` and `gfx950`.
* All other tests from `op_tests/triton_tests/test_mha.py` produce the same results as before the sink was added. This happens on both `gfx942` and `gfx950`. So we can conclude that the newly added sink feature didn't break anything that was already working.

## Performance Assessment

Target attention shapes:

* Data type: `bf16`
* TP1: HQ = 64, HKV = 8, D = 64, SQ = SK = 8192.
* TP8: HQ = 8, HKV = 1, D = 64, SQ = SK = 8192.
* Batch sizes 8-16 for `thd` layout and 1 for `bshd` layout.

Forward performance in `gfx950`:

| TP | Batch size | Layout | Forward time without sink (ms) | Forward time with sink (ms) | Speedup |
|---:|-----------:|--------|-------------------------------:|----------------------------:|--------:|
|  1 |          1 | bshd   |                          0.906 |                       0.910 |    1.00 |
|  1 |          8 | thd    |                          7.169 |                       7.115 |    1.01 |
|  1 |          9 | thd    |                          8.060 |                       7.955 |    1.01 |
|  1 |         10 | thd    |                          8.856 |                       8.827 |    1.00 |
|  1 |         11 | thd    |                          9.778 |                       9.802 |    1.00 |
|  1 |         12 | thd    |                         10.694 |                      10.736 |    1.00 |
|  1 |         13 | thd    |                         11.546 |                      11.487 |    1.01 |
|  1 |         14 | thd    |                         12.414 |                      12.362 |    1.00 |
|  1 |         15 | thd    |                         13.347 |                      13.280 |    1.01 |
|  1 |         16 | thd    |                         14.193 |                      14.188 |    1.00 |
|  8 |          1 | bshd   |                          0.191 |                       0.190 |    1.00 |
|  8 |          8 | thd    |                          0.965 |                       0.963 |    1.00 |
|  8 |          9 | thd    |                          1.079 |                       1.081 |    1.00 |
|  8 |         10 | thd    |                          1.185 |                       1.187 |    1.00 |
|  8 |         11 | thd    |                          1.299 |                       1.317 |    0.99 |
|  8 |         12 | thd    |                          1.409 |                       1.422 |    0.99 |
|  8 |         13 | thd    |                          1.528 |                       1.524 |    1.00 |
|  8 |         14 | thd    |                          1.644 |                       1.640 |    1.00 |
|  8 |         15 | thd    |                          1.760 |                       1.757 |    1.00 |
|  8 |         16 | thd    |                          1.864 |                       1.874 |    0.99 |
|    |            |        |                                |                     Geomean |    1.00 |

Backward performance in `gfx950`:

| TP | Batch size | Layout | "One kernel" backward   time without sink (ms) | "One kernel" backward   time with sink (ms) | Speedup |
|---:|-----------:|--------|-----------------------------------------------:|--------------------------------------------:|--------:|
|  1 |          1 | bshd   | 6.284                                          | 6.342                                       |    0.99 |
|  1 |          8 | thd    | 52.947                                         | 52.272                                      |    1.01 |
|  1 |          9 | thd    | 59.567                                         | 58.790                                      |    1.01 |
|  1 |         10 | thd    | 65.465                                         | 65.748                                      |    1.00 |
|  1 |         11 | thd    | 72.319                                         | 72.383                                      |    1.00 |
|  1 |         12 | thd    | 79.073                                         | 78.599                                      |    1.01 |
|  1 |         13 | thd    | 85.665                                         | 84.760                                      |    1.01 |
|  1 |         14 | thd    | 91.813                                         | 91.301                                      |    1.01 |
|  1 |         15 | thd    | 98.538                                         | 98.074                                      |    1.00 |
|  1 |         16 | thd    | 105.008                                        | 104.472                                     |    1.01 |
|  8 |          1 | bshd   | 4.423                                          | 4.336                                       |    1.02 |
|  8 |          8 | thd    | 9.548                                          | 9.479                                       |    1.01 |
|  8 |          9 | thd    | 11.594                                         | 11.661                                      |    0.99 |
|  8 |         10 | thd    | 12.819                                         | 13.071                                      |    0.98 |
|  8 |         11 | thd    | 13.014                                         | 13.088                                      |    0.99 |
|  8 |         12 | thd    | 13.958                                         | 14.067                                      |    0.99 |
|  8 |         13 | thd    | 15.211                                         | 15.439                                      |    0.99 |
|  8 |         14 | thd    | 16.498                                         | 16.555                                      |    1.00 |
|  8 |         15 | thd    | 17.525                                         | 17.593                                      |    1.00 |
|  8 |         16 | thd    | 17.827                                         | 17.768                                      |    1.00 |
|    |            |        |                                                |                                     Geomean |    1.00 |

Conclusion: Attention sink feature doesn't change performance on `gfx950`.

I did the same analysis on `gfx942` and got the same conclusion. I'm not publishing the numbers in the PR for the sake of brevity.

## Submission Checklist

### DIFF
--- a/op_tests/triton_tests/test_mha.py
+++ b/op_tests/triton_tests/test_mha.py
@@ -1448,8 +1448,8 @@ def test_mha_with_sink(
     torch.testing.assert_close(
         triton_dsink,
         torch_dsink,
-        atol=1e-1,  # higher tolerance due to atomic add
-        rtol=1e-1,  # higher tolerance due to atomic add
+        atol=5e-2,  # higher tolerance due to summation over exp
+        rtol=5e-2,  # higher tolerance due to summation over exp
         msg=lambda msg: f"bwd dsink mismatch\n\n{msg}\n",
     )
 
@@ -1631,7 +1631,7 @@ def test_mha_varlen_with_sink(
     torch.testing.assert_close(
         triton_dsink,
         torch_dsink,
-        atol=1e-1,  # higher tolerance due to atomic add
-        rtol=1e-1,  # higher tolerance due to atomic add
+        atol=5e-2,  # higher tolerance due to summation over exp
+        rtol=5e-2,  # higher tolerance due to summation over exp
         msg=lambda msg: f"bwd dsink mismatch\n\n{msg}\n",
     )


op_tests/triton_tests/test_mha.py
```diff
@@ -1303,3 +1303,335 @@ def test_mha_backward_varlen_with_pe(
...
+        triton_dv,
+        torch_dv,
+        atol=2e-2 if relax_dv_err_tol else bwd_atol,
+        rtol=2e-2 if relax_dv_err_tol else bwd_rtol,
+        msg=lambda msg: f"bwd dv mismatch\n\n{msg}\n",
+    )
+    torch.testing.assert_close(
+        triton_dsink,
+        torch_dsink,
+        atol=1e-1,  # higher tolerance due to atomic add
```
cagrikymk: Could you explain this further?

dsink is defined as fp32, so I wouldnt expect atomic add to cause any issues.
I wonder if this is due to the use of "exp2" and the summation over the exp.

op_tests/triton_tests/test_mha.py
```diff
@@ -1303,3 +1303,335 @@ def test_mha_backward_varlen_with_pe(
...
+        msg=lambda msg: f"bwd dq mismatch\n\n{msg}\n",
+    )
+    torch.testing.assert_close(
+        triton_dk,
+        torch_dk,
+        atol=bwd_atol,
+        rtol=bwd_rtol,
+        msg=lambda msg: f"bwd dk mismatch\n\n{msg}\n",
+    )
+    # Case [True-0.0-64-64-8-1024-1024-3] was failing on "gfx942" due to 1 / 1572864 mismatched element.
```
cagrikymk, line 1437: In a typical randomly generated test, what is the range for the dv? I am curious about the exact values for the mismatch.

op_tests/triton_tests/test_mha.py
brunomazzottiamd: I think you're right. I tried `bf16` type for `dsink` and I was getting high errors when comparing to PyTorch reference. After that, I changed `dsink` to `fp32` and didn't reevaluate the error threshold.

I ran all sink tests again with the default error tolerance `bwd_atol = bwd_rtol = 1.5e-2`. Everything passed except the following cases:

On `gfx950`:

```
test_mha_with_sink[True-0.0-64-64-8-128-64-3]:
E   AssertionError: bwd dsink mismatch
E   Tensor-likes are not close!
E   Mismatched elements: 1 / 64 (1.6%)
E   Greatest absolute difference: 0.028441041707992554 at index (28,) (up to 0.015 allowed)
E   Greatest relative difference: 0.18332210183143616 at index (28,) (up to 0.015 allowed)
```

On `gfx942`:

```
test_mha_varlen_with_sink[True-0.0-64-64-8-128-64-2]:
E   AssertionError: bwd dsink mismatch
E   Tensor-likes are not close!
E   Mismatched elements: 1 / 64 (1.6%)
E   Greatest absolute difference: 0.03745710849761963 at index (52,) (up to 0.015 allowed)
E   Greatest relative difference: 0.07790300995111465 at index (52,) (up to 0.015 allowed)

test_mha_varlen_with_sink[True-0.0-128-64-8-128-64-2]:
E   AssertionError: bwd dsink mismatch
E   Tensor-likes are not close!
E   Mismatched elements: 1 / 64 (1.6%)
E   Greatest absolute difference: 0.021396517753601074 at index (22,) (up to 0.015 allowed)
E   Greatest relative difference: 0.07519805431365967 at index (22,) (up to 0.015 allowed)
```

brunomazzottiamd: Everything is working with `bwd_atol = bwd_rtol = 5e-2` and `fp32` sink gradient.

op_tests/triton_tests/test_mha.py
brunomazzottiamd: @cagrikymk, are you interested in the range of `dv` values (i.e. min and max)? Or are you talking about the error magnitude?

I don't have any statistics about `dv`, but I can get them for you if this is really what you need.

The failure on `gfx942` due to `dv` mismatch is the following one:
```
test_mha_with_sink[True-0.0-64-64-8-1024-1024-3]:
E   AssertionError: bwd dv mismatch
E   Tensor-likes are not close!
E   Mismatched elements: 1 / 1572864 (0.0%)
E   Greatest absolute difference: 0.01953125 at index (2, 0, 0, 25) (up to 0.015 allowed)
E   Greatest relative difference: 0.09716796875 at index (2, 0, 0, 25) (up to 0.015 allowed)
```

brunomazzottiamd: I tried to play with the random seed, but there is always one ore more backward assertion that fails on `gfx942` OR `gfx950` with 1 or 2 element mismatches. IMHO `1.5e-2` is a good compromise, I don't want to loose the tolerance more than this. If `q`, `k` and `v` are `fp32` then all tests pass no matter the architecture and / or random seed.

cagrikymk: Could you try turning off the exp2 and see if that affects error rate?

I wonder if the error we see are the accumulation of that approximation.

brunomazzottiamd: @cagrikymk, I changed the exponentials from `exp2` to `exp`. I had to do some hacking on forward because `exp2` is hardcoded. On backward, it was just a matter of setting `USE_EXP2` to `False`. My changes had no effect on the accuracy of `dq`, `dk` and `dv` but I was able to decrease error tolerance on `dsink` assertion (from `5e-2` to `1.5e-2` for BSHD layout and from `5e-2` to `3e-2` for THD layout).

I think we can say that `exp2` results in lower `dsink` precision. `dsink` computation is the following one:

```python
if USE_EXP2:
    RCP_LN2: tl.constexpr = 1.4426950408889634
    psink = tl.math.exp2(sink * RCP_LN2 - m * RCP_LN2)
else:
    psink = tl.math.exp(sink - m)
dsink = tl.sum(-psink * delta[:, None])
```

---

Attaching the diff that changes `exp2` to `exp`:

```diff
diff --git a/aiter/ops/triton/_triton_kernels/mha.py b/aiter/ops/triton/_triton_kernels/mha.py
index b9343061..4070f746 100644
--- a/aiter/ops/triton/_triton_kernels/mha.py
+++ b/aiter/ops/triton/_triton_kernels/mha.py
@@ -173,7 +173,7 @@ def _attn_fwd_inner(
         q_mask = OFFS_M[:, None] < seqlen_q
         k_mask = (start_n + tl.arange(0, BLOCK_N))[None, :] < seqlen_k
         p_mask = q_mask & k_mask
-        qk_scale = SM_SCALE * RCP_LN2
+        qk_scale = SM_SCALE # * RCP_LN2
         # -- compute qk ----
         qk = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
         if HAS_PE:
@@ -197,12 +197,12 @@ def _attn_fwd_inner(
             alibi_block = _compute_alibi_block(
                 alibi_slope, seqlen_q, seqlen_k, global_m_positions, global_n_positions
             )
-            qk += alibi_block * RCP_LN2
+            qk += alibi_block # * RCP_LN2
         # get max scores so far
         m_ij = tl.maximum(m_i, tl.max(qk, 1))

         # Compute scaled QK and softmax probabilities
-        p = tl.math.exp2(qk - m_ij[:, None])
+        p = tl.math.exp(qk - m_ij[:, None])

         # CAVEAT: Must update l_ij before applying dropout
         l_ij = tl.sum(p, 1)
@@ -226,7 +226,7 @@ def _attn_fwd_inner(
         # -- update output accumulator --
         # alpha is an adjustment factor for acc and li as we loop and find new maxes
         # store the diff in maxes to adjust acc and li as we discover new maxes
-        alpha = tl.math.exp2(m_i - m_ij)
+        alpha = tl.math.exp(m_i - m_ij)
         acc = acc * alpha[:, None]
         # -- update m_i and l_i
         l_i = l_i * alpha + l_ij
@@ -636,7 +636,7 @@ def _attn_fwd(

     if ENABLE_SINK:
         RCP_LN2: tl.constexpr = 1.4426950408889634
-        m_i_value = tl.load(sink_ptr + off_q_head).to(tl.float32) * RCP_LN2
+        m_i_value = tl.load(sink_ptr + off_q_head).to(tl.float32) # * RCP_LN2
     else:
         m_i_value = float("-inf")

@@ -832,9 +832,9 @@ def _attn_fwd(
     if softmax_lse_ptr is not None:
         LN2: tl.constexpr = 0.6931471824645996
         # compute log-sum-exp in base 2 units
-        softmax_lse = m_i + tl.math.log2(l_i)
+        softmax_lse = m_i + tl.math.log(l_i)
         # convert back to natural units
-        softmax_lse *= LN2
+        # softmax_lse *= LN2

         if IS_CAUSAL:
             # zero out nans caused by -infs when doing causal
diff --git a/aiter/ops/triton/mha_onekernel_bwd.py b/aiter/ops/triton/mha_onekernel_bwd.py
index 625fdfd2..3c93d3e6 100644
--- a/aiter/ops/triton/mha_onekernel_bwd.py
+++ b/aiter/ops/triton/mha_onekernel_bwd.py
@@ -307,7 +307,7 @@ def flash_attn_onekernel_backward(
             ENABLE_DROPOUT=use_dropout,
             IS_VARLEN=IS_VARLEN,
             USE_ALIBI=use_alibi,
-            USE_EXP2=True,
+            USE_EXP2=False,
             IS_FP8=IS_FP8,
             FP8_MAX=FP8_MAX,
             DEBUG_TRITON=False,
@@ -363,7 +363,7 @@ def flash_attn_onekernel_backward(
             ENABLE_DROPOUT=use_dropout,
             IS_VARLEN=IS_VARLEN,
             USE_ALIBI=use_alibi,
-            USE_EXP2=True,
+            USE_EXP2=False,
             IS_FP8=IS_FP8,
             FP8_MAX=FP8_MAX,
             DEBUG_TRITON=False,
diff --git a/op_tests/triton_tests/test_mha.py b/op_tests/triton_tests/test_mha.py
index 0387d084..90459fc1 100644
--- a/op_tests/triton_tests/test_mha.py
+++ b/op_tests/triton_tests/test_mha.py
@@ -1448,8 +1448,8 @@ def test_mha_with_sink(
     torch.testing.assert_close(
         triton_dsink,
         torch_dsink,
-        atol=5e-2,  # higher tolerance due to summation over exp
-        rtol=5e-2,  # higher tolerance due to summation over exp
+        atol=1.5e-2,  # higher tolerance due to summation over exp
+        rtol=1.5e-2,  # higher tolerance due to summation over exp
         msg=lambda msg: f"bwd dsink mismatch\n\n{msg}\n",
     )

@@ -1631,7 +1631,7 @@ def test_mha_varlen_with_sink(
     torch.testing.assert_close(
         triton_dsink,
         torch_dsink,
-        atol=5e-2,  # higher tolerance due to summation over exp
-        rtol=5e-2,  # higher tolerance due to summation over exp
+        atol=3e-2,  # higher tolerance due to summation over exp
+        rtol=3e-2,  # higher tolerance due to summation over exp
         msg=lambda msg: f"bwd dsink mismatch\n\n{msg}\n",
     )
```

cagrikymk (approved):

op_tests/triton_tests/test_mha.py
cagrikymk: Thanks for bearing with my nitpicking and detailed evaluation :)

brunomazzottiamd: Anytime! It's always a pleasure to address review suggestions. By this way we can greatly improve the quality of our codebase.

k50112113 (approved):

PR #1606, opened 2025-12-10
Fix MLA issues in the case that nhead=128 and seqlen_q=2

ruanjm:
Memory allocated for metadata_work_info_set is not sufficient.
Also fix some trivial issue as well.

### DIFF
--- a/aiter/ops/attention.py
+++ b/aiter/ops/attention.py
@@ -635,7 +635,7 @@ def get_mla_metadata_info_v1(
     tile_cnt = batch_size * max_qo_tiles_per_batch
 
     if fast_mode:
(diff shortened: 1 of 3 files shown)


valarLip (approved):

valarLip:
> @valarLip Yesterday I tried a new method that could reduce the number of splits when nhead=128, qseqlen=2 to improve performance. However, it failed. This problem is more complicated than I expected and will require more time, so I’m planning to address it in a future PR. Therefore, I rolled back this PR and this PR does only fix the issues.

ic.. thanks

valarLip (approved):

PR #1636, opened 2025-12-12
[TRITON] Move triton_tests into respective folders

Boss2002n:
Move the fusion tests into fusion folder
And remove split_qkv files as they are no longer required


brunomazzottiamd (approved):
LGTM!

PR #1638, opened 2025-12-12
[TRITON] Move triton files into respective folders

Boss2002n:
Reorganize ops/triton into categorized folder structure

Restructured the ops/triton directory to improve code organization and maintainability by grouping related files into logical categories.

### DIFF
--- a/op_tests/test_concat_cache_mla.py
+++ b/op_tests/test_concat_cache_mla.py
@@ -5,7 +5,6 @@ from aiter import dtypes
 import argparse
 import pandas as pd
 import random
-from typing_extensions import List
 
 # torch.set_printoptions(threshold=torch.inf)
 


aiter/ops/triton/__init__.py
```diff
@@ -36,3 +40,133 @@
             "IRIS_COMM_AVAILABLE",
         ]
     )
+
+
+_BACKWARD_COMPAT_MAP = {
```
brunomazzottiamd, line 50: Can you please elaborate why the changes on `aiter/ops/triton/__init__.py` are required?

It's confusing and the code is complex. It lacks documentation (both in the PR and in the code itself). It's no trivial IMHO. I may be wrong, but I have a feeling that these changes will become a maintenance burden.

Hacking the `import` subsystem may be dangerous...

Boss2002n: These changes implement backward-compatibility for modules that were reorganized so that external repos (like sglang for example) can still import it the old "way" of importing.

I can  provide the documentation I went thru for this 

For __getattr__, this helps with importing modules,
https://peps.python.org/pep-0562/

image

(This helped me understand why -https://stackoverflow.com/questions/2447353/getattr-on-a-module)

for _backward_compat_find_spec or find_spec, this helps with importing particular functions/objects from inside modules directly
https://stackoverflow.com/questions/24322927/python-how-to-alias-module-name-rename-with-preserving-backward-compatibility

image

(original python documentation - although I find this not so intuitive -  https://docs.python.org/3/reference/import.html#the-meta-path)

the SimpleNamespace is just to avoid creatin a class and then creating an object - this keeps it simple by lettting us treat it like an object and makes it much more readable

brunomazzottiamd: This is the missing part:
> These changes implement backward-compatibility for modules that were reorganized so that external repos (like sglang for example) can still import it the old "way" of importing.

It's the justification for the black magic that's going on.

brunomazzottiamd (dismissed):

brunomazzottiamd (approved):

PR #1653, opened 2025-12-16
Make setup.py PyPI-compatible by separating git dependencies

xiaohuguo2023:
- Removed git dependency from setup.py extras_require
- Created requirements-triton.txt for optional Iris dependency
- Updated installation documentation in README and Triton comms README
- Kept empty triton_comms and all extras for API compatibility

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -332,8 +332,8 @@ setup(
     extras_require={
         # Triton-based communication using Iris
         # Note: Iris is not available on PyPI and must be installed separately
-        # Install with: pip install -r requirements-triton.txt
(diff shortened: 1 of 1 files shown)


amd-ruitang3:
We added iris@git to requirements.txt. However, we’re happy to adopt your solution if you think requirements-triton-comms.txt is better.

amd-ruitang3:
> > We added iris@git to requirements.txt. However, we’re happy to adopt your solution if you think requirements-triton-comms.txt is better.
> 
> with requirements-triton-comms.txt, it won't install iris if users don't need it.

OK. we wiil remove iris@git from  requirements.txt

PR #1667, opened 2025-12-17
Dev/a8w4 and a8w8splitk

yadaish:
## Submission Checklist


aiter/fused_moe.py
```diff
@@ -512,7 +548,7 @@ def nextPow2(n):
 def get_padded_M(M):
     padded_m = M
     if M >= 1 and M <= 16:
-        padded_m = 16
+        return padded_m
```
coderfeli: Is this safe ? tuned config only has 16?

aiter/fused_moe.py
```diff
@@ -645,7 +695,7 @@ def FinalFunc():
             doweight_stage1,
         ) in fused_moe_1stage_dict[get_gfx()]:
             if q_type == QuantType.per_1x128:
-                run_1stage = True and (inter_dim % 256 == 0)
+                run_1stage = token > 32 and (inter_dim % 256 == 0)
```
coderfeli, line 696: is this good for 308?

aiter/fused_moe.py
yadaish: historically  we padded to 16, it's safe but it may have performance downgrade risk.
So it's  better to tune every possible token for decoding problem I think.

aiter/fused_moe.py
yadaish: It is not the best option in 308.  We can make experiments later.
But hardcode "True" is not scalable for any problemsize and hardware.
Maybe we keep it first then improve it later?

aiter/fused_moe.py
```diff
@@ -1319,7 +1469,7 @@ def cktile_moe_stage1(
     if w1.dtype is torch.uint32:
         D = D * 8
     out = torch.empty(
-        (token_num, topk, D), dtype=hidden_states.dtype, device=hidden_states.device
+        (token_num, topk, D), dtype=dtypes.bf16, device=hidden_states.device
```
coderfeli: forced bf16?

Zzz9990: removed it and replaced it with the data type of out.

aiter/ops/triton/fused_mxfp4_quant.py
```diff
@@ -650,3 +649,196 @@ def fused_dynamic_mxfp4_quant_moe_sort(
...
+    offs = (
+        offs_0[:, None] * stride_scale_o0
+        + offs_1[None, :] * stride_scale_o1
+        + offs_2 * stride_scale_o2
+        + offs_3 * stride_scale_o3
+    )
+    tl.store(scale_sorted_ptr + offs, out)
+
+
+def fused_quant_fp8_sort(
```
valarLip, line 771: no use of this one?

aiter/ops/moe_op.py
```diff
@@ -431,6 +438,11 @@ def moe_cktile2stages_gemm2(
     torch.int4: "i4",
 }
 
+str2dtype_dict = {
```
valarLip, line 441: https://github.com/ROCm/aiter/blob/84b2b2f624a68777985099a8c81e55689753b81c/aiter/utility/dtypes.py#L31 try this one...

aiter/fused_moe.py
```diff
@@ -620,8 +657,22 @@ def FinalFunc():
         )
         logger.info("\033[0m")
 
+    def use_cfg():
```
valarLip, line 657: why bypass...

aiter/fused_moe.py
```diff
@@ -630,7 +681,7 @@ def FinalFunc():
         cfg = cfg_2stages.get(keys, None) if cfg_2stages else None
         if cfg is None:
             logger.warning(f"Fmoe tuning not support for {keys}")
-    if cfg is None:
+    if cfg is None or int(os.environ.get("AITER_HEURISTIC_ONLY", "0")):
```
valarLip: AITER_HEURISTIC_ONLY looks also a bypass env? we can have a flag for debug/test, but i would like rename it to AITER_BYPASS_TUNE_CONFIG

aiter/fused_moe.py
valarLip: keep this inter_dim % 128 == 0

aiter/fused_moe.py
```diff
@@ -1293,6 +1392,60 @@ def torch_moe_stage2(
...
+        block_m,
+        sorted_weights,
+        quant_type,
+        activation,
+        splitk,
+        out.dtype,
+    )
+    if splitk > 1:
+        if activation == ActivationType.Silu:
+            aiter.silu_and_mul(out, tmp_out.view(dtypes.fp32).to(out.dtype))
```
valarLip, line 1449: let's have a todo for silu_and_mul @zufayu  input fp32, output bf16/fp16

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_ck2stages.cu
```diff
@@ -56,13 +56,23 @@ void ck_moe_stage1(torch::Tensor &hidden_states,     // [m, k], input token
                    std::optional<int> block_m = 32,
                    std::optional<torch::Tensor> sorted_weights = std::nullopt,
                    int quant_type = 0,
-                   int activation = 0)
+                   int activation = 0,
+                   int splitk = 1,
+                   std::optional<std::string> dst_type = std::nullopt)
```
valarLip, line 61: no use for this one?

coderfeli (dismissed):

op_tests/test_moe_2stage.py
```diff
@@ -463,6 +464,30 @@ def calc_diff(x: torch.Tensor, y: torch.Tensor):
...
+                    inter_dim,
+                    args.expert,
+                    args.topk,
+                    aiter.ActivationType.Swiglu,
+                    quant_type,
+                    aq_dtype,
+                    wq_dtype,
+                    use_g1u1=True,
+                    doweight_stage1=doweight_stage1,
+                    hidden_pad=hidden_pad,
```
valarLip, line 487: @coderfeli when will we refine these pad logic.......i hate these

aiter/ops/triton/fused_mxfp4_quant.py
Zzz9990: Yes, this operator introduces a bubble and needs to be replaced with HIP. Just keep it here as a reference

aiter/fused_moe.py
yadaish: Currently tuners cannot support split-k  parameters. So we have to bypass tuners for specific problem set.

aiter/fused_moe.py
yadaish: ok.

aiter/fused_moe.py
yadaish: @valarLip "keep this inter_dim % 128 == 0"
I am not sure 128 can work here. Although quant block is 128, there's no guarantee that kernel Kblock 128.
Besides above, is there any difference between 128 and 256 dim here?

valarLip: fei add 128 kernel this week, and main change here to 128 already..

valarLip (approved):

PR #1673, opened 2025-12-17
LLfp4 qr cap for atom

amirumoAMD:
## Motivation

Enables QR when QR will help improve performance (>4MB of data)

## Technical Details

Creates a cap that will enable QR when the matrix size and datatype are below 4MB of data

## Test Plan

Manual trace review on ATOM to see which comm kernel it uses, lm_eval.

## Test Result

Performed as expected

## Submission Checklist


azaidy:
Hi @TennyWang1223 

Our goal is to use INT4 QR in ATOM. However, INT4 QR only works better than AR for transfer sizes >= 4 MB.
Hence, this initial implementation. I assume its not comprehensive so would appreciate your comments on how we can make it solid.

PR #1680, opened 2025-12-18
add dealing with memory access fault in Mp tuner 

yzhou103:
## Motivation

when add new implements, there are some cases that will cause memory access fault, which may interrupt tuning processing,
This will deal with the problem to continue tuning.

## Technical Details

1. fix memory access fault tuning, us=inf means memory fault or hang, us=-1 means other error (example: not support)
2. warm iterators, for some cases, kernel time is very short, which may cause inaccurate profiling time. it is better to set warm up time >100us ?
3. add three parameters:
 --timeout,  timeout for get result per test (s) 
 --warmup, warmup iteration before profiling
 --iters, iterations to gather perf

## Submission Checklist


aiter/utility/mp_tuner.py
```diff
@@ -133,72 +151,168 @@ def work_group(gpuIDMap, fast_mode, err_ratio, in_data, tasks):
...
+    try:
+        # Retrieve GPU ID from the map
+        pid = mp.current_process().pid
+        # if pid not in GPUIDMap:
+        #    # Fallback: Use round-robin GPU assignment based on PID
+        #    gpu_num = torch.cuda.device_count()
+        #    gpu_id = pid % gpu_num
+        #    warning_msg = (
+        #        f"[Warning] Process {pid} not found in GPUIDMap. "
+        #        f"Available PIDs: {list(GPUIDMap.keys())}. "
```
yzhou103, line 160: remove

PR #1714, opened 2025-12-23
add fake for MLA RoPE operator

mqhc2020:
## Motivation

For the function fused_qk_rope_cat_and_cache_mla, SGLang needs fake for it to pass torch compile.

## Technical Details

This commit will need another SGLang commit merged simultaneously, because the API is changed.

## Submission Checklist


k50112113 (approved):
Looks good! Thanks for the addition, I think we are not going to let torch compile see inside this function in any cases, so this is a pretty decent change.

PR #1737, opened 2025-12-25
use a16w4 to replace a4w4 as default policy

Zzz9990:
## Motivation

a4w4 decode call a16w4 to replace quant kernel.
enble splitk in a16w4.

## Submission Checklist

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -797,17 +797,20 @@ def get_2stage_cfgs(
             ksplit,
             run_1stage,
         )
-    if (
-        "ck2stages" in kernelName1
-        or (q_type == QuantType.per_1x128 and doweight_stage1)
-        or q_dtype_w
-        in [
-            dtypes.bf16,
-            dtypes.fp16,
-            torch.uint32,
-            dtypes.fp4x2,
-            dtypes.fp8,
-        ]
+
+    if (kernelName1 and "ck2stages" in kernelName1) or (
+        not kernelName1
+        and (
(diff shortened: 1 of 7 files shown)


op_tests/test_moe_2stage.py
```diff
@@ -207,7 +207,7 @@ def weight_per_128x128_quant(weight, quant_dtype):
         a2_scale = a2_scale.view(token, topk, -1)
     elif (
         qType == aiter.QuantType.per_1x32
-        and (AQDType in [dtypes.bf16, dtypes.fp16, dtypes.fp8])
+        and (AQDType in [dtypes.bf16, dtypes.fp16, dtypes.fp8] or token <= 4)
```
yadaish: why we need tokens <= 4 here? it seems unsafe

aiter/fused_moe.py
```diff
@@ -585,6 +587,8 @@ def get_2stage_cfgs(
     doweight_stage1,
     hidden_pad,
     intermediate_pad,
+    bias1,
+    bias2,
```
yadaish: bias1/bias2 should be removed from lru cache decorated function.
MOEMetadata has member has_bias to indicate whether to use bias.

op_tests/test_moe_2stage.py
lalala-sh: removed

aiter/fused_moe.py
lalala-sh: removed

PR #1762, opened 2025-12-31
[Docs] Add README for Triton Ops detailing general maintenance points

Boss2002n:
Add README for Triton Ops detailing general maintenance points

### DIFF
--- a/aiter/ops/triton/README.md
+++ b/aiter/ops/triton/README.md
@@ -251,7 +251,7 @@ Avoid product-name parsing like:
 
 ```python
 # Don't do this
-if impl == "gluon" and int(DEVICE_ARCH.split("MI")[1].replace("X", "")) < 350:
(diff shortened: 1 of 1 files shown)


aiter/ops/triton/README.md
```diff
@@ -0,0 +1,290 @@
...
+```python
+if impl == "gluon" and DEVICE_ARCH not in ("gfx950",):
+    ...
+```
+
+Avoid product-name parsing like:
+
+```python
+# Don't do this
+if impl == "gluon" and int(DEVICE_ARCH.split("MI")[1].replace("X", "")) < 350:
```
vgokhale: Can you change the "350" part of the text to something like 

if impl == "gluon" and int(DEVICE_ARCH.split("MI")[1].replace("X", "")) < mi_awesome_chip_number:

Just to be sure we don't have actual product name here either.

vgokhale (dismissed):

PR #1783, opened 2026-01-07
Fix fused qk concat cache mla

yzhou103:
## Motivation

report error when running in docker rocm6.4.1

## Technical Details

1. in docker rocm6.4.1 ck_tile::bf16_t is actually unsigned short, which caused the data mismatch, now we convert the data type to fp32 explicitly
2. also found perf is very bad at rocm6.4.1. It is caused by the buffer_o.template set_raw interface. 
I used set_raw replaced the set interface as i found it will cause 2 buffer_store_dwordx2 previously. This problem is fixed in Rocm7.x
## Test Plan

python op_tests/test_concat_cache_mla.py -c fused_qk -hd 16 -k 512 -t 1 -kvd auto -qd auto

## Submission Checklist


yzhou103:
Not merge, it seems there is some problem in test

valarLip (approved):

PR #1821, opened 2026-01-12
Implement a new api that will be switching between asm and hip pa internally

JohnNikolay84:
Inference engines should be calling paged_attention_common now with shuffled kv cache layout and aiter internally will decide between asm or hip kernel. HIP is more performant for lower concurrencies ( < 128). Also test_pa.py unit test has been updated to include the new interface.

Since asm and hip kernels require scales in different layouts the client of this api will be expected to provide both ( a scale value for hip per k/v tensor and an expanded per block scale for asm ).

## Motivation
vLLM has had a binary choice between HIP ( c++ ) and asm kernels provided by aiter framework. HIP kernel is more performant on lower concurrencies, asm kernel is more performant on higher concurrencies.  It would be great if Aiter could pick the best depending on the input.

## Technical Details

HIP pa kernel under csrc/cpp_itfs will now support a variant when kv_cache is provided as 5D tensor (asm style)
Aiter internally will switch between HIP and asm based on expected concurrency.
Inference engines should be calling new API with asm shuffled kv cache layouts - paged_attention_common, but also provide a scratch space  in case a 2-stage HIP kernel will be selected.
No legacy api is affected.

## Test Plan

Run qwen3 model and make sure accuracy has not regressed and performance is improved
Create a new aiter unit test for the new api and make sure it pass.

## Test Result

25% performance improvement for CONC > 64 on qwen3
8% overall improvement for Qwen_Qwen3-235B-A22B in geomean-OTPS

image


op_tests/README_pa_merged_tests.md
JohnNikolay84, line 1: I should remove this

op_tests/test_pa_merged.py
JohnNikolay84, line 1: this file is not needed either

csrc/cpp_itfs/pa/pa_v1.cuh
```diff
@@ -35,13 +35,13 @@ template <typename scalar_t,
...
-          bool SLIDING_WINDOW_ENABLED>
+          bool SLIDING_WINDOW_ENABLED,
+          bool USE_5D_VCACHE = false>
 __global__ __launch_bounds__(NUM_THREADS) void paged_attention_ll4mi_QKV_mfma16_kernel(
     const scalar_t* __restrict__ q,      // [num_seqs, num_heads, head_size]
-    const cache_t* __restrict__ k_cache, // [num_blocks, block_size, num_kv_heads,
-                                         // head_size]
-    const cache_t* __restrict__ v_cache, // [num_blocks, block_size, num_kv_heads,
-                                         // head_size]
+    const cache_t* __restrict__ k_cache, // [num_blocks, num_kv_heads, head_size/x, block_size, x]
```
fsx950223: For 5D k_cache, you should use paged_attention_rocm instead,

JohnNikolay84: Done. I am blind, I thought HIP kernels for pa and pa_v1 are identical. :)

PR #1834, opened 2026-01-13
Fix INT4 QR TP8 boundary condition

azaidy:
## Motivation

#1673 added a broad condition to use QR when input size > 4MB. This PR attempts to narrow the scope down to fp16/TP8/INT4 QR. The performance of QR is better than AR only above 8 MB. At 4MB QR achieves 32 us as opposed to 27 us for AR.

## Submission Checklist


valarLip (approved):
LGTM

PR #1866, opened 2026-01-19
Fix code style after updating Black to 26.1.0

gyohuangxin:


dbyoung18:
https://github.com/ROCm/aiter/blob/main/CONTRIBUTE.md
Need also align the fixed version here.

PR #1875, opened 2026-01-20
mla ps support page64 and add paged 3buffer layout for ds3.2

minmengdie:
## Motivation
mla ps add paged 3buffer layout for ds3.2

## Submission Checklist

### DIFF
--- a/csrc/kernels/mla/metadata/v1_comm.cuh
+++ b/csrc/kernels/mla/metadata/v1_comm.cuh
@@ -3,13 +3,14 @@
 
 #pragma once
 
+#include <torch/python.h>
+#include <ATen/hip/HIPContext.h>
+#include <ATen/hip/impl/HIPGuardImplMasqueradingAsCUDA.h>
 #include "aiter_hip_common.h"
 #include "custom_all_reduce.cuh"
 #include "mla.h"
 #include "pa.h"
-#include <ATen/hip/HIPContext.h>
-#include <ATen/hip/impl/HIPGuardImplMasqueradingAsCUDA.h>
-#include <torch/python.h>
+
 
 CK_TILE_HOST_DEVICE int32_t cal_cost(const int32_t qo_len, const int32_t kv_len)
 {


op_tests/test_mla_persistent.py
```diff
@@ -552,7 +782,23 @@ def test_absorb_decode_fp8():
     help="""variable kv seqlens per batch. Default: False.
     --varlen # True""",
 )
-
+parser.add_argument(
```
Zzz9990, line 932: add description for "LEGACY" & "3BUFFER“

op_tests/test_mla_persistent.py
```diff
@@ -291,7 +438,8 @@ def test_mla(
         reduce_indptr,
         reduce_final_map,
         reduce_partial_map,
-        kv_granularity=max(page_size, 16),
+        page_size=page_size,
+        kv_granularity=2,
```
ruanjm: Is `2` is a optimized setting for the case `page_size=64` ? I thought `kv_granularity` should be set as `max(1, 16/page_size)`. However, this formula doesn't match your setting here. If it doesn't work, I think we need a function to get the granularity which takes `max(1, 16/page_size)` as default.

csrc/kernels/mla/metadata/v1_2_device.cuh
```diff
@@ -119,7 +118,7 @@ __launch_bounds__(ck_tile::get_warp_size(), 1) __global__
     // expected payload handled by each cu part.
     const int32_t payload = ck_tile::integer_divide_ceil(sum_blocks, params.num_splits) +
                             Traits::kFixedOverheadNumBlocks;
```
ruanjm: I think we need to change the fixed coefficient `kFixedOverheadNumBlocks` based on page size. It is not a divisor so we may can make it be a runtime parameter.

ruanjm: It is good to me to do this in further PR.

PR #1900, opened 2026-01-23
Remove uses of AITER_ASM_DIR

draganmladjenovic:
## Motivation

Makes sure that user doesn't have to distribute kernels nor set up AITER_ASM_DIR.

## Technical Details

Embed code objects into binary. Use hipRegisterFatBinary to make it seamlessly work on multiple gpus. Make CFG tables read-only and AiterAsmKernels statically allocated.

## Test Plan

Selected tests from op_tests on gfx942

### DIFF
--- a/csrc/py_itfs_cu/asm_fmoe.cu
+++ b/csrc/py_itfs_cu/asm_fmoe.cu
@@ -2,6 +2,7 @@
 // Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
 #include "aiter_hip_common.h"
 #include "asm_fmoe_configs.hpp"
+#include "asm_fmoe_code_objects.hpp"
 #include "moe_op.h"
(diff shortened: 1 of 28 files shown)


csrc/py_itfs_cu/asm_fmoe.cu
```diff
@@ -2,6 +2,7 @@
 // Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
 #include "aiter_hip_common.h"
 #include "asm_fmoe_configs.hpp"
+#include "asm_fmoe_code_objects.hpp"
```
yuguo68, line 5: wondering when do we need both codegen files?

draganmladjenovic: Still has few standalone kernels that are not in the tables but manually handled.

PR #1933, opened 2026-01-30
[OPUS] enhance cast

carlushuang:
test: https://github.com/carlushuang/gcnasm/tree/master/opus_cast


kaiyang-1:
The `fp32_to_bf16` appears to cause precision loss based on my gemm test. Simply use `static_cast` for fp32_t <-> bf16_t may work as well?

carlushuang:
added a `gfx950` specialization for f32->bf16 cast with `static_cast`

PR #1949, opened 2026-02-02
Add allreduce rmsnorm 1stage fusion pass

xytpai:
Add one stage allreduce+rmsnorm fusion pass for decoding:

image


TennyWang1223 (commented):
LGTM

PR #1967, opened 2026-02-04
 Fix the precision issue of allreduce

yanboshao:
## Motivation

Fix the precision issue of write mode allreduce

## Technical Details

Since synchronization between ranks can only guarantee block-level synchronization, it is necessary to ensure that the data blocks in different stages are consistent.

## Test Plan

Cycle pressure test test case - (1271, 3584) &bf16. Compare whether the results on each card are consistent, and whether the results of each two rounds are consistent.

## Test Result

pass

## Submission Checklist


valarLip (approved):
LGTM

PR #1973, opened 2026-02-04
Defer expensive build operations to build_ext.run()

### LINKED ISSUE #1785: [Issue]: Aiter setup.py causing arbitrary code execution
### Problem Description

If we import setup.py to build the sdist or gather dependencies and we have the variables IS_ROCM and PREBUILD_KERNELS set the code will build the kernels outside of the build wheel step.

https://github.com/ROCm/aiter/blob/main/setup.py#L73-L192

### Operating System

N/A

### CPU

N/A

### GPU

N/A

### ROCm Version

N/A

### ROCm Component

_No response_

### Steps to Reproduce

import setup.py in an environment with the variables set.

_No response_

### Additional Information

_No response_

paradigm:
Previously, setup.py imported jit modules and performed build detection at module load time, causing metadata-only operations (egg_info, --version, etc.) to trigger the full build process.

Move heavy imports (jit.core, cpp_extension, torch) and build logic into NinjaBuildExtension.run(), which only executes during actual build commands. Directory copies for aiter_meta remain at module level as they're needed for setuptools package validation.

This reduces metadata operation time from ~45 minutes to ~1 second.

## Motivation

Make non-build operations significantly faster by avoiding unnecessary builds.

## Technical Details

- Moved build logic into build hook
- Some very minor code clean-ups, e.g. changed conditional `if cond: var = true else: var = false` into a straight `var = cond` assignment.

## Test Plan

- Ran metadata-only operation `python ./setup.py --name`
- Ran a build

## Test Result

- Metadata-only operations such as `python ./setup.py --name` and succeed relatively quickly without a long build
- Build completed as expected.

## Submission Checklist

Fixes #1785


amd-ruitang3:
Thanks @paradigm. Your PR looks good. You’ve considered metadata-only operations, which makes the setup much better. We are testing~

valarLip (approved):

PR #1981, opened 2026-02-05
Optimization for _fused_add_rmsnorm_pad Kernel

amd-wsung102:
## Kernel Optimization

This PR optimizes the `_fused_add_rmsnorm_pad` kernel by selecting the optimal BLOCK_SIZE_N, num_warps, and num_stages for an input shape. The autotune keys `["N", "N_OUT"]` allow Triton to select the better configuration based on the hidden dimension size. Smaller dimensions benefit from smaller block sizes and larger dimensions can leverage larger blocks with more parallelism.

## Performance Results

Setup: MI355, GPT-OSS-120B, 1k / 8k, KV cache FP8, TP8

- Reduced the kernel execution time from 6us 479ns to 4us 519ns, for certain input shapes, yielding around 1.4x speedup.
- Improved E2E total token throughput from 8652 tok/s to 9859 tok/s.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/wsung102/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/wsung102/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

Input shape | BLOCK_SIZE_N | num_warps | num_stages | Runtime
-- | -- | -- | -- | --
M=1,N=2880 | 64 | 2 | 2 | 4us 519ns
M=2,N=2880 | 64 | 2 | 2 | 4us 519ns
M=4,N=2880 | 64 | 2 | 2 | 4us 519ns
M=8,N=2880 | 64 | 2 | 2 | 4us 519ns
M=16,N=2880 | 64 | 2 | 2 | 4us 519ns
M=32,N=2880 | 64 | 2 | 2 | 4us 519ns
M=48,N=2880 | 64 | 2 | 2 | 4us 919ns
M=64,N=2880 | 64 | 2 | 2 | 4us 919ns
M=128,N=2880 | 64 | 2 | 2 | 5us 79ns
M=256,N=2880 | 64 | 2 | 2 | 5us 79ns
M=8192,N=2880 | 64 | 2 | 2 | 5us 559ns
M=16384,N=2880 | 64 | 2 | 2 | 5us 559ns

</body>

</html>

## Unit Test
Ran the `test_fused_add_rmsnorm_pad.py` unit test and all tests passed.
`===================================== 420 passed in 36.96s ======================================`

### DIFF
--- a/aiter/ops/triton/_triton_kernels/fused_add_rmsnorm_pad.py
+++ b/aiter/ops/triton/_triton_kernels/fused_add_rmsnorm_pad.py
@@ -2,6 +2,15 @@ import triton
 import triton.language as tl
 
 
+CONFIGS = [
+    triton.Config({"BLOCK_SIZE_N": 64}, num_warps=2, num_stages=2),
+    triton.Config({"BLOCK_SIZE_N": 128}, num_warps=4, num_stages=2),
+    triton.Config({"BLOCK_SIZE_N": 256}, num_warps=4, num_stages=3),
+    triton.Config({"BLOCK_SIZE_N": 512}, num_warps=8, num_stages=3),
+    triton.Config({"BLOCK_SIZE_N": 1024}, num_warps=8, num_stages=4),
+    triton.Config({"BLOCK_SIZE_N": 2048}, num_warps=8, num_stages=4),
+]
+
 @triton.jit
 def _rmsmorm_op(row, weight, n_cols, epsilon):
     row_norm = row * row
@@ -11,6 +20,10 @@ def _rmsmorm_op(row, weight, n_cols, epsilon):
     rms_norm = row * norm_factor * weight
     return rms_norm
 
+@triton.autotune(
+    configs=CONFIGS,
+    key=["N", "N_OUT"],
+)
 
 @triton.jit
 def _fused_add_rmsnorm_pad(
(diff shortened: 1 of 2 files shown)


aiter/ops/triton/_triton_kernels/fused_add_rmsnorm_pad.py
```diff
@@ -11,6 +26,11 @@ def _rmsmorm_op(row, weight, n_cols, epsilon):
     rms_norm = row * norm_factor * weight
     return rms_norm
 
+@triton.autotune(
```
vgokhale: We don't use autotune in aiter. The recommended way is to use a config file that is cached. You can see how this is done for the GEMM kernels.

amd-wsung102:
> Hi @amd-wsung102
> 
> Can you provide the shape of the input for which you got improvements on and the optimal configs corresponding to them? I'd suggest that we add a heuristic of determining the BLOCK_SIZE_N, num_warps, and num_stages, ... etc. before calling the kernel, it would be something like https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/quant/fused_fp8_quant.py#L103, the reason behind this is because 1. We don't generally use autotune in AITER, it'll cause unnecessary overhead if we are able to know the optimal configs for the shapes are running. 2. config files are heavily used by kernels such as GEMMs, but for this RMS norm, there is probably not much thing to tune, so I prefer adding heuristics if it's not too complicated.
> 
> Thanks

Hi Shao-Chun, thanks for reviewing this PR. I have included the input shapes that received performance improvement as well as the corresponding configs. They are in the PR description. While there's improvement for most shapes, smaller shapes received the most improvement. Since the optimal config is now known, I will proceed to add it as a heuristic, as you suggested.

k50112113 (dismissed):
Hi thanks for the update, sorry for the late reply, looks good, I think you can just merge after resolving the conflict

zufayu:

PR #2036, opened 2026-02-11
gfx12 gemm a8w8

ahmed-bsod:


aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+    HAS_BIAS: gl.constexpr,
+    BLOCK_SIZE_M: gl.constexpr,
+    BLOCK_SIZE_N: gl.constexpr,
+    BLOCK_SIZE_K: gl.constexpr,
+    GROUP_SIZE_M: gl.constexpr,
+    EVEN_K: gl.constexpr,
+    GRID_MN: gl.constexpr,
+    NUM_XCDS: gl.constexpr,
+    NUM_WARPS: gl.constexpr,
+    K_DIM: gl.constexpr,
```
vgokhale: Is this just to allow both int8 and fp8 support and their respective WMMAs? If yes, I think we can skip it. INT8 is not as critical anymore and having support for both makes things harder to read and maintain.

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+        version=3,
+        transposed=True,
+        warp_bases=[[0, 1], [0, 2], [1, 0]],  # [2, 4] layout
+        reg_bases=[],
+        instr_shape=[16, 16, K_DIM],
+    )
+
+    shared_a: gl.constexpr = gl.SwizzledSharedLayout(
+        vec=8, per_phase=1, max_phase=16, order=[1, 0]
+    ) # CHECK VEC=16, rn doing ds_load_2addr_stride64_b64. Will ds_load_b128 be better?
```
vgokhale: I think padded shared layout is supposed to help with this?

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
ahmed-bsod: yeah. i'll skip it then

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
ahmed-bsod: i looked into this but forgot to change it here. vec=16 & k_width=16 would be better here since we would have two bank conflicts per lds read vs three if we use vec=8 & kwidth=8. I was confused because the triton ttgir was using vec=8 and k_width=8. And that's because the wmma layout for 16x16x128 uses k_width=8 according to shader guide. However, since I am using k_width=16 for both A and B, it's a permute on both A and B which cancel out. I checked the asm as well to confirm and it doesn't use any permutes even if i use kwidth 16.

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+        )
+
+    offs_bk = gl.arange(0, BLOCK_SIZE_K, layout=gl.SliceLayout(1, blocked_kn))
+    offs_bn = (
+        pid_n * BLOCK_SIZE_N
+        + gl.arange(0, BLOCK_SIZE_N, layout=gl.SliceLayout(0, blocked_kn))
+    ) % N
+    offs_b = offs_bk[:, None] * stride_bk + offs_bn[None, :] * stride_bn
+    if EVEN_K:
+        b = gl.amd.gfx1250.buffer_load(
```
vgokhale: Does this have the ability to do async loads? So loads directly to LDS? Or not yet?

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+            )
+        else:
+            a = gl.amd.gfx1250.buffer_load(
+                ptr=a_ptr,
+                offsets=offs_a,
+                mask=(offs_ak[None, :] < K - (k + 1) * BLOCK_SIZE_K),
+            )
+
+        # LDS write current block of B
+        smem_b.store(b)
```
vgokhale: Why not pipeline this one too and store it into LDS outside the loop for the first iter?

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
ahmed-bsod: that's the kernel below this using tdm to do async loads: _gemm_a8w8_kernel_async
if you mean using global_to_shared() i can add that to the async version

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
ahmed-bsod: on cdna4 arranging the ds stores and loads like this improved performance

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+        offsets=offs_a_scale,
+        mask=offs_a_scale < M,
+    )
+    b_scale = gl.amd.gfx1250.buffer_load(
+        ptr=b_scale_ptr,
+        offsets=offs_b_scale,
+        mask=offs_b_scale < N,
+    )
+
+    producer=0
```
vgokhale: I would suggest changing the producer/consumer nomenclature. It is really referring to wave specialization which, looking at the rest of the code, this isn't doing (yet). Did you add it because you plan to do it in the future?

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,590 @@
...
+from aiter.ops.triton.utils._triton.pid_preprocessing import pid_grid
+from aiter.ops.triton.utils.core import AITER_TRITON_CONFIGS_PATH
+from triton.experimental import gluon
+from triton.experimental.gluon import language as gl
+
+USE_TDM = True
+TRANSPOSE_B = False
+
+@gluon.jit
+def issue_loads(producer, a_desc, b_desc, off_am, off_bn, a_buffer, b_buffer, BLOCK_K: gl.constexpr,
```
vgokhale: Can we separate this to do a single TDM load, and then put it in the utils? This is kind of highly specialized to do GEMMs and I don't think it is worth this level of specialization (what if I have just one load? Or 3 as in flash attn?).

PR #2050, opened 2026-02-13
[TRITON] Add Model Benchmarking Tool

lucas-santos-amd:
## Motivation

Provide a script to benchmark multiple GEMM, MoE, RMSNorm and RoPE Kernels easily.

## Technical Details

- The called benchmarks were modified to accept arguments from the caller script.
- The shapes to be benchmarked are stored into a JSON file, making it easy to add new kernels and shapes (as long as their output format is already supported by the script).
- It's possible to benchmark the shapes for a list of specific models or all of the available ones.
- At the end, the results are printed to the CLI and saved to a CSV file.

Help text of `bench_models.py`:

```
usage: bench_models.py [-h] [--M M [M ...]] [--TP {1,2,4,8}] [--metric {throughput,bandwidth,time}] [--model MODEL] [--layout {TN,TT,NN,NT}]
                       [--output_file OUTPUT_FILE]
 
Model benchmarking tool
 
options:
  -h, --help            show this help message and exit
  --M M [M ...]         Input size M (seq_len for RoPE). Accepts:
                          Single value:            --M 512
                          Multiple values:         --M 256 512 1024
                          Range start:stop:step:   --M 128:1024:128
                          Combinations of values and ranges are also accepted.
                        Default: 4096.
  --TP {1,2,4,8}        Tensor parallel size. Default: 8.
  --metric {throughput,bandwidth,time}
                        Metric to report (throughput=TFLOPS, bandwidth=GB/s, time=ms). Default: throughput. RoPE reports total flops (total floating-point operations) in a separate column (see note in output).
  --model MODEL         model name filter: case-insensitive regex matched against model name (default: all models). e.g. 'llama3' to include only Llama3 family, 'llama|qwen' to include both Llama and Qwen families, '^(?!.*deepseek)' to exclude DeepSeek family.
                        Available models: Llama3 405B, Llama3 70B, Llama3 8B, GPT-OSS 120B, DeepSeek-R1, Llama4 Maverick, Qwen3-235B-A22B.
  --layout {TN,TT,NN,NT}
                        GEMM layout. Default: TN.
  --output_file OUTPUT_FILE
                        Name for the CSV output file. Default: bench_results.
```

Description of output CSV file:

CSV   column | Description | Type | Relevant kernels
-- | -- | -- | --
Model | LLM   model name | String | All
Kernel | kernel   name | String | All
B | Batch   size | Integer | Batched   GEMM
M/S | M   dimension / sequence length | Integer | All
N | N   dimension | Integer | GEMM,   batched GEMM, RMSNorm
K | K   dimension | Integer | GEMM,   batched GEMM
layout | GEMM   layout | String | GEMM,   batched GEMM
Q | Number   of Q heads per KV head | Integer | RoPE
H | Number   of KV heads | Integer | RoPE
D | QKV   head size | Integer | RoPE
rotate_style | RoPE   rotate style | String | RoPE
E | Number   of experts | Integer | MoE
Dim1 | First   MoE dimension (i.e. hidden_size) | Integer | MoE
Dim2 | Second   MoE dimension (i.e. moe_intermediate_size*2) | Integer | MoE
TopK | Number   of experts per token | Integer | MoE
Performance metric | Can   be time in ms, throughput in tflops or bandwidth in gpbs | Decimal | GEMM,   batched GEMM, RMSNorm, MoE
total_flops(tflops) | Total floating-point   operations | Decimal | RoPE

## Submission Checklist


lucas-santos-amd:
Added support for RMSNorm and RoPE kernels

op_tests/op_benchmarks/triton/bench_batched_gemm_a8w8.py
```diff
@@ -139,17 +139,17 @@ def parse_args():
...
-    args, defaults = parse_args()
-    if args.print_vgpr:
+def main(args: list[str] | None = None) -> None:
+    parsed_args, defaults = parse_args(args=args)
+    if parsed_args.print_vgpr:
         print("Retrieving VGPR usage for Triton kernels...")
-        fun = lambda: run_benchmark(args, defaults)  # noqa: E731
+        fun = lambda: run_benchmark(parsed_args, defaults)  # noqa: E731
         print_vgpr(fun, get_caller_name_no_ext())
         return 0
```
brunomazzottiamd: **Non-blocker nitpick:**

This `return 0` statement is useless. The script entry-point does `sys.exit(main())`, which can be evaluated to `sys.exit(0)` if `parsed_args.print_vgpr` or `sys.exit(None)` otherwise. `sys.exit(None)` is treated as `sys.exit()`, which exits with success, i.e. `0` status code.

I'd simply call `main`, without calling `sys.exit()`, to simplify everything.

brunomazzottiamd: I think this pattern is replicated over and over again... Maybe it's a nice opportunity to do some cleanup. What do you think?

op_tests/op_benchmarks/triton/bench_moe_gemm_a8w4.py
```diff
@@ -271,7 +271,7 @@ def roofline_mlp(
 ):
     out_path = Path(f"logs/{name}/{x_dtype}x-{w_dtype}w-TP{TP}/")
     out_path.mkdir(parents=True, exist_ok=True)
-    csv_path = compute_roofline(
+    compute_roofline(
```
brunomazzottiamd, line 274: **Question:**

I didn't get what happened here... Unused return value? Does `compute_roofline` return anything?

If `compute_roofline` actually returns something and we aren't using the returned value prefer `_ = compute_roofline(...)` idiom.

brunomazzottiamd: I see this pattern again in `op_tests/op_benchmarks/triton/bench_moe_gemm_a8w8.py`. I'll keep just this discussion for brevity sake, but please let's resolve the issue everywhere.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
+from abc import ABC, abstractmethod
```
brunomazzottiamd, line 1: Let's rename `bench_models_script.py‎` to just `bench_models.py‎`. It's shorter and keeps the purpose perfectly understandable. `_script` suffix is just pollution IMHO.

op_tests/op_benchmarks/triton/bench_moe_gemm_a8w4.py
lucas-santos-amd: `compute_roofline` indeed doesn't return anything, maybe it was changed and whoever changed it forgot to remove the `csv_path =` assignment.

brunomazzottiamd: That's fine. Let's close this discussion.

op_tests/op_benchmarks/triton/bench_batched_gemm_a8w8.py
brunomazzottiamd: Since you've committed to do the cleanup, replace the `return 0`'s by plain `return`'s.

op_tests/op_benchmarks/triton/model_benchmarking_tool/model_shapes.json
```diff
@@ -0,0 +1,504 @@
+{
```
brunomazzottiamd, line 1: **Suggestion:**

Create a `model_shapes.md` file, briefly documenting the expected structure and fields of this JSON file. You can ask an AI tool to do the majority of the work for you (given the JSON file as input). Later on, you can fix the AI mistakes and improve on what's missing.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+import os
+import pandas as pd
+import json
+import re
+import aiter.ops.triton.utils._triton.arch_info as arch_info
+import matplotlib.pyplot as plt
+import argparse
+import sys
+
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
brunomazzottiamd: Why do you need this `import` hack?

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+from bench_rope import main as bench_rope_main
+
+
+def disable_aiter_logs() -> None:
+    logging.getLogger("aiter").disabled = True
+
+
+disable_aiter_logs()
+
+kernel_dict: dict[str, Callable[[list[str]], None]] = {
```
brunomazzottiamd: I think this is a constant, right? If this is the case then let's rename it to `ALL_CAPS`  to match Python standards (`KERNEL_DICT`).

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+            return f"--shape {B} {M} {N} {K} --metric {self._metric} --layout {self._layout}"
+        return f"--shape {M} {N} {K} --metric {self._metric} --layout {self._layout}"
+
+    def parse_stdout(self, stdout: str) -> float:
+        lines = [line for line in stdout.splitlines() if line.strip()]
+        data_line = lines[-1]
+        last_row_values = list(map(float, re.findall(r"-?\d+(?:\.\d+)?", data_line)))
+        if len(last_row_values) not in (5, 6):
+            raise ValueError(f"Unexpected GEMM bench output format: {last_row_values}")
+        return round(last_row_values[-1], 4)
```
brunomazzottiamd: Why are you rounding the performance metric? If it's just for presentation sake, I won't do it. Capture the performance metric as-is and leave rounding to the user of `bench_models.py`. For instance, the user can open the output CSV file in Excel and round as he/she whishes. Other related aspect: the number of decimal places that's best suited to the benchmark may vary according to unit (milliseconds, TFLOPS) or batch size.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+    if kernel == "rope":
+        return RopeKernelHandler()
+    if "gemm" in kernel and "moe" not in kernel:
+        return GemmKernelHandler()
+    raise ValueError(f"Kernel {kernel} not supported")
+
+
+def read_json(json_path: str) -> ModelShapesData:
+    script_dir = os.path.dirname(os.path.realpath(__file__))
+    with open(script_dir + json_path, "r") as f:
```
brunomazzottiamd: Do not concatenate paths with `+`, you can use `os.path.join` or `pathlib`'s `Path` `/` operator.

brunomazzottiamd: As far as I can understand, `bench_models.py` won't work in the absence of `model_shapes.json`. Add some assertion or error handling logic to warn the user about this unexpected condition.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+                jdf.drop(columns=["Model", "Kernel"])
+                .dropna(axis=1)
+                .to_string(index=False)
+            )
+
+    if (df["Kernel"] == "rope").any():
+        print(f"\n{ROPE_METRIC_NOTE}")
+
+    # Save results to CSV file
+    output_path = f"{os.path.dirname(os.path.realpath(__file__))}/{output_file}.csv"
```
brunomazzottiamd: Use `os.path.join` to concatenated filesystem entities.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+    args = parser.parse_args()
+    args.M = parse_M_values(args.M, parser)
+    if "all" in args.models and len(args.models) > 1:
+        parser.error("'all' cannot be combined with other models.")
+
+    return args
+
+
+def main() -> None:
+    data = read_json("/model_shapes.json")
```
brunomazzottiamd: Remove this leading `/` from file name.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -0,0 +1,501 @@
...
+    args = parse_args(available_models)
+
+    models = args.models
+    M_values = args.M
+    TP = args.TP
+    metric = args.metric
+    layout = args.layout
+    output_file = args.output_file
+
+    if "all" not in models:
```
brunomazzottiamd: Can you implement a simple model filtering logic instead of using all models or just a single model? Someone may be interested in all Llama models or Llama3 family.

Take a look at this: https://github.com/ROCm/aiter/pull/2111/changes#diff-483006d820b2f05f82cf46ba818b2f03dd6a44333c2534b5a934f046b6153523R666

I used a regex, but I think a contains logic may be good enough for this purpose.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
lucas-santos-amd: Apparently I don't need it anymore. During early stages I wasn't importing anything from "aiter.ops" and the repo root wasn't being added to sys.path, this was also before we found out about this behavior

brunomazzottiamd: Awesome!

op_tests/op_benchmarks/triton/model_benchmarking_tool/model_shapes.json
```diff
@@ -58,6 +58,14 @@
                 "positions": "true",
                 "rotate_style": "neox"
             }
+        ],
+        "mha": [
+            {
+                "hq": 128,
+                "khv": 8,
```
brunomazzottiamd: It's `"hkv"` instead of `"khv"`... it was a typo of mine, my bad...

azaidy (approved):
LGTM!

PR #2075, opened 2026-02-23
Use unreg path for custom all-reduce during CUDA graph capture

zyzshishui:
## Motivation
Same as https://github.com/sgl-project/sglang/pull/19162.

Super tiny fix, needed to be compatible with `torch_memory_saver`. Error path:
1. `torch_memory_saver` hook `hipMalloc` with `hipMemAddressReserve`+ `hipMemMap`(VMM APIs)
2. During `register_graph_buffers`, `hipIpcGetMemHandle`expect a ptr by `hipMalloc`but in fact, got buffer from `hipMemAddressReserve`+ `hipMemMap`
3. `hipIpcGetMemHandle` does not check and accept this invalid handle (invalid because Runtime API and VMM API use different Allocation Table), which I raised a fix [here](https://github.com/ROCm/clr/pull/266) 
4. Other ranks call `hipIpcOpenMemHandle(invalid_handle)` and fail, causing a hang during cuda graph capturing

## Submission Checklist

### DIFF
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -15,7 +15,6 @@
 * limitations under the License.
 """
 
-import os
 from contextlib import contextmanager
 from typing import Any, List, Optional, Union
 
@@ -56,6 +55,7 @@ class CustomAllreduce:
         group: ProcessGroup,
         device: Union[int, str, torch.device],
         max_size=8192 * 1024 * 8 * 2, # In allreduce 2stage writemode, use 2x tmp buffer
+        enable_register_for_capturing: bool = True,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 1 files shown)


aiter/dist/device_communicators/custom_all_reduce.py
```diff
@@ -147,6 +148,7 @@ def __init__(
         #     return
 
         self.disabled = False
+        self.tms_cudagraph = os.getenv("SGLANG_MEMORY_SAVER_CUDA_GRAPH", "0")
```
valarLip: prefix with SGLANG_MEMORY_SAVER_CUDA_GRAPH?

zyzshishui: Good catch. How about adding a tms_cudagraph parameter to `__init__` and parse the param from sglang? But this would need sglang using updated aiter. Any suggestions?

yushengsu-thu (approved):

aiter/dist/device_communicators/custom_all_reduce.py
valarLip: good idea, let's add "enable_register" param to __init__

zyzshishui: Added "enable_register_for_capturing" since this only control behavior for capturing not real calling. But again, we cannot make subsequent change before sglang bump up aiter's version. I will keep an eye on that

PR #2082, opened 2026-02-24
use regex to extract arch from rocminfo string

ahmed-bsod:
## Motivation
Fails to extract arch from rocminfo string in some cases and throws error

## Testing
Tested on gfx950 to make sure change doesn't break anything


valarLip:
can you please descript more about the background

valarLip (approved):

PR #2085, opened 2026-02-24
pa_decode_gluon_aot C++ api

amd-yilizhao:
### Motivation

aiter's Triton-based paged attention decode kernels are currently only accessible via Python. C++ inference frameworks cannot use them without a Python runtime in the serving path. This PR exposes the kernels as a C++ API to enable direct integration.

### Usage scenario

The API is packaged as a shared library (`pa_decode_gluon_aot_lib.so`). C++ callers link against it and invoke `aiter::pa_decode_gluon_aot()` with standard PyTorch C++ tensors.

### Performance implications

- The underlying GPU kernels are identical to the Python path -- no performance difference in kernel execution.
- The C++ dispatch path avoids Python interpreter and GIL overhead, which benefits latency-sensitive serving scenarios with high request throughput.
- First invocation per kernel variant incurs a one-time Triton JIT compilation cost; subsequent calls use cached kernel handles with negligible dispatch overhead.

### DIFF
--- a/csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot.cpp
+++ b/csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot.cpp
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <sstream>
 #include <stdexcept>
+#include <unordered_map>
 
 namespace py = pybind11;
 
@@ -22,6 +23,7 @@ namespace aiter {
 static std::unique_ptr<LRUCache<std::string, PaDecodeCacheEntry>> g_kernel_cache;
 static std::once_flag init_kernel_cache_flag;
 static std::mutex g_cache_mutex;
+static std::unordered_map<std::string, std::shared_ptr<std::mutex>> g_key_mutexes;
 
 static CachedKernel load_cached_kernel(
     const py::bytes& hsaco_bytes,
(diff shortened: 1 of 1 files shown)


amd-yilizhao:
> Hi, I found some issues, pls see:
> 
> 1. Locking / potential deadlock?
>    In csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot.cpp, could the combination of the GIL acquisition around line ~63 and the mutex lock around line ~187 lead to a lock-order issue or deadlock in some scenarios?
> 2. Thread blocking during warmup
>    Lines ~187–200 lock a global mutex and then call warmup_and_load() on a cache miss. If one thread triggers a compile, all other threads are blocked on that lock, effectively serializing compilation. This can cause long tail latency during cold start.
> 3. Some input validation
>    aiter::get_cdna_version() returns -1 on unsupported architectures, but the C++ path doesn’t seem to handle this case.
>    pa_decode_gluon_aot() could use additional argument checks similar to the Python path, eg:
>    query.size(0) % query_length == 0
>    num_query_heads % num_kv_heads == 0
>    required tensors are on the same device / correct dtype / expected dims
>    These checks would help avoid silent failures or hard-to-debug runtime errors.

Thanks for the review. These four points have been addressed:

1. **Locking / potential deadlock**: Restructured so that `warmup_and_load()` (which acquires the GIL) runs entirely outside the mutex scope. The mutex is now only held briefly for cache get/put, eliminating the GIL-mutex lock-order inversion.

2. **Thread blocking during warmup**: As a result of (1), concurrent threads are no longer serialized during compilation. A double-check-then-insert pattern after warmup ensures correctness when multiple threads compile the same kernel concurrently.

3. **CDNA version check**: Added `TORCH_CHECK(cdna_ver == 3 || cdna_ver == 4, ...)` to catch unsupported architectures (including the `-1` fallback) with a descriptive error message.

4. **Input validation**: Added comprehensive `TORCH_CHECK` guards: CUDA tensor checks, dimension checks (query 3D, output 3D, key_cache 5D, value_cache 4D/5D), `query.size(0) % query_length == 0`, `num_query_heads % num_kv_heads == 0`, `query_length` range, `kv_block_size` whitelist, and `query_length * query_group_size <= 64`.

PR #2100, opened 2026-02-24
Fix CI prebuild: use build_ext so kernels are actually compiled

### LINKED ISSUE #2099: CI prebuild step not actually compiling kernels - causes JIT timeouts on GPU nodes
## Summary

The `build_aiter_image` job in the **Aiter Test** workflow is not actually prebuilding kernels. The Docker image gets pushed without any compiled kernel `.so` files, causing all kernels to be JIT-compiled at runtime on GPU nodes. This leads to timeouts in downstream jobs (e.g., ATOM benchmark timing out after 30 minutes).

## Root Cause

PR [#1973](https://github.com/ROCm/aiter/pull/1973) ("Defer expensive build operations to build_ext.run()") merged on **Feb 12** moved heavy imports and build logic into `NinjaBuildExtension.run()`, which only executes during actual build commands. However, `python setup.py develop` in the Dockerfile is intercepted by pip/pyproject.toml's build backend, which performs a lightweight **editable install** that never calls `build_ext.run()`. The prebuild logic is therefore completely skipped.

The setuptools deprecation warning confirms this:

```
DevelopDeprecationWarning: develop command is deprecated.
Please avoid running ``setup.py`` and ``develop``.
```

The result is a 4KB editable wheel with **zero compiled kernels**:

```
Building editable for amd-aiter (pyproject.toml): started
Building editable for amd-aiter (pyproject.toml): finished with status 'done'
Created wheel for amd-aiter: ... size=4349   # <-- no compiled kernels
```

## When It Broke

| Run | Date (UTC) | Commit | Prebuild Duration | Status |
|---|---|---|---|---|
| [21931680895](https://github.com/ROCm/aiter/actions/runs/21931680895) | Feb 12 02:53 | `7c09c7e` (before #1973) | **~40 min** | Working |
| [21932964701](https://github.com/ROCm/aiter/actions/runs/21932964701) | Feb 12 06:19 | `d96aff2` (after #1973) | **~20 sec** | **Broken** |

The breaking commit is [`2616865`](https://github.com/ROCm/aiter/commit/2616865d1a) from PR [#1973](https://github.com/ROCm/aiter/pull/1973), merged Feb 12 03:23 UTC.

## Evidence

**Build job (broken)** — [job link](https://github.com/ROCm/aiter/actions/runs/22367816402/job/64741600353):
- `setup.py develop` finishes in ~8 seconds total
- No `hipcc` invocations in the log
- "Prebuilding kernels completed" printed but no kernels were built

**Downstream failure** — [job link](https://github.com/ROCm/aiter/actions/runs/22360806399/job/64713677039):
- ATOM benchmark on DeepSeek-R1-0528 (MI325, tp=8) times out after 30 minutes
- Logs show JIT compilation at runtime: `[aiter] start build [module_rope_pos_fwd]`
- 7 GPU ranks blocked waiting on JIT build lock: `[aiter] waiting for baton release at .../lock_module_rope_pos_fwd`
- Multiple `not found tuned config` warnings confirming no prebuilt kernels available

## Impact

- **Every CI run since Feb 12** has been shipping Docker images without prebuilt kernels
- GPU test nodes waste time on JIT compilation instead of running tests
- Jobs with timeouts (e.g., ATOM benchmark at 30 min) fail because JIT compilation takes too long
- All GPU ranks except one sit idle waiting on the build lock during JIT

## Suggested Fix

The prebuild logic needs to actually execute during the Docker image build. Options:

1. **Separate the prebuild step** from `setup.py develop` — run the kernel compilation as a standalone script in the Dockerfile after installation
2. **Invoke `build_ext` explicitly** — add `python setup.py build_ext --inplace` before or after `setup.py develop` so `NinjaBuildExtension.run()` actually fires
3. **Move the prebuild logic** into a dedicated build script (e.g., `python -m aiter.prebuild`) that's called explicitly in the Dockerfile after installation

okakarpa:
Since PR #1973 (Feb 12), `setup.py develop` no longer triggers NinjaBuildExtension.run() because pip intercepts it as an editable install via pyproject.toml. This means the Docker image ships without any prebuilt kernels, causing JIT compilation at runtime on GPU nodes and 30-min timeouts in ATOM benchmarks.

Split into explicit `build_ext --inplace` (which fires the prebuild logic) followed by `pip install -e .` for the editable install.

Fixes #2099

## Submission Checklist

### DIFF
--- a/.github/workflows/atom-test.yaml
+++ b/.github/workflows/atom-test.yaml
@@ -117,7 +117,7 @@ jobs:
           --ulimit memlock=-1 \
           --ulimit stack=67108864 \
           -e ATOM_DISABLE_MMAP=true \
-          -e HF_TOKEN=${{ secrets.HF_TOKEN_TEST }} \
+          -e HF_TOKEN="${HF_TOKEN:-${{ secrets.HF_TOKEN_TEST }}}" \
           -v "${{ github.workspace }}:/workspace" \
           -w /workspace \
           --name atom_aiter_test \


gyohuangxin:
After discussion with @amd-ruitang3 and @zufayu , the root cause should be the combination with setuptools version (8.90.0) used in CI and the [PR](https://github.com/ROCm/aiter/pull/1973) changes, the PREBUILD variable cannot be read anymore, causing prebuild to be skipped. The Aiter team plans to migrate from setup.py to pyproject.toml. 

Therefore, I will merge this workaround to make prebuild functionality works in CI, and will apply the new method after migration.

@okakarpa Thanks for your fix.

PR #2143, opened 2026-03-01
fix: quant dispatch bugs and dead code cleanup in quant.py

brucechanglongxu:
Spotted a few issues while reading through `aiter/ops/quant.py` — all in the dispatch helpers and one function body. Single-file change, no new dependencies.

`get_torch_act` uses `dict.get(aType, NotImplementedError)` as its fallback, which returns the class object itself rather than raising. Callers in `fused_moe.py` and `fused_moe_dp_shared_expert.py` do `torch_act = get_torch_act(activation)` and then call `torch_act(x)` — for an unsupported activation type this ends up constructing `NotImplementedError(tensor)` and returning it as a value instead of blowing up. The sibling `get_*_quant` functions already handle this properly with a local closure that raises, so I matched that pattern.

`get_dtype_max` has a bare `except:` around the `torch.finfo` / `torch.iinfo` fallback. The only expected failure is `TypeError` (finfo on an integer dtype), but bare except also catches `KeyboardInterrupt`, `SystemExit`, etc. Narrowed it to `except TypeError:`.

`per_token_quant_hip` had an `if 1:` guarding the active code path with unreachable `elif`/`else` branches — the `elif` was an older smoothquant-based implementation that's been superseded by `dynamic_per_token_scaled_quant`. Removed the dead branches.

`get_hip_quant` keyed its dispatch dict on `QuantType.*.value` (raw ints) and looked up with `qType.value`, while `get_torch_quant` and `get_triton_quant` both key on enum members directly. Not a bug, but inconsistent for no reason — switched to enum members to match.

All changes are behavior-preserving on valid inputs. The only difference is on error paths: unsupported activation types now raise properly, and `get_dtype_max` won't swallow signals. Existing `op_tests/test_quant.py` covers the valid dispatch paths.

### DIFF
--- a/aiter/ops/quant.py
+++ b/aiter/ops/quant.py
@@ -33,7 +33,7 @@ def moe_smoothquant_fwd(
 def get_dtype_max(dtype):
     try:
         dtypeMax = torch.finfo(dtype).max
-    except:
+    except TypeError:
         dtypeMax = torch.iinfo(dtype).max
     return dtypeMax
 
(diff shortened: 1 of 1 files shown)


zufayu:

PR #2169, opened 2026-03-04
Fix LRU cache pollution causing BLOCK_SIZE_S3 KeyError in gemm_afp4wfp4

Duyi-Wang:
## Motivation

Fix `KeyError: 'Keyword argument BLOCK_SIZE_S3 was specified but unrecognised'` crash in `gemm_afp4wfp4` triggered when using `quark_w4a4_mxfp4` quantization in SGLang.

## Technical Details

`get_gemm_config()` in `aiter/ops/triton/utils/gemm_config_utils.py` uses `@functools.lru_cache`, which caches the returned `(dict, bool)` tuple by reference. The fused kernel `fused_gemm_afp4wfp4_split_cat` mutates the cached dict in-place by injecting `BLOCK_SIZE_S3`:

```python
config["BLOCK_SIZE_S3"] = triton.next_power_of_2(...)
```

Subsequent calls to `gemm_afp4wfp4` with the same `(M, N, K)` receive the same polluted dict from the cache. When unpacked via `**config` into a Triton kernel that doesn't accept `BLOCK_SIZE_S3`, it raises a `KeyError`.

**Fix:** Added `config.pop("BLOCK_SIZE_S3", None)` in the three affected functions in `gemm_afp4wfp4.py` (`gemm_afp4wfp4_`, `gemm_afp4wfp4_preshuffled_scales`, `gemm_afp4wfp4_preshuffle`) immediately after obtaining the config dict. This strips the polluted key before it reaches the kernel call, regardless of whether the cache was polluted or not.

This is a minimal workaround. A more thorough fix would be to have `get_gemm_config()` return defensive copies to prevent cache mutation entirely.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/gemm/fused/fused_gemm_afp4wfp4_split_cat.py
+++ b/aiter/ops/triton/gemm/fused/fused_gemm_afp4wfp4_split_cat.py
@@ -73,7 +73,6 @@ def fused_gemm_afp4wfp4_split_cat(
 
     if config is None:
         config, _ = _get_config(M, N, K)
-    config = dict(config)
 
     c1 = torch.empty((M, D, S1 + S3), dtype=dtype, device=x.device)
     c2 = torch.empty((M, D, S2), dtype=dtype, device=x.device)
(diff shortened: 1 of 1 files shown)


Duyi-Wang:
Similar one merged #2173. However, this is the same workaround and is only applicable to gemm_afp4wfp4. @k50112113

PR #2178, opened 2026-03-04
Add new GEMM configs for GPT-OSS

apicciau:
## Motivation

When running GPT-OSS, many messages regarding missing configurations appear on screen.
```
not found tuned config in /tmp/aiter_configs/bf16_tuned_gemm.csv ...
```
which means those GEMMs are implemented naively in Pytorch. This PR addresses this issue and provides more configs.

## Technical Details

The untuned GEMMs were collected with `AITER_TUNE_GEMM=1` and re-tuned with gradlib's `gemm_tuner.py`. Also, the the script `.github/scripts/op_tune.sh` was run to retuned batched and other GEMM families for MI355.

## Test Plan

ATOM was benchmarked with [this script](https://github.com/kimbochen/bench_serving.git) with the InferenceX parameters:
```
ISL=1024
OSL=8192
CONC_VALUES=(4 8 16 32 64 128)
```

## Test Result

image
This translates to about 14%-13% throughput uplift for concurrency values of 64 and 128

image
This translates to about 12%-13% end-to-end latency reduction for concurrency values of 64 and 128


apicciau:
With the latest changes in the main branch, the benefit of this approach is no longer clear, so I’m closing the PR and switching to a different strategy.

PR #2179, opened 2026-03-04
Adds the ability to build static archives in addition to shared objects 

Micky774:
## Motivation

TE currently relies on `libmha_{b, f}wd.so` to provide AITER symbols at runtime, however if other projects (e.g. pytorch) also build the AITER library, but of a different version, then loading the same symbols can create conflict. To avoid this, we at TE would like to be able to generate static archives as part of the build triggered through `aiter/op_tests/cpp/mha/compile.py`.

## Technical Details

Gated by a new environment variable `AITER_BUILD_STATIC`, there is now a new rule inserted which will build `libmha_*wd.a` using `AITER_AR_BIN` (set to `ar` by default). Backwards compatibility is ensured by `AITER_BUILD_STATIC` being defaulted to `0`.

## Test Plan

Run `aiter/op_tests/cpp/mha/compile.py` and validate that the expected artifacts are produced.

## Submission Checklist

### DIFF
--- a/aiter/jit/utils/cpp_extension.py
+++ b/aiter/jit/utils/cpp_extension.py
@@ -1720,10 +1720,14 @@ def _write_ninja_file(
     else:
         devlink_rule, devlink = [], []
 
+    aiter_build_static = os.environ.get("AITER_BUILD_STATIC", "0")
     emit_static_archive = (
-        library_target is not None
-        and os.environ.get("AITER_BUILD_STATIC", "0") == "1"
-        and os.path.basename(library_target) in {"libmha_fwd.so", "libmha_bwd.so"}
+        library_target is not None and
+        aiter_build_static != "0" and
+        (
+            aiter_build_static in ("1", "all") or
+            os.path.splitext(library_target)[0] in aiter_build_static.split(",")
+        )
     )
 
     if library_target is not None:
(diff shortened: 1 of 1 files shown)


aiter/jit/utils/cpp_extension.py
wangye805, line 1603: Can we detect whether it's .a or .so based on the library target ext?

aiter/jit/utils/cpp_extension.py
```diff
@@ -1723,24 +1724,55 @@ def sanitize_flags(flags):
     else:
         devlink_rule, devlink = [], []
 
+    emit_static_archive = (
+        library_target is not None
+        and os.environ.get("AITER_BUILD_STATIC", "0") == "1"
+        and os.path.basename(library_target) in {"libmha_fwd.so", "libmha_bwd.so"}
```
wangye805: Probably we can add some argument/option and pass the library_target down from compile.py in op_tests/cpp/mha as libmha_fwd/bwd.a instead of libmha_fwd/bwd.so?

zufayu:
@Micky774 This PR has been inactive for over 2 months. Is the static-archive build support still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2180, opened 2026-03-04
[TRITON] Add support for splitk to A8W8 kernel

vgokhale:
## Motivation

A8W8 did not have support for splitk.

## Technical Details

Splitk is needed to get good performance on very small M dim GEMMs. The shapes tuned in this PR are requested by a customer.

Tuned using the GEMM tuning script. 

N | M | K=2048 | K=4096 | K=6144
-- | -- | -- | -- | --
128 | 16 | $4.891 \mu s$ | $6.095 \mu s$ | $6.777 \mu s$
128 | 64 | $4.490 \mu s$ | $6.536 \mu s$ | $7.698 \mu s$
128 | 128 | $4.531 \mu s$ | $6.895 \mu s$ | $7.537 \mu s$
256 | 16 | $4.370 \mu s$ | $6.094 \mu s$ | $6.856 \mu s$
256 | 64 | $4.490 \mu s$ | $6.775 \mu s$ | $7.497 \mu s$
256 | 128 | $4.571 \mu s$ | $7.377 \mu s$ | $8.179 \mu s$

## Test Plan

Unit test modified to support splitk

## Test Result

All UTs pass.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
+++ b/aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
@@ -172,8 +172,7 @@ def _gemm_a8w8_kernel(
 
         # Add bias (only when not splitting K; bias is added after reduce)
         if HAS_BIAS:
-            offs_bias = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N) % N
-            bias = tl.load(bias_ptr + offs_bias)
+            bias = tl.load(bias_ptr + offs_bn)
             accumulator = accumulator.to(bias_ptr.type.element_ty) + bias[None, :]
 
         c = accumulator.to(c_ptr.type.element_ty)


aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
```diff
@@ -87,107 +98,171 @@ def _gemm_a8w8_kernel(
...
+        offs_bn = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N
+        a_ptrs = a_ptr + (
+            offs_am[:, None] * stride_am + offs_k_split[None, :] * stride_ak
+        )
+        b_ptrs = b_ptr + (
+            offs_k_split[:, None] * stride_bk + offs_bn[None, :] * stride_bn
+        )
+
+        # Create pointers for the scale tensors and load them
+        offs_a_scale = pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M) % M
```
ahmed-bsod: this should be: 
`offs_a_scale = (pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)) % M`

and same for offs_b_scale

aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
```diff
@@ -87,107 +98,171 @@ def _gemm_a8w8_kernel(
     tl.assume(stride_cm > 0)
     tl.assume(stride_cn > 0)
 
+    GRID_MN = tl.cdiv(M, BLOCK_SIZE_M) * tl.cdiv(N, BLOCK_SIZE_N)
```
ahmed-bsod: We already compute GRID_MN in triton heuristics so idt we need this here?

aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
```diff
@@ -87,107 +98,171 @@ def _gemm_a8w8_kernel(
...
     tl.assume(pid_n >= 0)
+    tl.assume(pid_k >= 0)
+
+    if (pid_k * SPLITK_BLOCK_SIZE) < K:
+
+        num_k_iter = tl.cdiv(SPLITK_BLOCK_SIZE, BLOCK_SIZE_K)
+
+        # Create pointers for first block of A and B input matrices
+        offs_k = tl.arange(0, BLOCK_SIZE_K)
+        offs_k_split = pid_k * SPLITK_BLOCK_SIZE + offs_k
```
ahmed-bsod: also the kernel is failing for uneven k shapes with SPLIT_K > 1. I tried (1024, 1024, 1000) and SPLIT_K=2 and it fails. Initially I thought the masking was wrong in reduce_kernel when storing to C but it is due to SPLITK_BLOCK_SIZE not being a multiple of BLOCK_SIZE_K. When K=1000 and K_SPLIT=2, SPLIT_K_BLOCK_SIZE=500 and BLOCK_SIZE_K=128. pid_k=0 handles 0..511 (even though logically it should only use 0...499) and pid_k=1 handles 500...1000. So there is an overlap of k=500..511 which causes the failures.

I think the potential solution is changing this line to:
`offs_k_split = pid_k * num_k_iter * BLOCK_SIZE_K + offs_k`
so pid_k=1 starts from 512 instead of 500.

I tried a bunch of diff shapes and this seems to work. But you can double check and see this doesn't introduce a new bug.

vgokhale: Good catch. Looks like this bug also exists on gemm a16w16. Fixed.

aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
vgokhale: Removed.

aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
vgokhale: Thanks. I instead just reused the offsets computed for am and bn.

aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8w8.py
```diff
@@ -90,104 +101,165 @@ def _gemm_a8w8_kernel(
...
+            # Advance the ptrs to the next K block.
+            a_ptrs += BLOCK_SIZE_K * stride_ak
+            b_ptrs += BLOCK_SIZE_K * stride_bk
+
+        # Apply scale
+        accumulator *= a_scale[:, None] * b_scale[None, :]
+
+        # Add bias (only when not splitting K; bias is added after reduce)
+        if HAS_BIAS:
+            offs_bias = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N) % N
```
ahmed-bsod: `offs_bias = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N`

vgokhale: Done. This should just be offs_bn so I replaced it as well.

PR #2190, opened 2026-03-05
[TRITON] Qwen3 next unified attention fix

juuso-oskari:
Qwen 3 Next has kv_cache block_size=544 . This causes unified attention to crash as it tries to set TILE_SIZE=block_size. TILE_SIZE must be a power of 2, and this mismatch causes the crash.

This PR fixes this issue by changing the TILE_SIZE setting from:

```
TILE_SIZE = block_size # causes crash if block_size is not a power of 2, and can overflow
```
to
```
TILE_SIZE = min(triton.next_power_of_2(block_size), 64)
```
This PR also adds benching for unified attention kernel, in order to evaluate what should be the default TILE_SIZE. Based on this evaluation,  TILE_SIZE = min(..., 64) is chosen as default.


op_tests/op_benchmarks/triton/bench_unified_attention.py
```diff
@@ -0,0 +1,669 @@
...
+        )
+        key_cache = torch.randn(
+            num_blocks, block_size, num_kv_heads, head_size, dtype=dtype, device="cuda"
+        )
+        value_cache = torch.randn_like(key_cache)
+        cu_seqlens_q = torch.zeros(len(seqlens_q) + 1, dtype=torch.int32, device="cuda")
+        cu_seqlens_q[1:] = seqlens_q.cumsum(dim=0, dtype=torch.int32)
+        cu_seqlens_k = torch.zeros(len(seqlens_k) + 1, dtype=torch.int32, device="cuda")
+        cu_seqlens_k[1:] = seqlens_k.cumsum(dim=0, dtype=torch.int32)
+        block_tables = torch.randint(
```
cagrikymk: While randint is fine when num_blocks * block_size >> needed tokens, there might be some repetitions which might make the results less accuate. I suggest doing it this way, or something similar so that there are no repetitions in the block table:

```
    total_ind_count = num_seqs * max_num_blocks_per_seq
    values = torch.arange(0, total_ind_count, dtype=torch.int)
    values = values[torch.randperm(total_ind_count)]
    block_tables = values.view(num_seqs, max_num_blocks_per_seq).contiguous()
```

op_tests/op_benchmarks/triton/bench_unified_attention.py
```diff
@@ -0,0 +1,669 @@
...
+from op_tests.op_benchmarks.triton.utils.benchmark_utils import (
+    get_model_configs,
+    print_vgpr,
+    get_caller_name_no_ext,
+)
+from op_tests.triton_tests.attention.test_unified_attention import ref_paged_attn
+from op_tests.triton_tests.attention.test_pa_prefill import seed_everything
+
+
+def input_helper(
```
cagrikymk, line 21: I think this is not being used in the code if I am not missing something?

cagrikymk (dismissed):

op_tests/op_benchmarks/triton/bench_unified_attention.py
anhminhnguyenhoang: thanks, it's a good point!

op_tests/op_benchmarks/triton/bench_unified_attention.py
anhminhnguyenhoang: no, it's currently not. It's an alternative way of initializing the input referenced from `pa_prefill.py` for more realistic values but it doesn't make a difference nonetheless for the benchmark

PR #2223, opened 2026-03-09
[TRITON] Improve RDNA config selection for unified attention

skysnow2001:
## Motivation

Improve performance of unified attention for RDNA GPUs.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/attention/unified_attention.py
+++ b/aiter/ops/triton/attention/unified_attention.py
@@ -30,7 +30,7 @@ def select_2d_config(
     )
 
     TILE_SIZE = 32 if arch.name == "gfx1201" else 16 if arch.is_rdna else 64
-    waves_per_eu = 6 if arch.is_rdna else 2
+    waves_per_eu = 8 if arch.name == "gfx1151" else 6 if arch.is_rdna else 2
 
     max_num_stages_2d = 2 if head_size > 128 else 4
 


aiter/ops/triton/attention/unified_attention.py
```diff
@@ -21,35 +23,40 @@ def select_2d_config(
...
-    # in case head_size is large
-    max_num_stages_2d = 4
-    if head_size > 128:
-        max_num_stages_2d = 2
-    if all_decode == False:
-        num_stages_2d = 1
-        num_warps = 2
+
+    TILE_SIZE = 32 if arch.name == "gfx1201" else 16 if arch.is_rdna else 64
+    waves_per_eu = 6 if arch.is_rdna else 2
```
cagrikymk: Overall, it looks good to me but isnt 6 waves per eu too high?
In older Triton versions (< 3.6), waves_per_eu wasnt as strict and compiler was mostly using that as an upper bound. After 3.6, it tries to match this occupancy. So, I suggest checking if it leads to any spill with >=3.6.

Other than that, LGTM!

skysnow2001: I benchmarked this with Triton 3.6 and checked for VGPR spilling, and I didn't observe any vgpr spills with waves_per_eu=6. In fact, it gave better performance in my tests.

I also just added a specific value for Strix-Halo (gfx1151), where waves_per_eu=8 performs better.

Thanks!

PR #2240, opened 2026-03-10
[TRITON] Sagev2 patch

juuso-oskari:
This PR concerns the sage attention kernels (the vanilla and the mxfp4):

- Use int64 offsets to prevent B x S x H x D from overflowing int32 values
- Move the sage attention quantization kernels to their own files.
- Revert to un-fused way of quantization for perf boost for mxfp4 sage attention.


Chi-Chu319 (dismissed):

jcaraban (dismissed):

azaidy (approved):
LGTM!

PR #2267, opened 2026-03-13
[TRITON] Unified attention FP8 cleanup

cagrikymk:
This PR does the following:

1. Extend the FP8 scaling to FP8 x FP8 attention as well, not just bf16 x fp8
2. Add support for q scaling
3. Fuse the QKV descaling with the existing ops to hide the cost
4. Code cleanup
5. Extending tests to better check the correctness for the descaling/scaling
6. Update/simplify the kernel API while keeping the wrapper API same
7. Instead of inverting out_scale before kernel call and multiplying inside, divide by within the kernel, to not cause any extra torch call

Main goal is to be able to apply scaling for any FP8 input/output to better preserve accuracy, if the scales/descales provided, while not causing any extra VALU/type conversion.

TODO:
- [ ] Verify the logic matches what downstream tasks expect, such as:
  1. descaling implies tensor * descale
  2. scaling implies tensor / scale
  3. This matches the existing use, but verify
- [x] Verify the application of the descaling/scaling doesnt cause any extra VALU


op_tests/triton_tests/attention/test_unified_attention.py
```diff
@@ -91,21 +107,32 @@ def ref_paged_attn(
 @pytest.mark.parametrize("head_size", HEAD_SIZES)
 @pytest.mark.parametrize("block_size", BLOCK_SIZES)
 @pytest.mark.parametrize("sliding_window", [None, 256])
-@pytest.mark.parametrize("dtype", DTYPES)
-@pytest.mark.parametrize("soft_cap", [None, 10.0, 50.0])
+@pytest.mark.parametrize("soft_cap", [None, 50.0])
 @pytest.mark.parametrize("num_blocks", NUM_BLOCKS)
-@pytest.mark.parametrize("q_dtype", QDTYPES)
+@pytest.mark.parametrize(
+    "q_dtype, kv_dtype, use_kv_descale, use_out_scale, use_q_descale",
```
k50112113, line 113: the use_kv_descale and use_q_descale are automatically True if not bf16, I guess we can remove them and just use q_dtype and kv_dtype to determine the descale variables

k50112113: we can also replace use_out_scale with out_dtype

PR #2277, opened 2026-03-14
[Triton MoE] Add optimized Gluon kernel for AMD CDNA3 with K-dimension unrolling

jwu10003:
## Motivation

Performance optimization request from Ali

- topk: 8
- expert_num: 128
- gemm0: A = [8192, 4096], B = [128, 384, 4096], C = [65536, 384]
- gemm1: A = [65536, 192], B = [128, 4096, 192], C = [8192, 8, 4096]
- w8a8

## Technical Details

- Implements manual LICM since Gluon doesn't support automatic loop-invariant code motion
- Pre-loads A matrix data outside the N-dimension loop to reduce memory traffic
- Uses optimized BlockedLayout for efficient memory access patterns
- Supports FP8 and INT8 quantization with per-tensor or block-wise scaling

## Key Features

- Layout optimization for efficient memory access and MFMA operations
- K-dimension loop unrolling (up to 3 times) for small K sizes
- Support for larger BLOCK_N tile sizes (up to 1024)
- Manual loop-invariant code motion (LICM) for A matrix data loading
- Efficient use of AMD CDNA3 MFMA instructions

## Test Plan

```
pytest -v op_tests/triton_tests/moe/test_moe.py::test_fused_moe --junitxml=test-reports/triton.xml -s
```

### Performance Improvement
- Test case: test_fused_moe[False-False-dtype0-True-False-True-8192-4096-192-8-128]
- Kernel duration: 2.33ms -> 1.0ms (57% improvement)

## Submission Checklist


zufayu:
@jwu10003 This PR has been open since March 2026 with merge conflicts, no review activity, and no pushes to your fork since April. Is the CDNA3 K-unrolled Gluon MoE kernel still something you plan to pursue? If yes, please rebase onto current main. If there's no update within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

zufayu:

**Adds a gfx942-only Gluon MoE GEMM kernel that fully unrolls short-K (≤192) blocks into preloaded MFMA steps and silently reroutes the default fused_moe dispatch to it when N≥1024 and all buffers fit the 2GB buffer_load limit.**

Validation (deterministic): NOT RUN — the PR changes runtime code but ships no test target (triage found no collectible target; auto-validation disabled); hand-run on MI300X: plain-path tests pass, fp8 per-tensor and int8 match an fp32 reference, block-wise fp8 fails (first finding)
Perf (advisory): MEASURED — MI300X (gfx942), base vs head back to back, 30 samples median, M=8192/N=4096/K=192/E=128/topk=8: fp8 per-tensor 2.538ms → 1.034ms (2.45x), int8 2.448ms → 1.847ms (1.32x), bf16 2.631ms → 2.261ms (1.16x); the PR's 2.33ms → 1.0ms claim reproduced; no tested path regressed

🔴 [verified] `_gluon_fused_moe_unroll_k_kernel` in aiter/ops/triton/_triton_kernels/moe/moe_op.py loads block-wise fp8 scales at the wrong indices — the b_scale column offset uses `offs_bn * stride_bsn` where the base `_fused_moe_kernel` divides offs_bn by group_n first, and the k-group scale index advances by `BLOCK_SIZE_K // group_k` (=0 when group_k=128 > BLOCK_SIZE_K=64) instead of `(k * BLOCK_SIZE_K) // group_k` — while the new dispatch gate in aiter/ops/triton/moe/moe_op.py checks only arch/K/N/2GB and never block_shape, so any block-quantized fp8 caller on gfx942 with K≤192 and N≥1024 silently gets wrong output: measured on MI300X at the PR's own target shape (M=8192, N=4096, K=192, E=128, topk=8, block_shape=[128,128]) the base kernel matches an fp32 reference at 5e-4 max rel error while the Gluon kernel is off by up to 291x. **Author must** fix both scale-index computations or exclude block_shape != None from the dispatch gate.

⚠️ [verified] The perf evidence is attached to `test_fused_moe[False-False-dtype0-True-False-True-8192-4096-192-8-128]`, a case pytest cannot collect — the (8192, 4096, 192, 8, 128) shape is absent from the parametrize list and the fp8/int8 parametrize is commented out in op_tests/triton_tests/moe/test_moe.py — so the Gluon path added to aiter/ops/triton/moe/moe_op.py ships with zero collectible coverage of its fp8, int8 and block-wise paths (only the plain fp16/bf16 path is exercised by existing shapes), and the one path with a verified wrong-output bug is exactly the one no test can catch. **Author must** add the target shape and re-enable the fp8/int8 parametrize so the claimed case actually runs in CI.

⚠️ [verified] The PR no longer applies to main: commit 4f9c57d5e (#4833, 2026-08-25 "Remove legacy MOE code") deleted both aiter/ops/triton/moe/moe_op.py and aiter/ops/triton/_triton_kernels/moe/moe_op.py, so the fused_moe entry point this PR modifies no longer exists, and main has since grown its own Gluon gfx942 MoE home under aiter/ops/triton/_gluon_kernels/gfx942/moe/ — any CI result on this branch describes a tree that has moved. **Author must** rebase onto current main's MoE structure (or close and resubmit) before this change can be evaluated.

⚠️ [verified] The change silently reroutes the default fused_moe dispatch on gfx942 with no opt-in, but the PR carries no ci: downstream label, so the ATOM/SGLang/vLLM jobs that consume aiter MoE paths would never run pre-merge on a shared-path behavior change. **Reviewer should ask** for ci:all (or the minimal ci:atom/ci:sglang/ci:vllm set) green before merge.

📝 [verified] The new gate in aiter/ops/triton/moe/moe_op.py has no runtime kill switch: once merged, every gfx942 fused_moe call with K≤192 and N≥1024 takes the Gluon kernel, and a production shape hitting a defect (as block-wise fp8 already does) leaves serving stacks no fallback without a code change. **Author must** add an env-var gate or document the fallback path.

PR #2309, opened 2026-03-17
Fix mp_tuner exception handling that silently drops tuning results

samremes:
The mp_tuner refactor in PR #1680 introduced two bugs in the polling loop's exception handlers that cause shapes to be silently dropped during tuning:

1. Non-KeyError exceptions (process crash/segfault): task was marked completed but no dummy result was stored in result_dict, causing IndexError during result reconstruction that crashes the entire mp_tuner call and all subsequent batches.

2. KeyError exceptions (PID mapping issue after pool worker restart): task was never marked completed and pool_restart_needed was commented out, causing an infinite loop.

Fix both handlers to mirror the existing timeout handler: store dummy results in result_dict, mark tasks as completed, and trigger pool restart. Add a defensive guard in the reconstruction loop as well.

Also fix a latent bug in gemm_a8w8_blockscale_tune.py where tasks_data recorded cumulative task counts instead of per-shape counts.

Made-with: Cursor

## Submission Checklist

### DIFF
--- a/aiter/utility/base_tuner.py
+++ b/aiter/utility/base_tuner.py
@@ -481,7 +481,6 @@ class TunerCommon:
                     f"Retrying {len(retry_keys)} failed shapes (attempt {retry + 1}/{max_retries})"
                 )
                 self.failed = pd.DataFrame(columns=self.columns)
-                retry_batches = (len(retry_keys) + batch_size - 1) // batch_size
                 for j in range(0, len(retry_keys), batch_size):
                     retry_batch = retry_keys.iloc[j : j + batch_size].reset_index(drop=True)
                     all_results = self.tune(retry_batch, self.tunedf, args)


aiter/utility/mp_tuner.py
```diff
@@ -467,11 +467,25 @@ def add_dummy_result(k, results_list):
...
-                    failed_tasks.append((k, "timeout"))
+                    failed_tasks.append((k, "crash"))
+
+                    dummy_results = []
+                    add_dummy_result(k, dummy_results)
+                    result_dict[k] = (
+                        dummy_results if shape_grouped else [dummy_results[0]]
+                    )
                     completed_this_round.append((k, async_result))
+                    pool_restart_needed = True
```
yzhou103, line 488: pool_restart_needed should not be set True here, as it runs into here when the kernel or calling kernel error happens.

aiter/utility/mp_tuner.py
```diff
@@ -467,11 +467,25 @@ def add_dummy_result(k, results_list):
                 if is_mapping_error:
                     error_msg = f"[Mapping Error] Task {k} - Process PID not in GPU map (triggering pool restart): {error_type} - {e}"
                     dummy_failed_tasks.append((k, "mapping error"))
-                    # pool_restart_needed = True
```
yzhou103, line 470: I remembered I commented this line as the mapping error is actually caused by other problem so I didn't add it to completed_this_round.

PR #2282, opened 2026-03-15
Add gfx1250 support: GFX_MAP + default GEMM/MHA configs

sunway513:
## Summary
- Add GFX ID mapping (`18: "gfx1250"`) to `GFX_MAP` in `chip_info.py`
- Add 27 default config files for gfx1250 (26 base GEMM + 1 MHA), copied from gfx950

## Changes
1. `chip_info.py`: Add `18: "gfx1250"` to GFX_MAP
2. 27 new config files under `aiter/ops/triton/configs/`:
   - `gemm/gfx1250-*.json` (26 files) — base GEMM configs for all kernel types
   - `gfx1250-MHA-DEFAULT.json` — MHA attention config
   - **No N/K specialized tuning files** — those will come from GemmTuner in a follow-up PR
   - **No code changes to config loading logic** — just data files

## Why configs instead of code fallback
Per review feedback: explicit config files are preferred over silent fallbacks to avoid masking debug issues. 27 base configs (vs 116 total with N/K specializations) provides functional coverage without bloat.

## Verification on gfx1250
| Test | Before | After |
|------|--------|-------|
| test_gemm_a8w8_blockscale | KeyError crash | **126 passed** |
| test_batched_gemm_bf16 | KeyError crash | **196 passed** |
| test_ff_a16w16 | KeyError crash | **passing** |
| test_mha (dropout+small head) | FileNotFoundError | **passing** |
| test_fused_add_rmsnorm_pad | 216 passed | **216 passed** (sanity) |

## Test plan
- [x] Verified KeyError/FileNotFoundError eliminated on gfx1250
- [x] Verified no impact on existing gfx942/gfx950 configs
- [x] Sanity check: previously passing tests still pass
- [ ] Follow-up: gfx1250-tuned configs via GemmTuner (separate PR)
- [ ] Follow-up: N/K specialized configs for performance (separate PR)

PR #2315, opened 2026-03-17
[TRITON] Add gfx1250 support: GFX_MAP + default GEMM/MHA configs

vgokhale:
Port changes from ROCm/aiter#2282.


sunway513:
cc @valarLip

PR #2360, opened 2026-03-19
[TRITON] Unified attention FP8 cleanup

cagrikymk:
Reopening the PR from [#2267](https://github.com/ROCm/aiter/pull/2267) after git history related aiter/main changes.

This PR does the following:

1. Extend the FP8 scaling to FP8 x FP8 attention as well, not just bf16 x fp8
2. Add support for q scaling
3. Fuse the QKV descaling with the existing ops to hide the cost
4. Code cleanup
5. Extending tests to better check the correctness for the descaling/scaling
6. Update/simplify the kernel API while keeping the wrapper API same
7. Instead of inverting out_scale before kernel call and multiplying inside, divide by within the kernel, to not cause any extra torch call

Main goal is to be able to apply scaling for any FP8 input/output to better preserve accuracy, if the scales/descales provided, while not causing any extra VALU/type conversion.

TODO:
- [x] Verify the logic matches what downstream tasks expect, such as:
  1. descaling implies tensor * descale
  2. scaling implies tensor / scale
  3. This matches the existing use, but verify
- [x] Verify the application of the descaling/scaling doesnt cause any extra VALU

Related vllm: [triton_reshape_and_cache_flash](https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/ops/triton_reshape_and_cache_flash.py)
  
 Scaling KV before flushing to the cache, in vllm:
```
 value_tile = value_load / tl.load(v_scale)
 key_tile = key_load / tl.load(k_scale)
```
 
 In attention, we have:
```
 value_tile = value_load * tl.load(v_scale)
 key_tile = key_tile * tl.load(k_scale)
```
 
The same validation needs to be done for `q_scale `and `out_scale `as well.

Bench code and fp8 changes also provided by @Chi-Chu319  @juuso-oskari

### DIFF
--- a/op_tests/op_benchmarks/triton/bench_unified_attention.py
+++ b/op_tests/op_benchmarks/triton/bench_unified_attention.py
@@ -9,7 +9,6 @@ from aiter.ops.triton.utils.types import e4m3_dtype
 from op_tests.op_benchmarks.triton.utils.argparse import get_parser
 from op_tests.triton_tests.attention.test_unified_attention import ref_paged_attn
 
-
 FP8_TYPE = e4m3_dtype
 FP8_MAX = torch.finfo(FP8_TYPE).max
 
@@ -22,10 +21,7 @@ def default_benchmark_configs():
     head_dim = 128
     v_head_dim = head_dim
     configs = list(itertools.product(batch_sizes, n_heads, seq_len_q, seq_len_k))
-    return [
-        (bs, nh, nh, sq, sk, head_dim, v_head_dim)
-        for bs, nh, sq, sk in configs
-    ]
+    return [(bs, nh, nh, sq, sk, head_dim, v_head_dim) for bs, nh, sq, sk in configs]
 
 
 def quantize_to_fp8(tensor):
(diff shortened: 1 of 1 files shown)


op_tests/triton_tests/attention/test_unified_attention.py
```diff
@@ -91,21 +107,32 @@ def ref_paged_attn(
...
 @pytest.mark.parametrize("num_blocks", NUM_BLOCKS)
-@pytest.mark.parametrize("q_dtype", QDTYPES)
+@pytest.mark.parametrize(
+    "q_dtype, kv_dtype, use_kv_descale, use_out_scale, use_q_descale",
+    [
+        (torch.bfloat16, torch.bfloat16, False, False, False),
+        (e4m3_dtype, e4m3_dtype, True, True, True),
+        (e4m3_dtype, e4m3_dtype, True, False, True),
+        (torch.bfloat16, e4m3_dtype, True, False, False),
+        (torch.bfloat16, e4m3_dtype, True, False, False),
```
divakar-amd: (nit) Duplicated?

aiter/ops/triton/_triton_kernels/attention/unified_attention.py
```diff
@@ -358,10 +357,13 @@ def kernel_unified_attention_2d(
...
+    if v_descale is not None:
+        one_over_L = v_descale / L[:, None]
+    else:
+        one_over_L = 1.0 / L[:, None]
     acc = acc * one_over_L
-    if USE_FP8:
-        acc = acc * tl.load(out_scale)
+    if out_scale_ptr is not None:
+        acc = acc / tl.load(out_scale_ptr)
         acc = tl.clamp(acc, FP8_MIN, FP8_MAX)
```
divakar-amd, line 369: In case of fp8xfp8 attention, the final output of the kernel might still be in fp16/bf16. In that case, do we need to clamp the accumulator to fp8 range?

divakar-amd (commented):
A few comments. I'll check the vllm runs as well and get back.

aiter/ops/triton/_triton_kernels/attention/unified_attention.py
cagrikymk: Good point, it makes more sense to clamp only if the output dtype is fp8. My code couples that with the existence of out_scale, which might be a problem for certain cases.

Resolved!

op_tests/triton_tests/attention/test_unified_attention.py
cagrikymk: Addressed

PR #2396, opened 2026-03-20
[TRITON] Unified Attention V2

juuso-oskari:
# Motivation
This PR adds:

### 1) better per tensor fp8 quantization

actually allow per tensor fp8 scales, rather than assume them to be 1.0 or only active when Q is not fp8.

### 2) MLA 

i.e. allow q and k head dim to be "d+r" where as the v head dim is "dv". If "d+r", i.e. lora + rope is a power of two, this decoupling is all we do. If "d+r" is not a power of two, d is set as the previous power of two, and r is set as the excess, which gets processed separately. This way, if the "d+r" is for example 576 (as in original deepseek), d gets set as 512, and r as 64, and we do two tl.dots to accumulate the whole QK. We could instead load a next power of two (i.e. 1024), but this would result in massive excess loading and compute.

### 3) support thd layout for kv

we can emulate the cached layout (i.e. num_blocks, block_size, num_heads, head_size) with thd layout, by simply unsqueezing (k.unsqueeze(1)). We also don't need to do the paging anymore inside the loop as its a contiguous layout, we only need to set the base offset once to the start of the kv sequence before the loop by reading cu_seqlens_k (that needs to passed).

### 4) add non causal as option

all that needs to change is how we calculate the inner loop length and masking. Trivial change. 

### 5) sage v2 quantization scheme on top of the unified attention.

- Apply hadamard rotation to QK, to remove outliers, and quantize the QK to mxfp4 per token.
- quantize V to fp8 per channel.

qkv is expected to have shapes:
- q,  # [num_tokens, num_query_heads, head_size//2] uint8 as we pack 2 mxfp4 values in the same uint8
- k,  # [num_blks, blk_size, num_kv_heads, head_size//2] uint8 as we pack 2 mxfp4 values in the same uint8
- v,  # [num_blks, blk_size, num_kv_heads, head_size] fp8

we expect the scales to have shapes:
- q_scale,  # [num_tokens, num_kv_heads, head_size // 32] uint8
- k_scale,  # [num_blks, blk_size, num_kv_heads, head_size // 32] uint8
- v_scale,  # [num_kv_heads, head_size]

### TODO: How will this quantization scheme work in vllm?
we will provide a modified version of the following:
https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/ops/triton_reshape_and_cache_flash.py
 
The quantization and store process goes like this:
the k,v come in as thd tensors. 
They are stored into cache token by token based on slot_mapping (: (t,) tensor)
for a token index t_i, you get the cache offset (block_idx, block_offset) as:
slot_idx = slot_mapping[t_i]
block_idx = slot_idx // block_size
block_offset = slot_idx % block_size 
 
The sagev2 is easy because we can just do the per token quantization while we store to cache, and we can define a k_descale_cache as (num_blocks, blk_size, num_kv_heads, head_dim // 32) so the slot mapping can be used to map to that aswell.


zufayu:
@juuso-oskari This PR has been inactive for over 2 months. Is Unified Attention V2 still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2401, opened 2026-03-20
Fix kernel map collision on MGPU context

Micky774:
## Motivation

Note: This is **not necessary** with https://github.com/ROCm/aiter/pull/2221 so I do think ideally that gets merged instead. This is just required as long as we use the standard hip module load.

In TE's JAX integration, the MGPU tests fail due to multiple devices registering separate kernels in the `impl_ptr_map`, leading to devices attempting to launch kernels that may have come from other devices resulting in a mismatched device ordinal launch error.

## Technical Details

Adds device ID to map key

## Test Plan

Verify TE JAX integration tests pass

## Test Result

TE JAX integration tests do indeed pass

## Submission Checklist


zufayu:
Closing this PR since the preferred fix (#2221, SynchronizedCache) was merged on 2026-04-15, and as noted in the description this PR was only a stopgap while the standard hip malloc path was in use. The branch also has merge conflicts now. Happy to reopen if the MGPU kernel-map collision resurfaces on the new path.

PR #2441, opened 2026-03-24
[Triton MoE]: Add optimized Gluon kernel for AMD CDNA3 with K-dimension unrolling

jwu10003:
## Motivation

Performance optimization of MoE Int8 SmoothQuant GEMM kernel when the K dimension is small (K <= 192). This is common in the second layer (layer2) of MoE MLP blocks where the intermediate dimension is reduced.

This PR targets the **new MoE path** (`moe_op_gemm_int8_smoothquant`), adding a Gluon-optimized kernel that leverages AMD CDNA3 MFMA instructions for better performance on small K dimensions.

### Key Optimizations

- **Manual LICM (Loop-Invariant Code Motion)**: Pre-loads A matrix data outside the N-dimension loop to reduce redundant memory traffic, since Gluon doesn't support automatic LICM
- **K-dimension loop unrolling**: Fully unrolls the K loop (up to 3 iterations for K <= 192, BLOCK_K=64), eliminating loop overhead
- **Explicit BlockedLayout + MFMA**: Uses `gl.amd.AMDMFMALayout` with `v_mfma_i32_32x32x16_i8` instruction for optimal INT8 register usage
- **SUB_BLOCK_SIZE_N inner loop**: Processes large BLOCK_N (512/1024) in 64-wide chunks for better occupancy
- **Buffer load/store intrinsics**: Uses `gl.amd.cdna3.buffer_load/store` for efficient memory access

## Performance Results

Benchmarked on MI308X using `bench_moe_gemm_int8_smoothquant.py` with proton profiler (3x100 reps averaged).

Full MoE MLP stack (routing + layer1 GEMM + layer2 GEMM). Gluon only accelerates layer2.

### K=192 (--shape 4096 384, E_tot=8, E_act=2)

| Batch | Standard (us) | Gluon (us) | Speedup |
|---:|---:|---:|---:|
| 4096 | 299.2 | 255.1 | **1.17x** |
| 8192 | 538.9 | 426.5 | **1.26x** |

### K=128 (--shape 4096 256, E_tot=8, E_act=2)

| Batch | Standard (us) | Gluon (us) | Speedup |
|---:|---:|---:|---:|
| 4096 | 228.5 | 203.2 | **1.12x** |
| 8192 | 441.7 | 363.3 | **1.22x** |

### K=64 (--shape 4096 128, E_tot=8, E_act=2)

| Batch | Standard (us) | Gluon (us) | Speedup |
|---:|---:|---:|---:|
| 4096 | 211.9 | 154.0 | **1.38x** |
| 8192 | 391.6 | 256.4 | **1.53x** |

### Summary

- **Up to 1.53x end-to-end speedup** (K=64, batch=8192)
- Smaller K yields larger speedup (K=64 > K=128 > K=192)
- Gluon dispatched only when `M >= 4096`, no regression at smaller batches

### How to reproduce

```bash
python op_tests/op_benchmarks/triton/bench_moe_gemm_int8_smoothquant.py --shape 4096 384 --experts 8 2 --num-weight-inits 3
python op_tests/op_benchmarks/triton/bench_moe_gemm_int8_smoothquant.py --shape 4096 256 --experts 8 2 --num-weight-inits 3
python op_tests/op_benchmarks/triton/bench_moe_gemm_int8_smoothquant.py --shape 4096 128 --experts 8 2 --num-weight-inits 3
```

## Test Plan

```bash
pytest -v op_tests/triton_tests/moe/test_moe_gemm_int8_smoothquant.py -s
```

## Test Result

All 336 tests passed (including 32 new small-K test cases across all parameter combinations)

## Submission Checklist


jwu10003:
> We want to deprecate this path of moe. The new path uses helper files like this https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/moe/moe_op_gemm_int8_smoothquant.py or this https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/moe/moe_op_gemm_a8w8.py and each ubench provides an example of how to call the full moe stack including token routing and quants
> 
> https://github.com/ROCm/aiter/blob/569d6afdbad4a8c7784aa48ae18c7ebd3b455361/op_tests/op_benchmarks/triton/bench_moe_gemm_int8_smoothquant.py#L91
> 
> If this PR is related to sglang improvements is it possible to have sglang use the new moe path? We would welcome a gluon kernel there assuming improved performance over current triton kernels.

Thanks for the feedback! I have port to new path, could you please review the PR if you are free?

lburzawa:
@nsusanto maybe you could give quick look at the PR too since you worked on int8 as well

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_int8_smoothquant.py
```diff
@@ -224,7 +226,7 @@ def _moe_gemm_int8_smoothquant(
     Int8 MoE GEMM with SmoothQuant support and per-token per-channel scaling.
 
     SmoothQuant formula:
-        Y = (X diag(s)^-1) · (diag(s) W)
+        Y = (X diag(s)^-1) ? (diag(s) W)
```
brunomazzottiamd: I think some linter or formatter tool automatically removes characters it considers "weird" ones. IMHO, it harms the comments.

Check these:

* `Y = (X diag(s)^-1) ? (diag(s) W)` - what's `?` operator?
* `(dequant scale for X?)` - turns the sentence into a question and changes the variable (`X` instead of `X̂`)
* `(dequant scale for ?)` - dequant scale for what?

I won't request changes due to this, it's too much... However, it would nice if you could  rewrite the affected parts using non-"weird" characters, while preserving the mathematical meaning they are explaining.

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_int8_smoothquant.py
```diff
@@ -414,3 +416,352 @@ def _moe_gemm_int8_smoothquant(
...
+                mask=mask_m[:, None] & (offs_ak[None, :] < K - 2 * BLOCK_K),
+                other=0.0,
+            )
+
+    # Convert A to MFMA operand layout once
+    a0_converted = gl.convert_layout(a0, mfma_a_layout)
+    if UNROLL_TIMES > 1:
+        a1_converted = gl.convert_layout(a1, mfma_a_layout)
+    if UNROLL_TIMES > 2:
+        a2_converted = gl.convert_layout(a2, mfma_a_layout)
```
nsusanto: i think there are a lot of convert_layout calls which from my experience hits perf pretty bad. Can't we use LDS to load/store a and b tensors and load it in the correct layout? Is there a reason why we're not using LDS?

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_int8_smoothquant.py
```diff
@@ -414,3 +416,352 @@ def _moe_gemm_int8_smoothquant(
...
+                offs_x_m[:, None] * stride_x_m + offs_ak[None, :] * stride_x_k,
+                mask=mask_m[:, None] & (offs_ak[None, :] < K - 2 * BLOCK_K),
+                other=0.0,
+            )
+
+    # Convert A to MFMA operand layout once
+    a0_converted = gl.convert_layout(a0, mfma_a_layout)
+    if UNROLL_TIMES > 1:
+        a1_converted = gl.convert_layout(a1, mfma_a_layout)
+    if UNROLL_TIMES > 2:
```
nsusanto: can't we use gl.static_range to do the loop unrolling? I think we can probably do something 

```
for k in gl.static_range(UNROLL_TIMES):
    .....
```

jwu10003: Thanks for the suggestion!

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_int8_smoothquant.py
jwu10003: > i think there are a lot of convert_layout calls which from my experience hits perf pretty bad. Can't we use LDS to load/store a and b tensors and load it in the correct layout? Is there a reason why we're not using LDS?

Thanks for the suggestion! I evaluated this:

1. A operand: A is loop-invariant — it has been hoisted outside the N-dimension loop (LICM), so convert_layout only executes UNROLL_TIMES times per kernel invocation. No need to optimize.

2. B operand: This kernel targets small-K scenarios where compute is light and output store I/O dominates. Using explicit smem_b.store() + smem_b.load() allocates additional LDS, which reduces occupancy. Higher occupancy is critical here to hide output I/O latency via wavefront interleaving. I benchmarked 6 different LDS layout configs — all regressed by +5% to +57% .

jwu10003:
> Actually let's use "_gluon_kernels/gfx942/moe" to also recognize the arch of the gluon kernel, I edited my previous comment.

Thanks for the suggestion! Done.

lburzawa (approved):
Looks good, thanks for changes.

brunomazzottiamd:
I can approve this PR on behalf of **ROCm/team_aiter** if @lburzawa and @nsusanto think it's good! We still have an [unresolved discussion](https://github.com/ROCm/aiter/pull/2441#discussion_r3040496321) to be addressed.

nsusanto (approved):
LGTM!

nsusanto:
> I can approve this PR on behalf of **ROCm/team_aiter** if @lburzawa and @nsusanto think it's good! We still have an [unresolved discussion](https://github.com/ROCm/aiter/pull/2441#discussion_r3040496321) to be addressed.

I think the main argument against LDS usage is because the kernel targets small-K scenarios which actually slows down the kernel. I resolved the issue there.

brunomazzottiamd (approved):

PR #2454, opened 2026-03-24
added gfx1250 gemm_a8w8 kernel, test, and bench support

ahmed-bsod:
gfx1250 gemm_a8w8 kernel

### DIFF
--- a/aiter/ops/triton/gemm/basic/gemm_a8w8.py
+++ b/aiter/ops/triton/gemm/basic/gemm_a8w8.py
@@ -51,7 +51,7 @@ def gemm_a8w8(
     """
 
     _LOGGER.info(
-        f"GEMM_A8W8: x={tuple(x.shape)} w={tuple(w.shape)} x_scale={tuple(x_scale.shape)} w_scale={tuple(w_scale.shape)}"
+        f"GEMM_A8W8: x={tuple(x.shape)} w={tuple(w.shape)} x_scale={tuple(x_scale.shape)} w_scale={tuple(w_scale.shape)} use_gluon={use_gluon}"
     )
     
     if use_gluon:
--- a/aiter/ops/triton/gemm/basic/gemm_a8w8.py
+++ b/aiter/ops/triton/gemm/basic/gemm_a8w8.py
@@ -8,6 +8,9 @@ from aiter.ops.triton._triton_kernels.gemm.basic.gemm_a8w8 import (
     _gemm_a8w8_kernel,
     _get_config,
 )
+from aiter.ops.triton.gluon.gfx1250_gemm_a8w8 import (
+    gfx1250_gemm_a8w8 as _gluon_gemm_a8w8,
+)
 from aiter.ops.triton.utils.device_info import get_num_xcds
 
 from aiter.ops.triton.utils.logger import AiterTritonLogger
@@ -24,6 +27,7 @@ def gemm_a8w8(
     dtype: Optional[float] = torch.bfloat16,
     y: Optional[torch.Tensor] = None,
     config: Optional[dict] = None,
+    use_gluon: bool = False,
 ):
     """
     Computes 8 bit matrix multiplication Y = (X @ W^T) * (x_scale * w_scale) with optional bias.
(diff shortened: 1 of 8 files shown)


aiter/ops/triton/gemm/basic/gemm_a8w8.py
```diff
@@ -24,6 +27,7 @@ def gemm_a8w8(
     dtype: Optional[float] = torch.bfloat16,
     y: Optional[torch.Tensor] = None,
     config: Optional[dict] = None,
+    use_gluon: bool = False,
```
Boss2002n, line 30: something like 
use_gluon = arch_info.get_arch() == "gfx1250"
is better imo instead of passing it as a param

aiter/ops/triton/gemm/basic/gemm_a8w8.py
Boss2002n, line 80: maybe rename this fn to show that this is a triton fn, so that its easier to read/debug when needed
since the gluon kernel will also be called by this wrapper fn.

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,313 @@
...
+            instr_shape=[16, 16, 128],
+        )
+    elif NUM_WARPS == 2:
+        wmma_layout: gl.constexpr = gl.amd.AMDWMMALayout(
+            version=3,
+            transposed=True,
+            warp_bases=[[0, 1]],  # [1, 2] layout
+            reg_bases=[],
+            instr_shape=[16, 16, 128],
+        )
```
Boss2002n, line 127: something like 
    tiles_per_warp = 2
    if num_warps == 4:
        warp_bases = [[0, tiles_per_warp], [tiles_per_warp, 0]]
    else:
        warp_bases = [[0, tiles_per_warp], [0, tiles_per_warp * 2], [tiles_per_warp, 0]]
would be nicer imo (just a suggestion)

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
```diff
@@ -0,0 +1,313 @@
...
+        w.stride(0),
+        w.stride(1),
+        y.stride(0),
+        y.stride(1),
+        bias is not None,
+        NUM_WARPS=config["num_warps"],
+        **config,
+    )
+
+    return y
```
Boss2002n, line 313: I thought we were doing wrappers in the same place as triton wrappers?

aiter/ops/triton/utils/gemm_config_utils.py
Boss2002n, line 1: If we are separating gluon configs like this then I think we should also move triton configs and add that change as well. Not in this PR but if this is the way we are going about it then that would be nice to have

op_tests/op_benchmarks/triton/bench_gemm_a8w8.py
Boss2002n, line 1: IIRC there was a discussion to use the same wrapper calls for gfx1250 and the old ones - if thats the case we should handle which kernel it goes to in that function and will not need different benchmarking functions - we can just call the wrapper and the logic as to which kernel it should benchmark will be handled in the wrapper

op_tests/triton_tests/gemm/basic/test_gemm_a8w8.py
```diff
@@ -194,6 +197,16 @@ def test_gemm(in_dtype, out_dtype, m, n, k, layout, output, impl: str):
             "Gluon implementation is not supported on this device (requires CDNA4)."
         )
 
+    if impl == "gfx1250_gluon" and DEVICE_ARCH != "gfx1250":
+        pytest.skip(
+            "gfx1250 Gluon implementation is not supported on this device (requires gfx1250)."
+        )
```
Boss2002n, line 203: is_gluon_avail()
(aiter/ops/triton/utils/_triton/arch_info.py)
is available - lets use that to keep things uniform

op_tests/triton_tests/gemm/basic/test_gemm_a8w8.py
```diff
@@ -194,6 +197,16 @@ def test_gemm(in_dtype, out_dtype, m, n, k, layout, output, impl: str):
...
+
+    if impl == "gfx1250_gluon" and (in_dtype == "int8" or out_dtype == "int32"):
+        pytest.skip(
+            "gfx1250 Gluon implementation does not support int8 input or int32 output."
+        )
```
Boss2002n, line 208: This seems like something that could be used by others as well - worth it to add it to the respective utils file

Boss2002n (changes requested):

aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py
Boss2002n: image

PR #2489, opened 2026-03-26
Fix CPU/GPU device mismatch in _yarn_linear_ramp_mask

JohnQinAMD:
`_yarn_linear_ramp_mask` creates `torch.arange` on CPU (default device) but callers in `YaRNScalingRotaryEmbedding._compute_inv_freq` and `DeepseekScalingRotaryEmbedding._compute_inv_freq` use the result with GPU tensors (`pos_freqs` created with `device="cuda"`), causing:

  RuntimeError: Expected all tensors to be on the same device,
  but found at least two devices, cuda:N and cpu!

This manifests during CUDA graph capture in SGLang when serving models with DeepseekScaling/YaRN rotary embeddings (e.g., Kimi-K2.5).

Fix: Add `device` parameter to `_yarn_linear_ramp_mask` (default "cuda") and pass `device="cuda"` from both call sites.

Made-with: Cursor

## Submission Checklist

### DIFF
--- a/aiter/rotary_embedding.py
+++ b/aiter/rotary_embedding.py
@@ -595,12 +595,12 @@ def _yarn_find_correction_range(
 
 
 def _yarn_linear_ramp_mask(
-    low: float, high: float, dim: int, dtype: torch.dtype
+    low: float, high: float, dim: int, dtype: torch.dtype, device=None
 ) -> torch.Tensor:
     if low == high:
         high += 0.001  # Prevent singularity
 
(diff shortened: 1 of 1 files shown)


zufayu:
@JohnQinAMD This PR has been inactive for over 2 months. Is the CPU/GPU device mismatch fix in _yarn_linear_ramp_mask still applicable? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2490, opened 2026-03-26
[TRITON] Add kernel filter to the bench_models.py script

lucas-santos-amd:
## Motivation

Add a regex kernel filtering functionality to the bench_models.py script.

## Technical Details

The new `--kernel` arg expects a case-insensitive regex to be matched against all the available kernel names.
E.g. `'gemm'` to include any kernel name containing gemm, `'moe|rmsnorm'` for MoE and RMSNorm kernels, `'^(?=.*gemm)(?!.*batched)'` to include GEMM and exclude batched GEMM kernels.

## Submission Checklist

### DIFF
--- a/op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
+++ b/op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
@@ -605,10 +605,12 @@ def run_benchmarks(
                 )
                 continue
 
+            # MLA only reports time (ms)
+            run_metric = "time" if kernel == "mla" else metric
+
             bench_fn = KERNEL_DICT[kernel]
             handler = _get_handler(kernel)
-            # MLA only reports time (ms); use "time" metric for it
-            run_metric = "time" if kernel == "mla" else metric
+
             handler.set_run(model, kernel, run_metric, gemm_layout, mha_layout, TP)
             tp_shapes = handler.get_tp_shapes(shapes)
             for shape in tp_shapes:
@@ -662,7 +664,9 @@ def parse_arg_list(
     return sorted(set(result))  # Remove duplicates and sort
 
 
-def parse_args(available_models: list[str]) -> argparse.Namespace:
+def parse_args(
+    available_models: list[str], available_kernels: list[str]
+) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         description="Model benchmarking tool",
         allow_abbrev=False,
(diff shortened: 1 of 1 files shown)


op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -757,9 +771,11 @@ def parse_args(available_models: list[str]) -> argparse.Namespace:
 def main() -> None:
     data = read_json("model_shapes.json")
     available_models = list(data.keys())
```
brunomazzottiamd: **Suggestion:** I think it'd be nice to sort the model names as well, as you did with the kernel names.

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -778,6 +795,21 @@ def main() -> None:
         except re.error:
             print(f"Invalid model filter regex: {models!r} - running all models.")
 
+    if kernels is not None:
```
brunomazzottiamd: Can you please refactor the regex matching logic, to avoid code duplication?

I'm thinking about something like this:

**Filter function:**
```python
def filter_by_regex(pattern_str: str, candidates: list[str]) -> list[str] | None:
    """
    Filter *candidates* by a regex string (case-insensitive).

    Returns:
        A filtered list on success, the original list if the regex is invalid,
        or None if the regex is valid but nothing matched.
    """
    try:
        pattern: re.Pattern[str] = re.compile(pattern_str, re.IGNORECASE)
    except re.error:
        return candidates  # invalid regex, keep everything

    matched = [s for s in candidates if pattern.search(s)]
    return matched if matched else None # None signals "nothing survived"
```

**Model filtering:**
```python
if models is not None:
    matched_models = filter_by_regex(models, available_models)
    if matched_models is None:
        print("There are no models after filtering by model name.")
        return
    if matched_models is not available_models:  # regex was valid
        data = {m: data[m] for m in matched_models}
    # else: invalid regex, data unchanged
```

**Kernel filtering:**
```python
if kernels is not None:
    filtered: ModelShapesData = {}
    for m, kernel_map in data.items():
        matched_kernels = filter_by_regex(kernels, list(kernel_map))
        if matched_kernels is None:
            continue  # no kernels kept for this model
        kept_keys = set(matched_kernels) if matched_kernels is not list(kernel_map) else kernel_map
        filtered[m] = {k: v for k, v in kernel_map.items() if k in set(matched_kernels)}
    data = filtered
    if not data:
        print("There are no models/kernels after filtering by kernel name.")
        return
```

**Warning:** Please don't take this code for granted, I played a bit with Claude to draft a suggestion. Use your brain and test the suggested changes by yourself.

brunomazzottiamd (changes requested):
@lucas-santos-amd, can you please try to refactor the regex matching logic as suggested? Thanks!

op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py
```diff
@@ -768,15 +829,10 @@ def main() -> None:
...
-            pattern: re.Pattern[str] = re.compile(models, re.IGNORECASE)
-            data = {m: data[m] for m in available_models if pattern.search(m)}
-            if not data:
-                print("There are no models after filtering by model name.")
-                return
-        except re.error:
-            print(f"Invalid model filter regex: {models!r} - running all models.")
+    filtered_data = filter_models_and_kernels(data, available_models, models, kernels)
+    if filtered_data is None:
+        return
```
brunomazzottiamd, line 834: **Question:** I think `filter_models_and_kernels` function already emits proper error / warning messages. Am I right? Is this the reason you are just returning here?

lucas-santos-amd: Yes, you are correct. `filter_models_and_kernels` handles the error messages. The conditional on `main` is only for terminating the program.

PR #2498, opened 2026-03-27
Add ctypes C-ABI error bridging to prevent worker crashes during kern…

yzhou103:
…el tuning

AITER_CHECK and HIP_CALL now throw std::runtime_error instead of calling std::terminate()/exit(0), so exceptions can be caught at the C-ABI boundary.

New header aiter_ctypes_error.h provides:
- AITER_CTYPES_ERROR_DEF: per-TU thread-local error storage + ABI version probe
- AITER_CTYPES_DEFINE_ENTRYPOINT_VOID: macro that generates extern "C" int wrapper with automatic try/catch bridging (developer writes normal function body)
- aiter_safe_call: template that catches C++ exceptions, stores in TLS, returns -1

Python side (core.py) probes each .so for aiter_ctypes_abi_version to auto-detect the new int-returning convention and raises RuntimeError on failure.

This is mainly for tuning, not terminating the process when there is checking error.

## Submission Checklist


yzhou103:
This pr will check the error that some test not supported in mi355. fixed the tests in https://github.com/ROCm/aiter/pull/2553.
old:image
new:
image

csrc/include/aiter_hip_common.h
```diff
@@ -15,41 +15,44 @@
...
 {
-    std::cerr << "[AITER] " << file << ":" << line << " ";
-    (std::cerr << ... << std::forward<Args>(args)) << std::endl;
-    std::abort();
+    std::ostringstream oss;
+    oss << "[AITER] " << file << ":" << line << " ";
+    (oss << ... << std::forward<Args>(args));
+    std::string msg = oss.str();
+    std::cerr << msg << std::endl;
+    throw std::runtime_error(msg);
```
draganmladjenovic: std::move(msg)

csrc/include/aiter_hip_common.h
```diff
@@ -15,41 +15,44 @@
...
-    do                                                                  \
-    {                                                                   \
-        if(!(x)) [[unlikely]]                                           \
-        {                                                               \
-            detail::aiter_check_fatal(__FILE__, __LINE__, __VA_ARGS__); \
-        }                                                               \
+#define AITER_CHECK(x, ...)                                            \
+    do                                                                 \
+    {                                                                  \
+        if(!(x))                                                       \
```
draganmladjenovic: I would leave unlikely. You are using c++20?

csrc/include/aiter_hip_common.h
```diff
@@ -15,41 +15,44 @@
...
-                                      "fail to call " #call " ---> [HIP error](", \
-                                      hipGetErrorString(err),                     \
-                                      ')');                                       \
-        }                                                                         \
+#define HIP_CALL(call)                                                          \
+    do                                                                          \
+    {                                                                           \
+        hipError_t err = call;                                                  \
+        if(err != hipSuccess)                                                   \
+        {                                                                       \
```
draganmladjenovic: Call ((void)hipGetLastError()) to clear hip error sticky bit for the current thread. This wouldn't be needed if we were aborting. I'm torn of this one. This failing is usualy a bug on aiter side and should not just be shrugged off by the user.

yzhou103: You raise a valid point. In practice, HIP_CALL errors usually indicate real bugs and should not be silently ignored. The wrapping is opt-in — only applied to specific kernels used in tuning, where AITER_CHECK failures (unsupported config/shape/dtype) are expected and the tuner needs to skip and continue. HIP_CALL errors can also occur during tuning (e.g. hipModuleLoad file-not-found for missing .co files on certain architectures), so catching them in the wrapped path is still useful. For unwrapped functions, check_fail calls std::abort() directly.

csrc/include/aiter_hip_common.h
yzhou103: Done, Thanks

csrc/include/aiter_hip_common.h
yzhou103: Yes, i'm using C++20 (-std=c++20). Added [[unlikely]] to both AITER_CHECK and HIP_CALL

csrc/include/aiter_hip_common.h
yzhou103: I think you are right. HIP_CALL failures during tuning are unexpected, it should not continue to run.

PR #2505, opened 2026-03-27
[Triton] Add sliding window support for sink attention

kyle-256:
## Motivation

The Triton MHA path already supports sink attention, but it cannot be combined with a left sliding window today. That makes it difficult to validate and use sink + sliding-window attention in downstream integrations. This PR adds the missing left-window support while keeping the current restriction that `window_size_right` is still unsupported.

## Technical Details

- thread `window_size_left` through the Triton MHA wrapper into the forward kernel and one-kernel backward kernel
- keep rejecting `window_size_right != -1`, so this change only enables the supported left-window use case
- apply sliding-window masking in both forward and backward Triton kernels
- add guards for fully masked blocks to avoid NaNs when a tile falls completely outside the valid window
- add regression coverage in `op_tests/triton_tests/attention/test_mha.py` for sink + sliding window, and for sliding window without sink

## Test Plan

- `pytest op_tests/triton_tests/attention/test_mha.py -k test_mha_with_sink_sliding_window`

## Test Result

- `test_mha_with_sink_sliding_window` passed locally
- forward and backward results matched the reference implementation within the configured tolerances
- this PR also adds regression coverage for the left sliding-window path without sink in `test_mha.py`

## Submission Checklist

### DIFF
--- a/op_tests/triton_tests/attention/test_mha_with_sink.py
+++ b/op_tests/triton_tests/attention/test_mha_with_sink.py
@@ -41,8 +41,8 @@ def test_mha_with_sink(
     CAUSAL: bool,
 ):
     HAS_DROPOUT: bool = DROPOUT > 0.0
-    # Causal + Dropout use case is disabled in `test_mha_backward`.
-    # FIXME: We should fix it in the base implementation before adding sink to the mix.
+    # Keep sink coverage aligned with the baseline MHA tests.
+    # Causal + dropout backward is still disabled in `test_mha_backward`.
     TEST_BWD: bool = not (CAUSAL and HAS_DROPOUT)
     device: str = "cuda"
     dtype: torch.dtype = torch.bfloat16
@@ -176,8 +176,8 @@ def test_mha_varlen_with_sink(
     CAUSAL: bool,
 ):
     HAS_DROPOUT: bool = DROPOUT > 0.0
-    # Dropout use case is disabled in `test_mha_backward_varlen`.
-    # FIXME: We should fix it in the base implementation before adding sink to the mix.
+    # Keep sink coverage aligned with the baseline MHA tests.
+    # Dropout backward is still disabled in `test_mha_backward_varlen`.
     TEST_BWD: bool = not HAS_DROPOUT
     device: str = "cuda"
     dtype: torch.dtype = torch.bfloat16


op_tests/triton_tests/attention/test_mha.py
```diff
@@ -1560,12 +1560,247 @@ def test_mha_varlen_with_sink(
...
+
+@pytest.mark.parametrize("BATCH", [1, 2])
+@pytest.mark.parametrize(
+    "SEQLEN_Q, SEQLEN_K", [(64, 64), (128, 128), (256, 256), (128, 64), (32, 128)]
+)
+@pytest.mark.parametrize("NUM_Q_HEADS, NUM_K_HEADS", [(8, 1), (16, 4)])
+@pytest.mark.parametrize("HEAD_SZ", [32, 64])
+@pytest.mark.parametrize("CAUSAL", [True])
+@pytest.mark.parametrize("WINDOW_SIZE_LEFT", [4, 16, 64])
+def test_mha_with_sink_sliding_window(
```
brunomazzottiamd: `test_mha.py` is growing quickly, there are a lot of MHA flavors in this test file...

Can we move all the sink related tests to a new file? Maybe `test_mha_with_sink.py`? I think this can benefit CI granularity and help identify possible issues with distinct attention flavors.

aiter/ops/triton/attention/mha.py
```diff
@@ -390,6 +392,7 @@ def backward(ctx, do, *args):
                 USE_INT64_STRIDES=_USE_INT64_STRIDES,
                 sink=sink,
                 dsink=dsink,
+                sliding_window=int(ctx.window_size[0]) if int(ctx.window_size[0]) >= 0 else 0,
             )
```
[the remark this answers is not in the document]
brunomazzottiamd: This makes sense. Please implement this suggestion:

> (b) explicitly error out when `_USE_FUSED_BWD_KERNEL` is requested with sliding-window attention.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -1560,12 +1560,247 @@ def test_mha_varlen_with_sink(
         torch_dv,
         atol=bwd_atol,
         rtol=bwd_rtol,
-        msg=lambda msg: f"bwd dv mismatch\n\n{msg}\n",
+        msg=lambda msg: f"bwd dv mismatch (varlen)\n\n{msg}\n",
```
brunomazzottiamd: I think this change can be discarded since the test is already named `test_mha_varlen_with_sink`, i.e. has `varlen` in its name.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -1560,12 +1560,247 @@ def test_mha_varlen_with_sink(
...
     )
     torch.testing.assert_close(
         triton_dsink,
         torch_dsink,
         atol=5e-2,  # higher tolerance due to summation over exp
         rtol=5e-2,  # higher tolerance due to summation over exp
+        msg=lambda msg: f"bwd dsink mismatch (varlen)\n\n{msg}\n",
```
brunomazzottiamd: I think this change can be discarded since the test is already named `test_mha_varlen_with_sink`, i.e. has `varlen` in its name.

op_tests/triton_tests/attention/test_mha.py
brunomazzottiamd: This a must have. I contacted AITER CI team, `test_mha.py` is the 2nd most time-consuming test. We need to split it into multiple files. Use any criteria you judge appropriate.

brunomazzottiamd (changes requested):
Hi @kyle-256. Thanks for your PR.

I think the code is fine, but I have other requests:

* Correctness:
  * [x] The new tests (`test_mha_with_sink_sliding_window` and `test_mha_sliding_window_no_sink`) should pass on `gfx950`, with the TOT Triton compiler from upstream. Our CI runs only on `gfx942`, `gfx950` support on CI is a work in progress. You can publish the `gfx950` test results as a PR comment or add them to the PR description.
  * [x] We're currently facing some accuracy issues with `test_mha.py` on `gfx950`. Run all `test_mha.py` tests on `gfx950`, using `main` branch and your feature branch. Be sure that the failures are the same and the feature branch doesn't introduce any new failures. You can add this to the PR as well.
* Missing benchmarking pieces:
  * [x] Update [`bench_mha.py`](https://github.com/ROCm/aiter/blob/main/op_tests/op_benchmarks/triton/bench_mha.py): Add sliding window support to the core benchmark script.
  * [x] Update [`bench_attn_models.py`](https://github.com/ROCm/aiter/blob/main/op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_attn_models.py) and [`model_shapes.json`](https://github.com/ROCm/aiter/blob/main/op_tests/op_benchmarks/triton/model_benchmarking_tool/model_shapes.json): Add sink and sliding window support, both features are missing as far as I can remember. I think GPT-OSS models use both features, right?
* Performance considerations - add performance experiments to the PR:
  * [x] Make sure sliding window feature doesn't introduce performance regressions in the non-sliding window case, in forward and backward kernels. You can use the benchmark scripts I mentioned before, compare `main` branch with your feature branch.
  * [x] It would be great to have a performance comparison between regular MHA and MHA with sliding window, in forward and backward kernels. As far as I can see, sliding window adds more computations to the mix (more code under `if SLIDING_WINDOW > 0:` statetements).
* Refactoring for CI purposes:
  * [x] Split `test_mha.py` in two, this test file is growing and it's not good for CI granularity.

op_tests/op_benchmarks/triton/bench_mha.py
```diff
@@ -390,7 +437,12 @@ def bench_mha(
             requires_grad=requires_grad,
         )
         sink = (
-            torch.randn((HQ,), device=device, dtype=dtype, requires_grad=requires_grad)
+            torch.randn(
+                (HQ,),
+                device=device,
+                dtype=torch.float32,
```
brunomazzottiamd, line 443: **Question:** Is the sink always `fp32`?

kyle-256: The sink is not required to be fp32 at the API boundary, but for this Triton MHA path it is effectively treated as fp32. In aiter/ops/mha.py we upcast sink_ptr to fp32 before dispatch, and the backward path accumulates dsink in fp32 as well. I used fp32 in the benchmark so we don't benchmark the repeated cast overhead instead of the attention kernel itself.

brunomazzottiamd:
> > Hi @kyle-256. Are you facing any issues related to this PR? Do you see any blockers for merging it? Please let me know.
> 
> No issues from my side, and I don't see any blockers either. Please feel free to merge it into the main branch. Thanks!

Can you please rebase on top of latest `main` and let CI run again? I've tried to re-run the failed jobs, but it didn't work. Once CI is all green we can merge it. Thanks!

PR #2510, opened 2026-03-27
gemm_a8w8 gfx1250 gluon kernel, + wrapper + test + bench

ahmed-bsod:
## Submission Checklist

### DIFF
--- a/aiter/ops/triton/configs/gemm/gluon/gfx1250-GEMM-A8W8-N=5120-K=2880.json
+++ b/aiter/ops/triton/configs/gemm/gluon/gfx1250-GEMM-A8W8-N=5120-K=2880.json
@@ -1,0 +1,25 @@
+{
+    "M_LEQ_64": {
+        "BLOCK_SIZE_M": 64,
+        "BLOCK_SIZE_N": 256,
+        "BLOCK_SIZE_K": 256,
+        "GROUP_SIZE_M": 1,
+        "num_warps": 8,
+        "waves_per_eu": 0,
+        "matrix_instr_nonkdim": 16,
+        "kpack": 1,
+        "NUM_BUFFERS": 3
+    },
+    "any": {
+        "BLOCK_SIZE_M": 256,
+        "BLOCK_SIZE_N": 256,
+        "BLOCK_SIZE_K": 128,
+        "GROUP_SIZE_M": 8,
+        "num_warps": 8,
+        "waves_per_eu": 0,
+        "matrix_instr_nonkdim": 16,
+        "kpack": 1,
+        "NUM_BUFFERS": 2
+    }
+
+}
--- a/aiter/ops/triton/configs/gemm/gluon/gfx1250-GEMM-A8W8-N=7168-K=4096.json
+++ b/aiter/ops/triton/configs/gemm/gluon/gfx1250-GEMM-A8W8-N=7168-K=4096.json
@@ -1,0 +1,24 @@
+{
+    "M_LEQ_64": {
+        "BLOCK_SIZE_M": 32,
+        "BLOCK_SIZE_N": 256,
+        "BLOCK_SIZE_K": 256,
+        "GROUP_SIZE_M": 1,
+        "num_warps": 8,
+        "waves_per_eu": 0,
+        "matrix_instr_nonkdim": 16,
+        "kpack": 1,
+        "NUM_BUFFERS": 4
+    },
+    "any": {
+        "BLOCK_SIZE_M": 256,
+        "BLOCK_SIZE_N": 256,
+        "BLOCK_SIZE_K": 128,
+        "GROUP_SIZE_M": 8,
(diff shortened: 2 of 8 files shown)


zufayu:
@ahmed-bsod This PR has been inactive for over 2 months. Is the gemm_a8w8 gfx1250 gluon kernel still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

zufayu:

**Adds a gfx1250-only Gluon A8W8 GEMM kernel (TDM async-load pipeline rotating 2-4 LDS buffers into a 16x16x128 WMMA) behind a new use_gluon flag on the triton gemm_a8w8 wrapper, plus three tuned config JSONs, a pytest parameter and a bench flag.**

Validation (deterministic): NOT RUN — auto-validation disabled in this environment (REVIEW_AUTO_VALIDATE=0) and no report supplied; the single test target op_tests/triton_tests/gemm/basic/test_gemm_a8w8.py reaches the new kernel only through the gfx1250_gluon parameter, which skips on every arch available here
Perf (advisory): NOT RUN — the triaged target exposes no benchmark entry point; the PR's own --gfx1250-gluon flag in op_tests/op_benchmarks/triton/bench_gemm_a8w8.py was never run or reported, and no gfx1250 GPU exists in this environment

🔴 [verified] The PR is based on a tree that main has refactored away: the diff rejects on aiter/ops/triton/gemm/basic/gemm_a8w8.py, aiter/ops/triton/utils/gemm_config_utils.py and both op_tests files; main's wrapper now dispatches gluon through a backend string parameter instead of this use_gluon bool, gluon kernels moved under _gluon_kernels/, and GEMM configs now resolve from an <arch>/<backend>/gemm/<name>/DEFAULT.json layout — the three new JSONs would land in configs/gemm/gluon/, a directory main no longer reads, so any CI result on this PR describes a tree that has moved (and the PR has been inactive 2+ months under a close-pending maintainer ping). **Author must** rebase onto main and re-integrate the kernel under the backend-parameter dispatch and the new config layout before anything below can be validated.
⚠️ [verified] The use_gluon branch in aiter/ops/triton/gemm/basic/gemm_a8w8.py forwards bias/dtype/y/config but silently drops skip_reduce, and the Gluon wrapper has no split-K: a caller passing skip_reduce=True together with use_gluon=True — the documented fusion enabler for FP8/FP4 quant and RMSNorm — gets a plain (M, N) result instead of the documented (NUM_KSPLIT, M, N) partials, so downstream fusion code mis-indexes the output with no error. **Author must** assert skip_reduce is False (or raise) in the use_gluon branch, or implement split-K in the Gluon kernel.
⚠️ [verified] The kernel is documented FP8-only and the test says it "does not support int8", but nothing executable enforces that: gfx1250_gemm_a8w8 in aiter/ops/triton/gluon/gfx1250_gemm_a8w8.py asserts only x.dtype == w.dtype, the int8 skip in test_gemm_fp8 sits where in_dtype is parametrized only over fp8e4m3/fp8e5m2 so it can never fire, and test_gemm_int8 omits the impl entirely — an int8 caller on gfx1250 reaches the WMMA path, and whether that errors or silently mis-computes is untested. **Author must** assert the fp8 dtype in the wrapper, or test int8 and delete the dead skip.
⚠️ [verified] No timing evidence exists anywhere for a PR whose whole purpose is a faster gfx1250 kernel: the body is an unfilled template (Motivation/Test Plan/Test Result left as comments), no bench output accompanies the --gfx1250-gluon flag, and the per-shape configs in gfx1250-GEMM-A8W8-N=5120-K=2880.json and gfx1250-GEMM-A8W8-N=7168-K=4096.json (NUM_BUFFERS 3/4, distinct M_LEQ_64 buckets, identical "any" blocks) ship with no tuning log — while matrix_instr_nonkdim=16 and kpack=1 are set in all three JSONs although main's own gfx1250 Gluon configs set neither and this kernel lowers to WMMA, not MFMA; the unfilled template plus a silently-dropped parameter also pattern-match generated code, so the dispatch logic deserves extra manual scrutiny. **Author must** attach base-vs-head numbers (generic triton path vs the new kernel on the same shapes), state the tuning basis, and confirm matrix_instr_nonkdim/kpack are not no-ops on RDNA4.
📝 [verified] Nothing production-side can reach the new kernel: the public gemm_a8w8 in aiter/ops/gemm_op_a8w8.py routes RDNA archs to the triton wrapper without use_gluon, so all gfx1250 A8W8 traffic keeps the generic triton path and the kernel activates only for callers importing aiter/ops/triton/gemm/basic/gemm_a8w8.py directly; the only test exercising it (gfx1250_gluon) also skips on every arch CI runs, since the CI guide lists gfx942/gfx950 runners only. **Reviewer should ask** who wires gfx1250 dispatch to this kernel and where the gfx1250 test will actually run.

PR #2521, opened 2026-03-30
[Opt] Fused car+rms for gpt-oss and ensure to use 1-stage kernel

kkHuang-amd:
…ptized performance

## Motivation

This PR will fix two issues

1. Functionality is not work
PR #2329 to fix the issue "fused allreduce rmsnorm shape error" when enable "car+rmsnorm fusion" in gpt-oss model.
But the later PR #2346 remove this change and cause the run-time error in gpt-oss run.

2. Performance is worse than no-fusion version
Current code logic, the fusion kernel for last_dim 2880 will not use 1-stage kernel, it will introduce two kernels `reduce_scatter_cross_device_store` + `local_device_load_rmsnorm` that does not get the fusion benefits than un-fusion version.

## Technical Details

Add last_dim 2880 support for gpt-oss model

## Submission Checklist

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -2506,7 +2506,7 @@ void dispatchFusedAllReduceRMSNorm(hipStream_t stream,
     hipGetDeviceProperties(&dev_prop, dev);
     uint32_t num_cu = dev_prop.multiProcessorCount;
 
-    use_1stage = (use_1stage && (n == 4096 || n == 2048 || n == 1024 || n == 512));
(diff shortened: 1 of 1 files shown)


PerryZhang01:
hey, bro, can u recover the test in PR?#2329

zufayu:
@kkHuang-amd This PR has been inactive for over 2 months. Is the fused car+rms optimization for gpt-oss still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2562, opened 2026-03-31
Fix nondeterministic RNG in test_fused_mxfp4_quant

nidal567:
## Summary:
Tests in test_fused_mxfp4_quant.py were failing in CI, especially when executed as part of shard 3. The failures were not reproducible when running the test line in isolation. Thanks to Bruno for providing the command line.

## Root cause:
The random seed was previously set to be at the top-level part of the module just after imports via torch.manual_seed(). This caused test behaviour to depend on the global RNG state, which is affected by previously executed tests in the same shard (which makes sense why it worked in isolation, but not in the shard). As a result, the test outcomes were order-dependent and non-deterministic.

## Fix:
- Removed torch.manual_seed() from top-level part of module
- Added this deterministic seeding behaviour to the test case that was being impacted by this to ensure order-independent behaviour

## Validation:
- Reproduced failure using CI shard 3 command locally
- Verified the failures occuring in op_tests/triton_tests/quant/test_fused_mxfp4_quant.py::test_fused_rms_quant
- After fix:
	- All tests pass in shard 3 with TRITON_HIP_USE_ASYNC_COPY=0
	- Test_fused_rms_quant also passes with ASYNC_COPY enabled (in command line run with shard 3 and isolation)
	- Tests pass consistently in isolation and repeated runs

## Additional Notes:
- Remaining failures with TRITON_HIP_USE_ASYNC_COPY=1 are affected (MoE + GEMM known issues with ASYNC enabled). This is unrelated to the current task and can be addressed separately

### DIFF
--- a/op_tests/triton_tests/quant/test_fused_mxfp4_quant.py
+++ b/op_tests/triton_tests/quant/test_fused_mxfp4_quant.py
@@ -23,7 +23,6 @@ from aiter.ops.quant import per_1x32_f4_quant_hip
 from aiter.utility.fp4_utils import moe_mxfp4_sort, dynamic_mxfp4_quant
 
 
-
 def rmsnorm(input, weight, eps=1e-6):
     row_norm = input * input
     row_norm = torch.sum(row_norm, dim=-1)
@@ -129,6 +128,8 @@ def test_flatten_quant(B: int, M: int, N: int, dtype):
     if not (arch_info.is_fp4_avail()):
         pytest.skip("MXFP4 not supported on this architecture")
 
+    torch.manual_seed(0)
+
     torch.cuda.empty_cache()  # Helps avoid hangs in large tests
 
     x = torch.randn((B, M, N), dtype=dtype, device="cuda").transpose(0, 1)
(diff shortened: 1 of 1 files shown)


op_tests/triton_tests/quant/test_fused_mxfp4_quant.py
```diff
@@ -169,7 +168,7 @@ def test_fused_rms_quant(
     shuffle: bool,
     scale_shuffle_padding: bool,
 ):
-
+    torch.manual_seed(0)
```
brunomazzottiamd: Can you move `torch.manual_seed(0)` after skip condition?

```python
if not (arch_info.is_fp4_avail()):
    pytest.skip("MXFP4 not supported on this architecture")

torch.manual_seed(0)
```

By this way, we skip "at first glance" on `gfx942` and don't have to setup a RNG that won't be used at all.

brunomazzottiamd (changes requested):
Can you set RNG seed before all test cases? I mean, do it in all `pytest` functions:

* `test_flatten_quant`
* `test_fused_rms_quant`
* `test_fused_reduce_act_mul_mxfp4_group_quant`
* `test_fuse_reduce_rms_quant`
* `test_fused_dynamic_mxfp4_quant_moe_sort`

Let's make everything deterministic, not just the test that's breaking. I think it's a good improvement opportunity.

PR #2566, opened 2026-04-01
rope kernel positions parameter supports int32

daydayup-lh:
## Motivation

background:
the data type of `positions` is `int32` for a customer model( blaze-o1 LUM3 model). 
The data type of the `positions` parameter in the rope kernel is `int64`. 
if calling this rope kernel in the blaze-o1 framework, need to cast int32 to int64. it will takes 7us. 
so we want to optimize this data conversion operation. rope kernel positions parameter supports int32

## Submission Checklist


ruanjm:
LGTM at the code level. For the model level, refer to Linpeng's opinion.

i-chaochen:
> allocate and use int64 in framework, it won't be a model level "must"

AFAIK this PR provides a wider support (int32) on position data type that framework level can support better on more models/usage cases. 

and int32 is the default integer dtype for most users in [JAX](https://github.com/jax-ml/jax/blob/d36f6b3b75b1aef2cb28b07c4498c8ea2ae04fc5/jax/_src/dtypes.py#L285-L290)/[XLA](https://github.com/openxla/xla/blob/8ae49904a055a3a614a652fe2115dd3773b77c44/xla/xla_data.proto#L40) framework level

daydayup-lh:
> but seems this PR has caused compilation error @daydayup-lh please fix it and rebase again...
> 
> ```
> #13 7581.8 /aiter/aiter_meta/csrc/kernels/rope/general_1c_cached_positions_fwd_kernels.cu:66:9: error: no matching function for call to 'dispatch_1c_sbhd_cached_indirect'
> #13 7581.8    66 |         dispatch_1c_sbhd_cached_indirect<OpCachedFwd, RotateStyle, ReuseFreqsFrontPart, NopeFirst, true>(
> #13 7581.8       |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> #13 7581.8 /aiter/aiter_meta/csrc/kernels/rope/rope_common.h:6405:13: note: expanded from macro 'DISPATCH_ROPE_TYPES_PARAMS_WITH_POSITIONS'
> #13 7581.8  6405 |             __VA_ARGS__);                                                         \
> #13 7581.8       |             ^~~~~~~~~~~
> #13 7581.8 /aiter/aiter_meta/csrc/kernels/rope/rope_common.h:6613:25: note: expanded from macro 'DISPATCH_ROPE_TYPES_PARAMS'
> #13 7581.8  6613 |                         __VA_ARGS__;                                              \
> #13 7581.8       |                         ^~~~~~~~~~~
> #13 7581.8 /aiter/aiter_meta/csrc/kernels/rope/rope_common.h:5265:6: note: candidate function template not viable: no known conversion from 'int *' to 'const int64_t *__restrict' (aka 'const long *__restrict') for 5th argument
> #13 7581.8  5265 | void dispatch_1c_sbhd_cached_indirect(scalar_t* __restrict__ p_output,
> #13 7581.8       |      ^
> #13 7581.8  5266 |                                       const scalar_t* __restrict__ p_input,
> #13 7581.8  5267 |                                       const scalar_f_t* __restrict__ p_cos,
> #13 7581.8  5268 |                                       const scalar_f_t* __restrict__ p_sin,
> #13 7581.8  5269 |                                       const int64_t* __restrict__ p_indirect_buffer,
> #13 7581.8       |                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> #13 7581.8 fatal error: too many errors emitted, stopping now [-ferror-limit=]
> #13 7581.8 20 errors generated when compiling for gfx942.
> ```

have fixed the compilation error

ruanjm:
It would be better to add new modules/.so for b32 and b64 separately.

E.g. make `rope_cached_positions_fwd_impl` as a plain Python function without decorator `compile_ops`. Meanwhile, declare 2 new functions like `rope_cached_positions_32_fwd` and `rope_cached_position_64_fwd_impl` both decorated with `compile_ops` but differenct module. `rope_cached_positions_fwd_impl` is in charge of checking the `dtype` of `positions` and choose the most suitable one.

In this way, there would be less `#define` in code which make code become much cleaner. More importantly, we can use both b32 and b64 positions RoPE kernel in the same environment.

It is good to me if u are sure that u will do it in further PR.

csrc/kernels/rope/rope_common.h
valarLip: maybe add "typename pos_t" here, instead of lots of "#if...#else...#endif"

daydayup-lh: ok, I will modify the code by your comments.

csrc/kernels/rope/general_2c_cached_positions_fwd_kernels.cu
```diff
@@ -69,21 +69,22 @@ void rope_cached_positions_2c_fwd_impl(
...
+    DISPATCH_ROPE_TYPES_PARAMS_WITH_POSITIONS(
         input_x.scalar_type(),
         cos.scalar_type(),
+        positions.scalar_type(),
         rotate_style,
         reuse_freqs_front_part,
         nope_first,
         "dispatch_2c_sbhd_cached_indirect<OpCachedFwd, ...>",
-        dispatch_2c_sbhd_cached_indirect<OpCachedFwd, RotateStyle, ReuseFreqsFrontPart, NopeFirst, true>(
+        dispatch_2c_sbhd_cached_indirect<OpCachedFwd, RotateStyle, ReuseFreqsFrontPart, NopeFirst, true, scalar_t_0, scalar_t_1, pos_t>(
```
ruanjm: I think declaring the template parameter here is not required.

daydayup-lh: ok, I will delete it

csrc/kernels/rope/rope_common.h
```diff
@@ -6361,6 +6385,43 @@ void dispatch_1c_2d_cached(scalar_t* __restrict__ p_output,
...
+            TYPE0,                                                                \
+            TYPE1,                                                                \
+            ROTATE_STYLE,                                                         \
+            REUSE_FREQS_FRONT_PART,                                               \
+            NOPE_FIRST,                                                           \
+            NAME,                                                                 \
+            __VA_ARGS__);                                                         \
+        break;                                                                    \
+    }                                                                             \
+    case at::ScalarType::Int: {                                                   \
```
valarLip: i don't want compile both int32/int64 version, please add macro back here.....

daydayup-lh: ok

valarLip (approved):

PR #2587, opened 2026-04-02
Fix precision issue for 32x256 kernel, attempt #3

JohnNikolay84:
## Motivation

Fix precision issue for asm fmoe kernels.

## Technical Details

cl_gemm0_withup wait counters are not  being used correctly.

## Test Plan

DeepSeek model accuracy  is 95%
local 2 stage fmoe unit test show ~50% matching values  and  max diff should be 40% less than before.

## Test Result
DS:

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9591|±  |0.0055|
|     |       |strict-match    |     5|exact_match|↑  |0.9568|±  |0.0056|

## Submission Checklist


JohnNikolay84:
@valarLip  do you have any objections about this commit ? Should I run more tests ?

valarLip (approved):

PR #2594, opened 2026-04-02
Enabled rope Benchmarking CSV Output

etemadiamd:
## Submission Checklist


zufayu:
Rebased this onto current `main` and re-ran CI. It had been sitting since April, and the old CI logs had already aged past GitHub's retention window (HTTP 410), so the previous red marks were undiagnosable.

### Rebase

`2e6772f4` → `faa023db`. The old head was a `Merge branch 'main' into enable_rope_save_csv` commit; the rebase collapsed the branch to a single commit on top of `main`. No conflicts.

### The change still applies

The branch was ~319 commits behind, so I checked whether it had rotted:

- `get_caller_name_no_ext` still exists in `op_tests/op_benchmarks/triton/utils/benchmark_utils.py`
- `bench_rope.py` on `main` still has no CSV output, so this hasn't been superseded
- The `-o` / `--output-csv` flag is self-contained (`dest="o"`), so `args.o` resolves
- It follows existing convention — `get_caller_name_no_ext` is already used by ~13 other `bench_*.py` scripts

### CI: four failures, one root cause

Current state is `fail=4  pass=3  skipping=25`, but the four are one problem:

- **`Check Code Style with Ruff`** — the actual failure
- **`check-signal` ×3** — cascading. Each one logs:
  ```
  Pre-checks did not pass (conclusion: failure), skipping workflow.
  FAILED: Check Code Style with Ruff (failure)
  ```

So **Aiter Test / Triton Test / OPUS Test never ran** — they are gated behind pre-checks. There is currently no functional signal on this PR at all.

Ruff reports two findings, both in `bench_rope.py`:

| Rule | Line | Finding |
|---|---|---|
| `I001` | 1–36 | Import block is un-sorted or un-formatted |
| `FURB105` | 548 | `print("")` should be `print()` |

Worth flagging: **both already exist on `main`.** `print("")` is at line 527 there, and the `benchmark_utils` import block is already unsorted (`get_model_configs` before `get_available_models`). reviewdog runs with `filter-mode=diff_context`, so it reports context lines around the diff, not only added lines — this PR didn't introduce either finding, it just pulled them into scope. They still fail the job (`-fail-on-error=true`), so they need fixing here regardless.

### What's left

1. Fix the two Ruff findings — a few lines, and it unblocks the entire test suite
2. Let the real tests run for the first time on this change
3. An approving review — there are currently none, which is what `BLOCKED` is reporting

@etemadiamd — is this still wanted? If yes, the lint fix is small and the rest should follow. If it's been abandoned, closing it would be cleaner than leaving it open.

zufayu:
## [ROCm/aiter] PR #2594 — Enabled rope Benchmarking CSV Output

**Adds an opt-in `-o/--output-csv` flag to the RoPE benchmark that writes the run's config, the analytically estimated TFLOPs and memory GB, the repeat count and the model name to `bench_rope.csv`.**

⚠️ [verified] The diff no longer applies to merge target a84bd368 (patch fails at bench_rope.py:1): main has since re-sorted the import block, dropped the noqa comments and fixed `print("")` → `print()`, so the PR's head carries 26 ruff errors that are all already fixed on main — while the PR's own additions, applied to a copy of main's file, pass ruff 0.16.0 (the pinned version) clean. CI on the current head describes a tree that has moved, and the PR cannot merge as-is. **Author must** rebase onto current main, placing `import csv` in the re-sorted stdlib block (argparse, csv, blank line, torch).

📝 [verified] bench_rope.py:540-546 writes `bench_rope.csv` with `open(..., "w")` while `x_vals_list` (bench_rope.py:144-164) is a literal one-element list, so one invocation produces exactly one row and the next invocation silently overwrites the previous config's result; the output path is never printed (bench_flash_kda.py prints "Saved to ..."). For collecting a config sweep — the natural use of CSV output from a one-config-per-run script — every run but the last is lost. **Reviewer should ask** for append-with-header-if-absent (as bench_sage.py:2158-2161 does) or at least a printed output path.

📝 [verified] The PR description is an unfilled template: Motivation, Technical Details, Test Plan and Test Result are template comments and the submission checklist is unchecked, so nothing shows the `-o` path was ever executed (the CSV logic itself verifies standalone: header, units and values match the printed labels). **Author must** fill in the test plan with the command run and a sample of the output CSV.

Issue #2157, opened 2026-03-03
Support TheRock ROCm “Python-managed” install paths (non-/opt/rocm)


akao-amd:
After installing the latest ROCm 7.12 (TheRock / ROCm Core SDK stream) following [the official guide](https://github.com/ROCm/TheRock/blob/main/RELEASES.md#rocm-for-gfx950-dcgpu)  , and initializing with: ` rocm-sdk init`, the resulting SDK layout does not follow the traditional /opt/rocm or /opt/rocm-x.y.z convention on a vanilla ubuntu:22.04 base image.

Instead, the SDK is installed / expanded under a Python site-packages managed location, e.g.:

/usr/local/lib/python3.10/site-packages/_rocm_sdk_core
(and other _rocm_sdk_* directories)

This path change breaks tools and scripts that assume ROCm lives under /opt/rocm* (e.g., hardcoded discovery, PATH/CMAKE_PREFIX_PATH logic, or “rocm root” probing).

Request / ask
As TheRock-derived ROCm becomes a priority target, please consider updating the tooling to discover and support ROCm installations rooted in the _rocm_sdk_* site-packages paths (in addition to /opt/rocm*).

Issue #2470, opened 2026-03-25
[Feature]: Update aiter/jit/utils/cpp_extension.py from pytorch for Windows support


astrelsky:
### Suggestion Description

The code in `aiter/jit/utils/cpp_extension.py` is from https://github.com/pytorch/pytorch/blob/main/torch/utils/cpp_extension.py

https://github.com/ROCm/aiter/blob/28b2942591ff9a53e553abb2908b9fd1dadfb094/aiter/jit/utils/cpp_extension.py#L4-L6

The version used by `aiter` is way out of date, doesn't support building PyTorch extensions using rocm and Windows, and doesn't handle rocm from `TheRock` wheels. Using an updated version should largely resolve enabling jit support for Windows.

### Operating System

Win11

### ROCm Component

aiter

Related to #2072, #2428, #1943

CK also supports Windows, so there is absolutely no reason for aiter to not fully support Windows except for FlyDSL.

PR #2605, opened 2026-04-03
fix: replace hardcoded /opt/rocm paths with ROCM_HOME env var

zufayu:
## Summary

- Replace hardcoded `/opt/rocm` paths in `csrc/cpp_itfs/utils.py` with `ROCM_HOME`/`ROCM_PATH` env var lookup (fallback to `/opt/rocm`)
- Make `csrc/kernels/solver/Makefile` use `?=` so `ROCM_HOME` can override

This is the minimal fix to unblock TheRock (ROCm Core SDK) users who have ROCm installed under non-standard paths (e.g. Python site-packages). Users just need:
```bash
export ROCM_HOME=$(rocm-sdk path --root)
```

**Note:** `aiter/jit/utils/cpp_extension.py` already supports `ROCM_HOME`/`ROCM_PATH` — no changes needed there.

**Changes (2 files):**
| File | Change |
|------|--------|
| `csrc/cpp_itfs/utils.py` | `amdgpu-arch` and `hipconfig` paths use `_ROCM_HOME` variable |
| `csrc/kernels/solver/Makefile` | `ROCBLAS_DIR`/`ROCSOLVER_DIR` use `?=` with `ROCM_HOME` |

Ref #2157

## Test plan
- [x] Verified `DEFAULT_GPU_ARCH` and `get_hip_version()` work correctly on traditional `/opt/rocm` install
- [ ] TheRock environment test (requires TheRock setup)


csrc/kernels/solver/Makefile
astrelsky, line 1: Make is not available on Windows.

csrc/cpp_itfs/utils.py
```diff
@@ -39,9 +39,13 @@ def get_git_commit_id_short():
 AITER_CORE_DIR = os.path.abspath(f"{this_dir}/../../")
 if os.path.exists(os.path.join(AITER_CORE_DIR, "aiter_meta")):
     AITER_CORE_DIR = os.path.join(AITER_CORE_DIR, "aiter_meta")
+_ROCM_HOME = os.environ.get("ROCM_HOME") or os.environ.get("ROCM_PATH") or "/opt/rocm"
```
astrelsky: Environment variables `ROCM_HOME` and `ROCM_PATH` can **not** be required. Please use the `rocm-sdk` command line tool if it is available.

There really should be functions available in the `rocm` python package. Having to use a subprocess is stupid in my opinion.

astrelsky (changes requested):

csrc/cpp_itfs/utils.py
zufayu: Addressed all review feedback in 77b2b194e:

**@astrelsky:**
1. **Env vars are NOT required** — they're only an optional user override (escape hatch). The full discovery chain:
   1. `ROCM_HOME`/`ROCM_PATH` env var (optional override)
   2. `rocm_sdk._devel.get_devel_root()` Python import — **no subprocess needed**
   3. `hipconfig --rocmpath` (tool-based)
   4. `rocm-sdk path --root` CLI (fallback when Python import unavailable)
   5. `/opt/rocm` (last resort)

2. **Python package import added** — Step 2 now tries `from rocm_sdk._devel import get_devel_root` directly, addressing the "subprocess is stupid" concern. The CLI subprocess is kept as step 4 fallback only.

**Copilot fixes:**
3. `shell=True` → `shell=False` with argv lists for `DEFAULT_GPU_ARCH` and `get_hip_version()`
4. Makefile now honors `ROCM_PATH` via `$(if $(ROCM_PATH),$(ROCM_PATH),/opt/rocm)`

**Re: Makefile on Windows** — This Makefile (`csrc/kernels/solver/Makefile`) is an existing HIP benchmark (uses `hipcc`, `--offload-arch=gfx942`, links `rocsolver`/`rocblas`). It was already Make-only before this PR — we only changed the `ROCM_HOME` line to be overridable instead of hardcoded. The Windows concern is valid but out of scope for this fix.

csrc/kernels/solver/Makefile
zufayu: Acknowledged — this Makefile is an existing HIP benchmark that already requires `hipcc`, `rocsolver`, and `rocblas`. It was Make-only before this PR. We only made the `ROCM_HOME` line overridable instead of hardcoded. The Windows portability concern is valid but orthogonal to this fix.

astrelsky: The `hipcc`, `rocsolver` and `rocblas` requirements are irrelevant as they support Windows. The only 3 things blocking aiter from fully supporting Windows is `FlyDSL` (they seem to be ignoring the issue like aiter is), the hardcoded linux `/opt/rocm` that this pr fixes, and the outdated `aiter/jit/utils/cpp_extension.py` reported in #2470. Everything else fully supports Windows (yes, even CK supports Windows).

The Makefile is simple enough that it really isn't necessary, but understandable if you want to leave it here to avoid scope creep.

csrc/cpp_itfs/utils.py
astrelsky: Ok, thank you. I didn't know that function existed.

I won't approve it, since I don't work for AMD and don't want to cause confusion.

PR #2491, opened 2026-03-26
[TRITON] Fix unit tests on `gfx950` - part 2

brunomazzottiamd:
## Motivation

Triton test suite isn't passing on `gfx950`. This PR fixes `test_gemm_afp4wfp4.py` and `test_gemm_a8w8.py`, slightly improving the situation.

### Fix `test_gemm_afp4wfp4.py`

Triton commit [`de2ba3946b`](https://github.com/triton-lang/triton/commit/de2ba3946b) changed `AMDMFMALayout.instr_shape` from a 2-element `[M, N]` to a 3-element `[M, N, K]` list. Extend the previously 2-element `[32, 32]` to `[32, 32, 64]`. K=64 is the K dimension of the `mfma_scale_f32_32x32x64_f8f6f4` hardware instruction used for FP4 on `gfx950`.

Restrict AFP4/WFP4 AOT tests to Triton 3.5. Avoid using prebuilt AOT kernels on newer Triton versions where the metadata format is incompatible.

#### Fix `_gemm_a8w8_kernel`

Same `instr_shape` API break (Triton [`de2ba3946b`](https://github.com/triton-lang/triton/commit/de2ba3946b)). The kernel uses `mfma_scaled` for FP8 and plain `mfma` for INT8, which target different hardware instructions with different K dimensions:

- FP8 `mfma_scale_f32_16x16x128_f8f6f4` (K=128, K_WIDTH=32)
- INT8 `mfma_i32_16x16x64_i8` (K=64, K_WIDTH=16)

`SwizzledSharedLayout.vec` is updated to match K_WIDTH per data type specialization.

#### Fix `_gemm_a8w8_preshuffled_kernel`

The `linear_nk` layout and its `reshape - permute - reshape - trans` unshuffle sequence were designed for K=32 / K_WIDTH=16, so applying K=128 breaks the layout conversion. Since `mfma_scaled` was already invoked with `a_scale=None` and `b_scale=None` (per-tensor scale applied to the accumulator separately), replace it with plain `mfma`, targeting the unscaled `mfma_f32_16x16x32_fp8_fp8` (K=32) that the preshuffled layout was built for.

#### Fix `test_gemm_a8w8.py`

Relax absolute tolerance from 0.02 to 0.03 to accommodate the preshuffled FP8 path (unscaled dot + software accumulator scale).

### Fix `ff_a16w16_fused_ungated.py`
    
[...]

Issue #2312, opened 2026-03-17
CI: Run Triton tests on MI355


gyohuangxin:
## Summary

Currently Triton tests are not running on MI355 runners. We should add MI355 support for the Triton test jobs in the CI pipeline.

## Tasks

- [ ] Add MI355 runner to Triton test matrix
- [ ] Verify Triton tests pass on MI355

gyohuangxin:
# `gfx950` Failures

## Shard 0

All passed.

## Shard 1

`op_tests/triton_tests/attention/test_mha.py::test_mha_backward` aborts unexpectedly. I was able to reproduce it locally on Monday. I'm not sure what's going on. It needs to be investigated.

## Shard 2

All passed.

## Shard 3

**Overall result:** 3,754 failed | 17,066 passed | 3,680 skipped — runtime 31:10 min

**Test files in shard:**

- `op_tests/triton_tests/test_causal_conv1d.py`
- `op_tests/triton_tests/attention/test_flash_attn_kvcache.py`
- `op_tests/triton_tests/normalization/test_rmsnorm.py`
- `op_tests/triton_tests/gemm/basic/test_gemm_a8w8_per_token_scale.py`
- `op_tests/triton_tests/attention/test_fav3_sage.py`
- `op_tests/triton_tests/quant/test_quant_mxfp4.py`
- `op_tests/triton_tests/moe/test_moe_mx.py`
- `op_tests/triton_tests/gemm/fused/test_fused_gemm_afp4wfp4_mul_add.py`
- `op_tests/triton_tests/gemm/basic/test_gemm_afp4wfp4.py`
- `op_tests/triton_tests/attention/test_hstu_attn.py`

### Failing Tests

#### 1. `test_causal_conv1d.py` — 2 failures

**Test:** `test_causal_conv1d_update_with_batch_gather`
**Params:** `batch_size=512, dim=4096, width=3, seqlen=4, itype=torch.bfloat16` (both `has_bias=True` and `has_bias=False`)

**Cause:** Numerical correctness failure — `torch.allclose()` returns False with `rtol=0.01, atol=0.05`. The output values are close but marginally out of tolerance, likely due to bfloat16 precision accumulating over a very large batch (512) × dim (4096) product. This is only triggered at the largest combination of batch and dim, suggesting a borderline precision issue at scale.

#### 2. `test_gemm_afp4wfp4.py` — 3,752 failures (nearly all failures in the run)

[...]

PR #2612, opened 2026-04-03
[TRITON] Reduce MHA UTs

vgokhale:
Reduced MHA UTs.


op_tests/triton_tests/attention/test_chunked_pa_prefill.py
```diff
@@ -9,12 +9,12 @@
 from aiter.ops.triton.utils.types import str_to_torch_dtype
 
 NUM_HEADS = [64]
-NUM_QUERIES_PER_KV = [1, 8, 64]
-HEAD_SIZES = [128, 96, 24]
+NUM_QUERIES_PER_KV = [1, 8]
+HEAD_SIZES = [128]
 DTYPES = [torch.float16]
```
brunomazzottiamd, line 14: **Sincere question:** Should we use `bf16` here?

op_tests/triton_tests/attention/test_unified_attention.py
```diff
@@ -9,9 +9,9 @@
 from aiter.ops.triton.attention.unified_attention import unified_attention
 from aiter.ops.triton.utils.types import e4m3_dtype
 
-NUM_HEADS = [(4, 4), (8, 2), (16, 2)]
-HEAD_SIZES = [128, 256]
-BLOCK_SIZES = [16, 64, 48]
+NUM_HEADS = [(4, 4), (16, 2)]
+HEAD_SIZES = [64, 128]
```
k50112113, line 13: GPT-OSS only has `HEAD_SIZE = 64`, and we can maybe only set `NUM_HEADS = [(8, 1)]` for TP8 case that should also cover the correctness of TP1

k50112113 (commented):
GPT-OSS suggested changes

op_tests/triton_tests/attention/test_chunked_pa_prefill.py
```diff
@@ -9,12 +9,12 @@
...
-CUDA_DEVICES = [f"cuda:{i}" for i in range(1)]
-SLIDING_WINDOW = [0, 16, 64, 128, 256, 512, 2048]
-KV_CACHE_DTYPES = ["auto", "fp8e4m3", "fp8e5m2"]
+CUDA_DEVICES = [0, 3]
```
brunomazzottiamd: This caught my attention. I have some considerations...

1. What's the value of testing in more than one GPU? Can't we simply remove this parameterization and test on the default GPU (i.e., `device = "cuda"`)? It'd be a matter of doing this on `test_contexted_kv_attention` and `test_contexted_kv_attention_alibi`.
2. This test will fail on a host with just one GPU. In the early `gfx950` days, we had access to hosts with only one GPU.
3. This is bad for GPU sharing in the same host since it doesn't play well with `HIP_VISIBLE_DEVICES` environment variable. For instance, if I do `export HIP_VISIBLE_DEVICES=7` to reserve a GPU for me and let the other GPUs to the team then the test will break because there's only `cuda:0` from PyTorch's perspective.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -1007,8 +1106,8 @@ def test_mha_backward_with_pe(
 
     # Backward assertions
     # When dropout is active, some cases fail due to less than 1% mismatched elements.
```
brunomazzottiamd: I think this comment

```python
# When dropout is active, some cases fail due to less than 1% mismatched elements.
```

is deprecated given these changes:

```diff
-    bwd_atol = 1e-1 if HAS_DROPOUT else 1.5e-2
-    bwd_rtol = 1e-1 if HAS_DROPOUT else 1.5e-2
+    bwd_atol = 1e-1
+    bwd_rtol = 1e-1
```

I'd remove the comment.

brunomazzottiamd (approved):
LGTM!

I have some considerations in `op_tests/triton_tests/attention/test_chunked_pa_prefill.py` test file. Address them at your own judgement. It's not a blocker IMHO because it doesn't affect the main goal of this PR, i.e. reduce test cases.

brunomazzottiamd (changes requested):
@vgokhale, the following tests are failing on MI325X CI runners:

**Shard 2:**
* `op_tests/triton_tests/attention/test_unified_attention_sparse_mla.py::test_triton_unified_attn`
```
  def test_triton_unified_attn(
E       fixture 'rope_dim' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capteesys, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, subtests, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

**Shard 5:**
* `op_tests/triton_tests/attention/test_pa_prefill.py::test_contexted_kv_attention`
* `op_tests/triton_tests/attention/test_pa_prefill.py::test_contexted_kv_attention_alibi`
* `op_tests/triton_tests/gemm/basic/test_gemm_a16w8_blockscale.py::test_gemm`
```
E           torch.AcceleratorError: HIP error: invalid device ordinal
E           GPU device may be out of range, do you have enough GPUs?
E           HIP kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
E           For debugging consider passing AMD_SERIALIZE_KERNEL=3
E           Compile with `TORCH_USE_HIP_DSA` to enable device-side assertions.
```

**Shard 6:**
* `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink`
* `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink_sliding_window`
One or two mismatched elements in backward dk (`Mismatched elements: 1 / N` or `Mismatched elements: 2 / N`).

**Shard 7:**
* `op_tests/triton_tests/attention/test_fav3_sage.py`
```
   @pytest.mark.parametrize("BATCH", [1, 4, 57, 128])
   @pytest.mark.parametrize(
       "SEQLEN_Q, SEQLEN_K",
       [(1, 1), (4, 4), (128, 128), (2, 1), (1, 2), (32, 16), (64, 128)],
   )
   @pytest.mark.parametrize(
       "NUM_Q_HEADS, NUM_K_HEADS", [(1, 1), (16, 16), (2, 1), (48, 8)]
   )
   @pytest.mark.parametrize("layout", ["bhsd", "bshd"])
   def test_sage(
 E       fixture 'HEAD_SZ' not found
 >       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capteesys, doctest_namespace, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, subtests, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
 >       use 'pytest --fixtures [testpath]' for help on them.
```

Can you please check them? As far as I can see, all files were touched by this PR.

Thanks!

op_tests/triton_tests/attention/test_unified_attention.py
vgokhale: I think we want to be generic here. 256 is not really common but 64 and 128 are. 128 much more so.

op_tests/triton_tests/attention/test_chunked_pa_prefill.py
vgokhale: We should. I have a subsequent PR for that. It failed UTs so i didn't want to fix in this PR.

op_tests/triton_tests/attention/test_chunked_pa_prefill.py
vgokhale: Good point. I just assumed the author had some reason to try different GPUs and didn't want to spend too much time understanding what exactly this was testing. I can remove...

op_tests/triton_tests/attention/test_chunked_pa_prefill.py
brunomazzottiamd: That's fair. Let's consider this discussion resolved.

op_tests/triton_tests/attention/test_mha.py
vgokhale: Done.

op_tests/triton_tests/attention/test_chunked_pa_prefill.py
brunomazzottiamd: It was already testing on a single GPU... ` [f"cuda:{i}" for i in range(1)]` evaluates to `["cuda:0"]`. Adding GPU number 3 to the list broke CI, check https://github.com/ROCm/aiter/pull/2612#pullrequestreview-4062823174. I think we can change this to just `"cuda"` or even better, remove this parameterization.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -58,7 +58,7 @@ def _tol(ref_val, pt_val, is_forward=False):
         baseline = (pt_val - ref_val).abs().max().item()
         if is_fp8:
             mult = 4
-            atol_floor = 3e-1 if is_forward else 1.0
```
lucas-santos-amd, line 63: Hi, is this change intentional? Going from 0.3 to 1.0 seems like a lot... 

On the other hand, this would fix the gfx950 MHA CI failures that are raised from the forward output test inside test_mha.py::test_mha_backward (related issue: https://github.com/ROCm/aiter/issues/2312).

micmelesse (commented):
@vgokhale  What is driving this? Is this CI taking to long or was it an external request?

op_tests/triton_tests/attention/test_mha.py
vgokhale: I changed it to 5e-1. It does appear to be a tolerance issue and not a real test failure. WDYT?

lucas-santos-amd: It can very well be a tolerance issue given that both torch and triton outputs differ between gfx942 and gfx950, so it shouldn't be a triton kernel problem. By using 5e-1 there would still be failures, we have absolute differences of up to 0.75 on some [test cases](https://github.com/ROCm/aiter/actions/runs/23813463912/job/69412662136?pr=2491). I said that 1.0 seemed like a lot, but it might be necessary...

vgokhale: I don't see any failures in the UTs I ran as part of this. If after this PR we still see failures we can lower further.

brunomazzottiamd (dismissed):
LGTM!

Now the only test that's failing is `op_tests/triton_tests/gemm/feed_forward/test_ff_a16w16_fused.py`. I'm investigating it, it seems to be compiler related.

brunomazzottiamd:
> LGTM!
> 
> Now the only test that's failing is `op_tests/triton_tests/gemm/feed_forward/test_ff_a16w16_fused.py`. I'm investigating it, it seems to be compiler related.

The failing test is fixed by https://github.com/ROCm/aiter/pull/2491/changes/fc154d33b5f50ec42e2f94038b0d82dd04420e65.

op_tests/triton_tests/attention/test_mha.py
lucas-santos-amd: The current AITER CI runs only on gfx942, the failures I mentioned happen on gfx950. The goal of https://github.com/ROCm/aiter/issues/2312 is to fix test failures on gfx950 to allow its integration into the CI.

vgokhale: Yeah I meant I don't see any failures on my gfx950 machine with this reduced set of UTs.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -205,29 +191,88 @@ def test_mha(
         torch.testing.assert_close(triton_out, torch_out, atol=1e-2, rtol=1e-2)
 
 
-# LLaMA 3 405B config
-@pytest.mark.parametrize("BATCH", [1])
+@pytest.mark.parametrize("BATCH", [1, 30, 64])
 @pytest.mark.parametrize(
     "SEQLEN_Q, SEQLEN_K",
-    [(1, 1)],
+    [(1, 1), (128, 128), (1, 2), (32, 16), (64, 128)],
```
micmelesse: Make the sequence lengths bigger to be more like production usage here would be good. But if you are just moving stuff around it is fine for now. We can do it later

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -205,29 +191,88 @@ def test_mha(
...
+
+
+# LLaMA 3 405B config
+def test_mha_int64_strides(
     dtype=torch.float16,
     device="cuda",
     test_backward=True,
 ):
+    BATCH = 1
+    SEQLEN_Q, SEQLEN_K = 1, 1
```
micmelesse, line 279: A seqlens of 1 is maybe good for catching kernel masking bugs but it doesnot really represent a production 405B workload.

op_tests/triton_tests/attention/test_mha.py
```diff
@@ -474,6 +505,78 @@ def test_mha_varlen(
         )
 
 
+@pytest.mark.parametrize("BATCH", [1, 4, 57, 128])
+@pytest.mark.parametrize(
+    "SEQLEN_Q, SEQLEN_K",
+    [(1, 1), (4, 4), (128, 128), (2, 1), (1, 2), (32, 16), (64, 128)],
```
micmelesse: Ditto

micmelesse (commented):
I think having the pr test cases reflect production workloads might help catch common issues better. It can probably be follow up work. I have noted some of the spots where I see an opportunity. Maybe take configs from Llama3 (8B/70B/405B), Mixtral (7B/22B), and DeepSeek-V.

op_tests/triton_tests/attention/test_mha.py
vgokhale: I added a 8192, 8192

op_tests/triton_tests/attention/test_mha.py
vgokhale: I changed it to 1024, 1024

brunomazzottiamd:
We're experiencing the following test failures:

**Triton Tests (MI325) / Shard 4**

Minor correctness issue in `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink[True-0.0-128-8-1-1024-1024-1]`:

```text
AssertionError: bwd dk mismatch

Tensor-likes are not close!

Mismatched elements: 1 / 131072 (0.0%)
Greatest absolute difference: 0.0224609375 at index (0, 0, 0, 126) (up to 0.015 allowed)
Greatest relative difference: 0.1318359375 at index (0, 0, 0, 126) (up to 0.015 allowed)
```

**Triton Tests (MI325) / Shard 0**

~Segfault~ Abort `in op_tests/triton_tests/attention/test_mha.py::test_mha[True-False-128-48-8-8192-8192-30]` (may be in other test case, but it's happening in `test_mha.py`).

Here's the top of the stack trace:

```text
Fatal Python error: Aborted

Thread 0x00007f0c45d3c080 (most recent call first):
  File "/opt/venv/lib/python3.12/site-packages/torch/functional.py", line 373 in einsum
  File "/workspace/aiter/test_mha_common.py", line 434 in attention_ref
  File "/workspace/op_tests/triton_tests/attention/test_mha.py", line 178 in _test_mha_impl
  File "/workspace/op_tests/triton_tests/attention/test_mha.py", line 214 in test_mha
```

As far as I can understand, the ~segfault~ abort happens in the reference PyTorch implementation, i.e. `attention_ref` function.

gyohuangxin:
@lucas-santos-amd Can you check the failures in shard 0 and shard 5

brunomazzottiamd:
> @lucas-santos-amd Can you check the failures in shard 0 and shard 5

**Shard 0**

It's an abort on `op_tests/triton_tests/attention/test_mha.py`, i.e. `Fatal Python error: Aborted`.

**Shard 5**

It's one element mismatch in `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink[True-0.0-128-8-1-1024-1024-1]`:

```text
AssertionError: bwd dk mismatch

Tensor-likes are not close!

Mismatched elements: 1 / 131072 (0.0%)
Greatest absolute difference: 0.02197265625 at index (0, 0, 0, 126) (up to 0.015 allowed)
Greatest relative difference: 0.23046875 at index (0, 0, 0, 126) (up to 0.015 allowed)
```

lucas-santos-amd:
> > @lucas-santos-amd Can you check the failures in shard 0 and shard 5
> 
> **Shard 0**
> 
> It's an abort on `op_tests/triton_tests/attention/test_mha.py`, i.e. `Fatal Python error: Aborted`.
> 
> **Shard 5**
> 
> It's one element mismatch in `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink[True-0.0-128-8-1-1024-1024-1]`:
> 
> ```
> AssertionError: bwd dk mismatch
> 
> Tensor-likes are not close!
> 
> Mismatched elements: 1 / 131072 (0.0%)
> Greatest absolute difference: 0.02197265625 at index (0, 0, 0, 126) (up to 0.015 allowed)
> Greatest relative difference: 0.23046875 at index (0, 0, 0, 126) (up to 0.015 allowed)
> ```

For Shard 0, before the abort happened, there is also 82 failures, all of them on the new SEQLEN_Q, SEQLEN_K = (8192, 8192) test case added.

gyohuangxin:
> > > @lucas-santos-amd Can you check the failures in shard 0 and shard 5
> > 
> > 
> > **Shard 0**
> > It's an abort on `op_tests/triton_tests/attention/test_mha.py`, i.e. `Fatal Python error: Aborted`.
> > **Shard 5**
> > It's one element mismatch in `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink[True-0.0-128-8-1-1024-1024-1]`:
> > ```
> > AssertionError: bwd dk mismatch
> > 
> > Tensor-likes are not close!
> > 
> > Mismatched elements: 1 / 131072 (0.0%)
> > Greatest absolute difference: 0.02197265625 at index (0, 0, 0, 126) (up to 0.015 allowed)
> > Greatest relative difference: 0.23046875 at index (0, 0, 0, 126) (up to 0.015 allowed)
> > ```
> 
> For Shard 0, before the abort happened, there is also 82 failures, all of them on the new SEQLEN_Q, SEQLEN_K = (8192, 8192) test case added.

Do you have any ideas to fix those failures?

gyohuangxin:
@lucas-santos-amd the failures are still failing. 
cc @brunomazzottiamd

brunomazzottiamd:
> Do you have any ideas to fix those failures?

@gyohuangxin, I have an hypothesis that's being investigated by @lucas-santos-amd right now:

* This PR has added new test cases with bigger sequence lengths like 4K and 8K. This is stressing the memory subsystem.
* When testing locally, we've been experiencing aborts with seq. len. of 8K and OOM exceptions with seq. len. of 4K. Probably the aborts are coming from lower layers in the stack - unrecoverable allocator states, due to fragmentation or something else that's related to the stress on the memory subsystem.
* Our [reference implementation](https://github.com/ROCm/aiter/blob/main/aiter/test_mha_common.py#L504) isn't Flash Attention, so it has a quadratic cost on memory when materializing the attention tensors. When the reference implementation is commented out, we can run forward using seq. len. of 4K on MI355 without experiencing OOM exceptions.
* The memory requirements of backward are even heavier: in addition to forward requirements, there are the $\partial Q$, $\partial K$, $\partial V$ gradients for both Triton and reference implementation. Reference backward is computed with PyTorch autograd, this can impose even more memory related stress.

I think the solution is to decrease sequence length for forward tests and decrease it even further for backwards tests. We should also take into account that distinct GPUs have distinct memory capacities (MI35X 288 GB, MI325 256 GB, MI300 192 GB), the lowest common denominator should be used IMHO. It isn't ideal for testing real production workloads, but it's the best we can do on UTs without a complete rework.

FYI: @vgokhale + @azaidy

brunomazzottiamd:
> @lucas-santos-amd the failures are still failing. cc @brunomazzottiamd

Yes, I known... it's sad. We're still working on it!

lucas-santos-amd:
> @lucas-santos-amd the failures are still failing. cc @brunomazzottiamd

Just pushed a commit that should hopefully fix the MHA failures.

gyohuangxin:
> > Do you have any ideas to fix those failures?
> 
> @gyohuangxin, I have an hypothesis that's being investigated by @lucas-santos-amd right now:
> 
> * This PR has added new test cases with bigger sequence lengths like 4K and 8K. This is stressing the memory subsystem.
> * When testing locally, we've been experiencing aborts with seq. len. of 8K and OOM exceptions with seq. len. of 4K. Probably the aborts are coming from lower layers in the stack - unrecoverable allocator states, due to fragmentation or something else that's related to the stress on the memory subsystem.
> * Our [reference implementation](https://github.com/ROCm/aiter/blob/main/aiter/test_mha_common.py#L504) isn't Flash Attention, so it has a quadratic cost on memory when materializing the attention tensors. When the reference implementation is commented out, we can run forward using seq. len. of 4K on MI355 without experiencing OOM exceptions.
> * The memory requirements of backward are even heavier: in addition to forward requirements, there are the 
>     ∂
>     Q
>   , 
>     ∂
>     K
>   , 
>     ∂
>     V
>    gradients for both Triton and reference implementation. Reference backward is computed with PyTorch autograd, this can impose even more memory related stress.
> 
> I think the solution is to decrease sequence length for forward tests and decrease it even further for backwards tests. We should also take into account that distinct GPUs have distinct memory capacities (MI35X 288 GB, MI325 256 GB, MI300 192 GB), the lowest common denominator should be used IMHO. It isn't ideal for testing real production workloads, but it's the best we can do on UTs without a complete rework.
> 
> FYI: @vgokhale + @azaidy

Thanks for looking into this — makes sense to me, and I’m on board with dialing down seqlen (especially for backward) and using the smallest GPU as the baseline.

Quick question though: how were you able to run this locally with the longer seqlens (4K / 8K)? I’m wondering if your setup is different.

brunomazzottiamd:
> Quick question though: how were you able to run this locally with the longer seqlens (4K / 8K)? I’m wondering if your setup is different.

The Triton kernels run fine with 4K / 8K sequence length in isolation. The issue is related to the reference implementation backed by PyTorch - it isn't Flash Attention, thus it's quadratic on memory. If we comment out the reference implementation from the test then it pass. In the test we need to keep the output tensors from Triton kernel and reference implementation alive, so we can compare both tensors - this imposes further pressure on memory.

brunomazzottiamd:
As far as I can see, `test_mha.py` is now passing on both GPU architectures (`gfx942`, `gfx950`)! 🥳🥳🥳

However, we still have just one minor accuracy issue in `op_tests/triton_tests/attention/test_mha_with_sink.py::test_mha_with_sink[True-0.0-128-8-1-1024-1024-1]`:

```
AssertionError: bwd dk mismatch

Tensor-likes are not close!

Mismatched elements: 1 / 131072 (0.0%)
Greatest absolute difference: 0.0224609375 at index (0, 0, 0, 126) (up to 0.015 allowed)
Greatest relative difference: 0.1318359375 at index (0, 0, 0, 126) (up to 0.015 allowed)
```

I'm in favor of relaxing backward error tolerance for all gradients if sequence length is greater of equal to 1024. We already have this relaxing in place for $\partial V$:

```python
    bwd_atol = 1.5e-2
    bwd_rtol = 1.5e-2
    torch.testing.assert_close(
        triton_dq,
        torch_dq,
        atol=bwd_atol,
        rtol=bwd_rtol,
        msg=lambda msg: f"bwd dq mismatch\n\n{msg}\n",
    )
    torch.testing.assert_close(
        triton_dk,
        torch_dk,
        atol=bwd_atol,
        rtol=bwd_rtol,
        msg=lambda msg: f"bwd dk mismatch\n\n{msg}\n",
    )
    relax_dv_err_tol: bool = (
        arch == "gfx942" and BATCH > 1 and SEQLEN_Q >= 1024 and SEQLEN_K >= 1024
    )
    torch.testing.assert_close(
        triton_dv,
        torch_dv,
        atol=2e-2 if relax_dv_err_tol else bwd_atol,
        rtol=2e-2 if relax_dv_err_tol else bwd_rtol,
        msg=lambda msg: f"bwd dv mismatch\n\n{msg}\n",
    )
```

Can you do it @lucas-santos-amd?

brunomazzottiamd (approved):
LGTM!

brunomazzottiamd:
Failure on **Triton Tests (MI325) / Shard 3**, coming from `op_tests/triton_tests/attention/test_pa_decode.py::test_paged_attn`, is intermittent. ~I've triggered a rerun.~ Let's trigger a rerun whenever possible just to be sure we're good to merge.

gyohuangxin:
@brunomazzottiamd  @lucas-santos-amd Thanks for the excellent work. 

> Failure on **Triton Tests (MI325) / Shard 3**, coming from `op_tests/triton_tests/attention/test_pa_decode.py::test_paged_attn`, is intermittent. ~I've triggered a rerun.~ Let's trigger a rerun whenever possible just to be sure we're good to merge.

I am seeing the same issue on other PRs too. Could we open another issue to track and fix it? Keeping our test runs stable is important as well. Thanks.

brunomazzottiamd:
> I am seeing the same issue on other PRs too. Could we open another issue to track and fix it? Keeping our test runs stable is important as well. Thanks.

@gyohuangxin, I created https://github.com/ROCm/aiter/issues/2882 to track the bug.

PR #2620, opened 2026-04-05
Add fused_dynamic_mxfp4_quant_moe_sort_hip

junhaha666:
python3 op_tests/test_moe_sorting_mxfp4.py -ek 256,8 -dim 7168
test on MI355:
image

image

## Submission Checklist


valarLip (approved):

csrc/kernels/quant_kernels.cu
```diff
@@ -1614,4 +1614,360 @@ void moe_smooth_per_token_scaled_quant_v2(
...
+
+            const int sorted_row = sorted_ids_offset + i;
+            if(threadIdx.x % num_thread_per_group == 0 && scale_k < scaleN_valid)
+            {
+                uint8_t bs_e8m0 = (__builtin_bit_cast(uint32_t, row_scale) >> 23) & 0xFF;
+                int addr        = fp4_scale_shuffle_id(scaleN_pad, sorted_row, scale_k);
+                scale[addr]     = bs_e8m0;
+            }
+
+            if(topk_id < topk)
```
zejunchen-zejun, line 1728: Hi @junhaha666
We found EP MXFP4 acc issue may be from here.
For stage1 and topk=1, the code is equal to `if(topk_id == 0)`. When TP mode, there is no issue because all tokens' topk_id is in this rank, while for EP mode, the token's top-1 expert maybe on other rank and here if condition exclude this token computation even when this token picks the expert in this rank.
CC: @ZhangLirong-amd

Issue #2517, opened 2026-03-28
[Bug] mha_batch_prefill returns all-zeros when KV cache > 8.4M pages (int32 overflow in CK element_space_size)


ChangyiYang:
## Description

`mha_batch_prefill_func` returns all-zero outputs when the paged KV cache has more than ~8,388,608 pages (with `page_block_size=1`). The root cause is an `int32` multiplication overflow in CK's `tensor_descriptor.hpp` that corrupts `buffer_size_`, causing AMD GPU buffer loads to silently return zeros for valid memory.

## Minimal Reproduction

```python
import torch
from aiter.ops.mha import mha_batch_prefill_func

NUM_PAGES = 8_401_239   # > 8,388,608 threshold
KV_START  = 16_435      # > 12,630 (beyond overflowed SRD range)

k = torch.randn(NUM_PAGES, 8, 64, dtype=torch.bfloat16, device="cuda")
v = torch.randn(NUM_PAGES, 8, 64, dtype=torch.bfloat16, device="cuda")
q = torch.randn(512, 16, 64, dtype=torch.bfloat16, device="cuda")
cu = torch.arange(0, 513, 128, dtype=torch.int32, device="cuda")
idx = torch.arange(KV_START, KV_START + 512, dtype=torch.int32, device="cuda")

o = mha_batch_prefill_func(q, k, v, cu, cu, idx, 128, 128, causal=True)
print(f"max_abs={o.float().abs().max():.4e}")  # 0.0000e+00 — BUG!
# Change NUM_PAGES to 5,000,000 → works fine (output ≈ 3.34e+00)
```

**Platform**: AMD MI300X, ROCm, aiter with CK `eb033ef20` (also verified on CK tag `rocm-7.2.1`)

## Root Cause

In CK's `tensor_descriptor.hpp` (line ~244 in `calculate_element_space_size_impl`):

```cpp
auto acc_new = acc_old + (lengths[i] - number<1>{}) * strides[i];
//                        ^^^^^^^^^ int32  ×  ^^^^^^^^^ int32  →  overflow!
```

With `num_pages=8,401,239` and `stride=512` (= `num_kv_heads × head_dim = 8 × 64`):
- `(8,401,238) × 512 = 4,301,433,856` — exceeds `uint32_max` (4,294,967,296)
- Wraps to `6,466,560`, making `buffer_size_ = 6,466,624`
[...]

PR #2641, opened 2026-04-07
[CK_TILE] Temporarily remove batch prefill KV cache overflow asserts

### LINKED ISSUE #2608: [Issue][CK] batch prefill int32 overflow assert aborts server when page_size=1 and KV cache > 4GB
## Summary

CK commit `6e558658e` (rocm-libraries#4999) added an int32 overflow assertion in `fmha_batch_prefill_kernel.hpp` that **aborts** the SGLang server process on MI300X GPUs whenever the per-layer KV cache exceeds ~4 GB with `page_size < kN0` (128). This is hit easily on MI300X (192 GB HBM) with small/medium models (e.g. Llama-3.2-1B allocates 6.3M tokens → 6 GB per-layer K cache).

**Before this assertion was added the overflow happened silently** (potential silent data corruption). The assertion correctly catches the bug, but now crashes the server at startup instead. We need the CK team to **extend the SRD-rebasing fix (already done for `page_size >= kN0`) to also cover `page_size < kN0` (especially page_size=1)**, so the overflow is actually fixed rather than just asserted.

## Reproduction

```bash
python -m sglang.launch_server \
  --model meta-llama/Llama-3.2-1B-Instruct \
  --port 11000 \
  --attention-backend aiter \
  --page-size 1
```

Server aborts with:
```
fmha_batch_prefill_kernel.hpp:661: Assertion `static_cast<int64_t>(num_total_pages - 1) * batch_stride_k
  <= static_cast<int64_t>(std::numeric_limits<index_t>::max())
  && "KV cache K offset overflow: exceed int32 max"' failed.
Fatal Python error: Aborted
```

## Root Cause Analysis

| Parameter | Value |
|---|---|
| GPU | MI300X (192 GB HBM) |
| Model | Llama-3.2-1B (8 KV heads, 64 head_dim, 16 layers) |
| `page_size` | 1 (default on ROCm) |
| KV tokens allocated | 6,320,125 |
| `batch_stride_k` | `1 × 8 × 64 = 512` |
| Max offset | `(6,320,125 − 1) × 512 = 3,235,903,488` |
| `INT32_MAX` | `2,147,483,647` |
| **Overflow** | **1.51× over limit** |

CK commit `7654fd478` / `6e558658e` added SRD-rebasing that uses 48-bit pointer arithmetic to fix the overflow **only when `kPageBlockSize >= kN0`** (i.e., page_size ≥ 128). For `page_size < kN0` the old int32-offset code path is preserved and the assertion catches the overflow:

```cpp
// fmha_batch_prefill_kernel.hpp, lines 487-496
if constexpr(kPageBlockSize < FmhaPipeline::kN0) {  // page_size=1 < kN0=128
    if(num_total_pages > 1) {
        assert(static_cast<int64_t>(num_total_pages - 1) * batch_stride_k <=
               static_cast<int64_t>(std::numeric_limits<index_t>::max()) &&
               "KV cache K offset overflow: exceed int32 max");
    }
}
```

## Requested Fix

1. **Extend the SRD-rebasing to page_size=1 / page_size < kN0** so the kernel actually handles >4GB KV caches instead of asserting. This is the proper CK-level fix.
2. As a short-term measure, consider converting the hard `assert` to a runtime error or returning an error code so the caller (SGLang/aiter) can handle it gracefully (e.g., cap allocation) rather than `abort()`ing the process.

## Environment

- Container: `rocm/sgl-dev:v0.5.10rc0-rocm700-mi30x-test-aiter-upgrade-20260330`
- aiter commit: `8e84e79d9`
- CK submodule commit: `7968368d9` (HEAD of CK in aiter)
- CK commit that introduced assertion: `6e558658e` (rocm-libraries#4999)
- SGLang commit: `4c0639036`

Jeff-Huang:
## Motivation

CK commit `6e558658e` (rocm-libraries#4999) added int32 overflow assertions in `FmhaBatchPrefillWithPagedKVCacheKernel` that **abort the SGLang server process** when the per-layer KV cache exceeds ~4GB with `page_size < kN0` (128). This is easily triggered on MI300X (192GB HBM) with small/medium models — e.g., Llama-3.2-1B with `page_size=1` allocates 6.3M tokens → 6GB per-layer K cache, causing `assert()` → `abort()` at server startup.

The assertions are correct (they catch real int32 offset overflow), but the fix for `page_size < kN0` is not yet ready. This PR temporarily removes the asserts to unblock serving while the proper fix (per-tile SRD rebase) is completed in a follow-up PR.

Fixes: #2608
Related: #2517

## Technical Details

Updates CK submodule to include rocm-libraries#6201, which removes 28 lines of `assert()` checks from `fmha_batch_prefill_kernel.hpp` (two identical blocks for the two `MakeKargs` overloads).

**Note:** Without the asserts, the overflow may produce silent incorrect results for `page_size < kN0` with KV cache > 4GB. A follow-up PR will add proper overflow handling via per-tile SRD rebase with 64-bit pointer arithmetic.

## Test Plan

- Verified SGLang server no longer aborts on startup with `page_size=1` and large KV cache (MI300X, Llama-3.2-1B)
- Verified no regression on existing batch prefill tests


HaiShaw (approved):

HaiShaw:
@valarLip please review and merge.

sammysun0711:
Hi @Jeff-Huang, thanks for quick workaround, I see CK merged overflow fix: https://github.com/ROCm/composable_kernel/commit/60ff5693c47074abefcde3625fa09392f8c3343b, could you please help to update the AITER submodule to check if it solves the original problem?

PR #1585, opened 2025-12-08
[Config][Qwen] tune gemm and moe

gbyu-amd:
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #2645, opened 2026-04-07
fix(ck_gemm): fix multi-arch build targeting and kernel dispatch across all CK GEMM modules

eppaneamd:
## Problem

Three related bugs caused incorrect kernel dispatch and non-reproducible builds, plus a fourth critical limitation in the build system.

**1. Non-deterministic builds.** CSV rows were filtered by the live GPU's `cu_num` at build time. CI builds without a GPU skipped this filter, accidentally compiling all CSV rows. The same `GPU_ARCHS` setting produced different `.so` files depending on whether a GPU was present — making builds non-reproducible and CI coverage misleading.

**2. Multi-arch dispatch collision.** C++ dispatch maps were keyed by `(M, N, K)` only. In `GPU_ARCHS=gfx942;gfx950` builds, shapes shared between architectures silently overwrote each other (last-writer-wins), selecting the wrong kernel with no error or warning.

**3. Hash collisions.** `IntTupleHash` used plain XOR, which is order-independent — permutations like `(304, 128, 7168)` and `(304, 7168, 128)` hash identically. On 160 real GEMM shapes, only 64 unique buckets were produced (60% collision rate), degrading most lookups to linear scan.

**4. No retuning path during `PREBUILD_KERNELS` builds.** When installing aiter with `PREBUILD_KERNELS=1`, there was no way to retune GEMM shapes against the target machine. Tuning CSVs shipped with the repo are tuned on specific hardware — if the target machine differs (different architecture, binned variant, or simply newer hardware with better kernels), all inference uses the pre-existing CSV results unchanged. There was no mechanism to rebenchmark those shapes on the actual deployment GPU and update the kernels accordingly during installation.

## Fix

**Build targeting** — `get_build_targets()` in `chip_info.py` returns `(gfx, cu_num)` pairs from `GPU_ARCHS` (or the live GPU when unset), deterministically regardless of GPU presence. Adds `gfx` as the first column in all tuning CSV schemas. `GPU_ARCHS=native` now correctly routes to live-GPU detection instead of raising.

**Runtime arch detection** — `get_gfx_runtime()` in `chip_info.py` always queries the live GPU via rocminfo, independent of `GPU_ARCHS`. The four GEMM Python dispatch modules (`gemm_op_a8w8`, `gemm_op_a4w4`, `batched_gemm_op_a8w8`, `batched_gemm_op_bf16`) use this for runtime arch selection.

**C++ dispatch key** — extended to `(gfx, cu_num, M, N, K)` for non-batched modules and `(gfx, cu_num, B, M, N, K)` for batched modules. Both are resolved at runtime from the current HIP device. `get_device_cu_num()` and `get_device_gfx()` are new helpers in `gemm_dispatch_utils.h`, cached per device ID via `SynchronizedCache` so that a single process calling `hipSetDevice()` across GPUs of different architectures always dispatches to the correct kernel.

**Hash function** — replaces per-module XOR `IntTupleHash` with a shared boost-style mixing hash (`0x9e3779b9` constant). On 160 real GEMM shapes: 160 unique buckets vs 64 before, zero collisions.

**CSV pretune workflow** — `aiter/utility/pretune.py` adds a retuning path for `PREBUILD_KERNELS` builds. All tuned shapes in the CSVs are rebenchmarked on the live GPU, results tagged with `(gfx, cu_num)`, and the inference `.so` rebuilt. Two modes:
- **Build-time**: `PRETUNE_MODULES=module_gemm_a8w8_blockscale_tune python setup.py develop` — runs automatically after the `PREBUILD_KERNELS` compilation phase.
- **Standalone**: `python3 aiter/utility/pretune.py module_gemm_a8w8_blockscale_tune` — retunes on an already-installed aiter without a full rebuild.

Existing rows for other architectures are preserved — a single CSV can contain tuning results for multiple GPUs.

## Scope

The diff is large because the same correctness fix is applied mechanically across 9 modules. The novel logic is confined to a small number of files — see the reviewer guide. All 10 codegen scripts and 10 C++ files must move together; any partial split leaves a broken dispatch map.

## Files changed

| File | Change |
|---|---|
| `aiter/jit/utils/build_targets.py` | **New** — `GFX_CU_NUM_MAP`, `get_build_targets_env()`, `filter_tune_df()`, `_parse_gpu_archs_env()` (torch-free) |
| `aiter/jit/utils/chip_info.py` | `get_build_targets()`, `get_gfx_runtime()`, `build_tune_dict()`, `write_lookup_header()`; fix `GPU_ARCHS=native` |
| `aiter/ops/gemm_op_*.py` (4 files) | `get_gfx()` → `get_gfx_runtime()` |
| `csrc/include/gemm_dispatch_utils.h` | **New** — `GemmDispatchHash`, `GemmDispatchMap<>`, `get_device_cu_num()`, `get_device_gfx()` |
| 9× `gen_instances.py` + 1× `gen_instances_cktile.py` | Dispatch key `(M,N,K)` → `(gfx,cu_num,M,N,K)`; shared helpers |
| 10× C++ `.cu` files | `IntTupleHash` removed; `GemmDispatchMap<>` type alias |
| 8 primary + 17 model_configs tuning CSVs | `gfx` column prepended (fmoe CSVs intentionally excluded — different dispatch path) |
| 8× `README.md` | CSV schema examples updated |
| `aiter/jit/core.py` | Fix `update_config_files()`: duplicate CSV entries → `logger.warning` instead of crash; `dedup_keys` now includes `gfx` when present in merged CSV (prevents false duplicate detection across arch-sharing-cu_num targets) |
| `aiter/utility/base_tuner.py`, `gemm_a8w8_blockscale_tune.py`, `gemm_a4w4_blockscale_tune.py` | Filter tuning data by `(gfx, cu_num)` instead of `cu_num` alone; auto-populate `gfx` when absent; `get_gfx` → `get_gfx_runtime` so tuned CSV rows are tagged with live GPU arch |
| `setup.py` | Read `PRETUNE_MODULES` env var; invoke `run_pretune_modules()` after `ThreadPoolExecutor` completes |
| `aiter/utility/pretune.py` | **New** — `PRETUNE_MODULES` build-time and standalone retuning CLI; `run_pretune()`, `run_pretune_modules()`, `_parse_module_list()` |
| `op_tests/test_pretune.py` | **New** — 9 test functions (script resolution, CSV deduplication, multi-path merging, module list parsing, `_unsupported` exclusion), no GPU required |
| `op_tests/test_gemm_codegen.py` | New unit tests for build targeting and runtime dispatch; separator-only `GPU_ARCHS` guard |
| `csrc/ck_gemm_a8w8{,_blockscale,_bpreshuffle}/gemm_*_tune.py` (3 files) | Fix `_clear_op_caches()`: clear module-level CSV dict caches alongside `lru_cache` — `--compare`/`--run_config` benchmark flows now pick up updated CSVs correctly |

## Follow-up

- Python CSV lookup (`get_CKGEMM_config`) and C++ dispatch are decoupled — the Python side finding a shape does not guarantee the C++ map dispatches it. Whether to unify these is left as a follow-up.
- `get_gfx()` picks the last arch in multi-arch `GPU_ARCHS` lists (e.g. `gfx950` on MI300X with `GPU_ARCHS=gfx942;gfx950`). The codegen path (`gen_instances.py`) is resolved via `get_build_targets()` + `filter_tune_df()`. The dispatch key and tuning path now use `get_gfx_runtime()`. Broader runtime callers (`fused_moe.py`, `mla.py`, `attention.py`, capability gating in general) still use `get_gfx()` — left as a follow-up.
- CI validation only checks `.so` count. Inspecting generated `*_lookup.h` headers for `{"gfx942",` and `{"gfx950",` key prefixes would catch dispatch regressions earlier.

## Testing

- **Unit tests** (`test_gemm_codegen.py`): 31 tests covering `get_build_targets()` routing (including separator-only `GPU_ARCHS` guard), `build_tune_dict()` filtering, `write_lookup_header()` key format — no GPU required. Runtime dispatch section skipped gracefully without torch.
- **Pretune tests** (`test_pretune.py`): 84 tests covering script resolution, CSV deduplication, multi-path merging, module list parsing, `_unsupported` exclusion — no GPU required.
- **Repro CSVs**: shapes from the original regression verified correct dispatch on MI300X.
- **Multi-arch**: `GPU_ARCHS=gfx942;gfx950` build — lookup headers contain both `{"gfx942",` and `{"gfx950",` keys.
- **Runtime**: `test_gemm_a8w8.py` passes on MI300X.


csrc/include/gemm_dispatch_utils.h
```diff
@@ -0,0 +1,117 @@
...
+// multiProcessorCounts are not a supported inference scenario; the cached value
+// reflects whichever device is current at the time of the first dispatch call.
+// ---------------------------------------------------------------------------
+// Note: HIP_CALL from aiter_hip_common.h is not used here to keep this header
+// self-contained — aiter_hip_common.h pulls in CK and logger headers that may
+// not be available in all build configurations.
+inline int get_device_cu_num()
+{
+    static const int cu_num = []() {
+        int device = 0;
```
valarLip: force to device 0 may introduce some error..

eppaneamd: @valarLip thanks! The function should behave identically whether the device is 0 or e.g. -1, it is just an initialization value. hipGetDevice writes to that variable anyway upon successful query and otherwise gives a runtime error. I will set it to -1 for better readability and clearer intent.

eppaneamd:
## Reviewer guide — what changed and why

Large diff, small core change. The actual new logic is confined to a small number of files. Everything else is the same fix applied mechanically across 9 modules, plus data-only CSV column additions. There are no reasonable boundaries for splitting this PR: the helpers, the 10 codegen scripts, and the 10 C++ files must all move together — any partial split leaves either dead code or a broken runtime dispatch map.

---

### Four bugs, four fixes

```
BUILD TIME                    BEFORE                  AFTER
  Kernel selection            GPU_ARCHS != .so        GPU_ARCHS == .so
  depends on...               (env non-deterministic) filter by (gfx, cu_num)

RUNTIME DISPATCH              key = (M, N, K)         key = (gfx, cu_num, M, N, K)
  C++ static map              multi-arch collision     gfx+cu_num from HIP at runtime

HASH FUNCTION                 XOR (commutative)       boost-style mixing
  for dispatch map            high collision rate      0 collisions on GEMM shapes

RETUNING                      no path during          PRETUNE_MODULES env var
  PREBUILD_KERNELS build       PREBUILD_KERNELS        rebenchmarks shapes on live GPU;
                                                       rebuilds inference .so in-place
```

---

### Where the novel logic lives — 9 files to review

| File | What is new |
|---|---|
| `aiter/jit/utils/build_targets.py` | **New file.** `GFX_CU_NUM_MAP`; `get_build_targets_env()` (torch-free, safe in CI); `filter_tune_df()`; `_parse_gpu_archs_env()` (raises `RuntimeError` on separator-only `GPU_ARCHS` like `" ; "` that would otherwise silently produce an empty target list). |
| `aiter/jit/utils/chip_info.py` | **New functions.** `get_build_targets()` (priority: `GPU_ARCHS` → live GPU → error); `get_gfx_runtime()` (rocminfo, ignores `GPU_ARCHS`, for runtime dispatch); `build_tune_dict()` / `build_tune_dict_batched()`; `write_lookup_header()`. Imports `GFX_MAP`, `filter_tune_df`, `get_build_targets_env` from `build_targets.py`. Fix: `GPU_ARCHS=native` routing. |
| `csrc/include/gemm_dispatch_utils.h` | **New file.** `GemmDispatchHash` / `BatchedGemmDispatchHash` (boost-style, replaces per-module XOR); `GemmDispatchMap<>` / `BatchedGemmDispatchMap<>` type aliases; `get_device_cu_num()` and `get_device_gfx()` — both cached per device ID via `SynchronizedCache<int, T>` keyed by device ID, so that a single process calling `hipSetDevice()` across GPUs of different architectures always gets the correct values. `hipGetDeviceProperties(&prop, device)` is called with an explicit device ID in both factories — no dependency on the thread-local current device. |
| `aiter/jit/core.py` | **Pre-existing bug fix** (commit `94a6edda3`, found during CI). `update_config_files()`: `raise RuntimeError` → `logger.warning` on duplicate CSV entries; filter in memory so `/tmp/` output is clean. Blocked `build_aiter_image` non-recoverably in Docker. Also: `dedup_keys` now includes `gfx` when present in the merged tuned CSV — prevents false duplicate detection when merging tuning results from two architectures that share the same `cu_num` (e.g., a future gfx12xx and MI355X both at 256 CUs). |
| `aiter/utility/pretune.py` | **New file.** `PRETUNE_MODULES` env var triggers retuning after the `PREBUILD_KERNELS` build phase. `run_pretune()` builds the tune `.so`, extracts unique shapes from the merged CSV as an untune input, runs the tune script with `--all`, and rebuilds the inference `.so`. Standalone CLI (`python3 aiter/utility/pretune.py`) covers the same flow without a full rebuild. `_parse_module_list()` is a shared helper used by both paths — expands `"all"` to supported modules only and splits comma-separated lists. `run_pretune_modules()` (setup.py path) deduplicates by `(script, config_attr)` key to avoid retuning the same CSV twice. `_SCRIPT_FALLBACK` maps cktile variants to their parent scripts; two `blockscale_bpreshuffle` variants are skipped (no tune script writes to their CSV). |
| `aiter/utility/base_tuner.py` | Tuner now filters by `(gfx, cu_num)` instead of `cu_num` alone. Auto-populates `gfx` when absent for backwards compatibility with old untune CSVs. Import switched to `get_gfx_runtime` so tuning tags CSV rows with live GPU arch. |
| `setup.py` | Reads `PRETUNE_MODULES` env var; invokes `run_pretune_modules()` after `ThreadPoolExecutor`; passes `build_one_module` closure so pretune reuses the same compiler flags. |
| `op_tests/test_gemm_codegen.py` | **New tests.** 31 tests: `get_build_targets()` env routing including separator-only `GPU_ARCHS` guard (no GPU needed); runtime dispatch key selection (skipped gracefully without GPU). |
| `op_tests/test_pretune.py` | **New tests.** 84 tests: script resolution for all 10 tune modules, CSV deduplication and multi-path merging, `_SCRIPT_FALLBACK` key validity, module list parsing, `_unsupported` exclusion in "all" expansion. No GPU required. |
| `csrc/ck_gemm_a8w8/gemm_a8w8_tune.py`, `ck_gemm_a8w8_bpreshuffle/gemm_a8w8_bpreshuffle_tune.py`, `ck_gemm_a8w8_blockscale/gemm_a8w8_blockscale_tune.py` | Fix `_clear_op_caches()`: this PR introduced module-level dict caches (`_CKGEMM_CONFIG_CACHE`, `_GEMM_QUANT_TYPE_CACHE`) alongside `@lru_cache`. The existing cache-clear hooks only called `cache_clear()` on the lru — the next call then re-entered the function body, found the dict already populated, and returned stale CSV data. Both layers must be cleared for `--compare`/`--run_config` benchmark flows to pick up updated CSVs. The blockscale tuner additionally cleared the wrong function entirely. |

---

### Mechanical adaptation

**10x codegen scripts (9× `gen_instances.py` + 1× `gen_instances_cktile.py`)** — structurally identical substitution: `import torch` removed; `chip_info` path boilerplate added; `gen_lookup_dict` loop → `write_lookup_header()` (identical across all). The `gen_instances_cktile.py` in `ck_gemm_a8w8_blockscale/` follows the same pattern. The two MoE modules (`ck_gemm_moe_2stages_codegen`, `ck_tile_gemm_moe_2stages`) are intentionally out of scope — they use a different codegen path. All non-batched modules write a 5-tuple C++ key `(gfx, cu_num, M, N, K)` → kernelId; batched modules write a 6-tuple `(gfx, cu_num, B, M, N, K)` → kernelId. The `build_tune_dict` call has four variants driven by each module's CSV schema:
- `build_tune_dict(df, defaults, kernels_list)` — `ck_gemm_a8w8`, `ck_deepgemm`, `ck_gemm_a4w4_blockscale`
- `build_tune_dict(..., libtype="ck"/"cktile")` — modules with multi-libtype CSVs (`ck_gemm_a8w8_blockscale`, `ck_gemm_a8w8_bpreshuffle`, `ck_gemm_a8w8_blockscale_bpreshuffle`, `cktile_gemm_a8w8_bpreshuffle`)
- `build_tune_dict(..., kernels_by_name=...)` — `cktile_gemm_a8w8_bpreshuffle` only: cktile matches by kernel name string rather than integer ID
- `build_tune_dict_batched(...)` — `ck_batched_gemm_a8w8`, `ck_batched_gemm_bf16`: 6-tuple key `(gfx, cu_num, B, M, N, K)` vs the 5-tuple `(gfx, cu_num, M, N, K)` used elsewhere

**10x C++ `.cu` source files** — `#include "gemm_dispatch_utils.h"`; per-module `IntTupleHash` removed; old `unordered_map` definition replaced by `GemmDispatchMap<>` / `BatchedGemmDispatchMap<>` type alias.

**4x Python dispatch modules** — `get_gfx()` → `get_gfx_runtime()`; key gains `gfx` prefix.

**3x Python tuner files** (`base_tuner.py`, `gemm_a8w8_blockscale_tune.py`, `gemm_a4w4_blockscale_tune.py`) — import changed from `get_gfx` to `get_gfx_runtime` so tuned CSV rows are always tagged with the live GPU arch, not `GPU_ARCHS` env (which may contain a multi-arch list like `gfx942;gfx950`).

**8 primary + 17 model_configs tuning CSVs** — `gfx` prepended as first column, no data values changed. fmoe CSVs intentionally excluded (different dispatch path).

**`aiter/utility/base_tuner.py`** — tuner now filters by `(gfx, cu_num)` instead of `cu_num` alone, matching the updated CSV key schema. Auto-populates `gfx` column when absent (backwards compatibility with old untune CSVs).

**`setup.py`** — reads `PRETUNE_MODULES` env var; invokes `run_pretune_modules()` after the `ThreadPoolExecutor` completes (before the new FlyDSL AOT block from main); passes `build_one_module` closure so pretune reuses the same compiler flags as the main build.

**8x READMEs + chip_info.py logging** — docs and `print()` → `logger.warning()`, no logic change.

---

### Suggested review order

1. `build_targets.py` — small, self-contained
2. `chip_info.py` — `get_build_targets()`, `get_gfx_runtime()`, `build_tune_dict()`
3. `gemm_dispatch_utils.h` — `GemmDispatchHash`, `get_device_cu_num()`
4. One `gen_instances.py` (e.g. `csrc/ck_gemm_a8w8_blockscale/`) — establishes the pattern; the other 9 codegen scripts are structurally identical
5. `test_gemm_codegen.py` — validates build-time filter and runtime dispatch key
6. `base_tuner.py` — `(gfx, cu_num)` filter change; small, focused
7. `setup.py` — `PRETUNE_MODULES` integration; small addition after the executor block
8. `pretune.py` — `run_pretune()` and `run_pretune_modules()`; `_SCRIPT_FALLBACK` table
9. `test_pretune.py` — validates script resolution, CSV handling, module list parsing
10. `core.py` — small, independent

csrc/include/gemm_dispatch_utils.h
```diff
@@ -0,0 +1,121 @@
...
+// its CU count are fixed once the process starts.
+//
+// Note: multi-GPU processes where different devices have different
+// multiProcessorCounts are not a supported inference scenario; the cached value
+// reflects whichever device is current at the time of the first dispatch call.
+// ---------------------------------------------------------------------------
+inline int get_device_cu_num()
+{
+    static const int cu_num = []() {
+        int device = -1;
```
valarLip: this part is not safe if we are on a node that have different gfx

aiter/jit/core.py
yzhou103, line 270: I changed this back to raising an error intentionally. The goal here is not just to auto-resolve duplicate shape entries, but to make the submitter explicitly aware that duplicates were introduced. Silent auto-fix can hide config quality issues and let bad updates slip through unnoticed. Failing fast gives clear visibility, forces cleanup at the source, and helps keep the merged tuning configs maintainable.

eppaneamd: Thanks @yzhou103! I see - I made the change initially because this was breaking the build_aiter_image CI (raise aborts the wheel build). Would it work to raise only when the save-back actually fails, and log a warning otherwise? The warning still includes the full duplicate list and which files were cleaned, so visibility is preserved. Or have I misunderstood something?

yzhou103: Thanks, that makes sense, and I agree warning-only would be functionally safe if merged_df is also updated in the current run. My hesitation is more about policy than correctness: we would still like main to stay as clean and stable as possible, and a patch introducing many duplicate shape entries is not really the expected pattern. And we also want the person updating the CSVs to be very explicitly aware of exactly what happened and which shape entries were duplicated, rather than this becoming a mostly transparent cleanup path.

yzhou103: I noticed there are many duplicate shape entries in this PR, but my understanding is that it is not adding any new tuned shapes, so I am a bit confused about where these duplicates are coming from.

aiter/configs/model_configs/a8w8_blockscale_bpreshuffle_tuned_gemm_qwen3.5_397b.csv
```diff
@@ -1,554 +1,1114 @@
...
+gfx950,256,512,20480,4096,cktile,26,0,58.6077,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_2,1465.67,1824.93,0.0
+gfx950,256,1024,20480,4096,cktile,11,0,112.2264,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_1,1530.82,1158.58,0.0
+gfx950,256,2048,20480,4096,cktile,26,0,198.7021,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_2,1729.21,886.56,0.0
+gfx950,256,4096,20480,4096,cktile,28,0,371.7097,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_4,1848.74,722.16,0.0
+gfx950,256,8192,20480,4096,cktile,11,0,708.3984,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_1,1940.14,639.45,0.0
+gfx950,256,16384,20480,4096,cktile,27,0,1387.4271,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_3,1981.21,592.52,0.0
+gfx950,256,32768,20480,4096,cktile,28,0,2721.3003,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_4,2020.2,573.36,0.0
+gfx950,256,1,16,4096,cktile,12,0,9.9757,a8w8_blockscale_cktile_16x128x256_1x4x1_16x16x128_intrawave_0x1x0_3,0.01,6.98,0.0
```
yzhou103: there are duplicated shapes in this csv.

eppaneamd: @yzhou103 thanks, fixed now.

aiter/jit/core.py
eppaneamd: @yzhou103 quick comment: there should not be duplicates within CSVs now, and duplicates across CSVs (e.g. model configs) are not introduced by this PR.

eppaneamd: For example:

qwen3_235b — introduced in ba596dc3c (tune gemm and moe for Qwen3 MoE models #1585)
qwen3.5_397b — introduced in 3fcfeac70 (tuned qwen3.5 gemm #2485)

yzhou103:
to resolve the duplicated shapes quickly, we can run relative test like

>  python op_tests/test_gemm_a8w8_blockscale.py

valarLip (approved):

valarLip:
hold till atom test pass

PR #2648, opened 2026-04-08
feat: _flash_attn_forward add out args

RuibinCheung:
## Motivation

The output tensor of `_flash_attn_forward` is default bshd. But the aiter mha C++ API support pass out args to support user-defined layout (such as sbhd). The user call aiter through python api. If the output tensor is bshd that causes extra permute overhead. We hope to align python and C++ API to support multi layouts.

## Submission Checklist


minmengdie:
please add testcase in test_mha.py for the different layout of out tensor to test the feature has been ready.

aiter/ops/mha.py
minmengdie, line 1245: user call aiter python api through aiter.flash_attn_func, only modify the _flash_attn_forward api maybe not enough

minmengdie, line 1245: please add testcase in test_mha.py for the different layout of out tensor to test the feature has been ready.

minmengdie, line 1245: please check whether the aiter.flash_attn_varlen_func api also needs to support this feature.

RuibinCheung: Our user call `_flash_attn_forward` currenttly, update it is enough to our. The aiter.flash_attn_func is exposed to user. Change arguments of API has more risk.

RuibinCheung: The `_flash_attn_forward` is an internal API. It will not effect top-level user.

RuibinCheung: The `_flash_attn_varlen_forward` has supported this feature. Ref: https://github.com/ROCm/aiter/blob/b025f0940f000176bb76901bf7949b9522d0cdfd/aiter/ops/mha.py#L2020

valarLip (approved):

PR #2652, opened 2026-04-08
[OPUS] mma step_k, dword copy by set_slice and inline asm for tr_load

kaiyang-1:
**Enhancements to tiled MMA operations:**
* Added multiple overloads of the `step_k` method to `tiled_mma_adaptor`, enabling stepwise computation over the K dimension for both arrays and vectors, with and without scaling factors. These overloads support more flexible and efficient tiled MMA operations, allowing partial computation and improved composability.
 
**Performance optimization for memory copy:**
* Updated `set_slice_impl` to optimize memory copying for sub-dword scalar types (such as bf16, fp16, fp8) when source and destination slices are dword-aligned and contiguous. This copies at dword granularity to avoid v_bfi_b32 generated by compiler from per-element insert.
 
**Shared memory transpose load refactor:**
* Refactored `smem::_tr_load` to use inline assembly and support immediate offsets and generalized the implementation for different scalar types, improving code clarity and enabling more efficient loads with compile-time known offsets.


carlushuang (approved):
LGTM

PR #2655, opened 2026-04-08
[FIX] fix ck tile shim perf eval in mha bwd v3

JaxChen29:
## Motivation

ENABLE_CK=0 builds rely on ck_tile_shim.h, but the shim previously only launched the callables and returned 0, which made asm-only bwd_v3 perf evaluation report 0.000 ms and inf throughput/bandwidth. This PR brings the shim closer to CK host-side timing behavior so asm paths can be performance-tested without enabling full CK.

## Technical Details

Extend csrc/include/ck_tile_shim.h with CK-like timing and launch helpers, including gpu_timer, cpu_timer, launch_and_check, and warmup/repeat timing loops.
Update ck_tile::launch_kernel() to honor time_kernel_, cold_niters_, nrepeat_, and is_gpu_timer_ instead of always returning 0.
Add preprocess-aware and cache-flush-aware timing support via flush_icache(), preprocess_profiling_impl(), timing_loop_flush_cache_impl(), launch_kernel_time_mask(), and launch_kernel_time_mask_flush_cache().
Keep the implementation HIP-only so ENABLE_CK=0 asm builds gain usable perf timing without depending on the full CK host runtime.

## Test Plan

run smoke test

## Test Result

all passed, evaluate performance correctly.

## Submission Checklist


csrc/include/ck_tile_shim.h
```diff
@@ -35,11 +37,278 @@ constexpr T log2e_v = static_cast<T>(1.4426950408889634);
 
 inline int get_warp_size() { return 64; }
```
valarLip: can't hardcode to 64...

valarLip (approved):

PR #2661, opened 2026-04-09
[TRITON] Prevent NUM_KSPLIT from reducing K dim below GROUP_K

vgokhale:
This sets NUM_KSPLIT such that the K dim doesn't drop below GROUP_K.


k50112113 (dismissed):

azaidy (approved):
LGTM!

PR #2664, opened 2026-04-09
fix(setup.py): accept FlyDSL dev/rc builds when version matches pinned release

guangzlu:
**Problem**
In develop mode, setup.py ensures FlyDSL matches the pinned version by:

Reading the installed version with importlib.metadata.version("flydsl").
Comparing it to the pinned string (0.1.2).
On any mismatch or import error, running pip install flydsl==0.1.2.
For an editable/local FlyDSL install, the reported version is often 0.1.2.dev462 (or similar), which is not equal to the string "0.1.2", so the script always took the “mismatch” path and invoked pip install. That is unnecessary and can break workflows where:

- FlyDSL is installed from a git checkout (dev/local version), or
- PyPI has no wheel for the current Python ABI (e.g. cp313 while wheels exist only for cp310/cp312).

**Solution**
Add _flydsl_matches_pinned(installed_ver, pinned_ver) that uses packaging.version.Version and compares the .release tuples (e.g. (0, 1, 2)), so 0.1.2, 0.1.2.dev462, and other suffixes on the same X.Y.Z are treated as matching the pin.
If packaging is not importable, fall back to the previous strict string comparison for safety.

### DIFF
--- a/aiter/ops/flydsl/__init__.py
+++ b/aiter/ops/flydsl/__init__.py
@@ -14,6 +14,16 @@ from .utils import is_flydsl_available
 
 _REQUIRED_FLYDSL_VERSION = "0.1.2"
 
+
+def _flydsl_matches_pinned(installed_ver: str, pinned_ver: str) -> bool:
+    """Same X.Y.Z as pin, allowing dev/rc/post/local segments (e.g. 0.1.2.dev462 == 0.1.2)."""
+    try:
+        from packaging.version import Version
+    except ImportError:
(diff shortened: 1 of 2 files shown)


zufayu:
@guangzlu This PR has been inactive for over 2 months. Is the setup.py fix for FlyDSL dev/rc builds still applicable? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2672, opened 2026-04-09
[TRITON] Add separate ROPE computation path for unified attention

anhminhnguyenhoang:
## Motivation

The unified attention kernel on main treats the full Q/K head dimension as a single block, requiring it to be padded to the next power of two for Triton's `tl.dot`. For MLA-style architectures (e.g. DeepSeek, GLM5) where Q/K head dim is d+r (lora + rope), this can be very wasteful -- for example, d+r = 576 would pad to 1024, resulting in massive excess loading and compute for nearly half-empty tiles.

This PR enables independent head dimensions for Q/K and V and splits non-power-of-2 Q/K head dims into two efficient dot products, each for lora and rope considering they have power-of-2 dims.

## Technical Details

Core idea: Allow Q and K head dim to be d+r (e.g. lora rank + rope dim) while V head dim is dv, which can differ. The behavior depends on whether d+r is a power of two:

1. **Kernel** (`aiter/ops/triton/_triton_kernels/attention/unified_attention.py`):

   * Added `HEAD_SIZE_V`, `HEAD_SIZE_V_PADDED`, `ROPE_SIZE`, `ROPE_SIZE_PADDED`, `HAS_ROPE` constexpr params to both 2d and 3d kernels.
   * If d+r is a power of two (e.g. 128, 256): only the Q/K vs V decoupling is applied. The full Q/K head dim is loaded and processed in a single `tl.dot`. No RoPE separation is needed.
   * If d+r is not a power of two (e.g. 576, or 192): d is set to the power-of-two lora part, and r is set as the remaining part. Q_rope and K_rope are loaded from the `[d, d+r)` slice and accumulated via a second `tl.dot`: `S = scale * (Q[:d] @ K[:d]^T + Q[d:d+r] @ K[d:d+r]^T)`. For example, d+r = 192 splits into d=128, r=64 -- two dot products over power-of-2 tiles instead of one padded to 256.
   * V loading and output store use separate `offs_dv` / `dim_mask_v` for the independent V head dimension.
   * No structural changes to main's kernel layout -- no extracted helper functions or tile loop refactoring.

2. **Wrapper** (`aiter/ops/triton/attention/unified_attention.py`):

   * Added `rope_size` parameter with auto-detection: when d+r is not a power of 2, it automatically splits into d and r, or can be explicitly set.
   * Added `head_size_v` / `head_size_qk` decoupling throughout the dispatch logic.
   * Passes new kernel params: `HEAD_SIZE_V`, `HEAD_SIZE_V_PADDED`, `ROPE_SIZE`, `ROPE_SIZE_PADDED`, `HAS_ROPE`.
   * TILE_SIZE fix: enforced `TILE_SIZE >= 32` to prevent Triton compiler crashes for FP8 on gfx950 and correctness failures for BF16 with BLOCK_M=16. Added `TILE_SIZE=128` for quantized large-prefill on CDNA.

3. **Tests** (`op_tests/triton_tests/attention/test_unified_attention.py`):

   * Changed `HEAD_SIZES` to tuples: `[(128, 128), (256, 256), (192, 128)]` -- the (192, 128) case triggers the two-dot-product RoPE split (d=128, r=64, dv=128).
   * Added `make_unified_attn_inputs` helper for constructing test inputs with asymmetric head sizes.
   * Updated `ref_paged_attn` to handle `head_size_v != head_size_qk` for the V dimension.
   * Minimal diff vs main (115 insertions, 10 deletions) -- main's test structure, parametrize, and FP8 descale logic are fully preserved.

4. **Benchmark** (`op_tests/op_benchmarks/triton/bench_unified_attention.py`):

   * Added `-d`, `-dv`, `-rope_size`, `-decode`, `-fp8_full`, `-causal` CLI arguments.
   * Uses `make_unified_attn_inputs` for input construction.
   * Wired `rope_size` into the `unified_attention()` call.
   * Added `-test` correctness verification mode.

## Test Plan

- Full pytest suite: all head sizes (128,128), (256,256), (192,128) x FP8/no-quant x BF16 x sliding window x softcap x block sizes x sequence patterns
- Benchmark `-test` mode with d=192/dv=128 (RoPE split active), d=128/dv=128 (no split), FP8, batch decode, prefill, and default 18-config sweep

## Test Result

- Full pytest: 2160/2160 passed (0 failures)
- Benchmark correctness (`-test`): passed for all 6 configurations -- RoPE-split (d+r=192, d=128, r=64), non-split (d+r=128), FP8, batch decode, prefill, and default sweep. Validated with `torch.testing.assert_close` (atol=1.5e-2, rtol=1e-2 for BF16/FP16; atol=1.5e-1 for FP8)

## Submission Checklist


aiter/ops/triton/attention/unified_attention.py
```diff
@@ -22,14 +22,17 @@ def select_2d_config(
...
 
     BLOCK_M = (
         16 if num_queries_per_kv <= 16 else triton.next_power_of_2(num_queries_per_kv)
     )
-
-    TILE_SIZE = 32 if arch.name == "gfx1201" else 16 if arch.is_rdna else 64
+    # TILE_SIZE >= 32 is required: TILE_SIZE=16 causes Triton compiler
+    # crashes for FP8 on gfx950 (unsupported 16x16 FP8 MFMA) and
+    # correctness failures for BF16 with BLOCK_M=16.
+    TILE_SIZE = 32 if arch.is_rdna else 64
```
cagrikymk: `TILE_SIZE = 32 if arch.name == "gfx1201" else 16 if arch.is_rdna else 64`
Existing code seems to use TILE_SIZE=64 on cdna archs.

How does the crash happen in that case?

aiter/ops/triton/attention/unified_attention.py
```diff
@@ -69,12 +74,11 @@ def select_3d_config(
 ):
     reduce_num_warps = 2
     attn_warps = 2
-    TILE_SIZE = min(64, triton.next_power_of_2(block_size))
-    # MAX_SEGMENTS = min(128, math.ceil(max_seqlen_k / TILE_SIZE))
+    TILE_SIZE = max(32, min(64, triton.next_power_of_2(block_size)))
     num_segments = math.ceil(target_num_prgms / num_2d_prgms)
     num_segments = triton.next_power_of_2(num_segments)
     num_segments = min(num_segments, 128)
-    MIN_SEGMENTS = 16 if TILE_SIZE <= 16 else 8
```
juuso-oskari: I would reconsider these

aiter/ops/triton/attention/unified_attention.py
```diff
@@ -40,7 +43,7 @@ def select_2d_config(
     # pure decode config
     else:
         # to not have masking when loading KV
-        TILE_SIZE = min(64, triton.next_power_of_2(block_size))
+        TILE_SIZE = max(32, min(64, triton.next_power_of_2(block_size)))
```
juuso-oskari: and this

aiter/ops/triton/attention/unified_attention.py
anhminhnguyenhoang: The issue was with Triton 3.5, Claude reasons the IR picks an invalid 2d instr `amd_mfma<version=4, instrShape=[16,16]>` for gfx950.

```
/home/anguyenh/aiter_ua/aiter/ops/triton/_triton_kernels/attention/unified_attention.py:396:27: error: Unsupported conversion from 'f8E4M3FN' to 'f32' with rounding mode rtne
        acc += tl.dot(P.to(V.dtype), V)
                          ^
/home/anguyenh/aiter_ua/aiter/ops/triton/_triton_kernels/attention/unified_attention.py:396:27: error: failed to legalize operation 'tt.fp_to_fp' that was explicitly marked illegal
        acc += tl.dot(P.to(V.dtype), V)
                          ^
#blocked = #ttg.blocked<{sizePerThread = [16, 1], threadsPerWarp = [8, 8], warpsPerCTA = [1, 2], order = [0, 1]}>
#blocked1 = #ttg.blocked<{sizePerThread = [1, 16], threadsPerWarp = [8, 8], warpsPerCTA = [2, 1], order = [1, 0]}>
#blocked2 = #ttg.blocked<{sizePerThread = [4, 4], threadsPerWarp = [2, 32], warpsPerCTA = [2, 1], order = [1, 0]}>
#mma = #ttg.amd_mfma<{version = 4, warpsPerCTA = [2, 1], instrShape = [16, 16], isTransposed = true}>
```

With Triton 3.6, it uses
```
amd_mfma<{version = 4, warpsPerCTA = [1, 2], instrShape = [16, 16, 32], isTransposed = true}>
amd_mfma<{version = 4, warpsPerCTA = [2, 1], instrShape = [16, 16, 128], isTransposed = true}>
```

I'll revert these changes since after revisiting, the original config is doing better

cagrikymk (approved):
LGTM!

op_tests/triton_tests/attention/test_unified_attention.py
```diff
@@ -9,9 +9,9 @@
 from aiter.ops.triton.attention.unified_attention import unified_attention
 from aiter.ops.triton.utils.types import e4m3_dtype
 
-NUM_HEADS = [(4, 4), (16, 2)]
-HEAD_SIZES = [64, 128]
-BLOCK_SIZES = [16, 64]
+NUM_HEADS = [(4, 4), (8, 2), (16, 2)]
+HEAD_SIZES = [(128, 128), (256, 256), (192, 128)]
```
cagrikymk, line 13: Could we have head size 64 back? I think the test cases can contain:
[(64,64), (128,128),(192,128)]

Given head_size=64 is used for GPT-OSS, I think we can use that in the unit test instead of 256.

vgokhale: Yeah @anhminhnguyenhoang  please don't increase number of UTs to the extent possible. Is block size 48 critical? Why not use 64? What is 192 used for? We cannot do deepseek with unified attention anyways.

anhminhnguyenhoang: I'll bring back GPT-OSS head size and removed block size 48 (it's to test non-power-of-2 value which probably run similar test to 64).

Also, could you shed more light on why UA can't be used for Deepseek? The current `unified_attention_sparse_mla` seems to cater Deepseek v3.2 or GLM5 in decode. Should we try to merge both to one kernel eventually?

zufayu:
@anhminhnguyenhoang This PR has been inactive for over 2 months. Is the separate ROPE computation path still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2687, opened 2026-04-10
Allow preallocated moe sorting buffer

tpopp:
## Motivation

In locations like vLLM, they have abstracted the calling code to accept preallocated workspaces to use as output buffers.

## Technical Details

Allow an output buffer here as an optional argument and otherwise allocate a new buffer.

## Test Plan

Correctness testing is extended to also check that values are correct and the output buffer was used.

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -35,6 +35,7 @@ def _moe_sorting_impl(
     num_local_tokens,
     dispatch_policy,
     use_opus,
+    moe_buf=None,
 ):
     device = topk_ids.device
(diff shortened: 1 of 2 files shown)


tpopp:
A concern was raised around why the fix can't be done on the caller side. The reason output buffers are desired is because in non HIPGraph cases, they don't want ~3 different temporary buffer allocations for intermediate workspaces and the output but a single allocation to limit overhead.

valarLip:
we have internal logic and wlll use it for different way, like fuse quant or not fuse quant, the size and datatype for this buffer will change, this is can't for outside

PR #2688, opened 2026-04-10
[TRITON] fix torch.compile graph break in arch_info.get_arch()

lauri9:
## Motivation

When any AITER Triton operator is used inside a `torch.compile(fullgraph=True)` region, the call to `triton.runtime.driver.active.get_current_target().arch` inside `arch_info.get_arch()` causes a **graph break**. Dynamo cannot symbolically trace this opaque runtime call, so it falls back to eager execution for that subgraph. With `fullgraph=True` this surfaces as a hard `torch._dynamo.exc.Unsupported` error; without it, the graph is silently split, losing compilation benefits.

The graph break potentially affects every operator that depends on `arch_info`. There are approximately 25 call sites spanning attention kernels (sage, lean_atten, mha, hstu, extend, mla_decode_rope), GEMM kernels (a8w8, afp4wfp4, a16wfp4, a8wfp4, blockscale, batched variants, gmm), MoE (quant_moe, routing), gluon (pa_decode, gemm variants), fusions (fused_bmm_rope_kv_cache), and utility modules (types, gemm_config_utils, moe_config_utils). All of these call `get_arch()` to select tuned kernel configs or check hardware feature availability.

## Technical Details

The GPU architecture string is invariant for the lifetime of a process, so the `lru_cache`-based lazy evaluation in `get_arch()` can be replaced with eager evaluation at module import time.

**Before:**
```python
@lru_cache(maxsize=1)
def get_arch():
    try:
        arch = triton.runtime.driver.active.get_current_target().arch
    except RuntimeError:
        from jax._src.lib import gpu_triton as triton_kernel_call_lib
        arch = triton_kernel_call_lib.get_arch_details("0")
        arch = arch.split(":")[0]
    return arch
```

**After:**
```python
try:
    _CACHED_ARCH = triton.runtime.driver.active.get_current_target().arch
except RuntimeError:
    from jax._src.lib import gpu_triton as triton_kernel_call_lib
    _CACHED_ARCH = triton_kernel_call_lib.get_arch_details("0").split(":")[0]

def get_arch():
    return _CACHED_ARCH
```

When `torch.compile` traces `get_arch()`, it now sees a function returning a plain module-level string constant: no opaque runtime calls, no graph break. The public API (`get_arch()`, `is_gluon_avail()`, `is_fp4_avail()`, `is_fp8_avail()`) is unchanged.

## Test Plan

- Add `test_fav3_sage_compile.py` with two tests:
  - **`test_sage_compile_fullgraph`**: Compiles `fav3_sage_wrapper_func` with `torch.compile(fullgraph=True)` across multiple batch sizes, sequence lengths, head configurations, and layouts (bhsd/bshd). Asserts no graph break occurs and that compiled output matches eager within the same fp8 tolerances used by the existing sage correctness tests (atol=0.3, rtol=0.25, ≤0.5% outliers).
  - **`test_sage_compile_no_recompilation`**: Runs the compiled function 3 times with same-shaped inputs to verify Dynamo does not trigger unnecessary recompilation.
- Verify that existing `test_fav3_sage.py` tests continue to pass (no behavioral change to eager execution).
- Test impact on E2E performance on a diffusion model inference workload using FAv3 Sage attention

## Test Result
Tested on platform: gfx950.

- The fullgraph compilation tests in `test_fav3_sage_compile.py` fail on main but pass with this PR
- Existing Sage attention tests pass
- On a diffusion model workload using AITER Sage attention, E2E runtime improvement of 5.3% was observed. Current E2E performance: 24.40s, with this PR: 23.15s.

The image below compares side-by-side trace profiles from running with current version (left hand side) and with this PR (right hand side) indicates higher efficiency for `torch.compile` evidenced by larger uniform region compiled instead of many small regions.

image

## Submission Checklist


jcaraban (dismissed):

vgokhale:
We have been running this successfully for a year now so something seems off about this. Why does this cause a graph break? The arch target should not change during the run so the fix done here doesn't seem to really change anything. 

Also, even IF this fix was correct, the existing lru_cache decorator should handle it.

Do you know what causes the graph to break? Or more specifically which of these rules is getting violated

https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/cuda-graphs.html#device-graph-creation

lauri9:
# What causes the graph break?
The `lru_cache` decorator here provides runtime memoization of the architecture, but it still runs in eager Python once per invocation. The cached value is being reused at runtime, but Dynamo itself derives no benefit from the cache.

To verify this, consider the following sample code:
```
# test_lru_cache.py
import torch, functools, triton

@functools.lru_cache(maxsize=1)
def get_arch():
    return triton.runtime.driver.active.get_current_target().arch

# Warm the cache eagerly so any later "miss" is unambiguously Dynamo's doing:
_ = get_arch()
print("after warmup:", get_arch.cache_info())

def f(x):
    kpack = 2 if get_arch() == "gfx942" else 1
    return x * kpack

x = torch.randn(4, device="cuda")

torch._dynamo.reset()
fc = torch.compile(f, fullgraph=False)
_ = fc(x)                                 # compile + run
print("after 1st compiled call:", get_arch.cache_info())
_ = fc(x); _ = fc(x); _ = fc(x)
print("after 4th compiled call:", get_arch.cache_info())
```

Output:
```
after warmup: CacheInfo(hits=0, misses=1, maxsize=1, currsize=1)
after 1st compiled call: CacheInfo(hits=1, misses=1, maxsize=1, currsize=1)
after 4th compiled call: CacheInfo(hits=4, misses=1, maxsize=1, currsize=1)
```
Hits grows by exactly 1 per invocation of the compiled function, not per compile. Four compiled calls -> four hits.
- If Dynamo had constant-folded get_arch() at trace time, the compiled code would not call the wrapper at all -> hits would have stayed at 1, from a trace-time consultation, then frozen.
- If Dynamo had folded on the first invocation and baked the result into a guard for subsequent calls, you'd see hits = 1 after the 4th call, not hits = 4.

It seems to me the actual C-level `_lru_cache_wrapper.__call__` are happening, instead of Dynamo hoisting the constant value into the compiled graph.

The reason that `lru_cache` is not effective here is that `triton.runtime.driver.active.get_current_target()` calls `HIPDriver.get_current_target` which is explicitly in a file on the `SKIP_DIRS` list of function to "do not trace" (see "observation 1" below for more info).

# Recap of what the issue is and simple way to reproduce
I am following the [guidelines to use `fullgraph=True` to find and analyze graph breaks](https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/compile/programming_model.fullgraph_true.html).

Consider the following example script called `test_fav3_compile.py`:
```python
import argparse
import torch
from aiter.ops.triton.attention.fav3_sage_attention_mxfp4_wrapper import (
    fav3_sage_mxfp4_wrapper,
)

def f(q, k, v, w):
    # add a dummy matrix multiplication to demonstrate compiler graph 
    # behavior with other kernels
    q = q @ w
    out = fav3_sage_mxfp4_wrapper(q, k, v, causal=False, layout="bshd")
    return out


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--fullgraph", action=argparse.BooleanOptionalAction, default=False)
    args = parser.parse_args()

    fc = torch.compile(f, fullgraph=args.fullgraph)

    q = torch.randn(1, 1024, 40, 128, device="cuda")
    k = torch.randn(1, 1024, 40, 128, device="cuda")
    v = torch.randn(1, 1024, 40, 128, device="cuda")
    w = torch.randn(128, 128, device="cuda")

    trace_filename = "trace" + ("_fullgraph" if args.fullgraph else "") + ".json"
    with torch.profiler.profile(
        activities=[
            torch.profiler.ProfilerActivity.CPU,
            torch.profiler.ProfilerActivity.CUDA,
        ],
        schedule=torch.profiler.schedule(
            wait=2,
            warmup=2,
            active=1,
        )
    ) as p:
        for _ in range(2 + 2 + 1): # 2 warmup, 2 active, 1 wait step
            out = fc(q, k, v, w)
            p.step()
    p.export_chrome_trace(trace_filename)
    print(f"Exported trace to {trace_filename}")
```

## Observation 1: `fullgraph=True` triggers a hard crash on AITER main
Running `python test_fav3_compile.py` works, but `python test_fav3_compile.py --fullgraph` crashes with the following error:
```
  File "/app/diffusion/test_fav3_compile.py", line 40, in <module>
    out = fc(q, k, v, w)
          ^^^^^^^^^^^^^^
  File "/venv/lib/python3.12/site-packages/torch/_dynamo/eval_frame.py", line 841, in compile_wrapper
    raise e.with_traceback(None) from e.__cause__  # User compiler error
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch._dynamo.exc.Unsupported: Attempted to inline function marked as skipped
  Explanation: Dynamo developers have intentionally marked that the function `HIPDriver.get_current_target` should not be traced.
  Hint: Avoid calling the function `HIPDriver.get_current_target`.
  Hint: Apply `@torch._dynamo.dont_skip_tracing` to the function `HIPDriver.get_current_target` to force tracing into the function. More graph breaks may occur as a result of attempting to trace into the function.
  Hint: Please file an issue to PyTorch.

  Developer debug context: qualname: HIPDriver.get_current_target, name: get_current_target, filename: `/venv/lib/python3.12/site-packages/triton/backends/amd/driver.py`, skip reason: skipped according trace_rules.lookup SKIP_DIRS

 For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0008.html

from user code:
   File "/app/diffusion/test_fav3_compile.py", line 11, in f
    out = fav3_sage_mxfp4_wrapper(q, k, v, causal=False, layout="bshd")
  File "/app/external/aiter/aiter/ops/triton/attention/fav3_sage_attention_mxfp4_wrapper.py", line 130, in fav3_sage_mxfp4_wrapper
    return _FAv3SageMXFP4WrapperFunc.apply(
  File "/app/external/aiter/aiter/ops/triton/attention/fav3_sage_attention_mxfp4_wrapper.py", line 57, in forward
    config = get_sage_fwd_configs_mxfp4()
  File "/app/external/aiter/aiter/ops/triton/attention/fav3_sage_attention_mxfp4_wrapper.py", line 18, in get_sage_fwd_configs_mxfp4
    arch = arch_info.get_arch()
  File "/venv/lib/python3.12/site-packages/torch/_dynamo/polyfills/__init__.py", line 264, in getattr_and_trace
    return fn(*args[2:], **kwargs)
  File "/app/external/aiter/aiter/ops/triton/utils/_triton/arch_info.py", line 9, in get_arch
    triton.runtime.driver.active.get_current_target().arch

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

EDIT: I suppose this reveals the root cause why `lru_cache` is not effective here; `triton.runtime.driver.active.get_current_target()` calls `HIPDriver.get_current_target` which is explicitly in a file on the `SKIP_DIRS` list of "do not trace".

## Observation 2: operations adjacent to triton ops are in different compile graph parts
We can collect a profile running `python test_fav3_compile.py`. Attached for convenience here: [trace.json](https://github.com/user-attachments/files/26929368/trace.json)


Viewing the profile in a tool such as [Perfetto](ui.perfetto.dev) we see evidence of graph breaks: nested torch compile regions on CPU side, and split CompiledFxGraph on GPU side:

two_regions


## Observation 3: applying this PR resolves the issues
If we apply the fix in this PR, we can now run `python test_fav3_compile.py --fullgraph` successfully.

Also in the default compile mode, we can collect a profile running again `python test_fav3_compile.py`. Attached here: 
[trace_fixed.json](https://github.com/user-attachments/files/26929615/trace_fixed.json)

Viewing the trace visually, we see the issues above are resolved: now we have one torch compiled region both on CPU and GPU side.
single_region

lauri9:
By the way, when you say

> We have been running this successfully for a year now

do you mean you are successfully using `fullgraph=True` with `torch.compile`? The tests I added here for that were failing on AITER main, and I did not find any other tests for compile behavior. This is not a functional issue in the sense that using compile in default mode works, but it leaves performance on the table due to the silent breaks.

dllehr-amd:
@lauri9  Thanks for the PR.  I had a few additional questions to go along with Vinayak.  I work primarily in the vLLM space.  Can  you tell me what base problem led you to this?  was it training, inference, etc?  and which model?  

I think your fix has merit and would avoid breaks in some areas.  But for vLLM we have a pattern matching utility that takes a first stab at fusion patterns before letting inductor handle the rest.  So we're relatively protected to my knowledge.  But that doesn't mean something like this shouldn't be implemented for more traditional torch.compile fullgraph=True scenarios

jcaraban:
@dllehr-amd diffusion inference, more specifically for the overlap of comms and pre-attention quantization in a ulysses parallel setup. Do other areas besides vLLM employ similar pre-inductor fusions? That could explain why nobody has raised this issue before 🤔 

Anyway, this is a simple change with clear benefits and no visible drawbacks.

lauri9:
@dllehr-amd As mentioned by @jcaraban I noticed this on diffusion model inference (Wan 2.2-14B, but other models are affected as well). While vLLM-Omni can run some diffusion models, generally the framework space is still more fragmented with SGL-D, xDiT, DiffSynth, at least relevant. FWIW, this issue was discovered running through the xDiT inference engine.

The change here is functionally identical to the prior  `lru_cache` approach, with import-time initialization & cache reuse at runtime. But since it avoids the problematic `HIPDriver.get_current_target` call during tracing, it is also an improvement for the currently affected workloads that do not have pre-inductor fusion stages, with seemingly little chance of backfiring.

There might be other ways to check the architecture, but IMO this is the minimally invasive change, simply removing the `lru_cache` wrapper and not changing any of the implementation details nor interfaces.

ksikiric:
I would like to add a second concrete data point in favor of merging this. Independent of the fullgraph=True failure mode @lauri9 documented above, the same arch_info.get_arch() graph break also produces a hard crash under torch.compile(..., dynamic=True), even with the default fullgraph=False. 

The graph break causes the int literals returned by get_sage_fwd_configs_mxfp4() (notably num_warps and num_stages) to be tracked as SymNodeVariables rather than constants, which then crashes PyTorch's Triton-launch bridge in torch/_higher_order_ops/triton_kernel_wrap.py with `InternalTorchDynamoError: AttributeError: 'SymNodeVariable' object has no attribute 'value'.` 

So the PR fixes not just the fullgraph=True failure and the silent perf regressions, but also a third, harder-to-diagnose hard-crash mode that affects callers using dynamic=True on Triton ops with shape-dependent dispatch.

Happy to share the standalone repro and full Dynamo trace if useful.

vgokhale (approved):

PR #2742, opened 2026-04-14
Hoist introspection out of per-call ctype dispatch

gronsti-amd:
Hoist inspect.signature/typing.get_type_hints out of per-call ctype dispatch

These two introspection calls were recomputed on every invocation of the ctypes caller closure (~79µs + ~91µs per call). Since the decorated function's signature and type hints are immutable, compute them once at decoration time and capture via closure.

## Submission Checklist


amd-ruitang3:
@gronsti-amd That's a good idea~.

valarLip (approved):

PR #2762, opened 2026-04-16
feat(moe): support multi-B weight tensors (DWDP) in FlyDSL MOE kernels

AMD-yanfeiwang:
Add Distributed Weight Data Parallelism support to MOE GEMM kernels, allowing expert weights to be split across multiple tensors (up to 4). This enables DWDP where local and remote expert weights reside in separate buffers, eliminating the need for D2D copy before kernel launch.

Changes:
- moe_gemm_2stage.py (FP8): stage1/stage2 kernels accept arg_w_0..3 and arg_scale_w_0..3, create per-partition buffer resources, and use arith.select chains to dispatch correct (w_rsrc, expert_off_idx) at runtime based on expert index vs compile-time partition thresholds
- mixed_moe_gemm_2stage.py (FP4): same multi-B pattern for mixed-precision kernels, with nonlocal declarations for nested body functions
- moe_kernels.py: public API flydsl_moe_stage1/2 accepts Union[Tensor, List[Tensor]] for weights/scales, auto-computes b_tensor_l_offsets
- test_flydsl_moe_multi_b.py: unit tests for backward compat (single tensor) and 2/3/4-way splits, all 10/10 passing
- Fix pre-existing bug: i32 -> T.i32 in FP8 kernel epilogue

Design: num_b_tensors and b_tensor_l_offsets are compile-time params (part of lru_cache key). Preshuffle layout strides are independent of total expert count, so the same layout works for all partitions.

Note:
The current kernel has only been tested and passed on the gfx950 (MI355x). The kernel for other gfx1250 (MI450x) devices will be added as soon as possible.

## Submission Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,4 +79,4 @@ scripts/results/*.csv
 
 # Jupyter notebook checkpoints and outputs
 .ipynb_checkpoints/
-**/.ipynb_checkpoints/
+**/.ipynb_checkpoints/aiter/jit/flydsl_cache/
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -115,6 +115,193 @@ def get_inter_dim(w1_shape, w2_shape):
     return E, model_dim, inter_dim
 
 
+def _fused_moe_multi_b(
+    hidden_states,
+    w1_list,  # list of [E_i, inter_dim*2, dim] tensors
+    w2_list,  # list of [E_i, dim, inter_dim] tensors
+    topk_weight,
+    topk_ids,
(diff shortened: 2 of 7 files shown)


zufayu:
@AMD-yanfeiwang This PR has been inactive for over 2 months. Is the multi-B weight tensor (DWDP) FlyDSL MoE support still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2766, opened 2026-04-16
Fix fused_gemm_a8w8_blockscale_a16w16 unit test failures

nidal567:
## Motivation

This PR fixes numerical correctness failures in the `fused_gemm_a8w8_blockscale_a16w16` Triton kernel when running with `TRITON_HIP_USE_ASYNC_COPY` enabled (the default in the new triton compiler). The failures are known as assertion errors in pytest across a range of shapes, specifically around large M `(M=8192)` and non-power-of-2 K values `(K=7168)`. The root cause was a combination of LDS exhaustion from aggressive pipelining and a bias double-addition bug introduced during debugging.

### Fix - Kernel: Replace `tl.broadcast_to` with explicit zero-initialization + elementwise add
The issue came from initializing the bias using tl.broadcast_to, which creates a non-contiguous memory view. When ASYNC_COPY is enabled, the compiler can’t reliably treat that as a writable accumulator, which can lead to incorrect results. The fix is to first allocate the accumulator with tl.zeros, and then add the bias separately as a standard elementwise operation.

This change is applied to **both** the FP8 and BF16 accumulator paths in the kernel.

## Test Plan

Ran an isolated pytest for `test_fused_gemm_a8w8_blockscale_a16w16.py` covering:

- `M ∈ {1, 8, 32, 64, 128, 8192}`
- `N1 ∈ {256, 512}, N2 ∈ {256, 512}`
- `K ∈ {1024, 7168, 8192}`
- `dtype ∈ {float16, bfloat16}`
- `output ∈ {True, False}`
- `skip_reduce ∈ {True, False}`

Correctness was verified against PyTorch reference implementations using
`atol=0.1`, `rtol=0.1` (provided by the error prints).

## Test Result

All 584 test cases passed across all shape, dtypes, and config combinations, even at large values of `M` and non-power of 2 values of `K`.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a8w8_blockscale_a16w16.py
+++ b/aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a8w8_blockscale_a16w16.py
@@ -145,22 +145,11 @@ def _fused_gemm_a8w8_blockscale_a16w16_kernel(
                 + offs_bsn * stride_b_fp8_scale_n
             )
 
+            accumulator_fp8 = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype)
             if ADD_BIAS_FP8:
                 if NUM_KSPLIT == 1 or (SKIP_REDUCE and pid_k == 0):
-                    accumulator_fp8 = tl.load(bias_fp8_ptr + offs_b_fp8_n).to(
-                        dtype=acc_dtype
-                    )
-                    accumulator_fp8 = tl.broadcast_to(
-                        accumulator_fp8[None, :], (BLOCK_SIZE_M, BLOCK_SIZE_N)
-                    )
-                else:
-                    accumulator_fp8 = tl.zeros(
-                        (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                    )
-            else:
-                accumulator_fp8 = tl.zeros(
-                    (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                )
+                    bias_fp8_vals = tl.load(bias_fp8_ptr + offs_b_fp8_n).to(dtype=acc_dtype)
+                    accumulator_fp8 += bias_fp8_vals[None, :]
 
             for k in range(pid_k * num_k_iter, (pid_k + 1) * num_k_iter):
                 if EVEN_K:
@@ -222,22 +211,11 @@ def _fused_gemm_a8w8_blockscale_a16w16_kernel(
                 + offs_b_bf16_n[None, :] * stride_b_bf16_n
             )
 
+            accumulator_bf16 = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype)
             if ADD_BIAS_BF16:
                 if NUM_KSPLIT == 1 or (SKIP_REDUCE and pid_k == 0):
-                    accumulator_bf16 = tl.load(bias_bf16_ptr + offs_b_bf16_n).to(
-                        dtype=acc_dtype
-                    )
-                    accumulator_bf16 = tl.broadcast_to(
-                        accumulator_bf16[None, :], (BLOCK_SIZE_M, BLOCK_SIZE_N)
-                    )
-                else:
-                    accumulator_bf16 = tl.zeros(
-                        (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                    )
-            else:
-                accumulator_bf16 = tl.zeros(
-                    (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                )
+                    bias_bf16_vals = tl.load(bias_bf16_ptr + offs_b_bf16_n).to(dtype=acc_dtype)
+                    accumulator_bf16 = accumulator_bf16 + bias_bf16_vals[None, :]
 
             for k in range(pid_k * num_k_iter, (pid_k + 1) * num_k_iter):
                 if EVEN_K:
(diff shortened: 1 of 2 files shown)


aiter/ops/triton/gemm/fused/fused_gemm_a8w8_blockscale_a16w16.py
```diff
@@ -114,6 +114,10 @@ def fused_gemm_a8w8_blockscale_a16w16(
     config["GROUP_K"] = triton.next_power_of_2(triton.cdiv(K, w_fp8_scale.shape[0]))
     config["GROUP_N"] = triton.next_power_of_2(triton.cdiv(N_fp8, w_fp8_scale.shape[1]))
 
+    if M >= 4096 and config.get("num_stages", 1) > 1:
```
brunomazzottiamd: 1. Why not editing `aiter/ops/triton/configs/gemm/gfx950-FUSED-GEMM-A8W8_BLOCKSCALE-A16W16.json` and `aiter/ops/triton/configs/gemm/gfx950-FUSED-GEMM-A8W8_BLOCKSCALE-A16W16-N8=512-N16=256-K=7168.json` instead of doing it through code?
2. Content from `aiter/ops/triton/configs/gemm/gfx942-FUSED-GEMM-A8W8_BLOCKSCALE-A16W16.json` is overwritten, this isn't correct.

I'd prefer a way of doing this that's 100% based on config files.

nidal567: Got it, I see where I went wrong. I'll remove the guard and instead set `num_stages = 1` directly in the gfx950 config file (both `aiter/ops/triton/configs/gemm/gfx950-FUSED-GEMM-A8W8_BLOCKSCALE-A16W16.json and aiter/ops/triton/configs/gemm/gfx950-FUSED-GEMM-A8W8_BLOCKSCALE-A16W16-N8=512-N16=256-K=7168.json`, as I see where the overwriting is happening with regards to gfx942. Thank you Bruno, will implement this fix!

nidal567: Interestingly enough, Fix no. 1 is a standalone fix for this PR. I ran the test cases, it seems to work without needing to change the config files. I will keep the config files as is, and when we can investigate any differences in setting `num_stages` when a benchmark is developed!

brunomazzottiamd: > Interestingly enough, Fix no. 1 is a standalone fix for this PR.

This is awesome.

aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a8w8_blockscale_a16w16.py
```diff
@@ -222,22 +213,13 @@ def _fused_gemm_a8w8_blockscale_a16w16_kernel(
...
-                    )
-                else:
-                    accumulator_bf16 = tl.zeros(
-                        (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                    )
-            else:
-                accumulator_bf16 = tl.zeros(
-                    (BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=acc_dtype
-                )
+                    accumulator_bf16 = accumulator_bf16 + bias_bf16_vals[None, :]
```
brunomazzottiamd, line 222: **Nitpick:** What do you think of `accumulator_bf16 += bias_bf16_vals[None, :]` as you did with `accumulator_fp8`?

brunomazzottiamd (approved):
@nidal567, it LGTM! Let's wait for CI to run, if it goes green then we can merge it.

brunomazzottiamd:
The following failures on **Triton Tests (MI35X) / Shard 0** are expected ones:

```text
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-512-512-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-512-512-4]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-1024-1024-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-1024-1024-4]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-2048-2048-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-32-2048-2048-4]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-512-512-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-512-512-4]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-1024-1024-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-1024-1024-4]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-2048-2048-1]
op_tests/triton_tests/attention/test_mha.py::test_mha_backward[True-False-0.0-True-128-8-64-2048-2048-4]
```

It isn't a blocker for merging.

aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a8w8_blockscale_a16w16.py
nidal567: Good catch! This is cleaner and consistent, I'll work on that

brunomazzottiamd: This code aesthetics change will be addressed in a future PR.

PR #2775, opened 2026-04-17
Xiaobing/gptoss small m

XiaobingSuper:
## Motivation

This PR adds a dedicated FlyDSL small-M HGEMM path for decode-like BF16 GEMMs where `M` is tiny while `N/K` remain large.

## Technical Details

- Add a dedicated FlyDSL `small_m` HGEMM family for decode-like BF16 GEMMs with `1 <= M < 17`, instead of treating them as just another tuning point of generic `hgemm`.
- Collapse M-side parallelism (`TILE_M=16`, `BLOCK_M_WARPS=1`) so wave budget is spent on the heavier N/K directions rather than the tiny M dimension.
- Change warp mapping to an N-centric layout (`warp_m_idx=0`, `warp_n_idx=wid * WARP_N`), so one block works on one small-M slice with more N-side workers.
- Add `N_TILE_REPEAT` to reuse loaded A fragments across multiple N tiles in the non-`B_TO_LDS` path, improving arithmetic intensity for small-M / wide-N shapes.
- Add `PERSISTENT_N_TILES` plus dedicated `B_TO_LDS_UNROLL` / `waves_per_eu` tuning knobs for the `B_TO_LDS` path, so wide-N scheduling and load/compute overlap are tuned specifically for this regime.

## Test Plan

Accuracy and performance

## Test Result
image

image

## Submission Checklist


aiter/aot/flydsl/gemm.py
```diff
@@ -259,27 +231,38 @@ def _compile_hgemm_to_cache(
...
+        waves_per_eu=waves_per_eu,
+        b_to_lds_unroll=b_to_lds_unroll,
+        async_copy=async_copy,
+        b_to_lds=b_to_lds,
+        b_preshuffle=b_preshuffle,
+        c_to_lds=c_to_lds,
+        has_bias=has_bias,
+    )
+    _compile_executable_to_cache(
+        exe, out, a, b, bias if has_bias else None, m, counter, 0, stream
```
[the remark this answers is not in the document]
coderfeli: @XiaobingSuper  this looks reasonable

xytpai (dismissed):

aiter/aot/flydsl/gemm.py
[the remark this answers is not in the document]
coderfeli: Confirmed. FlyDSL JIT does not accept `None` for tensor slots (`NoneType` is not a registered JIT arg), so this can break no-bias AOT jobs. Runtime already handles this by passing a dummy tensor (`b`) when bias is absent.

aiter/ops/flydsl/kernels/small_m_hgemm.py
```diff
@@ -0,0 +1,1381 @@
...
+    192,
+    224,
+    256,
+    384,
+    512,
+    768,
+    1024,
+)
+SMALL_M_NON_B_TO_LDS_WAVES_PER_EU_OPTIONS = (0, 2, 4)
+SMALL_M_B_TO_LDS_WAVES_PER_EU_OPTIONS = (0, 2, 4)
```
coderfeli, line 100: Minor catalog issue: `0` is treated as a separate tuning value here, but the compiler later normalizes wide-N `B_TO_LDS` `waves_per_eu=0` to `2`, and `b_to_lds_unroll=0` to `8`. That means some distinct kernel names/candidates collapse to the same generated kernel.

aiter/ops/flydsl/kernels/small_m_hgemm.py
```diff
@@ -0,0 +1,1381 @@
...
+        c_zero_d = arith.constant(0.0, type=dtype_)
+        acc_init = arith.constant_vector(0.0, T.vec(WMMA_C_FRAG_VALUES, T.f32))
+        zero_a_vec = vector.broadcast(T.vec(LDG_VEC_SIZE, dtype_), c_zero_d)
+        zero_a_async_vec = vector.broadcast(T.vec(LDG_ASYNC_VEC_SIZE, dtype_), c_zero_d)
+
+        A_ = GTensor(A, dtype=dtype_, shape=(-1, k))
+        B_ = GTensor(B, dtype=dtype_, shape=(n, k))
+        C_ = GTensor(C, dtype=dtype_, shape=(-1, n))
+        if HAS_BIAS:
+            BIAS_ = GTensor(BIAS, dtype=dtype_, shape=(n,))
```
xudoyuan: Flydsl does not support this type of variable definition within an if statement. If the variable needs to be used outside the if statement (e.g., on line 625), it needs to be initialized outside the if statement. Please add initialization functionality for rectification. 

Other areas requiring modification: 
L552-559: bs_, smem_b_ptr
L564-565: COUNTER_

Thanks~

aiter/ops/flydsl/kernels/splitk_hgemm.py
```diff
@@ -230,6 +234,8 @@ def hgemm_kernel(
         A_ = GTensor(A, dtype=dtype_, shape=(-1, k))
         B_ = GTensor(B, dtype=dtype_, shape=(n, k))
         C_ = GTensor(C, dtype=dtype_, shape=(-1, n))
+        if HAS_BIAS:
```
xudoyuan: Outer variable undefined.

Other areas requiring modification:
L244-248: bs_、smem_b_ptr
L244-248: SHUFFLED_B_
L265-266: COUNTER_
L799: a_regs_next

aiter/ops/flydsl/kernels/small_m_hgemm.py
```diff
@@ -0,0 +1,1381 @@
...
+        def zero_c_tile(tile_n_offset):
+            zero_vec = vector.broadcast(T.vec(LDG_VEC_SIZE, dtype_), c_zero_d)
+            for i in range_constexpr(LDG_REG_C_COUNT):
+                global_tid = BLOCK_THREADS * i + tid
+                m_local_idx = global_tid // LDG_C_X_THREADS
+                n_local_idx = global_tid % LDG_C_X_THREADS * LDG_VEC_SIZE
+                row_idx = m_offset + fx.Index(m_local_idx)
+                init_vec = zero_vec
+                if HAS_BIAS:
+                    init_vec = BIAS_.vec_load(
```
xudoyuan: If a function definition inside the kernel needs to modify a variable defined outside the kernel, it(BIAS_) must use the variable as a parameter in the function call.

Other areas requiring modification:
L688: COUNTER_
Etc.

aiter/aot/flydsl/gemm.py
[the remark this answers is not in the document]
XiaobingSuper: Updated.

aiter/ops/flydsl/kernels/small_m_hgemm.py
XiaobingSuper: Updated. The compiler only normalizes waves_per_eu=0 to 2 for the wide-N B_TO_LDS subset. For non-wide-N B_TO_LDS configs, 0 is still a real tuning candidate. We already have tuned entries that rely on this. For example, in gptoss_bf16_tuned_gemm.csv, the M=8, N=2880, K=4096 small-M kernel has no _WPE suffix in the kernel name, which corresponds to waves_per_eu=0, and this path also passes runtime validation.

So the safer fix is not to remove 0 globally, but to:

keep SMALL_M_B_TO_LDS_WAVES_PER_EU_OPTIONS = (0, 2, 4)
canonicalize / deduplicate only the wide-N B_TO_LDS cases that collapse to the same effective compiled kernel
I updated it that way, and also verified both the registry dedup behavior and a runtime smoke case.

xudoyuan (approved):
LGTM

coderfeli (approved):

PR #2779, opened 2026-04-17
[Don't merge] Gluon pa bad case reproducer

ganyi1996ppo:
## Motivation
use 
```
pytest op_tests/triton_tests/test_pa_decode_gluon.py -k "gluon_vs_hip_performance"
```
 to reproduce the bad case of gluon shuffle layout pa. non-persistent mode is much better, but still have some case worse than hip version

## Submission Checklist


zufayu:
Closing this reproducer PR — it was explicitly marked [Don't merge] and served its purpose as a bad-case reproducer. The related paged-attention / pa_mqa fixes have since landed (#4440 block sizes > 1, #4825 pa_mqa logits block size fix, #4680 row/block addressing fix). If the bad case still reproduces on current main, please open a fresh issue with the reproducer details.

PR #2783, opened 2026-04-17
Gluon gemma8w8 blockscale wrap-up

amirumoAMD:
## Motivation

Wrapped-up GEMM a8w8 blockscale kernel for gfx12

## Technical Details

Buffer loaded scales, TDM all matrices. Single pass block scale.

## Test Plan

FFM & AM

## Test Result

Passed testing, closed with torch

## Submission Checklist


aiter/ops/triton/_gluon_kernels/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,348 @@
...
+        offsets=offs_a_scale,
+        cache=cache_modifier,
+    )
+    b_scale = gl.amd.cdna4.buffer_load(
+        ptr=b_scale_ptr,
+        offsets=offs_b_scale,
+        cache=cache_modifier,
+    )
+    # prefetch lds
+    if L2_PREFETCH_DISTANCE > NUM_BUFFERS:
```
vgokhale: Can we remove everything related to prefetch? I am not sure if it helps. Typically my experience has been to not optimize aggressively. That doesn't mean just write non-performant code and get UTs passing of course. But I try to bucket optimizations into ones that have obvious benefits and ones that do not. I am not sure why prefetch will help here. IT is certainly an optimization, but one that should be done based on information gained from ATT or other tools.

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,191 @@
...
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+from typing import Optional
+import functools
+import json
+import os
+import torch
+import triton
+import math
+from aiter.ops.triton.utils._triton.pid_preprocessing import pid_grid, remap_xcd
```
vgokhale: Seems unused (and probably okay for 1250) so remove?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,191 @@
...
+from aiter.ops.triton._gluon_kernels.gemm_a8w8_blockscale import *
+
+_LOGGER = AiterTritonLogger()
+from triton.experimental import gluon
+from triton.experimental.gluon import language as gl
+
+
+@triton.heuristics(
+    {
+        "EVEN_K": lambda args: args["K"] % args["BLOCK_SIZE_K"] == 0,
```
vgokhale: TDM means we don't need to detect this so remove?

aiter/ops/triton/_gluon_kernels/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,348 @@
...
+    NUM_BUFFERS: gl.constexpr,
+    L2_PREFETCH_DISTANCE: gl.constexpr,
+):
+    """
+    Note: this is Triton jited function and not meant to be called directly. Call gemm_a8w8_blockscale function
+    below
+
+    Computes the 8 bit matmul C = A x B using the block-scale quantization approach, with block shape assumed to be the same as BLOCK_SIZE_N/K. 
+
+    Split-K not supported due to design decision.
```
vgokhale, line 67: So I see a lot of code to handle splitk. But I don't see a reduce kernel. Did you eventually want to add splitk support?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,191 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+from typing import Optional
+import functools
```
vgokhale: What are all these imports used for?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,191 @@
...
+
+    _LOGGER.warning("enters here")
+
+    M, K = x.shape
+    N, K = w.shape
+
+    # Check constraints.
+    assert x.shape[1] == w.shape[1], "Incompatible dimensions!!!"
+
+    # Transpose w and w_scale
```
vgokhale: Can you add a bit more on why?

aiter/ops/triton/gemm_a8w8_blockscale.py
```diff
@@ -0,0 +1,191 @@
...
+            * triton.cdiv(M, META["BLOCK_SIZE_M"])
+            * triton.cdiv(N, META["BLOCK_SIZE_N"])
+        ),
+    )
+    NUM_WARPS=config["num_warps"]
+    warp_bases = [(0, 1)]
+    for i in range(int(math.log2(NUM_WARPS // 2))):
+        warp_bases.append((1 << i, 0))
+    warp_bases = tuple(warp_bases)
+    print("this is being called")
```
vgokhale: Remove?

op_tests/triton_tests/gemm/basic/test_gemm_a8w8_blockscale_gfx12.py
```diff
@@ -0,0 +1,231 @@
...
+        weight = (
+            (torch.rand((K, N), dtype=torch.float16, device="cuda") / 10)
+            .to(e4m3_type)
+            .T
+        )
+
+    x_scale = torch.rand([M, scale_k], dtype=torch.float32, device="cuda")
+    w_scale = torch.rand([scale_n, scale_k], dtype=torch.float32, device="cuda")
+
+    # Note: for now. weight shuffle was buggy, removed for wrapup
```
vgokhale: Remove comment and code associated with preshuffling.

op_tests/triton_tests/gemm/basic/test_gemm_a8w8_blockscale_gfx12.py
```diff
@@ -0,0 +1,231 @@
+# SPDX-License-Identifier: MIT
```
vgokhale: Can we reuse the existing test?

aiter/ops/triton/_gluon_kernels/gemm_a8w8_blockscale.py
amirumoAMD: I was originally trying to, but decided shortly before wrap-up that it wouldn't help with TDM. Cleaned up remnants, thanks

zufayu:
@amirumoAMD This PR has been inactive for over 2 months. Is the gluon gemm a8w8 blockscale wrap-up still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

zufayu:

**Adds a gfx1250 Gluon kernel for block-scale a8w8 GEMM — TDM descriptor loads through a 3-buffer pipeline with fp32 scales folded in after each wmma — behind a new arch-gated Python wrapper wired into the existing test and benchmark.**

Validation (deterministic): NOT RUN — REVIEW_AUTO_VALIDATE=0 and no gfx1250 GPU on this box (gfx942 only); the triaged target op_tests/triton_tests/gemm/basic/test_gemm_a8w8_blockscale.py exists — environment gap, not a PR defect
Perf (advisory): NOT RUN — the triaged target exposes no benchmark entry point, the PR's bench harness targets gfx1250 which this gfx942 box cannot run, and the PR ships no timing data of its own

⚠️ The patch no longer applies to main, and the op it adds already landed through #3816 in a different layout: main now ships aiter/ops/triton/_gluon_kernels/gfx1250/gemm/basic/gemm_a8w8_blockscale.py (split-K plus bandwidth/compute-bound variants) dispatched from the unified aiter/ops/triton/gemm/basic/gemm_a8w8_blockscale.py wrapper, while this PR's separate aiter/ops/triton/gemm_a8w8_blockscale_gfx1250.py imports aiter.ops.triton.utils.core, which main renamed to config_utils.py — so on current main the new wrapper and both op_tests files that import it fail at import, and any green CI on the PR describes a tree that has moved. **Author must** rebase onto current main and either fold this kernel into the landed _gluon_kernels/gfx1250/ structure or close the PR in favor of #3816. [verified]

⚠️ op_tests/triton_tests/gemm/basic/test_gemm_a8w8_blockscale.py ends with a module-level parser.parse_args() plus four direct test_gemm() calls and no `if __name__ == "__main__"` guard — under the CI's `pytest -v op_tests/triton_tests/` invocation (triton-test.yaml) argparse sees pytest's own argv and exits with "unrecognized arguments" (reproduced here: SystemExit(2)), so the file fails collection, and op_tests/op_benchmarks/triton/bench_gemm_a8w8_blockscale.py imports this module, so every benchmark run also dies at import (or, invoked bare, silently runs four tests before benchmarking). **Author must** move the block under a __main__ guard or switch to parse_known_args. [verified]

⚠️ In aiter/ops/triton/_gluon_kernels/gemm_a8w8_blockscale.py every scale load is an unmasked cdna4.buffer_load, and when K <= BLOCK_SIZE_K (k_tiles=1, e.g. the PR's own (16,16,128) shape) the epilogue's a_scale_ptr += stride_ascale_k advances one block past a scale tensor with scale_k=1, reading past x_scale/w_scale; the garbage multiplies a zero-filled tile, so the output corrupts only if the out-of-bounds float decodes as NaN (0×NaN=NaN poisons the whole tile), and ragged M=159 with BLOCK_SIZE_M=64 reads 32 rows past x_scale outright. **Author must** mask the scale loads or assert k_tiles >= 2 and M % BLOCK_SIZE_M == 0. [verified]

⚠️ A new kernel and a new tuned config ship with zero timing data — no numbers in the description, no benchmark output attached, and no base-vs-head run exists (auto-validation disabled here, no gfx1250 GPU on this box), while aiter/ops/triton/configs/gemm/gluon/gfx1250-GEMM-A8W8_BLOCKSCALE.json gives M_LEQ_64 and M_LEQ_128 16×32 tiles identical to M_LEQ_16 and NUM_BUFFERS=3 is hardcoded against num_stages=2, none of it measured. **Author must** attach benchmark numbers for aiter/ops/triton/_gluon_kernels/gemm_a8w8_blockscale.py against the triton path (and against main's landed gluon kernel after rebase). [verified]

📝 The arch guard in aiter/ops/triton/gemm_a8w8_blockscale_gfx1250.py (int(dev.split("gfx")[1]) < 950) raises "requires CDNA4" but numerically admits RDNA3 gfx1100/gfx1150, which have no gfx1250 TDM ops — a direct caller on those archs passes the guard and then fails inside Gluon codegen instead of seeing the intended error. **Reviewer should ask** for an explicit gfx1250 allowlist (or a real CDNA-family check) when the PR is rebased. [verified]

PR #2784, opened 2026-04-17
Fix LDS Exhaustion in `fused_gemm_afp4wfp4_a16w16` for ASYNC_COPY Enabled

nidal567:
## Motivation

This PR cleans a line in a previous code for consistency and fixes `OutOfResources: out of resource: shared memory` failures in the `fused_gemm_afp4wfp4_a16w16` Triton kernel when running with `TRITON_HIP_USE_ASYNC_COPY` enabled (which is the default in the current Triton compiler and has caused lots of breakage ([see here](https://github.com/ROCm/triton-internal/issues/1772)). The failures occur for **specific shape combinations**: `N1=512, N2=256, K=7168` with `M ∈ {64, 8192}` and **non-preshuffled** kernel path (`fp4_shuffle=False`). 

Experimentally, reducing `BLOCK_SIZE_M` in the `M_LEQ_64` and `any` config buckets is an alternative path to the same fix, but reducing `num_stages` is the cleaner and less risky change since it does not alter the tile geometry or affect the config files how they were desired to be tuned by the original implementation.

## Technical Details

Reduce num_stages from 2 to 1 in the two affected config buckets in:
`aiter/ops/triton/configs/gemm/gfx950-FUSED-GEMM-AFP4WFP4-A16W16-N4=512-N16=256-K=7168.json`

Specifically (key values):
`M_LEQ_64: num_stages: 2 -> 1`
`any: num_stages: 2 -> 1`

With `num_stages = 1`, there is no double-buffering - only one copy of each tile buffer lives in LDS at a time, halving the staging buffer LDS requirement and bringing total LDS usage within the hardware limit.

#### Why Only `N4=512-N16=256-K=7168` Config file?
The config system loads a specific config file when one exists for the exact (N4, N16, K) combination, and falls back to the general config otherwise. The general config `gfx950-FUSED-GEMM-AFP4WFP4-A16W16.json` uses `BLOCK_SIZE_M=32`, `BLOCK_SIZE_K=256`, which are small enough tiles that they don't overflow LDS even with ASYNC_COPY on. All other (N4, N16, K) combinations either use the general config with these smaller safe tiles, or have their own specialized configs that happen to have safe tile sizes. Only this specific `N4=512, N16=256, K=7168` specialized file had tile configurations large enough to overflow.

The LDS overflow only occurs when the tile footprint `BLOCK_SIZE_M × BLOCK_SIZE_K` is large enough that doubling it via `num_stages=2` exceeds the hardware limit. Across all buckets in the file:

| Bucket       | BLOCK_SIZE_M | BLOCK_SIZE_K | M × K product | Safe with num_stages=2? |
|--------------|--------------|--------------|---------------|--------------------------|
| M_LEQ_8      | 8            | 256          | 2,048         | ✅ Yes                   |
| M_LEQ_16     | 8            | 512          | 4,096         | ✅ Yes                   |
| M_LEQ_32     | 32           | 512          | 16,384        | ✅ Yes                   |
| M_LEQ_64     | 64           | 512          | 32,768        | ❌ No -> fixed            |
| M_LEQ_128    | 16           | 256          | 4,096         | ✅ Yes                   |
| M_LEQ_256    | 16           | 256          | 4,096         | ✅ Yes                   |
| any          | 128          | 256          | 32,768        | ❌ No -> fixed            |

`M_LEQ_64` and `any` both have a tile footprint of 32,768 — the threshold at which num_stages=2 + ASYNC_COPY pushes LDS usage over the hardware limit. Also, `M_LEQ_128` and `M_LEQ_256` actually use smaller tile sizes (specifically `BLOCK_SIZE_M`) than `M_LEQ_64`, which kept them safe from this issue. Only the two largest tile configurations needed fixing.

## Test Plan

Ran the full pytest suite for test_fused_gemm_afp4wfp4_a16w16.py covering [see here for test](https://github.com/ROCm/aiter/blob/1bcab14c8da65ead63e42fd54e4a20b2ed23615c/op_tests/triton_tests/gemm/fused/test_fused_gemm_afp4wfp4_a16w16.py#L73):

- `M ∈ {1, 8, 32, 64, 128, 8192}`
- `N1 ∈ {256, 512}` 
- `N2 ∈ {256, 512}`
- `K ∈ {1024, 7168, 8192}`
- `dtype ∈ {float16, bfloat16}`
- `output ∈ {True, False}`
- `skip_reduce ∈ {True, False}`
- `fp4_shuffle ∈ {True, False}`

Correctness verified against PyTorch reference implementations using `atol=0.1, rtol=0.1` (provided by the error prints).

## Test Result

All 1152 test cases passed across all shape, dtypes, and config combinations, even at large values of M and non-power of 2 values of K

## Submission Checklist


brunomazzottiamd (approved):
LGTM!

brunomazzottiamd:
Failure on **Triton Tests (MI35X) / Shard 0** is already fixed by https://github.com/ROCm/aiter/pull/2612.

Failure on **Triton Tests (MI325) / Shard 3**, coming from `op_tests/triton_tests/attention/test_pa_decode.py::test_paged_attn` is intermittent. I've triggered a rerun.

PR #2790, opened 2026-04-19
fix(pa_mqa_logits): handle ChunkQ > heads-per-GPU for high tensor parallelism (TP>=8)

jatseng-ai:
## Problem

When serving models with many attention heads at high tensor parallelism (TP>=8), each GPU holds fewer heads than the hardcoded `ChunkQ` values, causing two crashes in `deepgemm_fp8_paged_mqa_logits_stage1_ragged_k` and `deepgemm_fp8_paged_mqa_logits_stage1`:

**Example**: GLM-5.1-FP8 (128 heads) with TP=8 → 16 heads/GPU

1. `ZeroDivisionError` at `TileQCount = batch_size * next_n * (heads // ChunkQ)`  
   → `16 // 64 = 0`, then `TotalCuCount // 0` crashes

2. `AssertionError` at `assert heads % config["ChunkQ"] == 0`  
   → `16 % 64 = 16 ≠ 0` (similarly `16 % 32 ≠ 0` in the ragged_k variant)

## Fix

- Guard `TileQCount` with `max(1, ...)` to prevent the `ZeroDivisionError`
- Before each assertion, halve `ChunkQ` until it evenly divides `heads`  
  (gracefully falls back; ChunkQ=1 always divides any positive integer)

```python
# Example: heads=16, starting ChunkQ=64
# 16 % 64 ≠ 0 → halve → 32
# 16 % 32 ≠ 0 → halve → 16
# 16 % 16 = 0 ✓  →  use ChunkQ=16
```

## Validation

Tested on AMD MI300X (gfx942) with **GLM-5.1-FP8** served via vLLM at **TP=8**:
- Server starts successfully (previously crashed at first inference)
- 4/4 multi-turn tool-call conversations completed correctly
- KV cache: 76.47 GiB available as expected
- Generation throughput: ~9 tok/s (enforce-eager mode)


jatseng-ai:
Ready for review.

Rebased onto latest `main` (conflict resolved in `pa_mqa_logits.py`; ChunkQ adjust loops + `TileQCount = max(1, …)` guard preserved).

The high-TP ChunkQ bug is still present on `main`: with heads-per-GPU < hardcoded/default ChunkQ (e.g. 128 heads @ TP=8 → 16 heads/GPU vs ChunkQ 32/64), stage1 hits `ZeroDivisionError` / `assert heads % ChunkQ == 0`. This PR halves ChunkQ to the largest power-of-2 divisor of `heads` before those asserts.

CI re-triggered on the rebase; prior shard failures after an old merge commit looked infra-ish (exit 134 / MI35X shard) while the original fix commit was green — will watch the new run.

PR #2794, opened 2026-04-20
[CK] Add StreamLLM sink token support for batch_prefill pipeline

LJ-underdog:
## Dependencies

  This PR depends on **ROCm/rocm-libraries#6479** (CK-side change that promotes `kHasSink_` to a real template parameter in `fmha_fwd_batch_prefill_traits_`, enabling `_sink` / `_nsink` kernel
  variant codegen for the batch_prefill pipeline).

  - Must be merged **after** ROCm/rocm-libraries#6479.
  - The `3rdparty/composable_kernel` submodule in this PR is bumped to commit `feea6be`, which already contains #6479.

  ---

  ## Motivation

  The existing `mha_batch_prefill` pipeline did not support StreamingLLM-style sink tokens: `kHasSink_` was hard-coded to `false` in the CK traits, and `sink_size` was not threaded through the
  AITER → C++ → CK call chain. This blocked workloads (e.g. GPT-OSS-style attention) that require a fixed prefix of always-attended KV tokens together with a sliding window and/or a virtual
  sink-logit per head.

  This PR adds end-to-end sink token support for `mha_batch_prefill` on top of the CK changes in #6479.

  ## Technical Details

  ### CK submodule
  - Bump `3rdparty/composable_kernel` to `feea6be` (includes ROCm/rocm-libraries#6479), which makes `kHasSink_` a real template parameter and emits separate `_sink` / `_nsink` kernel
  instantiations for batch_prefill.

  ### Python (`aiter/ops/mha.py`)
  - Add `sink_size: int = 0` to `mha_batch_prefill_func`, `_mha_batch_prefill`, the `mha_batch_prefill` stub, and `cmdGenFunc_mha_batch_prefill`.
  - Select the `_sink` / `_nsink` kernel module name by `has_effective_sink = sink_size > 0 and (causal or has_window_mask)`, matching the C++ mask logic so we don't compile an unused sink
  variant when the mask cannot expose it.

  ### C++ interface (`csrc/`)
  - `mha_batch_prefill_traits`: expose `has_sink` (was hard-coded `false`).
  - `mha_fwd_batch_prefill.cu`: derive `has_sink = args.sink_size > 0`.
  - `mha_batch_prefill_kernels.cu`:
    - Add `int sink_size` parameter; include it in the `mask_identify` string used for kernel-variant selection.
    - Set `args.sink_size = mask.sink`.
    - Zero-initialize `fmha_batch_prefill_args{}` to avoid UB on fields not explicitly written.
  - PyBind (`csrc/include/rocm_ops.hpp`) and the public declaration (`csrc/include/torch/mha_batch_prefill.h`): add `sink_size` as a positional parameter immediately after `window_size_right`.

  ### Semantics (clarified and documented)
  - `sink_size`: number of leading KV tokens that are always attended (drives the sink-phase loop in the kernel).
  - `sink_ptr[nhead]`: a fixed logit injected for a single virtual sink token in softmax (`null` ⇒ `-inf`, non-null ⇒ user-provided value); independent of `sink_size`.
  - `window_size_left = L` means k ∈ `[abs_q - L, abs_q]` (i.e. `L + 1` tokens). Confirmed via code derivation (`block_masking.hpp`) and discriminating tests.

  ## Test Plan

  `op_tests/test_batch_prefill.py` is extended with StreamingLLM-style sink configurations covering:
  - No-sink baseline (regression check, `sink_size = 0`, `sink_ptr = None`).
  - `sink_ptr`-only path (virtual sink token in softmax, no sink-phase tokens).
  - Combined `window_size_left = 1024` + `sink_size = 4` + `sink_ptr`.

  Each case is cross-checked against a torch reference implementation.

  ## Test Result

  | Configuration | max_diff vs torch ref |
  |---|---|
  | No-sink paths (regression) | < 0.004 |
  | `sink_ptr` virtual token only | < 0.004 |
  | `window=1024` + `sink_size=4` + `sink_ptr` | 0.000488 |

  All numerically within the expected fp16/bf16 tolerance.

  ## Submission Checklist


LJ-underdog:
Aiter Test build failure is an upstream main issue (FlyDSL GEMM AOT requires GPU at build time, but docker build env has none). Same failure reproduces
  on main head e.g. run 24758662572. Will resolve once main is fixed.

valarLip (approved):

PR #2839, opened 2026-04-21
fix(build): add missing c10/hip/HIPException.h include in gated_rmsnorm_quant_kernels.cu

ChuanLi1101:
﻿## Summary

Adds the missing `#include <c10/hip/HIPException.h>` to `csrc/kernels/gated_rmsnorm_quant_kernels.cu` so that wheel builds with `PREBUILD_KERNELS=1` succeed.

Fixes #2837.

## Root cause

`gated_rmsnorm_quant_kernels.cu` line 223 uses `C10_HIP_KERNEL_LAUNCH_CHECK()` but the file never includes the header that defines it. The macro lives in `c10/hip/HIPException.h` (analogue of `C10_CUDA_KERNEL_LAUNCH_CHECK` in `c10/cuda/CUDAException.h`).

- **JIT path**: the AITER hipify pipeline under `aiter/jit/utils/hipify/` injects these c10 headers implicitly, so the missing include was invisible in per-PR CI.
- **`PREBUILD_KERNELS=1` path**: compiles the `.cu` directly via `setup.py` extensions, without that injection, and fails with

```
csrc/kernels/gated_rmsnorm_quant_kernels.cu:223:5:
  error: use of undeclared identifier 'C10_HIP_KERNEL_LAUNCH_CHECK'
```

The regression was introduced in 875f6b7cd (#2499, "Fuse gated rmsnorm + group_quant for qwen3next and qwen3.5") and is present in v0.1.12.post1.

## Fix

One-line change: add `#include <c10/hip/HIPException.h>` to the file's include block.

## Test plan

- [ ] Local: `PREBUILD_KERNELS=1 pip install .` no longer errors on `gated_rmsnorm_quant_kernels.cu`.
- [ ] JIT path still builds the kernel unchanged (no behavioral change; only adds a header).
- [ ] No runtime semantics change - same macro, same call site.

## Follow-up suggestion

This class of bug (compiles via JIT, fails via `PREBUILD_KERNELS=1`) is only catchable by a wheel-build CI lane. Consider adding a per-PR job that runs `PREBUILD_KERNELS=1 pip wheel .` on at least one config; happy to send a separate proposal.

PR #2842, opened 2026-04-21
[release/v0.1.12] Cherry-pick #2839: add missing c10/hip/HIPException.h include

### LINKED ISSUE #2837: Building 0.1.12.post1 with PREBUILD_KERNELS fails
Trying to build v0.1.12.post1 with PREBUILD_KERNELS=1 fails with the following error:
[2026-04-21T02:45:09.750Z] #42 3621.8 /app/aiter/aiter_meta/csrc/kernels/gated_rmsnorm_quant_kernels.cu:223:5: error: use of undeclared identifier 'C10_HIP_KERNEL_LAUNCH_CHECK'
Full log can be found at https://ml-ci-internal.amd.com/job/vLLM/job/vLLM-Base/job/main/229/console

sunway513:
Cherry-picks #2839 (by @ChuanLi1101) onto `release/v0.1.12` to prepare `v0.1.12.post2`.

## Why

Building v0.1.12.post1 from source with `PREBUILD_KERNELS=1` fails in vLLM's `ml-ci-internal` base image:

```
csrc/kernels/gated_rmsnorm_quant_kernels.cu:223:5: error: use of undeclared identifier 'C10_HIP_KERNEL_LAUNCH_CHECK'
```

AITER CI doesn't catch this because `rocm/pytorch:latest` happens to pull `<c10/hip/HIPException.h>` transitively through other includes in this file. vLLM ml-ci-internal's base image doesn't, so the build fails. Per Richard's audit, this macro is referenced in exactly one `.cu` file across all of `csrc/`, so a single explicit include fully resolves it.

Fixes #2837. See #2839 for the full root-cause analysis from @ChuanLi1101.

## Diff

One line added after the existing `#include <ATen/...>` block:

```diff
 #include "dispatch_utils.h"
 #include <ATen/hip/impl/HIPGuardImplMasqueradingAsCUDA.h>
+#include <c10/hip/HIPException.h>

 namespace aiter {
```

## CI

Adding `ci:vllm`, `ci:atom`, `ci:sglang`, `ci:triton-355` to validate downstream impact before merge — this is the `v0.1.12.post2` candidate so we want full coverage.

## Follow-up

Lingpeng raised the option of removing `C10_HIP_KERNEL_LAUNCH_CHECK` entirely instead of adding the header. We're going with the explicit include here to unblock @gshtras / vLLM 0.20 train today; macro removal will be tracked as a separate audit (no other `csrc/` files use it).

### DIFF
--- a/csrc/kernels/gated_rmsnorm_quant_kernels.cu
+++ b/csrc/kernels/gated_rmsnorm_quant_kernels.cu
@@ -6,6 +6,7 @@
 #include "aiter_opus_plus.h"
 #include "dispatch_utils.h"
 #include <ATen/hip/impl/HIPGuardImplMasqueradingAsCUDA.h>
+#include <c10/hip/HIPException.h>
 
 namespace aiter {
 


sunway513:
Heads up — opened #2846 as the supersede for this PR, per Lingpeng's preferred approach (remove the macro outright rather than rely on `#include <c10/hip/HIPException.h>`).

Context: Greg confirmed today (2026-04-21) that even with this PR's include applied, his build against an older PyTorch still fails to find `C10_HIP_KERNEL_LAUNCH_CHECK` — the macro is only defined in recent PyTorch versions of `c10/hip/HIPException.h`, so the include-based fix is brittle across PyTorch versions. Removing the macro (it's a one-liner used in exactly one place across all of `csrc/`, wrapping `hipGetLastError()`) eliminates the dependency entirely.

Leaving this PR open for Peng's call on whether to close.

sunway513:
Superseded by #2846 (Lingpeng's macro-removal fix), which is robust to PyTorch versions that don't define `C10_HIP_KERNEL_LAUNCH_CHECK` even when the c10/hip header is included — confirmed by @gshtras 2026-04-21 that the include-only fix in this PR was insufficient against vLLM ml-ci-internal's older PyTorch.

Closing in favor of #2846. Issue #2837 stays open until #2846 merges and a v0.1.12.post2 tag ships.

PR #2850, opened 2026-04-22
fix(flydsl): support AOT cache generation without GPUs

zhiding512:
Use CPU placeholders and cache-hit short-circuiting so FlyDSL AOT can populate caches in environments without visible HIP devices. Fail setup.py early when any FlyDSL AOT kernel compilation errors out.

Made-with: Cursor


coderfeli:
Not a good idea to hack cache and jit pass outside dsl

PR #2898, opened 2026-04-24
Fix CK 2stages MoE (always use BK1 = 16)

ex-rzr:
## Motivation

Fix for AICK-1084 and AICK-1160: FP8 2stages CK MoE produces incorrect results for intermediate dim < 256

## Technical Details

It is incorrect to pass AK1Value = BK1Value = 8 to work around static asserts triggered in thread transfers for some configs.
Preshuffling (shuffle_weight) uses KPack = 16 for FP8, so BK1 must be the same otherwise the kernel loads wrong values.
Instead ABlockTransferSrcScalarPerVector/ABlockTransferDstScalarPerVector_AK1 should be decreased in such cases so each thread load less data but all threads participate in loading.

## Submission Checklist


zufayu:
@ex-rzr This PR has been open since April 2026 with merge conflicts and no review activity since. Is the CK 2-stage MoE BK1=16 fix still something you plan to pursue? If yes, please rebase onto current main. If there's no update within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2907, opened 2026-04-24
[MLA] Fix qh128 ASM nullptr write; enable native qh128 fp8 on gfx950	

inkcherry:
for OSL/ISL=8k/1k  con=256 InferenceMax DSV3 decode case cc @Duyi-Wang  @slippedJim 
Restore the native `qh128` fp8 decode path on gfx950 that PR #2204 had to
disable for this case, by fixing the underlying nullptr-write bug in the qh128 ASM kernel
call site rather than working around it. This recovers MLA decode performance
on gfx950 for `nhead=128, fp8/fp8` while keeping gfx942 unchanged.

```bash
python3 op_tests/test_mla_persistent.py \
    -n 128,1 -d fp8 -kvd fp8 \
    -b  32 64 128 256 -c 8192
```
| bs  | without this PR     | with this PR        | speedup |
| --: | :------------------ | :------------------ | :-----: |
|  32 | 210 µs / 0.75 TB/s  | 105 µs / 1.50 TB/s  |  2.0×   |
|  64 | 413 µs / 0.76 TB/s  | 174 µs / 1.82 TB/s  |  2.4×   |
| 128 | 800 µs / 0.79 TB/s  | 300 µs / 2.10 TB/s  |  2.7×   |
| 256 | 1812 µs / 0.70 TB/s | 526 µs / 2.40 TB/s  |  3.4×   |

### DIFF
--- a/aiter/mla.py
+++ b/aiter/mla.py
@@ -248,9 +248,7 @@ def mla_decode_fwd(
         # Plan A safety: always allocate final_lse buffer.
         # qh128 ASM kernel (mla_a8w8_qh128_m32x4_n16x2_msk0_ps) writes ptr_LSEP
         # unconditionally; passing nullptr crashes on gfx950 at large batch.
-        final_lse_buf = torch.empty(
-            (total_s, nhead), dtype=dtypes.fp32, device=device
-        )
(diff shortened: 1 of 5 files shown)


inkcherry:
closed, fixed in https://github.com/ROCm/aiter/pull/2927

PR #2912, opened 2026-04-24
[Triton/Gluon] rmsnorm gluon kernel created for gfx1250

amd-jrosas:
## Motivation
Create rmsnorm kernel in gluon for gfx1250

## Technical Details
Translated existing triton implementation into a gluon equivalent. 

## Test Plan
Added a test reference in existing test_rmsnorm.py for gluon implementation.

## Test Result
Passed all test condition


aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,225 @@
...
+    col_offsets = gl.arange(0, BLOCK_SIZE, layout=col_layout)
+
+    # create a swizzled shared layout for the input
+    sharedLayoutInput: gl.constexpr = gl.SwizzledSharedLayout(1, 1, 1, order=[1, 0])
+
+    # create a swizzled shared layout for the weights
+    sharedLayoutWeights: gl.constexpr = gl.SwizzledSharedLayout(1, 1, 1, order=[0])
+
+    # create a swizzled shared layout for the output
+    gl.SwizzledSharedLayout(1, 1, 1, order=[1, 0])
```
vgokhale: This isn't assigned to anything but is probably also not needed since the output is not TDM stored so you don't need a shared layout for it?

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,225 @@
...
+            weights_ptr.dtype.element_ty, [nStages, BLOCK_SIZE], sharedLayoutWeights
+        )
+        # one time creation for handling the last block if n_cols is not a multiple of BLOCK_SIZE
+        n_col_blk = gl.cdiv(n_cols, BLOCK_SIZE) - 1
+        last_col_idx = n_col_blk * BLOCK_SIZE
+        mask = col_offsets < (n_cols - last_col_idx)
+
+        # Loop through the rows of the input tensor by NUM_PROG blocks
+        for row_idx in range(row_start, n_rows, NUM_PROG):
+            input_ptr + (row_idx * input_row_stride)
```
vgokhale: This isn't assigned to anything?

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,225 @@
...
+            # compute the sum of the square of the input
+            row_sq_sum = gl.sum(sum_sq, axis=0)
+            # compute the norm factor
+            norm_factor = gl.rsqrt((row_sq_sum / n_cols) + epsilon)
+
+            # compute the rms norm
+            rms_norm = a * norm_factor * weights
+            # store rms norm and the norm factor
+            gl.store(
+                rsigma_ptr + row_start, norm_factor.to(rsigma_ptr.dtype.element_ty)
```
vgokhale: Did you mean row_idx?

aiter/ops/triton/gluon/rmsnorm.py
```diff
@@ -0,0 +1,39 @@
...
+    ROW, COL = input.shape
+    output = torch.empty_like(input, device=input.device)
+    rsigma = torch.empty((ROW,), device=input.device, dtype=input.dtype)
+
+    MAX_FUSED_SIZE = 65536 // input.element_size()
+    BLOCK_SIZE = min(MAX_FUSED_SIZE, triton.next_power_of_2(COL))
+    USE_BLOCK = COL > BLOCK_SIZE
+    NUM_PROG = min(ROW, get_num_sms())
+
+    grid = (NUM_PROG,)
```
vgokhale: I think you can put  min(ROW, get_num_sms()) here.

aiter/ops/triton/gluon/rmsnorm.py
```diff
@@ -0,0 +1,39 @@
...
+def gluon_rms_norm_kernel(
+    input: torch.Tensor,
+    weights: torch.Tensor,
+    epsilon: float,
+):
+    ROW, COL = input.shape
+    output = torch.empty_like(input, device=input.device)
+    rsigma = torch.empty((ROW,), device=input.device, dtype=input.dtype)
+
+    MAX_FUSED_SIZE = 65536 // input.element_size()
```
vgokhale: Comment for the magic number?

aiter/ops/triton/gluon/rmsnorm.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: MIT
```
vgokhale: I don't think we want two different files. We want a single API and the wrapper decides whether to call triton (gfx950 and earlier) or gluon (gfx1250, if a gluon kernel exists).

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
amd-jrosas: That's correct. This version did not use TDM stores and any unused layouts/variables should be removed.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
amd-jrosas: Left over code from previous implementation and will be removed.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
vgokhale: Yeah I mean can you delete this line of deadcode?

vgokhale: Also I see below we are now using TDM stores?

aiter/ops/triton/gluon/rmsnorm.py
amd-jrosas: Okay. I can remove this file and create a reference in the existing API file that calls the triton implementation.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
amd-jrosas: Yes, I can delete that section of dead code. Correct, the latest version is using TDM stores.

aiter/ops/triton/gluon/rmsnorm.py
amd-jrosas: I can put that reference in line 24 but the variable it is assigned to is used in the kernel for both the blocking and non-blocking case.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
amd-jrosas: Yes, that should be row_idx. Updated in a later commit.

op_tests/triton_tests/normalization/test_rmsnorm.py
```diff
@@ -384,3 +386,30 @@ def test_rms_norm_dynamic_per_token_fp8_quant(
 
     torch.testing.assert_close(xq_dequant, ref_xq_dequant, atol=atol, rtol=rtol)
     torch.testing.assert_close(x_normed, ref_x_normed, atol=atol, rtol=rtol)
+
+
+_IS_GFX1250 = get_gfx() == "gfx1250"
+
+
+@pytest.mark.skipif(not _IS_GFX1250, reason="gluon rmsnorm requires gfx1250 hardware")
+@pytest.mark.parametrize("in_dtype_str", ["fp16", "bf16"])
```
vgokhale, line 395: Can you remove f16? We can just use bf16 for all tests.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,239 @@
...
+        n_col_blk = gl.cdiv(n_cols, BLOCK_SIZE) - 1
+        last_col_idx = n_col_blk * BLOCK_SIZE
+
+        # Loop through the rows of the input tensor by NUM_PROG blocks
+        for row_idx in range(row_start, n_rows, NUM_PROG):
+            # sum_sq store
+            sum_sq = 0.0
+            # preload the inputs and weights
+            gl.amd.gfx1250.tdm.async_load(input_desc, [row_idx, 0], smemInput.index(0))
+            gl.amd.gfx1250.tdm.async_load(weights_desc, [0], smemWeights.index(0))
```
vgokhale, line 92: This is loaded here but never used and reloaded below?

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,239 @@
...
+            # compute the rms norm
+            rms_norm = a * norm_factor * weights
+            # store rms norm and the norm factor
+            gl.store(rsigma_ptr + row_idx, norm_factor.to(rsigma_ptr.dtype.element_ty))
+            smemOutput1d = smemOutput.index(current_stage).reshape([BLOCK_SIZE])
+            smemOutput1d.store(rms_norm.to(output_ptr.dtype.element_ty))
+            gl.amd.gfx1250.tdm.async_store(
+                output_desc, [row_idx, 0], smemOutput.index(current_stage)
+            )
+            gl.amd.gfx1250.tdm.async_wait(0)
```
vgokhale, line 239: You want to double buffer this. Otherwise it is not overlapping store and compute.

aiter/ops/triton/_gluon_kernels/normalization/rmsnorm.py
```diff
@@ -0,0 +1,239 @@
...
+):
+
+    # map program id to row start and column offsets
+    row_start = gl.program_id(axis=0)
+
+    # Rank-1D layout: used for col_offsets, masks, zeros, etc.
+    col_layout: gl.constexpr = gl.BlockedLayout(
+        [8],  # sizePerThread
+        [32],  # threadsPerWarp  (warp size = 32 on gfx1250)
+        [4],  # warpsPerCTA
```
vgokhale, line 28: This is assuming 4 warps. Can you add an assert in the wrapper?

zufayu:

**Adds a gfx1250-only Gluon (TDM tensor-descriptor) RMSNorm kernel with a two-stage LDS software pipeline, and makes it the default `rms_norm` backend on gfx1250 via a new optional `backend="triton"|"gluon"` kwarg threaded through `rms_norm`/`_RMSNorm`.**

Validation (deterministic): NOT RUN — auto-validation disabled in this environment (REVIEW_AUTO_VALIDATE=0) and no head-matched report supplied; the single test target is op_tests/triton_tests/normalization/test_rmsnorm.py
Perf (advisory): NOT RUN — the runtime kernel path changed so timing was required, but the test target exposes no benchmark entry point and auto-validation was disabled; the repo's op_tests/op_benchmarks/triton/bench_rmsnorm.py already covers rms_norm and was neither run nor extended

⚠️ [verified] Default-path flip with zero CI coverage. aiter/ops/triton/normalization/rmsnorm.py resolves backend=None to "gluon" on gfx1250, so every rms_norm call on gfx1250 — including rmsnorm_forward_inference, which has no backend parameter and cannot opt out — runs the new kernel by default; but op_tests/triton_tests/normalization/test_rmsnorm.py skips every gluon case off gfx1250 and no default CI job runs gfx1250 hardware (triton CI runs on MI35X/gfx950; the gfx1250 suite is opt-in via the ci:gfx1250-ffm-triton label). The new default path can merge having never run in CI, and the description states only WHAT ("translated existing triton implementation") with no mechanism or numbers — elevated AI-code risk. **Reviewer should ask** for a gfx1250 suite run (or the ci:gfx1250-ffm-triton label) before merge, or keep the default on "triton" until the gluon path is validated on hardware.

⚠️ [verified] backend silently discarded on large-M/small-N shapes. In aiter/ops/triton/normalization/rmsnorm.py, _RMSNorm.forward routes M>8192, N<=2048 (e.g. 16380x1536 or 364800x128) to _rmsnorm_forward_large_m_small_n without forwarding or checking backend, so on gfx1250 both the default gluon selection and an explicit backend="gluon" are silently ignored with no warning; the test rows at those shapes pass through that branch, so they assert nothing about the gluon kernel even on gfx1250. **Author must** warn or assert when backend="gluon" is dropped, and stop counting the large-M rows as gluon coverage.

⚠️ [inferred] No measurement behind a default kernel swap. The PR ships zero timing data (no number in the description, no benchmark run), while op_tests/op_benchmarks/triton/bench_rmsnorm.py already covers rms_norm and the repo's own convention says a kernel ships as kernel + wrapper + test + benchmark; NUM_WARPS=4 and nStages=2 are hardcoded in aiter/ops/triton/normalization/rmsnorm.py and in the new kernel, outside the config registry (get_normalization_config / configs/gfx1250/) that the backward path in the same file already uses. Is the gluon kernel actually faster than the triton kernel it replaces as the default, and are NUM_WARPS=4/nStages=2 tuned or inherited? **Author must** run base-vs-head bench_rmsnorm.py on gfx1250 and move the launch config into the registry.

⚠️ [verified] Missing input guards the sibling gluon wrapper carries. The new branch in aiter/ops/triton/normalization/rmsnorm.py builds TDM descriptors on x/weight with strides [x.stride(0), 1] and no contiguity or shape/dtype checks, while the merged gluon sibling aiter/ops/triton/normalization/fused_rmsnorm_add.py asserts x.dim/weight shape/dtype and does x = x.contiguous() before make_tensor_descriptor. If a caller passes a column-sliced or transposed view (stride(1) != 1, e.g. bigger[:, ::2]) or a misaligned base, the descriptor would silently misdescribe the layout and read wrong values with no crash. **Author must** match the sibling wrapper's contiguous()/assert guards or state why the TDM path tolerates strided inputs.

📝 [verified] Housekeeping before merge. The new kernel file aiter/ops/triton/_gluon_kernels/gfx1250/norm/rmsnorm.py carries no SPDX/copyright header (every other kernel in the tree has one) and uses bare @gluon.jit where the repo's ops-triton convention requires make_kernel_repr + @gluon.jit(repr=...) for Gluon entry kernels (the triton kernel it translates has one); the PR is also stale against main — the large-M/small-N specialization (commit 4b6184fef) touched the same two files after this PR's base, the patch no longer applies directly (3-way applies), so any CI result on it describes a moved tree. **Author must** add the header/repr and rebase before merge.

PR #2917, opened 2026-04-25
Add a new kernel that supports decode mla with sub_kv=64 and sub_qh=8

JohnNikolay84:
## Motivation

Customers have been asking for a mla decode kernel with the following config:

qType bf16
kvType bf16
QH 8
ps 0
prefill 0
causal 0
seqlen 1

## Technical Details

This is really a cosmetic change to the existing qh16 kernel. with seqlen1 whether we use 8 or 16 lanes in matrix Q does not really matter. It is still an M=16 gemm and those extra lines will be just zeroes and dropped.

QH8 non-persistent kernels are introduced for mi300 and mi350.

## Test Plan

modify existing aiter mla test to support qh=8 and make sure it passes with the new kernel.

## Test Result

python op_tests/test_mla.py     -n 8,1     -k 512 -qn 192 -qr 64 -vh 256     -d bf16 -kvd bf16     -b 1 4 16     -c 64 256

PASS

## Submission Checklist


Byeong-Chan:
@JohnNikolay84 

Hello! Thanks for great works!

I have one question.

Do you have any plan of sub_qh=64 for Context Parallel?

*edit

Do you have any plan of fp8 sub_qh=64 for Context Parallel in gfx942?

JohnNikolay84:
@Byeong-Chan 

https://github.com/ROCm/aiter/tree/main/hsa/gfx950/mla

Isn't qh=64  already being supported for both BF16 and FP8? with decode (qseqlen1) and small-batch (qseqlen4) variants, and with/without LSE output.

Byeong-Chan:
ahh, I'm sorry about my question.

What I actually wanted to know is qh=64 for fp8 in gfx942.

JohnNikolay84:
@Byeong-Chan , do you need analogs of these  kernels for gfx942?

https://github.com/ROCm/aiter/blob/main/hsa/gfx950/mla/mla_a8w8_qh64_qseqlen1_gqaratio64_lse_v3_ps.co
https://github.com/ROCm/aiter/blob/main/hsa/gfx950/mla/mla_a8w8_qh64_qseqlen1_gqaratio64_v3_ps.co
https://github.com/ROCm/aiter/blob/main/hsa/gfx950/mla/mla_a8w8_qh64_qseqlen4_gqaratio16.co

I do not think those exist at the moment. gfx950 and gfx942 are different enough for the kernels in asm to be quite different too.  We can add implementing those kernels into our backlog.

Byeong-Chan:
I really appreciate it.

valarLip:
can we have some perf data compare to the solution that pad head to 16

tvirolai-amd:
> can we have some perf data compare to the solution that pad head to 16

Even if this would be slower we'd still like this in for functional reasons, and then padding can be added as performance improvement.

Reason for this is that currently we need to make upstream PR for every new model arch that happens to have QH of 8. With this we'd work out of the box which is vast improvement for user experience.

valarLip (approved):

JohnNikolay84:
This kernel is doing better than padded qh16 with bf16 kv cache
image

PR #2921, opened 2026-04-27
[TRITON] Set RNG seed in Triton tests

brunomazzottiamd:
## Motivation

Triton tests can be flaky and non-reproducible because RNG state leaks across test cases. Seeds were either missing, set once at module import time, or set before `pytest.skip` checks - so a single test could have different inputs depending on collection order or which earlier tests ran. This PR makes the Triton test suite deterministic at the file, function, and case level so failures are reliably reproducible.

## Technical Details

* Added `torch.manual_seed(...)` at the start of every parametrized test case (and inside shared input-generation helpers like `generate_batched_gemm_a8w8_inputs`) so each case re-seeds the RNG before allocating tensors.
* Removed the file-level `torch.manual_seed(0)` in `test_fused_fp8_quant.py` in favor of per-test seeding.
* In `test_mha.py`, moved `torch.cuda.empty_cache()` / `torch.manual_seed(20)` to *after* the `pytest.skip` guards in `test_mha_backward` and `test_mha_backward_varlen`, so skipped configurations no longer perturb the RNG stream of subsequent tests.

## Test Plan

Run the entire Triton test suite through CI, on both `gfx942` and `gfx950` runners.

## Test Result

All Triton tests have passed. ✅

## Submission Checklist

### DIFF
--- a/op_tests/triton_tests/attention/test_fav3_sage.py
+++ b/op_tests/triton_tests/attention/test_fav3_sage.py
@@ -185,6 +185,7 @@ def input_helper(
         k_shape = (BATCH, N_CTX_K, HK, D_HEAD)
         v_shape = (BATCH, N_CTX_K, HK, D_HEAD_V)
 
+    torch.manual_seed(20)
     q = torch.randn(q_shape, device="cuda", dtype=dtype)
     k = torch.randn(k_shape, device="cuda", dtype=dtype)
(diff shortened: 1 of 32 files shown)


nidal567 (approved):
LGTM! Thanks Bruno

k50112113 (approved):
LGTM

PR #2943, opened 2026-04-29
Make `rmsnorm2d_fwd` Handle Strided Higher-Rank Inputs Safely

hubertlu-tw:
## Summary

This PR fixes a gfx942 memory fault in Aiter's public `rmsnorm2d_fwd` wrapper when callers pass higher-rank strided tensors, such as Q/K views sliced from a packed QKV projection.

The wrapper now normalizes inputs to a contiguous 2D view before dispatching to the existing 2D RMSNorm kernels, then reshapes the output back to the caller's original shape.

## Motivation

SGLang's Qwen3 Q/K normalization path can pass tensors shaped like:

```text
q: shape=(2048, 32, 128), stride=(6144, 128, 1)
k: shape=(2048,  8, 128), stride=(6144, 128, 1)
```

These tensors are strided 3D views sliced from a packed QKV projection. Calling Aiter's low-level `rmsnorm` kernel directly on those views memory-faults on MI325/gfx942:

```text
Memory access fault by GPU node-2 ... Reason: Unknown.
Fatal Python error: Aborted
```

The same workload does not reproduce on MI355/gfx950, which suggests a gfx942-sensitive kernel/layout issue.

## Root Cause

`rmsnorm2d_fwd` is a public API, but its fast path passed `input` directly into a low-level kernel that expects a contiguous 2D tensor:

```python
out = torch.empty_like(input, dtype=input.dtype, device=input.device)
rmsnorm(out, input, weight, epsilon)
```

For higher-rank strided inputs, `torch.empty_like(input)` preserves the higher-rank shape, and the low-level 2D kernel receives a layout it does not safely support.

## Fix

Add layout normalization in `aiter/ops/rmsnorm.py`:

```python
input_2d, original_shape = _as_2d_contiguous(input)
...
out = rmsnorm2d_kernel(input_2d, ...)
return _restore_shape(out, original_shape)
```

The public API keeps the same output shape as before, but the low-level 2D kernel only sees a safe contiguous 2D tensor.

## Files Changed

```text
aiter/ops/rmsnorm.py
op_tests/repro_rmsnorm2d_strided_layout.py
```

## Reproducer

Safe fixed path:

```bash
cd /sgl-workspace/aiter
python op_tests/repro_rmsnorm2d_strided_layout.py --mode fixed
```

Expected:

```text
q_input: shape=(2048, 32, 128) stride=(6144, 128, 1) contiguous=False ...
k_input: shape=(2048, 8, 128) stride=(6144, 128, 1) contiguous=False ...
q_output: shape=(2048, 32, 128) ...
k_output: shape=(2048, 8, 128) ...
FIXED_PUBLIC_API_OK
```

Compare fixed public API against the old unsafe low-level path:

```bash
cd /sgl-workspace/aiter
python op_tests/repro_rmsnorm2d_strided_layout.py --mode both
```

On gfx942, expected behavior:

```text
=== Fixed public API path ===
FIXED_PUBLIC_API_OK

=== Old low-level path in subprocess ===
Calling low-level rmsnorm(...) directly on strided 3D views...
Memory access fault by GPU node-2 ...
OLD_LOW_LEVEL_SUBPROCESS_EXIT_CODE=-6
```

The unsafe path runs in a subprocess so the parent process can report the failure and clean `gpucore.*` files.

## Validation

Validated on MI325/gfx942:

```bash
python op_tests/repro_rmsnorm2d_strided_layout.py --mode fixed
```

Result:

```text
FIXED_PUBLIC_API_OK
```

Validated that the old low-level path still reproduces the gfx942 fault:

```bash
python op_tests/repro_rmsnorm2d_strided_layout.py --mode both
```

Result:

```text
OLD_LOW_LEVEL_SUBPROCESS_EXIT_CODE=-6
```

## Notes

- This is not a backend fallback.
- This preserves Aiter RMSNorm use.
- This is a public wrapper layout fix: callers can pass higher-rank tensors and receive the same higher-rank output shape, while the kernel receives the 2D contiguous layout it expects.
- SGLang also needs to keep its LFM/Qwen call sites correct, but this Aiter-side change makes the public API robust and prevents GPU faults from unsafe layouts.

### DIFF
--- a/aiter/ops/rmsnorm.py
+++ b/aiter/ops/rmsnorm.py
@@ -9,6 +9,28 @@ from typing import Optional
 MD_NAME = "module_rmsnorm"
 
 
+def _prepare_rmsnorm2d_input(
+    input: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Size | None]:
+    """Return a safe 2D contiguous tensor for RMSNorm kernels.
+
+    The low-level RMSNorm kernels operate on 2D contiguous inputs. Keep the
+    common already-safe layout as a zero-copy fast path, and only normalize
+    higher-rank or strided views.
+    """
+    if input.dim() == 2 and input.is_contiguous():
+        return input, None
+
+    original_shape = input.shape
+    return input.contiguous().reshape(-1, original_shape[-1]), original_shape
+
+
+def _restore_shape(output: torch.Tensor, original_shape: torch.Size | None) -> torch.Tensor:
+    if original_shape is None:
+        return output
+    return output.reshape(original_shape)
+
+
 @compile_ops("module_rmsnorm")
 def rms_norm_cu(
     out: Tensor,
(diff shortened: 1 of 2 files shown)


zufayu:
@hubertlu-tw This PR has been open since April 2026 with merge conflicts and no review activity since. Is the rmsnorm2d_fwd strided higher-rank input fix still something you plan to pursue? If yes, please rebase onto current main. If there's no update within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2964, opened 2026-04-29
[TRITON] Fix: Prevent null pointer dereference with empty descale tensors

juuso-oskari:
Fixes a memory access fault in unified_attention kernel when used with empty descale tensors.

Root cause:
- The kernel checks `if q_descale_ptr is not None` but doesn't validate that the tensor is non-empty before calling `tl.load(q_descale_ptr)`
- When vLLM passes empty tensors (torch.empty(0)), the data_ptr() is 0x0
- This causes a null pointer dereference during CUDA graph capture

Symptoms:
- Memory access fault by GPU on address (nil)
- Crash at ~98% during "Capturing CUDA graphs (decode, FULL)"
- vLLM server fails to start with FP8 models

Fix:
- Added _normalize_empty_tensor() helper that converts empty tensors to None
- Applied to q_descale, k_descale, v_descale, and output_scale
- Works for both 2D and 3D kernel paths
- Zero performance impact (one-time check before kernel launch)

Tested with:
- Unit tests covering None, valid, and empty tensor cases
- CUDA graph capture and replay
- vLLM server with Qwen/Qwen3-Next-80B-A3B-Instruct-FP8 model
- All tests pass, server starts successfully

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/attention/unified_attention.py
+++ b/aiter/ops/triton/attention/unified_attention.py
@@ -139,6 +139,21 @@ def unified_attention(
     if sinks is not None:
         assert sinks.shape[0] == q.shape[1], "Sinks must be num_query_heads size"
 
+    # FIX: Normalize empty tensors to None to prevent null pointer dereference
+    # Empty tensors have data_ptr()=0 which causes memory access faults in the kernel.
+    # This happens when vLLM passes empty tensors instead of None for unused descale factors.
+    # See: https://github.com/vllm-project/vllm/issues/...
+    def _normalize_empty_tensor(tensor):
+        """Convert empty tensors to None to prevent null pointer dereference in kernel."""
+        if tensor is not None and tensor.numel() == 0:
+            return None
+        return tensor
+
+    q_descale = _normalize_empty_tensor(q_descale)
+    k_descale = _normalize_empty_tensor(k_descale)
+    v_descale = _normalize_empty_tensor(v_descale)
+    output_scale = _normalize_empty_tensor(output_scale)
+
     use_alibi_slopes = alibi_slopes is not None
     use_qq_bias = qq_bias is not None
     SLIDING_WINDOW = 1 + window_size[0]


aiter/ops/triton/attention/unified_attention.py
```diff
@@ -139,6 +139,21 @@ def unified_attention(
     if sinks is not None:
         assert sinks.shape[0] == q.shape[1], "Sinks must be num_query_heads size"
 
+    # FIX: Normalize empty tensors to None to prevent null pointer dereference
+    # Empty tensors have data_ptr()=0 which causes memory access faults in the kernel.
+    # This happens when vLLM passes empty tensors instead of None for unused descale factors.
+    # See: https://github.com/vllm-project/vllm/issues/...
```
brunomazzottiamd, line 145: Is it possible to link to a specific VLLM issue?

brunomazzottiamd (approved):
LGTM!

**Question:** Do we need this kind of defensive behavior against empty tensors in other kernels?

aiter/ops/triton/attention/unified_attention.py
juuso-oskari: Ill change this still into a draft, when we are investigating form VLLM side

juuso-oskari: I think other kernels have defaults as None for tensors that are not needed to be passed (such as descales). Likely the same should be done here. I think the way the VLLM calls this is rather weird and passes empty tensors when it should pass Nones.

brunomazzottiamd: OK! Please let me know when it's ready to be reviewed!

zufayu:
@juuso-oskari This PR has been inactive for over 2 months. Is the null-pointer-dereference fix still applicable? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #2986, opened 2026-04-30
[Bugfix] Fix TypeError for per_1x128 FP8 blockscale path on gfx950 with ntok <= 32

peymanr:
## Motivation

On gfx950 (MI300X) with the per_1x128 FP8 blockscale quantisation path, calling `fused_moe` with `ntok <= 32` (decode regime) raises a `TypeError` at runtime. The root cause is that the `xbf16` fast-path branch in `fused_moe_1stage` skipped the normal `q_dtype_a` resolution logic, leaving the activation dtype in an inconsistent state when `hidden_states.dtype == fp8` and the token count is small. This PR fixes the bug, removes the now-unnecessary `xbf16` parameter, and cleans up several related dead-code paths that accumulated around the same logic.

### Bug fix — `q_dtype_a` resolution (`fused_moe_`)

The previous code contained a multi-branch heuristic that set `q_dtype_a` differently depending on `get_gfx()`, `M`, and `hidden_states.dtype`. The branch for `QuantType.per_1x128` with a pre-quantised FP8 input on gfx950 could leave `q_dtype_a` pointing at `w1.dtype` (bf16) while `hidden_states` was already fp8, causing a shape/type mismatch downstream. The fix collapses the heuristic to a single line:

```python
q_dtype_a = dtypes.fp4x2 if quant_type == QuantType.per_1x32 else q_dtype_a
```

All other quant types inherit `q_dtype_a` from the weight dtype, which is the correct default.

### Removal of `xbf16` parameter (`fused_moe_1stage`)

The `xbf16` flag was introduced to skip host-side quantisation for the blockscale bf16 ASM kernel. The same effect is now achieved by the existing `hidden_states.dtype != q_dtype_a` guard, so the flag and its branch are removed. `fused_moe_1stage` and `get_2stage_cfgs` / `MOEMetadata` no longer accept or propagate `xbf16`.

### `mxfp4_moe_sort_fwd` → `fp4_utils.moe_mxfp4_sort`

The sort helper is replaced with the canonical `fp4_utils.moe_mxfp4_sort` API, which accepts `block_size_M` directly instead of `cols`, matching the actual sorted-scale layout expected by the MXFP4 kernel.

### `_get_scale_t_buf` for partial-transpose output

The transposed-scale buffer is now obtained via `_get_scale_t_buf(a1_scale)` (a cached allocation helper introduced in PR 4) instead of `torch.empty_like`, avoiding a repeated HBM allocation on every decode step.

### `get_2stage_cfgs` — block_m and dead-code removal

- **block_m for force_1stage decode**: changed from `BLOCK_SIZE_M=32` to `16`. Using 32 doubles the moe_sorting output buffer (256 experts × 32 = 8 192 padded tokens vs 256 × 16 = 4 096), wasting ~4× HBM bandwidth for sorted-id/weight arrays in the decode regime.
- **Presorted-ps kernel resolution**: kernel names for the force_1stage override are now resolved via module-level constants (`_FAST_PATH_KERNELNAME_BF16`, `_GFX950_BLOCKSCALE_NOVS_KERNELS`) rather than repeated dict lookups.
- **Dead code removed**: `ksplit` override via `AITER_KSPLIT=-1`, `use_non_temporal_load`, the entire flydsl two-stage wrapper path, and the `get_block_m()` inner function are all removed.
- **`fused_moe_1stage_dict` pruned**: stale entries for `(Gelu, per_1x128)` and `(Silu/Gelu, per_Token, fp8, non-tkw1)` are removed; these dispatch targets are no longer reachable after the quant-type simplification.

### Related PRs in this series

| PR | Focus |
|----|-------|
| **This PR (PR 1)** | Bugfix: TypeError on gfx950 ntok ≤ 32, per_1x128 FP8 blockscale |
| PR 2 | Perf: Close doweight_stage1=False CSV gap for MiniMax ntok 33–512 on gfx950 |
| PR 3 | Tuning: Add MiniMax a8w8 blockscale tuned configs for gfx950 |
| PR 4 | Perf: Cache gfx/cu_num queries and scale-transpose buffer in fused_moe |

## Test Plan

1. **Regression — decode path (ntok ≤ 32, per_1x128 FP8 blockscale, gfx950)**
   Run a minimal reproduce script that calls `fused_moe` with `ntok=1` and `ntok=32`, `quant_type=QuantType.per_1x128`, pre-quantised FP8 `hidden_states`, on a gfx950 device. Confirm no `TypeError` is raised and outputs match a bf16 reference within tolerance.

2. **Regression — prefill path (ntok > 32, per_1x128 FP8 blockscale, gfx950)**
   Confirm existing behaviour is preserved for `ntok=64`, `ntok=512`, `ntok=2048`.

3. **Regression — per_1x32 MXFP4 path**
   Confirm `q_dtype_a = dtypes.fp4x2` is set correctly and `fp4_utils.moe_mxfp4_sort` produces the same sorted-scale layout as the previous `mxfp4_moe_sort_fwd` call.

4. **Regression — per_Token paths (bf16, fp8 tkw1)**
   Confirm the pruned `fused_moe_1stage_dict` entries do not affect the remaining dispatch targets.

5. **Full test suite**
   ```bash
   ```

6. **Multi-GPU**
   ```bash
   MULTIGPU=TRUE bash .github/scripts/aiter_test.sh
   ```

## Test Result

| Scenario | Before PR | After PR |
|----------|-----------|----------|
| `fused_moe` ntok=1, per_1x128 FP8, gfx950 | ❌ `TypeError` | ✅ Correct output |
| `fused_moe` ntok=32, per_1x128 FP8, gfx950 | ❌ `TypeError` | ✅ Correct output |
| `fused_moe` ntok=64, per_1x128 FP8, gfx950 | ✅ Correct output | ✅ Correct output |
| `fused_moe` ntok=1, per_1x32 MXFP4, gfx950 | ✅ Correct output | ✅ Correct output |
| Full test suite (`aiter_test.sh`) | TBD | TBD |

> **Note:** Before-PR numbers for the broken paths are unavailable as a numeric baseline because the code path raised a `TypeError` before producing any output. Performance benchmarks for the decode path will be provided in PR 2 (perf) and PR 4 (perf) of this series.

## Submission Checklist

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -255,23 +255,7 @@ def fused_moe_(
     quant_type = quant_remap.get(quant_type, quant_type)
     q_dtype_w = w1.dtype
     q_dtype_a = w1.dtype if w1.dtype != torch.uint32 else dtypes.fp8
-    # If input is already FP8-quantized (e.g. from FP8 dispatch) with block scale,
-    # use FP8 as activation dtype to skip redundant re-quantization
-    if (
-        quant_type == QuantType.per_1x128
-        and hidden_states.dtype == dtypes.fp8
-        and a1_scale is not None
-    ):
-        q_dtype_a = dtypes.fp8
-    bf16_fp8_bound = 256
-    if quant_type == QuantType.per_1x32:
-        if activation == ActivationType.Swiglu:
-            if get_gfx() != "gfx950" or M < bf16_fp8_bound:
-                q_dtype_a = dtypes.bf16
-            elif M >= bf16_fp8_bound:
-                q_dtype_a = dtypes.fp8
-        else:
-            q_dtype_a = dtypes.fp4x2
+    q_dtype_a = dtypes.fp4x2 if quant_type == QuantType.per_1x32 else q_dtype_a
 
     metadata = get_2stage_cfgs(
         get_padded_M(M),  # consider token_num > 1024 as prefill
@@ -378,7 +362,6 @@ def fused_moe_1stage(
     block_size_M=32,
     activation=ActivationType.Silu,
     quant_type=QuantType.No,
-    xbf16=False,
     kernelName: str = "",
     # following for quant
     q_dtype_a=None,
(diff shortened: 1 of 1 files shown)


peymanr:
Closing to resubmit as a cleaner series — architecture fix in progress.

Issue #2996, opened 2026-05-01
bf16 MLA decode-stage1 silently produces wrong outputs at page_size > 1 (no _ps kernel shipped)


kzjeef:
# bf16 MLA decode-stage1 silently produces wrong outputs at `page_size > 1` (no `_ps` kernel shipped)

## Summary

`aiter.mla.mla_decode_fwd` accepts `page_size` as a parameter, but on the bf16
absorbed-MLA path used by DeepSeek-V3 / Kimi-K2 / Kimi-K2.5, **any
`page_size > 1` silently produces incorrect outputs** instead of erroring out.
The Python wrapper passes `s_log2_plen = log2(page_size)` to the kernel, but
the underlying assembly blob loaded in this case
(`mla_dec_stage1_bf16_a16w16_subQ128_mqa128.co` /
`mla_dec_stage1_bf16_a16w16_subQ16_mqa16.co`) is `ps=0` per
`hsa/gfx950/mla/mla_asm.csv` — i.e. it doesn't honor `s_log2_plen`. There is no
`_ps` (paged-aware) variant of these two kernels shipped in
`hsa/gfx950/mla/`.

End-user impact: serving stacks that flip `--page-size 32`/`64`/etc. on
ROCm + Kimi/DeepSeek (e.g. sglang, vLLM with the aiter backend) generate
"fast but wrong" tokens — model output becomes incoherent word-salad while the
kernel reports up to **170× higher decode throughput**.

This is also why sglang has the comment

```
# Setting page size to 1 for DeepSeek DSA on ROCm.
```

in `server_args.py:1654`. It's not a sglang choice — it's an AITER
ship-list gap.

## Reproduction (AITER's own unit test)

`op_tests/test_mla.py` already covers this — the default `-blk 1` passes,
every other `-blk` fails the absorb path:

```
$ python3 op_tests/test_mla.py -blk 1   -c 256 -b 4 -n 16,1
mla_prefill-normal    [torch vs  aiter_ck]:    17.73 us......PASSED
mla_prefill-absorb    [torch vs aiter_asm]:    40.59 us......PASSED
mla_decode-absorb     [golden vs aiter_asm]:    13.54 us......PASSED

$ python3 op_tests/test_mla.py -blk 16  -c 256 -b 4 -n 16,1
mla_prefill-normal    [torch vs  aiter_ck]:    17.39 us......PASSED
mla_prefill-absorb    [torch vs aiter_asm]:    40.27 us......FAILED
[...]

PR #2997, opened 2026-05-01
mla: refuse page_size > 1 on bf16 decode-stage1 kernel (no _ps variant shipped)

kzjeef:
Refs #2996

## What

Refuse `page_size != 1` for the bf16 decode-stage1 MLA path used by
DeepSeek-V3 / Kimi-K2 / Kimi-K2.5, raising `NotImplementedError` instead of
silently producing wrong outputs. Adds an `op_tests/test_mla_page_guard.py`
covering the new behavior.

## Why

The kernel selector in `csrc/cpp_itfs/mla/asm_mla_decode_fwd.py:46` always
loads the `ps=0` blob (`mla_dec_stage1_bf16_a16w16_subQ{16,128}_mqa{16,128}.co`).
The launcher template still packs `s_log2_plen=log2(page_size)`, but the
assembly ignores the scalar and reads the cache with `page_size=1` addressing
even when `page_size>1` is requested.

End-to-end on sglang + Kimi-K2.5-MXFP4 at TP=4: `--page-size 32` makes the
model emit word-salad while reporting +65 % decode throughput (per-call
latency of `kn_mla_reduce_v1<128,16,4>` drops from ~4 ms to ~23 µs). AITER's
own `op_tests/test_mla.py` shows the same regression at `-blk {16,32,64,128}`
on the prefill-absorb / decode-absorb checks:

```
$ python3 op_tests/test_mla.py -blk 1   -c 256 -b 4 -n 16,1
mla_prefill-absorb [torch vs aiter_asm]: ... PASSED
mla_decode-absorb  [golden vs aiter_asm]: ... PASSED

$ python3 op_tests/test_mla.py -blk 32  -c 256 -b 4 -n 16,1
mla_prefill-absorb [torch vs aiter_asm]: ... FAILED
mla_decode-absorb  [golden vs aiter_asm]: ... FAILED
```

CI only invokes `python3 op_tests/test_mla.py` with no flags, so the default
`--block_size 1` is the only configuration validated.

## What this PR does NOT do

Does **not** enable multi-token paging on the bf16 path. The real fix is to
ship `mla_dec_stage1_bf16_a16w16_*_ps.co` kernels and update the selector +
`hsa/gfx950/mla/mla_asm.csv`. That work is tracked as TODOs in #2996. This
PR is a fail-loud guard so users stop being silently miscomputed at while
that lands.

## Files

- `aiter/mla.py:mla_decode_fwd` — guard at the user-facing entry. Fires on
  every call, no JIT-cache race.
- `csrc/cpp_itfs/mla/asm_mla_decode_fwd.py:compile()` — guard at the JIT
  compile entry, which is what the C++ wrapper invokes via subprocess for
  not-yet-built `(gqa_ratio, page_size, ...)` cache keys.
- `op_tests/test_mla_page_guard.py` — new unit test:
  - `page_size=1` does **not** raise (the documented good path)
  - `page_size ∈ {16, 32, 64, 128}` all raise `NotImplementedError`
  - both `gqa_ratio=128` (Kimi/DSV3) and `gqa_ratio=16` are guarded

## Test plan

- [x] `python3 op_tests/test_mla.py -blk 1` still passes (decode/prefill absorb)
- [x] `python3 op_tests/test_mla.py -blk 32` raises `NotImplementedError` from
      `aiter.mla.mla_decode_fwd` instead of producing wrong outputs
- [x] `python3 op_tests/test_mla_page_guard.py` returns 0 (all 3 sub-tests pass)
- [x] Verified end-to-end: sglang serving Kimi-K2.5-MXFP4 at `--page-size 1`
      produces coherent output (math/code/factual prompts all sensible);
      `--page-size 32` now errors at startup instead of silently corrupting

## Suggested follow-ups (out of scope for this PR)

See the TODO list in #2996. In short: implement `_ps` variants of the bf16
decode-stage1 kernels, update the selector to pick them when `page_size>1`,
extend `mla_asm.csv`, and add a `-blk` matrix to `aiter_test.sh`'s CI
invocation so multi-page regressions surface in CI rather than user
deployments.

### DIFF
--- a/aiter/mla.py
+++ b/aiter/mla.py
@@ -190,6 +190,23 @@ def mla_decode_fwd(
     if sm_scale is None:
         sm_scale = 1.0 / (qk_head_dim**0.5)
 
+    # AITER ships ps=0 (page-1-only) decode-stage1 kernels for the bf16
+    # absorbed MLA path used by DeepSeek-V3 / Kimi-K2 / Kimi-K2.5. The launcher
+    # template still passes s_log2_plen=log2(page_size), but the underlying
+    # kernel binary ignores it, silently producing wrong outputs. Refuse loudly
+    # rather than corrupting generation. See hsa/gfx950/mla/mla_asm.csv column
+    # "ps" (0/1) — bf16 a16w16 dec_stage1 has no _ps variant shipped.
+    _bf16_path = (kv_buffer.dtype == torch.bfloat16
+                  or (kv_buffer.dtype == torch.uint8 and q.dtype == torch.bfloat16))
+    if page_size != 1 and _bf16_path:
+        raise NotImplementedError(
+            f"aiter.mla.mla_decode_fwd: page_size={page_size} is not safe with "
+            "the bf16 decode-stage1 kernel. Only page_size=1 is correct in "
+            "this AITER build (kernel asm has no paged variant). "
+            "Pass page_size=1 or wait for an AITER release shipping "
+            "mla_dec_stage1_bf16_a16w16_*_ps.co."
+        )
+
     ori_total_s, ori_nhead, ori_v_head_dim = o.shape
     total_s, nhead, v_head_dim = o.shape
     bs = qo_indptr.shape[0] - 1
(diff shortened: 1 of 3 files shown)


zufayu:
@kzjeef This PR has been inactive for over 2 months. Is the page_size>1 refusal for bf16 decode-stage1 still applicable? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3016, opened 2026-05-04
[MoE] Cache split-K scratch buffers to avoid per-call hipMalloc

frida-andersson:
## Summary

Replace per-call `torch.zeros`/`torch.empty` for split-K intermediate buffers with a persistent cache keyed by `(dtype, device)`. The buffer grows monotonically and is zeroed in-place, eliminating repeated `hipMalloc`/`hipFree` overhead in the MoE decode path.

## Motivation

During decode, each MoE forward call allocates and frees a split-K scratch buffer (`hipMalloc` + `hipFree`). On MI355X with DeepSeek-V3.2 (256 experts, TP4), this adds measurable stalls between MoE compute kernels visible in profiler traces. Caching the buffer removes these allocation gaps entirely.

## Changes

- Add `_splitk_scratch_cache` dict and `_get_splitk_scratch()` helper in `aiter/fused_moe.py`
- Replace allocation sites in `asm_stage1`, `ck_moe_stage1`, and `cktile_moe_stage1`

## Correctness

- Buffer is `.zero_()`'d on every call (split-K accumulates partial results)
- Keyed by `(dtype, device)` — no cross-type aliasing
- Grows monotonically — always large enough
- Sequential MoE execution in inference — no concurrent aliasing risk
- HIP graph compatible (persistent pointer stable across replays)

## Test plan

- Run DeepSeek-V3.2 TP4 decode benchmark on MI355X
- Verify no `hipMalloc` calls between MoE kernels in torch profiler trace
- Compare TPOT before/after


azaidy:
@valarLip need urgent review on this

valarLip:
this one not reasonable to me, please provide info maybe like the trace.. @frida-andersson

PR #3004, opened 2026-05-01
[Silo] Bulk merge: tuned GEMM and FMoE configs (GLM-4.7, Kimi-K2.5, MiniMax-M2.5, DeepSeek-V3.2)

sunway513:
## Summary

Bulk merge of 5 Silo config-only PRs for v0.1.13 deadline (vLLM 0.21 freeze 2026-05-08).

**Pure CSV tuning configs — no code changes.**

### Included PRs

- #2923 — GLM-4.7 FP8 tuned/untuned FMoE configs
- #2938 — Kimi K2.5 FP4 fused MoE tunings (TP2 / 256 CU refresh)
- #2979 — MiniMax M2.5 A8W8 blockscale GEMM tunings
- #2981 — DeepSeek-V3.2 MI355X tuned GEMM and FMoE configs
- #2982 — MiniMax M2.5 FMoE tunings

### Files Changed

All under `aiter/configs/`:
- `a8w8_blockscale_tuned_gemm.csv`
- `tuned_fmoe.csv`
- `model_configs/glm47_fp8_tuned_fmoe.csv` (new)
- `model_configs/glm47_fp8_untuned_fmoe.csv` (new)
- `model_configs/kimik2_fp4_tuned_fmoe.csv`

### Risk

Low — CSV config files only. No kernel code, no API changes, no build changes.

### Test Plan

- [ ] CI passes (no code changes, config load only)
- [ ] Spot check: ATOM loads new configs without error

PR #3023, opened 2026-05-04
Add MiniMax M25 FMoE Instances support

akii96:
Adds the MiniMax M25 FMoE codegen change that was left out of PR #3004 but is referenced by some of the tuned FMoE config.csv


akii96:
Ping @sunway513 @azaidy

PR #3026, opened 2026-05-04
[Perf] Cache GFX/CU-count, moe_sorting buffers, scale-transpose bufs, and quant partials

peymanr:
Part of [ROCm/aiter#3156](https://github.com/ROCm/aiter/issues/3156) — (PR 2/3 in series)

Stacks on [#3157](https://github.com/ROCm/aiter/pull/3157).

## Summary

Pre-allocates and reuses `moe_sorting` output buffers, `scale_t` transpose buffers, and quant function partials across decode steps to eliminate per-call HIP allocations in the `fused_moe` hot path. (`get_gfx()` and `get_cu_num()` already use `@functools.lru_cache` in `chip_info.py` and needed no change.)

### Changes
- **`moe_sorting` buffer cache** (`_moe_sorting_buf_cache`): the five output tensors (`sorted_ids`, `sorted_weights`, `sorted_expert_ids`, `num_valid_ids`, `moe_buf`) are allocated once per unique `(device, M, topk, num_experts, block_size, model_dim, dtype)` key and reused across calls when shape matches.
- **`scale_t` transpose buffer cache** (`_scale_t_cache`): the scale transpose buffer is cached per `(device, cols, rows, dtype)`.
- **Quant function partial cache** (`_quant_func_t_cache`): `functools.partial` objects wrapping the quantization dispatch are cached so Python object construction does not happen on every forward pass.
- **`block_m` default changed 32 → 16** for the 256-expert case: halves the `moe_sorting` buffer footprint and reduces HBM bandwidth waste in the sorting kernel. The cache key includes `block_m` so per-shape tuning remains possible.

### Motivation
In the MiniMax-M2.5 decode loop `fused_moe` is called on every token step. Each call was freshly allocating 6+ HIP tensors, adding allocator overhead that scales linearly with request rate. At CONC=64 this is the dominant non-compute cost.

## Test Plan
- `benchmark_moe_sorting_gemm_mi300x.py` — throughput regression check on MI300X
- Verify no regression at low concurrency (CONC=4) where allocation overhead is less significant

## Test Result

TBD — to be filled after CI run.

### DIFF
--- a/_apply.patch
+++ b/_apply.patch
@@ -1,0 +1,1488 @@
+--- a/aiter/fused_moe.py
++++ b/aiter/fused_moe.py
+@@ -1,5 +1,5 @@
+ # SPDX-License-Identifier: MIT
+-# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
++# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+ 
+ import functools
+ import os
+@@ -13,86 +13,13 @@
+ from aiter import ActivationType, QuantType, dtypes
+ from aiter import get_hip_quant as get_quant
+ from aiter import logger
+-from aiter.jit.core import AITER_CONFIGS, AITER_CSRC_DIR, PY, bd_dir, mp_lock
++from aiter.jit.core import AITER_CONFIGS, PY, bd_dir, mp_lock, get_asm_dir
+ from aiter.jit.utils.chip_info import get_cu_num, get_gfx
+ from aiter.jit.utils.torch_guard import torch_compile_guard
+-from aiter.ops.flydsl.utils import is_flydsl_available
+-from aiter.ops.flydsl.moe_common import GateMode
+-from aiter import fused_dynamic_mxfp4_quant_moe_sort, mxfp4_moe_sort_fwd
++from aiter.ops.triton.fused_mxfp4_quant import fused_dynamic_mxfp4_quant_moe_sort
++from aiter.utility import fp4_utils
+ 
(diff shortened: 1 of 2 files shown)


peymanr:
Good catch — `get_gfx()` and `get_cu_num()` are both decorated with `@functools.lru_cache(maxsize=1)` in `chip_info.py` so they already memoize after the first call. Removed the redundant module-level `_cached_gfx` / `_cached_cu_num` variables. The buffer pre-allocation caching (`_moe_sorting_buf_cache`, `_scale_t_cache`) is the actual perf change in this PR and is unaffected.

peymanr:
## Blocking issue: GPU crash on E=256, topk=8

This PR crashes with a signal 6 / abort (GPU-level, no torch exception) when called with the MiniMax-M2.5 / DeepSeek-V3 expert configuration:

- `num_experts=256`, `topk=8`, `hidden_size=7168`, `intermediate=2048`
- Reproduces on both **MI355X (gfx950)** and **MI300X (gfx942)** — not hardware-specific
- Mixtral shape (E=8, topk=2) works fine; only large expert counts trigger it

**Suspected cause:** Buffer aliasing in the `moe_sorting` cache. The cache is keyed on `(num_tokens, num_experts, block_m)` and returns tensors for reuse. At E=256/topk=8, a dispatch may receive a buffer still in-flight on the device (write-after-read hazard at the HIP level). The `block_m` default change from 32→16 for the 256-expert case may also affect cache key behavior for callers that don't pass `block_m` explicitly.

Holding as Draft until this is investigated. Happy to provide full crash logs if helpful.

PR #3041, opened 2026-05-06
[Triton] fused_flatten_fp8_group_quant: add transpose_scale param

Jacob0226:
## Motivation

`aiter/ops/triton/quant/fused_fp8_quant.py::fused_rms_fp8_group_quant` already supports `transpose_scale=True` (writes the per-token-group scale tensor in a pre-transposed `(num_bs_cols, M)` row-major storage so downstream consumers like CK bpreshuffle GEMM can read it directly without an extra `.transpose(-1, -2).contiguous()` copy on the consumer side).

`fused_flatten_fp8_group_quant` in the same file produces the same shape of scale tensor (`(M, cdiv(N1*N2, group_size))`), is fed into the same downstream GEMM path in sglang's `aiter_w8a8_block_fp8_linear`, but does NOT currently support `transpose_scale`. As a result, on the o_proj path of GLM-5.1-FP8 (and any other Linear consuming a flatten-quanted activation), sglang has to do an explicit `.transpose(-1, -2).contiguous()` on the scale tensor before each bpreshuffle GEMM, costing ~4us per call.

This PR mirrors the same `transpose_scale` flag onto `fused_flatten_fp8_group_quant` so the consumer can opt in and skip that redundant copy. Companion sglang patch: [Jacob0226/sglang:jacob/fused-transpose-scale](https://github.com/Jacob0226/sglang/tree/jacob/fused-transpose-scale) (also opened upstream as sgl-project/sglang#24476; this aiter PR is its prerequisite).

## Technical Details

`aiter/ops/triton/quant/fused_fp8_quant.py::fused_flatten_fp8_group_quant` (Python wrapper only — inner `_fused_flatten_fp8_group_quant_kernel` is unchanged):

* New keyword arg `transpose_scale: bool = False` (default preserves existing behavior; default callers see no change).
* When `transpose_scale=True`:
  - Allocate `out_block_scales` as `(num_bs_cols, M)` instead of `(M, num_bs_cols)`.
  - Pass swapped `out_bs_row_stride / out_bs_col_stride` (1, M) to the inner kernel so it writes column-major into the same buffer.
  - At the end, `view(M, num_bs_cols)` to keep the public shape identical.
* The pattern is identical to the existing `transpose_scale` handling on `fused_rms_fp8_group_quant` in the same file; this is just propagating it to the flatten variant.

Net diff: +33 / -4 lines, single function, single file.

## Test Plan

1. Run the entire existing UT file `op_tests/triton_tests/quant/test_fused_fp8_quant.py` to confirm:
   - `test_fused_flatten_fp8_group_quant` (default path of the patched function) still passes — verifies `transpose_scale=False` behavior is unchanged.
   - `test_fused_rms_fp8_group_quant` and `test_fused_rms_fp8_group_quant_transpose_scale` (the function whose `transpose_scale` handling this patch mirrors) still pass — verifies no cross-function regression.
2. Add `test_fused_flatten_fp8_group_quant_transpose_scale` to the same file — a new pytest case mirroring the sister `test_fused_rms_fp8_group_quant_transpose_scale`. Parametrized over `M ∈ {1, 32, 256} × (N1, N2) ∈ {(16, 128)} × dtype ∈ {float16, bfloat16}` (6 cases, same coverage as the existing default `test_fused_flatten_fp8_group_quant`). For each case it:
   - Calls `fused_flatten_fp8_group_quant` with `transpose_scale=False` and `transpose_scale=True`.
   - Asserts both outputs have shape `(M, num_bs_cols)` and are contiguous after the inner `.view()`.
   - Asserts `scale_transposed` is byte-equivalent to `scale_default.transpose(0, 1).contiguous().view(*scale_default.shape)` (the manual transform consumers like sglang's `aiter_w8a8_block_fp8_linear` perform on the bpreshuffle path).
   - Asserts the FP8 quanted tensor is bit-identical between the two calls (atol=0, rtol=0).
3. End-to-end performance measurement via the companion sglang PR (sgl-project/sglang#24476) on GLM-5.1-FP8 / MI355X / TP=8.

Hardware / software: MI355X (gfx950), ROCm 7.2, docker `rocm/sgl-dev:v0.5.10.post1-rocm720-mi35x-20260503`.

### Reproduce

```bash
cd <aiter>
# Full file (300 cases — regression + new transpose_scale UT):
pytest op_tests/triton_tests/quant/test_fused_fp8_quant.py -v

# Just the new transpose_scale case (6 cases):
pytest op_tests/triton_tests/quant/test_fused_fp8_quant.py::test_fused_flatten_fp8_group_quant_transpose_scale -v
```

### `op_tests/triton_tests/quant/test_fused_fp8_quant.py` — 300 / 300 PASS on MI355X (27s)

300 = 294 existing + 6 newly added. Highlights:

* `test_fused_flatten_fp8_group_quant` (default path of the patched function): **6 / 6 PASS** — confirms `transpose_scale=False` behavior is unchanged.
* `test_fused_flatten_fp8_group_quant_transpose_scale` (NEW, added in this PR): **6 / 6 PASS** — confirms `transpose_scale=True` produces a scale tensor byte-equivalent to `default.transpose(0, 1).contiguous().view(...)` and a bit-identical FP8 quantized tensor.
* `test_fused_rms_fp8_group_quant` and `test_fused_rms_fp8_group_quant_transpose_scale` (the sister function this patch mirrors): **36 / 36 PASS** — confirms no cross-function regression.

### End-to-end performance (via companion sglang PR sgl-project/sglang#24476)

* GLM-5.1-FP8 on MI355X TP=8, attention block: o_proj direct-copy kernel (~4 us / layer) eliminated in trace.
* E2E throughput: **+0.6% incremental** on top of the rms variant (`fused_rms_fp8_group_quant(transpose_scale=True)`) opt-in. Combined with the rms opt-in, **total +3.1% throughput / -3.1% TPOT** averaged across 10 ISL/OSL/concurrency points (1k1k, 8k1k × concurrency 4/8/16/32/64).
* GSM8K accuracy: 0.943 (within run-to-run noise of baseline 0.948) — confirms the optimization is byte-equivalent at the GEMM level.

## Submission Checklist


aiter/ops/triton/quant/fused_fp8_quant.py
```diff
@@ -497,25 +497,46 @@ def fused_flatten_fp8_group_quant(
...
+
+    if transpose_scale:
+        # Allocate as (num_bs_cols, M) so the row-major storage of this tensor
+        # is equivalent to a column-major (M, num_bs_cols) layout. We then
+        # tell the inner kernel to write with swapped strides, and view back
+        # to (M, num_bs_cols) at the end.
+        out_block_scales = torch.empty(
+            (num_bs_cols, M), dtype=torch.float32, device=x.device
+        )
+        out_bs_row_stride = out_block_scales.stride(1)  # = 1
```
vgokhale: I don't think you need all this. You are assigning rows and col strides based on layout (transpose or not) but then doing a view. Instead you can not do any of that and just supply the strides to the kernel. The addrs are computed like this in the kernel:

```
out_scales_ptr
    + m * out_scales_stride_m
    + (n1 * NUM_QUANT_BLOCKS + block_scale_offs) * out_scales_stride_n
```

It will do the correct thing regardless of whether it is row or col major.

op_tests/triton_tests/quant/test_fused_fp8_quant.py
```diff
@@ -365,6 +365,66 @@ def test_fused_flatten_fp8_group_quant(M: int, N1: int, N2: int, dtype):
     torch.testing.assert_close(y_upcast_torch, y_upcast_triton, atol=0.1, rtol=0.1)
 
 
+@pytest.mark.parametrize("M", [1, 32, 256])
+@pytest.mark.parametrize("N1, N2", [(16, 128)])
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
```
vgokhale: Please remove float16.

aiter/ops/triton/quant/fused_fp8_quant.py
Jacob0226: The same allocate-`(num_bs_cols, M)` + stride-swap + trailing `.view()` pattern is already in `fused_rms_fp8_group_quant` in this same file ([line 215–227](https://github.com/ROCm/aiter/blob/99855d59/aiter/ops/triton/quant/fused_fp8_quant.py#L215-L227), [281–287](https://github.com/ROCm/aiter/blob/99855d59/aiter/ops/triton/quant/fused_fp8_quant.py#L281-L287), [332–333](https://github.com/ROCm/aiter/blob/99855d59/aiter/ops/triton/quant/fused_fp8_quant.py#L332-L333)) — this PR is just mirroring it onto the flatten variant.

Verified that dropping the swap or the view does break the UT:

- **Drop the stride swap, just unpack `*out_block_scales.stride()`**: the buffer's natural strides are `(M, 1)`, kernel writes at `m*M + n` and overruns the buffer for any M ≥ 32. UT fails on `M=32` and `M=256` cases (OOB writes give `inf`/`2.7e+38`, 99.6% mismatch); `M=1` accidentally passes because `M*M + (num_bs_cols-1) < num_bs_cols*M` only holds at M=1.
- **Drop `.view()`**: output stride becomes non-row-major, fails the existing assertions `y_s_orig.stride() == y_s_transposed.stride()` and `is_contiguous()` (lines 411–416). The companion sglang fallback also calls `x_scale.view(num_bs_cols, M).transpose(0, 1).contiguous()` which needs contiguous input.

What I can clean up is the `.stride(0)/.stride(1)` attribute access → literals (UT 3/3 PASS):

```python
    if transpose_scale:
        out_block_scales = torch.empty(
            (num_bs_cols, M), dtype=torch.float32, device=x.device
        )
        out_bs_row_stride, out_bs_col_stride = 1, M
    else:
        out_block_scales = torch.empty(
            (M, num_bs_cols), dtype=torch.float32, device=x.device
        )
        out_bs_row_stride, out_bs_col_stride = num_bs_cols, 1
```

Happy to apply the same literal-stride cleanup to `fused_rms_fp8_group_quant` in this PR or a follow-up if you'd like consistency.

op_tests/triton_tests/quant/test_fused_fp8_quant.py
Jacob0226: Done — bf16 only on the new test in the next push. Left fp16 on the older `test_fused_rms_fp8_group_quant_transpose_scale` for now since it predates the convention; happy to clean that up in a follow-up.

Jacob0226:
@vgokhale please review again thanks

aiter/ops/triton/quant/fused_fp8_quant.py
vgokhale: This AI generated response does not do what I requested. Please re-read what I said and respond with a human in the loop understanding the intent of my comment.

vgokhale (approved):

PR #3046, opened 2026-05-06
Add nhead128,1 mask=1 to support mtp < 4, nhead128,4 fold to m16x4 kernel

fangche123:
## Motivation

Multi-token prediction (MTP > 0) on gfx950 with nhead=128, fp8 requires causal masking during decode to ensure tokens only attend to valid preceding positions. Previously, the qh128 persistent kernel only had the non-causal (mask=0) variant, which produced incorrect results when max_seqlen_q > 1.

## Technical Details

1.Dispatch logic (asm_mla.cu): For gfx950 fp8/fp8 with gqa_ratio=128 and max_seqlen_q > 1, set causal = 1 to route to the masked kernel variant.

2.New kernel binaries:

mla_a8w8_qh128_m32x4_n16x2_msk1_ps.co — causal mask, no LSE output (regenerated with LSE_FINAL_STORE=0)
mla_a8w8_qh128_m32x4_n16x2_msk1_lse_ps.co — causal mask, with LSE output (LSE_FINAL_STORE=1)
3.CSV update (mla_asm.csv): Added two entries for causal=1 (lse=0 and lse=1) to enable kernel lookup.

## Test Plan

python3 op_tests/test_mla_persistent.py -n 128,2 -d fp8 -kvd fp8 -b 32 64 128 256 -c 8192
python3 op_tests/test_mla_persistent.py -n 128,1 128,2 128,3 128,4-d fp8 -kvd fp8 -b 1 32 64 -c 1024 8192 16384

## Test Result

image
python3 op_tests/test_mla_persistent.py -n 128,1 128,2 128,3 128,4-d fp8 -kvd fp8 -b 1 32 64 -c 1024 8192 16384
image

## Submission Checklist


hubertlu-tw:
@fangche123 thank you very much for the fix. I have confirmed that the gsm8k accuracy in SGLang with DP enabled has recovered.
CC: @kkHuang-amd

shengnxu (approved):

PR #3050, opened 2026-05-06
Fix CK MoE split-k dispatch for blockscale FP8

kliuae:
## Motivation
In `ck_moe_stage1`, split-k kernels write to a separate, differently sized buffer from the module output. However, currently the blockscale quant type predicate for split-k is compared through object identity rather than values, and may always be evaluated to False as `QuantType` is not a strict python enum.
The downstream `cmdGenFunc_ck_moe_stage` also determines its own split-k conditions based on name matching from tuned configs. While this may be fine when tuned configs are not used as the non-splitk versions are always chosen, when the split-k kernels are selected from tuned configs through name matching, the split-k kernels would consume wrong-sized buffers, resulting in memory access faults.

## Technical Details
This PR replaces the `QuantType` comparison with value equality in the detection of CK MoE stage1's split-k so that the correct buffer can be allocated for the blockscale splitk paths.

## Submission Checklist

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -1828,7 +1828,7 @@ def ck_moe_stage1(
     dtype=None,
 ):
     token_num = hidden_states.shape[0]
-    is_splitk = quant_type is aiter.QuantType.per_1x128 and splitk > 1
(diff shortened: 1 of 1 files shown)


valarLip (approved):
lgtm

PR #3070, opened 2026-05-07
Fix dsink bf16 noise in Triton MHA one-kernel backward

kyle-256:
## Motivation

[ROCM-23870](https://amd-hub.atlassian.net/browse/ROCM-23870): `flash_attn_onekernel_backward` returns wrong `dsink` for sink + sliding-window with large enough Q. `out`/`dq`/`dk`/`dv` are fine — only `dsink` fails the test tolerance.

Failing shapes (bf16, causal, GQA 28/4, head_dim=128, S=1024):
```
B=4, window_left=32 → dsink atol=5e-2 fails
B=2, window_left=64 → dsink atol=5e-2 fails
```

## Technical Details

`dsink_partial = sum_t (-p_sink(t) * delta(t))` where `delta = sum_d do_d * o_d`.

Old path loaded `delta` from `_bwd_preprocess`, which reads bf16 `o`. The per-token bf16 round-trip is harmless for `dq/dk/dv` (each `delta(t)` is multiplied by `p` and contributes per-(q,k); bias cancels), but `dsink` sums `delta(t)` directly across `batch * seqlen_q` per head — bias accumulates linearly past tolerance.

Fix: recompute `delta = sum_j(p_ij * dp_ij)` inside `_bwd_dq_inner` from the fp32 `p`/`dp` already materialized for `dq`, gated on a new `RECOMPUTE_DELTA` constexpr (= `ENABLE_SINK`). Move the `dsink` atomic_add to after the masked + unmasked dq inner loops so it sees the fully-accumulated value. Same change in `bwd_kernel_causal` and `bwd_kernel_noncausal`.

Non-sink path is unchanged (constexpr-gated, same code at compile time).

## Test Plan

Standalone reproducer from the ticket on MI300X / gfx942 / bf16.

## Test Result

Before:
```
window32_batch4 dsink allclose(5e-2, 5e-2) = False
window64_batch2 dsink allclose(5e-2, 5e-2) = False
```

After:
```
window32_batch4 dsink allclose(5e-2, 5e-2) = True
window64_batch2 dsink allclose(5e-2, 5e-2) = True
```

## Submission Checklist

[ROCM-23870]: https://amd-hub.atlassian.net/browse/ROCM-23870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aiter/ops/triton/_triton_kernels/attention/mha_onekernel_bwd.py
+++ b/aiter/ops/triton/_triton_kernels/attention/mha_onekernel_bwd.py
@@ -345,7 +345,6 @@ def _bwd_dkdv_inner(
 def _bwd_dq_inner(
     dq,  # output
     dq_pe,  # optional output, pass None for non-PE case
-    delta_recomp,
     q,
     q_pe,
     K,
@@ -393,7 +392,7 @@ def _bwd_dq_inner(
     DEBUG_TRITON: tl.constexpr,
     DEBUG_TRITON_DETAIL: tl.constexpr,
     SLIDING_WINDOW: tl.constexpr,
-    RECOMPUTE_DELTA: tl.constexpr,
+    ENABLE_SINK: tl.constexpr,
 ):
     # if HEAD_DIM is padded
     PADDED_HEAD: tl.constexpr = ACTUAL_HEAD_DIM != HEAD_DIM
@@ -419,6 +418,7 @@ def _bwd_dq_inner(
     curr_philox_offset = batch_philox_offset
     curr_dropout_offset = dropout_offset
     RCP_LN2: tl.constexpr = 1.4426950408889634  # = 1.0 / ln(2)
+    delta_recomp = tl.zeros([BLOCK_M2], dtype=tl.float32)
     for blk_idx in range(num_steps):
         if DEBUG_TRITON:
             print(f"iter {blk_idx}: curr_n = {curr_n}")  # noqa: E701
@@ -497,7 +497,7 @@ def _bwd_dq_inner(
             dp = tl.dot(do, vT)
         if ENABLE_DROPOUT:
             dp = tl.where(dropout_mask, dp, 0.0) * dropout_scale
-        if RECOMPUTE_DELTA:
+        if ENABLE_SINK:
             # Equal to do_i . o_i but skips the bf16 round-trip through forward `o`.
             delta_recomp += tl.sum(p * dp, axis=1)
         delta_i = Di[:, None]
@@ -1112,11 +1112,9 @@ def bwd_kernel_causal(  # grid = (tl.cdiv(max_seqlen_q // BLOCK_M2), batch, nhea
                 dq_pe = tl.zeros([BLOCK_M2, PE_HEAD_DIM], dtype=tl.float32)
             else:
                 dq_pe = dq  # Couldn't assign None to dq_pe because _bwd_dq_inner can't return None.
-            delta_recomp = tl.zeros([BLOCK_M2], dtype=tl.float32)
-            dq, dq_pe, delta_recomp = _bwd_dq_inner(
+            dq, dq_pe, delta_recomp_masked = _bwd_dq_inner(
                 dq,  # output tensor
                 dq_pe,  # optional output tensor
-                delta_recomp,
                 q,
                 q_pe,
                 K,
@@ -1162,7 +1160,7 @@ def bwd_kernel_causal(  # grid = (tl.cdiv(max_seqlen_q // BLOCK_M2), batch, nhea
                 DEBUG_TRITON=DEBUG_TRITON,
                 DEBUG_TRITON_DETAIL=DEBUG_TRITON_DETAIL,
                 SLIDING_WINDOW=SLIDING_WINDOW,
-                RECOMPUTE_DELTA=ENABLE_SINK,
+                ENABLE_SINK=ENABLE_SINK,
             )
             end_n -= num_steps * MASK_BLOCK_N2
             window_start_n = 0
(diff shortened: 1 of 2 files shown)


brunomazzottiamd:
Added `ci:triton-300x` label to run unit tests in `gfx942`. Currently this is opt-in.

aiter/ops/triton/_triton_kernels/attention/mha_onekernel_bwd.py
```diff
@@ -495,6 +497,9 @@ def _bwd_dq_inner(
             dp = tl.dot(do, vT)
         if ENABLE_DROPOUT:
             dp = tl.where(dropout_mask, dp, 0.0) * dropout_scale
+        if RECOMPUTE_DELTA:
```
brunomazzottiamd: Why do we need the new `RECOMPUTE_DELTA` `tl.constexpr` if it's always equal to `ENABLE_SINK`? It is possible to use `ENABLE_SINK` here? Do you foresee any use case in which `RECOMPUTE_DELTA` isn't equal to `ENABLE_SINK`?

aiter/ops/triton/_triton_kernels/attention/mha_onekernel_bwd.py
```diff
@@ -1117,9 +1112,11 @@ def bwd_kernel_causal(  # grid = (tl.cdiv(max_seqlen_q // BLOCK_M2), batch, nhea
                 dq_pe = tl.zeros([BLOCK_M2, PE_HEAD_DIM], dtype=tl.float32)
             else:
                 dq_pe = dq  # Couldn't assign None to dq_pe because _bwd_dq_inner can't return None.
-            dq, dq_pe = _bwd_dq_inner(
+            delta_recomp = tl.zeros([BLOCK_M2], dtype=tl.float32)
```
brunomazzottiamd: `delta_recomp` is unused if `RECOMPUTE_DELTA` is `False`. Are you sure that the compiler is eliminating `delta_recomp` in this case?

Wouldn't it be more prudent to do something like this?

```python
if RECOMPUTE_DELTA:
    delta_recomp = tl.zeros([BLOCK_M2], dtype=tl.float32)
else:
    delta_recomp = None
```

We may face some issues with this approach... In the past, I was unable to return a `None` value from `_bwd_dq_inner` JIT function, just pass `None` as an input argument.

brunomazzottiamd: This is also applicable to other instances of `delta_recomp = tl.zeros([BLOCK_M2], dtype=tl.float32)`.

brunomazzottiamd (commented):
Hi @kyle-256. Thanks for your PR!

Generally speaking, it LGTM! I've left some comments, but no blockers. I can approve the PR when CI goes green.

Can you please post some performance data comparing the old way of computing `dskink` vs. the new way of computing `dsink`? Instinctively, I believe the performance will be the same, but it's more prudent to at least conduct a quick experiment.

brunomazzottiamd:
Is it possible to add a new test case to `op_tests/triton_tests/attention/test_mha_with_sink.py`? So, we can be sure this bug won't surface again.

brunomazzottiamd:
> Added `ci:triton-300x` label to run unit tests in `gfx942`. Currently this is opt-in.

I think tests passing in `gfx942` are mandatory to merge this PR. Now I see the bug happens in MI300X.

kyle-256:
Thanks @brunomazzottiamd! Pushed `6a1f33838`:

**1. `RECOMPUTE_DELTA` → `ENABLE_SINK`** — done, dropped the extra constexpr.

**2. `delta_recomp` allocation** — moved inside `_bwd_dq_inner` so callers don't see it at all (no conditional allocation / `None` plumbing needed at the call site). Same "always allocate, constexpr-gated, trust DCE" pattern as `dq_pe = dq` already in that function. Causal kernel sums the two returns when `ENABLE_SINK`.

**3. Test coverage** — extended `test_mha_with_sink_sliding_window` parametrize with the Jira-flagged gaps: `SEQLEN=(1024,1024)`, `NUM_HEADS=(28,4)`, `WINDOW=32`. 64 → 180 cases, ~40s total.

**4. Perf** (MI300X, seqlen=1024, 28/4 heads, head_dim=128, causal):

| case | no-sink | with-sink | overhead |
|---|---|---|---|
| sw=32, batch=4 | 0.335 ms | 0.342 ms | +2.2% |
| sw=64, batch=2 | 0.358 ms | 0.366 ms | +2.4% |
| no sw, batch=4 | 0.982 ms | 0.991 ms | +1.0% |

Overhead includes all sink work (load/exp/atomic_add + new `tl.sum(p*dp)`); the new accumulation alone is negligible vs the 3 matmuls. Non-sink path unchanged.

Accuracy: dsink max-abs error **9.8e-2 → 1.3e-5** on the failing combo.

> tests passing in gfx942 are mandatory

Local gfx942: 330 sink tests + 236 MHA bwd tests pass.

brunomazzottiamd:
> Thanks @brunomazzottiamd! Pushed `6a1f33838`:
> 
> **1. `RECOMPUTE_DELTA` → `ENABLE_SINK`** — done, dropped the extra constexpr.
> 
> **2. `delta_recomp` allocation** — moved inside `_bwd_dq_inner` so callers don't see it at all (no conditional allocation / `None` plumbing needed at the call site). Same "always allocate, constexpr-gated, trust DCE" pattern as `dq_pe = dq` already in that function. Causal kernel sums the two returns when `ENABLE_SINK`.
> 
> **3. Test coverage** — extended `test_mha_with_sink_sliding_window` parametrize with the Jira-flagged gaps: `SEQLEN=(1024,1024)`, `NUM_HEADS=(28,4)`, `WINDOW=32`. 64 → 180 cases, ~40s total.
> 
> **4. Perf** (MI300X, seqlen=1024, 28/4 heads, head_dim=128, causal):
> 
> case	no-sink	with-sink	overhead
> sw=32, batch=4	0.335 ms	0.342 ms	+2.2%
> sw=64, batch=2	0.358 ms	0.366 ms	+2.4%
> no sw, batch=4	0.982 ms	0.991 ms	+1.0%
> Overhead includes all sink work (load/exp/atomic_add + new `tl.sum(p*dp)`); the new accumulation alone is negligible vs the 3 matmuls. Non-sink path unchanged.
> 
> Accuracy: dsink max-abs error **9.8e-2 → 1.3e-5** on the failing combo.
> 
> > tests passing in gfx942 are mandatory
> 
> Local gfx942: 330 sink tests + 236 MHA bwd tests pass.

Amazing work!

brunomazzottiamd (approved):
LGTM! Let's merge it once CI passes.

brunomazzottiamd:
CI is green, but we can't merge due to some unresolved discussions created by Copilot. I think we're good to merge and Copilot suggestions don't make sense at all. I'll resolve them and merge the PR.

PR #3116, opened 2026-05-11
perf(moe_gemm_a8w4): tune block_m=32 & block_m=128 config for MI355X / gpt-oss-120b

JohnQinAMD:
3-line change in `get_kernel_config()`: switches `num_warps = 8 → 4` for `block_m ∈ {32, 128}` on gfx950. These are the routing block sizes selected by `routing.block_m = max(16, min(next_pow_2(tokens_per_expt), 128))` for gpt-oss-120b (E=128, top_k=4): **block_m=32** at batch=1024, **block_m=128** at every batch ≥ 4096 (the dominant chunked-prefill case). Non-gfx950 and `block_m ∈ {16, 64}` paths are byte-for-byte unchanged.

```diff
     elif block_m == 32:
         if n <= 1024:
             block_n = 128
             num_warps = 4
         elif n <= 4096:
             block_n = 256
-            num_warps = 8
+            num_warps = 4 
         else:
             block_n = 512
-            num_warps = 8
+            num_warps = 4

     else:
         block_n = 512
-        num_warps = 8
+        num_warps = 4 if (block_m == 128) else 8
```

> Gated on `block_m == 128` (not the full outer `else`) because `num_warps=4` regresses at `block_m=64` by ~7% on this shape. The outer `else` handles both 64 and 128.

## Microbench (MI355X gfx950, `bench_moe_gemm_a8w4.py --num-weight-inits=10`)

```
python bench_moe_gemm_a8w4.py --shape 2880 5760 --experts 128 4 --op-regex ".*<layer>.*" --num-weight-inits=10
```

| batch | block_m (routing) | layer | Pre kernel µs (nw=8) | After kernel µs (nw=4) | Δ |
|---:|:---:|---|---:|---:|---:|
| 1024 | 32  | layer1 (gate_up, N=5760) | 514.30  | 499.14  | +2.95% |
| 1024 | 32  | layer2 (down,    N=2880) | 261.29  | 239.65  | +8.3% |
| 4096 | 128 | layer1                   | 2134.27 | 1060.24 | **+50.3% (2.01×)** |
| 4096 | 128 | layer2                   | 1124.46 | 545.38  | **+51.5% (2.06×)** |
| 8192 | 128 | layer1                   | 3392.18 | 1638.85 | **+51.7% (2.07×)** |
| 8192 | 128 | layer2                   | 1761.58 | 836.27  | **+52.5% (2.11×)** |

At batches where `block_m ∈ {16, 64}` (batches 1–256 and 1088–2048 in the default M sweep), pre and after are identical within ±2% noise — those branches are not touched.

## End-to-end (vLLM gpt-oss-120b MXFP4, MI355X TP=1)

Same image (`vllm-rocm:v1-rc5`), same env, only this file toggled:

| ISL   | OSL  | conc | Pre TTFT (ms) | After TTFT (ms) | **Δ TTFT** | output_tps           |
|------:|-----:|-----:|--------------:|----------------:|-----------:|---------------------:|
| 1000  | 100  | 4    | 139.9         | 93.9            | **−32.9%** | 382 → 423 (+10.6%)   |
| 1000  | 100  | 8    | 235.8         | 136.5           | **−42.1%** | 1028 → 1093 (+6.3%)  |
| 10000 | 1000 | 4    | 923.8         | 521.9           | **−43.5%** | 700 → 750 (+7.1%)    |
| 10000 | 1000 | 8    | 1484.2        | 829.5           | **−44.1%** | 1040 → 1153 (+10.9%) |
| 10000 | 1000 | 16   | 2012.9        | 1083.7          | **−46.2%** | 1413 → 1632 (+15.5%) |
| 10000 | 1000 | 64   | 1804.6        | 1008.4          | **−44.1%** | 2257 → 2887 (+27.9%) |

ITL unchanged (±2%); decode hits `block_m=16` which is not touched.

### DIFF
--- a/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
+++ b/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
@@ -94,14 +94,14 @@ def get_kernel_config(m, n, k, routing_data):
             num_warps = 4
         elif n <= 4096:
             block_n = 256
-            num_warps = 4 if arch == "gfx950" else 8
+            num_warps = 4
         else:
             block_n = 512
-            num_warps = 4 if arch == "gfx950" else 8
+            num_warps = 4
 
     else:
         block_n = 512
-        num_warps = 4 if (block_m == 128 and arch == "gfx950") else 8
+        num_warps = 4 if block_m == 128 else 8
 
     ret = {
         "block_m": block_m,
--- a/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
+++ b/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
@@ -89,15 +89,30 @@ def get_kernel_config(m, n, k, routing_data):
             grid = grid_m * grid_n * split_k
 
     elif block_m == 32:
-        if n <= 1024:
-            block_n = 128
-            num_warps = 4
-        elif n <= 4096:
-            block_n = 256
-            num_warps = 8
+        if arch == "gfx950":
+            group_m = 4 if m >= 4096 else 1
+            if n <= 1024:
+                block_n = 128
+                num_warps = 4
+            elif n <= 4096:
+                block_n = 256
+                num_warps = 4
+            elif m < 1024:
+                block_n = 128
+                num_warps = 4
+            else:
+                block_n = 512
+                num_warps = 4
         else:
-            block_n = 512
-            num_warps = 8
+            if n <= 1024:
+                block_n = 128
+                num_warps = 4
+            elif n <= 4096:
+                block_n = 256
+                num_warps = 8
+            else:
+                block_n = 512
+                num_warps = 8
 
     else:
         block_n = 512
--- a/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
+++ b/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
@@ -101,6 +101,8 @@ def get_kernel_config(m, n, k, routing_data):
 
     else:
         block_n = 512
+        # routing caps block_m at 128; nw=4 wins ~2x at block_m=128 on gpt-oss
+        # shapes (MI355X) but regresses ~7% at block_m=64, so 64 stays at 8.
         num_warps = 4 if block_m == 128 else 8
 
     ret = {
--- a/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
+++ b/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
@@ -89,34 +89,19 @@ def get_kernel_config(m, n, k, routing_data):
             grid = grid_m * grid_n * split_k
 
     elif block_m == 32:
-        if arch == "gfx950":
-            group_m = 4 if m >= 4096 else 1
-            if n <= 1024:
-                block_n = 128
-                num_warps = 4
-            elif n <= 4096:
-                block_n = 256
-                num_warps = 4
-            elif m < 1024:
-                block_n = 128
-                num_warps = 4
-            else:
-                block_n = 512
-                num_warps = 4
+        if n <= 1024:
+            block_n = 128
+            num_warps = 4
+        elif n <= 4096:
+            block_n = 256
+            num_warps = 4 if arch == "gfx950" else 8
         else:
-            if n <= 1024:
-                block_n = 128
-                num_warps = 4
-            elif n <= 4096:
-                block_n = 256
-                num_warps = 8
-            else:
-                block_n = 512
-                num_warps = 8
+            block_n = 512
+            num_warps = 4 if arch == "gfx950" else 8
 
     else:
         block_n = 512
-        num_warps = 8
+        num_warps = 4 if (block_m == 128 and arch == "gfx950") else 8
 
     ret = {
         "block_m": block_m,


aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,27 @@ def get_kernel_config(m, n, k, routing_data):
             grid = grid_m * grid_n * split_k
 
     elif block_m == 32:
+        # Tuned on MI355X (gfx950) for MoE shapes with K=2880, N in {2880, 5760}
+        # (gpt-oss-120b gate_up + down). Microbench wins vs the prior config:
+        #   * num_warps=4 over num_warps=8 at N>1024: +8.6% (K=2880, all M).
+        #   * For small-M / large-N (m<1024, n>4096), block_n=128 over
+        #     block_n=512: +6.2% (M=256, N=5760 verified).
```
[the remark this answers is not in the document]
JohnQinAMD: Addressed in 23585f7cd: the new `block_m == 32` heuristic is now gated on `arch == "gfx950"`, and the non-gfx950 path falls back to the exact pre-PR (num_warps=8 / group_m=4) config so behavior on other architectures is byte-for-byte unchanged.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,27 @@ def get_kernel_config(m, n, k, routing_data):
...
+        #   * group_m=1 over group_m=4 for m<4096: +3-8% across the swept
+        #     M range; group_m=4 still wins at m>=4096.
+        # See PR description for the full sweep table.
```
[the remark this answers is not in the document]
JohnQinAMD: Addressed in 71aa19350 (kept in 23585f7cd): the inline comment block with microbench deltas and the PR-description reference was removed. The patch is now config values only; rationale lives in the commit message.

brunomazzottiamd (commented):
LGTM! I've requested a review from our MoE experts!

azaidy:
> ## Summary
> Retunes `get_kernel_config()` in `aiter/ops/triton/moe/moe_op_gemm_a8w4.py` for the `block_m == 32` branch on MI355X (gfx950) — the branch selected for gpt-oss-120b (K=2880, N ∈ {2880, 5760}). Scoped to `block_m == 32` only; other branches are byte-for-byte unchanged.
> 
> Param	Before	After	Microbench gain
> `num_warps` (n > 1024)	8	4	+8.6%
> `block_n` (m < 1024, n > 4096)	512	128	+6.2%
> `group_m`	4	1 if m<4096 else 4	+3–8% small/mid M
> ## End-to-end
> vLLM serving gpt-oss-120b, MI355X, MXFP4 + FP8, ISL=10000 OSL=1000 conc=4: **TTFT 920 ms → 518 ms (−43.6%)**. ITL unchanged.
> 
> ## Risk
> Low — tuning-only change (no kernel code touched), and only the `block_m == 32` arm of the heuristic is modified. Existing tests in `op_tests/triton_tests/moe/test_moe_gemm_a8w4.py` cover this code path with no behavioral assumptions on `block_n`/`num_warps`/`group_m`.

@JohnQinAMD 
So the multiplicative gain here is 1.086*1.062*1.08 = 1.25x approx. are you saying that led to 920/518 = 1.797x gain on end 2 end. seems like there are cache warmup effects that are being accounted for here.

nsusanto (dismissed):
LGTM!

nsusanto:
@JohnQinAMD I ran the microbenchmark on 950 with the same shape from the PR but I am not seeing as big of a performance gain as what you got. Can you explain how the microbench numbers in the PR were collected? 

block_m=32 only applies to batch=1024. FYI use `--num-weight-inits` flag to set the number of iterations to average out noise
(N=5760)
`python bench_moe_gemm_a8w4.py --shape 2880 5760 --experts 128 4 --op-regex ".*layer1.*" --num-weight-inits=10` 
Before patch:
batch:  1024 | Total latency (us): 805.56 | Kernel latency (us): 510.53 | TFLOPS: 266.2 | TBPS: 2.11

After patch
batch:  1024 | Total latency (us): 767.43 | Kernel latency (us): 497.66 | TFLOPS: 273.1 | TBPS: 2.16

(N=2880)
`python bench_moe_gemm_a8w4.py --shape 2880 5760 --experts 128 4 --op-regex ".*layer2.*" --num-weight-inits=10` 
Before patch:
batch:  1024 | Total latency (us): 802.36 | Kernel latency (us): 262.89 | TFLOPS: 258.5 | TBPS: 2.15

After patch:
batch:  1024 | Total latency (us): 764.55 | Kernel latency (us): 237.37 | TFLOPS: 286.2 | TBPS: 2.39

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,30 @@ def get_kernel_config(m, n, k, routing_data):
             grid = grid_m * grid_n * split_k
 
     elif block_m == 32:
-        if n <= 1024:
-            block_n = 128
-            num_warps = 4
-        elif n <= 4096:
-            block_n = 256
-            num_warps = 8
+        if arch == "gfx950":
```
ChuanLi1101: Minor refactor: every sub-case inside the `gfx950` branch sets `num_warps = 4`, so it can be hoisted out and the body collapses to just `block_n` selection. Easier to see at a glance what actually differs from the legacy path (just `num_warps` and the large-`n` / small-`m` `block_n`):

```suggestion
        if arch == "gfx950":
            num_warps = 4
            group_m = 4 if m >= 4096 else 1
            if n <= 1024:
                block_n = 128
            elif n <= 4096:
                block_n = 256
            elif m < 1024:
                block_n = 128
            else:
                block_n = 512
```

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,30 @@ def get_kernel_config(m, n, k, routing_data):
...
+            group_m = 4 if m >= 4096 else 1
```
ChuanLi1101: Hard cliff at `m == 4096`: this flips `group_m` from 1 to 4, and combined with the `m < 1024` branch below it flips `block_n` between 128 and 512. With vLLM dynamic batching the routed-token count `m` can oscillate around this boundary across consecutive prefills, so two very-close `m` values may pick very different configs.

Could you sweep `m` in `[3500, 4500]` (step ~128) for the target shapes (K=2880, N in {2880, 5760}) and confirm there's no >10% regression at the boundary? If there is, either move the threshold or add an intermediate config (e.g. `group_m=2`) for the transition region.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,30 @@ def get_kernel_config(m, n, k, routing_data):
...
+            if n <= 1024:
+                block_n = 128
+                num_warps = 4
+            elif n <= 4096:
+                block_n = 256
+                num_warps = 4
+            elif m < 1024:
```
ChuanLi1101: Two things on this branch:

1. **Readability.** `elif m < 1024:` implicitly relies on `n > 4096` from the chain above. Reads cleaner as an explicit 2-D condition:

   ```python
   elif n > 4096 and m < 1024:
       block_n = 128
   else:  # n > 4096, m >= 1024
       block_n = 512
   ```

2. **Cliff concern.** Same issue as line 93: `m == 1024` is a hard switch from `block_n=128` to `block_n=512` (a 4x jump). Please verify `m` in `[800, 1300]` doesn't regress on the target shapes before merge.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -89,15 +89,30 @@ def get_kernel_config(m, n, k, routing_data):
...
+                block_n = 128
+                num_warps = 4
+            else:
+                block_n = 512
+                num_warps = 4
         else:
```
ChuanLi1101, line 98: Worth a one-line in-file comment here to make the intent explicit, e.g.:

```python
        else:
            # Non-gfx950: preserve pre-PR block_m == 32 config byte-for-byte.
            if n <= 1024:
                ...
```

Otherwise a future reader sees a duplicated chain and may not realize it's deliberately a frozen legacy path — and the next person tuning gfx942 will benefit from knowing this is the baseline to beat.

ChuanLi1101 (commented):
Direction is good — arch-gated, scope tightly limited to `block_m == 32` on gfx950, legacy paths preserved. A few things to address before merge:

**1. Microbench numbers don't reconcile (please re-run with `--num-weight-inits=10`).**
@nsusanto's repro on the same shape only shows ~2.5% (layer1, N=5760) / ~9.7% (layer2, N=2880) kernel speedup, which doesn't match the +8.6 / +6.2 / +3–8% deltas in the PR description. And as @azaidy pointed out, the multiplicative microbench gain (~1.25x) cannot explain the 1.797x e2e TTFT improvement (920 -> 518 ms). Could you:
  - Re-run microbench with `--num-weight-inits=10` and paste the raw before/after logs;
  - If the e2e -43.6% TTFT is real, attach a per-kernel breakdown (rocprofv3 ATT or vLLM per-layer trace) that isolates this kernel as the source of the prefill speedup.
As-is the headline numbers don't reconcile, which makes it hard to trust the tuning generalizes.

**2. Threshold cliffs at `m=1024` and `m=4096`.** See inline comments on lines 93 and 100. Both thresholds are hard switches between very different config families (`block_n` and `group_m` both flip), which is risky for dynamic batching where `m` oscillates around them. Please sweep `m` in `[800, 1300]` and `[3500, 4500]` and confirm there's no regression near the boundaries.

**3. CI coverage.** `op_tests/triton_tests/moe/test_moe_gemm_a8w4.py` exercises `block_m == 32`, but only the `arch == "gfx950"` branch contains new logic — the MI300X (gfx942) job will exercise the unchanged `else` path. Please confirm the MI35X Triton CI job actually executed the new gfx950 sub-branch and link to it in the PR description.

**4. Code clarity.** See inline refactor suggestion on line 92 and the legacy-preservation note on line 106.

Scope itself (block_m==32 only, fully arch-gated, legacy path byte-for-byte preserved) is appropriately conservative — once (1) is resolved I'm OK with this approach.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Simplified further in ef0e7f0b2 — the entire `if arch == "gfx950": ... else: ...` block inside `block_m == 32` is collapsed into a single inline `num_warps = 4 if arch == "gfx950" else 8` ternary on the two `n > 1024` sub-cases. There are no sub-cases left to hoist from; block_m=32's branch structure is now identical to pre-PR except for those two ternaries.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Simplified in ef0e7f0b2 — the m-dependent `group_m = 4 if m >= 4096 else 1` is removed; `group_m` stays at the constant default 4 across all m. The cliff at m=4096 no longer exists.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Simplified in ef0e7f0b2 — the `elif m < 1024:` if/else branch is removed entirely. The cliff at m=1024 no longer exists, and the implicit `n > 4096` dependency goes away with it.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Resolved in ef0e7f0b2 — there's no longer a separate non-gfx950 else inside `block_m == 32`. Gating is now an inline ternary (`num_warps = 4 if arch == "gfx950" else 8`), so non-gfx950 falls back to `num_warps = 8` exactly as pre-PR without a duplicate code block to annotate.

JohnQinAMD:
@azaidy  the multiplicative math doesn't work, and the original PR body's framing was misleading. The wins don't multiply because each parameter only fires in specific (m, n) regimes, and at the only batch where `block_m == 32` is selected in production (batch=1024) the multiplicative components reduce to a single ~+5% per-call kernel win, not 1.25×.


The PR body is updated with the corrected per-shape numbers with `--num-weight-inits=10`  and the E2E A/B across the different workload set.

JohnQinAMD:
@ChuanLi1101  Thanks for the careful review — addressed all three comments:

**1. Microbench reconcile (`--num-weight-inits=10`).** Re-ran on MI355X; @nsusanto's reproduction was correct. At production batch=1024 (the only batch where block_m=32 is selected by routing), the kernel win is ~+5% combined (+2.95% layer1, +8.3% layer2), not +8.6%. The headline +8.6 / +6.2 / +3–8% numbers were from a custom microbench script that overrode `routing_data.block_m=32` at M values where production routing picks `block_m=16` (a bug in the bench script). Raw logs in the updated PR body.

**2. Threshold cliffs at m=4096 and m=1024.** Both cliffs are removed by ef0e7f0b2 — the m-dependent `group_m` and the `m < 1024` block_n override are gone (there are headroom when tune 3 parameters together, but tune combinations need more sweep and tests.). Microbench sweep at batch ∈ [768, 1280] confirms no regression across the original cliff regions.

**3. Per-kernel breakdown isolating the source of the 43% TTFT win.** Post-patch prefill trace (eager / cudagraph_mode=NONE for kernel visibility, conc=4 ISL=10K): `_moe_gemm_a8w4` accounts for **29.4% of GPU time** (854 calls × avg 1334 µs). All 854 calls hit `block_m=128` (tokens_per_expt = 3454*4/128 = 108 → next_pow_2 = 128 → routing cap). Pre-patch the same calls take 2× longer (verified by microbench at block_m=128 across M ∈ {4096, 8192}). Clean A/B end-to-end measurement today (same image, same env, only this file toggled): TTFT 923.8 → 521.9 ms at conc=4 ISL=10K (−43.5%), matches the original claim.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
vgokhale: You should remove the conditional entirely. "non-gfx950" is just gfx942. And 942 has no support for fp4, let along mixed fp8 x fp4. It is going to upcast in all cases and using fp8 will result in strictly worse perf because it now has to upcast two inputs instead of just one if using bf16 x fp4. 

So this should be statically set to 4.

JohnQinAMD: @vgokhale  Thanks — good point, applied in fde465b40. Dropped the `arch == "gfx950"` conditional from the `block_m == 32` branches (both n>1024 sub-cases now just set `num_warps = 4` unconditionally). The cumulative diff vs main is now 3 plain constant flips plus one `block_m == 128` ternary, no arch checks anywhere.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -94,14 +94,14 @@ def get_kernel_config(m, n, k, routing_data):
             num_warps = 4
         elif n <= 4096:
             block_n = 256
-            num_warps = 8
+            num_warps = 4
```
ChuanLi1101, line 97: Per @vgokhale this is now arch-agnostic (was `4 if arch == "gfx950" else 8`), so `num_warps = 4` becomes the permanent default on gfx942 too. The reasoning (gfx942 has no native FP8×FP4 MFMA, the path is upcast-bound, nw choice is unmeasurable) is plausible, but PR data is MI355X-only. Same flip on the `n > 4096` branch below (line 100) — both rely on the same assumption.

Could you post the `ci:triton-300x` MI300X result link (or a quick one-shape A/B on gfx942 at `K=2880, N ∈ {2880, 5760}, batch=1024`) once CI finishes? Just so the "no regression on 942" claim is pinned to an actual run rather than a theoretical argument.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -94,14 +94,14 @@ def get_kernel_config(m, n, k, routing_data):
...
         else:
             block_n = 512
-            num_warps = 8
+            num_warps = 4
```
ChuanLi1101, line 100: Minor: this `n > 4096` sub-branch isn't hit by gpt-oss-120b in production (its `N ∈ {2880, 5760}` are both ≤ 4096) — at production batch=1024 the call lands on the `elif n <= 4096` branch above. So this flip is forward-looking for future shapes rather than load-bearing today; that's fine, just worth knowing the e2e TTFT win in the PR body comes entirely from line 104 (block_m=128 path) and line 97 — not this line.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -94,14 +94,14 @@ def get_kernel_config(m, n, k, routing_data):
...
 
     else:
         block_n = 512
-        num_warps = 8
+        num_warps = 4 if block_m == 128 else 8
```
ChuanLi1101, line 106: Two things on this line:

**1. Please add a one-line in-file comment for the `block_m == 128` selector.** Right now the rationale lives only in the commit message — a future reader sees a bare `4 if block_m == 128 else 8` and won't know that (a) routing caps at `block_m = max(16, min(next_pow_2(tokens_per_expt), 128))` so the only other value reaching this `else` in practice is `block_m == 64`, and (b) `num_warps = 4` was measured at `block_m == 64` on gpt-oss shapes and regressed ~7%. Suggested:

```python
else:
    block_n = 512
    # routing caps block_m at 128; nw=4 wins ~2x at block_m=128 on gpt-oss
    # shapes (MI355X) but regresses ~7% at block_m=64, so 64 stays at 8.
    num_warps = 4 if block_m == 128 else 8
```

Makes `git blame` self-contained; saves the next person retuning this from re-deriving the routing-cap reasoning.

**2. Same gfx942 concern as lines 97/100, but stronger here.** This ternary is also arch-agnostic now, and `block_m == 128` *is* the dominant prefill path for gpt-oss at conc≥4. The commit message argues "the regression at block_m=64 with nw=4 is independent of arch," but all data is gfx950. Please link the MI300X CI run that exercises `block_m == 128`, or attach one A/B point on gfx942 at K=2880, N∈{2880, 5760}.

ChuanLi1101 (commented):
Third pass — scope is now nicely contained: net diff is 3 constant flips + 1 `block_m == 128` ternary, and the earlier `group_m` / `m<1024` cliffs are gone. Direction LGTM. Two remaining items before approval:

1. **Add an in-file comment on the `block_m == 128` ternary** (line 104) so the magic constant `128` and the `block_m == 64` exclusion are discoverable from `git blame` instead of only the commit message.
2. **Post the MI300X (`ci:triton-300x`) CI link in the PR** once it's green. The previous `arch == "gfx950"` guards were dropped per @vgokhale's reasoning, which is plausible but means three branches now permanently default to `num_warps = 4` on gfx942 with no posted A/B. A CI link closes that out.

Once those two are addressed I'm OK to approve.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Tried to get the gfx942 A/B and the result is "not applicable" — the kernel is gfx950-only, so the `num_warps` change is a no-op on gfx942.

Attempted to run `bench_moe_gemm_a8w4.py` on MI300X with both pre- and post-patch `get_kernel_config()`. The `_moe_gemm_a8w4` Triton kernel uses [`tl.dot_scaled`](https://github.com/ROCm/aiter/blob/c016f43fc/aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_a8w4.py#L339) for the FP8×MXFP4 microscaling matmul — a CDNA4-only hardware path. It does not compile on gfx942: confirmed across Triton 3.5.1, 3.6.0, and 3.7.0, all failing to lower `tl.dot_scaled` in the AMD backend (`Unsupported DotScaleOp`).

The repo already gates this codepath on gfx950 — see [`op_tests/triton_tests/moe/test_moe_gemm_a8w4.py#L227`](https://github.com/ROCm/aiter/blob/c016f43fc/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py#L227):

```python
if get_arch() != "gfx950":
    pytest.skip("float8 x mx only supported on CDNA4")
```

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Good catch — at production batch=1024 for gpt-oss-120b (N ∈ {2880, 5760}), layer1 (N=5760) does land here (n>4096) and layer2 (N=2880) lands on the `elif n <= 4096` branch above. The microbench data in the PR body confirms this — layer1 at batch=1024 hits this branch and shows +2.95% (`block_m=32`, `bn=512`, `nw=8 → 4`); layer2 at batch=1024 hits the elif above with +8.3%.

Your point is correct: the e2e TTFT win in the PR body comes mostly from the `block_m == 128` path (line 104), not from these `block_m == 32` ternaries — the `block_m == 32` branch only fires at one specific batch value (=1024) in production, while `block_m == 128` is the dominant prefill case at any conc≥4 ISL≥1000.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: Both addressed

**(1) In-file comment** — added per your suggestion, 

```python
else:
    block_n = 512
    # routing caps block_m at 128; nw=4 wins ~2x at block_m=128 on
    # gpt-oss shapes (MI355X) but regresses ~7% at block_m=64.
```

git blame is now self-contained: future readers see why block_m=64 stays at nw=8, why the routing caps at 128, and why the arch gate is there despite being dead code today.

**(2) gfx942 test**  `tl.dot_scaled` is CDNA4-only and doesn't compile on gfx942 across Triton 3.5-3.7, so the `num_warps` change is unreachable on gfx942. The repo already gates this codepath on gfx950 — see [`test_moe_gemm_a8w4.py:L227`](https://github.com/ROCm/aiter/blob/c016f43fc/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py#L227).  Also confirmed by @vgokhale this kernel will not be enabled on gfx942 due to hardware datatype support.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
JohnQinAMD: @vgokhale  re-added the `arch == "gfx950"` guard in c016f43fc. You were correct that it's dead code today: `tl.dot_scaled` only lowers to CDNA4 (Triton 3.5/3.6/3.7 all fail on gfx942 with `Unsupported DotScaleOp` in the AMD backend), so the non-gfx950 branch is never taken. But the tuning in this PR was measured only on gfx950, and re-adding the explicit gate keeps that gfx950-specific intent self-documenting and forward-compatible with any future Triton release that adds gfx942 lowering of `tl.dot_scaled`. Cost is one ternary per branch; benefit is we don't silently apply gfx950 `num_warps` choices to a different arch.

Also addresses @ChuanLi1101's separate request to make the `block_m == 128` rationale visible to future readers — see line-104 thread.

vgokhale: > forward-compatible with any future Triton release that adds gfx942 lowering of tl.dot_scaled.

It is not possible to lower dot_scaled on gfx942 because there is no native support for these dtypes. Please remove the check. It makes no sense to use this kernel on 942.

JohnQinAMD: @vgokhale  Thanks for confirming, arch check has been removed. Thanks

ChuanLi1101 (approved):
LGTM

JohnQinAMD:
> @JohnQinAMD I ran the microbenchmark on 950 with the same shape from the PR but I am not seeing as big of a performance gain as what you got. Can you explain how the microbench numbers in the PR were collected?
> 
> block_m=32 only applies to batch=1024. FYI use `--num-weight-inits` flag to set the number of iterations to average out noise (N=5760) `python bench_moe_gemm_a8w4.py --shape 2880 5760 --experts 128 4 --op-regex ".*layer1.*" --num-weight-inits=10` Before patch: batch: 1024 | Total latency (us): 805.56 | Kernel latency (us): 510.53 | TFLOPS: 266.2 | TBPS: 2.11
> 
> After patch batch: 1024 | Total latency (us): 767.43 | Kernel latency (us): 497.66 | TFLOPS: 273.1 | TBPS: 2.16
> 
> (N=2880) `python bench_moe_gemm_a8w4.py --shape 2880 5760 --experts 128 4 --op-regex ".*layer2.*" --num-weight-inits=10` Before patch: batch: 1024 | Total latency (us): 802.36 | Kernel latency (us): 262.89 | TFLOPS: 258.5 | TBPS: 2.15
> 
> After patch: batch: 1024 | Total latency (us): 764.55 | Kernel latency (us): 237.37 | TFLOPS: 286.2 | TBPS: 2.39

@nsusanto The newly tested perf has been updated in the pr body with aligned approach. Also added more perf numbers to illustrate the perf uplift. Thanks

nsusanto:
Verifed the performance numbers. Seems good to me

Before patch
batch:     1 | Total latency (us): 50.10 | Kernel latency (us): 15.34 | TFLOPS: 4.326 | TBPS: 1.08
batch:     2 | Total latency (us): 58.93 | Kernel latency (us): 16.45 | TFLOPS: 8.067 | TBPS: 2.00
batch:     4 | Total latency (us): 84.47 | Kernel latency (us): 24.42 | TFLOPS: 10.87 | TBPS: 2.60
batch:     8 | Total latency (us): 138.17 | Kernel latency (us): 41.81 | TFLOPS: 12.70 | TBPS: 2.91
batch:    16 | Total latency (us): 211.01 | Kernel latency (us): 66.55 | TFLOPS: 15.95 | TBPS: 3.24
batch:    32 | Total latency (us): 312.73 | Kernel latency (us): 100.65 | TFLOPS: 21.10 | TBPS: 3.39
batch:    64 | Total latency (us): 413.30 | Kernel latency (us): 132.35 | TFLOPS: 32.09 | TBPS: 3.53
batch:   128 | Total latency (us): 467.46 | Kernel latency (us): 150.30 | TFLOPS: 56.51 | TBPS: 3.51
batch:   256 | Total latency (us): 477.51 | Kernel latency (us): 153.28 | TFLOPS: 110.8 | TBPS: 3.52
batch:  1024 | Total latency (us): 758.52 | Kernel latency (us): 237.18 | TFLOPS: 286.5 | TBPS: 2.39
batch:  4096 | Total latency (us): 3225.71 | Kernel latency (us): 1103.03 | TFLOPS: 246.4 | TBPS: 0.61
batch:  8192 | Total latency (us): 5153.19 | Kernel latency (us): 1749.76 | TFLOPS: 310.7 | TBPS: 0.47
 
After patch
batch:     1 | Total latency (us): 50.17 | Kernel latency (us): 15.32 | TFLOPS: 4.331 | TBPS: 1.09
batch:     2 | Total latency (us): 58.78 | Kernel latency (us): 16.39 | TFLOPS: 8.098 | TBPS: 1.95
batch:     4 | Total latency (us): 84.49 | Kernel latency (us): 24.45 | TFLOPS: 10.86 | TBPS: 2.60
batch:     8 | Total latency (us): 135.75 | Kernel latency (us): 41.48 | TFLOPS: 12.80 | TBPS: 2.86
batch:    16 | Total latency (us): 211.76 | Kernel latency (us): 66.51 | TFLOPS: 15.96 | TBPS: 3.24
batch:    32 | Total latency (us): 310.35 | Kernel latency (us): 99.35 | TFLOPS: 21.37 | TBPS: 3.40
batch:    64 | Total latency (us): 408.63 | Kernel latency (us): 131.08 | TFLOPS: 32.40 | TBPS: 3.53
batch:   128 | Total latency (us): 467.44 | Kernel latency (us): 150.43 | TFLOPS: 56.46 | TBPS: 3.51
batch:   256 | Total latency (us): 479.08 | Kernel latency (us): 153.35 | TFLOPS: 110.8 | TBPS: 3.52
batch:  1024 | Total latency (us): 727.53 | Kernel latency (us): 212.36 | TFLOPS: 320.0 | TBPS: 2.67
batch:  4096 | Total latency (us): 1648.41 | Kernel latency (us): 538.75 | TFLOPS: 504.5 | TBPS: 1.25
batch:  8192 | Total latency (us): 2555.70 | Kernel latency (us): 824.87 | TFLOPS: 659.0 | TBPS: 0.99

lburzawa (approved):

nsusanto (approved):
LGTM

vgokhale (approved):

PR #3127, opened 2026-05-11
AiterAsmKernel: add init-time sanity checks for .co registration

alexioslyrakis-amd:
## Summary

- Adds `validate_hsaco_lds()` to `AiterAsmKernelFast::init()`: scans the raw `.co` ELF blob for `group_segment_fixed_size` via msgpack decode and checks it against the device LDS limit (`hipDeviceGetAttribute`). Fires before registration, giving an actionable error instead of a silent null from `hipGetFuncBySymbol`.
- Adds a registration probe in `init()`: since `__hipRegisterFunction` returns `void`, a `hipGetFuncBySymbol` probe detects silent rejection by the runtime (LDS limit exceeded, arch mismatch, corrupted binary, etc.).
- Both checks run **once per kernel variant** at construction time — zero impact on the `launch_kernel()` hot path.

## Motivation

On gfx942 (MI300X) the LDS limit is 64 KB. A `.co` built targeting gfx950 (MI355X, ~160 KB LDS) would be silently rejected at registration, `hipGetFuncBySymbol` would return null, and `hipModuleLaunchKernel` would emit only a generic `hipErrorIllegalState` with no indication of the root cause.

## Test plan

- [x] Run with a `.co` declaring `group_segment_fixed_size > 65536` on gfx942 — expect clear error from `validate_hsaco_lds()` at init
- [x] Run with a correctly built `.co` on gfx942 — expect no error, kernel executes normally

### DIFF
--- a/csrc/include/aiter_hip_common.h
+++ b/csrc/include/aiter_hip_common.h
@@ -21,6 +21,8 @@
 #include <fstream>
 #include <mutex>
 #include <memory>
+#include <string_view>
+#include <algorithm>
 #ifdef AITER_EMBEDDED_HSA_HEADER
 #include AITER_EMBEDDED_HSA_HEADER
 #endif
@@ -194,6 +196,14 @@ class AiterAsmKernelFast
                                             nullptr,
                                             nullptr,
                                             nullptr);
+        // Verify registration succeeded. __hipRegisterFunction returns void so
+        // we probe via hipGetFuncBySymbol — if it returns null the runtime silently
+        // rejected the kernel (e.g. resource limits, arch mismatch). This runs
+        // once per kernel variant at init time, not on every launch.
+        hipFunction_t probe = nullptr;
+        (void)hipGetFuncBySymbol(&probe, reinterpret_cast<void*>(this));
+        AITER_CHECK(probe != nullptr,
+                    "kernel registration failed for '", kernel_name, "'.");
     }
 
     public:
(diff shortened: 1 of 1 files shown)


valarLip (approved):
LGTM

amd-ruitang3:
Hi @alexioslyrakis-amd,
Thanks your commit make aiter more robust. I have a question for you:
The probe can already detect the issue of hipregister in the early phase. The validate_hsaco_lds() use msgpack to parse, but msgpack is fragile, is there the midjudgments?

alexioslyrakis-amd:
> Hi @alexioslyrakis-amd, Thanks your commit make aiter more robust. I have a question for you: The probe can already detect the issue of hipregister in the early phase. The validate_hsaco_lds() use msgpack to parse, but msgpack is fragile, is there the midjudgments?

Hi @amd-ruitang3, the probe is the safety net, it catches all failures, but it can only say "registration failed", not why. `validate_hsaco_lds()` runs before registration specifically to surface the LDS root cause with the actual numbers. 

On the fragility concern: the scanner is conservative (skips anything it doesn't recognize) so false negatives are possible (it misses the check), but I can't think of how false positives might appear. I could also alternatively use LLVM's built-in msgpack parser (llvm/BinaryFormat/MsgPack.h), I guess this won't introduce a new dependency since LLVM is already required by ROCm.

PR #3132, opened 2026-05-11
[Triton] Add s_barrier to sync waves after writing expert hist

vgokhale:
Without this it is possible that wave 0 in a WG is reading addr X, wave 1 in the same WG is responsible for writing addr X, but hasn't yet arrived at its tl.store(X) while wave 0 is at its tl.load(X)


azaidy (approved):
LGTM!

PR #3145, opened 2026-05-12
add silu_and_mul_quant and Opt silu_and_mul

yzhou103:
## Motivation

Add HIP silu_and_mul group quantization support and refactor shared fp4 scale/shuffle helpers for kernel consistency 
optimization

## Technical Details

1. Non-temporal (NT) load/store for act_and_mul_kernel
Added a compile-time AUX template parameter to act_and_mul_kernel. The dispatch macro passes GROUP_NT (non-temporal hint), which tells the hardware to bypass L2 cache for both loads and stores. Since activation kernels are streaming workloads with no data reuse, skipping L2 avoids cache pollution and improves effective HBM bandwidth.
2. Vectorized NT store + NT load for scaled_act_and_mul_kernel
The original kernel wrote fp8 output with per-element scalar stores (out[idx] = ...), which generated one 1-byte store instruction per element. Replaced with:
store_vector_nbytes for vectorized stores (up to 16 bytes per store instruction)
GROUP_NT hint on both loads and stores
This reduces total store instruction count by up to 16x and avoids L2 cache thrashing.
3. Block-size specialization for scaled_silu_and_mul
Added COMPUTE_SCALED_ACTIVATION_KERNEL_PARAMS with a conditional rule: when 512 < d <= 2048 and num_tokens <= 4096, increase num_wave to 4 (256 threads per block) instead of the default 2 (128 threads). This better saturates the CU for mid-range hidden dimensions typical of DeepSeek V4 Pro TP4 (d=768) and Flash TP1 (d=2048), matching the parallelism level used by equivalent Triton kernels.

## Test Result

image

## Submission Checklist


yzhou103:
test on mi355 
image

valarLip (approved):

PR #2695, opened 2026-04-10
[Triton] Declare triton>=3.6.0 dependency 

micmelesse:
## Motivation

This PR declares aiter's dependency on triton in setup.py. This is standard practice for python libraries. We declare a minimum version of `3.6.0`. We remove a lot of ad-hoc code that was branching on triton versions.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3149, opened 2026-05-12
[Gluon] Fix MFMA instr_shape for Triton version compatibility

yanxuer-999:
Triton >= 3.6.0 requires 3D instr_shape (M, N, K) for AMDMFMALayout, while older versions use 2D (M, N). Select the shape at compile time via a constexpr derived from triton_version so the gluon GEMM and MLA decode kernels build on both.

Fixes ROCm/triton-internal#1670

## Motivation

Similar to the fix in PR2491, this addresses the remaining compatibility issues for legacy Gluon kernels.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/gluon/gemm_a8w8_blockscale.py
+++ b/aiter/ops/triton/gluon/gemm_a8w8_blockscale.py
@@ -16,6 +16,10 @@ from triton import language as tl
 _LOGGER = AiterTritonLogger()
 from triton.experimental import gluon
 from triton.experimental.gluon import language as gl
+import aiter.ops.triton.gluon.triton_version as tv
+
+# Pre-compute version check as constexpr for use in JIT kernels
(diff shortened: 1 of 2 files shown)


brunomazzottiamd (changes requested):
**Sharing for the sake of visibility:** https://github.com/ROCm/aiter/pull/2695 proposes the complete removal of Triton version checks in Gluon kernels and enforcement of Triton >= 3.6 in AITER.

I'm requesting changes just to prevent merging. I can't see anything wrong with this PR. Let's do some team work and agree on what should be done before proceeding with this PR and https://github.com/ROCm/aiter/pull/2695.

vgokhale:
I think this is effectively saying we will support older triton versions, which we don't want to do. So I am closing this PR (I also closed the corresponding issue on triton-internal).

PR #3161, opened 2026-05-13
Update triton wheel

mengfei-jiang:
## Motivation

This PR addresses two issues discovered during Triton 3.7.0 compatibility testing:

  1. torch.compile constexpr mutation regression — Triton 3.7.0 introduced a behavior change where tl.constexpr arguments in Triton kernels can be mutated during compilation, causing incorrect results when the same kernel is invoked multiple times with different parameters under torch.compile. A regression test (test_compile_constexpr_mutation.py) is added to catch this. and we update triton wheel for this issue.
  2. Triton reinstall breaks older PyTorch — When torch < 2.9.1 is installed, reinstalling triton via install_triton.sh during setup.py can pull an incompatible version, breaking the environment. A guard is added to skip triton reinstall when torch < 2.9.1 is detected.

## Technical Details

  1. setup.py — Skip triton reinstall for torch < 2.9.1

  Added _torch_version_below("2.9.1") guard before the triton install logic. When an older PyTorch is detected, setup.py prints a message and preserves the existing triton installation instead of invoking install_triton.sh, which could pull a wheel incompatible with the older torch ABI.

  2. test_compile_constexpr_mutation.py — Regression test

  Adds a test that invokes a tl.constexpr-parameterized RMSNorm kernel multiple times with different N values under torch.compile. In Triton 3.7.0, a bug causes the constexpr value from the first invocation to leak into subsequent invocations, producing incorrect results. The test asserts output correctness across multiple N values to catch this regression.

## Test Plan

Add test_compile_constexpr_mutation.py for regression test.

## Submission Checklist


README.md
```diff
@@ -115,7 +115,7 @@ pip install -r requirements.txt
 
 ### Triton
 
-AITER includes Triton-based operators that require triton from AMD PyPI ([ROCm 7.0](https://pypi.amd.com/triton/release/rocm-7.0.0/simple/), [ROCm 7.1](https://pypi.amd.com/triton/release/rocm-7.1.0/simple/), [ROCm 7.2](https://pypi.amd.com/triton/release/rocm-7.2.0/simple/)), with the correct version selected based on your ROCm installation.
+AITER includes Triton-based operators that require triton from AMD PyPI, with the correct version selected based on your ROCm installation.
```
Dewei-Wang-sh, line 118: add the env var usage later.

setup.py
```diff
@@ -97,8 +97,26 @@ def _run_install_triton():
...
+        return Version(torch.__version__.split("+")[0].split("dev")[0]) < Version(
+            min_version
+        )
+    except Exception:
+        return False
+
+
 _triton_info = _is_triton_installed()
-if AITER_USE_SYSTEM_TRITON and _triton_info:
+if _torch_version_below("2.9.1"):
```
Dewei-Wang-sh, line 114: later, add comment to state why

Dewei-Wang-sh (approved):
lgtm

README.md
mengfei-jiang: OK

setup.py
mengfei-jiang: OK

PR #3164, opened 2026-05-13
feat(aot): align FlyDSL AOT precompile with runtime cache keys

zhiding512:
## Summary
- Update FlyDSL MoE AOT precompile to call the same runtime wrapper path as `fused_moe_2stages`, using fake tensors and runtime-aligned tensor shapes so generated cache keys match inference-time lookups.
- Add GEMM AOT fake tensor precompile support to avoid requiring real GPU allocations during cache generation.
- Add MoE AOT cache-miss detection in `test_moe_2stage.py` so CSV-derived runtime cases fail when AOT coverage is incomplete.
- Support optional `swiglu_limit` in MoE AOT CSV jobs and align the stage2 `a2_scale` shape logic with `fused_moe.py`.


coderfeli:
CI still failed  @zhiding512

coderfeli (approved):

PR #3174, opened 2026-05-13
[TRITON][GLUON] Add fp8_mqa_logits gluon kernel for gfx950 and gfx1250

cagrikymk:
This PR adds fp8_mqa_logits kernels for gfx950 and gfx1250.

Other related changes:

- Add clean_logits flag, when set to False, the values beyond the limits will be untouched. This saves computation as we dont have to fill them with -inf. However, the topk kernel needs to be mindful of that and onlu read the inbound values.
- Extend benchmark and tests based on this flag and other small changes
- On gfx950, certain features are not available in the latest triton release so I have hacks to extract if they are present in the existing installation and gate them accordingly. While this is hacky, it is needed for backward compat. and performance
- ops/triton/attention/fp8_mqa_logits acts as the main router, the calls are routed to the gluon kernels if the arch. is supported


vgokhale (approved):

valarLip:
have to revert this one due to break...
https://github.com/ROCm/ATOM/actions/runs/25915963950/job/76172886513#step:21:2087
triton.compiler.errors.CompilationError: at 75:16:

cagrikymk:
> have to revert this one due to break... https://github.com/ROCm/ATOM/actions/runs/25915963950/job/76172886513#step:21:2087 triton.compiler.errors.CompilationError: at 75:16:

Sorry for the break, I missed this one because I only tested with Triton 3.6 and 3.7. The aiter CI also uses 3.7 but the problem exists for < 3.6. I will PR it again after adding proper gating based on Triton version.

PR #3175, opened 2026-05-13
DeepSeek FP4 stage1 FQ compatibility override

josusanmartin:
# AITER PR Draft

## Title

DeepSeek FP4 fused-MoE stage1 kernel compatibility and optional import robustness

## Summary

This PR contains a small AITER-side compatibility patch used by the DeepSeek R1 FP4 ATOM submission stack:

- Broaden optional Triton/Iris communication import error handling.
- Add a guarded stage1 FP4/FQ adaptive override for the narrow DSR1 small-token MoE shape.
- Normalize older `_fq` kernel names to the current `_fp4` FlyDSL naming convention.

## Motivation

The current ATOM/AITER stack uses newer FlyDSL naming and package versions than older DSR1 experiment scripts. The small-token MoE path needed a narrow compatibility layer so existing token16 stage1-FQ kernel selections could resolve cleanly on the latest AITER stack.

## Implementation Notes

- `aiter/__init__.py`: catch `RuntimeError`, `OSError`, and `KeyError` in optional Triton/Iris communication import setup, and preserve optional distributed init/destroy imports when CK/HIP ops are unavailable.
- `aiter/fused_moe.py`: add env-gated adaptive stage1-FQ override:
  - `AITER_FMOE_STAGE1_FQ_ADAPTIVE`
  - `AITER_FMOE_STAGE1_FQ_ADAPTIVE_MIN_TOKEN`
  - `AITER_FMOE_STAGE1_FQ_ADAPTIVE_MAX_TOKEN`
  - `AITER_FMOE_STAGE1_FQ_ADAPTIVE_TOKEN16_KERNEL`

The override is restricted to the DSR1 FP4 per-1x32 SiLU MoE shape with `inter_dim == 512`, BF16 output, FP4x2 inputs/weights, grouped gate/up layout, and no `doweight_stage1`.

## Validation

Submitted DeepSeek R1 FP4 leaderboard runs passed GSM8K:

| CONC | TP | GSM8K |
|---:|---:|---:|
| 4 | 4 | 0.9378 |
| 32 | 4 | 0.9363 |
| 128 | 8 | 0.9378 |

Submitted C4/C32 server logs show the token16 stage1 override active. C128 uses TP8 `inter_dim=256`, so the override is intentionally not active there.

### DIFF
--- a/aiter/__init__.py
+++ b/aiter/__init__.py
@@ -115,6 +115,10 @@ else:
             "CK and HIP ops are disabled. Triton ops remain available.",
             e,
         )
+        try:
+            from .ops.communication import destroy_dist_env, init_dist_env  # noqa: F401,E402
+        except Exception:
+            pass
 
 # Import Triton-based communication primitives from ops.triton.comms (optional, only if Iris is available)
 try:
@@ -126,6 +130,6 @@ try:
         reduce_scatter_rmsnorm_quant_all_gather,  # noqa: F401
         IRIS_COMM_AVAILABLE,  # noqa: F401
     )
-except (ImportError, AttributeError):
+except (ImportError, AttributeError, RuntimeError, OSError, KeyError):
     # Iris or triton not available, skip import
     IRIS_COMM_AVAILABLE = False
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -698,6 +698,81 @@ def _normalize_bias_for_kernel(
     return bias
 
 
+@functools.lru_cache(maxsize=None)
+def _fmoe_stage1_fq_adaptive_enabled() -> bool:
+    return os.environ.get("AITER_FMOE_STAGE1_FQ_ADAPTIVE", "0") == "1"
+
+
+@functools.lru_cache(maxsize=None)
+def _fmoe_stage1_fq_adaptive_min_token() -> int:
+    raw = os.environ.get("AITER_FMOE_STAGE1_FQ_ADAPTIVE_MIN_TOKEN", "16")
+    try:
+        return int(raw)
+    except ValueError:
+        return 16
+
+
+@functools.lru_cache(maxsize=None)
+def _fmoe_stage1_fq_adaptive_max_token() -> int:
+    raw = os.environ.get("AITER_FMOE_STAGE1_FQ_ADAPTIVE_MAX_TOKEN", "32")
+    try:
+        return int(raw)
+    except ValueError:
+        return 32
+
+
+def _normalize_stage1_fq_kernel_name(value: str) -> Optional[str]:
+    value = value.strip()
+    if value.lower() in {"", "0", "1", "false", "true", "no", "yes", "off", "on"}:
+        return None
+    # May 8 C4 scripts used the old "_fq" suffix; current FlyDSL uses "_fp4".
+    value = value.replace("_go_fq", "_go_fp4")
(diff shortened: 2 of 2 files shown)


zufayu:
Hi @josusanmartin, a couple of questions on this PR:

1. **Still needed?** This looks like it was written for the ATOM DeepSeek R1 FP4 leaderboard submission. Now that the run is done, is this patch still relevant to keep long-term in the codebase?

2. **Design concern:** Adding 4 env vars () for a very specific shape (, FP4x2, BF16 out) feels like a lot of knobs for a narrow use case. If this is intended to stay, could the selection logic be folded into the existing dispatch path more cleanly rather than as an opt-in override layer?

Happy to move forward if this is still needed — just want to make sure we're not accumulating one-off env vars for past submissions.

zufayu:
Hi @josusanmartin, thanks for the detailed PR description. A few thoughts on the approach:

**Design concern: env vars bypass the existing tune system**

The 4 `AITER_FMOE_STAGE1_FQ_ADAPTIVE*` env vars essentially hand-route a specific shape outside of aiter's standard dispatch/tune flow. This creates maintenance risk — future contributors won't know when these knobs are safe to remove, and they accumulate as dead weight after each submission cycle.

**The right architecture: extend `tuned_fmoe.csv`**

The existing `tuned_fmoe.csv` already handles per-shape kernel selection for bf16/fp8 MoE. The correct path for this token16 + `inter_dim=512` FP4 shape would be:
1. Extend `tuned_fmoe.csv` schema and dispatch logic to support FP4 (`a4w4_blockscale`) quant types for fused MoE
2. Register the optimal stage1 kernel for this shape directly in the CSV

That way the selection is automatic, consistent with the rest of aiter, and survives future refactors cleanly.

**Question for you:**
- Is the ATOM DSR1 FP4 submission still active, or is this a one-off run that's already done?
- If this shape + kernel is genuinely better and worth keeping, are you open to reworking it through the tune table path instead?

Happy to help review a revised approach if you decide to go that route.

josusanmartin:
Thanks @zufayu   I think your read is right.

Just to clarify motivation: I opened this because this was a real compatibility issue I hit while running the ATOM / DeepSeek R1 FP4 stack, not because I wanted to add a permanent leaderboard-specific knob. In that setup, older scripts/kernel names and the current AITER/FlyDSL naming/dispatch path did not line up cleanly, and the token16 stage1 path needed an override to run.

That said, I agree with your design concern. The env-var override is probably not the right long-term shape for AITER, especially now that the specific submission is done. I also no longer have access to the AMD hardware I used during the hackathon, so I cannot responsibly rework and revalidate this through the `tuned_fmoe.csv` path.

Given that, I’m happy to close this PR unless maintainers want to pick up the underlying compatibility issue. The import robustness part may still be worth splitting into a smaller PR if useful.

zufayu:
Thanks @josusanmartin — that context helps a lot. Agreed on closing this PR as-is.

The env-var override layer is the right call to drop; that part's too narrow to carry long-term.

The `aiter/__init__.py` import robustness changes (catching `RuntimeError`/`OSError`/`KeyError` in the Iris import block, and guarding `destroy_dist_env`/`init_dist_env` when CK/HIP is unavailable) are genuinely useful and environment-agnostic. If you're up for it, a small follow-up PR with just those two hunks from `__init__.py` would be worth merging independently — no hardware needed to validate since it's purely defensive import handling.

PR #2911, opened 2026-04-24
F8 fmha ASM gfx950

JohnNikolay84:
## Motivation

There is no f8 fmha ASM kernel for gfx950, despite a similar kernel exists for gfx942.
Diffusion model spend 60-70% flops just in attention, so f8 can gives even 25% e2e improvement over bf16.

## Technical Details

I used MLA ASM kernel for hd192 and reworked it as a pure fmha with hd128. In the end of the day it is not that hard as it is essentially fmha and mfma is 32x32x64 which is perfectly compatible with hd128 too.

## Test Plan

- Run video generation prompts with xDiT.
- aiter fmha f8 tests.

## Test Result

The improvement is 60->45s. No noticeable artifacts.

```
python3 op_tests/test_mha_fp8.py -b 1 -n 5 -q 75600 -d 128
    [aiter] import [module_aiter_core] under /workspace/aiter/aiter/jit/module_aiter_core.so
    [aiter] import [module_fmha_v3_fwd] under /workspace/aiter/aiter/jit/module_fmha_v3_fwd.so
    [aiter] type hints mismatch, override to --> fmha_v3_fwd(q: torch.Tensor, k: torch.Tensor, v: torch.Tensor, dropout_p: float, softmax_scale: float, is_causal: bool, window_size_left: int, window_size_right: int, return_softmax_lse: bool, return_dropout_randval: bool, how_v3_bf16_cvt: int, out: Optional[torch.Tensor] = None, bias: Optional[torch.Tensor] = None, alibi_slopes: Optional[torch.Tensor] = None, q_descale: Optional[torch.Tensor] = None, k_descale: Optional[torch.Tensor] = None, v_descale: Optional[torch.Tensor] = None, gen: Optional[torch.Generator] = None) -> List[torch.Tensor]
    [aiter] LoadKernel: _ZN5aiter24fmha_fwd_hd128_fp8_gfx950E hsaco: /workspace/aiter/hsa//gfx950/fmha_v3_fwd/fwd_hd128_fp8.co
    [W424 14:32:50.280727107 collection.cpp:1133] Warning: ROCTracer produced duplicate flow start: 4 (function operator())
    [aiter] LoadKernel: _ZN5aiter19fmha_fwd_hd128_bf16E hsaco: /workspace/aiter/hsa//gfx950/fmha_v3_fwd/fwd_hd128_bf16.co
Output nrms: 0.004180908203125
Output max diff: 0.001953125
```
```
[aiter] mha summary:
[...]

PR #3186, opened 2026-05-14
i8fp8 fmha gfx950/942 asm

jcaraban:
## Motivation

`fp8` FMHA delivers large speedups over `bf16`, but its precision is insufficient for diffusion models — the quantization error shows up as visible artifacts in generated images and videos.

This PR adds a **Sage-like `i8fp8` FMHA** (int8 Q/K, fp8 V, bf16 output) as a middle point between `bf16` and `fp8`. It is more precise than fp8 (enough to prevent artifacts) while running faster than bf16, about 95% of fp8  performance.

## Technical Details

New asm FMHA forward kernel for head-dim 128 on **gfx950 (MI350)** and **gfx942 (MI300)**:

- **Quantization scheme:** int8 Q and K (QK product accumulated in int32),   fp8 (e4m3) V, bf16 output. Per-tensor or per-(batch, kv-head) `q/k/v_descale`  (float32) applied on the descale path.
- **Dispatch:** routed **implicitly by tensor dtype** (`q,k = int8` and  `v = fp8`) — no new API surface or override flag is required. The new   `data_type` string `"i8fp8bf16"` threads through the existing v3 dispatch:
  - `csrc/py_itfs_cu/asm_mha_fwd.cu` — detects `is_i8fp8`, validates q/k(int8) and v(fp8) dtypes and the descale tensors, and forces a bf16 output.
  - `csrc/cpp_itfs/mha_fwd.cu` — sets `in_bpe = 1`, wires the descale pointers, and admits `i8fp8bf16` through the v3 support guard.
  - `hsa/gfx950/.../fmha_fwd.csv` and `hsa/gfx942/.../fmha_fwd.csv` — one lookup  row each pointing at the new `fwd_hd128_i8fp8.co` kernel.
- **Python entry point:** `flash_attn_i8fp8_pertensor_func(q, k, v, q_descale, k_descale, v_descale, ...)` in `aiter/ops/mha.py`, plus an `is_fmha_v3_i8fp8()` gate in `_flash_attn_forward`.
- **Constraints:** head-dim 128, non-causal full attention, bf16 output only, descales required.

_Note:_ causal and grouped variants could be generated if useful for othe rusers

## Test Plan

Benchmarked and accuracy-checked via `op_tests/op_benchmarks/triton/bench_sage.py`, which was extended with an `aiter_i8fp8` runner, a transformer-like input distribution, QK rotation/clipping for better quantized quality, and a refactored accuracy-metrics report (cosine similarity against the bf16 and Sage references):

```bash
  python op_tests/op_benchmarks/triton/bench_sage.py --b 1 --hq 16 --sq 16384 --d 128 --kernel all
```

## Test Result

- Functional: i8fp8 kernel runs on gfx950 (MI350) and gfx942 (MI300) for hd128.
- Accuracy: cosine similarity 0.998198 vs BF16 — above FP8 although below Triton Sage, but sufficient
- Performance: ~1860 TFLOPS (B1, H16, S16384, D128) for full attention on gfx950

```
bench_sage --kernel=all  (b=1 hq=5 sq=75600 sk=75600 d=128 input=transformer):
kernel             time(ms)     TFLOPS          MAE         MaxE       Cosine
------------------------------------------------------------------------------
sage_fp8             9.6745    1512.37    5.198e-04    4.297e-02     0.998552
sage_mxfp4           7.9293    1845.21    2.386e-03    2.988e-01     0.969782
aiter_fp8            7.1052    2059.24    7.851e-04    8.008e-02     0.996940
aiter_i8fp8          7.8580    1861.96    6.900e-04    4.492e-02     0.997761
aiter_bf16          12.2508    1194.32    0.000e+00    0.000e+00     1.000000
```

## Submission Checklist


aiter/ops/mha.py
```diff
@@ -3106,6 +3118,56 @@ def flash_attn_fp8_pertensor_func(
     return out
 
 
+def flash_attn_i8fp8_pertensor_func(
```
valarLip, line 3121: why we need new python api..

aiter/ops/mha.py
```diff
@@ -1308,22 +1308,37 @@ def is_fmha_v3_fp8():
...
         ret = ret and (hdim_q == 128 or hdim_q == 192)
         ret = ret and (nhead_q % nhead_k == 0)
         ret = ret and (not swa)
-        ret = ret and (q.dtype == dtypes.bf16 or is_fmha_v3_fp8())
+        ret = ret and (q.dtype == dtypes.bf16 or is_fmha_v3_fp8() or is_fmha_v3_i8fp8())
         ret = ret and (cu_seqlens_q is None and cu_seqlens_kv is None)
         # FP8 ASM kernels assemble the GQA-shift from a fixed log2 table
         # (1,2,4,8,16); arbitrary divisor ratios route to CK.
-        if is_fmha_v3_fp8():
+        if is_fmha_v3_fp8() or is_fmha_v3_i8fp8():
```
JohnNikolay84, line 1341: That is a lot of code duplication. Why not having a single function that takes tensor types to check against as an argument ?

jcaraban: The whole AITER mha API is long due for a refactor, precisely so different Q/K/V/O dtypes combinations can go through with minimal code duplication. However in this PR I tried to keep my changes to a minimum, thus introducing temporal duplication. My proposal is to refactor after the mxfp4 and mxfp6 attention variants have also merged, to minimize the chances of breaking something (like e.g [here](https://github.com/ROCm/aiter/pull/2911/changes#diff-7a6efcd8433829b8d1ba9c938cf5e3536a7adca55ef8d02a5ebe21f3a4217c43R186-R189) )

aiter/ops/mha.py
jcaraban: I tried to minimize changes to existing logic on this PR, and leave those for a holistic refactor later, but yes in this case `flash_attn_fp8_pertensor_func` could have been reused with minimal intrusion

jcaraban: but even then, I find the duplication less severe than making the fp8 wrapper dispatch Sage-like i8f8, when the later doesn't fall back to `flash_attn_func_v3_triton()`.

PR #3187, opened 2026-05-14
[TRITON] Remove precision ieee from non-F32 tl.dots

vgokhale:
This is needed only for f32 tl.dots because the AMD backend default is TF32 (and AFAICT only for gfx942).


azaidy (dismissed):
LGTM!

brunomazzottiamd (dismissed):
LGTM!

I added `ci:triton-300x` label to the PR because you said "...and AFAICT only for gfx942". So we can run UTs in gfx942.

azaidy (approved):
LGTM!

PR #3208, opened 2026-05-15
[Bugfix][CKTile] Use local expert ids for MoE bias activation

XiaobingSuper:
## Summary
- Fix the GPT-OSS MoE EP crash where CKTile two-stage bias activation could index local bias tensors with global expert ids and hit `hipErrorIllegalAddress`.
- Make Opus MoE sorting emit `local_topk_ids` on GPU, including the multi-phase MP sorter.
- Add a fused `gelu_and_mul_bias` activation kernel and route the bias GELU path through fused activation instead of a Python helper.

## Problem
With expert parallelism enabled, `topk_ids` contains global expert ids. CKTile split-k/two-stage activation receives bias tensors that are local to the current rank, so using global ids as bias row indices can read past the local bias allocation. This shows up in ATOM GPT-OSS TP+EP startup/cudagraph capture as a GPU illegal memory access.

## Solution
- `moe_sorting_opus` accepts an optional `local_topk_ids` output and fills it directly from the expert mask/cumulative local expert mapping.
- Both Opus sorting policies now support local id emission: the single-kernel path and the multi-phase MP path.
- `fused_moe_` requires sorter-provided local ids for the CKTile bias path; CK sorting falls back to Opus when local ids are needed instead of silently using the unsupported path.
- The CKTile bias activation path now passes local expert ids into fused activation.
- `gelu_and_mul_bias` was added alongside the existing fused bias activations, and invalid expert ids produce zero output rather than indexing bias out of bounds.

## Validation
- `black aiter/fused_moe.py aiter/ops/activation.py op_tests/test_moe_local_expert_ids.py` (3 files unchanged)
- `python3 -m pytest op_tests/test_moe_local_expert_ids.py -q` in ROCm `atom_test` container
- ATOM GPT-OSS-120B validation using this PR branch via `PYTHONPATH=/shared/amdgpu/home/xiaobing_zhang_qle/aiter_main`:
  - Command shape: `python3 -m atom.entrypoints.openai_server --model /shared/data/amd_int/models/gpt-oss-120b --kv_cache_dtype fp8 -tp 2 --enable-dp-attention --enable-expert-parallel --enable-tbo all --gpu-memory-utilization 0.5 --max-num-batched-tokens 16384`
  - Logs show `enable_dp_attention=True`, `enable_tbo=True`, `enable_tbo_decode=True`, DP ranks 0/1, EP ranks 0/1, MORI intra-node EP setup, CKTile two-stage MoE paths, TBO cudagraph capture `[1, 2, 4, 8, 16, 32, 48, 64, 128, 256, 512]`, and server ready.
  - A `/v1/chat/completions` request returned HTTP 200 with usage reported; no `hipErrorIllegalAddress`, illegal memory access, traceback, or runtime error was observed in the server log.


coderfeli:
I updated to main with aot cache miss check by default. @XiaobingSuper

valarLip (approved):

PR #3210, opened 2026-05-15
[feat](gpt-oss): add a8w8 gemm tunefile for gpt-oss

PerryZhang01:
## Motivation

This PR is for a8w8 gemm for gpt-oss.


yifehuan:
PR is based on a base from May 19, but gptoss_bf16_tuned_gemm.csv has since been re-tuned on main with a faster flydsl set. Merging this would overwrite the newer configs with slower ones: of 109 overlapping shapes, 51 regress (>3%) and only 2 improve — +23.4% slower overall, up to +66% (e.g. M=2,N=2880,K=2048: main flydsl 5.56us → this PR asm 9.26us). It also drops 16 shapes that exist on main. Please rebase onto current main and re-tune (or diff against the existing csv and only add genuinely faster / missing shapes).  Also: please drop the hipBLASLt kernelName column values — the runtime selects hipBLASLt by solidx, not by name, and those mangled names are unstable across library versions.

zufayu:
@PerryZhang01 @yifehuan This PR has been inactive for over 2 months. Is the gpt-oss a8w8 gemm tunefile still needed? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3214, opened 2026-05-15
configs: add DSv4-Flash shape to dsv4_fp8fp4_tuned_fmoe.csv

XinyuJiangCMU:
Co-authored-by: @JessicaJiang-123

## Problem

`aiter/configs/model_configs/dsv4_fp8fp4_tuned_fmoe.csv` ships only DSv4-**Pro** shapes:

| model_dim | inter_dim | expert | topk |
|---|---|---|---|
| 7168 | 512 / 768 / 1536 | 384 / 385 | 6 / 7 |

DSv4-**Flash** has a different shape — `(model_dim=4096, inter_dim=256, expert=256, topk=6)` per-rank on TP=8 — and has no entry.

Runtime fmoe dispatch MISSes for every Flash MoE call and falls back to the default CK heuristic, which trips:

```
RuntimeError: Unsupported kernel config for moe heuristic dispatch
```

because the underlying CK template has no `Run()` overload for FP8 act × FP4_mxfp4 weight stage1.

## Fix

Add 34 rows (17 token buckets 1..131072, each with a primary all-FlyDSL row and a `flydsl_fallback` `cktile_a8w4` row) for Flash. Same dtype combo as Pro rows:
- `q_dtype_a = torch.float8_e4m3fn`
- `q_dtype_w = torch.float4_e2m1fn_x2`
- `q_type = QuantType.per_1x32`
- `dtype = torch.bfloat16`, `act_type = ActivationType.Silu`

**Data is produced by running `gemm_moe_tune.py` end-to-end on this shape on 8x MI350X (gfx950, cu_num=256).** All us / tflops / bw numbers in this patch are real per-bucket measurements.

### Note on cktile_a8w4 stage2 for small batch

The raw tuner picks `cktile_a8w4_bm32` as the best stage2 for `token=2` and `token=4` (slightly faster than the runner-up FlyDSL kernel — 20.20us vs 23.62us at token=2, 23.71us vs 26.67us at token=4). However the live `aiter.fused_moe` dispatch path does not route `cktile_a8w4` stage2 for F8 × FP4_mxfp4 — that combination falls through into the CK heuristic dispatch where the missing `Run()` overload throws, exactly the failure mode this PR aims to fix.

To keep the primary rows actually dispatchable end-to-end, the committed primary rows use the **next-best all-FlyDSL stage2** from `profile_flash.csv` (~15% slower than the cktile raw winner for token=2,4; identical to the raw winner for token≥8). Pure-cktile rows are kept as `flydsl_fallback` so deployments without FlyDSL still dispatch cleanly through that path.

A follow-up could either:
1. Wire cktile_a8w4 dispatch through the F8×FP4_mxfp4 path so the raw winner becomes usable, or
2. Leave as-is (the gap is small).

### Per-bucket selections (primary, all-FlyDSL)

| token | us_total | TFLOPS | stage1 | stage2 |
|---|---|---|---|---|
| 1 | 20.49 | 1.84 | `t32x128x256_w4_gui` | `t32x128x256_atomic` |
| 2 | 23.62 | 3.20 | `t64x128x256_w4_gui` | `t64x128x256_atomic_bnt2` |
| 4 | 26.67 | 5.66 | `t64x128x256_w4_gui` | `t64x128x256_reduce_bnt2` |
| 8 | 28.47 | 10.61 | `t32x128x256_w2_bnt0_gui` | `t32x256x256_reduce` |
| 16 | 35.70 | 16.92 | `t32x64x256_w4_bnt0_gui` | `t32x256x256_reduce_persist` |
| 32 | 56.01 | 21.57 | `t32x128x256_w4_bnt0_gui` | `t32x256x256_reduce_persist` |
| 64 | 72.41 | 33.36 | `t32x64x256_w4_gui` | `t32x256x256_reduce` |
| 128 | 82.94 | 58.25 | `t32x64x256_w4_gui` | `t32x256x256_reduce_bnt2_persist` |

(token 256..131072 in CSV; trends continue with larger tile selection.)

All stage1/stage2 kernels are `flydsl_moe{1,2}_afp8_wfp4_bf16_*`.

## Verification + benchmarks

`rocm/sgl-dev:rocm720-mi35x-721f045-20260513-DSv4` image serving DeepSeek-V4-Flash on 8x MI350X (TP=8), HAI's official `run_dsv4.sh` script (RADIX=0 + `--disable-radix-cache`):

- All CFG_DBG lookups HIT (no MISS, no CK heuristic fallback)
- Dispatch routes to `flydsl_moe1/2_afp8_wfp4_*` throughout
- CUDA graph capture completes (~97 s on rerun thanks to FlyDSL JIT cache from the tune sweep; ~2085 s cold)
- `/v1/completions` returns correct output end-to-end

**Single-stream baseline** (`sglang.bench_serving`, random 256-in/256-out, conc=1, 5 prompts after 2 warmup):

| metric | value |
|---|---|
| Output throughput | 71.70 tok/s |
| TTFT (median) | 125 ms |
| TPOT (median) | 13.15 ms |
| ITL (median) | 13.15 ms |

**Concurrent throughput** (random 256-in/256-out, max-concurrency=32, request-rate=16, 100 prompts after 4 warmup):

| metric | value |
|---|---|
| Total throughput (in+out) | 1197 tok/s |
| Output throughput (mean) | 639 tok/s |
| Peak output throughput | 1635 tok/s |
| TTFT (median / P99) | 220 ms / 3.12 s |
| TPOT (median / P99) | 33 ms / 93 ms |

These are clean numbers from a fully warm server. They confirm the kernels selected here run end-to-end at a reasonable rate; an `AITER_ONLINE_TUNE` refresh could squeeze a few more percent but the current selection is already production-usable.


HaiShaw (approved):
LGTM

HaiShaw:
Enable V4-Flash

valarLip (approved):

PR #3223, opened 2026-05-15
fix mi350 bf16gemm issue when m is not multiple of 16

tingchen988:
## Submission Checklist


valarLip:
only have issue with 256x256?

tingchen988:
I think so. This kernel only generates two 256x256.co for use.

valarLip (approved):

PR #3235, opened 2026-05-16
Guard sparse indexer padding and mapped tokens

XiaobingSuper:
## Summary
- Move the `slot_mapping[token_idx] < 0` check earlier in `indexer_qk_rope_quant_and_cache_kernel`.
- Skip padded sparse indexer rows before reading `positions` or RoPE cache pointers, preventing graph padding from reaching invalid addresses in the fused QK cache path.
- Limit both sparse indexer cache update wrappers to the mapped token count when graph padding leaves `slot_mapping` shorter than padded `k` tensors, avoiding out-of-bounds slot reads.

## Test plan
- Rebuilt `module_cache` in the local zxb_vllm_0516 environment for the early guard change.
- Ran the padded-token reproducer with `slot_mapping=-1` and invalid padded `positions`; no HIP illegal memory access was triggered.
- Added runtime diagnostics in zxb_vllm_0516 and reproduced the ATOM/vLLM PIECEWISE case where `num_tokens_padded=512` and `slot_mapping_num_tokens=509`, which this mapped-token limit addresses.


valarLip (approved):
LGTM

PR #3251, opened 2026-05-18
[Triton] Support non-interleaved tensor layout in fused reshape causal conv1d update kernel for Qwen3.5

hellozhuo-amd:
# Support non-interleaved GQA in fused reshape causal conv1d update

## Summary

Adds `qkvz_layout="interleaved"|"flat"` to `fused_reshape_causal_conv1d_update_single_token` so the AITER GDN decode fast path can serve both **Qwen3-Next** (interleaved layout) and **Qwen3.5** (flat layout) without falling back to the generic, slower GDN core path.

## Background

The GDN decode fast path introduced for Qwen3-Next fuses the reshape and
causal-conv1d update into a single Triton kernel.  Qwen3-Next packs its
combined projection as an **interleaved** `[q,k,v,z]` tensor — one chunk per
K-head — and a matching `[b,a]` tensor interleaved per K-head group.

Qwen3.5 uses a different **non-interleaved (flat)** layout:

| Tensor | Qwen3-Next (interleaved) | Qwen3.5 (flat) |
|--------|--------------------------|---------------------------|
| `x` (qkvz) | `[q₀\|k₀\|v₀\|z₀, q₁\|k₁\|v₁\|z₁, …]` per K-head | `[q_all \| k_all \| v_all \| z_all]` flat |
| `ba` | `[b₀\|a₀, b₁\|a₁, …]` per K-head group | `[b_all \| a_all]` flat |

Before this PR, sending Qwen3.5's packed tensors to the fused kernel caused it
to read the wrong columns, producing near-zero model accuracy (GSM8K 0.5%
instead of the expected ~85%).

The downstream `fused_rearrange_sigmoid_gated_delta_rule` kernel already
consumes a **non-interleaved** `[q_all|k_all|v_all]` conv output regardless of
model, so no changes are needed there.

### `aiter/ops/triton/causal_conv1d_update_single_token.py`

- Added `qkvz_layout: str = "interleaved"` parameter to
  `fused_reshape_causal_conv1d_update_single_token` (valid values: `"interleaved"`, `"flat"`).
- Updated `expected_qkvz_dim` validation to account for the flat layout's
  different total dimension:
  - Interleaved: `num_k_heads × (2·Hk + 2·Hv·num_v_heads/num_k_heads)`
  - Flat: `2·num_k_heads·Hk + 2·num_v_heads·Hv`
- Derives `INTERLEAVED_QKVZ` bool and forwards it as a `tl.constexpr` to the kernel so the
  layout branch is resolved at compile time with zero runtime overhead.
- Added docstring clarifying both layout modes.

### `aiter/ops/triton/_triton_kernels/causal_conv1d_update_single_token.py`

Three `@triton.jit` helpers encapsulate the layout-dependent index math,
each taking `INTERLEAVED_QKVZ: tl.constexpr`:

- `_ba_source_offsets` — maps v-head index to `(b_off, a_off)` in the packed `ba` tensor
- `_z_source_idx` — maps flat z index to source column in packed `x`
- `_feat_source_idx` — maps logical conv-output feature index to source column in packed `x`

The main kernel (`_reshape_causal_conv1d_update_single_token_kernel`) calls these helpers in its three layout-sensitive phases:

**`ba` extraction (block 0 — write b and a)**

| Layout | Logic |
|--------|-------|
| Interleaved | `b_source_offset = idx_h × (2·G) + idx_v`; `a_source_offset = idx_h × (2·G) + G + idx_v` |
| Flat | `b_source_offset = idx_hv`; `a_source_offset = num_v_heads + idx_hv` |

**`z` extraction (blocks 1…num\_z)**

| Layout | Logic |
|--------|-------|
| Interleaved | Per-K-head slot calculation |
| Flat | `idx_z_x = 2·num_k_heads·Hk + num_v_heads·Hv + idx_z` |

**`q/k/v` feature mapping (remaining blocks)**

| Layout | Logic |
|--------|-------|
| Interleaved | Scattered gather via `head_qkvz_dim` stride |
| Flat | Identity: `idx_feats_x = idx_feats` |

Because `INTERLEAVED_QKVZ` is `tl.constexpr`, Triton emits separate
compiled variants for each model family — there is no branch overhead at
runtime.

### `op_tests/triton_tests/test_causal_conv1d_update_single_token.py`

- Updated `@pytest.mark.parametrize("qkvz_layout", ["interleaved", "flat"])`.
- Conditional `qkvz_dim` construction in the reference path mirrors both packing layouts.
- Both variants pass.

## Backward Compatibility

`qkvz_layout` defaults to `"interleaved"`, so all existing callers are
**bit-identical** — no API break.

---

## Integration with vLLM (vllm-project/vllm#42880)

[vllm-project/vllm#42880](https://github.com/vllm-project/vllm/pull/42880)
added a `gqa_interleaved_layout` guard in `_forward_core_rocm` to fix
Qwen3.5 accuracy.  It routes Qwen3.5 through the slow generic GDN path and
keeps Qwen3-Next on the fast path:

```python
# vLLM before this aiter PR (gdn_linear_attn.py)
if (
    self.gqa_interleaved_layout          # <-- guard: Qwen3.5 falls through
    and attn_metadata.spec_sequence_masks is None
    and attn_metadata.num_prefills == 0
    and attn_metadata.num_decodes > 0
):
    # AITER fast path — fused reshape + conv1d update
    ...
else:
    # Generic (slow) GDN core path
    ...
```

Once this aiter PR is available in the installed `aiter` package, vLLM can
update `_forward_core_rocm` to **pass `qkvz_layout` directly into
the kernel** and remove the guard that excludes Qwen3.5 from the fast path:

```python
# vLLM after adopting this aiter PR (proposed change in gdn_linear_attn.py)
if (
    attn_metadata.spec_sequence_masks is None
    and attn_metadata.num_prefills == 0
    and attn_metadata.num_decodes > 0
):
    # Both interleaved (Qwen3-Next) and flat (Qwen3.5) now served
    # by the same fused kernel — layout handled inside AITER.
    qkv, conv_states = aiter.fused_reshape_causal_conv1d_update_single_token(
        x=x,
        conv_state=conv_state,
        weight=weight,
        bias=bias,
        ba=ba,
        # ... other args ...
        qkvz_layout=self.qkvz_layout,  # <-- new arg
    )
```

## Files changed

| File | Change |
|------|--------|
| `aiter/ops/triton/causal_conv1d_update_single_token.py` | `+qkvz_layout` param, updated validation, kernel forwarding, docstring |
| `aiter/ops/triton/_triton_kernels/causal_conv1d_update_single_token.py` | `+INTERLEAVED_QKVZ: tl.constexpr`, three `@triton.jit` helper functions |
| `op_tests/triton_tests/test_causal_conv1d_update_single_token.py` | Parametrized on both layouts |


tuukkjs:
I tested this with Qwen 3.5 from the vLLM side with `vllm/vllm-openai-rocm:nightly-6e889b582b6a0b11f22b3764be174266faa9ff5e`, AITER PR 3251 source mounted into the image, and a small vLLM overlay that removes the Qwen3.5 guard and passes `gqa_interleaved_layout=self.gqa_interleaved_layout` into `fused_reshape_causal_conv1d_update_single_token`.

| Path | GSM8K flexible-extract | GSM8K strict-match |
|---|---:|---:|
| Previous vLLM guard / non-fast path | `0.8658 ± 0.0094` | `0.8469 ± 0.0099` |
| AITER PR 3251 fast path | `0.8666 ± 0.0094` | `0.8446 ± 0.0100` |

The scores are effectively identical within noise, so this looks good from the vLLM correctness side.
`lm_eval` command:

```bash
lm_eval \
  --model local-completions \
  --model_args model=Qwen/Qwen3.5-397B-A17B-FP8,base_url=http://localhost:8000/v1/completions \
  --tasks gsm8k \
  --batch_size 128 \
  --seed 4321
```

hellozhuo-amd:
For comments from https://github.com/ROCm/frameworks-internal/issues/16675:
Original comments:

> ### Issues Found
> - [Code duplication]: All three branches (write ba, write z, write feats) gained near-identical if/else pairs. A small helper that returns the offset given (layout, …) would reduce visual noise, especially in the long idx_feats_x computation.
> - [Naming]: gqa_interleaved_layout mixes "GQA" (which the kernel always implements) with "interleaved" (the actual variable). Something like qkvz_layout="interleaved"|"flat" (or a flag named after the packing, not "GQA") would describe the option more directly and survive future per-K-head variants.
> - [Test]: New parametrization is added only to test_fused_reshape_causal_conv1d_update_single_token_matches_ref; the older test_causal_conv1d_update_single_token_matches_ref does not cover the new flag.
> 
> ### Overall Assessment
> Looks good — correctness-preserving extension with default-True for back-compat; only minor polish opportunities.

### Resolutions:
- [Code duplication] Extracted three @triton.jit helpers — _ba_source_offsets, _z_source_idx, _feat_source_idx — each taking INTERLEAVED_QKVZ: tl.constexpr. The three inline if/else blocks in the kernel body are replaced with calls to these helpers.
- [Naming] Renamed gqa_interleaved_layout: bool → qkvz_layout: str with values "interleaved" (default, Qwen3-Next) and "flat" (Qwen3.5), with a validation assert. The internal kernel constexpr is renamed GQA_INTERLEAVED_LAYOUT → INTERLEAVED_QKVZ accordingly.
- [Test] The qkvz_layout flag only applies to fused_reshape_causal_conv1d_update_single_token (the _reshape kernel). test_causal_conv1d_update_single_token_matches_ref tests the non-fused causal_conv1d_update_single_token path which has no layout parameter, so no change is needed there.

vgokhale (approved):

PR #3252, opened 2026-05-18
Update triton wheel

mengfei-jiang:
- Add triton_kernels test suite from upstream triton (removed NVIDIA-specific tests)
- Add torch.compile constexpr mutation regression test for Triton 3.7.0
- Update install_triton.sh: use release_ PyPI URL, add triton-kernels install
- Skip triton reinstall when torch < 2.9.1 in setup.py
- Add triton_kernels test timing to split_tests.sh

## Submission Checklist

### DIFF
--- a/.github/scripts/install_triton.sh
+++ b/.github/scripts/install_triton.sh
@@ -2,13 +2,15 @@
 set -e
 
 pip uninstall -y triton pytorch-triton pytorch-triton-rocm triton-rocm amd-triton || true
+pip uninstall -y triton-kernels || true
 
-TRITON_INDEX_URL="https://pypi.amd.com/triton/release/rocm-7.0.0/simple/"
+TRITON_INDEX_URL="https://pypi.amd.com/triton/release_/rocm-7.0.0/simple/"
 ROCM_VERSION=$(dpkg -l rocm-core 2>/dev/null | awk '/^ii/{print $3}')
 if [[ -n "$ROCM_VERSION" ]]; then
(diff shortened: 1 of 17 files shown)


vgokhale:
@mengfei-jiang  we have all kernels from triton_kernels in AITER. Why do we need this?

mengfei-jiang:
> we have all kernels from triton_kernels in AITER. Why do we need this?

@vgokhale  If all kernels from triton_kernels are already available in AITER, then this part is indeed unnecessary. It’s only meant for testing triton_kernels for the new triton wheel.

Dewei-Wang-sh:
> @mengfei-jiang we have all kernels from triton_kernels in AITER. Why do we need this?
this change comes from that Lingpeng report using the stand alone installed triton_kernels.
if all customers can switch to the kernels in aiter and the kernels basically the same as in triton_kernels, it would be great.

vgokhale:
@mengfei-jiang  yes all kernels from triton_kernels are available here. In general we should not test third party kernels inside AITER. So even if they are not available (say triton_kernels adds something new tomorrow) then we should add that support inside AITER.

Can you close this PR after confirming everything in triton_kernels is available here?

vgokhale:
> if all customers can switch to the kernels in aiter and the kernels basically the same as in triton_kernels, it would be great.

Hi @Dewei-Wang-sh, if customers cannot use AITER, they should file issues on openai triton. We can suggest they switch. We have kept the API identical for seamless migration. But if they insist on using triton_kernels then that repo owner should be responsible for testing, not us. Basically we cannot test someone else's code.

PR #3257, opened 2026-05-18
[Bugfix][Triton][Hardware][MI300X] fp8_mqa_logits: shrink BLOCK_KV/num_stages on gfx942 to fit 64 KB LDS

jin-amd:
## Summary

In the Triton path of `fp8_mqa_logits`, gate the tile choice
(`BLOCK_KV`, `num_stages`) on a **conservative LDS-budget estimate**
instead of the architecture string. The smaller `(64, 1)` tile is used
only when the default `(128, 2)` would push gfx942's 64 KB-per-CU
shared memory over budget; smaller shapes (e.g. `NUM_HEADS=16,
HEAD_SIZE=64`) keep the original `(128, 2)` tile and incur no
throughput penalty.

## Motivation

The Triton branch of `fp8_mqa_logits` is the path taken on every
architecture that doesn't have the new Gluon kernel — i.e. **gfx942
and CUDA today**. Launched at `BLOCK_KV=128, num_stages=2`, the
`_fp8_mqa_logits_kernel` can request up to ~96 KB of LDS per
workgroup (double-buffered KV tile + fp32 scores accumulator + Q
tile) on shapes like the DeepSeek-V4 indexer (`NUM_HEADS=64`,
`HEAD_SIZE=128`). gfx942 has only 64 KB per CU, so the first JIT
compilation aborts with:

```
triton.runtime.errors.OutOfResources:
out of resource: shared memory, Required: 98304, Hardware limit: 65536.
```

This makes any model that exercises the DSv4 indexer
(e.g. `deepseek-ai/DeepSeek-V4-Flash` served through
[vllm-project/vllm#42893](https://github.com/vllm-project/vllm/pull/42893))
**unrunnable on MI300X** — the engine crashes before a single prefill
completes.

## Why this wasn't caught earlier

`op_tests/triton_tests/attention/test_fp8_mqa_logits.py` already
covers `num_heads=64`, `head_dim=128`, `s_q/s_k` up to `1024×1024` —
exactly the DSv4 indexer shape. The gap is purely in CI coverage:

* `.github/workflows/triton-test.yaml`'s default `triton:` matrix
  runs on `linux-aiter-mi35x-1` (gfx950), which takes the Gluon
  kernel branch added in #3226. The Triton path is never
  JIT-compiled.
* The `triton-mi300x:` matrix (gfx942) is opt-in via the
  `ci:triton-300x` PR label. It runs on `main` but not on PRs
  unless labeled.

I've requested the `ci:triton-300x` label on this PR so the existing
test pinpoints the fix on gfx942 before merge.

## Changes

* `aiter/ops/triton/attention/fp8_mqa_logits.py`:
  * New module-level helper `_gfx942_default_tile_fits_lds(num_heads,
    head_size)` that conservatively models the kernel's LDS footprint
    at the default `(BLOCK_KV=128, num_stages=2)` tile:

    ```
    q_bytes      = num_heads * head_size           # fp8 Q tile
    kv_bytes     = head_size * 128 * 2             # double-buffered fp8 KV
    scores_bytes = num_heads * 128 * 4             # fp32 scores accumulator
    fits          ⟺ q + kv + scores ≤ 64 KB
    ```

    Triton may keep some of these in registers depending on version
    and layout, so the model is an upper bound. The safety direction
    is the correct one — "false positives" only shrink the tile,
    never the reverse.

  * In the Triton branch of `fp8_mqa_logits`, replace `block_kv = 128`
    with a gated choice: `(block_kv, num_stages) = (64, 1)` only when
    `arch == "gfx942"` *and* the estimator says the default wouldn't
    fit; everywhere else the original `(128, 2)` tile is preserved
    unchanged.

Cumulative diff: 27 added / 2 removed.

```diff
@@ -58,6 +58,23 @@ def _permute_accepts_constexpr_tuple() -> bool:
 ASYNC_COPY_SUPPORTS_DISTRIBUTED = _async_copy_accepts_distributed_layout()
 FOLDED_REDUCTED_SUPPORT = _permute_accepts_constexpr_tuple()

+# gfx942 (MI300X) has 64 KB of LDS per CU. Conservative upper bound on the
+# shared-memory `_fp8_mqa_logits_kernel` may request with BLOCK_KV=128,
+# num_stages=2 -- models the double-buffered KV tile, the fp32 scores
+# accumulator, and the Q tile. Triton may keep some of these in registers
+# depending on version/layout, so this can over-predict; the safety direction
+# is the right one (false positives just shrink the tile, never the reverse).
+_GFX942_LDS_BYTES = 64 * 1024
+
+
+def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
+    BLOCK_KV = 128
+    NUM_STAGES = 2
+    kv_bytes = head_size * BLOCK_KV * NUM_STAGES
+    scores_bytes = num_heads * BLOCK_KV * 4
+    q_bytes = num_heads * head_size
+    return q_bytes + kv_bytes + scores_bytes <= _GFX942_LDS_BYTES
+
@@ -114,7 +131,17 @@ def fp8_mqa_logits(
     if not use_gluon:
-        block_kv = 128
+        # gfx942 (MI300X) can JIT-abort with `OutOfResources: shared memory`
+        # at the default (BLOCK_KV=128, num_stages=2) tile on shapes that push
+        # the kernel over 64 KB of LDS (e.g. the DSv4 indexer's NUM_HEADS=64,
+        # HEAD_SIZE=128). Shrink only when the default would overflow so
+        # smaller shapes keep their original throughput.
+        if arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size):
+            block_kv = 64
+            num_stages = 1
+        else:
+            block_kv = 128
+            num_stages = 2
@@ -144,7 +171,7 @@ def fp8_mqa_logits(
             BLOCK_KV=block_kv,
             num_warps=4,
-            num_stages=2,
+            num_stages=num_stages,
             waves_per_eu=2,
             matrix_instr_nonkdim=matrix_instr_nonkdim,
         )
```

### Micro-bench — `_fp8_mqa_logits_kernel` (MI300X gfx942, Triton 3.6.0, seq_q_l = seq_kv_l = 4096)
Same kernel, same wrapper, only `(BLOCK_KV, num_stages)` differs. Each
cell is the mean of 4 measurements taken on 4 different MI300X GPUs (GPUs
4–7 on the same host); 25 warmup + 100 rep per measurement via
`triton.testing.do_bench`. "KV LDS (128,2)" is the per-CU occupancy
estimate `occupancy * num_stages * block_kv * head_size`; the gate shrinks
when it exceeds `0.9 * 64 KB ≈ 58.98 KB`.
| Shape (NH × HD) | stock (128, 2) time | (64, 1) time     | Δ if shrunk | KV LDS (128,2) | This PR picks       |
| --------------- | ------------------- | ---------------- | ----------- | -------------- | ------------------- |
| **16 × 64**     | 0.205 ± 0.003 ms    | 0.655 ± 0.002 ms | +220.2 %    | 32 KB          | **(128, 2)** — fits |
| **32 × 64**     | 0.229 ± 0.003 ms    | 0.767 ± 0.009 ms | +235.3 %    | 32 KB          | **(128, 2)** — fits |
| **128 × 64**    | 0.533 ± 0.006 ms    | 0.553 ± 0.005 ms | +3.8 %      | 32 KB          | **(128, 2)** — fits |
| **256 × 64**    | 1.167 ± 0.008 ms    | 1.015 ± 0.012 ms | −13.0 %     | 32 KB          | **(128, 2)** — fits |
| **64 × 128**    | 0.332 ± 0.003 ms    | 0.578 ± 0.007 ms | +74.3 %     | 64 KB          | **(64, 1)** — DSv4  |
| **128 × 128**   | 0.785 ± 0.009 ms    | 0.662 ± 0.004 ms | −15.7 %     | 64 KB          | **(64, 1)**         |
| **64 × 256**    | 0.612 ± 0.010 ms    | 0.702 ± 0.010 ms | +14.9 %     | 128 KB         | **(64, 1)**         |
Notes:
* **No regression on `HEAD_SIZE=64` shapes.** The KV tile is 32 KB at the
  default `(128, 2)`, under the 58.98 KB occupancy budget, so all four
  `HD=64` shapes keep the fast default tile. The `256×64` row would have
  been ~13 % faster under `(64, 1)`, but the occupancy gate (correctly)
  doesn't shrink it since it fits; this is the deliberate consequence of
  gating on a clean occupancy model rather than the old upper bound.
* **DSv4 still safe.** The DSv4 indexer shape (`64 × 128`) shrinks to
  `(64, 1)`, which is what the original `OutOfResources` deployment
  needed.
* **`128×128` is also a win.** Its `(64, 1)` pick is ~16 % faster than
  the stock tile (better register pressure), so the shrink buys
  throughput there too.
Variance across the 4 MI300X GPUs is tight (stddev ≤ 2.6 % for 6 of 7
shapes, max 5.6 % at 16×64), and run-to-run on the same GPU is within
± 2 %.

### End-to-end (MI300X, gfx942, ROCm 6.x, TP=4 × 4× MI300X)

DeepSeek-V4-Flash serving with this PR's fix + the vLLM-side fixes
from [vllm-project/vllm#42893](https://github.com/vllm-project/vllm/pull/42893).
**GSM8K 5-shot, n=200, exact_match** (vs the upstream PR's reported
0.005 without these fixes — i.e. random-chance gibberish):

| Mode  | exact_match | Stderr   | Wall time |
| ----- | ----------: | -------: | --------: |
| Eager |   **0.955** | ± 0.0147 |     227 s |
| Graph |   **0.955** | ± 0.0147 |      86 s |

Throughput sweep (graph mode, `--gpu-memory-utilization=0.85`,
`vllm bench serve --num-prompts 128 --max-concurrency 32`):

| in / out (tok) | dur (s) | req/s | output tok/s | total tok/s | mean TPOT (ms) | failed |
|----------------|--------:|------:|-------------:|------------:|---------------:|-------:|
| 1 k / 1 k      |   236.2 |  0.54 |        555.0 |     1 110.0 |           52.5 |      0 |
| 1 k / 8 k      | 1 699.7 |  0.08 |        616.9 |       694.0 |           51.6 |      0 |
| 8 k / 1 k      |   446.2 |  0.29 |        293.7 |     2 643.5 |           95.9 |      0 |
| 16 k / 1 k     |   907.8 |  0.14 |        144.4 |     2 454.5 |          187.4 |      0 |
| 24 k / 4 k     | 2 264.3 |  0.06 |        231.5 |     1 620.8 |          121.8 |      0 |
| 30 k / 2 k     | 2 481.1 |  0.05 |        105.7 |     1 690.5 |          256.9 |      0 |

Zero engine errors and zero preemption across all six cells.

## Testing

- [x] **Existing CI test covers the DSv4 shape — needs `ci:triton-300x`
      label to run on gfx942.**
      `op_tests/triton_tests/attention/test_fp8_mqa_logits.py`
      parametrizes `num_heads=64`, `head_dim={64, 128}`, and 9
      `(s_q, s_k)` combinations up to `(1024, 1024)`. With the
      `ci:triton-300x` label applied, the `triton-mi300x:` job runs
      this test on gfx942 and exercises the patched tile end-to-end.
- [x] **Micro-bench on MI300X** — 7 shapes × 4 GPUs × 2 configs (see
      the table above), variance ≤ 5.6 % on the worst shape.
- [x] **End-to-end DSv4-Flash serving on MI300X** — GSM8K (n=200, eager
      and graph) and a 6-cell throughput sweep (see above).
- [ ] Tested on MI350X (gfx950) — **N/A by construction.** gfx950
      takes the Gluon kernel branch (added in #3226) and never reaches
      this code; the `arch == "gfx942"` precondition on the shrink
      keeps `(128, 2)` for any other arch that does fall back to the
      Triton path (CUDA today).

## Documentation

- [x] Module-level comment + helper docstring explain what's modeled
      and the safety direction of the bound.
- [x] Inline comment above the `if` describes the failure mode being
      defended against and the small-shape preservation rationale.
- [ ] No public API change; no docstring update needed.

## Dependencies

- [x] No new third-party dependencies. Uses `arch_info.get_arch()`
      already imported in this file as of #3226.

## Breaking changes

None. CUDA, gfx950, and gfx1250 codepaths are byte-identical to
before this PR. On gfx942, only shapes that the conservative LDS
estimate predicts would not fit the default tile are affected; their
behavior changes from "may JIT-abort" to "smaller tile that fits".

## Related

* Blocks: [vllm-project/vllm#42893](https://github.com/vllm-project/vllm/pull/42893)
  (DeepSeek-V4-Flash on MI300X)
* Related: #3226 (Gluon kernel for gfx950/gfx1250 — that PR fixed
  the same kernel for those archs by routing them away; this PR
  fixes the remaining Triton path for gfx942)

### DIFF
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -61,13 +61,22 @@ def _permute_accepts_constexpr_tuple() -> bool:
 ASYNC_COPY_SUPPORTS_DISTRIBUTED = _async_copy_accepts_distributed_layout()
 FOLDED_REDUCTED_SUPPORT = _permute_accepts_constexpr_tuple()
 
-# gfx942 (MI300X) has 64 KB of LDS per CU. Conservative upper bound on the
-# shared-memory `_fp8_mqa_logits_kernel` may request with BLOCK_KV=128,
-# num_stages=2 -- models the double-buffered KV tile, the fp32 scores
-# accumulator, and the Q tile. Triton may keep some of these in registers
-# depending on version/layout, so this can over-predict; the safety direction
-# is the right one (false positives just shrink the tile, never the reverse).
-_GFX942_LDS_BYTES = 64 * 1024
+# gfx942 (MI300X) has 64 KB of LDS per CU. We accept the default
+# (BLOCK_KV=128, num_stages=2) tile only when *both* of these hold:
+#
+# 1. Occupancy gate. With waves_per_eu=2 and num_warps=4 we target two
+#    workgroups co-resident on a CU -> per-WG LDS budget = 32 KB. Triton 3.6+
+#    keeps Q in registers (loop-invariant) and the fp32 scores accumulator
+#    in VGPRs (heavy VALU), so only the double-buffered KV tile lives in
+#    LDS. A 0.9 safety factor leaves headroom for any LDS overhead the
+#    compiler may add.
+#
+# 2. Hardware ceiling. Defensive upper bound that also counts Q and scores
+#    against the 64 KB CU limit, in case a Triton version (older or future)
+#    decides to spill them to LDS. False positives here only shrink the
+#    tile; false negatives are JIT-aborts, so we lean conservative.
+_GFX942_CU_LDS_BYTES = 64 * 1024
+_GFX942_PER_WG_LDS_BUDGET_BYTES = _GFX942_CU_LDS_BYTES * 9 // 20  # ~28.8 KiB
 
 
 def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
@@ -76,7 +85,9 @@ def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
     kv_bytes = head_size * BLOCK_KV * NUM_STAGES
     scores_bytes = num_heads * BLOCK_KV * 4
     q_bytes = num_heads * head_size
-    return q_bytes + kv_bytes + scores_bytes <= _GFX942_LDS_BYTES
+    fits_occupancy = kv_bytes < _GFX942_PER_WG_LDS_BUDGET_BYTES
+    fits_hardware = q_bytes + kv_bytes + scores_bytes <= _GFX942_CU_LDS_BYTES
+    return fits_occupancy and fits_hardware
 
 
 def fp8_mqa_logits(
@@ -134,11 +145,13 @@ def fp8_mqa_logits(
     stride_w_s, stride_w_h = weights.stride()
     stride_logits_s, stride_logits_k = logits.stride()
     if not use_gluon:
-        # gfx942 (MI300X) can JIT-abort with `OutOfResources: shared memory`
-        # at the default (BLOCK_KV=128, num_stages=2) tile on shapes that push
-        # the kernel over 64 KB of LDS (e.g. the DSv4 indexer's NUM_HEADS=64,
-        # HEAD_SIZE=128). Shrink only when the default would overflow so
-        # smaller shapes keep their original throughput.
+        # gfx942 (MI300X): the default (BLOCK_KV=128, num_stages=2) tile
+        # double-buffers a `HEAD_SIZE * 128`-byte KV tile in LDS. With
+        # waves_per_eu=2, num_warps=4 we want two workgroups co-resident on a
+        # CU (per-WG LDS budget ~32 KB), so shapes with HEAD_SIZE >= 128 spill
+        # past that budget and either lose occupancy or, on older Triton,
+        # JIT-abort with `OutOfResources: shared memory`. Shrink only when
+        # the default would overflow so smaller shapes keep their throughput.
         # TODO: gate on computed LDS budget instead of arch string
         if arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size):
             block_kv = 64
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -139,12 +139,13 @@ def fp8_mqa_logits(
         # the kernel over 64 KB of LDS (e.g. the DSv4 indexer's NUM_HEADS=64,
         # HEAD_SIZE=128). Shrink only when the default would overflow so
         # smaller shapes keep their original throughput.
+        # TODO: gate on computed LDS budget instead of arch string
         if arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size):
             block_kv = 64
-            triton_num_stages = 1
+            num_stages = 1
         else:
             block_kv = 128
-            triton_num_stages = 2
+            num_stages = 2
 
         # heuristic for MFMA instruction shape
         matrix_instr_nonkdim = 32
@@ -190,7 +191,7 @@ def fp8_mqa_logits(
             stride_logits_k=stride_logits_k,
             BLOCK_KV=block_kv,
             num_warps=4,
-            num_stages=triton_num_stages,
+            num_stages=num_stages,
             waves_per_eu=2,
             matrix_instr_nonkdim=matrix_instr_nonkdim,
         )
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -191,6 +191,7 @@ def fp8_mqa_logits(
             stride_logits_k=stride_logits_k,
             BLOCK_KV=block_kv,
             num_warps=4,
+            num_stages=num_stages,
             waves_per_eu=2,
             matrix_instr_nonkdim=matrix_instr_nonkdim,
         )
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -72,6 +72,7 @@ def _gfx942_tile_fits_lds(
     # accumulator stay in registers in Triton 3.6+). Account for `occupancy`
     # co-resident workgroups and keep a 0.9 safety factor for compiler
     # overhead.
+    # If a future Triton spills Q or scores to LDS, re-add a `q + kv + scores <= 64 KB` upper-bound term here to avoid re-triggering the JIT abort.
     lds_bytes = occupancy * num_stages * block_kv * head_size
     return lds_bytes <= 0.9 * _GFX942_CU_LDS_BYTES
 
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -117,7 +117,19 @@ def fp8_mqa_logits(
     stride_w_s, stride_w_h = weights.stride()
     stride_logits_s, stride_logits_k = logits.stride()
     if not use_gluon:
-        block_kv = 128
+        # gfx942 (MI300X) has 64 KB of LDS per CU. NUM_HEADS=64 + HEAD_SIZE=128
+        # (DeepSeek-V4 indexer config) needs ~96 KB at block_kv=128 +
+        # num_stages=2, and the kernel JIT aborts with
+        # `triton.runtime.errors.OutOfResources: Required: 98304,
+        # Hardware limit: 65536`. Drop to block_kv=64 + num_stages=1 (~33 KB)
+        # on gfx942; other architectures keep the original tile.
+        # gfx950/gfx1250 use the Gluon kernel above and don't reach this path.
+        if arch == "gfx942":
+            block_kv = 64
+            triton_num_stages = 1
+        else:
+            block_kv = 128
+            triton_num_stages = 2
 
         # heuristic for MFMA instruction shape
         matrix_instr_nonkdim = 32
@@ -163,7 +175,7 @@ def fp8_mqa_logits(
             stride_logits_k=stride_logits_k,
             BLOCK_KV=block_kv,
             num_warps=4,
-            num_stages=2,
+            num_stages=triton_num_stages,
             waves_per_eu=2,
             matrix_instr_nonkdim=matrix_instr_nonkdim,
         )
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -61,33 +61,19 @@ def _permute_accepts_constexpr_tuple() -> bool:
 ASYNC_COPY_SUPPORTS_DISTRIBUTED = _async_copy_accepts_distributed_layout()
 FOLDED_REDUCTED_SUPPORT = _permute_accepts_constexpr_tuple()
 
-# gfx942 (MI300X) has 64 KB of LDS per CU. We accept the default
-# (BLOCK_KV=128, num_stages=2) tile only when *both* of these hold:
-#
-# 1. Occupancy gate. With waves_per_eu=2 and num_warps=4 we target two
-#    workgroups co-resident on a CU -> per-WG LDS budget = 32 KB. Triton 3.6+
-#    keeps Q in registers (loop-invariant) and the fp32 scores accumulator
-#    in VGPRs (heavy VALU), so only the double-buffered KV tile lives in
-#    LDS. A 0.9 safety factor leaves headroom for any LDS overhead the
-#    compiler may add.
-#
-# 2. Hardware ceiling. Defensive upper bound that also counts Q and scores
-#    against the 64 KB CU limit, in case a Triton version (older or future)
-#    decides to spill them to LDS. False positives here only shrink the
-#    tile; false negatives are JIT-aborts, so we lean conservative.
+# gfx942 (MI300X) LDS size per CU.
 _GFX942_CU_LDS_BYTES = 64 * 1024
-_GFX942_PER_WG_LDS_BUDGET_BYTES = _GFX942_CU_LDS_BYTES * 9 // 20  # ~28.8 KiB
 
 
-def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
-    BLOCK_KV = 128
-    NUM_STAGES = 2
-    kv_bytes = head_size * BLOCK_KV * NUM_STAGES
-    scores_bytes = num_heads * BLOCK_KV * 4
-    q_bytes = num_heads * head_size
-    fits_occupancy = kv_bytes < _GFX942_PER_WG_LDS_BUDGET_BYTES
-    fits_hardware = q_bytes + kv_bytes + scores_bytes <= _GFX942_CU_LDS_BYTES
-    return fits_occupancy and fits_hardware
+def _gfx942_tile_fits_lds(
+    block_kv: int, head_size: int, num_stages: int, occupancy: int
+) -> bool:
+    # Only the double-buffered KV tile lives in LDS (Q and the fp32 scores
+    # accumulator stay in registers in Triton 3.6+). Account for `occupancy`
+    # co-resident workgroups and keep a 0.9 safety factor for compiler
+    # overhead.
+    lds_bytes = occupancy * num_stages * block_kv * head_size
+    return lds_bytes <= 0.9 * _GFX942_CU_LDS_BYTES
 
 
 def fp8_mqa_logits(
@@ -145,15 +131,12 @@ def fp8_mqa_logits(
     stride_w_s, stride_w_h = weights.stride()
     stride_logits_s, stride_logits_k = logits.stride()
     if not use_gluon:
-        # gfx942 (MI300X): the default (BLOCK_KV=128, num_stages=2) tile
-        # double-buffers a `HEAD_SIZE * 128`-byte KV tile in LDS. With
-        # waves_per_eu=2, num_warps=4 we want two workgroups co-resident on a
-        # CU (per-WG LDS budget ~32 KB), so shapes with HEAD_SIZE >= 128 spill
-        # past that budget and either lose occupancy or, on older Triton,
-        # JIT-abort with `OutOfResources: shared memory`. Shrink only when
-        # the default would overflow so smaller shapes keep their throughput.
-        # TODO: gate on computed LDS budget instead of arch string
-        if arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size):
+        # On gfx942 (MI300X), drop to (64, 1) when our LDS estimate predicts
+        # the default (128, 2) tile would not fit two co-resident workgroups
+        # on a CU; keep the default tile otherwise.
+        if arch == "gfx942" and not _gfx942_tile_fits_lds(
+            block_kv=128, head_size=head_size, num_stages=2, occupancy=2
+        ):
             block_kv = 64
             num_stages = 1
         else:
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -117,6 +117,8 @@ def fp8_mqa_logits(
     stride_w_s, stride_w_h = weights.stride()
     stride_logits_s, stride_logits_k = logits.stride()
     if not use_gluon:
+        # gfx942 has 64 KB LDS/CU; default (128, 2) tile needs ~96 KB. Use
+        # (64, 1) (~33 KB) so DSv4 indexer (NUM_HEADS=64, HEAD_SIZE=128) fits.
         if arch == "gfx942":
             block_kv = 64
             triton_num_stages = 1
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -61,6 +61,23 @@ def _permute_accepts_constexpr_tuple() -> bool:
 ASYNC_COPY_SUPPORTS_DISTRIBUTED = _async_copy_accepts_distributed_layout()
 FOLDED_REDUCTED_SUPPORT = _permute_accepts_constexpr_tuple()
 
+# gfx942 (MI300X) has 64 KB of LDS per CU. Conservative upper bound on the
+# shared-memory `_fp8_mqa_logits_kernel` may request with BLOCK_KV=128,
+# num_stages=2 -- models the double-buffered KV tile, the fp32 scores
+# accumulator, and the Q tile. Triton may keep some of these in registers
+# depending on version/layout, so this can over-predict; the safety direction
+# is the right one (false positives just shrink the tile, never the reverse).
+_GFX942_LDS_BYTES = 64 * 1024
+
+
+def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
+    BLOCK_KV = 128
+    NUM_STAGES = 2
+    kv_bytes = head_size * BLOCK_KV * NUM_STAGES
+    scores_bytes = num_heads * BLOCK_KV * 4
+    q_bytes = num_heads * head_size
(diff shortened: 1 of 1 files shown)


ChuanLi1101:
Reviewed the diff against `main` + the PR head branch. Correctness is fine — `arch` symbol exists at module scope (from #3226), both branches define `triton_num_stages`, gfx950/gfx1250 take the Gluon path and are byte-identical, CUDA `else` arm is unchanged, LDS math (~33 KB vs. 64 KB cap) checks out.

A few things worth fixing before merge:

**1. PR description doesn't match the diff.**
PR body claims an inline comment was added explaining the LDS budget arithmetic, but commit `bda0c8e` ("remove unnecessary explanation") stripped it. Either restore a one-liner — e.g. `# gfx942 has 64 KB LDS/CU; default (128, 2) tile needs ~96 KB. Use (64, 1) (~33 KB) so DSv4 indexer (NUM_HEADS=64, HEAD_SIZE=128) fits.` — or update the description. Recommend the former; nobody will remember the `== "gfx942"` magic in six months.

**2. The "unit test follow-up" is not actually needed.**
`op_tests/triton_tests/attention/test_fp8_mqa_logits.py` already parametrizes `num_heads ∈ {16, 64}` × `head_dim ∈ {64, 128}` × 5×5 seq combos. The DSv4 shape is covered today. What's missing is just **running it on gfx942** — please add the `ci:triton-300x` label to this PR so we get the regression coverage for free. This also answers the awkward question of why this LDS overflow was never caught: default Triton CI runs on MI35X (Gluon path), so gfx942 + this shape has literally never been exercised in CI.

**3. Tile choice is shape-agnostic — risk for small shapes.**
Hardcoding `(64, 1)` for *all* gfx942 callers means small-shape callers (e.g. `num_heads=16, head_dim=64`) eat reduced pipelining unnecessarily. The end-to-end vLLM numbers only cover the DSv4 shape; the indexer kernel isn't the bottleneck there. Please run `op_tests/op_benchmarks/triton/bench_fp8_mqa_logits.py` for at least one small shape (e.g. `--num_heads_q 16 --head_dim 64`) before/after to confirm no micro-bench regression, and add a `# TODO: gate on computed LDS budget instead of arch string` next to the new `if`.

**4. Nit.**
`triton_num_stages` → `num_stages` inside the `if not use_gluon:` branch — the `triton_` prefix is tautological here.

Summary: (1) and (2) are blocking from my side; (3) is strongly suggested given the perf-table-only-for-one-shape framing in the PR body; (4) is optional.

jin-amd:
> Reviewed the diff against `main` + the PR head branch. Correctness is fine — `arch` symbol exists at module scope (from #3226), both branches define `triton_num_stages`, gfx950/gfx1250 take the Gluon path and are byte-identical, CUDA `else` arm is unchanged, LDS math (~33 KB vs. 64 KB cap) checks out.
> 
> A few things worth fixing before merge:
> 
> **1. PR description doesn't match the diff.** PR body claims an inline comment was added explaining the LDS budget arithmetic, but commit `bda0c8e` ("remove unnecessary explanation") stripped it. Either restore a one-liner — e.g. `# gfx942 has 64 KB LDS/CU; default (128, 2) tile needs ~96 KB. Use (64, 1) (~33 KB) so DSv4 indexer (NUM_HEADS=64, HEAD_SIZE=128) fits.` — or update the description. Recommend the former; nobody will remember the `== "gfx942"` magic in six months.
> 
> **2. The "unit test follow-up" is not actually needed.** `op_tests/triton_tests/attention/test_fp8_mqa_logits.py` already parametrizes `num_heads ∈ {16, 64}` × `head_dim ∈ {64, 128}` × 5×5 seq combos. The DSv4 shape is covered today. What's missing is just **running it on gfx942** — please add the `ci:triton-300x` label to this PR so we get the regression coverage for free. This also answers the awkward question of why this LDS overflow was never caught: default Triton CI runs on MI35X (Gluon path), so gfx942 + this shape has literally never been exercised in CI.
> 
> **3. Tile choice is shape-agnostic — risk for small shapes.** Hardcoding `(64, 1)` for _all_ gfx942 callers means small-shape callers (e.g. `num_heads=16, head_dim=64`) eat reduced pipelining unnecessarily. The end-to-end vLLM numbers only cover the DSv4 shape; the indexer kernel isn't the bottleneck there. Please run `op_tests/op_benchmarks/triton/bench_fp8_mqa_logits.py` for at least one small shape (e.g. `--num_heads_q 16 --head_dim 64`) before/after to confirm no micro-bench regression, and add a `# TODO: gate on computed LDS budget instead of arch string` next to the new `if`.
> 
> **4. Nit.** `triton_num_stages` → `num_stages` inside the `if not use_gluon:` branch — the `triton_` prefix is tautological here.
> 
> Summary: (1) and (2) are blocking from my side; (3) is strongly suggested given the perf-table-only-for-one-shape framing in the PR body; (4) is optional.

@ChuanLi1101  Thanks for the careful review — all four addressed:

1. **Inline LDS-budget comment restored** (a6c04e2): module-level explainer above `_gfx942_default_tile_fits_lds`, plus at the `if` calling out that DSv4's `NUM_HEADS=64, HEAD_SIZE=128` is what overflows.

2. **PR description updated** to drop the "needs unit test" follow-up — agreed, the existing `test_fp8_mqa_logits.py` parametrization already covers (16,64) / (64,128). **Could you add `ci:triton-300x`?** I don't have the rights on the repo so I can't apply it myself.

3. **Now shape-aware** This is the biggest change from previous solution! (0183a77): the `if` is `arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size)`. Small shapes (e.g. `16 × 64`, conservative LDS ~9 KB) stay on the fast default `(128, 2)`; only callers whose default-tile footprint exceeds 64 KB drop to `(64, 1)`. Reran `bench_fp8_mqa_logits.py` for `(16, 64)` and the DSv4 `(64, 128)` shape on multiple MI300X devices — **no small-shape regression**, DSv4 goes from crash → works. Numbers in the PR body. TODO added next to the `if` (d77c81d).

4. **`triton_num_stages` → `num_stages`** (aea791b + 8f9c1b7 to also update the launch kwarg).

FWIW I also reran the full DSv4-Flash GSM8K-200 on MI300X with this shape-aware version as a regression sanity check — eager 0.955 flex / 0.950 strict, graph 0.960 / 0.960, matches the prior arch-only-gate numbers (expected: same kernel launch args at the DSv4 indexer shape).

aiter/ops/triton/attention/fp8_mqa_logits.py
```diff
@@ -114,7 +131,18 @@ def fp8_mqa_logits(
...
     stride_logits_s, stride_logits_k = logits.stride()
     if not use_gluon:
-        block_kv = 128
+        # gfx942 (MI300X) can JIT-abort with `OutOfResources: shared memory`
+        # at the default (BLOCK_KV=128, num_stages=2) tile on shapes that push
+        # the kernel over 64 KB of LDS (e.g. the DSv4 indexer's NUM_HEADS=64,
+        # HEAD_SIZE=128). Shrink only when the default would overflow so
+        # smaller shapes keep their original throughput.
+        # TODO: gate on computed LDS budget instead of arch string
+        if arch == "gfx942" and not _gfx942_default_tile_fits_lds(num_heads, head_size):
```
vgokhale: @jin-amd  does this work with cudagraph? I don't think you can have host side compute like this.Can you statically set it for gfx942?

aiter/ops/triton/attention/fp8_mqa_logits.py
```diff
@@ -58,6 +58,23 @@ def _permute_accepts_constexpr_tuple() -> bool:
...
+_GFX942_LDS_BYTES = 64 * 1024
+
+
+def _gfx942_default_tile_fits_lds(num_heads: int, head_size: int) -> bool:
+    BLOCK_KV = 128
+    NUM_STAGES = 2
+    kv_bytes = head_size * BLOCK_KV * NUM_STAGES
+    scores_bytes = num_heads * BLOCK_KV * 4
+    q_bytes = num_heads * head_size
+    return q_bytes + kv_bytes + scores_bytes <= _GFX942_LDS_BYTES
```
cagrikymk: Given wpeu=2, num_waves=4, to hit occupancy of 2, we need to cap the LDS usage at 32 KB per WG so to WG can coexec. on a CU.

I dont think triton compiler will place q or scores on LDS, we have fixed q for the loop body, and scores goes through a lot of VALU ops that require keeping it on registers.

cagrikymk: In that case, default config gives us (128 * 128 * 2) = 32 KB, just for KV. Not sure if there is any LDS overhead somewhere but maybe we can use < LIMIT, or < LIMIT * 0.9 to be conservative, as the limit here.

aiter/ops/triton/attention/fp8_mqa_logits.py
jin-amd: Yes, this works with graphs.

The branch runs in plain Python before any kernel launch — it only picks BLOCK_KV / num_stages. Under capture, that Python runs once at capture time and the kernel launch (with the chosen values baked in) is what gets recorded; at replay only the launch executes. Its inputs are deterministic for a captured graph: arch is a module constant and num_heads / head_size come from Q.shape, which is fixed across captures. The function already does the same kind of capture-time host compute for seq_len_kv_aligned, matrix_instr_nonkdim, the clean_logits allocation, and the use_gluon arch gate one level up.

Empirically: the PR description's GSM8K row in graph mode gives exact_match = 0.955 (vs 0.005 = random without the fix) at 86 s vs 227 s eager, so capture and replay both clean.

**On statically setting (64, 1) for gfx942** — I tried that first; the micro-bench in the PR body shows it costs +220 % on 16×64 and +235 % on 32×64 (both use ~25–35 KB LDS, well under the 64 KB cap), so an unconditional shrink would slow those callers down for no LDS reason. The gate keeps them on (128, 2) and shrinks only when the conservative estimate predicts overflow. This is a question raised by @ChuanLi1101, the static setting for gfx942 will lead to recession for small shapes (e.g., 16 × 64, conservative LDS ~9 KB), thus this method is used.

aiter/ops/triton/attention/fp8_mqa_logits.py
jin-amd: Good catches on both points — adopted both.

Switched the threshold from the 64 KB hardware cap to the **per-WG occupancy budget** (32 KB × 0.9 ≈ 28.8 KiB for `waves_per_eu=2, num_warps=4` → 2 WGs/CU), and made KV the primary estimator since Q stays in registers (loop-invariant) and the fp32 scores accumulator stays in VGPRs.

I kept a secondary `q + kv + scores ≤ 64 KB` upper-bound check as a paranoia layer — its only job is to defend against a Triton version (older deployments, or a future regression) that decides to spill Q or scores to LDS. False positives there only shrink the tile; false negatives are a JIT-abort, so the asymmetry favors leaving it in. Net cost is one extra arithmetic check.

Resulting picks on gfx942 with the default `(128, 2)` tile (rows where the secondary gate changes the answer in **bold**):

| NH × HD | KV-only | q+kv+scores | Picks |
|---:|---:|---:|:--:|
| 16 × 64, 32 × 64, 64 × 64 | 16 KB | ≤ 56 KB | `(128, 2)` |
| 64 × 128 (DSv4) | 32 KB | 72 KB | `(64, 1)` |
| 128 × 16 | 32 KB | 42 KB | `(64, 1)` |
| **128 × 64** | 16 KB | 90 KB | **`(64, 1)`** (only the upper bound shrinks this) |
| 256 × 64 | 16 KB | 168 KB | **`(64, 1)`** |
| 64 × 256 | 64 KB | 112 KB | `(64, 1)` |

Let me know if you'd rather drop the upper-bound check and rely purely on the occupancy gate — happy to do that if you think the spill behavior in pinned Triton 3.6+ is stable enough to count on.

aiter/ops/triton/attention/fp8_mqa_logits.py
```diff
@@ -58,6 +58,34 @@ def _permute_accepts_constexpr_tuple() -> bool:
...
+    NUM_STAGES = 2
+    kv_bytes = head_size * BLOCK_KV * NUM_STAGES
+    scores_bytes = num_heads * BLOCK_KV * 4
+    q_bytes = num_heads * head_size
+    fits_occupancy = kv_bytes < _GFX942_PER_WG_LDS_BUDGET_BYTES
+    fits_hardware = q_bytes + kv_bytes + scores_bytes <= _GFX942_CU_LDS_BYTES
+    return fits_occupancy and fits_hardware
+
 
 def fp8_mqa_logits(
```
cagrikymk, line 80: Can we make the comments less verbose?

Also, I think fits_occupancy should be enough? 

One more thing, you can carry the LDS size calculation inside the function as it is only used by that.

Basically, the function can take block_kv, num_heds, head_size, num_stages, occupancy and calculate the LDS usage via:

occupancy * num_stages * num_heads * head_size, if this value beyond 0.9 * _GFX942_CU_LDS_BYTES, then it can return false.

As long as the comments make it clear we lower the the config size to fit into the LDS because of our estimation, I think it should be enough.

jin-amd: @cagrikymk Thanks — all asks addressed in [a9b13a0](https://github.com/ROCm/aiter/commit/a9b13a053):

1. Less verbose comments. Module-level dropped from a 14-line block to one line; call-site from 8 lines + a stale TODO to 3 lines.

2. fits_occupancy is enough. Dropped the secondary q + kv + scores ≤ 64 KB hardware-cap check. Q and the fp32 scores accumulator stay in registers in Triton 3.6+ as you noted, so only the double-buffered KV tile is modeled.

3. LDS calculation inside the function. Removed _GFX942_PER_WG_LDS_BUDGET_BYTES from module scope; only _GFX942_CU_LDS_BYTES remains as a constant. The 0.9 factor and the per-CU multiplication now live inside the helper.

4. Function takes the tile params explicitly. Renamed _gfx942_default_tile_fits_lds → _gfx942_tile_fits_lds, signature is now (block_kv, head_size, num_stages, occupancy), body collapses to:

```
lds_bytes = occupancy * num_stages * block_kv * head_size
return lds_bytes <= 0.9 * _GFX942_CU_LDS_BYTES
```

cagrikymk (dismissed):

aiter/ops/triton/attention/fp8_mqa_logits.py
cagrikymk: Thanks, approved!
We can merge it once the CI passes

tjtanaa:
@jin-amd are we proceeding with this?

ChuanLi1101:
LGTM overall — well-scoped fix and the occupancy-based gate is the right call. A couple of non-blocking notes before merge:

1. **PR description is out of sync with the final code.** The diff and the micro-bench `This PR picks` column still reflect the old `q + kv + scores <= 64KB` upper-bound estimator. The final implementation is a pure occupancy gate (`occupancy * num_stages * block_kv * head_size <= 0.9 * 64KB`), which changes the picks for some shapes (e.g. `128x64` and `256x64` now keep `(128, 2)`, consistent with the commit message in b3a130e). Worth updating the description so it matches.

2. **Dropping the secondary hardware-cap check is a deliberate tradeoff worth flagging.** The gate now relies on `Triton 3.6+ keeps Q and the fp32 scores accumulator in registers.` If a future Triton version spills either to LDS, the estimate under-predicts and could re-trigger the JIT abort. Fine as-is per the earlier discussion, but might be worth a note somewhere.

3. Minor: `_gfx942_tile_fits_lds` takes `block_kv` but it's always called with `128` — harmless, just noting.

Main thing is to confirm the `ci:triton-300x` job actually passes on gfx942, since that's the real validation for this fix.

jin-amd:
> LGTM overall — well-scoped fix and the occupancy-based gate is the right call. A couple of non-blocking notes before merge:
> 
> 1. **PR description is out of sync with the final code.** The diff and the micro-bench `This PR picks` column still reflect the old `q + kv + scores <= 64KB` upper-bound estimator. The final implementation is a pure occupancy gate (`occupancy * num_stages * block_kv * head_size <= 0.9 * 64KB`), which changes the picks for some shapes (e.g. `128x64` and `256x64` now keep `(128, 2)`, consistent with the commit message in [b3a130e](https://github.com/ROCm/aiter/commit/b3a130e2eec7ea0c0049ecf2215953cbd072598e)). Worth updating the description so it matches.
> 2. **Dropping the secondary hardware-cap check is a deliberate tradeoff worth flagging.** The gate now relies on `Triton 3.6+ keeps Q and the fp32 scores accumulator in registers.` If a future Triton version spills either to LDS, the estimate under-predicts and could re-trigger the JIT abort. Fine as-is per the earlier discussion, but might be worth a note somewhere.
> 3. Minor: `_gfx942_tile_fits_lds` takes `block_kv` but it's always called with `128` — harmless, just noting.
> 
> Main thing is to confirm the `ci:triton-300x` job actually passes on gfx942, since that's the real validation for this fix.

Thanks @ChuanLi1101:

1. Description out of sync — updated the body to describe the final occupancy gate (occupancy * num_stages * block_kv * head_size <= 0.9 * 64 KB) and corrected the micro-bench "This PR picks" column: 128×64 and 256×64 now keep (128, 2), matching b3a130e/a9b13a0. Since the formula has no num_heads term, the pick depends only on HEAD_SIZE (≤64 keeps (128, 2), ≥128 shrinks to (64, 1)).

2. Dropped hardware-cap check — flagged in code (d07d14a6): added a note in _gfx942_tile_fits_lds that the gate relies on Triton 3.6+ keeping Q and the fp32 scores accumulator in registers, and that the q + kv + scores <= 64 KB upper-bound term should be re-added if a future Triton spills either to LDS.

3. _gfx942_tile_fits_lds always called with block_kv=128 — intentional; the explicit signature was per cagrikymk's request so the LDS formula is self-contained. Happy to hardcode if preferred.

4. ci:triton-300x — I believe the label is applied; will confirm the gfx942 job goes green.

valarLip (approved):

jin-amd:
Hi @valarLip, thank you for your approval. But there was a failure in the CI test: `op_tests/triton_tests/attention/test_mha.py::test_mha[True-128-48-8-2048-2048-50]`
I reran the same test locally, and it passed without any failure. How can I solve this issue? 

Aslo I need approval for all the workflows.

Also ping @ChuanLi1101 

```
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-64-128-1] PASSED                              [ 97%]
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-64-128-30] PASSED                             [ 97%]
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-64-128-50] PASSED                             [ 98%]
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-2048-2048-1] PASSED                           [ 98%]
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-2048-2048-30] PASSED                          [ 99%]
op_tests/triton_tests/attention/test_mha.py::test_mha[False-128-48-8-2048-2048-50] PASSED                          [100%]

==================================================== warnings summary ====================================================
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
../../../usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101
  /usr/local/lib/python3.12/dist-packages/triton/runtime/autotuner.py:101: DeprecationWarning: warmup, rep, and use_cuda_graph parameters are deprecated. See https://github.com/triton-lang/triton/pull/4496 for details.
    warnings.warn(("warmup, rep, and use_cuda_graph parameters are deprecated. See "

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================ 180 passed, 6 warnings in 28.12s ============================================
```

PR #3265, opened 2026-05-19
[feat](rmsnorm): support pad args for rmsnorm kernel

PerryZhang01:
## Motivation

- Add `x_pad_to_multiple` to the fused all-reduce RMSNorm API so callers can request padded output widths while keeping the residual path on the original hidden dimension.
- Introduce a padded custom all-reduce RMSNorm kernel path that distinguishes input width, valid hidden width, and output width, zero-fills the padded tail, and validates shape constraints more explicitly.
- Improve the communicator fallback logic to handle weak-contiguous and supported strided layouts more robustly, including packing eligible strided 2D inputs into the registered buffer instead of forcing a Python-side contiguous copy.
- Add multi-GPU coverage for padded outputs, padded input storage, and strided input/residual layouts.


valarLip (dismissed):

csrc/include/custom_all_reduce.cuh
```diff
@@ -1325,11 +1341,18 @@ __global__ void __launch_bounds__(tnum, 1) local_device_load_rmsnorm(RankSignals
...
-                int write_idx = bid * (n / pack_size) + n_iter * tnum + threadIdx.x;
-                *(reinterpret_cast<P*>(results) + write_idx)      = rmsnorm_rslt;
-                *(reinterpret_cast<P*>(residual_out) + write_idx) = rmsnorm_inp;
+                int read_idx         = bid * in_pack_count + n_iter * tnum + threadIdx.x;
+                int output_write_idx = bid * out_pack_count + n_iter * tnum + threadIdx.x;
+                *(reinterpret_cast<P*>(results) + output_write_idx) = rmsnorm_rslt;
+                *(reinterpret_cast<P*>(residual_out) + read_idx)    = rmsnorm_inp;
             }
         }
+        if(tail_pack_count > 0 && threadIdx.x < tail_pack_count)
```
TennyWang1223: Each block only lets threads with threadIdx.x < tail_pack_count write one pack each, with no loop. The dispatcher computes n_loop based on n_packs (valid hidden), not out_n / pack_size:

  int n_loop = (n_packs + 511) / 512;
  launch_fused_allreduce_rmsnorm_pad((local_device_load_rmsnorm<T, 512, 1>));

  If the tail_pack_count corresponding to out_n - n exceeds tnum (block size 512 or 256), some tail positions are never written and the output retains stale memory.

  Example: bf16, n=4096 (n_packs=256, dispatched as tnum=512 / n_loop=1), caller requests x_pad_to_multiple=16384 → out_n=16384, tail = 12288 elements = 1536 packs. Only threadIdx.x < 512 write, leaving 1024 pack slots with garbage.

aiter/dist/device_communicators/communicator_cuda.py
```diff
@@ -199,32 +199,109 @@ def fused_allreduce_rmsnorm(
...
+                weight_,
+                eps,
+                use_1stage,
+                out_hidden_dim=out_n,
+            )
+            assert out is not None
+            assert res_out is not None
+            return out, res_out
+
+        if (
```
TennyWang1223, line 260: only checks input_.stride(-1) == 1 and does not constrain dim, but the C++ helper _can_pack_2d_last_dim_slice   only matches dim() == 2. A 3D (or higher) strided input falls through to:
  HIP_CALL(hipMemcpyAsync((void*)reg_ptr, inp.data_ptr(), data_bytes,
                          hipMemcpyDeviceToDevice, stream));

  Here data_bytes = numel * element_size, but the underlying storage is non-contiguous, so this copies the wrong bytes and reads out of bounds.

TennyWang1223 (changes requested):

csrc/include/custom_all_reduce.cuh
PerryZhang01: Confirmed, this was a real bug.

The padded-tail zero-fill in `local_device_load_rmsnorm` only wrote at most one pack per thread, so large `out_n - n` tails could leave stale values in the padded region. I fixed it by striding the tail zero-fill loop across the whole tail (`tail_offset += blockDim.x`) instead of writing a single pack per thread.

I also added a multi-GPU regression that covers a large padded output case (`shape=(13, 4096)`, `x_pad_to_multiple=16384`) so we exercise tails much larger than the block size.

aiter/dist/device_communicators/communicator_cuda.py
PerryZhang01: I fixed this in two parts:
1. aligned the Python predicate with the C++ packable-layout check, so only safe 2D last-dim slices use the packed-input fast path;
2. updated the general fallback path to flatten leading dims to 2D for Triton RMSNorm and then reshape back, so strided N-D inputs still work correctly without relying on the packed-input path.

I also added a regression covering a strided 3D input.

TennyWang1223 (approved):
LGTM

valarLip (approved):

PR #3274, opened 2026-05-19
[MOE] remove fill(0) in flydsl_moe_stage2

k50112113:
This PR removes the `.fill(0)` in flydsl_moe_stage2 helper function as the moe_sorting kernel already allocates the zero tensor, both are inside the universal `fused_moe` entry point so the zero tensor should always be allocated

on ATOM side:
```
local-completions ({'model': '/data/deepseek-ai/DeepSeek-V4-Pro', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 65, 'max_retries': 2, 'tokenized_requests': False}), gen_kwargs: ({}), limit: 100.0, num_fewshot: 3, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     3|exact_match|↑  | 0.97|±  |0.0171|
|     |       |strict-match    |     3|exact_match|↑  | 0.97|±  |0.0171|
```

Baseline
image

The fill kernel between the 2 stages of MOE is gone
image

### DIFF
--- a/aiter/ops/flydsl/moe_kernels.py
+++ b/aiter/ops/flydsl/moe_kernels.py
@@ -1040,7 +1040,6 @@ def flydsl_moe_stage2(
     inter_dim_pad: int = 0,
     xcd_swizzle: int = 0,
     bias: Optional[torch.Tensor] = None,
-    out_already_filled_with_zeros: Optional[bool] = False,
 ) -> torch.Tensor:
     """Down-projection GEMM (MOE stage2). Supports atomic/reduce modes.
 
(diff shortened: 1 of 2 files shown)


aiter/ops/flydsl/moe_kernels.py
```diff
@@ -1040,6 +1040,7 @@ def flydsl_moe_stage2(
     inter_dim_pad: int = 0,
     xcd_swizzle: int = 0,
     bias: Optional[torch.Tensor] = None,
+    out_already_filled_with_zeros: Optional[bool] = False,
```
lalala-sh: I think just rm fill is enough. No additional parameter is needed.

k50112113: done

azaidy (approved):
LGTM!

valarLip (approved):

PR #3275, opened 2026-05-19
[Triton] remove MOE activation downcast

k50112113:
This PR removes the BF16->FP8 downcast kernel by adding a fused A16W16 GEMM with a downcast copy kernel of the gating operation (the integration will be on the ATOM side, the hidden_states that is sent into fused_moe is fp8 but the dtype argument will have to be explicitly set to BF16 to make sure the MOE config can be correctly selected


zufayu:
@k50112113 This PR has been inactive for over 2 months. Is the MOE activation downcast removal still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3278, opened 2026-05-19
[bugfix] fix mhc split-k acc_sq mask

RolaoDenthu:
## Motivation

When running the fused mHC post/pre split-K path, we will hit a GPU memory access fault. The issue happens when `KSPLIT_POW2 > ACTUAL_KSPLIT`: the kernel creates padded split-K lanes, but the `acc_sq` load in the post/res branches did not mask out those padded lanes.
As a result, the kernel may read invalid split-K partials from `acc_sq`, which can lead to a GPU memory access fault.

## Technical Details

 Applies the split-K mask to those acc_sq loads as well:
```
mask=ks_mask[:, None] & m_mask_post_res[None, :]
```

## Submission Checklist


k50112113 (approved):
Thanks for the debug

valarLip (approved):

PR #3290, opened 2026-05-20
Workaround custom_all_reduce issue (By Chengzhe Fan @zx3xyy)

zjing14:
**Note: the issue is identified and the fix is implemented by Chengzhe Fan @zx3xyy**

Issue: Crash with custom all-reduce + CUDA graph enabled (ROCm / Aiter)

Root Cause Hypothesis: In Aiter's ROCm registered-input CUDA graph custom all-reduce path, a GPU can signal "ready" and peers begin reading the output buffer before the producer kernel's memory writes are globally visible across the PCIe/xGMI fabric.

Workaround: Insert a system-scope memory fence (e.g., `__threadfence_system()` or equivalent ROCm barrier) before the readiness signal, ensuring all prior stores from producer kernels within the CUDA graph are visible to peer GPUs before any consumer on another GPU reads the buffer.

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -144,9 +144,9 @@ DINLINE O downcast(V val)
 }
 
 // This function is meant to be used as the first synchronization in the all
-// reduce kernel. Thus, it doesn't need to make any visibility guarantees for
-// prior memory accesses. Note: volatile writes will not be reordered against
-// other volatile writes.
+// reduce kernel. When using registered CUDA graph inputs, prior memory accesses
+// from producer kernels must be system-visible before peers read the input
+// buffers. Issue __threadfence_system() to ensure global visibility.
 template <int ngpus>
 DINLINE void start_sync(const RankSignals& sg,
 #ifndef USE_ROCM
@@ -156,6 +156,11 @@ DINLINE void start_sync(const RankSignals& sg,
                         int rank)
 {
 #ifdef USE_ROCM
+    // Ensure prior memory writes (e.g. from producer kernels in CUDA graph)
+    // are visible to peer GPUs before signaling readiness.
+    if(threadIdx.x == 0)
+        __threadfence_system();
+    __syncthreads();
     uint32_t flag = self_sg->_flag[blockIdx.x] + 1;
     if(threadIdx.x < ngpus)
     {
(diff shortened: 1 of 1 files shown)


zjing14:
@HaiShaw @carlushuang

TennyWang1223:
● Hi @zjing14 @zx3xyy,

  Thanks for the detailed analysis and the proposed fix — the root cause hypothesis sounds reasonable to me.

  **1. Reproduction details**

  I ran a local stress test composing other kernels + all-reduce into a CUDA graph, but was unable to reproduce the crash after 10,000 iterations. To investigate further, could you share:

  - The **input shape(s)** at the time of the crash
  - The **ROCm version** you're using

  The ROCm version matters here because **ROCm 7.2.0** has a known runtime bug that causes all-reduce crashes, which has been fixed in later releases. Ruling this out first would help confirm we're looking at the same issue.

  **2. On the proposed fix**

  I agree that `__threadfence_system()` should correctly resolve the visibility issue. My only concern is its latency — it's quite heavy and would noticeably regress all-reduce latency on the hot path.

  As an alternative, I'd like to explore replacing it with `buffer_wbl2 sc0 sc1`, which should provide the same cross-GPU visibility guarantee at a much lower cost. However, to validate that this alternative actually fixes the
  underlying race, I first need to be able to reproduce the crash locally — otherwise I can't verify the change is effective.

zjing14:
> The ROCm version matters here because **ROCm 7.2.0** has a known runtime bug that causes all-reduce crashes, which has been fixed in later releases.

@TennyWang1223 Thanks for your comments. We are using ROCm 7.2.0. Let us try the latest ROCm (7.2.3) and create a mini-reproducer.

csrc/include/custom_all_reduce.cuh
```diff
@@ -158,9 +158,18 @@ DINLINE void start_sync(const RankSignals& sg,
 #ifdef USE_ROCM
     // Ensure prior memory writes (e.g. from producer kernels in CUDA graph)
     // are visible to peer GPUs before signaling readiness.
+#if 1
     if(threadIdx.x == 0)
         __threadfence_system();
     __syncthreads();
+#else
```
zjing14, line 165: @TennyWang1223 I am trying to create a minimal reproducer, but I'm still unsuccessful. I also tried `buffer_wbl2 sc0 sc1`, which works, but performance is slightly worse (approximately 3%) than using `__threadfence_system`.

HaiShaw: @TennyWang1223 Please help. Thanks.

TennyWang1223: I think this issue might share the same root cause as https://github.com/ROCm/aiter/pull/3514. Could you cherry-pick it and help verify whether adding end_sync after cross_device_reduce_1stage fixes your issue? If it does, there would
  be no need to modify start_sync. Thanks!

zjing14: @TennyWang1223 The fix works on my side. Let us drop the PR.

zjing14: Thanks @TennyWang1223 and @HaiShaw for your proactive help.

PR #3231, opened 2026-05-15
[TRITON] Use tl.dot(..., acc=...) accumulator form

vgokhale:
Convert `acc += tl.dot(...)` and `acc += tl.dot_scaled(...)` to `acc = tl.dot(..., acc=acc)` / `acc = tl.dot_scaled(..., acc=acc)` so the matrix multiply accumulates directly into acc without a separate add.

PR #3292, opened 2026-05-20
[Triton] [ATOM] MXFP8 GEMM and A8W4 MOE optimization

k50112113:
This PR is co-authored by @azaidy and @k50112113 

This PR includes:

1. Added MXFP8 GEMM (non-shuffled and pre-shuffled), tunned with DSV4 shapes 
2. Added all other mxfp8 GEMM related kernels, including: dynamic_mxfp8_quant, fused_rms_mxfp8_quant, fused_dual_rms_mxfp8_quant, fused_flatten_mxfp8_quant.
3. A8W4 MOE fused with activation and downcast between the 2 MOEs 
4. Optimized Topk and routing for the A8W4 MOE
6. Fix fused_flatten_fp8_group_qaunt bug when N is not power of 2


aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a16w16_quant_x.py
```diff
@@ -0,0 +1,265 @@
...
+    EVEN_MN: tl.constexpr,
+    cache_modifier: tl.constexpr,
+    activation: tl.constexpr,
+    use_activation: tl.constexpr,
+    ADD_BIAS: tl.constexpr,
+    SKIP_REDUCE: tl.constexpr,
+):
+    """Kernel that computes C = A x B and also emits an MXFP8-quantized A.
+
+    The grid is laid out as a single 1D program-id space split into two
```
cagrikymk, line 87: Is there a reason grid is not defined in an overlapping way such that:

`grid = max(gemm_grid, quant_grid)
`
And, in the kernel:

```
if pid < gemm grid:
   do gemm
if pid < quant_grid:
   do quant
```

aiter/ops/triton/_triton_kernels/quant/quant_mxfp8.py
```diff
@@ -0,0 +1,447 @@
...
+        x_2d = tl.reshape(x, (n_groups, QUANT_BLOCK_SIZE))
+        amax = tl.max(tl.abs(x_2d), axis=1, keep_dims=True)
+
+        amax_i32 = amax.to(tl.int32, bitcast=True)
+        amax_i32 = (amax_i32 + 0x200000).to(tl.uint32, bitcast=True) & 0xFF800000
+        amax_p2 = amax_i32.to(tl.float32, bitcast=True)
+        scale_unbiased = tl.log2(amax_p2).floor() - 8
+        scale_unbiased = tl.clamp(scale_unbiased, min=-127, max=127)
+        scale_e8m0 = (scale_unbiased.to(tl.int32) + 127).to(tl.uint8)
+        quant_scale = tl.exp2(-scale_unbiased)
```
cagrikymk: Nitpick:

Maybe we can have mxfp8 util function here so other kernels can import it? (Like the fused gemm + mxfp8 quant one).

aiter/ops/triton/gemm/basic/gemm_afp8wfp8.py
```diff
@@ -0,0 +1,287 @@
...
+#
+# Configs live under aiter.ops.triton.configs.gemm using the standard aiter
+# naming: gfx{arch}-GEMM-MXFP8-PRESHUFFLE-N={N}-K={K}.json, keyed by
+# M_LEQ_x / M_GEQ_x / any per STANDARD_M_BOUNDS. A generic fallback
+# gfx{arch}-GEMM-MXFP8-PRESHUFFLE.json covers untuned (N, K) shapes.
+# -----------------------------------------------------------------------------
+
+# Shapes that have OOM'd at runtime with the tuned config. Future calls
+# bypass the tuned lookup for these and use _DEFAULT_CONFIG. The tuned
+# benches don't run under HIP-graph capture, so they can pick configs that
```
cagrikymk: I am confused about how the LDS behavior/usage change with graph-capture

aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a16w16_quant_x.py
k50112113: I found that it's more performant.. cuz this is the gating, the N dim is small, there is gonna be only few pids along the N dim, so no need to overlap between the quant and the gemms

cagrikymk: Makes sense!

aiter/ops/triton/gemm/basic/gemm_afp8wfp8.py
k50112113: this is dead code, so I remove it. but I don't think LDS usage is going to change graph-capture

aiter/ops/triton/_triton_kernels/quant/quant_mxfp8.py
k50112113: refactored

cagrikymk (dismissed):

aiter/ops/triton/fusions/fused_clamp_act_mul.py
```diff
@@ -54,6 +57,25 @@ def fused_clamp_act_mul(
...
+        assert dtype_quant in (
+            torch.float8_e4m3fn,
+            torch.float8_e4m3fnuz,
+        ), f"ue8m0 requires fp8 e4m3, got {dtype_quant}"
+        if shuffle_scale and transpose_scale:
+            raise ValueError("shuffle_scale incompatible with transpose_scale")
+        _scale_storage_dtype = torch.uint8
+    else:
+        if shuffle_scale:
+            raise ValueError("shuffle_scale only valid with scale_dtype_fmt='ue8m0'")
```
brunomazzottiamd: **Suggestion:**

Theoretically speaking, `ValueError` is for validating user input / external data and `assert` is for internal debugging. You're throwing a `ValueError` to validate function input arguments, this is fine. However, the function is already using `assert` for everything. I'd convert these exceptions to assertions, for the sake of keeping the code consistent:

```python
assert not (shuffle_scale and transpose_scale), "shuffle_scale incompatible with transpose_scale"
```

```python
assert not shuffle_scale, "shuffle_scale only valid with scale_dtype_fmt='ue8m0'"
```

aiter/ops/triton/fusions/fused_clamp_act_mul.py
```diff
@@ -167,5 +190,9 @@ def fused_clamp_act_mul(
     if HAS_QUANT:
         if transpose_scale:
             scale = scale.view(M, num_bs_cols)
+        if shuffle_scale:
+            from aiter.utility import fp4_utils
```
brunomazzottiamd: **Question:** Is it really expensive to import `fp4_utils` at top level? Why do you need to import in a `if` statement? Please leave a comment with the rationale.

aiter/ops/triton/gemm/basic/gemm_afp8wfp8.py
```diff
@@ -0,0 +1,267 @@
...
+    _gemm_afp8wfp8_kernel,
+    _gemm_afp8wfp8_preshuffle_kernel,
+    _get_config,
+)
+from aiter.ops.triton._triton_kernels.common.splitk_reduce import (
+    _gemm_splitk_reduce_kernel,
+)
+
+
+def gemm_afp8wfp8(
```
brunomazzottiamd, line 19: **Question:** Do we really need a new wrapper function? We already have `gemm_a8w8` in `aiter/ops/triton/gemm/basic/gemm_a8w8.py`. After a quick look, I could see that the logic is practically the same. Wouldn't it be worthwhile to extend the existing wrapper instead of creating a new one?

aiter/ops/triton/moe/moe_routing/topk.py
```diff
@@ -1,50 +1,91 @@
...
+        UNBIASED sqrt(softplus(logits)). DeepSeek-V4 noaux_tc router.
 
-    BLOCK_M = 32
+    bias (fp32, [n_expts_tot]): added to scores for selection only, not for
+    returned weights. Only meaningful with score_mode='sqrtsoftplus'.
+
+    renorm: renormalize weights to sum=1 per row before multiplying by
+    routed_scaling_factor.
+    """
+    # print(f"topk {x.dtype=} {x.shape=} {k=} {apply_softmax=} {dim=} {return_bitmatrix=} {HIST_BLOCK_M=} {score_mode=} {bias.dtype} {bias.shape} {renorm=} {routed_scaling_factor=}")
```
brunomazzottiamd: Please remove this dead `print` or convert it to proper debug level log.

aiter/ops/triton/moe/moe_routing/topk.py
```diff
@@ -1,50 +1,91 @@
...
+    assert len(x.shape) == 2
+    n_rows, n_cols = x.shape
+
+    # BLOCK_M=1 for small n_rows keeps the grid wide enough to overlap with
+    # the bitmatrix-scratchpad memset pids (see /app/_test/verify-perf-topk.py).
```
brunomazzottiamd: Where can I find this `verify-perf-topk.py` file? It isn't added by this PR and I couldn't find it in AITER either.

aiter/ops/triton/moe/moe_routing/topk.py
```diff
@@ -1,50 +1,91 @@
...
+    BLOCK_M = 1 if n_rows <= 256 else 32
     BLOCK_N = 128  # triton.next_power_of_2(x_shape[1])  # 128
```
brunomazzottiamd: Remove dead commented code `# triton.next_power_of_2(x_shape[1])  # 128` or at least add a comment explaining why it's still here.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -57,6 +66,17 @@ def allocate_output(
...
 
+def recommend_block_m(m: int) -> int:
+    """Recommend block_m for moe_gemm_a8w4 based on M.
+
+    Prefill (M >= 256) → 64. Decode (M < 256) → 32.
+    Both regimes overridable via env: AITER_A8W4_PREFILL_BM / AITER_A8W4_DECODE_BM.
+    """
+    if m >= 256:
+        return int(os.environ.get("AITER_A8W4_PREFILL_BM", "64"))
+    return int(os.environ.get("AITER_A8W4_DECODE_BM", "16"))
```
brunomazzottiamd: Please be more defensive when reading environment variables. If I do `export AITER_A8W4_DECODE_BM=foo` this code will break with `ValueError: invalid literal for int() with base 10: 'foo'`.

Do we have further restrictions? Should it be positive? Should it be a power of 2? Please enforce these restrictions as well.

aiter/ops/triton/moe/reduce.py
```diff
@@ -14,6 +14,8 @@ def reduce_grouped(
     reduction_n=1,
     out_dtype=None,
     add_residual: bool = True,
+    # Step 9: optional external residual folded into the writeback.
+    residual: "torch.Tensor | None" = None,
```
brunomazzottiamd: Why type annotating with strings here? `torch` is already imported, you aren't forward referencing it. `residual: torch.Tensor | None = None,` should be perfectly fine.

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -300,6 +300,201 @@ def routing(logits, n_expts_act, sm_first=False):
...
+         block_pid_map) from the histogram for the chosen ``block_m``.
+
+    Returns ``(RoutingData, gather_indx, scatter_indx)`` where ``gather_indx``
+    and ``scatter_indx`` are raw int32 tensors — same contract as
+    ``routing_a8w4`` — so ``_a8w4_fused_experts`` consumes them unchanged.
+    """
+    from aiter.ops.triton.fusions.fused_routing_from_topk import (
+        fused_routing_from_topk,
+    )
+    from aiter.ops.triton._triton_kernels.moe.moe_routing.expt_data import (
```
brunomazzottiamd: All newly added functions import something. Do you have a reason for this implementation style? I can think of avoiding circular dependencies or speeding up startup time.

Are these functions supposed to be called in a loop? Even though Python caches imports, the namespace lookup inside a loop can degrade performance on CPU side.

aiter/ops/triton/gemm/fused/fused_gemm_a16w16_quant_x.py
```diff
@@ -0,0 +1,179 @@
...
+        config (Optional[dict]): Kernel tuning parameters.
+        activation (Optional[str]): Activation fused into Y.
+        skip_reduce (Optional[bool]): Skip split-K reduction for Y.
+
+    Returns:
+        Tuple[torch.Tensor, torch.Tensor, torch.Tensor]: (Y, x_quant, x_scales).
+        When skip_reduce=True and NUM_KSPLIT > 1, Y has shape (NUM_KSPLIT, M, N).
+    """
+
+    _LOGGER.info(f"FUSED_GEMM_A16W16_QUANT_X: x={tuple(x.shape)} w={tuple(w.shape)}")
```
brunomazzottiamd, line 70: Wouldn't it be better to use debug level here?

brunomazzottiamd (commented):
Shao, I'm publishing a partial review. I still need to check the kernels in `aiter/ops/triton/_triton_kernels` and the tests in `op_tests/triton_tests`. I've just checked the wrappers and the config files. I can review the rest tomorrow.

aiter/ops/triton/gemm/basic/gemm_afp8wfp8.py
k50112113: We would prefer not, because gemm_a8w8 is for 1x128 group fp8 quant and the scales are true scales to be multiplied back in fp32, while gemm_afp8wfp8 is the "mxfp8" format which we use the same naming convention as gemm_afp4wfp4, meaning the scales are uint8 representing the exponent.

aiter/ops/triton/moe/moe_routing/topk.py
k50112113: it's a temporary file I added elsewhere, I removed this comment

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
k50112113: This is added when tunning the perf, I removed this part and replaced with simple logic `return 64 if m >= 256 else 16 `. This function may be updated in the future when new shapes are being studied, but the logic is simple for now.

aiter/ops/triton/moe/reduce.py
k50112113: typo, changed it to `residual: Optional[torch.Tensor] = None`

aiter/ops/triton/moe/moe_routing/routing.py
k50112113: I think it's just CC, moved it to top

aiter/ops/triton/gemm/fused/fused_gemm_a16w16_quant_x.py
k50112113: I think we have been using `_LOGGER.info` for all other kernels, by Rahul, so I am using the same style.

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_a8w4.py
```diff
@@ -421,4 +430,45 @@ def _moe_gemm_a8w4(
         + offs_y_n.to(index_type)[None, :] * stride_y_n
     )
     mask = mask_m[:, None] & mask_n[None, :]
-    tl.store(YPtrs, out, mask=mask)
+    if HAS_MX_OUT and SPLIT_K == 1:
```
lburzawa, line 431: Could we factor this out into a function? Another option is to have some common function in quant folder that follows this logic. That way both moe kernel and quant kernel could call same piece of code if that's possible.

aiter/ops/triton/moe/moe_routing/topk.py
k50112113: removed

aiter/ops/triton/fusions/fused_clamp_act_mul.py
k50112113: thanks for this check, I forgot to move this into the kernel, we should not do the torch shuffle impl in the helper function

aiter/ops/triton/fusions/fused_clamp_act_mul.py
k50112113: make sense, you are right

aiter/ops/triton/_triton_kernels/moe/moe_routing/topk.py
```diff
@@ -37,45 +56,78 @@ def streaming_topk(
...
+    # -inf — because the transform of -inf is NOT -inf (sqrt(softplus(-inf))
+    # = 0) and would incorrectly win topk against small valid scores.
     X_ptrs = X + offs_m[:, None] * stride_xm + offs_x_n[None, :]
-    x = tl.load(X_ptrs, mask=(mask_m & mask_n), other=float("-inf"))
+    if SCORE_MODE == "softmax":
+        x = tl.load(X_ptrs, mask=(mask_m & mask_n), other=float("-inf"))
+    else:
+        x = tl.load(X_ptrs, mask=(mask_m & mask_n), other=0.0)
+        x = _apply_score_mode(x, SCORE_MODE)
+        if HAS_BIAS:
```
lburzawa, line 88: In deepseek bias is added after the activation function then, is that right?

aiter/ops/triton/moe/moe_routing/topk.py
k50112113: removed

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -300,6 +306,195 @@ def routing(logits, n_expts_act, sm_first=False):
     )
 
 
+def routing_a8w4(
```
lburzawa, line 315: Could we try to unify all the routing function through different arguments and options? It might become hard to maintain and understand so many separate ones.

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -300,6 +306,195 @@ def routing(logits, n_expts_act, sm_first=False):
...
+        gate_scal=gate_scal,
+        expt_hist=hist,
+        n_expts_tot=n_expts_tot,
+        n_expts_act=n_expts_act,
+        expt_data=expt_data,
+    )
+    return routing_data, topk_indx, gate_indx
+
+
+def routing_a8w4_from_hash(
```
lburzawa, line 383: Could we try to unify all the routing function through different arguments and options? It might become hard to maintain and understand so many separate ones.

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -300,6 +306,195 @@ def routing(logits, n_expts_act, sm_first=False):
...
+        gate_scal=gate_scal,
+        expt_hist=hist,
+        n_expts_tot=n_expts_tot,
+        n_expts_act=n_expts_act,
+        expt_data=expt_data,
+    )
+    return routing_data, topk_indx, gate_indx
+
+
+def routing_a8w4_from_topk(
```
lburzawa, line 445: Could we try to unify all the routing function through different arguments and options? It might become hard to maintain and understand so many separate ones.

aiter/ops/triton/moe/moe_routing/topk.py
```diff
@@ -79,3 +123,112 @@ def cdiv(a, b):
         scratchpad_partials=scratchpad_partials,
     )
     return y_vals, y_indx, bitmatrix
+
+
+def hash_routing(
```
lburzawa, line 128: Is this the hash routing function that replaces topk in deepseek? We could even add a separate file for it in moe_routing folder.

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -41,6 +42,14 @@ def allocate_output(
     split_k,
     device,
 ):
+    # ---- output ------
+    N = w.shape[-1]
```
lburzawa: M and N should already be retrieved in the main function, why do we need to retrieve them again instead of just using them as arguments?

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -57,6 +66,14 @@ def allocate_output(
     return matmul_output, final_output
 
 
+def recommend_block_m(m: int) -> int:
```
lburzawa: Why is this necessary and where is this used? Routing already function already determines block_m that has least amount of padding. Using block_m of 64 for batch 256 would result in average of 90%+ padding in your activations. Even block_m has a lot of padding in that case but thats the size of smallest mfma instruction we use.

op_tests/triton_tests/moe/test_moe_routing.py
```diff
@@ -139,6 +149,430 @@ def _assert_indx_equal(ref, tri):
...
+    "score_mode, has_bias, renorm, routed_scaling_factor",
+    [
+        ("sqrtsoftplus", True, True, 2.5),  # full V4 noaux_tc path
+        ("sqrtsoftplus", True, False, 1.0),  # bias, no renorm
+        ("sqrtsoftplus", False, True, 1.0),  # no bias
+        ("softmax", False, False, 1.0),  # identity transform, no renorm
+    ],
+)
+@pytest.mark.parametrize("block_m", [16, 32])
+def test_routing_a8w4(
```
lburzawa, line 384: test file would also benefit from more unification like I mentioned earlier for wrapper code.

aiter/ops/triton/_triton_kernels/moe/moe_routing/topk.py
k50112113: for DSV4, the score_mode is "sqrtsoftplus" -> sqrt(softplus(x)) + bias

aiter/ops/triton/_triton_kernels/moe/moe_op_gemm_a8w4.py
k50112113: factoered

aiter/ops/triton/moe/moe_op_gemm_a8w4.py
```diff
@@ -259,11 +319,21 @@ def moe_gemm_a8w4(
...
+        out_dtype_actual = torch.float8_e4m3fn
+    else:
+        out_dtype_actual = out_dtype
     y, y_final = allocate_output(
-        M,
-        N,
-        out_dtype,
+        x,
+        w,
+        out_dtype_actual,
```
lburzawa: nitpick but I dont think we need to call it "actual"

aiter/ops/triton/gemm/basic/gemm_afp8wfp8.py
brunomazzottiamd: Fine! Let's consider this discussion resolved.

aiter/ops/triton/gemm/fused/fused_gemm_a16w16_quant_x.py
brunomazzottiamd: Fine, let's keep it as-is.

lburzawa (dismissed):

lburzawa (approved):

aiter/ops/triton/_triton_kernels/moe/moe_routing/topk.py
```diff
@@ -26,6 +26,25 @@ def key_to_fpval(x):
     return x ^ tl.where((x & tm) == 0, fm, tm)
 
 
+@triton.jit
+def _apply_score_mode(x, SCORE_MODE: tl.constexpr):
+    """Pre-transform raw logits before topk selection.
+
+    SCORE_MODE values:
```
brunomazzottiamd, line 33: **Non-blocker comment:** It'd be nice to have some compile time checks in place, to guarantee valid values for `SCORE_MODE`. I'm thinking about something like this:

```python
tl.static_assert(
    SCORE_MODE == "softmax" or SCORE_MODE == "sqrtsoftplus",
    "Invalid SCORE_MODE - use softmax or sqrtsoftplus.",
)
```

aiter/ops/triton/_triton_kernels/moe/moe_routing/topk.py
```diff
@@ -26,6 +26,25 @@ def key_to_fpval(x):
...
+      - "softmax": no pre-transform (caller may apply softmax to selected
+        values via APPLY_SOFTMAX flag).
+      - "sqrtsoftplus": x → sqrt(softplus(x)) using numerically stable
+        softplus(x) = max(x, 0) + log(1 + exp(-|x|)). Matches
+        torch.nn.functional.softplus for the DeepSeek-V4 sqrtsoftplus router.
+    """
+    if SCORE_MODE == "sqrtsoftplus":
+        x_f = x.to(tl.float32)
+        softplus_x = tl.maximum(x_f, 0.0) + tl.log(1.0 + tl.exp(-tl.abs(x_f)))
```
brunomazzottiamd, line 42: **Non-blocker comment:**

We are computing $log(1 + z)$ here, where $z = exp(-|x|)$. Do you see any benefit in using something like https://docs.pytorch.org/docs/2.12/generated/torch.log1p.html in this case? $log1p(z)$ is more stable and accurate when $z$ is small.

GPT said that `log1p` is usually implemented as follows in ML frameworks (I didn't double check it):

```python
def log1p(z):
    if (abs(z) < eps):
        return z - 0.5 * z * z
    else:
        return log(1 + z)
```

aiter/ops/triton/_triton_kernels/gemm/fused/fused_gemm_a16w16_quant_x.py
```diff
@@ -0,0 +1,258 @@
...
+                        mask=offs_k[None, :] < k_span - k * BLOCK_SIZE_K,
+                        other=0.0,
+                    )
+                    b = tl.load(
+                        b_ptrs,
+                        mask=offs_k[:, None] < k_span - k * BLOCK_SIZE_K,
+                        other=0.0,
+                        cache_modifier=cache_modifier,
+                    )
+                accumulator += tl.dot(a, b)
```
brunomazzottiamd, line 189: Vinayak has changed all `tl.dot`'s from `acc += tl.dot(...)` form to `acc = tl.dot(..., acc=acc)` form in https://github.com/ROCm/aiter/pull/3231, so the matrix multiply accumulates directly into `acc` without a separate add.

Can you please double check the newly added GEMMs to comply with the current standard?

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -306,6 +312,195 @@ def routing(logits, n_expts_act, sm_first=False):
...
+      2. aiter `sort_tokens` (or `sort_tokens_fused` for tiny M): sort tokens by
+         expert and produce ExptData specialized for the given ``block_m``.
+
+    Returns (RoutingData, gather_indx, scatter_indx) where gather_indx and
+    scatter_indx are raw int32 tensors (no GatherIndx/ScatterIndx wrappers) —
+    consumed directly by ``moe_gemm_a8w4``.
+
+    No multi-block_m dict, no triton_kernels wrapper, no Python bridge step.
+    """
+    from .topk import topk
```
brunomazzottiamd, line 339: **Non-blocker comment:**

Can we move this `import` to file top-level? I think we should avoid the Python `import` machinery in the kernel dispatch path as much as we can. Even if it's cached, it incurs in a lookup.

PS: I don't have any trouble leaving the `import` here if we have a rationale for it.

aiter/ops/triton/moe/moe_routing/routing.py
```diff
@@ -306,6 +312,195 @@ def routing(logits, n_expts_act, sm_first=False):
...
+    """All-Triton routing for the a8w4 path on DeepSeek-V4 hash layers.
+
+    Single fused kernel ``hash_routing`` does tid2eid lookup + score transform
+    + gather + renorm + scale + bitmatrix in one launch, then
+    ``sort_tokens_fused`` (same as :func:`routing_a8w4`) produces ExptData.
+
+    Replaces the Python ``_hash_topk`` + multi-kernel ``fused_routing_from_topk``
+    counting-sort + ``compute_expt_data`` (with memset) chain entirely.
+    """
+    from .topk import hash_routing
```
brunomazzottiamd, line 403: _ditto_

op_tests/triton_tests/moe/test_moe_routing.py
```diff
@@ -1,12 +1,22 @@
...
+    compute_expt_data_torch,
+)
 from aiter.ops.triton.utils._triton.arch_info import get_arch
 
 
 def assert_equal(ref, tri):
     if isinstance(ref, torch.Tensor):
-        assert torch.all(ref == tri)
+        # CI may be failing using this:
+        # assert torch.all(ref == tri)
```
brunomazzottiamd, line 18: This is weird! Do you have any plausible justification for this failure in CI?

Another question: Do we really need to move data to CPU to do the assertion? Would doing `((ref - tri) ** 2).sum() == 0` on GPU work?

In the past, we had some cleanup PRs that reworked the assertions to avoid moving data to CPU as much as possible. This improved CI speed a lot. I'm worried about reintroducing this kind of assertions and killing CI performance by a thousand cuts.

op_tests/triton_tests/gemm/basic/test_gemm_afp8wfp8.py
```diff
@@ -0,0 +1,144 @@
...
+    gemm_afp8wfp8,
+    gemm_afp8wfp8_preshuffle,
+)
+from aiter.ops.shuffle import shuffle_weight
+import aiter.ops.triton.utils._triton.arch_info as arch_info
+
+SCALE_GROUP_SIZE = 32  # A: 1x32 e8m0 scale group
+W_SCALE_K_GROUP = 128  # B: 128 in K direction
+W_SCALE_N_GROUP = 128  # B: 128 in N direction
+FP8_MAX = 448.0  # e4m3 max
```
brunomazzottiamd, line 17: **Non-blocker comment:** `torch.finfo(torch.float8_e4m3fn).max` instead of hardcoding `448.0`?

op_tests/triton_tests/gemm/fused/test_fused_gemm_a16w16_quant_x.py
```diff
@@ -0,0 +1,143 @@
...
+    x_vals = [(16, 1024, 1024)]
+    x_vals += [(128, 8192, 512)]
+    x_vals += [(256, 512, 8192)]
+    x_vals += [(1024, 1024, 1024)]
+    return x_vals
+
+
+@pytest.mark.parametrize("activation", ["gelu", "gelu_tanh", "silu"])
+@pytest.mark.parametrize("M, N, K", get_fewer_x_vals())
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
```
brunomazzottiamd, line 97: Do we really need to test `fp16`? We dropped lots of `fp16` test cases in a recent PR, keeping just `bf16` test cases. The motivation was to speedup CI and keep only real production cases in the test suite.

Please take this comment in consideration for all tests introduced by this PR.

brunomazzottiamd (commented):
Shao, I can't see anything wrong with this PR. Furthermore, I trust Lukasz a lot and he approved the PR!

However, I have some minor requests for you:
* Standardize `tl.dot` form to `acc = tl.dot(..., acc=acc)`.
* Reason about `fp16` test cases: Do we really need them or is it fine to keep just `bf16` cases?
* Try to avoid CPU assertions in unit tests, if possible.

I've left other comments as well, but I think they aren't blockers for merging. Address them at your own judgement.

brunomazzottiamd (approved):
Approved! The comments from my previous review (https://github.com/ROCm/aiter/pull/3292#pullrequestreview-4418801849)  should be addressed in a new PR.

PR #3297, opened 2026-05-21
add pageattention with sliding window

ChengYao-amd:
## Motivation
  `_paged_attention_kernel` and `_paged_attention_kernel_EXPERIMENTAL` already implement
  sliding-window attention (SWA) at the token level — masking K/V positions that fall
  outside `[context_len - sliding_window, context_len)` inside the inner loop. The kernel
  still launches one workgroup per partition (`T_PAR_SIZE = 256` tokens) and runs the
  full QK / softmax / KV-output GEMMs for every partition, even when an entire partition
  lies before the window's lower bound and contributes nothing to the result.
  For long contexts paired with a small SWA window (e.g. `ctx_len = 9216`,
  `sliding_window = 8192`), the wasted partitions are a non-trivial fraction of the
  total work. This PR adds a partition-level early-out so those workgroups exit
  before any matmul, while keeping the reduce kernel correct.
  ## Technical Details
  ### Kernel changes — `csrc/cpp_itfs/pa/pa_kernels.cuh`
  Both `_paged_attention_kernel` and `_paged_attention_kernel_EXPERIMENTAL` get a new
  `if constexpr (SLIDING_WINDOW_ENABLED)` block right after the `wg_start_head_idx` /
  `total_num_heads` computation:
  ```cpp
  if constexpr (SLIDING_WINDOW_ENABLED) {
      const int kv_lo = context_len - sliding_window;
      if (kv_lo > 0 && partition_start_token_idx + T_PAR_SIZE <= kv_lo) {
          // partition is entirely outside the SWA window -> skip GEMMs
      }
  }
  ```
  The reduce kernel always iterates over every partition slot and computes
  `tmp_out[..] * shared_exp_sums[..]`, so leaving the skipped partition's outputs
  uninitialised would propagate NaN (`NaN * 0 = NaN`). The early-out therefore still
  writes well-defined sentinels for every head/MTP slot it owns, then returns:
  1. `max_logits[seq, head, partition] = -FLT_MAX`
  2. `exp_sums[seq, head, partition]   = 0.0f`
  3. `tmp_out[seq, head, partition, 0..HEAD_SIZE) = 0`
  With these sentinels the reduce path's logsumexp and weighted sum naturally drop
  the skipped partition without any branches on its side. The `if constexpr` gate
  ensures non-SWA specializations compile to the original code path with zero
  runtime overhead.
  ### Tests — `op_tests/test_pa_v1_swa.py` (new)
  Targets `torch.ops.aiter.paged_attention_v1`. Components:
  - `run_torch_swa(...)`: from-scratch PyTorch reference that gathers paged K/V from
    `block_tables`, builds the full `QK^T`, applies an explicit
    `position < max(0, seq_len - window_size)` mask, then softmax and V product.
  - `run_aiter_pa_v1_swa(...)`: thin wrapper that allocates the workspace buffer,
    sets `max_num_partitions` from `_PARTITION_SIZE_ROCM = 256`, and dispatches the
    v1 op with the `sliding_window` argument (`0` = disabled).
  - `test_pa_v1_swa_correctness`: parametrized over `dtype ∈ {bf16, fp16}`,
    `num_seqs ∈ {1, 8, 16}`, `(num_q_heads, num_kv_heads) ∈ {(8,1), (16,4), (32,4)}`,
    `head_size ∈ {64, 128}`, `block_size ∈ {16, 32}`, and `(ctx_len, window)` pairs
    covering: `window > ctx` (no masking), `window == ctx`, `ctx == window + 1`
    (one masked position), `ctx >> window` to exercise many fully-skipped partitions
    (e.g. `(9216, 8192)`, `(4097, 2048)`), and `window <= 0` (disabled).
  - `test_pa_v1_swa_matches_full_causal_when_window_ge_context`: equivalence check —
    when `window >= ctx_len`, SWA output must match the full-causal baseline.
  Tolerances: `atol = 1e-2` (bf16) / `5e-3` (fp16), `rtol = 1e-2`, with an additional
  `error_ratio < 0.01` gate via `aiter.test_common.checkAllclose`.
  ## Test Plan
  ```bash
  pytest op_tests/test_pa_v1_swa.py -v
  # Single-shape CLI for quick smoke checks:
  python op_tests/test_pa_v1_swa.py --ctx 9216 --window 8192 --num_seqs 8
  python op_tests/test_pa_v1_swa.py --ctx 4096 --window 1024 --dtype fp16
  ```
  ## Test Result
  Hardware: AMD Instinct MI300X (`gfx942:sramecc+:xnack-`), ROCm, PyTorch 2.9.1.
  ```
  ============================= test session starts ==============================
  platform linux -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0
  collected 793 items
  op_tests/test_pa_v1_swa.py ...........................................  [100%]
  ======================= 793 passed in 238.08s (0:03:58) ========================
  ```
  - `test_pa_v1_swa_correctness`: 792 / 792 passed across all parametrized
    `(dtype, num_seqs, num_heads, head_size, block_size, ctx_len, window_size)`
    combinations, including the long-context / short-window shapes that exercise
    the new partition-level early-out (`(ctx=9216, window=8192)`, `(ctx=4097,
    window=2048)`, `(ctx=2048, window=512)` …).
  - `test_pa_v1_swa_matches_full_causal_when_window_ge_context`: passed —
    SWA output with `window >= ctx_len` matches the full-causal baseline within
    `atol = rtol = 1e-3`.
  - 0 failed, 0 errored.
  ## Submission Checklist


fsx950223:
Could you share performance number

ChengYao-amd:
> Could you share performance number

# AITER `paged_attention_v1(sliding_window=W)` micro-benchmark

Operator-level wall-clock comparison between three decode paths:

| Tag | Implementation |
|---|---|
| `aiter_swa` | `torch.ops.aiter.paged_attention_v1(..., mtp=1, sliding_window=W)` (commit `698a944b`, partition-level + token-level early-exit on the SWA mask) |
| `aiter_full` | Same kernel called with `sliding_window=0` (full causal, no early-exit) |
| `triton_swa` | `aiter.ops.triton.unified_attention(window_size=(W-1, 0))` (the prior fallback for SWA decode) |

## 1. Test setup

| Item | Value |
|---|---|
| Hardware | 1× AMD MI300X (gfx942) |
| Tensor dtype | bf16 |
| KV cache layout | `NHD`, `block_size = 16` |
| Decode query length | 1 token / sequence |
| Q heads | `num_q_heads = 16` |
| KV heads | `num_kv_heads = 4` (GQA-4) |
| Head dim | `head_size = 128` |
| Sweep | `num_seqs ∈ {1, 4, 8, 16}` × `ctx_len ∈ {1024, 2048, 4096, 8192, 9216}` × `window ∈ {256, 512, 2048, 4096, 8192}` |
| Valid configs | 64 (combinations with `window ≥ ctx_len` are skipped — equivalent to full causal) |
| Timing | CUDA-event median, 50 iters after 10 warmup |
| Source | `op_tests/bench_pa_v1_swa_bmoe.py` (depends on `bench_pa_v1_swa.py:bench_one`) |

## 2. Aggregate results over 64 configs

| Comparison | Geomean | Min | Max |
|---|---:|---:|---:|
| `aiter_swa` vs `triton_swa` | **15.31×** | 3.05× | 38.55× |
| `aiter_swa` vs `aiter_full` | **1.21×** | 0.74× | 2.36× |

The `aiter_swa` vs `aiter_full` comparison isolates the gain coming from window-aware partition early-exit on top of the same mfma16 kernel. The 0.74× corner is `window` close to `ctx_len`, where early-exit no longer fires and SWA mask checking adds a small overhead.

## 3. Per-shape wall-clock (`num_seqs = 16`)

The triton fallback time is roughly proportional to `ctx_len`; `aiter_swa` time tracks `eff_ctx = min(ctx_len, window)`.

| ctx_len | window | eff_ctx | `aiter_swa` (µs) | `aiter_full` (µs) | `triton_swa` (µs) | `aiter_swa` vs `triton_swa` | `aiter_swa` vs `aiter_full` |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1024 | 256  | 256  | 55.28  | 58.93  | 216.64  | **3.92×**  | 1.07× |
| 1024 | 512  | 512  | 52.64  | 54.56  | 229.06  | **4.35×**  | 1.04× |
| 2048 | 512  | 512  | 58.33  | 64.74  | 398.91  | **6.84×**  | 1.11× |
| 4096 | 2048 | 2048 | 46.03  | 84.56  | 746.59  | **16.22×** | 1.84× |
| 8192 | 2048 | 2048 | 60.45  | 108.68 | 1752.24 | **28.99×** | 1.80× |
| 9216 | 2048 | 2048 | 67.54  | 141.05 | 1783.97 | **26.41×** | 2.09× |
| 9216 | 4096 | 4096 | 93.90  | 141.57 | 1748.32 | **18.62×** | 1.51× |
| 9216 | 8192 | 8192 | 127.67 | 134.96 | 1750.84 | **13.71×** | 1.06× |

## 4. Per-shape wall-clock (`num_seqs = 1`)

Lower-concurrency strip from the same sweep. Wall time at small `num_seqs` is launch-overhead-bound, so the absolute numbers compress into a narrow band, but the triton-fallback scaling with `ctx_len` is unchanged.

| ctx_len | window | eff_ctx | `aiter_swa` (µs) | `aiter_full` (µs) | `triton_swa` (µs) | `aiter_swa` vs `triton_swa` | `aiter_swa` vs `aiter_full` |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1024 | 256  | 256  | 46.27 | 51.36 | 166.73  | **3.60×**  | 1.11× |
| 2048 | 512  | 512  | 50.07 | 52.44 | 265.07  | **5.29×**  | 1.05× |
| 4096 | 2048 | 2048 | 53.16 | 45.59 | 389.62  | **7.33×**  | 0.86× |
| 8192 | 512  | 512  | 37.01 | 49.39 | 1427.04 | **38.55×** | 1.33× |
| 8192 | 2048 | 2048 | 42.94 | 50.27 | 1256.35 | **29.26×** | 1.17× |
| 9216 | 4096 | 4096 | 51.32 | 37.90 | 1628.46 | **31.73×** | 0.74× |
| 9216 | 8192 | 8192 | 52.36 | 51.20 | 1441.82 | **27.54×** | 0.98× |

## 5. Observations

1. **Triton fallback scales with `ctx_len`, AITER scales with `eff_ctx`.** Triton wall time grows from ≈170 µs at `ctx_len=1k` to ≈1700–1800 µs at `ctx_len=8–9k` regardless of window. AITER stays in the 40–130 µs band across the entire sweep, scaling only with `min(ctx_len, window)`.
2. **Speedup widens as `ctx_len` grows.** At `ctx_len=1k` the gap is ≈3.5–4×; at `ctx_len=8–9k` with a 2k window it widens to ≈26–32× per call. The widening comes from the triton kernel scanning the full context, not from `aiter_swa` getting faster.
3. **Window-aware early-exit pays off on top of the existing mfma16 kernel.** Across all 64 configs, `aiter_swa` is 1.21× faster than `aiter_full` (geomean), with the largest per-row gain (2.36×) appearing at `num_seqs=16, ctx_len=8192, window=256`. The corner where `aiter_swa < aiter_full` is exactly the case `window ≈ ctx_len`, where partition early-exit cannot skip any work.
4. **Small-`num_seqs` runs are launch-overhead-bound on the AITER side.** At `num_seqs=1`, `aiter_swa` time is essentially flat (40–55 µs) regardless of `eff_ctx`; the per-call gain over triton therefore looks larger here (geomean speedup column `aiter_swa vs triton_swa` peaks at 38.55×) but only because the triton time grows while the AITER time is dominated by kernel-launch latency.

## 6. Caveats

- Single-MI300X micro-benchmark; numbers are kernel-only, CUDA-event medians. Python dispatch, `torch.compile` warmup, KV-cache write, and any caller-side glue are excluded.
- `aiter_full` rows reuse the same kernel binary; the only difference vs `aiter_swa` is `sliding_window=0`. This is therefore the cleanest isolation of the SWA early-exit cost/benefit.
- `triton_swa` is the path the inference engine previously fell back to for SWA decode; it is the apples-to-apples baseline for the integration switch.
- Configurations with `window ≥ ctx_len` are skipped to keep the comparison meaningful (they would degenerate to full causal).
- Full per-row data (all 64 configs, including `num_seqs ∈ {1, 4, 8, 16}`) is in `bench_pa_v1_swa_bmoe.csv`.

op_tests/test_pa_v1_swa.py
```diff
@@ -0,0 +1,405 @@
...
+    kv_cache_layout: str = "NHD",
+) -> torch.Tensor:
+    """Wrapper for ``torch.ops.aiter.paged_attention_v1`` with SWA.
+
+    ``window_size <= 0`` means full causal (passes ``sliding_window=0``).
+    """
+    num_seqs, num_heads, head_size = query.shape
+    max_seq_len = int(seq_lens.max().item())
+    max_num_partitions = (
+        max_seq_len + _PARTITION_SIZE_ROCM - 1
```
fsx950223: For sliding window, the max_seq_len should be https://github.com/ROCm/aiter/blob/f7531b3d3239ab41cd1107110c9ba2a20932f498/op_tests/triton_tests/test_pa_decode_gluon.py#L1512.

ChengYao-amd: ## About the issue
Whether `max_seq_len` (and therefore
the `max_num_partitions` used to size the v1 op's workspace) could be tightened to
`min(ctx_len, sliding_window)` when SWA is enabled, mirroring the gluon decode test at
`aiter/op_tests/triton_tests/test_pa_decode_gluon.py:1512`:

```python
max_context_length = (
    min(context_lengths.max().item(), sliding_window)
    if sliding_window > 0
    else context_lengths.max().item()
)
```

## What I tried

I applied that change in `op_tests/test_pa_v1_swa.py` and ran the full parametrized
suite. Result on MI300X (`gfx942`): **260 / 288** of the SWA shapes failed correctness,
all of them in the `ctx_len > sliding_window` regime, with an error ratio of ~2% on bf16.
The 28 cases that still passed are exactly the `ctx_len == sliding_window` shapes,
where `min(ctx, sw) == ctx` and the change is a no-op.

## Why it fails

The reason is a difference in `partition_idx -> token_idx` convention between the two
kernels:

| Kernel | partition 0 starts at | required `max_num_partitions` |
|---|---|---|
| gluon (`test_pa_decode_gluon.py`) | `ctx - sliding_window` (SWA-aware offset) | `cdiv(min(ctx, sw), partition_size)` |
| v1 HIP (`pa_v1.cuh:82`) | `0` (always, regardless of SWA) | `cdiv(ctx, partition_size)` |

Concretely, for `ctx_len = 9216`, `sliding_window = 8192`, `partition_size = 256`:

- The in-window tokens are `[1024, 9216)`, which live in partitions `4 .. 35`.
- Setting `max_seq_len = sw = 8192` shrinks the grid Y dim to `cdiv(8192, 256) = 32`,
  so partitions `4 .. 31` are launched correctly, partitions `0 .. 3` enter the
  early-out path added in this PR...
- ...but partitions `32 .. 35` -- which cover the last 1024 in-window tokens
  immediately preceding the query -- are **never launched**. Their contribution is
  silently dropped, hence the ~2% error.

The gluon-style workspace shrink cannot be applied to the v1 op as-is: the v1 kernel
always treats `partition_idx = 0` as `token = 0`, while the gluon kernel internally
re-bases its iteration to `ctx - sliding_window` when SWA is on. The current PR's
partition-level early-out is the safe variant: it still launches
`cdiv(ctx, partition_size)` workgroups, but each out-of-window one returns after
writing sentinels, with no GEMMs.

## Question for the next step

Two options for getting the gluon-style win on the v1 path:

1. **Leave v1 as-is in this PR.** Keep partition-level early-out, ship the
   correctness/perf improvement on the existing `max_num_partitions = cdiv(ctx, ps)`
   shape. Open a follow-up PR for the workspace shrink.

2. **Extend v1 in a follow-up to support gluon-style partitioning.** Concretely:
   - In `pa_v1.cuh`, compute
     `partition_offset = (SLIDING_WINDOW_ENABLED && ctx > sw) ? floor((ctx - sw) / T_PAR_SIZE) * T_PAR_SIZE : 0`
     and use `partition_start_token_idx = partition_offset + partition_idx * T_PAR_SIZE`.
   - Mirror the same offset in `paged_attention_ll4mi_reduce_kernel` so the
     partition-axis indexing of `max_logits / exp_sums / tmp_out` lines up.
   - In `pa_v1.py`, set
     `max_num_partitions = cdiv(min(max_context_len, sliding_window), partition_size)`
     when SWA is enabled.
   - Update `test_pa_v1_swa.py` and `bench_pa_v1_swa.py` to pass the smaller
     `max_seq_len` and allocate a smaller workspace.

Option 2 saves the launch overhead for fully-out-of-window partitions and shrinks
`tmp_out / max_logits / exp_sums` from `O(ctx)` to `O(sw)` along the partition axis,
which is meaningful when `ctx >> sw` (e.g. `ctx = 64K`, `sw = 4K`).

Should I proceed with option 2 in a follow-up PR (kernel + python + tests + bench),
or keep this PR scoped to the partition-level early-out only?

fsx950223:
The kernel implementation is abandoned

ChengYao-amd:
> The kernel implementation is abandoned

But our customer is still using it, we give them a patch now. Hope to merge into aiter so that we can deliver a former version to them.

fsx950223:
> > The kernel implementation is abandoned
> 
> But our customer is still using it, we give them a patch now. Hope to merge into aiter so that we can deliver a former version to them.

Why not use pa_decode_gluon directly?

ChengYao-amd:
> > > The kernel implementation is abandoned
> > 
> > 
> > But our customer is still using it, we give them a patch now. Hope to merge into aiter so that we can deliver a former version to them.
> 
> Why not use pa_decode_gluon directly?

I tested this kernel 1x faster than the pa_decode_gluon in our case.

fsx950223:
> > > > The kernel implementation is abandoned
> > > 
> > > 
> > > But our customer is still using it, we give them a patch now. Hope to merge into aiter so that we can deliver a former version to them.
> > 
> > 
> > Why not use pa_decode_gluon directly?
> 
> I tested this kernel 1x faster than the pa_decode_gluon in our case.

Could you share the compare unit test script?

ChengYao-amd:
> > > > > The kernel implementation is abandoned
> > > > 
> > > > 
> > > > But our customer is still using it, we give them a patch now. Hope to merge into aiter so that we can deliver a former version to them.
> > > 
> > > 
> > > Why not use pa_decode_gluon directly?
> > 
> > 
> > I tested this kernel 1x faster than the pa_decode_gluon in our case.
> 
> Could you share the compare unit test script?

```

"""
Latency comparison for sliding-window-attention (SWA) decode kernels:

  1. ``v1_swa``     - HIP MFMA16 ``torch.ops.aiter.paged_attention_v1`` with
                      partition-level early-out (this PR).
  2. ``v1_full``    - same op with ``sliding_window=0``. Reference baseline
                      that does the full GEMM workload.
  3. ``gluon_swa``  - ``torch.ops.aiter.pa_decode_gluon`` with native
                      ``sliding_window`` (kernel re-bases partitions to
                      ``[ctx - sw, ctx)``, so its workspace is
                      ``cdiv(min(ctx, sw), partition_size)`` partitions).

Run::

    python op_tests/bench_pa_v1_swa_vs_gluon.py
    python op_tests/bench_pa_v1_swa_vs_gluon.py --quick
    python op_tests/bench_pa_v1_swa_vs_gluon.py --csv out.csv
"""
from __future__ import annotations

import argparse
import csv
from typing import List, Optional, Sequence, Tuple

import torch
from einops import rearrange

import aiter  # noqa: F401  -- registers torch.ops.aiter.paged_attention_v1
import aiter.ops.attention  # noqa: F401  -- registers torch.ops.aiter.pa_decode_gluon
from aiter import dtypes

_PARTITION_SIZE_ROCM = 256
_PARTITION_SIZE_GLUON = 256
UNIFORM_RANGE = (-1.0, 1.0)


def _api_sliding_window(window_size: int) -> int:
    return window_size if window_size > 0 else 0


def make_caches(
    num_blocks: int,
    block_size: int,
    num_kv_heads: int,
    head_size: int,
    dtype: torch.dtype,
    device: str,
) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
    """Return (k_5d, v_4d, k_aiter_hnd, v_aiter_hnd).

    * ``k_5d`` / ``v_4d`` are the vLLM-style internal paged layout consumed
      by ``pa_decode_gluon``:
        - ``k_5d``: ``[blocks, kv_heads, head_size//x, block_size, x]``
        - ``v_4d``: ``[blocks, kv_heads, head_size, block_size]``
    * ``k_aiter_hnd`` / ``v_aiter_hnd`` are the HND view used by
      ``paged_attention_v1`` with ``kv_cache_layout='HND'``:
        - ``[blocks, kv_heads, block_size, head_size]``
      They are populated by reshaping the same underlying values, so both
      kernels see identical KV content.
    """
    x = 16 // dtype.itemsize
    assert head_size % x == 0, f"head_size {head_size} not divisible by x={x}"
    k_5d = torch.empty(
        (num_blocks, num_kv_heads, head_size // x, block_size, x),
        dtype=dtype, device=device,
    ).uniform_(*UNIFORM_RANGE)
    v_4d = torch.empty(
        (num_blocks, num_kv_heads, head_size, block_size),
        dtype=dtype, device=device,
    ).uniform_(*UNIFORM_RANGE)
    k_aiter_hnd = rearrange(k_5d, "b h d1 s d2 -> b h s (d1 d2)").contiguous()
    v_aiter_hnd = rearrange(v_4d, "b h d s -> b h s d").contiguous()
    return k_5d, v_4d, k_aiter_hnd, v_aiter_hnd


def bench_callable(fn, warmup: int = 10, iters: int = 50) -> float:
    """Median wall-clock latency in microseconds via CUDA events."""
    for _ in range(warmup):
        fn()
    torch.cuda.synchronize()
    starts = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
    stops = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
    for i in range(iters):
        starts[i].record()
        fn()
        stops[i].record()
    torch.cuda.synchronize()
    lat_ms = sorted(s.elapsed_time(e) for s, e in zip(starts, stops))
    return lat_ms[len(lat_ms) // 2] * 1e3


# ---------------------------- kernels --------------------------------- #


def v1_pa_swa(
    query: torch.Tensor,
    k_aiter_hnd: torch.Tensor,
    v_aiter_hnd: torch.Tensor,
    block_tables: torch.Tensor,
    cu_query_lens: torch.Tensor,
    seq_lens: torch.Tensor,
    max_seq_len: int,
    output: torch.Tensor,
    workspace_buffer: torch.Tensor,
    k_scale: torch.Tensor,
    v_scale: torch.Tensor,
    scale: float,
    window_size: int,
) -> None:
    torch.ops.aiter.paged_attention_v1(
        output,
        workspace_buffer,
        query,
        k_aiter_hnd,
        v_aiter_hnd,
        scale,
        block_tables,
        cu_query_lens,
        seq_lens,
        max_seq_len,
        None,           # alibi_slopes
        "auto",         # kv_cache_dtype
        "HND",          # kv_cache_layout
        0.0,            # logits_soft_cap
        k_scale,
        v_scale,
        None,           # fp8_out_scale
        _PARTITION_SIZE_ROCM,
        1,              # mtp
        _api_sliding_window(window_size),
    )


def gluon_pa_swa(
    query: torch.Tensor,
    k_5d: torch.Tensor,
    v_4d: torch.Tensor,
    block_tables: torch.Tensor,
    seq_lens: torch.Tensor,
    output: torch.Tensor,
    exp_sums: torch.Tensor,
    max_logits: torch.Tensor,
    temporary_output: torch.Tensor,
    qs: torch.Tensor,
    ks: torch.Tensor,
    vs: torch.Tensor,
    scale: float,
    window_size: int,
    max_ctx_part: int,
    compute_type: torch.dtype,
) -> None:
    torch.ops.aiter.pa_decode_gluon(
        output,
        query,
        k_5d,
        v_4d,
        seq_lens,
        block_tables,
        scale,
        1,                                    # query_length
        max_ctx_part,
        _PARTITION_SIZE_GLUON,
        compute_type,
        qs, ks, vs,
        exp_sums=exp_sums,
        max_logits=max_logits,
        temporary_output=temporary_output,
        alibi_slopes=None,
        sinks=None,
        sliding_window=_api_sliding_window(window_size),
        ps=False,
    )


# ---------------------------- driver --------------------------------- #


def bench_one(
    num_seqs: int,
    num_q_heads: int,
    num_kv_heads: int,
    head_size: int,
    block_size: int,
    ctx_len: int,
    window_size: int,
    dtype: torch.dtype,
    device: str,
    warmup: int,
    iters: int,
) -> dict:
    torch.manual_seed(0)
    max_blocks_per_seq = (ctx_len + block_size - 1) // block_size
    num_blocks = max_blocks_per_seq * num_seqs

    query = torch.empty(num_seqs, num_q_heads, head_size, dtype=dtype, device=device)
    query.uniform_(*UNIFORM_RANGE)

    k_5d, v_4d, k_aiter_hnd, v_aiter_hnd = make_caches(
        num_blocks, block_size, num_kv_heads, head_size, dtype, device,
    )

    block_tables = torch.randperm(num_blocks, dtype=dtypes.i32, device=device)
    block_tables = block_tables[: num_seqs * max_blocks_per_seq].view(
        num_seqs, max_blocks_per_seq,
    )
    seq_lens = torch.full((num_seqs,), ctx_len, dtype=torch.int32, device=device)
    cu_query_lens = torch.arange(0, num_seqs + 1, dtype=torch.int32, device=device)

    scale = 1.0 / (head_size ** 0.5)
    output = torch.empty_like(query)

    # v1 workspace (sized by full ctx -- v1 partitions count from token 0)
    v1_max_parts = (ctx_len + _PARTITION_SIZE_ROCM - 1) // _PARTITION_SIZE_ROCM
    nbytes_per_qo_elem = torch.finfo(output.dtype).bits // 8
    workspace_buffer = torch.empty(
        (num_seqs * num_q_heads * v1_max_parts * head_size) * nbytes_per_qo_elem
        + 2 * (num_seqs * num_q_heads * v1_max_parts) * 4,
        dtype=torch.uint8, device=device,
    )

    k_scale = torch.tensor([1.0], dtype=dtypes.fp32, device=device)
    v_scale = torch.tensor([1.0], dtype=dtypes.fp32, device=device)

    # gluon workspace (sized by sw -- gluon kernel re-bases partitions
    # to [ctx - sw, ctx) when SWA is enabled)
    sw_for_gluon_workspace = (
        min(ctx_len, window_size) if window_size > 0 else ctx_len
    )
    gluon_max_parts = (
        sw_for_gluon_workspace + _PARTITION_SIZE_GLUON - 1
    ) // _PARTITION_SIZE_GLUON
    query_group_size = num_q_heads // num_kv_heads
    inter = (num_seqs, num_kv_heads, gluon_max_parts, query_group_size)
    exp_sums = torch.empty(inter, dtype=torch.float32, device=device)
    max_logits = torch.empty(inter, dtype=torch.float32, device=device)
    temporary_output = torch.empty(*inter, head_size, dtype=dtype, device=device)
    qs = torch.tensor([1.0], dtype=torch.float32, device=device)

    def fn_v1_swa() -> None:
        v1_pa_swa(
            query, k_aiter_hnd, v_aiter_hnd, block_tables, cu_query_lens,
            seq_lens, ctx_len, output, workspace_buffer, k_scale, v_scale,
            scale, window_size,
        )

    def fn_v1_full() -> None:
        v1_pa_swa(
            query, k_aiter_hnd, v_aiter_hnd, block_tables, cu_query_lens,
            seq_lens, ctx_len, output, workspace_buffer, k_scale, v_scale,
            scale, -1,
        )

    def fn_gluon_swa() -> None:
        gluon_pa_swa(
            query, k_5d, v_4d, block_tables, seq_lens, output,
            exp_sums, max_logits, temporary_output,
            qs, k_scale, v_scale,
            scale, window_size, gluon_max_parts, dtype,
        )

    t_v1_swa = bench_callable(fn_v1_swa, warmup, iters)
    t_v1_full = bench_callable(fn_v1_full, warmup, iters)
    try:
        t_gluon_swa = bench_callable(fn_gluon_swa, warmup, iters)
    except Exception as exc:  # noqa: BLE001
        print(f"  [gluon failed: {exc}]")
        t_gluon_swa = float("nan")

    return dict(
        num_seqs=num_seqs,
        num_q_heads=num_q_heads,
        num_kv_heads=num_kv_heads,
        head_size=head_size,
        block_size=block_size,
        ctx_len=ctx_len,
        window_size=window_size,
        dtype=str(dtype).replace("torch.", ""),
        v1_swa_us=t_v1_swa,
        v1_full_us=t_v1_full,
        gluon_swa_us=t_gluon_swa,
        v1_swa_vs_gluon=(
            t_gluon_swa / t_v1_swa
            if t_gluon_swa == t_gluon_swa
            else float("nan")
        ),
        v1_swa_vs_full=t_v1_full / t_v1_swa,
    )


def main(argv: Optional[Sequence[str]] = None) -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--quick", action="store_true",
                        help="Smaller config grid for sanity checks.")
    parser.add_argument("--csv", type=str, default=None,
                        help="Append CSV results to this file.")
    parser.add_argument("--warmup", type=int, default=10)
    parser.add_argument("--iters", type=int, default=50)
    parser.add_argument("--device", default="cuda:0")
    args = parser.parse_args(argv)

    torch.set_default_device(args.device)

    head_configs: List[Tuple[int, int, int]] = [
        (16, 4, 128),
        (32, 4, 128),
    ]
    block_sizes = [16]
    dtypes_to_test = [dtypes.bf16]

    if args.quick:
        ctx_lens = [4096, 16384]
        window_sizes = [512, 4096]
        num_seqs_list = [16]
    else:
        ctx_lens = [2048, 4096, 8192, 16384]
        window_sizes = [512, 2048, 8192]
        num_seqs_list = [1, 16, 32]

    rows: List[dict] = []
    header_printed = False
    for dtype in dtypes_to_test:
        for num_q, num_kv, hs in head_configs:
            for bs in block_sizes:
                for ns in num_seqs_list:
                    for ctx in ctx_lens:
                        for win in window_sizes:
                            if win >= ctx:
                                continue   # SWA collapses to full causal
                            try:
                                row = bench_one(
                                    num_seqs=ns,
                                    num_q_heads=num_q,
                                    num_kv_heads=num_kv,
                                    head_size=hs,
                                    block_size=bs,
                                    ctx_len=ctx,
                                    window_size=win,
                                    dtype=dtype,
                                    device=args.device,
                                    warmup=args.warmup,
                                    iters=args.iters,
                                )
                            except Exception as exc:  # noqa: BLE001
                                print(
                                    f"[skip] ns={ns} qh={num_q} kvh={num_kv} "
                                    f"hs={hs} bs={bs} ctx={ctx} win={win}: {exc}"
                                )
                                continue
                            rows.append(row)
                            if not header_printed:
                                print(
                                    f"{'ns':>3} {'qh':>3} {'kvh':>3} {'hs':>3} "
                                    f"{'bs':>2} {'ctx':>6} {'win':>5} "
                                    f"{'v1_swa(us)':>11} {'v1_full(us)':>12} "
                                    f"{'gluon_swa(us)':>14} "
                                    f"{'gluon/v1_swa':>13} {'v1_swa/full':>12}"
                                )
                                header_printed = True
                            print(
                                f"{row['num_seqs']:>3} {row['num_q_heads']:>3} "
                                f"{row['num_kv_heads']:>3} {row['head_size']:>3} "
                                f"{row['block_size']:>2} {row['ctx_len']:>6} "
                                f"{row['window_size']:>5} "
                                f"{row['v1_swa_us']:>11.2f} "
                                f"{row['v1_full_us']:>12.2f} "
                                f"{row['gluon_swa_us']:>14.2f} "
                                f"{row['v1_swa_vs_gluon']:>12.2f}x "
                                f"{row['v1_swa_vs_full']:>11.2f}x"
                            )

    if args.csv and rows:
        with open(args.csv, "w", newline="") as fh:
            w = csv.DictWriter(fh, fieldnames=list(rows[0].keys()))
            w.writeheader()
            w.writerows(rows)
        print(f"\nWrote {len(rows)} rows to {args.csv}")

    if rows:
        valid_gluon = [
            r["v1_swa_vs_gluon"]
            for r in rows
            if r["v1_swa_vs_gluon"] == r["v1_swa_vs_gluon"]
        ]
        valid_full = [r["v1_swa_vs_full"] for r in rows]
        if valid_gluon:
            print(f"\nSummary over {len(rows)} configs:")
            print(
                f"  v1_swa vs gluon_swa  geomean ratio: "
                f"{torch.tensor(valid_gluon).log().mean().exp().item():.2f}x "
                f"(min={min(valid_gluon):.2f}x, max={max(valid_gluon):.2f}x) "
                f"-- >1 means v1 is faster"
            )
            print(
                f"  v1_swa vs v1_full   geomean speedup: "
                f"{torch.tensor(valid_full).log().mean().exp().item():.2f}x "
                f"(min={min(valid_full):.2f}x, max={max(valid_full):.2f}x) "
                f"-- partition early-out gain"
            )

    return 0


if __name__ == "__main__":
    raise SystemExit(main())

```

fsx950223:
SWA performance droped with context length increased, which is not expected.
 ns  qh kvh  hs bs    ctx   win  v1_swa(us)  v1_full(us)  gluon_swa(us)  gluon/v1_swa  v1_swa/full
  1  16   4 128 16   8192  2048       67.56        67.44         139.52         2.07x        1.00x
  1  16   4 128 16  65536  2048       66.88        81.48         136.56         2.04x        1.22x
 32  16   4 128 16   8192  2048       95.28       260.96         139.60         1.47x        2.74x
 32  16   4 128 16  65536  2048      204.24      1965.57         140.24         0.69x        9.62x
128  16   4 128 16   8192  2048      320.92       991.12         337.32         1.05x        3.09x
128  16   4 128 16  65536  2048      743.08      7821.15         341.56         0.46x       10.53x
  1  32   4 128 16   8192  2048       66.72        65.56         137.56         2.06x        0.98x
  1  32   4 128 16  65536  2048       66.68        82.48         136.60         2.05x        1.24x
 32  32   4 128 16   8192  2048      100.48       269.88         139.56         1.39x        2.69x
 32  32   4 128 16  65536  2048      230.28      2005.25         138.72         0.60x        8.71x
128  32   4 128 16   8192  2048      347.44      1034.64         352.12         1.01x        2.98x
128  32   4 128 16  65536  2048      844.00      8058.15         350.16         0.41x        9.55x
For pa gluon, there is performance issue with bf16 compute type, maybe you could port same design to the hip pa.

ChengYao-amd:
> SWA performance droped with context length increased, which is not expected. ns qh kvh hs bs ctx win v1_swa(us) v1_full(us) gluon_swa(us) gluon/v1_swa v1_swa/full 1 16 4 128 16 8192 2048 67.56 67.44 139.52 2.07x 1.00x 1 16 4 128 16 65536 2048 66.88 81.48 136.56 2.04x 1.22x 32 16 4 128 16 8192 2048 95.28 260.96 139.60 1.47x 2.74x 32 16 4 128 16 65536 2048 204.24 1965.57 140.24 0.69x 9.62x 128 16 4 128 16 8192 2048 320.92 991.12 337.32 1.05x 3.09x 128 16 4 128 16 65536 2048 743.08 7821.15 341.56 0.46x 10.53x 1 32 4 128 16 8192 2048 66.72 65.56 137.56 2.06x 0.98x 1 32 4 128 16 65536 2048 66.68 82.48 136.60 2.05x 1.24x 32 32 4 128 16 8192 2048 100.48 269.88 139.56 1.39x 2.69x 32 32 4 128 16 65536 2048 230.28 2005.25 138.72 0.60x 8.71x 128 32 4 128 16 8192 2048 347.44 1034.64 352.12 1.01x 2.98x 128 32 4 128 16 65536 2048 844.00 8058.15 350.16 0.41x 9.55x For pa gluon, there is performance issue with bf16 compute type, maybe you could port same design to the hip pa.

I updated the code at the latest commit, Seems the performance droped issue fixed.


| ctx    | win  | v1_swa (us) | v1_full (us) | gluon_swa (us) | gluon / v1_swa | v1_swa / v1_full |
|-------:|-----:|------------:|-------------:|---------------:|---------------:|-----------------:|
| 4096   | 256  | 54.40       | 56.00        | 81.72          | 1.50x          | 1.03x            |
| 4096   | 512  | 57.89       | 59.77        | 108.08         | 1.87x          | 1.03x            |
| 4096   | 2048 | 61.01       | 60.49        | 105.44         | 1.73x          | 0.99x            |
| 8192   | 512  | 57.73       | 68.02        | 108.28         | 1.88x          | 1.18x            |
| 16384  | 512  | 57.41       | 157.71       | 107.56         | 1.87x          | 2.75x            |
| 32768  | 512  | 57.33       | 299.24       | 109.64         | 1.91x          | 5.22x            |
| 65536  | 512  | 57.73       | 654.17       | 108.16         | 1.87x          | 11.33x           |

fsx950223:
LGTM

fsx950223 (dismissed):

zufayu:
@ChengYao-amd This PR has been inactive for over 2 months. Is the page-attention with sliding window still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3313, opened 2026-05-22
Zhimding/fix aot moe pad 0522

coderfeli:
Zhimding/fix aot moe pad 0522. Enable aot for dim pad


aiter/aot/flydsl/moe.py
```diff
@@ -71,6 +71,21 @@ def _row_swiglu_limit(row: dict[str, str]) -> float:
     return _parse_optional_float(row.get("swiglu_limit"), "swiglu_limit") or 0.0
 
 
+def _row_optional_int(row: dict[str, str], name: str, *aliases: str) -> int:
+    for key in (name, *aliases):
```
coderfeli: Carrying the typo `hiddne_pad` as a permanent alias bakes the bug into the schema forever. Prefer a one-shot rename of the CSV column + code, and drop the alias. The runtime cache key is `(model_dim_pad, inter_dim_pad)` ints, so renaming the CSV column does not cause cache misses.

aiter/aot/flydsl/moe.py
```diff
@@ -747,6 +772,8 @@ def compile_one_config(
     shape_str = (
         f"{kernel_name}  "
         f"model_dim={model_dim} inter_dim={inter_dim} "
+        f"hidden_pad={kwargs.get('hidden_pad', 0)} "
```
coderfeli: `hidden_pad` / `intermediate_pad` are already declared as explicit kwargs on `_precompile_to_cache`; reading them back through `kwargs.get(...)` here is inconsistent. Either accept them as named params on `compile_one_config` or pass them through a single dict — picking one keeps the call graph honest.

aiter/fused_moe.py
```diff
@@ -880,12 +880,27 @@ def get_2stage_cfgs(
...
     ]
 
+    def _normalize_pad_cols(df):
+        if "hidden_pad" not in df.columns:
+            if "hiddne_pad" in df.columns:
+                df["hidden_pad"] = df["hiddne_pad"]
+            else:
+                df["hidden_pad"] = 0
+        df["hidden_pad"] = df["hidden_pad"].fillna(0).astype(int)
+        if "intermediate_pad" not in df.columns:
```
coderfeli: Three places now have to know about the pad columns (`_normalize_pad_cols`, `_FILL_DEFAULTS`, dedup keys). Consider a single `PAD_COLUMNS = ("hidden_pad", "intermediate_pad")` constant and iterate, so future pad fields don't have to be added in N spots.

aiter/jit/core.py
```diff
@@ -215,7 +215,14 @@ def update_config_files(self, file_path: str, merge_name: str):
                 f"when merging '{merge_name}'."
             )
 
-        _FILL_DEFAULTS = {"xbf16": 0, "run_1stage": 0, "ksplit": 0}
+        _FILL_DEFAULTS = {
```
coderfeli, line 218: Same comment as on the AOT side: please don't enshrine `hiddne_pad` in `_FILL_DEFAULTS`. Fix the CSV(s) and drop the misspelled key.

op_tests/test_moe_2stage.py
```diff
@@ -81,15 +88,27 @@ def test_fmoe(
             w1[:, -intermediate_pad:, :] = 0
             w1[:, inter_dim - intermediate_pad : inter_dim, :] = 0
         exp_bias1 = torch.clamp(torch.randn((E, inter_dim * 2), dtype=dtype), -1.0, 1.0)
+        # Dense torch reference still evaluates padded lanes; keep padded
```
coderfeli: Zeroing padded bias lanes in the *reference* path is masking a semantic mismatch — the dense torch reference still evaluates padded lanes, while the kernel does not. A cleaner approach: build the reference at logical dims (`model_dim - hidden_pad`, `inter_dim - intermediate_pad`), then pad to physical dims only for the kernel input. That way you don't need to patch bias/output to make the compare pass.

op_tests/test_moe_2stage.py
```diff
@@ -71,6 +71,13 @@ def test_fmoe(
 ):
     if get_gfx() not in ["gfx950"] and qType in [aiter.QuantType.per_1x32]:
         return
+    assert (
```
coderfeli, line 82: `assert 0 <= hidden_pad < model_dim` allows nonsense values like `model_dim - 1`. The real constraint is probably `(model_dim - hidden_pad) % tile_n == 0` (or similar tile alignment). Consider a stronger check, or at minimum a comment stating which invariant this guards.

op_tests/test_moe_2stage.py
```diff
@@ -359,9 +378,12 @@ def weight_per_128x128_quant(weight, quant_dtype):
         num_iters=5,
         num_warmup=2,
     )
+    valid_model_dim = model_dim - hidden_pad
+    out2_ref_check = out2_ref[:, :valid_model_dim]
+    out2_ck_check = out2_ck[:, :valid_model_dim]
     err = checkAllclose(
```
coderfeli, line 397: Slicing both ref and ck to `valid_model_dim` here is fine, but combined with the bias zeroing above it's a double patch. If the reference is rebuilt at logical dims (see comment on bias), this slice goes away.

coderfeli (commented):
Plumbing looks fine, but the pad column handling and the test-reference patching feel hacky. A few inline comments.

aiter/configs/model_configs/gptoss_fp4_tuned_fmoe.csv
```diff
@@ -1,25 +1,25 @@
...
-256,16384,3072,512,128,4,ActivationType.Swiglu,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,64,0,206.4827,flydsl_moe1_afp4_wfp4_bf16_t64x128x256_w4_bnt0_xcd4,0.0%,285.1907,flydsl_moe2_afp4_wfp4_bf16_t64x256x256_reduce_xcd4,0.6%,491.6734,0,0,1257.9,1535.52,
-256,32768,3072,512,128,4,ActivationType.Swiglu,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,64,0,367.6363,flydsl_moe1_afp4_wfp4_bf16_t64x128x256_w2_bnt0_xcd4,0.0%,517.2156,flydsl_moe2_afp4_wfp4_bf16_t64x256x256_reduce_xcd4,0.6%,884.8519,0,0,1397.92,1023.87,
+cu_num,token,model_dim,inter_dim,expert,topk,act_type,dtype,q_dtype_a,q_dtype_w,q_type,use_g1u1,doweight_stage1,block_m,ksplit,us1,kernelName1,err1,us2,kernelName2,err2,us,run_1stage,xbf16,hidden_pad,intermediate_pad,tflops,bw,_tag
+256,256,3072,3072,128,4,ActivationType.Swiglu,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,208.4462,flydsl_moe1_afp4_wfp4_bf16_t32x128x256_w2,0.0%,109.1431,flydsl_moe2_afp4_wfp4_bf16_t32x256x256_atomic_bnt2_persist,5.0%,317.5893,0,0,192,192,182.57,11418.01,
```
coderfeli: All 24 rows use the same 3 constants for the pad columns (192 / {192,96,152}). If pad is a function of `model_dim` / `inter_dim` + tile alignment, it should be derived in code rather than duplicated in the tuned CSV — otherwise every retune has to remember to fill these columns consistently.

PR #3319, opened 2026-05-24
[triton] kv_cache.reshape_and_cache (HND default, NHD option) + HIP→triton fallback in cache.reshape_and_cache

carlushuang:
## Summary

Three consolidated changes to make aiter's KV-cache scatter portable across arches:

### 1. New triton wrapper `aiter.ops.triton.kv_cache.reshape_and_cache`

Scatters `(key, value)` tokens into a paged KV cache at the slots given by `slot_mapping`. Slots equal to a negative value are skipped (no write) **inside the kernel** — callers can pre-fill `slot_mapping` with `-1` for padded positions without a Python-side branch, making the path CUDAGraph-safe.

Until now there was no triton-side `reshape_and_cache` for non-MLA paged GQA layouts — callers had to use `cat_and_cache_mla` (MLA-specific) or the HIP CK module `module_cache.reshape_and_cache` (prebuilt for CDNA only — SIGSEGVs on RDNA arches without a code object).

### 2. Layout option: `kv_cache_layout='HND'` (default) or `'NHD'`

The underlying `_reshape_and_cache_kernel` is already stride-based (uses `cache_stride_block` / `_head` / `_within` and indexes the cache via strides) — it doesn't care about tensor-dim ordering. Only the wrapper hard-coded HND. Same pattern as `aiter.ops.triton.rope.fused_qkv_split_qk_norm_rope_cache`:

```
HND (default): [num_blocks, num_kv_heads, block_size, head_dim]
NHD:           [num_blocks, block_size, num_kv_heads, head_dim]
```

HND is the default to preserve the kernel's pre-PR behavior. NHD is opt-in for callers whose paged cache is laid out for `aiter.ops.triton.attention.unified_attention.unified_attention` — that kernel reads `k.shape[1]` as `block_size` and `k.shape[2]` as `num_kv_heads` (NHD), so any caller that writes K/V into an NHD cache (e.g. ATOM PR ROCm/ATOM#811's `PagedAttentionImpl.forward` flash-layout branch) needs a layout-matched writer.

### 3. HIP → triton fallback in `aiter.ops.cache.reshape_and_cache`

The HIP CK module `module_cache` ships prebuilt code objects only for gfx94x/95x. On other arches (e.g. gfx1201 RDNA4 in `rocm/atom-dev:latest`) the HIP kernel launch SIGSEGVs at first forward — no catchable exception, so we allowlist arches **before** calling the HIP path. Same pattern as `aiter.ops.gemm_op_a8w8` PR #3343.

Adds `_CACHE_HIP_PREBUILT_ARCHES = {gfx940, gfx941, gfx942, gfx950}` + `_hip_cache_supported()` helper. When not supported, routes to the triton wrapper above (HND layout — the HIP CK and triton wrappers use the same layout, so this is a direct dispatch). Fallback constraints (asserted):
- `k_scale` / `v_scale` must be `None` (triton variant doesn't yet support FP8 KV scales)
- `asm_layout` must be `False` (CDNA-specific cache shape)

Covers the BF16 KV path. FP8 / asm_layout on non-CDNA arches is out of scope here.

## Why bundled

All three changes are about `reshape_and_cache` — adding a triton wrapper, exposing the existing layout flexibility, and letting the HIP entry point fall back to it on non-CDNA. Single coherent story; previously split into PRs #3319 / #3345 / #3351 which formed a chain (each based on the previous). Consolidated to avoid the review serialization.

## Tests

- `op_tests/triton_tests/test_kv_cache.py`: correctness vs torch reference scatter across HND layout (multiple shapes / dtypes / mixed-sign slot_mapping).
- Verified on gfx1201: NHD writes byte-identical to HND-then-permute; HIP wrapper auto-falls-back to triton; output matches torch reference.
- No CDNA behavior change: `_hip_cache_supported()` returns True for gfx94x/95x and the HIP CK path runs unchanged.

## Notes

A previous version of this PR also relaxed `paged_attention_decode`'s `k_scale` / `v_scale` to `Union[float, torch.Tensor]` for CUDAGraph-safety on ATOM's old gfx1201 attention path. **Reverted** in response to @valarLip's review — pa_decode keeps the `torch.Tensor`-only API. ATOM's gfx1201 path now goes through `unified_attention` instead of pa_decode, so that constraint is no longer load-bearing here.


aiter/ops/triton/attention/pa_decode.py
```diff
@@ -41,8 +41,8 @@ def paged_attention_decode(
     attn_scale: float,
     max_seq_len: int,
     compute_type,
-    k_scale: torch.Tensor,
-    v_scale: torch.Tensor,
+    k_scale: Union[float, torch.Tensor],
```
valarLip: no float.... keep tensor...

valarLip (changes requested):

aiter/ops/triton/attention/pa_decode.py
carlushuang: Thanks @valarLip. To check I understand: the goal is to drop the `float` from the type union so callers only pass tensors? The Union was added specifically to give CUDAGraph-safe callers a way to skip the `.item()` GPU→CPU sync (illegal inside graph capture). Without that path, callers inside a CUDAGraph have to either pre-resolve the scale outside the graph or accept a sync at first capture. Is there a different approach you had in mind for CUDAGraph safety — e.g. plumbing the scale tensor through to a `tl.load` inside the triton kernel? Happy to refactor either way once the direction is clear.

carlushuang: Done — reverted in ba35eef. `pa_decode` keeps the `torch.Tensor` k_scale / v_scale signature, and the float-path tests are dropped. PR title and description updated to reflect the trimmed scope (now just the `kv_cache.reshape_and_cache` triton wrapper).

carlushuang: addressed

carlushuang:
Superseded — closing.

The gfx1201 KV-cache write doesn't need a separate triton writer after all. aiter's native HIP `reshape_and_cache(asm_layout=True)` (in `module_cache`) writes the 5D-SHUFFLE layout correctly on gfx1201 — the kernel is arch-clean and validated bitwise vs a torch reference. A from-source build on gfx1201 already produces a working `module_cache`.

The actual bug was in the JIT loader: it imports a wrong-arch (CDNA-only) prebuilt `module_cache.so` instead of rebuilding for the running GPU, so it SIGSEGVs at launch. Tracked + fixed generally:
- Issue: ROCm/aiter#3846 (clean repro)
- Fix: ROCm/aiter#3839 (arch-coverage guard → JIT-rebuild when the prebuilt lacks the running arch)

ATOM PR ROCm/ATOM#811 has been updated to depend on #3846/#3839 instead of this PR. The dedicated HIP-module alternative (#3837) is being closed for the same reason.

PR #3321, opened 2026-05-24
[FlyDSL AOT] Skip kernels for unrequested arches when GPU_ARCHS is set

eppaneamd:
## Summary

When `GPU_ARCHS` is set to a specific arch at build time (e.g. `gfx942`), FlyDSL AOT previously compiled all kernels unconditionally, including hundreds of kernels tuned for other arches (that will not be used).

- Add `_job_arch(job)` helper that returns the target arch from any job dict (`cu_num` → `cu_num_to_arch` for GEMM/MoE; explicit `"arch"` field for CHUNK_GDN_H; `None` for untuned/arch-agnostic jobs that must always compile).
- In `start_aot()`, filter `all_jobs` against `GPU_ARCHS` after collection. Uses `_parse_gpu_archs_env` from `build_targets.py` (`;`-separated, consistent with the rest of the codebase). Import is deferred inside the branch to avoid triggering `aiter/__init__` during `setup.py`'s early import of `common.py`.
- `GPU_ARCHS` unset or `"native"` preserves existing behaviour.

## Test plan

- Unit-tested `_job_arch` and filter logic: tuned GEMM/MoE, CHUNK_GDN_H explicit arch, untuned jobs, single arch, multi-arch (`gfx942;gfx950`), unset, and `native`.
- Verified end-to-end on AITER `main` with `GPU_ARCHS=gfx942`: filter fires correctly, 2001 gfx950-only kernels skipped, build completes without errors.


coderfeli:
@zhiding512  take a look?

PR #3330, opened 2026-05-25
ew kernels support on gfx12.

stevenshenyj:
## Motivation

Enable building and running [aiter] on RDNA4 (gfx1201, e.g. Radeon RX 9700) GPUs. Several kernels currently fail to compile or dispatch on gfx1201:

v_pk_mul_f32 in fused-qk-norm kernel — this packed FP32 SIMD instruction exists on CDNA (gfx9xx) but is not available on RDNA4, causing a backend assembler error when building for gfx1201.

v_cmp_u_f32 with explicit SGPR dest + v_add3_u32 in [rope_common.h]— the [f32x2_to_bf16x2_rne()] function uses CDNA-specific inline assembly (explicit SGPR destination for v_cmp, 3-operand integer add). RDNA4's v_cmp can only write to implicit VCC, and v_add3_u32 does not exist, causing "invalid operand for instruction" errors.

fmha_v3 ASM kernels — hand-written gfx9 assembly (gfx942/gfx950 only). On non-gfx9 hosts, can_impl_fmha_v3_fwd() unconditionally returned True, causing dispatch failures ("invalid argument for fmha_fwd") instead of falling back to the ck-tile path.

ck-tile MHA generate.py targets — defaults to --targets gfx9,gfx950, so non-gfx9 hosts get an empty kernel set and mha_fwd fails at dispatch.

As a result, downstream users (sglang on Radeon) cannot use aiter's fused-qk-norm-rope, fused-qk-norm-rope-cache-quant, or flash-attention paths at all on RDNA4.

## Technical Details

csrc/kernels/fused_qk_norm.cu: Added [#if defined(__gfx906__) || ... || defined(__gfx950__)]positive-enumeration guard at both v_pk_mul_f32 sites. CDNA archs keep the asm fast path; all other archs fall back to portable element-wise multiplies (v_mul_f32 x2).

csrc/kernels/rope/rope_common.h: Added arch guard around the [f32x2_to_bf16x2_rne()] inline asm block. The CDNA path uses v_cmp_u_f32 (explicit SGPR dest) + v_add3_u32 + v_and_or_b32 for high-throughput packed bf16 conversion. The #else path implements identical RNE + NaN logic in portable C++ (compiler emits standard RDNA scalar instructions).

aiter/ops/mha.py: can_impl_fmha_v3_fwd() now gates on get_gfx() in ("gfx942", "gfx950") before enabling the v3 ASM path. Non-gfx9 archs cleanly fall through to the ck-tile FmhaFwd implementation.

aiter/jit/utils/mha_recipes.py: Added _ck_targets_flag() helper that detects runtime GPU arch. For non-gfx9 hosts, it appends --targets <runtime arch> (e.g. --targets gfx1201) to ck-tile's generate.py invocations, ensuring the correct kernel variants are generated instead of defaulting to gfx9-only.

## Files changed:
aiter_pr1

All changes are guarded by positive CDNA arch enumeration macros or runtime get_gfx() checks, so:
gfx1201 builds: take the scalar/portable path (compiles and runs); MHA dispatches via ck-tile.
gfx9 builds (gfx942/gfx950): bit-for-bit unchanged (same asm / same fmha_v3 as before).

## Submission Checklist

### DIFF
--- a/aiter/jit/utils/mha_recipes.py
+++ b/aiter/jit/utils/mha_recipes.py
@@ -1,6 +1,25 @@
 from typing import List, Dict, Tuple
 
 
+def _ck_targets_flag() -> str:
+    """Return ``--targets <runtime arch>`` when the runtime GPU is not gfx9.
+
+    ck-tile's ``generate.py`` defaults to ``--targets gfx9,gfx950``, so any
+    non-gfx9 host (gfx10/11/12) ends up with an empty kernel set and ``mha_fwd``
+    fails at dispatch with "invalid argument for fmha_fwd". For gfx9 hosts we
+    keep the default (covers both gfx942 and gfx950 like before).
+    """
+    try:
+        from chip_info import get_gfx
+
+        gfx = get_gfx()
+    except Exception:
+        return ""
+    if gfx.startswith("gfx9"):
+        return ""
+    return f" --targets {gfx}"
+
+
 def compose_mha_fwd_variant_suffix_and_filter(
     dtype: str,
     logits_positive: bool,
@@ -104,8 +123,8 @@ def get_mha_varlen_prebuild_variants_by_names(
             has_qscale=has_qscale,
         )
         blob_gen_cmd = [
-            f"{ck_dir}/example/ck_tile/01_fmha/generate.py -d fwd --receipt {receipt} --filter {filter_pattern} --output_dir {{}}",
-            f'{ck_dir}/example/ck_tile/01_fmha/generate.py -d fwd_splitkv --receipt {receipt} --filter " @ " --output_dir {{}}',
+            f"{ck_dir}/example/ck_tile/01_fmha/generate.py -d fwd --receipt {receipt} --filter {filter_pattern} --output_dir {{}}{_ck_targets_flag()}",
+            f'{ck_dir}/example/ck_tile/01_fmha/generate.py -d fwd_splitkv --receipt {receipt} --filter " @ " --output_dir {{}}{_ck_targets_flag()}',
         ]
         variants.append(
             {"md_name": f"mha_varlen_fwd{suffix}", "blob_gen_cmd": blob_gen_cmd}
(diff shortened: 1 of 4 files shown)


csrc/include/aiter_opus_plus.h
```diff
@@ -23,7 +23,16 @@ using index_t = int;
...
     fp32x2_t c;
+    // v_pk_mul_f32 is a CDNA2+ (gfx90a/gfx94x/gfx950) packed-FP32 instruction
+    // and is NOT available on RDNA. Currently only gfx1201 (RDNA4) has been
+    // verified to need this fallback; extend the guard here as more unsupported
+    // archs (e.g. other RDNA gfx10xx/gfx11xx/gfx12xx) are encountered.
+#if defined(__gfx1201__)
+    c.x = a.x * b.x;
+    c.y = a.y * b.y;
+#else
     asm volatile("v_pk_mul_f32 %0, %1, %2" : "=v"(c) : "v"(a), "v"(b));
```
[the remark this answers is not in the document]
stevenshenyj: This change has already been removed，please ignore.

junhaha666 (dismissed):

setup.py
```diff
@@ -397,6 +397,18 @@ def build_one_module(one_opt_args):
...
+        # starts from version 0 (unversioned .so names).  The import of
+        # aiter.aot.flydsl.common above may have triggered a module_aiter_core
+        # build with a different hash (no -DPREBUILD_KERNELS); clearing here
+        # avoids a version bump that would rename the output to *_v1.so.
+        # NOTE: core.py imports cpp_extension via bare `from cpp_extension import`
+        # (relying on sys.path), so the authoritative singleton lives under the
+        # "cpp_extension" key in sys.modules, NOT "jit.utils.cpp_extension".
+        import cpp_extension
+
+        cpp_extension.JIT_EXTENSION_VERSIONER.entries.clear()
```
amd-ruitang3: Hi @stevenshenyj , we discussed this with the FlyDSL team. They confirmed the issue has been resolved after merging AOT. Could you double check?

stevenshenyj: already fix the issue by chaning from aiter.jit.utils.chip_info import get_gfx  into from chip_info import get_gfx in mha_recipies.py to avoid redundant module_aiter_core build.

junhaha666 (approved):

PR #3334, opened 2026-05-25
[Triton] SAGE: compensate softmax_lse for K-smoothing shift to enable ring-attention compatibility

ksikiric:
## Motivation

`fav3_sage_wrapper_func` returns a softmax_lse that is computed against the K‑smoothed tensor (K − mean(K)), not the original K. Sage relies on the softmax shift‑invariance identity for the output, so the output is correct, but the LSE is offset by a per‑row constant δᵢ = sm_scale · Qᵢ · mean(K)ᵀ.

This is invisible in single‑call use, but it breaks any FA‑style merge of (out, lse) across shards (ring attention). Each shard sees a different K slice and uses a different local mean, so each returned lse_s has a different offset → wrong merge weights → structured low‑frequency artifacts in the merged output. We hit this in Wan2.2 video generation through xfuser/xdit; severity scales with ring degree.

## Technical Details

With `K = K₀ − c`, `c = mean(K₀, dim=seq)`, `δᵢ = sm_scale · Qᵢ · cᵀ`:

`qk_centered_ij = qk0_ij − δᵢ`
`mᵢ`, probabilities `pᵢⱼ`, and `lᵢ = Σⱼ pᵢⱼ` are unchanged (so `out` is unchanged)
`lse_centered_i = lse0_i − δᵢ`
So adding `δᵢ` to the kernel's LSE recovers the un‑smoothed LSE, making `(out, lse)` consistent with any FA ring merge.

### Changes

- `sage_quant`: new `return_lse: bool = False`. When True, additionally returns `δ` of shape `(B, H_q, S_q)`, fp32, with GQA/MQA broadcasting and both layouts handled.
- `fav3_sage_wrapper_func `/` _FAv3SageWrapperFunc.forward`: new `smooth_k: bool = True` forwarded to `sage_quant`; the wrapper adds `δ` to the kernel's `softmax_lse` when `return_lse=True`.
 
### Additional kernel-side consistency fixes

Aligns two outlier paths in `sage_fwd` (`_no_blocks` early-exit and the
per-row `causal_mask` override for rows above the diagonal) to write
`softmax_lse = -inf` instead of `0`, matching the `-inf` already produced
by the regular `invalid_mask` branch. Also casts `start_m` to `int32` at the `compute_block_masking`
call site to fix a Triton type-unification error under `causal=True`,
mirroring the cast already used in the mxfp4 path. Adds two regression
tests for both empty-row paths.

## Test Plan

E2E Runs with varying degree of ring attention

## Test Result

E2E runs now produce correct output videos for varying degrees of ring attention. 

## Submission Checklist


juuso-oskari (approved):
LGTM

vgokhale (approved):

PR #3337, opened 2026-05-25
CI: cache Triton wheel for standard tests

gyohuangxin:
## Summary
- Download the Triton wheel once in the Aiter test workflow using the same per-run artifact approach as Triton tests.
- Let standard test shards download the Triton wheel artifact and pass it into the container via `TRITON_WHEEL_DIR`.
- Keep the public Triton index fallback when the artifact is unavailable or unusable.

## Test plan
- python3 -c "import yaml; yaml.safe_load(open('.github/workflows/aiter-test.yaml'))"
- git diff --check -- .github/workflows/aiter-test.yaml


mengfei-jiang (approved):
LGTM

PR #3345, opened 2026-05-26
kv_cache: reshape_and_cache accepts kv_cache_layout=HND/NHD

carlushuang:
## Summary

Extend `aiter.ops.triton.kv_cache.reshape_and_cache` with a `kv_cache_layout` argument (`HND` default for backward compat, `NHD` opt-in).

```
HND: [num_blocks, num_kv_heads, block_size, head_dim]   # current default
NHD: [num_blocks, block_size, num_kv_heads, head_dim]   # new
```

NHD is required by callers that read the cache via `aiter.ops.triton.attention.unified_attention.unified_attention` — it indexes `k.shape[1]` as `block_size` and `k.shape[2]` as `num_kv_heads`. Without this option, callers either have to use the (slower) HND-targeted `pa_decode` + `context_attention_fwd` split, or permute the cache per call.

Same pattern as `aiter.ops.triton.rope.fused_qkv_split_qk_norm_rope_cache`, which has accepted both layouts since its introduction.

## Implementation

The underlying `_reshape_and_cache_kernel` is already layout-agnostic — it uses `(cache_stride_block, cache_stride_head, cache_stride_within)` and indexes via strides. Only the wrapper hard-coded HND. The patch teaches the wrapper to unpack shape + extract strides based on layout. Kernel unchanged.

## Test plan

- [x] Correctness on gfx1201: `reshape_and_cache(..., kv_cache_layout="NHD")` writes are **byte-identical** to `reshape_and_cache(..., kv_cache_layout="HND")` followed by `.permute(0, 2, 1, 3)`. Tested across N=8 tokens, BLOCK_SIZE=16, NUM_KV_HEADS=8, HEAD_DIM=128 with mixed positive / negative slots (the skip-on-neg semantics are preserved in both layouts).
- [x] No behavior or perf change for existing HND callers — default value preserves the old signature behavior, kernel unchanged.

## Notes

Based on `carhuang/triton_reshape_and_cache_and_pa_decode_float_scale` (PR #3319), since that PR introduced `reshape_and_cache`. If #3319 lands first, this can be rebased onto main; if this lands first, #3319's reshape_and_cache addition needs to ship with `kv_cache_layout` already in place.

Unblocks an ATOM-side simplification of `native_triton_attn.py`: replace the separate prefill + decode kernel calls with a single `unified_attention` call by allocating the KV cache in NHD and calling `reshape_and_cache(..., kv_cache_layout="NHD")`.


carlushuang:
Superseded by #3319 — folded into that PR as the layout-option commit. Closing to avoid the dependency chain.

PR #3351, opened 2026-05-26
cache: reshape_and_cache HIP → triton fallback for non-CDNA archs

carlushuang:
## Summary

The prebuilt HIP CK module `module_cache` (`aiter.ops.cache.reshape_and_cache`) ships code objects only for gfx94x/95x. On other arches (e.g. gfx1201 RDNA4 in `rocm/atom-dev:latest`) the HIP kernel launch SIGSEGVs at first forward call — no catchable exception, so we must allowlist arches **before** calling the HIP path. Same pattern as `gemm_a8w8_blockscale` (PR #3343 / commit 7644ed6 in `aiter/ops/gemm_op_a8w8.py`).

## Implementation

Adds:
- `_CACHE_HIP_PREBUILT_ARCHES = {gfx940, gfx941, gfx942, gfx950}` (same set as the blockscale GEMM allowlist)
- `_hip_cache_supported()` helper
- `_reshape_and_cache_ck` (renamed direct HIP binding)
- A new public Python wrapper `reshape_and_cache(...)` that dispatches: HIP CK on the allowlisted arches, triton on others.

The triton fallback routes to `aiter.ops.triton.kv_cache.reshape_and_cache` (added in PR #3319), which JIT-compiles for any arch and uses the same HND layout `[num_blocks, num_kv_heads, block_size, head_dim]`.

**Fallback constraints** (asserted, not silently dropped):
- `k_scale` / `v_scale` must be `None` — the triton variant does not yet accept FP8 KV cache scales
- `asm_layout` must be `False` — asm_layout uses a CDNA-specific cache shape (`x`-axis packing) that the triton kernel can't reconstruct

These cover the BF16 KV path. FP8 KV / asm_layout on non-CDNA arches need their own triton-side work.

## Why

Today, ATOM's `PagedAttentionImpl` calls `aiter.reshape_and_cache(...)` directly. On gfx1201 the call SIGSEGVs, forcing ATOM PR #811 to maintain a parallel `NativeTritonBackend` that bypasses `PagedAttentionImpl` entirely. With this fallback in place, the gfx1201 path can start using `PagedAttentionImpl` (or its `TritonMHABackend` wrapper) — for the BF16 + non-asm-layout case at least.

## Test plan

- [x] Standalone on gfx1201: `aiter.reshape_and_cache(...)` (HND layout, BF16 K/V, mixed positive/negative slots) writes correctly. `k_cache[block, :, within, :]` equals input `k[token]`; negative slots skipped.
- [x] No behavior change on CDNA — `_hip_cache_supported()` returns True for gfx94x/95x and the HIP CK path runs unchanged.

## Notes

Depends on PR #3319 (which adds the triton `reshape_and_cache` wrapper). PR base is the #3319 branch — when #3319 lands, rebase onto main.

This is **one piece** of a broader gfx1201 enablement. To actually run ATOM's `TritonMHABackend` on gfx1201, additional HIP-call sites in `PagedAttentionImpl` need similar fallbacks (`reshape_and_cache_with_pertoken_quant`, `asm_layout=True` paths, `fused_qk_rope_reshape_and_cache`, etc.). This PR addresses just the most common BF16-KV reshape_and_cache call.


carlushuang:
Superseded by #3319 — folded into that PR as the HIP→triton fallback commit. Closing to avoid the dependency chain.

PR #3353, opened 2026-05-26
add per-(batch, head) fp8 quant ops for fused QK norm/rope and V

LiuYinfeng01:
- fused_qk_norm_rope_2way_fp8_perhead_quant: 3-pass variant of the 2way fused RoPE+RMSNorm that emits fp8 Q/K plus per-(batch, head) descale tensors of shape [B, H], the layout CK FP8 flash attention accepts natively.
- v_per_head_fp8_quant: per-head fp8 quant for V producing matching [B, H] descale.

These keep the CK fp8 FA asm v3 fast path (which requires descale shape (1,) or [B, H_k]) while giving finer-granularity scales than per-tensor, improving accuracy without forcing the bf16 fallback.

## Submission Checklist


csrc/pybind/v_per_head_fp8_quant_pybind.cu
```diff
@@ -0,0 +1,13 @@
+#include <torch/extension.h>
```
valarLip: maybe switch new style

valarLip (approved):

PR #3369, opened 2026-05-27
fp8 einsum flydsl impl

ganyi1996ppo:
## Summary

  FP8 einsum kernel (flydsl, gfx950) — `Z[b,h,d] = einsum('bhr,hdr->bhd', X, Y)` with DeepGEMM packed-UE8M0 scales
  Lands a from-scratch flydsl implementation of the FP8 multi-head einsum used in attention QK/PV (and matmul-with-batched-rhs patterns), targeting **MI355X**. Both BF16 and
   FP8-with-D-128-group-quant output variants are supported through a single underlying kernel factory plus a unified `fp8_einsum()` user-facing entry point.

  Headline:

  - **1965 TF** at H=16 D=1024 R=4096 B=32768, BF16 output
  - Built-in autotune table covering 17 BF16 + 17 QZ tuned shapes
  - FP8 + per-(B, H, D/128) fp32 scale output in a single launch (no separate quant pass)
  
  ## Interface
  
  ```python 
  def fp8_einsum(
      x_fp8: torch.Tensor,             # (B, H, R) fp8 e4m3 — caller pre-quantizes
      y_pre: torch.Tensor,             # preshuffled per-head, (n0, k0, klane=4, nlane=16, kpack=16)
      sx:    torch.Tensor,             # (B, H, R//512) int32 — 4 packed UE8M0 bytes per i32
      *,
      out_dtype: torch.dtype = torch.bfloat16,   # or torch.float8_e4m3fn
      transpose_scale: bool = False,             # only meaningful for fp8 output
      stream: torch.cuda.Stream | None = None,
  ) -> tuple[torch.Tensor, torch.Tensor | None]:
      """Returns (z, sz). For bf16 output, sz is None. For fp8 output, sz is the fp32
      per-(B, H, D/128) scale tensor in the layout selected by transpose_scale."""
```

  Scale ABI matches DeepGEMM SM100 packed-UE8M0: 4 UE8M0 exponent bytes per consecutive 128-K-col group packed little-endian into one i32.

  Mirrors the preshuffle_gemm async-pingpong structure, adapted to feed packed UE8M0 scales directly into the MFMA opsel byte:

  - async-A: raw_ptr_buffer_load_lds with swizzle-the-source — A gmem→LDS direct, no register transit
  - B, sx_i32, sy_i32 all carried in registers across iters
  - Scale fed DIRECTLY to mfma_scale_f32_16x16x128_f8f6f4 via opsel byte index — no fp32 promote step
  - a0_prefetch: first A pack of NEXT tile pre-loaded from LDS to regs before next-iter MFMA fires (hides 1 ds_read from the inner-loop critical path)
  - Unrolled pingpong: each loop iter handles 2 K-tiles (compute on pong + prefetch ping, then compute on ping + prefetch pong)
  - Single s_waitcnt(vmcnt=N_b+N_sx+N_sy)+lgkmcnt(0) per half AFTER compute_tile
  
  LDS layout:
  - A fp8 ping/pong stages
  - sx slab (persistent, loaded once per WG)
  - sy slab (persistent, loaded once per WG)
  - qz amax slab (only allocated when quant_output=True)

  QZ epilogue (fp8 output mode)

  Per WG: tile_m rows × tile_n/128 D-128 blocks.

  Reduction structure depends on tile_n:
  - tile_n == 128: all 4 waves contribute to the single D-128 block per WG
  - tile_n == 256: waves 0+1 → block 0, waves 2+3 → block 1

  Algorithm per (mi, ii, d128):
  1. Local amax over the lane's num_acc_n accs at this (mi, ii)
  2. In-wave 16-lane reduce via xor butterfly (ds_bpermute × 4)
  3. Cross-wave reduce via LDS slab + barrier (waves_per_d128 partials)
  4. s = max(amax, 1e-30) / 448
  5. Clamp scaled accs to [-448, 448] (cvt_pk_fp8_f32 does NOT saturate — produces NaN above 448), then RNE convert + 1-byte store
  
  OOB-row predicate on scale store: when B < tile_m, padding rows would otherwise compute amax=0 and overwrite valid scales for other d128 blocks due to linear-index wrap.

## Perf

  | B     | tile (m, n, k, bsw) | BF16 ms | BF16 TF | QZ ms | QZ TF |
  |------:|---------------------|--------:|--------:|------:|------:|
  |     1 | (32, 128, 256, 2)   |   0.022 |     6.0 | 0.025 |   5.4 |
  |     2 | (32, 128, 256, 8)   |   0.023 |    11.9 | 0.025 |  10.9 |
  |     4 | (32, 128, 256, 0)   |   0.022 |    24.0 | 0.025 |  21.7 |
  |     8 | (32, 128, 256, 2)   |   0.022 |    47.9 | 0.025 |  43.7 |
  |    16 | (32, 128, 256, 0)   |   0.022 |      97 | 0.025 |    87 |
  |    32 | (32, 128, 256, 8)   |   0.022 |     195 | 0.025 |   173 |
  |    64 | (64, 128, 256, 0)   |   0.022 |     394 | 0.025 |   348 |
  |   128 | (64, 128, 256, 0)   |   0.023 |     761 | 0.026 |   661 |
  |   256 | (128, 128, 256, 0)  |   0.028 |    1230 | 0.034 |  1010 |
  |   512 | (64, 256, 128, 2)   |   0.043 |    1606 | 0.046 |  1502 |
  |  1024 | (64, 256, 128, 4)   |   0.073 |    1875 | 0.078 |  1766 |
  |  4096 | (128, 128, 128, 8)  |   0.291 |    1888 | 0.400 |  1376 |
  |  8192 | (128, 128, 128, 4)  |   0.567 |    1939 | 0.763 |  1441 |
  | 16384 | (128, 128, 128, 4)  |   1.124 |    1957 | 1.509 |  1457 |
  | 32768 | (128, 128, 128, 4)  |   2.238 |    1965 | 2.985 |  1473 |

## Submission Checklist

### DIFF
--- a/aiter/ops/flydsl/kernels/fp8_einsum.py
+++ b/aiter/ops/flydsl/kernels/fp8_einsum.py
@@ -1,46 +1,58 @@
-"""DeepGEMM-style clean FP8 einsum for gfx950 — v2 pipeline (preshuffle-style).
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
 
-Reference shape: H=8 D=8192 R=8192 B=2048, tm=128 tn=128 tk=256, bsw=4,
-sx+sy in LDS, B in regs via pingpong → ~1896 TF peak.
+"""Clean, submission-ready FP8 blockscaled batched einsum for gfx950 / MI355X.
+
+Computes  Z[b,h,d] = sum_r X[b,h,r] * Y[h,d,r]   (einsum 'bhr,hdr->bhd')
+in FP8 e4m3 with DeepGEMM-style packed-UE8M0 per-128-K block scales.
+
+This is the "reg-B" path: the proven-correct and peak-performant pipeline.
+  - A (X tile)  : gmem -> LDS via raw_ptr_buffer_load_lds (compiler intrinsic,
+                  no register transit; swizzle-the-source for bank-conflict-free
+                  ds_read). Double-buffered ping/pong across K-tiles.
+  - B (Y tile)  : gmem -> registers via load_b_tile (carried across iters).
+  - sx (X scale): staged once per WG to a small LDS slab, then read per K-tile.
+  - sy (Y scale): gmem -> registers per K-tile (tiny; packed UE8M0).
+  - Scale fed DIRECTLY to the MFMA (packed UE8M0 via the opsel byte index).
+  - Workgroup barrier sits AFTER compute_tile (WAR protection on the ping/pong
+    A LDS slab — the next K-tile's DMA must not overwrite a slab still being
+    read by a slower wave).
+
+Supports: bf16 output (default) and in-kernel D-128 group fp8 quant output
+(quant_output=True), each with optional split-K accumulation (split_k>1).
 
 Public entry points (all in this module):
   compile_fp8_einsum_clean_ue8m0(*, H, D, R, tile_m, tile_n, tile_k,
                                  block_swizzle_n=0, quant_output=False,
-                                 quant_transpose_scale=False)
+                                 quant_transpose_scale=False, split_k=1)
       Underlying factory. Returns a launcher for either bf16 output (default)
-      or in-kernel D-128 group fp8 quant output (quant_output=True).
+      or in-kernel D-128 group fp8 quant output (quant_output=True). split_k>1
+      returns a host wrapper that accumulates fp32 partials then casts to bf16.
 
   compile_fp8_einsum_clean_ue8m0_qz(*, H, D, R, ..., transpose_scale=False)
-      Convenience wrapper — calls the underlying factory with
-      quant_output=True. The qz launcher signature adds `arg_sz`
-      (fp32 D-128 group scale) between `arg_z` and `arg_x`.
+      Convenience wrapper — quant_output=True. The qz launcher signature adds
+      `arg_sz` (fp32 D-128 group scale) between `arg_z` and `arg_x`.
+
+  compile_fp8_einsum_clean_ue8m0_qz_splitk(*, H, D, R, ..., split_k)
+      qz + split-K convenience wrapper.
 
   compile_fp8_einsum_clean_ue8m0_auto(*, H, D, R, B=None, ...)
   compile_fp8_einsum_clean_ue8m0_qz_auto(*, H, D, R, B=None, ...)
-      Autotune-dispatched: looks up the best (tile_m, tile_n, tile_k, bsw)
-      from the per-shape table at the top of this file and forwards to the
-      underlying factory. Use these by default unless you need to override
-      the tile.
+      Autotune-dispatched: looks up the best
+      (tile_m, tile_n, tile_k, bsw, split_k) from the per-shape table at the top
+      of this file and forwards to the underlying factory. Use these by default
+      unless you need to override the tile.
 
-The experimental ni-rotated variant lives in fp8_einsum_clean.py as
-`compile_fp8_einsum_clean_ue8m0_ni_rotated` (not used by autotune dispatch).
+  fp8_einsum(x, y, sx, sy, *, out_dtype=..., ...)
+      User-facing entry: validates, compiles+caches, allocates output, launches.
 
 ABI (DeepGEMM SM100 packed-UE8M0):
-  arg_z : bf16 (B, H, D)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by zufayu** [comment]
Tested on MI355X at the TP=8 V4-Pro wo_a shape (H=2, D=1024, R=4096, B=512,
  i.e. c=4096 / dp_attn=8). All modes pass numerically (cos_err 1e-6 ~ 3e-4).

  Perf at this point (same shape, apples-to-apples):
    - BF16, no split-K : 19.3 us / 445 TF   (best tile 64x128x256_bsw0)
    - QZ (fp8 output)  : 37.5 us             -> 1.9x slower than BF16
    - BF16 split-K     : sk2 53.8us / sk4 97us / sk8 165us -> 2.8x+ slower

  So at H=2 / M=512 both the fp8-output epilogue and split-K are net losses
  on this exact shape - no peak assumption needed, each just loses to its own
  BF16 no-split-K baseline. For reference the same kernel reaches ~1965 TF at
  H=16 / large B (your reported number), so this regime is far off that, which
  matches the "minor improvement" note. Root cause is the structural batch=2
  (n_groups at TP=8), not K-parallelism.

  Directions that would actually help this regime:
    1. Map both the group (H) and token (M) axes onto the grid (persistent /
       cooperative kernel) so batch=2 no longer caps WG count.
    2. Fuse the two groups into a single GEMM where possible, instead of a
       batched einsum, to keep the MFMA pipeline full at small M.
    3. Skip the QZ (fp8-output) epilogue at high TP / small batch and keep BF16
       output - the dequant/quant cost outweighs the bandwidth saving here.

**Comment by zufayu** [comment]
Validated at the actual TP=8 DSv4-Pro wo_a shape (H=2, D=1024, R=4096, B=512):

  64x128x256_bsw0:  19.6 µs / 439 TF  ← best
  128x128x256_bsw0: 24.5 µs
  128x128x128_bsw0: 29.3 µs
  hipBLASLt baseline: 80 µs → 4.1× speedup confirmed.

Prefill (B=4096) best: 128x128x256_bsw0 at 36 µs / 1904 TF.

Three fixes needed before this can merge:

1. Import: `from aiter.ops.flydsl.kernels.fp8_einsum_ready import` →
           `from aiter.ops.flydsl.kernels.fp8_einsum import`
   (crashes at import as-is)

2. Test coverage: H is hardcoded to 8; add H=2 (the TP=8 production shape)
   and B=512 to the bf16 non-split-K sweep.

3. Exit code: add `sys.exit(n_fail)` at the end so CI catches failures.

The kernel itself is correct and the perf is real. Just needs these fixes to be CI-ready.

**Comment by ganyi1996ppo** [comment]
Updated

**Review by zufayu — DISMISSED** [review]

**Comment by zufayu** [comment]
MI300X  CI  ReRun

**Comment by zufayu** [comment]
CI Failed -> ACTIONS: MI300X CI Re Run

**Comment by ganyi1996ppo** [comment]
> CI Failed -> ACTIONS: MI300X CI Re Run

This kernel didn't support mi300 from start

**Comment by zufayu** [comment]
**Root cause:** `compile_fp8_einsum_clean_ue8m0` already has a gfx950 guard (line 773, `raise RuntimeError`), but the test file doesn't handle this exception — so on MI300X the tests **FAIL** instead of **SKIP**, blocking the merge.

**Fix:** Add a module-level skip marker at the top of `op_tests/test_fp8_einsum.py`:

```python
import pytest
from aiter.jit.utils.chip_info import get_gfx

pytestmark = pytest.mark.skipif(
    get_gfx() != "gfx950",
    reason="fp8_einsum targets gfx950 (MI355X) only",
)
```

This matches the pattern used in other arch-specific tests in the repo (e.g. `test_deepgemm.py`). With this change, MI300X CI shards will report SKIP instead of FAILURE and the merge will unblock.

zufayu:
@ganyi1996ppo This PR has been inactive for over 2 months. Is the fp8 einsum flydsl implementation still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

zufayu:
Closing this as announced — no response within the window. For future reference: the kernel itself is correct with real perf (4.1x vs hipBLASLt at the TP=8 wo_a shape); the remaining CI-ready fixes are listed in the June 23 review above. Anyone is welcome to pick this up and resubmit against the latest `main`.

PR #2723, opened 2026-04-13
Fix Triton MoE GEMM shared memory exhaustion by reducing stage count

nidal567:
- Reduce num_stages in kernel configs
- Lower LDS usage to avoid shared memory OOR
- Fix `triton.runtime.errors.OutOfResources` errors in MoE GEMM kernels

## Motivation

Several Triton kernels were failing with:
- `triton.runtime.errors.OutOfResources: shared memory`
- See [here](https://github.com/ROCm/triton-internal/issues/1772) for my up-to-date github issue showing the results of the Triton kernel test suite

This was observed across multiple test cases:
- test_ff_a16w16_fused.py
- test_fused_gemm_afp4wfp4_a16w16.py
- `test_moe_gemm_a4w4.py #current PR`
- `test_moe_gemm_a8w4.py #current PR`
- `test_moe_gemm_a8w8.py #current PR`

The issue became prominent in testing after async copy was enabled, which increased LDS (shared memory) usage.

This PR is about the MoE GEMMs labeled with #current PR as the other two kernels are in progress.

## Technical Details

The root cause is increased LDS pressure due to:
- Double buffering from `num_stages`
- Increased tile residency in shared memory
- Kernel configurations originally tuned for register staging exceeding LDS limits

To address this:
- Reduced `num_stages` from 2 -> 1 in affected Triton kernel configs:
  - `moe_op_gemm_a4w4.py`
  - `moe_op_gemm_a8w4.py`
  - `moe_op_gemm_a8w8.py`
- This lowers shared memory (LDS) usage and avoids exceeding hardware limits

No other changes were introduced beyond parameter adjustments in the config files.

## Test Plan

- Ran the following affected test cases:
  - `test_moe_gemm_a4w4.py`
  - `test_moe_gemm_a8w4.py`
  - `test_moe_gemm_a8w8.py`
- Executed tests multiple times to ensure stability and correctness.

## Test Result

- The following previously failing tests now pass: 
  - `test_moe_gemm_a4w4.py`
  - `test_moe_gemm_a8w4.py`
  - `test_moe_gemm_a8w8.py`
- Results are consistent across multiple runs

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3372, opened 2026-05-27
LDS-aware num_stages selection for gfx950 MoE GEMM

xiaohuguo2023:
**Context:** 

PR #2723 set `num_stages = 1` unconditionally on gfx950 to fix LDS exhaustion in the Triton MoE GEMMs (a8w4 / a4w4 / a8w8).
  The unconditional revert regresses shapes whose double-buffered staging actually fits — gpt-oss-120b at hidden=2880 on MI355X loses up  to **-31% throughput** at TP=1 8K context.

  ### Regression from PR #2723 (gpt-oss-120b w-mxfp4-a-fp8 on MI355X, TP=1, ISL 8192 / OSL 1024)

  | concurrency | AITER 0.1.13 (PR #2723) | AITER 0.1.10.post3 (pre-#2723) | Δ |
  |------------:|----------------------:|-------------------------------:|------:|
  |   4 |  6,309 |  6,432 |  -1.9% |
  |   8 |  9,194 |  9,846 |  -6.6% |
  |  16 | 13,586 | 15,688 | -13.4% |
  |  32 | 17,966 | 22,523 | -20.2% |
  |  64 | 22,804 | 30,394 | -25.0% |
  | **128** | **26,426** | **38,318** | **-31.0%** |

  (Units: total tokens/s across the deployment. Three runs per shape, cold cache.)

  ### Fix

  Pick `num_stages` per `(block_m, block_n, block_k)` using the actual LDS budget.


aiter/ops/triton/utils/_triton/arch_info.py
```diff
@@ -30,3 +30,23 @@ def is_mx_scale_preshuffling_avail():
 
 def is_tdm_avail():
     return get_arch() in ("gfx1250",)
+
+
+_LDS_CAP_BYTES = {"gfx950": 163840, "gfx942": 65536}
```
brunomazzottiamd, line 35: I think `_LDS_CAP_BYTES` definition is in the right place, since it's an architectural property.

However, `_gemm_lds_bytes` and `pick_gemm_num_stages` are related to GEMM kernels. I'd put these functions in a `gemm_utils.py` or `gemm_commons.py` file, I don't know if we have such utility hub in AITER.

As far as I can see, `pick_gemm_num_stages` is only used by MoE GEMMs. I'd put this function in `aiter/ops/triton/moe/moe_op_gemm_common.py` (or something like this), to reduce scope to the minimum common denominator (despite the fact that it's applicable to any GEMM, not only MoE GEMMs).

aiter/ops/triton/utils/_triton/arch_info.py
```diff
@@ -30,3 +30,23 @@ def is_mx_scale_preshuffling_avail():
...
+def _gemm_lds_bytes(block_m, block_n, block_k, bits_a, bits_b, num_stages):
+    # Non-async LDS usage (matches TensorAtlas calculate_lds_usage).
+    LDSA_bits = block_m * block_k * bits_a
+    LDSB_bits = block_n * block_k * bits_b
+    if num_stages <= 1:
+        return max(LDSA_bits, LDSB_bits) // 8
+    return (LDSA_bits + LDSB_bits) * (num_stages - 1) // 8
+
+
+def pick_gemm_num_stages(arch, block_m, block_n, block_k, bits_a, bits_b):
```
brunomazzottiamd: Can you be defensive and add some assertions on input arguments? I'm assuming "all input is evil" by default. I don't know, I think all arguments must be positive to this calculation make sense. Maybe in real world everything is also a power of 2, but asserting on this may be too much!

brunomazzottiamd (commented):
Xiaohu, I've left some non-blocking comments. Feel free to ask for another review round as soon as you resolve the padded size estimation issue.

brunomazzottiamd:
Added `ci:triton-300x` label to the PR, to run unit tests on `gfx942` too. Just to be sure the MoE kernels are running fine on `gfx942`.

nidal567:
Thanks for cleaning this up. I reviewed the changes and the improvement looks good to me.

This preserves the intent of the original fix, which is avoiding LDS/shared-memory exhaustion on the affected MoE GEMMs while avoiding the broader performance cost of forcing `num_stages=1` for every gfx950 shape. At the time of my PR we were focused on having them pass the triton test suite and Triton CI with the ongoing `async_copy` issues. Selecting the stage count based on the estimated LDS usage is a cleaner and more optimal approach towards the performance of these kernels. Thank you Xiaohu!

I don't have additional concerns beyond the existing review comments. Once those are addressed and CI is clean on the relevant gfx950/gfx942 coverage, I can check again. Otherwise, this looks good so far from my side.

aiter/ops/triton/utils/_triton/arch_info.py
xiaohuguo2023: thanks, moved to aiter/ops/triton/utils/gemm_config_utils.py

aiter/ops/triton/utils/_triton/arch_info.py
xiaohuguo2023: added assert. Thanks!

xiaohuguo2023:
> Thanks for cleaning this up. I reviewed the changes and the improvement looks good to me.
> 
> This preserves the intent of the original fix, which is avoiding LDS/shared-memory exhaustion on the affected MoE GEMMs while avoiding the broader performance cost of forcing `num_stages=1` for every gfx950 shape. At the time of my PR we were focused on having them pass the triton test suite and Triton CI with the ongoing `async_copy` issues. Selecting the stage count based on the estimated LDS usage is a cleaner and more optimal approach towards the performance of these kernels. Thank you Xiaohu!
> 
> I don't have additional concerns beyond the existing review comments. Once those are addressed and CI is clean on the relevant gfx950/gfx942 coverage, I can check again. Otherwise, this looks good so far from my side.

this sounds like we should do the retune for triton moe kernel ? the existed tile sizes are probably too big for async_copy ?

lburzawa:
To give some context we did a lot of tuning of these moe kernels in the fall of 2025 for GPT and Deepseek where the non-async load option performed better and we always used num_stages=2 for better performance. Then the compiler change made async loads the default option and as a result we made num_stages=1 the temporary fix so we dont get out of memory errors. I believe num_stages=2 would also perform better in the async load case, we just need to tune the block size and num_warps properly. Could we do that instead?

aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -235,3 +235,58 @@ def compute_splitk_params(config: dict, K: int) -> dict:
...
+def pick_gemm_num_stages(
+    arch, block_m, block_n, block_k, bits_a, bits_b, use_async_padding=False
+):
+    assert min(block_m, block_n, block_k, bits_a, bits_b) > 0
+    # bits_a / bits_b: element bit-widths (8 for fp8, 4 for mxfp4).
+    # use_async_padding: True when the kernel lowers to async direct-to-LDS
+    # with padded shared encoding (e.g. a4w4 on gfx950).
+    cap = arch_info._LDS_CAP_BYTES.get(arch)
+    if cap is None:
+        return 2
```
brunomazzottiamd, line 288: Wouldn't it be more prudent to use `num_stages=1` for architectures with unknown LDS capacity? It's not a blocker in my opinion, just asking your opinion...

brunomazzottiamd (approved):
LGTM! Let's wait for the opinion of other reviewers before merging.

brunomazzottiamd:
> To give some context we did a lot of tuning of these moe kernels in the fall of 2025 for GPT and Deepseek where the non-async load option performed better and we always used num_stages=2 for better performance. Then the compiler change made async loads the default option and as a result we made num_stages=1 the temporary fix so we dont get out of memory errors. I believe num_stages=2 would also perform better in the async load case, we just need to tune the block size and num_warps properly. Could we do that instead?

Lukasz, @nidal567 is retuning all `gfx950` kernels to take the "async load by default" change into consideration. He did it for batched GEMMs, currently working on fused GEMMs. MoE GEMMs can be the next target.

Do you think it's bad if we merge this PR as a temporary improvement? I mean, while the MoE GEMM retune isn't properly done... Other option would be to close this PR and do the retuning as you suggested.

Thanks for sharing your opinion on this PR.

xiaohuguo2023:
> > To give some context we did a lot of tuning of these moe kernels in the fall of 2025 for GPT and Deepseek where the non-async load option performed better and we always used num_stages=2 for better performance. Then the compiler change made async loads the default option and as a result we made num_stages=1 the temporary fix so we dont get out of memory errors. I believe num_stages=2 would also perform better in the async load case, we just need to tune the block size and num_warps properly. Could we do that instead?
> 
> Lukasz, @nidal567 is retuning all `gfx950` kernels to take the "async load by default" change into consideration. He did it for batched GEMMs, currently working on fused GEMMs. MoE GEMMs can be the next target.
> 
> Do you think it's bad if we merge this PR as a temporary improvement? I mean, while the MoE GEMM retune isn't properly done... Other option would be to close this PR and do the retuning as you suggested.
> 
> Thanks for sharing your opinion on this PR.

I think you need this PR to guard your LDS limit, otherwise, how would you do the tuning ? 

cc @brunomazzottiamd @lburzawa @nidal567

xiaohuguo2023:
And in fact, my next step is going to do some quick tuning particularly for gpt-oss 120b, as I need this asap.

lburzawa:
If the next step is to retune the blocks then sounds good to me.

lburzawa (approved):

brunomazzottiamd:
@xiaohuguo2023, feel free to merge. I think we have reached an agreement.

PR #3377, opened 2026-05-27
flydsl moe: EP reduce path with masked gather + tuned-key fix

yadaish:
## Summary
  - Add an EP-aware stage-2 reduce path in FlyDSL MOE: `compile_moe_reduction`
    now fuses the validity gather `valid = expert_mask[topk_ids[t,k]] != 0`,
    so masked slots cost no atomics and no torch.sum epilogue. Gated by
    `AITER_FLYDSL_FORCE_REDUCE=1` for debug; threaded through
    `_flydsl_stage2_wrapper` and `flydsl_moe_stage2`.
  - `moe_sorting` learns `need_reduce` / `has_reduce_mask` so `moe_buf` is
    sized correctly for the three downstream shapes (atomic accumulate vs
    per-topk slab vs caller-owned).
  - Fix tuned-config lookup miss in EP mode: callers append one fake-mask
    slot to `topk_ids`, inflating runtime `topk` by 1, but tuned CSVs are
    keyed on routed topk. `get_2stage_cfgs` now takes `is_ep` and does
    `topk -= int(is_ep)` before building the key, so the existing DSv3
    tuned rows (`topk=8`) are hit when the runtime sees `topk=9`.
  - Encode `use_mask` + dtype + topk + model_dim into the reduction kernel
    symbol via `@flyc.kernel(name=module_name)` so rocprofv3 traces can
    distinguish `moe_reduction_kernel_masked_…` from `_plain_…`.
  - Add `g1u1_a4w4_mxfp4` / `g1u1_a8w4_mxfp4` EP cases to `test_moe_ep.py`
    and a `test_ep.sh` driver that covers smoke shapes plus the DSv3
    shape (`model_dim=7168, inter_dim=2048, E=256, topk=8, ep=8`).


aiter/ops/flydsl/moe_kernels.py
```diff
@@ -1173,6 +1186,47 @@ def flydsl_moe_stage2(
...
+
+            reduce_exe = compile_moe_reduction(
+                topk=topk,
+                model_dim=model_dim,
+                dtype_str=_reduce_dtype_str,
+                use_mask=use_mask,
+            )
+            X = target.view(token_num, topk, model_dim)
+            if use_mask:
+                em = expert_mask.to(torch.int32).contiguous()
```
lalala-sh, line 1232: dose here import extra copy? seems the agent write redundant input check

aiter/fused_moe.py
```diff
@@ -149,6 +164,20 @@ def moe_sorting(
         raise e
 
 
+def get_topk_valid_mask(
```
lalala-sh, line 201: Not called func?

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -3967,15 +4002,18 @@ def __call__(
...
                 )
-            valid_mask = torch.empty(
-                (0, self._topk), device=arg_out.device, dtype=torch.uint8
-            )
-        self._reduce_exe(X, Y, valid_mask, tokens_in, stream)
+            em = expert_mask.to(torch.int32).contiguous()
+            tk = topk_ids.to(torch.int32).contiguous()
+        else:
+            # Placeholders; kernel ignores them when use_mask=False (compile-time).
+            em = torch.empty(0, device=arg_out.device, dtype=torch.int32)
```
lalala-sh, line 4003: Can we remove the placeholder

aiter/fused_moe.py
yadaish: it can be kept because it shows how to build valid mask.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
yadaish: if so,  one more interface will added.  This part of code is rarely used.

aiter/ops/flydsl/moe_kernels.py
yadaish: No copy here. Just one defense guard

valarLip (dismissed):

aiter/ops/flydsl/moe_kernels.py
valarLip: we may use stride tensor here...

valarLip (approved):

PR #3382, opened 2026-05-28
[OPUS]fix opus_gemm mono version m align assert error

demonsan:
Release the restriction of opus_gemm mono kernel. Now only N and K should be aligned due to performance issue.


demonsan (commented):
n,k will use other instances cover

valarLip (approved):

PR #3395, opened 2026-05-28
Fix reduce scatter when dim != 0

yitingw1:
## Motivation

`reduce_scatter` produces **incorrect results when `dim != 0`**, blocking DCP decode path which calls `reduce_scatter(out, dim=1)`. Root cause: `movedim` direction reversed, return values of `reshape`/`movedim` discarded (no-op), and `use_custom=False` path ignores `dim` entirely.

## Technical Details

**`communicator_cuda.py`** :
- Fix `movedim` direction: `input_.movedim(dim, 0)` (was `movedim(0, dim)`)
- Add `tmp` contiguous buffer for NCCL output when `dim != 0`, then `output_.copy_(tmp.movedim(0, dim))` to write back
- Restructure to dim-first branching, eliminating duplicated code between `ca_comm` and `pynccl` paths
- Add `dim == 0` divisibility assertion

**`parallel_state.py`**:
- `use_custom=True`: fix output shape to use `dim` parameter instead of always dividing `shape[0]`
- `use_custom=False`: add `movedim(dim, 0)` before `torch.distributed.reduce_scatter_tensor` + `movedim(0, dim)` after, since `reduce_scatter_tensor` always scatters along dim 0

**`test_reduce_scatter.py`**:
- Add `dim` parameter support, test dims `[0, 1, -1]`
- Add 3-way accuracy comparison: dist vs CPU ref, aiter vs CPU ref, dist vs aiter
- Add 3D shape `(4, 8, 8192)`, use `rtol=0.05, atol=0.2` for bf16 tolerance
- Return `output.cpu()` from worker to avoid HIP IPC errors with large tensors

## Submission Checklist


TennyWang1223:
aiter/dist/device_communicators/base_device_communicator.py:183 should also change movedim(0, dim) to movedim(dim, 0), same as CudaCommunicator

yitingw1:
> aiter/dist/device_communicators/base_device_communicator.py:183 should also change movedim(0, dim) to movedim(dim, 0), same as CudaCommunicator

Done.

TennyWang1223 (approved):
LGTM

yitingw1:
Closed as [PR-3464](https://github.com/ROCm/aiter/pull/3464) has fixed reduce-scatter dim!=0 issue in kernel side.

PR #3403, opened 2026-05-28
[OPUS]Opus gemm 4G reject

demonsan:
when tensor size over 4GB, it is invalid for single buffer inst. reject these shape for opus gemm

### DIFF
--- a/csrc/opus_gemm/include/gfx950/opus_gemm_arch_gfx950.cuh
+++ b/csrc/opus_gemm/include/gfx950/opus_gemm_arch_gfx950.cuh
@@ -186,6 +186,24 @@ opus_dispatch_a16w16_gfx950<bf16_t>(int M, int N, int K, int batch, bool has_bia
         return it->func;
     }
     (void)batch;  // heuristic does not currently use batch.
+    // 4 GiB buffer-resource guard. The heuristic returns one of
+    // HEURISTIC_DEFAULT_KIDS, all of which are legacy (non-4g_safe) and
+    // build a single AMDGPU buffer-resource over the whole A/B/C tensors;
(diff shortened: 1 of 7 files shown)


csrc/opus_gemm/include/gfx950/opus_gemm_arch_gfx950.cuh
```diff
@@ -210,6 +228,19 @@ opus_dispatch_a16w16_gfx950<fp32_t>(int M, int N, int K, int batch, bool has_bia
         return it->func;
     }
     (void)batch;
+    // 4 GiB buffer-resource guard (see <bf16_t> overload for rationale).
+    // C is fp32 here so the bound is 4 bytes/element.
+    constexpr uint64_t U32_MAX_BYTES = (1ULL << 32) - 1;
+    const uint64_t a_bytes = (uint64_t)M * (uint64_t)K * sizeof(bf16_t);
+    const uint64_t b_bytes = (uint64_t)N * (uint64_t)K * sizeof(bf16_t);
+    const uint64_t c_bytes = (uint64_t)M * (uint64_t)N * sizeof(fp32_t);
+    AITER_CHECK(a_bytes <= U32_MAX_BYTES && b_bytes <= U32_MAX_BYTES
```
valarLip, line 237: still need this check?

demonsan: yes, this check is for heuristic search. Currently, kids support over 4GB is not allowed in heuristic mode and can only be invoked in tuner and tuned path due to accuracy concern.

valarLip (approved):

PR #3408, opened 2026-05-28
[FlyDSL MoE] Add no combine feature

zx3xyy:
## Summary

This adds a small FlyDSL MoE stage2 escape hatch for callers that need the raw
per-route output instead of the usual topk-reduced output.

With `return_per_slot=True`, `flydsl_moe_stage2` writes stage2 results into a
`(token_num, topk, model_dim)` buffer and returns that buffer directly. The
caller can then apply `topk_weights` and reduce the topk dimension later, for
example after expert-parallel communication.

## Test plan
python -m pytest --collect-only -q aiter/ops/flydsl/test_flydsl_moe_a4w4.py


zx3xyy:
@HaiShaw

HaiShaw (dismissed):

HaiShaw:
@coderfeli would please help review and check the aiter tests as well?
cc @valarLip

HaiShaw (approved):
LGTM

HaiShaw:
@coderfeli @valarLip Please review again, code change is simplified by direct modifying `flydsl_moe_stage2`.

valarLip (approved):

PR #2642, opened 2026-04-07
fix: enable MXFP4 MoE at TP=4/8 via CKTile a4w4 kernels and quant fixes

thpereir:
Enable tensor-parallel MXFP4 MoE for non-power-of-2 scale dimensions (TP=4: scaleN=12, TP=8: scaleN=6) by fixing quant padding, sort kernel reshaping, and adding CKTile a4w4 kernel dispatch.

Fix bugs and add two performance optimizations to the MXFP4 a4w4 (fp4x2 activations + fp4x2 weights) two-stage MoE path, validated on MiniMax-M2.1-MXFP4 at TP=8 on gfx950.

Fixes:

1. Wrong stage1 kernel for a4w4:
   cktile_moe_stage1 routes fp4x2 activations through the fp8 pipeline
   (AQUANT_Pipeline), which misinterprets packed fp4 as fp8 -> garbage.
   Switch stage1 to ck_moe_stage1 (JIT CK DeviceMoeGemmMXBPreShuffle).

2. ksplit>1 steals a4w4 cases for shared experts:
   MiniMax shared experts have inter_dim=256 (per TP=8 rank).
   get_ksplit() returns 2, and the "ksplit>1 and is_shuffled" elif
   fires before the a4w4 branch because it only checks q_dtype_w.
   Fix: add "q_dtype_a not in [dtypes.fp4x2]" guard.

3. ksplit must be 0 for a4w4 in fused_moe_2stages:
   When metadata.ksplit>1, fused_moe_2stages skips fp4x2 activation
   quantization. Force ksplit=0 in the a4w4 MOEMetadata.

Performance optimizations:

4. Skip intermediate bf16->fp4x2 quantization between stages:
   Stage2 now uses CKTile AOT with bf16 activations (a16w4 path)
   instead of JIT CK with fp4x2 activations, eliminating a full
   quantization kernel launch + memory round-trip between stages.
   New skip_inter_quant flag on MOEMetadata controls this.

5. block_m 32->64 for fp4x2:
   Sweep across all MiniMax batch sizes shows block_m=64 is
   consistently faster than 32 on gfx950 (up to 1.54x for prefill).
   128 is not supported by current CKTile stage2 instances.

Additional changes:
- CKTile AOT: add pk_fp4 x pk_fp4 codegen for both gemm1 and gemm2,
  generate stage1+stage2 instances for a4w4, skip split_k for pk_fp4.
- cktile_moe_stage1 split_k>1: fix tmp_out buffer size (use sorted_size
[...]

PR #3412, opened 2026-05-29
[Bugfix] Enable MXFP4 MoE at TP=4/8 via CKTile a4w4 kernels and quant fixes

haoyangli0109:
Based on [PR](https://github.com/ROCm/aiter/pull/2642)

```
HIP_VISIBLE_DEVICES=0,5,6,7 \
AITER_LOG_LEVEL=WARNING \
python -m atom.entrypoints.openai_server \
  --model /shareddata/amd/MiniMax-M2.1-MXFP4 \
  --trust-remote-code \
  -tp 4 \
  --port 5679 \
  --server-port 7778

lm_eval \
  --model local-completions \
  --model_args "model=/shareddata/amd/MiniMax-M2.1-MXFP4,base_url=http://localhost:7778/v1/completions,tokenized_requests=False,tokenizer_backend=None,num_concurrent=32" \
  --tasks gsm8k \
  --num_fewshot 5 \
  --batch_size auto

```

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9287|±  |0.0071|
|     |       |strict-match    |     5|exact_match|↑  |0.9272|±  |0.0072|

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -767,6 +767,7 @@ class MOEMetadata:
     fuse_quant: str = ""
     stage2_has_bias: bool = False
     flat: bool = False
+    skip_inter_quant: bool = False
 
 
 def _needs_swiglu_bias_support(dtype, quant_type):
@@ -1206,6 +1207,12 @@ def get_2stage_cfgs(
     def get_block_m() -> int:
         if q_dtype_a == dtypes.fp8:
             return 32
+        elif q_dtype_a == dtypes.fp4x2:
+            # MXFP4 fused quant+sort requires block_size % 32 == 0.
+            # block_m=64 is significantly faster than 32 for fp4x2 on
+            # gfx950 across all tested batch sizes (up to 1.5x for
+            # prefill).  128 is not supported by current CKTile stage2.
+            return 64
         else:
             return 16 if token < 2048 else 32 if token < 16384 else 64
 
@@ -1451,6 +1458,7 @@ def get_2stage_cfgs(
         dtype in [dtypes.bf16, dtypes.fp16]
         and q_type == QuantType.per_1x32
         and q_dtype_w in [dtypes.fp4x2]
+        and q_dtype_a not in [dtypes.fp4x2]
         and is_shuffled
         and not (activation == ActivationType.Swiglu and q_dtype_a == dtypes.fp4x2)
         and (ksplit > 1 or swiglu_mxfp4_bf16_cktile)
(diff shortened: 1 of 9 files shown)


haoyangli0109:
hi,
@k50112113 
Thiago is on vacation, so I made some changes based on his branch. Could you help review this PR?
cc 
@thpereir @thiagocrepaldi

aiter/fused_moe.py
```diff
@@ -1189,6 +1190,12 @@ def _lookup_cfg(c2s):
     def get_block_m() -> int:
         if q_dtype_a == dtypes.fp8:
             return 32
+        elif q_dtype_a == dtypes.fp4x2:
```
lalala-sh, line 1210: please rm it first. If you want to use the config, choose it by moe tunner

aiter/fused_moe.py
```diff
@@ -1472,6 +1479,7 @@ def get_block_m() -> int:
         dtype in [dtypes.bf16, dtypes.fp16]
         and q_type == QuantType.per_1x32
         and q_dtype_w in [dtypes.fp4x2]
+        and q_dtype_a not in [dtypes.fp4x2]
```
lalala-sh, line 1461: also rm here, keep the dispatch policy.

aiter/fused_moe.py
lihaoyang-amd: If we delete it, an error will occur when block_M=16.
```
RuntimeError: Unsupported block_m value for moe heuristic dispatch: 16
```

aiter/fused_moe.py
lihaoyang-amd: If we remove this condition, w4a4 will incorrectly enter this section, which will still cause precision issues.

aiter/fused_moe.py
lalala-sh: Change the default dispatch policy is risky. You can try the moe tuner to tune the Minimax TP4 configs.

lihaoyang-amd: The heuristic get_block_m is the fallback when no tuned config matches, so it must produce a valid size for a4w4. Today it returns 16 for token < 2048, but a4w4 stage1 has no block_m=16 CK instance (and the MXFP4 fused quant+sort requires block_m % 32 == 0), so it hard-fails with Unsupported block_m value for moe heuristic dispatch: 16 for any untuned a4w4 decode shape — not just Minimax TP4.

So this elif is a correctness fix, not a tuning knob:

16 → no kernel instance + violates the %32 sort constraint → crash
128 → unsupported by the current CKTile stage2
→ only {32, 64} are valid, and 64 is ≥ 32 across all tested batch sizes (up to ~1.5x on prefill)
64 is therefore the only well-defined valid choice, not a hand-picked perf value, and tuned configs still override it.

This is not Minimax-specific. Other models hit the same accuracy problem on current main when running MoE w4a4 — e.g. I reproduced it on Qwen3.5-35B-A3B-MXFP4 (TP=2): prefill is fine (ksplit=0), but at decode the heuristic produces ksplit=2, which steals a4w4 into the CKTile a16w4 stage1 path and silently emits garbage tokens. Tuning only the TP4 config patches those specific shapes; the fallback stays broken for every other untuned a4w4 case. The proper fix is to make the default heuristic valid for a4w4.

lihaoyang-amd: Actually, we're using a specific value for fp8.

lalala-sh: We'll refine the moe dispatch code with flydsl in future. For now, I draft another fix PR in flydsl and moe tunner. https://github.com/ROCm/aiter/pull/3476

haoyangli0109: cc: @thpereir

PR #3414, opened 2026-05-29
fmoe: Expected m host hint

inkcherry:
## Submission Checklist


valarLip:
this way can't work for hipgraph

aiter/fused_moe.py
```diff
@@ -372,8 +342,15 @@ def fused_moe_(
         else:
             q_dtype_a = dtypes.fp4x2
 
+    # In DeepEP/Mori low-latency dispatch, topk_ids.shape[0] is the padded
```
yadaish: after protocol confirmed, please update the definition of M_for_schedule in comment.

inkcherry: thanks, update.

op_tests/test_fused_moe_expected_m.py
yadaish, line 1: remove this file

yadaish:
LGTM

inkcherry:
Thanks for the callout, @valarLip — that's a good point. Keeping expected_m unchanged within a graph is indeed necessary.
For a given graph, expected_m stays constant in practice, and we've validated it end-to-end. We also ran performance tests to confirm it holds under the actual workload.

op_tests/test_fused_moe_expected_m.py
inkcherry: thanks @yadaish  Could we keep this for now to prevent regressions? Of course, if you're confident that the changes here won't be broken, I'm happy to remove them — it's up to you. I'm not fully familiar with the specifics of this change

valarLip:
i can't assess this way...

inkcherry:
Thanks @valarLip, that's a fair thing to double-check, and I appreciate you raising it. Let me share some context to address the hipgraph concern precisely: this expected_m host hint is not something new introduced here — it's a general, already-established approach, and by construction it stays constant within a captured graph.

1. It is graph-invariant in the existing SGLang low-latency paths. Both LL dispatch backends already compute expected_m from capture-time-constant quantities only:

DeepEP: https://github.com/sgl-project/sglang/blob/0f5e14e1d9ff07b9627204450ce9576f42f6cc58/python/sglang/srt/layers/moe/token_dispatcher/deepep.py#L650-L654
Mooncake: https://github.com/sgl-project/sglang/blob/0f5e14e1d9ff07b9627204450ce9576f42f6cc58/python/sglang/srt/layers/moe/token_dispatcher/mooncake.py#L156-L160
expected_m = (hidden_states.shape[0] * buffer.group_size * topk_ids.shape[1] + num_experts) // num_experts
Every term — padded batch size, EP group size, top-k, expert count — is fixed at capture time; none of it depends on the runtime routing distribution. So expected_m stays constant for a given graph, and there is no graph re-capture or per-step host divergence.

2. Consistent with the established expected_m usage. This matches how expected_m is treated as a host-side hint elsewhere — e.g. the DeepGEMM masked grouped-GEMM path designed for "the inference decoding phase, when CUDA graph is enabled" (https://github.com/deepseek-ai/DeepGEMM/blob/88965b078186ee7510ab9fc4f1d5ebc19adfa8d1/README.md), with the parameter documented as "a value hint (which is a value on CPU) for the M expectation of each batch, correctly setting this value may lead to better performance." (https://docs.flashinfer.ai/generated/flashinfer.gemm.batch_deepgemm_fp8_nt_groupwise.html).

3. Validated end-to-end. Correctness and performance were confirmed under the real workload with the graph captured.

I've also already documented this invariant inline at the expected_m definition — "it MUST be identical for every call sharing the same HIP graph (one graph key → one fixed expected_m); None → bit-identical to old behavior." If there's a concrete case where it would change inside a captured graph, please point to it and I'll address it directly.
 FYI @Duyi-Wang

PR #3429, opened 2026-05-29
Fuse dynamic_per_tensor_quant_fp8_i8 into one launch for the decode regime

JohnQinAMD:
## Summary

`dynamic_per_tensor_quant_fp8_i8` launches **two** kernels: an atomic-max amax
pass (`_dynamic_per_tensor_quant_fp8_i8_kernel`) followed by a quantize pass
(`_static_per_tensor_quant_fp8_i8_kernel`). At decode-size tensors this is
**launch / CUDA-graph-node bound**, not bandwidth bound.

This PR adds `_fused_dynamic_per_tensor_quant_fp8_i8_kernel`: a single workgroup
computes the global amax and quantizes in **one launch** via a grid-stride loop
(no cross-block reduction is needed once one program walks the whole tensor).
`dynamic_per_tensor_quant_fp8_i8` routes to it for contiguous tensors with
`numel() <= 40960` (the decode regime); larger tensors keep the existing
two-launch path unchanged.

## Correctness

The fused kernel is **byte-identical** to the existing two-launch path:

- **scale:** `tl.max(|x|) / DTYPE_MAX`. `max` commutes with `/ DTYPE_MAX`, so a
  single global max divided once equals the existing per-row `max(row_amax/MAX)`
  via `atomic_max` — same fp32 bits.
- **quant:** `(x * (1 / scale)).to(out.dtype)` — identical to
  `_static_per_tensor_quant_fp8_i8_kernel`.
- **no clamp:** dynamic scaling guarantees `|x| / scale <= DTYPE_MAX`, matching
  the existing path (which also does not clamp).

Works for both fp8 and int8 (`DTYPE_MAX` derived from the output dtype).

## Motivation

Per-tensor dynamic fp8 activation quant fires once per linear input on the
decode hot path; collapsing 2 launches → 1 removes a CUDA-graph node per call.
Downstream (vLLM, Mistral-Small-3.1-24B FP8 on MI355X) an equivalent fusion
measured ~8% geomean interactivity uplift across concurrency, gsm8k unchanged.

## Testing

- [x] Imports / parse cleanly; dispatch threshold + block sizing reviewed.
- [x] Byte-identicality argued by construction (above) vs the two-launch path.
- [ ] **Recommended before merge:** GPU byte-equality test of the fused vs the
      two-launch path across decode shapes (M∈{1,8}, K∈{5120,4096,32768}; fp8 and
      int8), plus a microbench confirming the launch-count reduction.

## Follow-up

A separate change routes `per_tensor_quant_hip`'s dynamic-fp8 path to this fused
kernel so HIP-path consumers pick up the decode win without a caller change; kept
out of this PR to keep it byte-identical and self-contained.

### DIFF
--- a/aiter/ops/triton/_triton_kernels/quant/quant.py
+++ b/aiter/ops/triton/_triton_kernels/quant/quant.py
@@ -52,6 +52,47 @@ def _dynamic_per_tensor_quant_fp8_i8_kernel(
 
 
 @triton.jit
+def _fused_dynamic_per_tensor_quant_fp8_i8_kernel(
+    qx_ptr,
+    x_in_ptr,
+    scale_out_ptr,
+    N: int,
+    DTYPE_MAX: tl.constexpr,
+    BLOCK: tl.constexpr,
+):
+    # Single-workgroup fused amax + quantize for the small/decode regime.
+    # One program walks the whole (flattened, contiguous) tensor in a grid-stride
+    # loop, so the global scale is known without a cross-block reduction -> a
+    # single kernel launch instead of the atomic-max amax pass + separate quant
+    # pass. Byte-identical to _dynamic_per_tensor_quant_fp8_i8_kernel +
+    # _static_per_tensor_quant_fp8_i8_kernel: same scale (max commutes with
+    # / DTYPE_MAX), same 1 / scale, no clamp (dynamic scaling guarantees
+    # |x| / scale <= DTYPE_MAX).
+    offs = tl.arange(0, BLOCK)
+
+    amax = tl.zeros([BLOCK], dtype=tl.float32)
+    i = 0
+    while i < N:
+        idx = i + offs
+        mask = idx < N
+        x = tl.load(x_in_ptr + idx, mask=mask, other=0.0, cache_modifier=".cg")
+        amax = tl.maximum(amax, tl.abs(x.to(tl.float32)))
+        i += BLOCK
+    scale = tl.max(amax, axis=0) / DTYPE_MAX
+    tl.store(scale_out_ptr, scale)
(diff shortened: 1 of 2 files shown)


JohnQinAMD:
### ✅ GPU byte-equality validation — PASS (real MI355X)

Ran the fused 1-launch kernel against the existing 2-launch path on **chi2762
(AMD Instinct MI355X / gfx950)**, image `lmsysorg/sglang:v0.5.12-rocm720-mi35x`.
The test loads the actual kernels from this branch and compares output bytes +
scale directly.

**Result: 26/26 cases byte-identical** — `byte_mismatch=0`, `scale_eq=True`,
`max_ulp=0` for every case.

- dtypes: `float8_e4m3fn` and `int8`
- decode shapes: (1,5120), (1,4096), (1,32768), (8,5120)=cap, (8,4096),
  (2,4096), (4,5120), (16,2560)
- edge cases: **all-zero**, large-magnitude (5e4), asymmetric negative outlier,
  (1,1), and the exact 40960-element cap

This confirms the fused path is bit-for-bit equal to the two-launch path (same
scale, same `1/scale`, no clamp needed) — including the all-zero case. Remaining
checklist item (launch-count / perf microbench) is covered by the downstream E2E
numbers; happy to add a standalone microbench if useful.

zufayu:
@JohnQinAMD This PR has been inactive for over 2 months. Is the dynamic_per_tensor_quant_fp8_i8 single-launch fusion still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3451, opened 2026-06-01
Fix Q UE8M0 quant and require fp32 LN params in fused DSv3.2 indexer kernel

frida-andersson:
## Summary

The fused `indexer_qk_rope_quant_and_cache_kernel` (added in #3185) diverged from the unfused vLLM path it replaces, dropping DeepSeek-V3.2 GSM8K strict-match (lm-eval, 20-shot) from **0.95 → 0.49** when fusion was enabled. Two independent kernel bugs in `csrc/kernels/cache_kernels.cu`:

**Bug A — Q quant.** The unfused reference `per_token_group_quant_fp8` multiplies by `1 / fp8_max` (avoids GPU fast-division noise) and applies UE8M0 rounding when `scale_fmt = "ue8m0"`, which DSv3.2's indexer always passes. The fused kernel did neither, while the K-side of the same kernel applied UE8M0 correctly. Net: ~94% Q disagreement per call. Fix mirrors the K-side.

**Bug B — `norm_weight` / `norm_bias` typed `scalar_t*` (= bf16 at runtime).** vLLM stores `LayerNorm.weight` / `bias` as fp32 and `F.layer_norm` consumes them as fp32; the call site was casting fp32 → bf16 before launch, drifting K on ~3% of values per call. Fix hard-types both as `float*`, drops the now-redundant `static_cast<float>` in the kernel body, and asserts fp32 on the host. Also adds explicit `dim() == 1` and contiguity guards so non-1D or strided views are caught at launch rather than silently producing wrong results.

## Breaking change

`norm_weight` and `norm_bias` now **must be fp32**. The dtype contract tightened from "must match `q.dtype`" to "must be fp32"; any caller passing bf16 LN params will now get a loud `AITER_CHECK` failure at launch instead of silent ~3% K drift. Caller audit: vLLM DSv3.2 indexer and ROCm/ATOM `attention_mla_sparse.py` both store LN params as fp32 natively and are unaffected. SGLang does not call this kernel.

## End-to-end result

DSv3.2, MI355X, TP=4, lm-eval GSM8K (`num_fewshot=20`, `num_concurrent=256`, full 1319-question suite):

| Configuration | strict-match |
|---|---|
| Fusion off, fp8 KV cache (baseline) | 0.95 |
| Fusion on, no fixes, fp8 | 0.49 |
| Fusion on + this PR + companion vLLM `preshuffle=True`, fp8 | **0.95** |
| Fusion on + this PR + companion vLLM `preshuffle=True`, bf16 | **0.9591** |

The fix is independent of `--kv-cache-dtype` (the indexer's internal cache is always fp8). Drift harness: fused vs unfused Q fp8 + Q scale and K (with `preshuffle=True` on both) bit-exact post-fix.

## Companion vLLM PR

[vllm-project/vllm#43907](https://github.com/vllm-project/vllm/pull/43907) — ships `preshuffle=True` and switches the call site to pass fp32 LN params.


csrc/kernels/cache_kernels.cu
```diff
@@ -1256,8 +1256,9 @@ __global__ void indexer_qk_rope_quant_and_cache_kernel(
     const scalar_t* __restrict__ k,           // [num_tokens, head_dim]
     cache_t* __restrict__ kv_cache,           // [num_blocks, block_size, cache_stride]
     const int64_t* __restrict__ slot_mapping, // [num_tokens]
-    const scalar_t* __restrict__ norm_weight, // [head_dim]
-    const scalar_t* __restrict__ norm_bias,   // [head_dim]
+    // fp32 to match caller's LN params (bf16 cast drops precision).
+    const float* __restrict__ norm_weight,    // [head_dim]
+    const float* __restrict__ norm_bias,      // [head_dim]
```
samremes, line 1376: Are we sure there's no other users of this kernel that still want to pass these parameters as scalar_t/bf16?

csrc/kernels/cache_kernels.cu
```diff
@@ -1341,9 +1342,13 @@ __global__ void indexer_qk_rope_quant_and_cache_kernel(
...
     const float q_fp8_max = static_cast<float>(opus::finfo<cache_t>::max());
-    // Q scale is consumed by weights_out and must match the unfused quant path;
-    // only the K cache scale is encoded as UE8M0 for cache layout compatibility.
-    const float q_scale = fmaxf(q_amax, 1e-10f) / q_fp8_max;
+    // Match unfused per_token_group_quant_fp8: reciprocal-multiply + UE8M0.
+    const float q_inv_fp8_max = 1.0f / q_fp8_max;
+    float q_scale             = fmaxf(q_amax, 1e-10f) * q_inv_fp8_max;
+    if(use_ue8m0)
+    {
+        q_scale = exp2f(ceilf(log2f(q_scale)));
```
samremes, line 1465: This was also flagged in the original PR: https://github.com/ROCm/aiter/pull/3185#discussion_r3241426615 but ignored there, so could be it just slipped through the cracks.

frida-andersson: Yes, the unfused vLLM `per_token_group_quant_fp8` always passes `scale_fmt = "ue8m0"` for DSv3.2's indexer, so the asymmetry was unintentional - and is what produced the ~94% Q disagreement vs the unfused reference per call.

csrc/kernels/cache_kernels.cu
frida-andersson: Audited every caller of this kernel on GitHub. Two production paths:

1. `vllm-project/vllm` (DSv3.2 indexer) — `torch.nn.LayerNorm` ⇒ fp32. The `.to(q.dtype)` cast at the call site (= bf16) is the bug this PR fixes.
2. `ROCm/ATOM` `atom/plugin/attention_mla_sparse.py:631` — LN params fp32 by construction (`atom_parameter(torch.ones(dim))` / `torch.zeros(dim)` in `atom/model_ops/layernorm.py:970`). If an ATOM user does `model.to(bfloat16)` they hit the same silent ~3% K drift.

SGLang does **not** call this kernel — it only uses the K-only `indexer_k_quant_and_cache` (`python/sglang/srt/layers/attention/dsa/dsa_indexer.py:1247`), which takes no LN params. No callers inside aiter either (no `op_tests`, no benchmark).

The kernel body already does `static_cast<float>(norm_weight[dim])` and accumulates in fp32, so `scalar_t*` was only buying the implicit fp32→bf16 cast at the call site — i.e. the footgun. `float*` + the host `AITER_CHECK` turns a silent numerical regression into a loud launch failure.

If a future caller actually wants bf16 LN params we can split into two overloads or template on a separate `LN_T` at that point. Today every existing caller is fp32 upstream. However, feel free to verify this too

samremes: I agree that going back-and-forth with fp32->bf16->fp32 makes no sense and is harmful, and should be fixed at both sides. Thanks for more context!

sunway513:
Data point from a downstream verification run: I cherry-picked this PR's fix (`423f22a8`, cache_kernels.cu) onto a clean AITER branch and ran the MI35x DeepSeek-V3.2 accuracy suite (`test/registered/amd/accuracy/mi35x/test_deepseek_v32_eval_mi35x.py`, 8-GPU, linux-aiter-do-mi350x-8). It still **hangs and hits the 3600s timeout** (HIP backtrace in libamdhip64.so, exit 255) — i.e. this fix alone does not yet get V3.2 through the eval. Flagging in case there's a remaining indexer hang beyond the Q UE8M0 / fp32-LN-params changes. Happy to re-run once you push more. (FYI: I've temporarily disabled the DeepSeek-V3.2 accuracy entry in sglang_downstream.py to keep the downstream signal clean meanwhile; it flips back on when this lands.)

frida-andersson:
Thanks @sunway513 - worth flagging: SGLang's DSv3.2 indexer (`dsa_indexer.py`) only imports `indexer_k_quant_and_cache` (L68) and calls it once at L1247. No references to `indexer_qk_rope_quant_and_cache` anywhere in `sgl-project/sglang`.

This PR only touches `indexer_qk_rope_quant_and_cache_kernel` and its host wrapper, so it's a no-op for the SGLang path, re-enabling your `sglang_downstream.py` DSv3.2 entry shouldn't be gated on this PR. Whatever's hanging is somewhere else (probably `indexer_k_quant_and_cache`, the MQA logits kernels, or upstream in `dsa_indexer.py`)

ChuanLi1101:
The core fix is correct: aligning the Q scale with the unfused per_token_group_quant_fp8 (reciprocal-multiply) and making the LN params fp32 end-to-end. Two non-CI items to address before merge:

1. Add shape/contiguity guards for norm_weight / norm_bias (re: Copilot's comment). The kernel indexes them as a flat array (norm_weight[dim] / norm_bias[dim]), but the host only validates dtype and size(0) == head_dim. A non-contiguous or non-1D view would silently produce wrong results. Please add explicit dim() == 1 and contiguity checks (e.g. AITER_CHECK(norm_weight.is_contiguous(), ...)) for both tensors before launch — or, if the contract guarantees a 1D contiguous input, reply on the thread so it can be resolved.

2. Document the breaking dtype change. Host validation changed from "norm_weight/norm_bias dtype must match q.dtype" to "must be fp32". Any existing caller passing bf16 LN params will now hit the AITER_CHECK failure. The caller audit (vLLM / ATOM both fp32) was posted in the thread, which is good — please also note this as a breaking change in the PR description / changelog so downstream is aware.

frida-andersson:
Thanks for the review @ChuanLi1101! I've adressed both points:

1. Shape/contiguity guards - added `AITER_CHECK(norm_weight.dim() == 1, ...)`, `AITER_CHECK(norm_weight.is_contiguous(), ...)` and the same for norm_bias before launch in https://github.com/ROCm/aiter/pull/3451/changes/9a1bae8650aa62d02b56df019200fa2150c346a1
2. Breaking change documented - added a **Breaking change section** to the PR description noting the fp32 dtype contract and the caller audit (vLLM + ATOM both fp32 natively, SGLang doesn't call this kernel)

ChuanLi1101 (approved):
LGTM - the Q UE8M0 + fp32 LN fixes are correct, and my earlier review points are now addressed (contiguity/1D guards added, breaking dtype change documented in the ABI note + caller audit). One thing before merge (non-blocking on the code): please confirm the red CI gates (Kimi-K2.5 vllm, Kimi perf vllm, gpt-oss-120b fp8_kvcache benchmark) are pre-existing/unrelated - none of them exercise the DSv3.2 indexer path this PR touches, and DeepSeek-R1 / SGLang / Standard tests are all green.

amd-mghanimi:
> LGTM - the Q UE8M0 + fp32 LN fixes are correct, and my earlier review points are now addressed (contiguity/1D guards added, breaking dtype change documented in the ABI note + caller audit). One thing before merge (non-blocking on the code): please confirm the red CI gates (Kimi-K2.5 vllm, Kimi perf vllm, gpt-oss-120b fp8_kvcache benchmark) are pre-existing/unrelated - none of them exercise the DSv3.2 indexer path this PR touches, and DeepSeek-R1 / SGLang / Standard tests are all green.

Thanks for the review. The Kimi [errors here](https://github.com/ROCm/aiter/actions/runs/27539973061/job/81399063215?pr=3451#step:6:100) are due to this bug in upstream vllm that is hopefully going to be fixed in the next nightlies: https://github.com/vllm-project/vllm/issues/45596

The GPT-OSS error is related to refactoring triton version update. The current version in vllm is 3.6 but installing aiter automatically updated that to 3.7 (got released 2 weeks ago) this version moved `triton_kernels.matmul_ogs` to `triton_kernels.matmul` (this [PR in triton](https://github.com/triton-lang/triton/pull/8765/changes#diff-e3e2307ca17834b972e622ccf27187986752178ce15f31d2499742f89c08b80c)). Current vLLM has a non-blocking error message for this on some models but fails on some other models such as gptoss here.

valarLip:
let me know once ci passed

frida-andersson:
Thanks @valarLip! The CI failures are unrelated to this PR, see @amd-mghanimi's comment above. I rebased the PR but need approval to re-start the CI

amd-mghanimi:
New CI issues:
- Atom tests / Accuracy seems to be failing on python module/package import in the ATOM source code, which is unrelated to this PR.
`ImportError: cannot import name 'moe_shuffle_scale' from 'aiter.ops.shuffle' (/app/aiter-test/aiter/ops/shuffle.py)` 
- Kimi Downstream Test and Kimi Perf Downstream seems to be failing because the benchmark script is not waiting enough for the server to load the model and be ready for serving.  
These kimi tests are also failing on the main branch:
https://github.com/ROCm/aiter/actions/workflows/kimi-perf-downstream.yaml?query=branch%3Amain

In the past, some of these errors could get resolved by rerun. Because part of the slow loading time for these tests were due to model was getting downloaded on the runner machine, so on second attempts often they could pass but I think for Kimi tests the source script should add some waiting time to be able to pass the tests.

zufayu:
PR #3451 updates LayerNorm weight and bias to FP32, which requires matching changes in ATOM’s LayerNorm module. Merging this PR without corresponding ATOM adjustments will break existing production models.

**Background**
At present, only the QK normalization layer in the DSV32 workload uses FP32 weights, which is handled by the kernel indexer_qk_rope_quant_and_cache_kernel modified in this PR.

Required ATOM LayerNorm model-op updates
- Add a data type parameter during initialization.
- Add type conversion logic: convert FP32 weights to BF16 before feeding them into the CK LayerNorm kernel.

Merge Requirement
ATOM code changes must be submitted together with PR #3451 or merged beforehand. Conditional branch logic is needed to preserve backward compatibility and prevent runtime errors on existing models.

frida-andersson:
Thanks @zufayu for the comment, agreed this needs an ATOM-side companion. This PR only changes [the dtype contract](https://github.com/ROCm/aiter/pull/3451/files#diff-2b637aa4a1ce6281b0b1d955c37f7ecce22df2018d37e995cf0fe1b0e6dbd969R3696) on indexer_qk_rope_quant_and_cache (norm dtype == q.dtype → == fp32), it doesn't touch ATOM's LayerNorm or layernorm2d_fwd, so the CK kernel shouldn't need any change. 

Also, are you seeing an actual accuracy regression in ATOM, or is this from review? On vLLM this fix recovers GSM8K 0.49 → 0.95 (with the vLLM companion [#43907](https://github.com/vllm-project/vllm/pull/43907)) when running with `--num_fewshot 20`

zufayu:
Checked vLLM #43907 (deepseek_v2.py): it calls .to(torch.float32).contiguous() on k_norm.weight/bias explicitly, commenting "casting to bf16 here loses precision and drifts K from the unfused path." — so the companion ATOM change is required.

PS: :713
https://github.com/vllm-project/vllm/pull/43907/changes#diff-b637aa4a1ce6281b0b1d955c37f7ecce22df2018d37e995cf0fe1b0e6dbd969

frida-andersson:
Rebased cleanly on latest main (no conflicts) - needs approval from a maintainer to re-run the CI. Code is unchanged since @ChuanLi1101's LGTM: contiguity/1D guards and the breaking fp32 dtype-contract doc are in. @junhaha666 (already requested) / @valarLip - could one of you take a look and review this PR?

The ATOM dependency is covered by [ROCm/ATOM#1359](https://github.com/ROCm/ATOM/pull/1359) (optional `dtype` on `LayerNorm`, defaults `None` so existing models are unchanged; sets DSv3.2 `k_norm` to fp32) - suggest landing this with #1359.

valarLip (approved):

valarLip:
btw looks like no accuracy diff for ATOM side?

frida-andersson:
> btw looks like no accuracy diff for ATOM side?

No accuracy diff on ATOM, @cpersson-amd's numbers in ROCm/ATOM#1359 show DSv3.2 gsm8k 5-shot at 0.9560 vs 0.9591 strict-match on main. ATOM already calls the indexer with preshuffle=True, so it never hit the K-cache layout bug that tanked vLLM (fixed separately in [vllm#43907)](https://github.com/vllm-project/vllm/pull/43907) — here the fp32 LN change is just a precision alignment and accuracy stays flat. #1359 keeps dtype optional (defaults None) so all other models are unchanged, and only pins DSv3.2 k_norm to fp32. Suggest landing the two together. CI's green now

PR #3457, opened 2026-06-01
Fused SplitK zero-init for FP8 a8w8 blockscale GEMMs (y_is_zeroed) + re-enable CKTile SplitK

samremes:
## Summary
SplitK (KBatch > 1) blockscale GEMMs accumulate partial products into the output `Y`
via `atomic_add`, so `Y` must be zero before launch. Today the GEMM issues that
`Y.zero_()` itself — an extra full-output memset launch. This PR lets the upstream op
that already streams over memory (the activation quant / norm+quant producer feeding
the GEMM) pre-zero `Y` as a side effect, so the GEMM can run with `y_is_zeroed=True`
and skip its own zero-init. That removes a kernel launch + redundant write, which
matters most at decode / small-M.

This PR also re-enables the CKTile SplitK support due to fixes done in ROCm/rocm-libraries#7199.

## What changed
- **GEMM plumbing:** thread an explicit output buffer + `y_is_zeroed` flag through the
  a8w8 blockscale GEMMs (CKTile, bpreshuffle-CKTile, legacy CK) and their tune
  harnesses / codegen. `y_is_zeroed` defaults to `False`, so existing callers are
  unchanged.
- **Producers:** add an optional `gemm_out_zero_init` argument to the quant /
  norm+quant ops that feed these GEMMs — `per_group_quant_hip`,
  `fused_qk_rmsnorm_group_quant`, `gated_rmsnorm_fp8_group_quant` (HIP) and
  `fused_rms_fp8_group_quant`, `fused_rms_gated_fp8_group_quant`,
  `act_mul_and_fp8_group_quant` (Triton). When provided, each kernel zeros that
  buffer's full byte range as a grid-strided prologue; the producer's own output is
  unaffected.
- **CK bump:** `3rdparty/composable_kernel` is pinned to `ad4e2e7`, which contains
  [ROCm/rocm-libraries#7199](https://github.com/ROCm/rocm-libraries/pull/7199)
  ([CK_TILE] [QuantGEMM] Fix SplitK tail handling). #7199 restores `k_batch > 1` for
  `ABQuantGemmPipelineAgBgCrCompV3` and removes the device-side
  `assert(kargs.k_batch == 1)`, enabling the SplitK blockscale atomic-add accumulation
  this fusion relies on.
- **Test:** `op_tests/test_zero_init_splitk_fusion.py`.

## Testing
`pytest op_tests/test_zero_init_splitk_fusion.py` on gfx950 — scoped to the zeroing
feature only (each producer's numerical correctness is covered by its own op_test):
- **Producer side** (all 6 producers): the fused `gemm_out_zero_init` buffer is
  byte-for-byte identical to an explicit `torch.zeros`, and the producer's primary
  output is unchanged with vs. without the buffer.
- **GEMM side:** producer-zeroed + `y_is_zeroed=True` matches the GEMM's own zero-init
  across decode shapes (splitK=2), with a dirty-buffer negative control confirming that
  `y_is_zeroed=True` without a real zero-init genuinely fails.

  > Note: the GEMM-side split-K path requires CK containing #7199, which is now the
  > pinned submodule commit (`ad4e2e7`), so `splitK > 0` no longer hits the device-side
  > assert.

## Compatibility
Opt-in and backward compatible: `y_is_zeroed` defaults to `False` and
`gemm_out_zero_init` to `None`, preserving current behavior unless both are used
together.


valarLip:
the code base is not clean

samremes:
@valarLip This is passing the CI now. I removed the vllm ci label as some of jobs in the vllm ci are broken with some dependencies.

zufayu:
@samremes Two CI failures found, both appear to be regressions introduced by this PR:

**1. Shard 0 — `test_flydsl_qk_norm_rope_quant.py` + Quant checkAllclose failed**
This PR modifies `fused_qk_rmsnorm_group_quant` (cu + header) to add the `gemm_out_zero_init` prologue. The quant correctness test is failing with checkAllclose errors, suggesting the prologue change is affecting the primary output numerics.

**2. Shard 7 — `mla_decode-absorb_fp8` checkAllclose failed**
The PR does not directly touch MLA code, but bumps the `composable_kernel` submodule to `ad4e2e7`. The new CK version appears to affect MLA decode fp8 precision as a side effect.

Could you investigate and fix these two issues? Happy to approve once CI is green.

zufayu:
@valarLip @yzhou103 — this PR changes two stable downstream contracts (`aiter/ops/quant.py`, `aiter/ops/gemm_op_a8w8.py`) by adding the `gemm_out_zero_init` / `y_is_zeroed` fusion, and bumps `3rdparty/composable_kernel` to `f33252c` (105 commits past the `ad4e2e7` the PR body cites, includes `rocm-libraries#9740`). Since the author isn't a committer of these paths, requesting an explicit sign-off from the path owners before merge.

Two asks:
1. Please add the `ci:atom` label so the end-to-end accuracy gate runs (DeepSeek-R1-0528 / Kimi-K2.7 are MLA-heavy and would exercise the fused quant + CK-bumped MLA decode path). The downstream CI jobs are currently skipped.
2. @valarLip @yzhou103 could you review the contract change and approve? The op-test CI is green, but the quant prologue + CK bump's MLA decode side-effect needs the downstream accuracy check before this is safe to merge.

zufayu:
@samremes This PR has been inactive for over 2 months. Is the fused SplitK zero-init for FP8 a8w8 blockscale GEMMs still being pursued? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

zufayu:
Closing this as announced — no response within the window. For future reference: the fused SplitK zero-init approach is sound; the blocking items (two CI regressions + path-owner sign-off) are captured in the June 29 and July 28 reviews above. Happy to see this resubmitted against the latest `main`.

PR #3464, opened 2026-06-01
fea: reduce_scatter support all dim

TennyWang1223:
## Motivation

  `custom_reduce_scatter` previously only supported scattering along dim 0, with strict shape constraints (`numel % (ngpus * pack_size) == 0`). Any other case fell back to NCCL. This PR extends the custom reduce_scatter kernel to
  support an arbitrary scatter `dim`, covering more TP/EP use cases.

## Technical Details

  - Introduce a `ReduceScatterSplitDim` enum (`kFirst` / `kLast` / `kMid`). The Python wrapper collapses an arbitrary-rank input around the scatter dim into one of three canonical `(m, n, k)` shapes before calling into C++.
  - The C++ dispatcher routes to three kernel families based on split dim:
    - **First dim**: reuses the existing vectorized `reduce_scatter_split_first_dim` (requires `numel % (ngpus * pack_size) == 0`).
    - **Last / middle dim**: new vectorized kernels plus a naive fallback for shapes that cannot be 16B-vectorized along the inner dim.
  - `CustomAllreduce.should_custom_rs(inp, dim)` centralizes all hard requirements (size cap, contiguity, `shape[dim] % world_size == 0`, plus the first-dim vectorization condition). Unsupported cases fall back to NCCL at the caller.
  - `GroupCoordinator._reduce_scatter_out_place` now computes the output shape from `dim`, preserving rank and layout instead of flattening.
  - Pybind / `ops.reduce_scatter` / `custom_all_reduce.h` signatures updated to carry the new `(m, n, k, split_dim)` arguments.

## Test Plan

 Rewrote `op_tests/multigpu_tests/test_reduce_scatter.py` to parametrize across dtypes (fp16 / bf16 / fp32), world sizes, scatter dim (first / last / middle), and both vectorized and naive-fallback shapes. Results are validated against
  `torch.distributed.reduce_scatter_tensor`.

## Test Result

  All new and updated multi-GPU tests pass; numerics match the `torch.distributed` baseline.

## Submission Checklist

### DIFF
--- a/aiter/dist/parallel_state.py
+++ b/aiter/dist/parallel_state.py
@@ -532,14 +532,15 @@ class GroupCoordinator:
         prefill_support: bool = False,
         emit_bf16: bool = False,
     ):
-        return self.fused_allreduce_rmsnorm_quant(
+        if self.device_communicator is None:
+            raise ValueError("No device communicator found")
+        return self.device_communicator.fused_allreduce_rmsnorm_quant_per_group(
             input_,
             residual_inp_,
             weight_,
             eps,
+            group_size,
             prefill_support,
-            quant_type="per_group",
-            group_size=group_size,
             emit_bf16=emit_bf16,
         )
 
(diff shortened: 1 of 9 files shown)


yitingw1 (dismissed):
Verified that this PR can fix reduce_scatter bug when dim = 1 with DCP feature.

aiter/dist/parallel_state.py
```diff
@@ -593,14 +604,25 @@ def reduce_scatter_tensor(
...
+        # kernel writes elements in linear order into this layout, so no
+        # post-kernel reshape/copy is needed.
+        out_shape = (
+            input_.shape[:dim]
+            + (input_.shape[dim] // world_size,)
+            + input_.shape[dim + 1 :]
+        )
 
         output_ = torch.empty(out_shape, dtype=input_.dtype, device=input_.device)
         if use_custom:
```
yitingw1, line 661: Here `if` branch has supported dim!=0, but the `else` branch still not. Please add dim!=0 support for `else` branch.
```python
        if use_custom:
            outplace_reduce_scatter(
                input_, output_, group_name=self.unique_name, dim=dim
            )
        else:
            torch.distributed.reduce_scatter_tensor(
                output_, input_, group=self.device_group
            )
```

yitingw1 (approved):
LGTM

valarLip (approved):

PR #3470, opened 2026-06-01
[PERF] MXFP4 (a4w4) MoE backend for gfx950

ColorsWind:
# [PERF] mxfp4_moe — a native MXFP4 (a4w4) MoE backend for gfx950

## Motivation

Kimi-K2.5 / DeepSeek-R1 serve MoE FFNs with MXFP4 (a4w4) weights. The existing FlyDSL (MLIR) a4w4 MoE path leaves performance on the table across the whole token range — it is launch/latency-bound at decode-size M and not fully HBM-bound at prefill-size M. This PR adds **`mxfp4_moe`**, a native pure-HIP MXFP4 MoE backend that is faster than FlyDSL at every M bucket on MI350X.

## Technical Details

The backend is two JIT modules plus a `fused_moe` integration:

- **`module_moe_mxfp4_aux`** — token sort, mxfp4 quant (bf16 → fp4 + e8m0, group=32), scale sort/shuffle, and the weighted scatter-reduce.
- **`module_moe_mxfp4_gemm`** — the two a4w4 GEMMs (gemm1 fused-SwiGLU, gemm2 down-proj) with codegen'd kernel-name dispatch (`gen_instances.py`) and MFMA f4f4 microkernels.
- **`fused_moe` wiring** — `mxfp4_moe_*` kernel names route to an `_mxfp4_moe_run` pipeline via a `MOEMetadata` hook; per-M kernel selection comes from a `shuffle_kind`-tagged tuned CSV, so the backend ships its own tuned rows next to the default tuning without dedup conflicts.

### Several pipelines, dispatched per token-count M

The optimal fusion/accumulation strategy flips between decode-size and prefill-size M, so we designed several pipelines and pick one per M from the tuned CSV:

- **inline-quant + atomic** — small M (BM=16): `sort → quant-fused-gemm1 → gemm2 (atomic-add reduce)`. Quant is folded into gemm1 and gemm2 atomic-adds straight into the output, removing the per-sorted-row partial buffer and the separate quant / scatter passes. (The gemm2 B-load additionally has a per-M-tuned cached vs non-temporal choice.) Wins when M ≲ BM and the path is launch/latency-bound.
- **three-stage** — mid-to-large M (BM=32 / 128): `sort → quant → sort-scale → gemm1 → gemm2 [→ scatter-reduce]`. Prologue runs as separate full-throughput kernels. At mid M (BM=32) gemm2 atomic-accumulates straight into the output (no separate reduce); at large M (BM=128) gemm2 writes per-sorted-row `flat_out` and a dedicated non-atomic scatter-reduce does the weighted sum — avoiding the atomic contention that grows with M. HBM-bound regime.

The core tradeoff: fusion + atomics win when operator shapes are complementary and launch overhead dominates (small M); at large M they break down — the same token is read by multiple experts (duplicated quant work) and atomic contention grows — so the separate, non-atomic path wins.

On top of the large-M non-atomic path, an optional **mxfp4-intermediate** variant (CSV-gated, enabled on the largest prefill buckets) stages gemm2's `flat_out` as packed fp4 + e8m0 (mxfp4-out epilog) and reads it back with `scatter_reduce_q`, cutting the reduce's HBM traffic ~3.8×.

## Test Plan

**Machine / shape**

- AMD Instinct **MI350X** (gfx950)
- **Kimi-K2.5, TP=4**: NE=385, H=7168, D_INTER=512, TOPK=9
- Routing: **1 shared expert (always selected) + 8 routed experts**, softmax with a mild per-expert popularity skew (skew=0.5).
- Both backends fed the **same** per-1x32 MXFP4 weights (quantized once).
- FlyDSL baseline: ROCm/aiter @ `ca781d81`.

**Cache / timing**

- Correctness: end-to-end pipeline vs a full-precision bf16 ground truth.
- Performance: end-to-end MoE wall-clock (`perf_counter` + `synchronize`), median of 30 iters, **warm cache** (weights resident in L2; no per-iter flush). `mxfp4_moe` is driven through the production `fused_moe` entry — per-M kernel dispatch comes straight from the tuned CSV (including the mxfp4-intermediate path on the largest buckets), so the numbers carry real host-side dispatch overhead; FlyDSL is timed at its kernel stages (best block_m reported). Same routing and fp4 weights for both — if anything this is conservative for `mxfp4_moe` at small M.

### Correctness

End-to-end `cos(kernel, bf16_truth)` is **0.975–0.978** across the full M range and never worse than the fp4 reference — all pass.

### Performance (end-to-end, mxfp4_moe vs FlyDSL)

| M     | mxfp4_moe (µs) | flydsl (µs) | speedup | cos(mxfp4_moe, flydsl) |
|-------|----------------|-------------|---------|------------------------|
| 4     | 99.3           | 178.3       | 1.80×   | 0.985 |
| 8     | 116.2          | 201.7       | 1.74×   | 0.987 |
| 16    | 153.8          | 233.6       | 1.52×   | 0.987 |
| 32    | 204.5          | 300.3       | 1.47×   | 0.987 |
| 64    | 254.8          | 342.8       | 1.35×   | 0.987 |
| 128   | 304.2          | 412.4       | 1.36×   | 0.987 |
| 256   | 363.3          | 471.7       | 1.30×   | 0.987 |
| 4096  | 999.7          | 1115.2      | 1.12×   | 0.987 |
| 8192  | 1399.0         | 1785.9      | 1.28×   | 0.981 |
| 16384 | 2250.5         | 3176.7      | 1.41×   | 0.981 |

`mxfp4_moe` is faster than FlyDSL across the entire token range (**1.12×–1.80×**). At the largest buckets (M ≥ 8192) the CSV switches to the mxfp4-intermediate path; its cos vs FlyDSL drops to ~0.981 (the inherent loss of the before-sum fp4 quant) while still pulling ahead in latency.


benenzhu:
Performance bench script: https://gist.github.com/ftyghome/088870c544693da57258d22968573620

coderfeli:
@benenzhu it is launch/latency-bound' in your bench script you use s = time.perf_counter() and then time.perf_counter() - s. It counts all cpu overhead. The real run in e2e uses cudagraph that will not count.

ftyghome:
> @benenzhu it is launch/latency-bound' in your bench script you use s = time.perf_counter() and then time.perf_counter() - s. It counts all cpu overhead. The real run in e2e uses cudagraph that will not count.

Hi Felix,

Thanks for the review!
We just updated the [microbench script](https://gist.github.com/ftyghome/088870c544693da57258d22968573620) to capture 60 runs in a cudagraph. The new perf data is listed below:

| M | mxfp4 us | flydsl us | flydsl/mxfp4 | cos |
|------:|---------:|----------:|------------:|-------:|
| 4 | 43.5 | 69.9 | 1.61x | 0.9845 |
| 8 | 62.2 | 96.4 | 1.55x | 0.9868 |
| 16 | 101.6 | 132.9 | 1.31x | 0.9874 |
| 32 | 151.3 | 190.2 | 1.26x | 0.9868 |
| 64 | 206.8 | 236.6 | 1.14x | 0.9874 |
| 128 | 256.3 | 305.8 | 1.19x | 0.9875 |
| 256 | 310.2 | 341.0 | 1.10x | 0.9871 |
| 4096 | 977.4 | 1025.6 | 1.05x | 0.9875 |
| 8192 | 1381.3 | 1718.0 | 1.24x | 0.9812 |
| 16384 | 2254.4 | 3133.8 | 1.39x | 0.9811 |

lalala-sh:
> > @benenzhu it is launch/latency-bound' in your bench script you use s = time.perf_counter() and then time.perf_counter() - s. It counts all cpu overhead. The real run in e2e uses cudagraph that will not count.
> 
> Hi Felix,
> 
> Thanks for the review! We just updated the [microbench script](https://gist.github.com/ftyghome/088870c544693da57258d22968573620) to capture 60 runs in a cudagraph. The new perf data is listed below:
> 
> M	mxfp4 us	flydsl us	flydsl/mxfp4	cos
> 4	43.5	69.9	1.61x	0.9845
> 8	62.2	96.4	1.55x	0.9868
> 16	101.6	132.9	1.31x	0.9874
> 32	151.3	190.2	1.26x	0.9868
> 64	206.8	236.6	1.14x	0.9874
> 128	256.3	305.8	1.19x	0.9875
> 256	310.2	341.0	1.10x	0.9871
> 4096	977.4	1025.6	1.05x	0.9875
> 8192	1381.3	1718.0	1.24x	0.9812
> 16384	2254.4	3133.8	1.39x	0.9811

Hello @ftyghome, Could you please update the perf compare again? Collect the flydsl date with fused_moe interface, so that flydsl will be called through the moe  tuner too.

benenzhu:
> > > @benenzhu it is launch/latency-bound' in your bench script you use s = time.perf_counter() and then time.perf_counter() - s. It counts all cpu overhead. The real run in e2e uses cudagraph that will not count.
> > 
> > 
> > Hi Felix,
> > Thanks for the review! We just updated the [microbench script](https://gist.github.com/ftyghome/088870c544693da57258d22968573620) to capture 60 runs in a cudagraph. The new perf data is listed below:
> > M	mxfp4 us	flydsl us	flydsl/mxfp4	cos
> > 4	43.5	69.9	1.61x	0.9845
> > 8	62.2	96.4	1.55x	0.9868
> > 16	101.6	132.9	1.31x	0.9874
> > 32	151.3	190.2	1.26x	0.9868
> > 64	206.8	236.6	1.14x	0.9874
> > 128	256.3	305.8	1.19x	0.9875
> > 256	310.2	341.0	1.10x	0.9871
> > 4096	977.4	1025.6	1.05x	0.9875
> > 8192	1381.3	1718.0	1.24x	0.9812
> > 16384	2254.4	3133.8	1.39x	0.9811
> 
> Hello @ftyghome, Could you please update the perf compare again? Collect the flydsl date with fused_moe interface, so that flydsl will be called through the moe tuner too.


Hi updated: @lalala-sh 
Script: https://gist.github.com/benenzhu/ea38af8e5606ed5596abc1890e56a5a1 
Change by ai, not verify the correctness.
| M | mxfp4 us | bm | flydsl us | flydsl/mxfp4 | mx·fly | mx·ref | fly·ref |
|------:|---------:|----:|----------:|:------------:|-------:|-------:|--------:|
| 4 | 41.9 | 16 | 59.2 | 1.41x | 0.9845 | 0.9533 | 0.9482 |
| 8 | 61.2 | 16 | 74.6 | 1.22x | 0.9868 | 0.9524 | 0.9485 |
| 16 | 100.3 | 16 | 122.1 | 1.22x | 0.9876 | 0.9538 | 0.9514 |
| 32 | 147.8 | 16 | 165.0 | 1.12x | 0.9868 | 0.9548 | 0.9520 |
| 64 | 203.5 | 16 | 213.9 | 1.05x | 0.9874 | 0.9555 | 0.9529 |
| 128 | 254.2 | 16 | 263.8 | 1.04x | 0.9875 | 0.9562 | 0.9536 |
| 256 | 304.1 | 32 | 306.4 | 1.01x | 0.9871 | 0.9561 | 0.9538 |
| 8192 | 1137.3 | 128 | 1359.0 | 1.19x | 0.9811 | 0.9505 | 0.9541 |
| 16384 | 1821.7 | 128 | 2325.2 | 1.28x | 0.9810 | 0.9506 | 0.9541 |

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+        masked_m.data_ptr<int32_t>(), m_indices.data_ptr<int32_t>(),                           \
+        bf16_zero_ptr)
+
+    if (TOPK == 9 && D_HIDDEN == 7168) {
+        if (NE == 385) {
+            if (MB == 32) { LAUNCH(385, 9, 32, 7168); return; }
+        }
+        if (NE == 257) {
+            if (MB == 32) { LAUNCH(257, 9, 32, 7168); return; }
+        }
```
coderfeli: why so many constraints here? Could you make it more general ? I think the kernels don't have these limitation?

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+            masked_m.data_ptr<int32_t>(), m_indices.data_ptr<int32_t>(),                         \
+            bf16_zero_ptr, bf16_zero_ws_ptr, workspace_bytes)
+
+        if (TOPK == 9 && D_HIDDEN == 7168) {
+            if (NE == 385) {
+                if (MB == 16) { LAUNCH_IQ_ZI(385, 9, 16, 7168); return; }
+            }
+            if (NE == 257) {
+                if (MB == 16) { LAUNCH_IQ_ZI(257, 9, 16, 7168); return; }
+            }
```
coderfeli: +

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+            masked_m.data_ptr<int32_t>(), m_indices.data_ptr<int32_t>())
+
+        if (TOPK == 9 && D_HIDDEN == 7168) {
+            if (NE == 385) {
+                if (MB == 16) { LAUNCH_IQ(385, 9, 16, 7168); return; }
+            }
+            if (NE == 257) {
+                if (MB == 16) { LAUNCH_IQ(257, 9, 16, 7168); return; }
+            }
+        }
```
coderfeli: ++

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+        bf16_zero_ptr)
+
+    if (TOPK == 9 && D_HIDDEN == 7168) {
+        if (NE == 385) {
+            if (MB == 32)  { LAUNCH(385, 9, 32,  7168); return; }
+            if (MB == 128) { LAUNCH(385, 9, 128, 7168); return; }
+        }
+        if (NE == 257) {
+            if (MB == 32)  { LAUNCH(257, 9, 32,  7168); return; }
+            if (MB == 128) { LAUNCH(257, 9, 128, 7168); return; }
```
coderfeli: Also make quant a general usable kernel?

csrc/kernels/mxfp4_moe/gemm_a4w4/gemm2_a4w4.cuh
```diff
@@ -0,0 +1,544 @@
...
+template <int MAX_M, int NUM_EXPERTS, int K, int N_OUT, int TOPK, int BM,
+          EpilogPolicy kEpilog,
+          bool kUseNT = false,
+          int kXcdSwizzle = 0,
+          bool kMxfp4Out = false>
+__global__ void
+__launch_bounds__(256,
+                  is_nonatomic_v<kEpilog> ? 1 :
+                  ((BM == 16) ? 4 : 2))
+kernel(
```
coderfeli, line 60: https://github.com/ROCm/FlyDSL/pull/660 for gemm2, we try to impl a flydsl pipeline that's exactly the same as this one. Would you mind transfer the same style and infra for better long term maintaining?

csrc/kernels/mxfp4_moe/gemm_a4w4/mxfp4_moe_gemm.cu
```diff
@@ -0,0 +1,181 @@
...
+        stream,                                                                                  \
+        inter_sorted_quant.data_ptr(), inter_sorted_shuffled_scale.data_ptr(),                   \
+        w3_shuffled_quant.data_ptr(), w3_shuffled_scale.data_ptr(),                              \
+        sorted_expert_ids.data_ptr<int32_t>(), cumsum_tensor.data_ptr<int32_t>(),                \
+        static_cast<int>(max_sorted),                                                            \
+        flat_out_q.data_ptr(), flat_out_scale.data_ptr())
+
+    if (D_HIDDEN == 7168 && D_INTER == 512) {
+        if (NE == 385) { LAUNCH_G2_MXFP4(385, 512, 7168); return; }
+        if (NE == 257) { LAUNCH_G2_MXFP4(257, 512, 7168); return; }
```
coderfeli, line 173: hard code shapes. also couldn't be extended.

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+#define LAUNCH(D_HIDDEN_, TOPK_, NT_)                                                            \
+    aiter::mxfp4_moe::moe_scatter_reduce::launch<                                                \
+            D_HIDDEN_, TOPK_, kThreadsScatterReduce, kColsPerThread, NT_>(                       \
+        stream, M,                                                                               \
+        reinterpret_cast<const __hip_bfloat16*>(flat_out.data_ptr()),                            \
+        reverse_sorted.data_ptr<int32_t>(),                                                      \
+        sorted_weights.data_ptr<float>(),                                                        \
+        reinterpret_cast<__hip_bfloat16*>(out.data_ptr()))
+
+    if (D_HIDDEN == 7168 && TOPK == 9) {
```
coderfeli: ++

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
```diff
@@ -0,0 +1,384 @@
...
+    int64_t MB)
+{
+    (void)NE;
+    const at::hip::OptionalHIPGuardMasqueradingAsCUDA guard(device_of(flat_out_q));
+    const hipStream_t stream = at::hip::getCurrentHIPStream();
+    const int M = static_cast<int>(out.size(0));
+    const bool nt_hints = (MB >= 128);
+
+#define LAUNCH_Q(D_HIDDEN_, TOPK_, NT_)                                                           \
+    aiter::mxfp4_moe::moe_scatter_reduce::launch_mxfp4<                                           \
```
coderfeli: All input is quanted to fp4. It's ok for gsm8k but not in our many other real tests. could you also add flags to enable ?

csrc/kernels/mxfp4_moe/gemm_a4w4/gemm2_a4w4.cuh
```diff
@@ -0,0 +1,544 @@
...
+    const int wave   = __builtin_amdgcn_readfirstlane(tid / 64);
+    const int wave_n = wave;
+    const int lane   = tid % 64;
+
+    const buffer_rsrc_t A_q_rsrc =
+        make_buffer_rsrc(A_q, (uint32_t)(MAX_M * K_HALF * sizeof(__hip_fp4x2_storage_t)));
+    const buffer_rsrc_t B_q_rsrc =
+        make_buffer_rsrc(B_q,
+            (uint32_t)((long long)NUM_EXPERTS * N_OUT * K_HALF * sizeof(__hip_fp4x2_storage_t)));
+    constexpr int kAS_bound_div = kAtomic ? BM_GRID : 32;
```
coderfeli: The A_scale buffer-resource bound is 2× too small for BM=64 (atomic path).

  A_scale is laid out in fixed 32-row chunks, and the read side addresses it accordingly at line 333:
  const int chunk_base = (BM == 16) ? (m_row / BM_GRID) : (m_row / 32);
  a_scale_s_base[sub]  = (chunk_base + sub) * kAS_per_chunk_dw * 4;
  So the chunk-stride divisor is 16 for BM=16 and 32 for everything else — it is not BM. But here the bound divides by kAtomic ? BM_GRID : 
  32, which only matches the addressing for BM∈{16,32}:
For BM=64 the rsrc extent is (MAX_M/64)*kAS_per_chunk_dw, but addressing reaches (MAX_M/32)*kAS_per_chunk_dw. Once m_row ≥ MAX_M/2, the
  hardware buffer-bound check fails the scale loads and returns 0 → e8m0 = 0 → scale 2^-127 → those output rows silently collapse to ~0.
  Wrong results, no error.

csrc/kernels/mxfp4_moe/gemm_a4w4/codegen/gen_instances.py
```diff
@@ -0,0 +1,325 @@
...
+from pathlib import Path
+
+# ── Supported shape tuples ─────────────────────────────────────────────────
+# (NE, D_HIDDEN, D_INTER, TOPK)
+SHAPES = [
+    (385,  7168,      512,      9),  # Kimi-K2.5 TP=4
+    (257,  7168,      512,      9),  # DSR
+]
+
+MAX_M = 655360
```
coderfeli: this is the hard limit of topk * tokens? Then all prefill inputs >72.8K will fail and produce wrong outputs?

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
benenzhu: It's the deepseek & kimi shapes. only tested on this.

csrc/kernels/mxfp4_moe/gemm_a4w4/codegen/gen_instances.py
ColorsWind: You're right that `MAX_M` is a hard cap on the sorted token count, and inputs past it would produce wrong outputs. It's not really a bug in practice, since `MAX_M` is set far above any realistic single-call input, so nothing hits it. This change just removes the fixed cap and uses the actual sorted length (`M * topk`, i.e. `sorted_m`) for the buffer-resource bounds check, so it scales with the real input instead of relying on a large constant.

See 82e8e0ade84c1b6ef1ad4240b606d5ec8cf40226 for the fix and ee609101049c7cc05f8a4370f60e6742daddd8a3 for the 128k validation test.

csrc/kernels/mxfp4_moe/gemm_a4w4/gemm2_a4w4.cuh
ColorsWind: Good catch. A_scale is addressed in 32-row chunks (`(BM==16)?16:32`, not BM), but the bound divided by `kAtomic ? BM_GRID : 32`, so for BM=64 it covers only `max_sorted/64` chunks while addressing reaches `max_sorted/32`. Rows past `max_sorted/2` then read out of bounds, the e8m0 comes back 0, and those outputs collapse to ~0.

Fixed by matching the bound to the addressing: `(BM==16) ? BM_GRID : 32`. It was masked by the old large `MAX_M` and only surfaced once the bound was tightened to `max_sorted`. It also went unnoticed because the BM=64 gemm2 kernel is never dispatched (the tuned CSV only uses BM in {16,32,128}, and there is no BM=64 gemm1/sort to feed it).

Fix: bd5ba7938a62cad4d319d14de189167ed1bcde1e, test: 09e0e1d65401d5eadd030bd27f5250174e47d85e (BM=32 vs BM=64 on identical inputs; fails before, passes after).

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
ColorsWind: Made it opt-in: set `AITER_MXFP4_INTERMEDIATE=1` to enable the fp4 path, otherwise it falls back to the accurate bf16 reduce by default. So your other evals are safe unless you turn it on. a3c9875f808fd2da22db33740b43f8ced27e0f5a

csrc/kernels/mxfp4_moe/mxfp4_moe_aux.cu
ColorsWind: The general case is already covered by `dynamic_mxfp4_quant` in aiter, which work on arbitrary shapes. This `mxfp4_moe_quant` is a fused, shape-specialized variant (compile-time `D_HIDDEN` for vectorization, plus the optional fused zero-init) that only makes sense inside this MoE pipeline.

If you need it for more model shapes, the cheap path is to add template dispatch cases (keeps the compile-time perf) rather than making it fully runtime-general. Happy to do that as a follow-up if there's a concrete shape you need.

csrc/kernels/mxfp4_moe/gemm_a4w4/gemm2_a4w4.cuh
ftyghome: Sure. We also have a FlyDSL version of gemm1, and most of the performance already matches our HIP implementation in our tests (except M=4 case). We will upload it soon.

It would be great to sync with the ongoing gemm1/gemm2 FlyDSL work. We would be happy to work together on this and align the performance with our HIP implementation while moving toward FlyDSL.

ftyghome:
Hi @coderfeli,

Thanks for all the feedback!

We went through all your inline comments about the hardcoded shapes and constraints, and have made several commits to address them.

Initially, we only tuned and tested the Kimi-K2.5 TP4 shape in detail. The real constraint, however, is just `D_HIDDEN % 32 == 0`, which comes from the MXFP4 group size. Everything else is a template parameter.

We have now wired the aux kernels into the same CSV codegen flow as `gemm1` / `gemm2` in 28e1737. With this change, supporting a new shape only requires adding it to the CSV, without further kernel edits.

Could you please take another look when you get a chance? Please let us know if there is anything else we should address.

ftyghome:
Hi @coderfeli,

Based on the latest commits, we added a new end-to-end evaluation for the native HIP MXFP4 a4w4 MoE backend on Kimi-K2.5.

The test compares three configs:

| Name | MoE backend | `AITER_MXFP4_INTERMEDIATE` |
|---|---|---|
| base | base 2-stage baseline | n/a |
| MXFP4 intermediate OFF | native HIP MXFP4 a4w4 | OFF, bf16 intermediate/reduce |
| MXFP4 intermediate ON | native HIP MXFP4 a4w4 | ON, packed fp4 intermediate from gemm2 to scatter/reduce |

## Technical detail

The backend uses different pipelines for different token ranges:

- small M / decode, BM=16: `sort -> quant-fused-gemm1 -> gemm2 (atomic-add reduce)`
- mid-large M, BM=32 / 128: `sort -> quant -> sort-scale -> gemm1 -> gemm2`
- optional fp4 intermediate path: when `AITER_MXFP4_INTERMEDIATE=1`, gemm2 stages output packed fp4 intermediate values before scatter/reduce. By default this path is OFF and the backend uses the more accurate bf16 intermediate/reduce path.

## Test Plan

Environment:

```bash
# 4x gfx950 / MI355X, TP4
# torch / ROCm: 2.10.0+rocm7.2.3

export ATOM_USE_TRITON_MXFP4_BMM=1
export AITER_QUICK_REDUCE_QUANTIZATION=INT4
export ATOM_ENABLE_HIP_MLA_DECODE=1

python -m atom.entrypoints.openai_server --model Kimi-K2.5-MXFP4 \
    -tp 4 --kv_cache_dtype fp8 --max-model-len 10240 \
    --max-num-batched-tokens 16384 --gpu-memory-utilization 0.7 --no-enable_prefix_caching
```

## Accuracy

gsm8k, full 1319 examples, fewshot=5, eval concurrency=64.

| Config | strict-match | flexible-extract |
|---|---:|---:|
| base | 0.9697 | 0.9689 |
| MXFP4 intermediate OFF | 0.9689 | 0.9689 |
| MXFP4 intermediate ON | 0.9682 | 0.9674 |

The fp4 intermediate path only changes gsm8k by about `-0.0015` compared with intermediate OFF, which is within the expected evaluation noise.

## Performance

Median, Kimi-K2.5 TP4, ISL 8192 / OSL 1024.  
Improvement is computed against the base.

| Batch | E2E (s) | Impr | Interactivity (tok/s/user) | Impr | TPOT (ms) | Impr | Throughput/GPU (tok/s) | Impr |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 4 | 8.74 -> 7.84 | +10.3% | 106.4 -> 118.6 | +11.5% | 9.40 -> 8.43 | +10.3% | 913 -> 1017 | +11.4% |
| 8 | 10.67 -> 9.75 | +8.6% | 88.5 -> 96.9 | +9.5% | 11.30 -> 10.32 | +8.7% | 1516 -> 1659 | +9.4% |
| 16 | 15.09 -> 13.94 | +7.6% | 61.6 -> 66.8 | +8.4% | 16.23 -> 14.98 | +7.7% | 2130 -> 2313 | +8.6% |
| 32 | 21.58 -> 20.14 | +6.7% | 42.8 -> 45.9 | +7.2% | 23.34 -> 21.77 | +6.7% | 2979 -> 3206 | +7.6% |
| 64 | 33.05 -> 31.53 | +4.6% | 28.4 -> 29.5 | +3.9% | 35.27 -> 33.92 | +3.8% | 3860 -> 4033 | +4.5% |
| 128 | 51.30 -> 48.68 | +5.1% | 19.4 -> 20.4 | +5.2% | 51.43 -> 48.93 | +4.9% | 4901 -> 5177 | +5.6% |
| 256 | 92.62 -> 86.41 | +6.7% | 13.7 -> 14.2 | +3.6% | 73.18 -> 70.63 | +3.5% | 4889 -> 5294 | +8.3% |

## Detailed Performance

### Interactivity, tok/s/user

| Batch | base | MXFP4 intermediate OFF | MXFP4 intermediate ON |
|---:|---:|---:|---:|
| 4 | 106.4 | 118.4 | 118.6 |
| 8 | 88.5 | 96.1 | 96.9 |
| 16 | 61.6 | 66.3 | 66.8 |
| 32 | 42.8 | 45.4 | 45.9 |
| 64 | 28.4 | 29.4 | 29.5 |
| 128 | 19.4 | 20.1 | 20.4 |
| 256 | 13.7 | 13.7 | 14.2 |

### TPOT, ms

| Batch | base | MXFP4 intermediate OFF | MXFP4 intermediate ON |
|---:|---:|---:|---:|
| 4 | 9.40 | 8.45 | 8.43 |
| 8 | 11.30 | 10.40 | 10.32 |
| 16 | 16.23 | 15.09 | 14.98 |
| 32 | 23.34 | 22.01 | 21.77 |
| 64 | 35.27 | 34.04 | 33.92 |
| 128 | 51.43 | 49.70 | 48.93 |
| 256 | 73.18 | 73.12 | 70.63 |

### Throughput/GPU, tok/s

| Batch | base | MXFP4 intermediate OFF | MXFP4 intermediate ON |
|---:|---:|---:|---:|
| 4 | 913 | 1015 | 1017 |
| 8 | 1516 | 1647 | 1659 |
| 16 | 2130 | 2301 | 2313 |
| 32 | 2979 | 3155 | 3206 |
| 64 | 3860 | 4005 | 4033 |
| 128 | 4901 | 5082 | 5177 |
| 256 | 4889 | 5065 | 5294 |

### E2E latency, s

| Batch | base | MXFP4 intermediate OFF | MXFP4 intermediate ON |
|---:|---:|---:|---:|
| 4 | 8.74 | 7.86 | 7.84 |
| 8 | 10.67 | 9.84 | 9.75 |
| 16 | 15.09 | 14.02 | 13.94 |
| 32 | 21.58 | 20.41 | 20.14 |
| 64 | 33.05 | 31.67 | 31.53 |
| 128 | 51.30 | 49.54 | 48.68 |
| 256 | 92.62 | 89.04 | 86.41 |

csrc/kernels/mxfp4_moe/gemm_a4w4/gemm2_a4w4.cuh
coderfeli: Yes. Thanks a lot. HIP backend is also great. But currently there are ck in hip, flydsl, which we are trying to unify. Current PR add another new one, we hope to maintain in one infra for all optimization if possible.

coderfeli:
@ColorsWind conflicts

fsx950223:
Could the kernel support general weight and scale shuffle? And the performance of kernels which INTER=256 are not as good as INTER=512.

ftyghome:
Hi @fsx950223 

Thanks for the review!

> Could the kernel support general weight and scale shuffle?

AITER's `shuffle_weights` already provides both `guinterleave=True` and `guinterleave=False` variants, controlled by the `ATOM_MOE_GU_ITLV=1` environment variable. The DeepSeek-V4 serving recipe (`recipes/DeepSeek-V4.md`) enables this environment variable as well.

Our kernel performs the `silu(gate) * up` fusion within a wave, and the `guinterleave` layout is the most efficient format for this access pattern.  Our understanding is that this is a generally supported shuffle format. If we are misunderstanding the intended use case here, please let us know.


> And the performance of kernels which INTER=256 are not as good as INTER=512.

Would you mind sharing your performance results for `INTER=512`?

Our current kernel has been extensively tuned and validated on Kimi K2.5, but we would also be happy to investigate and improve performance for other shapes if there are gaps.

fsx950223:
> Hi @fsx950223
> 
> Thanks for the review!
> 
> > Could the kernel support general weight and scale shuffle?
> 
> AITER's `shuffle_weights` already provides both `guinterleave=True` and `guinterleave=False` variants, controlled by the `ATOM_MOE_GU_ITLV=1` environment variable. The DeepSeek-V4 serving recipe (`recipes/DeepSeek-V4.md`) enables this environment variable as well.
> 
> Our kernel performs the `silu(gate) * up` fusion within a wave, and the `guinterleave` layout is the most efficient format for this access pattern. Our understanding is that this is a generally supported shuffle format. If we are misunderstanding the intended use case here, please let us know.
> 
> > And the performance of kernels which INTER=256 are not as good as INTER=512.
> 
> Would you mind sharing your performance results for `INTER=512`?
> 
> Our current kernel has been extensively tuned and validated on Kimi K2.5, but we would also be happy to investigate and improve performance for other shapes if there are gaps.

This is current branch vs main branch moe performance compare 
image
It shows qwen35_397b is slower than main moe with large M, which INTER is 256

PR #3473, opened 2026-06-01
ci(atom-downstream): fix MiniMax-M2.7 via HSA_NO_SCRATCH_RECLAIM

sunway513:
## What

Fix the one InferenceX frontier model still failing the ATOM downstream
accuracy gate: **MiniMax-M2.7**.

## Root cause

HSA scratch reclaim corrupts MiniMax-M2.7's MoE scratch buffers on
gfx950. The run completes but produces garbage, so it misses the gsm8k
accuracy threshold. (Same fix validated previously in the Bedrock
benchmark runs.)

## Change

- Add a per-model env override map (`model_env_overrides`) in the
  `load-atom-models` generator that injects `HSA_NO_SCRATCH_RECLAIM=1`
  for MiniMax-M2.7.
- The value is **merged** with any `env_vars` already declared in the
  ATOM model config (newline-separated `KEY=VALUE`), so existing
  per-model env is preserved. Other models are untouched.

## Coverage

MiniMax-M2.7 continues to run under the `ci:atom_full` label (TP2,
`linux-aiter-do-mi350x-2`), not on every PR. With this fix the full
InferenceX frontier set on the ATOM lane (DeepSeek-V4-Pro,
Qwen3.5-397B-A17B-FP8, GLM-5.1-FP8, MiniMax-M2.7) is green.

## Note

Stacked on #3441 (the Kimi downstream gate adds the
`model_runner_overrides` / frontier-model plumbing this fix builds on).
Will auto-retarget to `main` once #3441 merges.

### DIFF
--- a/.github/scripts/kimi_sglang_accuracy.sh
+++ b/.github/scripts/kimi_sglang_accuracy.sh
@@ -4,8 +4,26 @@
 # PR's AITER already installed. Launches sglang.launch_server, runs lm_eval
 # gsm8k 3-shot, prints `KIMI_FLEX_EXTRACT=<value>`.
 #
-# NOTE: SGLang currently fails to load AMD Quark MXFP4 per-expert weights
-# (loader bug). This gate is wired non-blocking until that is fixed upstream.
+# Two launcher flags are REQUIRED to serve the Quark MXFP4 checkpoint on MI350X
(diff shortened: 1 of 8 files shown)


sunway513:
Superseded by a same-repo PR so CI runs without per-push fork approval. Reopening from ROCm/aiter:fix/minimax-scratch-reclaim with identical content.

PR #3483, opened 2026-06-02
[fix](moe): align MoE dims to 256 with real-dim pad in fmoe path

zhiding512:
Tuned fmoe CSVs now store each model's real (unpadded) MoE dims (e.g. gpt-oss 2880/1440/360). The kernel runs on 256-aligned dims and masks the padded tail, so derive the aligned dims and the pad widths from the real dims in a single shared helper used by both the model-CSV path and the legacy CLI sweep.

- fused_moe: align runtime/AOT lookup keys (model_dim/inter_dim) up to 256 so real dims resolve to the same tuned/compiled kernel as padded tensors.
- aot/flydsl/moe: parse real dims, pass real model_dim_pad/inter_dim_pad into the compile key for both bias variants so the AOT cache key matches the runtime compile key.
- fused_moe: drop the _get_padding_for_flydsl workaround that zeroed pad when bias was present; runtime now always passes the real pad.
- test_moe_2stage: derive aligned dims + pad from real dims, judge accuracy on the real region only (padded tail is discarded downstream and may hold NaN), and report the full tensor as a diagnostic. Remove the -hip flag.
- model_configs: rewrite gpt-oss fp4/fp8fp4 tuned/untuned CSVs to real dims.


coderfeli:
???

PR #3498, opened 2026-06-02
fix(triton/decode): honor real paged KV block stride (support non-contiguous cache)

lorri-rao:
attention_forward_decode_triton_impl's split-K kernel computed the paged block base address as `physical_block * BLOCK_SIZE_K * stride_kn`, i.e. it reconstructed the per-block stride as block_size * slot_stride and thus assumed the paged KV cache is contiguous.

This is wrong for callers whose key_cache / value_cache are non-contiguous views of the paged buffer. In particular, vLLM hybrid attention+mamba models store the KV cache block-major with K/V interleaved per block ([num_blocks, 2, block_size, num_kv_heads, head_size]); the attention halves obtained via kv_cache.unbind(0) are non-contiguous with block stride 2 * block_size * num_kv_heads * head_size, i.e. 2x what the kernel assumed. The kernel then reads the wrong block memory (straddling K and V) and produces garbage attention output.

Thread the real block stride (k_cache.stride(0) / v_cache.stride(0)) into the kernel and use it for the K/V block base instead of BLOCK_SIZE_K * stride_kn. This is a no-op for contiguous caches (stride(0) == BLOCK_SIZE_K * stride_kn) and makes the kernel correct for any regular paged layout, contiguous or interleaved.

Validated on an interleaved (stride(0)=8192) paged cache: max abs error vs a brute-force reference dropped from ~6.3 to bf16 noise (~0.09); contiguous caches are unchanged.

Internal ticket:
ROCM-25540

### DIFF
--- a/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/fwd_decode.py
+++ b/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/fwd_decode.py
@@ -332,6 +332,8 @@ def _fwd_kernel_splitK(
     stride_vn_d,
     stride_bt_b,
     stride_bt_s,
+    stride_kb,  # paged K block stride (k_cache.stride(0)); supports non-contiguous cache
+    stride_vb,  # paged V block stride (v_cache.stride(0))
     stride_az,
     stride_ah,
     stride_q_descale_z,  # FP8 descale strides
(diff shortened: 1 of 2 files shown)


micmelesse (dismissed):
LGTM. @lorri-rao Can you add a test so that we can pin this behavior.

lorri-rao:
> LGTM. @lorri-rao Can you add a test so that we can pin this behavior.

Thanks for looking into this, I just added a test @micmelesse

micmelesse (approved):
LGTM

vgokhale (approved):

PR #3504, opened 2026-06-03
Moe a8w4 optimization for decode 

lburzawa:
## Motivation

Update kernel after tuning performed on 450 silicon.

## Technical Details

- Separate decode and prefill kernels for now pending further testing.
- Use 2 buffers during decode.
- Load bias using TDM.
- Separate some TDM waits to reduce stalls.
- Use warp hints during prologue.
- Add mod M on gather offsets to prevent mem faults.

## Test Plan

Correctness tests on FFM and performance on silicon.

## Test Result

UTs pass and performance improves by up to 1.5 Tbs.

## Submission Checklist


aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
```diff
@@ -549,7 +1098,7 @@ def _moe_gemm_a8w4(
                 cur_x, 0, "e4m3", cur_w, cur_w_scales, "e2m1", acc
             )
 
-        gl.amd.gfx1250.tdm.async_wait((NUM_BUFFERS - 2 - k_ep) * NUM_TDM_OPS)
+        gl.amd.gfx1250.tdm.async_wait((NUM_BUFFERS - 2 - k_ep) * NUM_TDM_OPS + 1)
```
nsusanto: Is there a reason why we have a +1 in both decode and prefill epilogue? I'm assuming its from the extra TDM op from bias. if so I think we should gate based whether bias is not none.

I think we can just increment NUM_TDM_OPS in the "if B is not None:" check

aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
```diff
@@ -317,36 +317,578 @@ def _moe_gemm_a8w4(
...
+        )
+
+    read_idx = 0
+    write_idx = 0
+    for _ in gl.static_range(NUM_BUFFERS):
+        gl.amd.gfx1250.tdm.async_load(
+            w_desc,
+            [off_w_n, write_idx * PACKED_BLOCK_K_W],
+            w_buffer.index(write_idx % NUM_BUFFERS),
+            warp_used_hint=12,
```
nsusanto: not sure warp_used_hint does but is there a reason why it only appears on prefill main loop and not the others?  I'm assuming this is some tuning config.

aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
lburzawa: Yes it's from bias. We can't increment NUM_TDM_OPS because bias would only be used with one buffer not multiple. But yes we should gate based on it being None or not

aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
lburzawa: Removed it in newest version. It controls how many warps issue each tdm load using a bitmask.

aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
lburzawa: Fixed now

vgokhale (approved):

PR #3524, opened 2026-06-03
Fix OPUS warning on gfx1250

Boss2002n:
Only show the error when someone calls gemm_a16w16_opus or opus_gemma16w16_tune on a non-gfx950 device
Currently it shows up on every gfx1250 run


aiter/ops/opus/__init__.py
```diff
@@ -70,12 +68,11 @@ def _stub(*_args, **_kwargs):
     # propagate up through `from aiter.ops.opus import *` in
     # aiter/__init__.py, where it would be caught by the surrounding
     # try/except and silently disable the 30+ subsequent op imports.
-    warnings.warn(
```
valarLip, line 51: the descprition said "Only show the error when someone calls gemm_a16w16_opus", but here just remove the warning?

Boss2002n: import aiter nags every gfx1250 box even though opus is never touched there if we use warning like this 

the error part still exists below the warning which is thrown only when we use OPUS on gfx1250

valarLip (approved):

PR #3532, opened 2026-06-04
fix(moe-tune): bound registration-barrier deadlock + harden tune() reporting; add opt-in tuning depth

jhinpan:
## Summary

Hardens the `ck_gemm_moe_2stages_codegen` MoE tuner (`gemm_moe_tune.py` + `aiter/utility/mp_tuner.py`) and adds an opt-in benchmark-depth knob. **No change to which kernel config is selected or its measured performance.**

### Changes

**1. Bound the worker-registration barrier (deadlock fix).**
Registration uses a `Manager().Barrier(mp_num)` rendezvous (replacing a blind `time.sleep(3)`) so each `get_pid` task lands on a distinct worker. But the wait was unbounded: if a worker dies before reaching the barrier (segfault re-importing torch, OOM-kill, SIGKILL), every surviving worker — and then the parent's `.get()` — blocks **forever**. A real hazard on fault-prone fp4/i4 sweeps.
- `get_pid` now waits with `timeout=REGISTRATION_BARRIER_TIMEOUT` and returns its own pid on timeout/broken barrier (degrading to "register whoever showed up").
- The parent collects pids via `register_gpu_map`, bounding each `.get` with `REGISTRATION_GET_TIMEOUT`; a worker that fails to register leaves a recoverable gap (its tasks surface as mapping errors and retry via the existing pool-restart path) instead of hanging.
- Keeps the safe polling-interval improvements (`check_interval` 10s→1s, loop sleep 1s→0.2s) and the registration-manager teardown.

**2. Fix a `tune()` crash in failed-case reporting.**
`_, stage, kname, blockM = tag` raised `ValueError: too many values to unpack` whenever a failed candidate surfaced a bare `info` tuple instead of the usual 4-tuple tag — taking down the whole run on the first failed candidate (late-failing divisibility-invalid candidates are common). Now unpacks defensively via `_unpack_tag`.

**3. Drop dead CK split-K kernel-list block.**
`gen_2stages_task` built `ck_stage1_splitk_kernels` for fp8-blockscale shapes but never benchmarked or selected from it (the candidate loop iterates only `ck_stage1_kernels`; the only `ksplit` reaching selection comes from FlyDSL `_skN` names). Removed — pure wasted work.

**4. Opt-in `--tune_iters` / `--tune_warmup`.**
The tuning sweep always benchmarked at `run_perftest`'s defaults (101/2); the existing `--iters`/`--warmup` only reach `--run_config`. These flags set the tuning-sweep depth. **Default unchanged** (per-candidate kwargs stay `{}` → identical output); `--tune_iters` floors at 40 to stay above the IQR outlier-filter threshold so the winner doesn't flip on near-ties.

### Rejected: per-candidate datagen dedup (negative result, documented in-code)

I prototyped grouping same-`(shape, stage, block_m)` candidates so `generate_data_2stages` + the torch reference run **once per group** instead of once per candidate (datagen is ~200ms/candidate). It **measured 2.4× slower** on MI350X i4 (53.4s vs 22.6s, 4-shape warm): datagen is already parallelized across GPUs as independent tasks, and grouping pins each group's *dominant, uneven kernel benchmarks* serially onto one GPU, wrecking load balance. Not included; a comment in `tune()` warns against re-attempting it without solving the load-balance / data-locality trade-off.

### Validation (MI350X, gfx950)

4-shape i4 (`a16wi4`, per_1x32) warm A/B, `--mp 8`:
- **Perf-neutral**: 23.8s vs 22.6s baseline (within run-to-run noise).
- **Selection preserved**: `block_m` / `ksplit` / tile / stage2 kernel match baseline; total µs within the baseline's measured noise band. (The stage1 `k_batch` winner is a near-tie that flips run-to-run even at baseline — flagged so it isn't read as a regression.)
- Opt-in `--tune_iters 40 --tune_warmup 1` smoke-tested: completes, all candidates pass, selection in-band.


zufayu:
@jhinpan This PR has been inactive for over 2 months. Is the moe-tune registration-barrier deadlock fix still applicable? If you still plan to pursue it, please rebase onto current main and push an update. If there's no response within ~2 weeks, we'll close this to keep the PR queue clean — it can always be reopened or resubmitted.

PR #3543, opened 2026-06-04
[gfx12] enable per-tensor scaled-Q/K/V attention

quintinwang5:
## Motivation

Support  per-tensor scaled-Q/K/V attention
## Technical Details

For this kernel, Q/K/V is fp8, their scales are per-tensor scalar


sunway513:
@quintinwang5 — MI455X war room: per-tensor scaled Q/K/V attention (GPT-OSS pa) CI is red on **Check Code Style with Black** and **check-signal**. Black should be a quick local fix (run black and push); could you also check the check-signal failures? This is the GPT-OSS pa kernel on our critical path. Thanks.

valarLip (approved):

PR #3571, opened 2026-06-06
ci(sglang-downstream): add MoRI EP accuracy gate (guards moe_kernels stage-2 reduce)

sunway513:
## What

Add an AITER **downstream accuracy gate for the MoRI expert-parallel path**,
in direct response to the SemiAnalysis/InferenceX bug report
(sgl-project/sglang#27194).

## The bug this guards

On the MoRI EP backend, a small per-rank dispatch buffer **silently
corrupted decode output**: gsm8k dropped to 0 while the model kept
producing fluent text and even ran ~1.5% faster. Root-caused to AITER
(`moe_kernels.py` stage-2 reduce). A silent correctness failure that unit
and perf tests cannot catch — exactly the class that needs a real-model
accuracy gate.

| `SGLANG_MORI_NUM_MAX_DISPATCH_TOKENS_PER_RANK` | gsm8k |
|---|---|
| 32 / 64 | 0.00 |
| 256 | 0.94 |

## How

Reuses SGLang's own registered MoRI EP regression suite on `amd/aiter-ci`
(`stage-c-test-large-8-gpu-amd` / `test_moriep_small.py`):

- DeepSeek-V3, TP8 / EP8 / DP8, `--enable-dp-attention`
- `--moe-a2a-backend mori`, `--attention-backend aiter`
- MTP (EAGLE, num-draft-tokens=4)
- `SGLANG_MORI_NUM_MAX_DISPATCH_TOKENS_PER_RANK=128` (the small-buffer regime)
- gated on **gsm8k >= 0.90**

The MoRI EP stage-2 reduce is a kernel path, independent of the specific
DeepSeek model, so this exercises the same AITER code that regressed.
`run_on_pr` + `run_on_schedule` on `linux-aiter-do-mi350x-8`.

## Note

If `deepseek-ai/DeepSeek-V3-0324` is not staged on the runner `/models`
cache, the first run may hit a long HF download (same staging dependency
as other large-model gates). The CI log will show whether it resolves from
cache; if not, a staging request follows.


gyohuangxin:
```
========== TIMINGS BEGIN ==========
========== TIMINGS END ==========
+------------+------------------------------+-----------+-------------+
| Hardware   | Suite                        | Nightly   | Partition   |
|------------+------------------------------+-----------+-------------|
| AMD        | stage-c-test-large-8-gpu-amd | True      | full        |
+------------+------------------------------+-----------+-------------+
No tests found for hw=AMD, suite=stage-c-test-large-8-gpu-amd, nightly=True
This is expected during incremental migration. Skipping.

+ echo ''
+ exit 0
```

@bingxche Could you please check which tests suite should be used?

.github/scripts/sglang_downstream.py
```diff
@@ -106,6 +106,43 @@
...
+        "model_id": "deepseek-ai/DeepSeek-V3-0324",
+        "model_path_env": "DEEPEP_MODEL_PATH",
+        "draft_model_id": "lmsys/DeepSeek-V3-NextN",
+        "draft_model_path_env": "DEEPEP_NEXTN_MODEL_PATH",
+        "test_type": "Accuracy (MoRI EP low-latency)",
+        "timeout_minutes": 90,
+        "extra_exec_args": "",
+        "test_command": "python3 registered/amd/test_moriep_small.py TestLowLatency",
+        "run_on_pr": False,
+        "run_on_schedule": False,
```
gyohuangxin:
```suggestion
        "run_on_schedule": True,
```

bingxche:
Hi @gyohuangxin  Could you please trigger the test added in this PR? Because `run_on_pr` is false, it will not be automatically executed in the PR CI checks.

PR #3576, opened 2026-06-06
[gfx1250][FLYDSL] mha with qkdim 192 for DSv3 & KIMI K2

Zzz9990:
## Motivation

Add FlyDSL-based Flash Attention forward kernel for gfx1250, targeting DSv3/KIMI-style models with `D_qk=192, D_v=128, bf16` variable-length (varlen) THD layout.

- **New FlyDSL MHA kernel** (`aiter/ops/flydsl/kernels/fmha_gfx1250/`): hand-scheduled FMHA forward using WMMA, TDM loads, and 4-wave thread-group with LDS ping-pong. Supports causal/non-causal, variable seqlen (cu_seqlens), GQA, return_lse, and seqlen_k==0 zero-fill.
- **Integration into `flash_attn_varlen_func`** (`aiter/ops/mha.py`): auto-dispatches to FlyDSL when `gfx1250 + D_qk=192 + D_v=128 + bf16`; falls through to Triton/CK otherwise. Only 27 lines added, zero modification to existing code paths.
- **Comprehensive test suite** (`op_tests/test_mha_flydsl_varlen.py`): 124 test cases covering sq==sk, sq!=sk, sq<<sk (decode-like), multi-batch, chunked prefill (cu_q!=cu_k), seqlen_k==0, mixed zero/nonzero batches, causal x return_lse combos. Includes FlyDSL vs Triton performance comparison.

## Technical Details

| Component | File | Description |
|---|---|---|
| Prologue | `fmha_prologue.py` | Q load via buffer_load, K/V address gen, TDM descriptor setup |
| Core loop | `fmha_core_loop.py` | Interleaved GEMM1(QK) + softmax + GEMM2(PV) with fine-grained VGPR bank scheduling |
| Schedule | `fmha_schedule.py` | 160-row per-WMMA instruction schedule table (96 GEMM1 + 64 GEMM2 WMMAs) |
| Top-level | `fmha_kernel.py` | scf.for_ KV loop, ping-pong TDM, causal/OOB masking, O writeback |

- **Tile sizes**: BLOCK_M=128, TILE_N=128, 4 SUs per tile
- **TDM (Tensor Data Mover)**: async K/V loads to LDS with per-SU OOB protection
- **Ping-pong double buffering**: K_a/K_b + V_a/V_b LDS regions, overlapping compute with next-tile loads
- **VGPR bank allocation**: `@llvm.amdgcn.set.vgpr.bank` intrinsic for zero bank-conflict WMMA scheduling

## Submission Checklist


coderfeli:
The gate asserts on bias/alibi/sink but silently ignores `dropout_p`/`window_size`/`block_table`/`logits_soft_cap`. `block_table` is the bad one — a paged KV call gets computed as dense and returns wrong results with no error. Suggest folding these into the gate condition so non-default values fall through to CK instead.

coderfeli:
`fmha_kernel.py:48` inserts the file dir into global `sys.path` so the modules can do `from fmha_prologue import ...`. Could we add an `__init__.py` to `fmha_gfx1250/` and use relative imports (`from .fmha_prologue import ...`) instead? The sys.path insert leaks process-globally and the bare names (`fmha_prologue`/`fmha_core_loop`/`fmha_schedule`) can shadow same-named modules.

op_tests/test_mha_flydsl_varlen.py
```diff
@@ -0,0 +1,372 @@
...
+    bash run_mha_flydsl_varlen.sh
+"""
+
+import math
+import os
+import sys
+
+import torch
+
+os.environ.setdefault("ENABLE_CK", "0")
```
valarLip: ....

valarLip (changes requested):

op_tests/test_mha_flydsl_varlen.py
Zzz9990: removed

valarLip (approved):

PR #3614, opened 2026-06-08
flydsl fp8 einsum 

ganyi1996ppo:
## Summary

  FP8 einsum kernel (flydsl, gfx950) — `Z[b,h,d] = einsum('bhr,hdr->bhd', X, Y)` with DeepGEMM packed-UE8M0 scales
  Lands a from-scratch flydsl implementation of the FP8 multi-head einsum used in attention QK/PV (and matmul-with-batched-rhs patterns), targeting **MI355X**. Both BF16 and
   FP8-with-D-128-group-quant output variants are supported through a single underlying kernel factory plus a unified `fp8_einsum()` user-facing entry point.

  Headline:

  - **1965 TF** at H=16 D=1024 R=4096 B=32768, BF16 output
  - Built-in autotune table covering 17 BF16 + 17 QZ tuned shapes
  - FP8 + per-(B, H, D/128) fp32 scale output in a single launch (no separate quant pass)
  
  ## Interface
  
  ```python 
def fp8_einsum(
    x_fp8: torch.Tensor,
    y_pre: torch.Tensor,
    sx: torch.Tensor,
    sy: torch.Tensor,
    *,
    out_dtype: torch.dtype = torch.bfloat16,
    transpose_scale: bool = False,
    stream: torch.cuda.Stream | None = None,
) -> tuple[torch.Tensor, torch.Tensor | None]:
      """Returns (z, sz). For bf16 output, sz is None. For fp8 output, sz is the fp32
      per-(B, H, D/128) scale tensor in the layout selected by transpose_scale."""
```

  Scale ABI matches DeepGEMM SM100 packed-UE8M0: 4 UE8M0 exponent bytes per consecutive 128-K-col group packed little-endian into one i32.

  Mirrors the preshuffle_gemm async-pingpong structure, adapted to feed packed UE8M0 scales directly into the MFMA opsel byte:

  - async-A: raw_ptr_buffer_load_lds with swizzle-the-source — A gmem→LDS direct, no register transit
  - B, sx_i32, sy_i32 all carried in registers across iters
  - Scale fed DIRECTLY to mfma_scale_f32_16x16x128_f8f6f4 via opsel byte index — no fp32 promote step
  - a0_prefetch: first A pack of NEXT tile pre-loaded from LDS to regs before next-iter MFMA fires (hides 1 ds_read from the inner-loop critical path)
  - Unrolled pingpong: each loop iter handles 2 K-tiles (compute on pong + prefetch ping, then compute on ping + prefetch pong)
  - Single s_waitcnt(vmcnt=N_b+N_sx+N_sy)+lgkmcnt(0) per half AFTER compute_tile
  
  LDS layout:
  - A fp8 ping/pong stages
  - sx slab (persistent, loaded once per WG)
  - sy slab (persistent, loaded once per WG)
  - qz amax slab (only allocated when quant_output=True)

  QZ epilogue (fp8 output mode)

  Per WG: tile_m rows × tile_n/128 D-128 blocks.

  Reduction structure depends on tile_n:
  - tile_n == 128: all 4 waves contribute to the single D-128 block per WG
  - tile_n == 256: waves 0+1 → block 0, waves 2+3 → block 1

  Algorithm per (mi, ii, d128):
  1. Local amax over the lane's num_acc_n accs at this (mi, ii)
  2. In-wave 16-lane reduce via xor butterfly (ds_bpermute × 4)
  3. Cross-wave reduce via LDS slab + barrier (waves_per_d128 partials)
  4. s = max(amax, 1e-30) / 448
  5. Clamp scaled accs to [-448, 448] (cvt_pk_fp8_f32 does NOT saturate — produces NaN above 448), then RNE convert + 1-byte store
  
  OOB-row predicate on scale store: when B < tile_m, padding rows would otherwise compute amax=0 and overwrite valid scales for other d128 blocks due to linear-index wrap.

### DIFF
--- a/aiter/ops/flydsl/kernels/fp8_einsum.py
+++ b/aiter/ops/flydsl/kernels/fp8_einsum.py
@@ -1,46 +1,58 @@
-"""DeepGEMM-style clean FP8 einsum for gfx950 — v2 pipeline (preshuffle-style).
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
 
-Reference shape: H=8 D=8192 R=8192 B=2048, tm=128 tn=128 tk=256, bsw=4,
-sx+sy in LDS, B in regs via pingpong → ~1896 TF peak.
+"""Clean, submission-ready FP8 blockscaled batched einsum for gfx950 / MI355X.
+
+Computes  Z[b,h,d] = sum_r X[b,h,r] * Y[h,d,r]   (einsum 'bhr,hdr->bhd')
+in FP8 e4m3 with DeepGEMM-style packed-UE8M0 per-128-K block scales.
+
+This is the "reg-B" path: the proven-correct and peak-performant pipeline.
+  - A (X tile)  : gmem -> LDS via raw_ptr_buffer_load_lds (compiler intrinsic,
+                  no register transit; swizzle-the-source for bank-conflict-free
+                  ds_read). Double-buffered ping/pong across K-tiles.
+  - B (Y tile)  : gmem -> registers via load_b_tile (carried across iters).
+  - sx (X scale): staged once per WG to a small LDS slab, then read per K-tile.
+  - sy (Y scale): gmem -> registers per K-tile (tiny; packed UE8M0).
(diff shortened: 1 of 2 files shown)


ganyi1996ppo:
replace this with https://github.com/ROCm/aiter/pull/3614

PR #3615, opened 2026-06-08
[flydsl] Skip dead zero-fill on split-K MoE stage1 buffer

okorzh-amd:
The FlyDSL MoE stage1 kernel writes every element of tmp_out before any read, so the torch.zeros zero-fill is dead work. PyTorch 2.10 + ROCm 7.2 does NOT fast-path Tensor.zero_() on bf16/fp16 to hipMemsetAsync; it instead emits a FillFunctor elementwise kernel (~3.8 us per decode layer observed on Kimi-K2.6 MXFP4 / MI355X). Allocate with torch.empty instead.

Validated end-to-end on Kimi-K2.6-MXFP4 / MI355X / TP=4 across a full serving sweep (ISL/OSL = 1024,1024 / 1024,8192 / 8192,8192 / 8192,1024; conc 4..64) with no throughput or accuracy regression; gsm8k strict exact-match 0.940 (conc 4) and 0.952 (conc 64).

## Motivation

The FlyDSL MoE stage1 kernel writes every element of tmp_out before any read, so the torch.zeros zero-fill is dead work. PyTorch 2.10 + ROCm 7.2 does NOT fast-path Tensor.zero_() on bf16/fp16 to hipMemsetAsync; it instead emits a FillFunctor elementwise kernel (~3.8 us per decode layer observed on Kimi-K2.6 MXFP4 / MI355X). Allocate with torch.empty instead.

## Technical Details
`flydsl_moe_stage1` allocates its split-K accumulation buffer (`tmp_out`) with `torch.zeros(...)`. The FlyDSL stage1 kernel writes every element of `tmp_out` before any read, so the zero-fill is dead work.

On PyTorch 2.10 + ROCm 7.2 this is not free: `torch.zeros` (and `Tensor.zero_()`) on a bf16/fp16 tensor is **not** fast-pathed to `hipMemsetAsync` — it launches a separate `at::native::FillFunctor` elementwise kernel. In the Kimi-K2.6 MXFP4 decode path this showed up as an extra ~3.8 µs kernel per layer.

This PR allocates the split-K `tmp_out` with `torch.empty(...)` instead, which is safe because the kernel fully overwrites the buffer. Only the `_is_splitk` branch in `flydsl_moe_stage1` is touched; the stage2 `torch.zeros if accumulate` path (which relies on zero-init for accumulation) is unchanged.

- File: `aiter/ops/flydsl/moe_kernels.py`
- Net change: `torch.zeros` → `torch.empty` for the split-K stage1 buffer (+ comment)

## Test Plan
End-to-end vLLM serving benchmarks on **Kimi-K2.6-MXFP4 / MI355X / TP=4**:
- Perf sweep over ISL/OSL = `1024,1024`, `1024,8192`, `8192,1024` with concurrency `{4, 8, 16, 32, 64}` (15 points).
- gsm8k accuracy (250 samples) at concurrency `4` and `64`.
- Decode-layer kernel trace inspection to confirm the `FillFunctor` kernel is gone from the MoE stage1 path.

## Test Result
- All 15 perf points completed with no throughput regression vs. the `torch.zeros` baseline. 1% geomean perf uplift. 
- `FillFunctor` (~3.8 µs/layer) confirmed eliminated from the decode trace.
- gsm8k strict exact-match: **0.940** (conc 4), **0.952** (conc 64) — on par with baseline, no accuracy regression.


aiter/ops/flydsl/moe_kernels.py
```diff
@@ -1204,7 +1204,7 @@ def flydsl_moe_stage1(
 
     if _is_splitk:
         torch_tmp_out_dtype = dtypes.bf16 if _base_out_dtype == "bf16" else dtypes.fp16
-        tmp_out = torch.zeros(
+        tmp_out = torch.empty(
```
okorzh-amd, line 1207: closing PR as there is a failed CI test identifying requirement to actually zero memory before the ext kernel in the case of splitk > 1.

PR #3644, opened 2026-06-10
Updated MiniMax M2.5 FMoE tuned configs with new 64x384 kernels

akii96:
## Summary

Retuned the MiniMax M2.5 FMoE config after ROCm/aiter#3630 added the new 64x384 kernels.

The config is merged, sorted, and deduped so each FMoE shape keeps the best measured entry without overlapping other runtime FMoE configs.

## Validation
Checked no duplicate shape keys remain in the MiniMax file or against the merged FMoE config set. 

cc @sunway513  Might be of interest for ATOM too


JohnNikolay84:
should we update it for glm  too ?

akii96:
@JohnNikolay84 The 64x384 kernels from your PR targets `QuantType.per_1x128`, but GLM47 uses `QuantType.per_Token`, so those should not be picked here is my understanding 

BTW do you know why the CI keeps failing on ASM PA decode kernel ? Flakiness or something else ?

JohnNikolay84:
@akii96 I see 32x384 kernel being used here for example:

./aiter/configs/model_configs/glm47_fp8_tuned_fmoe.csv

but indeed the quantization method is per token. I confused kernels, sorry :)

yzhou103 (approved):

PR #3659, opened 2026-06-10
test: skip pa_decode_bf16_asm off gfx1250

yhl-amd:
## Summary\n- Skip test_pa_decode_bf16_asm.py on non-gfx1250 runners before setting the CUDA default device\n- Use aiter.dtypes.fp8 in the gfx1250-only test so the tensor dtype matches AITER's registered fp8 dtype\n\n## Validation\n- python3 -m py_compile op_tests/test_pa_decode_bf16_asm.py\n- git diff --check\n- Simulated non-gfx1250 import path with stubbed torch/pandas/aiter modules; verified SystemExit(0) before CUDA setup\n\nThis addresses the MI300X/gfx942 CI failure where the ctypes wrapper rejected torch.float8_e4m3fn because AITER registers fp8 as float8_e4m3fnuz on gfx942.


yhl-amd:
Closing: this was opened from the stacked PR branch and includes unrelated commits. I will move the fix to the correct branch/PR.

PR #3667, opened 2026-06-10
Guard `v_cvt_pk_fp8_f32` and `v_cvt_pk_bf8_f32` asm in aiter_opus_plus.h for gfx11 family

menglcai:
According to https://llvm.org/docs/AMDGPU/AMDGPUAsmGFX11.html , `v_cvt_pk_fp8_f32` and `v_cvt_pk_bf8_f32` are not available on gfx11xx. When kernels that include aiter_opus_plus.h (e.g. `module_rmsnorm_quant` triggered by `op_tests/test_rmsnorm2d.py`) are JIT-compiled on RDNA GPUs, the compiler reports `instruction not supported on this GPU`

Add `#if defined(__gfx11__)` guards to `fp32_to_fp8_scaled_x2` and `fp32_to_bf8_scaled_x2`, returning a dummy value on gfx11 targets. This allows affected JIT modules to compile successfully.

**Note**: On gfx11, the dummy return prevents compile failure; invoking the FP8/BF8 quantization path on gfx11 at runtime will produce incorrect results.

### DIFF
--- a/csrc/include/aiter_opus_plus.h
+++ b/csrc/include/aiter_opus_plus.h
@@ -44,6 +44,9 @@ OPUS_D fp32x2_t pk_mul_f32(fp32x2_t a, fp32x2_t b)
 template <typename S, std::enable_if_t<std::is_same_v<S, fp32x2_t>, bool> = true>
 OPUS_D decltype(auto) fp32_to_fp8_scaled_x2(const S& s, float inverted_scale)
 {
+#if defined(__gfx11__)
+    return fp8x2_t{};
+#else
     fp32x2_t tmp = pk_mul_f32(s, fp32x2_t{inverted_scale, inverted_scale});
 #if defined(__gfx942__)
     constexpr float hi = 240.0f, lo = -240.0f;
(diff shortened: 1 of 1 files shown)


csrc/include/aiter_opus_plus.h
```diff
@@ -44,6 +44,9 @@ OPUS_D fp32x2_t pk_mul_f32(fp32x2_t a, fp32x2_t b)
 template <typename S, std::enable_if_t<std::is_same_v<S, fp32x2_t>, bool> = true>
 OPUS_D decltype(auto) fp32_to_fp8_scaled_x2(const S& s, float inverted_scale)
 {
+#if defined(__gfx11__)
+    return fp8x2_t{};
```
mgehre-amd, line 49: Maybe add "assert(false);" or similar to abort at runtime instead of silently producing wrong values?

menglcai: Done in 2c94acd. Add assert(false) to make unsupported-arch failures explicit rather than silently wrong.

klin2024:
This is the same solution as #3860, which has already been merged. Closing this PR.

PR #3693, opened 2026-06-12
Add qwen3.5 397b mxfp4 GEMM tuning

Emmanuel0612:
## Submission Checklist


yzhou103:
Please avoid recording hipBLASLt kernel names in this CSV. hipBLASLt solution/kernel names are not stable across library versions, so a CSV generated with one hipBLASLt version may fail or select invalid solutions with another version

yzhou103:
Please also run:

`git add --renormalize aiter/configs/model_configs/qwen3_5_397b_*.csv`
and include the resulting line-ending normalization in this PR. These CSV files currently appear to be committed with CRLF line endings, while .gitattributes requires CSV files to use LF. Without normalization, checking out the branch leaves the working tree dirty.

aiter/configs/model_configs/qwen3_5_397b_untuned_gemm.csv
```diff
@@ -0,0 +1,15683 @@
+M,N,K,bias,dtype,outdtype,scaleAB,bpreshuffle
```
yifehuan, line 1: This CSV is bf16 dense GEMM, not mxfp4, which does not match the PR title. The filename is *_bf16_tuned_gemm.csv and every row is dtype=torch.bfloat16, so it is loaded by the a16w16 path. mxfp4 GEMM uses a separate path (gemm_op_a4w4 + a4w4_blockscale configs), and qwen3.5-397b already has qwen3_5_397b_fp4_tuned_fmoe.csv. Please confirm this file is in the intended path, or update the title accordingly.

aiter/configs/model_configs/qwen3_5_397b_untuned_gemm.csv
```diff
@@ -0,0 +1,15683 @@
+M,N,K,bias,dtype,outdtype,scaleAB,bpreshuffle
+1,1024,4096,FALSE,torch.bfloat16,torch.bfloat16,FALSE,FALSE
```
yifehuan, line 2: Please tune per padded_M bucket instead of per raw M. At runtime get_GEMM_A16W16_config (tuned_gemm.py) aligns M via getPaddedM(gl=0) before lookup, so these 16k rows collapse into only ~797 unique buckets, making ~95% of them redundant. Among the 514 buckets where different Ms picked different kernels, ~94% have a us spread within 10% (measurement noise). One row per (N,K)×padded_M bucket (~797 rows) would cover every M, including those unseen during tuning.

PR #3694, opened 2026-06-12
Pass --targets to ck-tile generate.py for non-gfx9 hosts

menglcai:
## Motivation

ck-tile's generate.py defaults to `--targets` gfx9,gfx950 (see [example/ck_tile/01_fmha/generate.py](https://github.com/ROCm/composable_kernel/blob/develop/example/ck_tile/01_fmha/generate.py#L102-L107)). On non-gfx9 hosts (e.g. gfx1151), the generated blob contains no matching kernels, causing mha_fwd dispatch to return t < 0 and triggering:  `"invalid argument for fmha_fwd"`

## Solution Description

According to https://github.com/ROCm/aiter/pull/3330,  `_ck_targets_flag()` already exists in mha_recipes.py to return `--targets <arch>` for non-gfx9 hosts and `""` for gfx9. Add it to blob_gen_cmd in all five cmdGenFunc_mha_* functions:
  - cmdGenFunc_mha_fwd
  - cmdGenFunc_mha_varlen_fwd (pagedkv branch, both fwd and fwd_splitkv)
  - cmdGenFunc_mha_bwd
  - cmdGenFunc_mha_varlen_bwd
  - cmdGenFunc_mha_batch_prefill

### DIFF
--- a/aiter/ops/mha.py
+++ b/aiter/ops/mha.py
@@ -12,6 +12,7 @@ from ..jit.utils.torch_guard import torch_compile_guard
 from ..jit.utils.mha_recipes import (
     compose_mha_fwd_variant_suffix_and_filter,
     get_mha_varlen_prebuild_variants_by_names,
+    _ck_targets_flag,
 )
 from ..utility import dtypes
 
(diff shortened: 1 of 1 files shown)


klin2024:
Hi reviewers, could you please take a look at this PR when you have a chance? We need to run aiter on gfx1151, and this change is important for our use case. Thanks!

PR #3698, opened 2026-06-12
[Triton] unified_attention: mask V load and output store by value head size

reger-men:
`unified_attention` sizes everything off `HEAD_SIZE = q.shape[-1]` and reuses one `dim_mask = offs_d < HEAD_SIZE` for the Q load, the K load, the V load and the output store. That's fine when the qk and v head dims match, but asymmetric-head models (MiMo: qk=192, v=128) have a V cache and an output whose per-head width is the value head size, not `HEAD_SIZE`.

With the single 192-wide mask:
- the V load reads `offs_d` 0..191 out of a 128-wide head, so it spills into the next kv-head and, on the last head, past the buffer;
- the output store writes 0..191 into a row whose per-head stride is 128, so each head clobbers the first 64 dims of the next one and the last head writes past the output buffer.

On MiMo decode that shows up as garbled logits at low concurrency and a `hipErrorIllegalAddress` (worker death) under concurrency from the last-head out-of-bounds store. Prefill is unaffected; this is decode-only.

The fix masks the V load by the V cache head stride (`stride_v_cache_2`) and the output store by the output per-head stride (`output_stride_1`), both already passed into the kernels. Four masks across the 2d and 3d kernels plus `reduce_segments`, no signature or launcher change. For symmetric models `HEAD_SIZE`, `stride_v_cache_2` and `output_stride_1` are all equal, so it's a no-op.

Validated live on MI300X (gfx942), TP=8, decode path (`--decode-attention-backend aiter`, eager), patch applied in-image:
- GSM8K, full 1319-question 5-shot: strict-match 0.9515, flexible-extract 0.9522, 0 generation errors. The triton backend baseline on the same harness is 0.9424; unpatched aiter produced garbage and scored ~0.625 on a 40-q subset.
- conc-1: coherent on control prompts (was `,#,###,#`-style garbage before).
- conc-64 burst (192 requests): 192 ok / 0 errors, no illegal-address or HIP-error hits in the serve log. Unpatched, workers died with `hipErrorIllegalAddress` under the same load.

One note for reviewers: using the stride as a stand-in for the value head size is correct for the contiguous KV-cache and output views the caller passes, but it reads as a dependency on the caller's layout. A cleaner form would be an explicit `VALUE_HEAD_SIZE: tl.constexpr` (mirroring `HEAD_SIZE`), plumbed from `v.shape[-1]` and defaulting to `HEAD_SIZE` so existing callers stay byte-identical. Happy to respin it that way if preferred; this PR is the minimal in-place fix we validated.

### DIFF
--- a/aiter/ops/triton/_triton_kernels/attention/unified_attention.py
+++ b/aiter/ops/triton/_triton_kernels/attention/unified_attention.py
@@ -80,6 +80,8 @@ def kernel_unified_attention_2d(
     TILE_SIZE: tl.constexpr,  # int must be power of 2
     HEAD_SIZE: tl.constexpr,  # int
     HEAD_SIZE_PADDED: tl.constexpr,  # int, must be power of 2
+    V_HEAD_SIZE: tl.constexpr,  # int, value head size (may differ from HEAD_SIZE)
+    V_HEAD_SIZE_PADDED: tl.constexpr,  # int, must be power of 2
     USE_ALIBI_SLOPES: tl.constexpr,  # bool
     USE_QQ_BIAS: tl.constexpr,  # bool
     USE_SOFTCAP: tl.constexpr,  # bool
@@ -134,6 +136,7 @@ def kernel_unified_attention_2d(
 
     offs_m = tl.arange(0, BLOCK_M)
     offs_d = tl.arange(0, HEAD_SIZE_PADDED)
+    offs_vd = tl.arange(0, V_HEAD_SIZE_PADDED)
     offs_t = tl.arange(0, TILE_SIZE)
     query_pos = q_block_local_idx * BLOCK_Q + offs_m // num_queries_per_kv
 
@@ -153,6 +156,10 @@ def kernel_unified_attention_2d(
         dim_mask = offs_d < HEAD_SIZE
     else:
         dim_mask = tl.full((1,), 1, dtype=tl.int1)
+    if V_HEAD_SIZE_PADDED != V_HEAD_SIZE:
+        v_dim_mask = offs_vd < V_HEAD_SIZE
+    else:
+        v_dim_mask = tl.full((1,), 1, dtype=tl.int1)
     query_mask_0 = query_pos < cur_batch_query_len
     query_mask_1 = query_offset_1 < num_query_heads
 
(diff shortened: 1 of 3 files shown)


cagrikymk:
Hello, thank you for the PR!

Having explicit constexpr argument would be nicer.

In your PR, we are relying on the fact that qk head size > v head size, as otherwise we would be loading v head dim partially.

Can we provide V_HEAD_SIZE and V_HEAD_SIZE mirroring existing approach?

Also, this requires explicit test cases as well, at least the MiMo case.

Thanks!

reger-men:
Thanks for the review, both points addressed in [5cf0f81](https://github.com/ROCm/aiter/pull/3698/commits/5cf0f81612f39433cc4f4c522becbec4f7bb5636)

cagrikymk:
Thanks!
One more thing, is it possible to check the perf. before and after these changes using:
[bench_unified_attention.py](https://github.com/ROCm/aiter/blob/main/op_tests/op_benchmarks/triton/bench_unified_attention.py)

You can do it for:
q heads=64, kv_heads=8, batch=1 ISL=8192, OSL=8192, head_size=64, block_size=64, dtype={bf16, fp8}.
q heads=64, kv_heads=8, batch=64 ISL=1, OSL=8192, head_size=64, block_size=64, dtype={bf16, fp8}.

I think compiler would make these changes no_op when qk head size = v head size but would be nice to double check.

reger-men:
Checked perf on MI300X (gfx942), bf16 and fp8, PR tip (5cf0f81) vs the pre-PR baseline (4c91afb2). Single GPU, `bench_unified_attention.py -hq 64 -hk 8 -d 64 -block_size 64 -equal_seqlens`, throughput in TFLOPS, median of 3 runs:

| config | baseline | PR tip | delta |
|---|---|---|---|
| b1, sq8192, sk8192, bf16 | 299.4 | 299.5 | +0.0% |
| b1, sq8192, sk8192, fp8  | 361.5 | 361.7 | +0.1% |
| b64, sq1, sk8192, bf16   | 24.6  | 24.6  | +0.0% |
| b64, sq1, sk8192, fp8    | 30.4  | 30.8  | +1.5% |

b64/sq1 is the small decode kernel and is noisy (per-config spread up to ~5% across repeats), so its delta is inside run-to-run noise. For symmetric qk==v heads the V_HEAD_SIZE path is a no-op vs baseline, as you expected.

cagrikymk:
@reger-men Thanks!

cagrikymk (dismissed):

cagrikymk:
CC'ing @vgokhale @k50112113 for final review to unblock the merge

vgokhale (dismissed):

reger-men:
@cagrikymk @vgokhale @k50112113 this one's just waiting on a re-approval. The force-push that addressed cagrikymk's notes (explicit V_HEAD_SIZE, and not partially loading V when the head sizes differ) dropped the earlier approvals. Checks are all green (Standard, Triton, OPUS) and the MI300X before/after is up above. Could one of you take another look?

cagrikymk (approved):

reger-men:
@vgokhale @k50112113 cagrikymk re-approved and CI's green (Standard + Triton + OPUS), this just needs the final codeowner review to merge, mind taking the last look?

PR #3700, opened 2026-06-12
[fused_moe] Opt-in fused stage1 activation-quant kernel selection

karverma-amd:
Adds AITER_FUSE_STAGE1_QUANT (default off): when a FlyDSL stage1 gate/up GEMM has a proven fused-quant (`_fp4`) sibling in the merged tuned config, swap to it so fused_moe can skip the standalone inter-stage activation-quant kernel (dynamic_mxfp4 quant + moe_sort).

## Motivation
On the FlyDSL 2-stage MoE path, the stage1 gate/up GEMM and the inter-stage activation quantization run as separate kernels: stage1 produces bf16 activations, then a standalone dynamic_mxfp4 quant + moe_sort pass converts them to fp4 with sorted e8m0 scales before stage2. Some FlyDSL stage1 kernels have a fused-quant (_fp4) sibling that emits the fp4 activations and sorted scales directly from the GEMM epilogue, which lets fused_moe skip that standalone quant/sort kernel entirely — a useful launch-latency win in decode where the MoE is latency-bound.

This PR adds an opt-in mechanism to select that fused-quant stage1 kernel when a tuned, validated variant is available, with safe fallbacks so default behavior is unchanged.

## Technical Details

Gated by AITER_FUSE_STAGE1_QUANT (default 0 = off), so behavior is identical to today unless explicitly enabled.

_get_proven_fp4_stage1(tune_file): collects the set of flydsl_moe1*_fp4 kernels that ship in the merged tuned config CSV. Membership means the kernel's codegen is AMD-validated/known-good, so it is safe to JIT-compile on demand. Result is cached per tune-file path.
_maybe_fuse_stage1_quant(kernelName1, q_dtype_a, tune_file, token): upgrades a stage1 kernel name to its _fp4 sibling only when all of the following hold — flag enabled, activation dtype is fp4x2, the kernel is a flydsl_moe1 kernel and not already fused, and a proven sibling exists. The _bnt0 (b-non-temporal) suffix is dropped before lookup because the fused variants are not tuned with that flag. Any miss returns the original kernel name unchanged.
AITER_FUSE_STAGE1_MIN_TOKENS (default 0 = no gate): the fused epilogue (quant+sort+swizzle) carries a fixed per-launch cost that is a larger fraction of a tiny GEMM, so at small token tiers it can be slightly slower. This optional threshold restricts the upgrade to token >= threshold, keeping the win at large-M decode and leaving the small-M tail on the stock split path.
Call site: hooked into the FlyDSL 2-stage selection in get_2stage_cfgs, immediately before stage1_func is bound, so the chosen kernel name flows through unchanged into the existing stage1 wrapper.
The selection logic is pure kernel-name/string handling driven by the tuned CSV — it does not change numerics or kernel codegen, only which (already-tuned) kernel is dispatched.

## Test Plan

1. Unit tests (op_tests/test_fused_stage1_quant.py, 13 cases, no GPU required):

  - _get_proven_fp4_stage1: collects only flydsl_moe1*_fp4 rows, is cached (post-cache file mutation ignored), empty set on missing file.
  - Gating/fallbacks: flag off is a no-op; upgrade when a proven sibling exists; _bnt0 suffix dropped before lookup; unchanged when no proven sibling, empty tune file, non-fp4x2 activation dtype, already-fused kernel, or non-flydsl_moe1 kernel.
  - AITER_FUSE_STAGE1_MIN_TOKENS: skips upgrade below threshold, upgrades at/above; 0 disables the gate.

All 13 passing on MI355X.

## Test Result

End-to-end accuracy (DeepSeek-R1, MoRI EP, MI355X) with AITER_FUSE_STAGE1_QUANT=1: GSM8k 1319 questions, 5-shot → 94.4%, Invalid 0.000, matching the expected R1 score (no regression).

Before - 
image

After -
image

## Submission Checklist

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -990,9 +990,7 @@ def _get_proven_fp4_stage1(tune_file):
                     if kn.startswith("flydsl_moe1") and kn.endswith("_fp4"):
                         result.add(kn)
     except Exception as e:  # pragma: no cover - defensive
-        logger.warning(
-            f"[fused_moe] could not collect fused-quant stage1 kernels: {e}"
-        )
+        logger.warning(f"[fused_moe] could not collect fused-quant stage1 kernels: {e}")
     _proven_fp4_stage1_cache[tune_file] = result
     return result
 
@@ -1041,9 +1039,7 @@ def _maybe_fuse_stage1_quant(kernelName1, q_dtype_a, tune_file, token=None):
         candidates.append(f"{kernelName1[: -len('_bnt0')]}_fp4")
     for cand in candidates:
         if cand in proven:
-            logger.info(
-                f"[fused_moe] AITER_FUSE_STAGE1_QUANT: {kernelName1} -> {cand}"
-            )
+            logger.info(f"[fused_moe] AITER_FUSE_STAGE1_QUANT: {kernelName1} -> {cand}")
             return cand
     return kernelName1
 
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -969,6 +969,85 @@ def _flydsl_stage2_wrapper(
     )
 
 
+_proven_fp4_stage1_cache = {}
+
+
+def _get_proven_fp4_stage1(tune_file):
+    """FlyDSL stage1 kernels with a fused activation-quant (``_fp4``) epilogue
+    that ship in the merged tuned config. Membership here means the kernel's
+    codegen is AMD-validated/known-good, so it is safe to JIT-compile on demand.
+    """
+    if tune_file in _proven_fp4_stage1_cache:
+        return _proven_fp4_stage1_cache[tune_file]
+    result = set()
+    try:
+        import pandas as pd
+
+        if os.path.exists(tune_file):
+            df = pd.read_csv(tune_file)
+            if "kernelName1" in df.columns:
+                for kn in df["kernelName1"].dropna().astype(str):
+                    if kn.startswith("flydsl_moe1") and kn.endswith("_fp4"):
+                        result.add(kn)
+    except Exception as e:  # pragma: no cover - defensive
+        logger.warning(
+            f"[fused_moe] could not collect fused-quant stage1 kernels: {e}"
+        )
+    _proven_fp4_stage1_cache[tune_file] = result
+    return result
+
+
+def _maybe_fuse_stage1_quant(kernelName1, q_dtype_a, tune_file, token=None):
+    """Opt-in upgrade of a FlyDSL stage1 kernel to its fused-quant sibling.
+
+    The fused stage1 kernel writes fp4 activations + sorted e8m0 scales directly
+    from the gate/up GEMM epilogue, which lets fused_moe skip the standalone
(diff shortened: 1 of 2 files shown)


coderfeli:
Why not use the tuning infra to get the config

Zzz9990:
Hi, could you please give us the tensor dump at the entrance of the [quant](https://github.com/ROCm/aiter/blob/main/aiter/fused_moe.py#L1800-L1808) ? The quant kernel is taking an unusually long time.

karverma-amd:
> Hi, could you please give us the tensor dump at the entrance of the [quant](https://github.com/ROCm/aiter/blob/main/aiter/fused_moe.py#L1800-L1808) ? The quant kernel is taking an unusually long time.

hi @Zzz9990 ,
Please find below dump logs - Dump at the entrance of the standalone inter-stage fused_dynamic_mxfp4_quant_moe_sort (the kernel this PR fuses away). Input a2 is [token_num*topk=262144, inter_dim=2048] bf16 ≈ 1.0 GB, padded to MoRI dispatch capacity, while only num_valid_ids (~64–832) rows are real in decode — so the quant scans an ~all-padding buffer

[dump.zip](https://github.com/user-attachments/files/28972008/dump.zip)

karverma-amd:
> Why not use the tuning infra to get the config
hi @coderfeli , Please see below response - 

1. FlyDSL emits the 2-stage MoE as fused GEMM kernels with programmable epilogues. On the DeepSeek-R1 MoE path, FlyDSL generates the stage1 (gate/up) and stage2 (down) expert GEMMs as templated kernels whose epilogue is part of the codegen. The default stage1 kernel emits bf16 activations, leaving the bf16→fp4 conversion to a separate dynamic_mxfp4 quant + moe_sort pass before stage2. The fused-quant variant (_fp4 sibling) folds that conversion into the stage1 GEMM epilogue — emitting fp4x2 activations plus sorted e8m0 scales directly — so the standalone quant/sort kernel launch is eliminated.
2. The _fp4 variants are a distinct, separately-tuned point in FlyDSL's kernel space, not a runtime flag on the same kernel. Each fused-quant kernel is its own FlyDSL codegen instance (different epilogue → different register/LDS allocation and scheduling), tuned independently. That's why membership in the merged tuned CSV matters: a flydsl_moe1*_fp4 row means that specific codegen has been benchmarked and validated, so it's safe to JIT on demand. Also note the _bnt0 (b-non-temporal) suffix is dropped before lookup, because the fused variants aren't tuned with that load flag — they're a different tuning subset.
3. The epilogue fusion is an M-sensitive (token-count) tradeoff, which is intrinsic to FlyDSL's launch model. The fused epilogue adds a fixed per-launch cost (quant + sort + scale swizzle inside the kernel). For large-M decode that cost is amortized across many rows and the saved kernel launch is a net win; at small M it's a larger fraction of a tiny GEMM and can regress. Since FlyDSL tunes/keys per GEMM shape rather than per runtime token tier, this large-M-vs-small-M decision lives naturally at dispatch time — hence the AITER_FUSE_STAGE1_MIN_TOKENS gate rather than a pure tuner pick.

PR #3702, opened 2026-06-12
perf(quant): coalesced LDS-staged store path for mxfp4_moe_sort_kernel

rbrugaro-amd:
# Coalesced LDS-staged store path for `mxfp4_moe_sort_kernel`

## Summary

`mxfp4_moe_sort_kernel` permutes MXFP4/MXFP8 e8m0 block-scale **bytes** from
per-token order into the sorted + swizzled layout consumed by the MX GEMM. The
loads are already vectorized, but the **store** is a per-byte scatter into the
swizzled output, which is fully uncoalesced and store-transaction-bound.

This change adds a **coalesced, LDS-staged store path**: the block's 32 rows ×
`scaleN_pad` scale bytes are staged into shared memory in natural `[row][col]`
order, then the threadblock emits the swizzled `out_scale` layout with
**coalesced packed 4-byte (dword) stores** instead of 32 strided single-byte
stores per thread.

The new path is **byte-exact** with the existing kernel (and the torch
reference) and is wrapped in a **host-side no-regression gate** so it engages
only where it is a net win on the target hardware (AMD Instinct MI355X /
gfx950), falling back to the original scatter path otherwise.

- No API change. Only `mxfp4_moe_sort_kernel` + its host launcher
  `mxfp4_moe_sort_hip` are touched.
- `num_rows ∈ {16, 64, 128}` shapes (model dims 256 / 512 / 16384) keep the
  original scatter path unconditionally (compile-time `if constexpr`).
- `persistent_mode` default is unchanged (`false`).

## Problem / motivation

The store loop in the original kernel is:

```cpp
vec_i vec_scale = load_vector_nbytes<...>(buffer_scale, scale_offset + scale_k);
for (int j = 0; j < vec_size_i; j++) {
    if ((scale_k + j) < scaleN_valid) {
        int addr = aiter::mx_scale_shuffle_idx(scaleN_pad, sorted_row, scale_k + j);
        out_scale[addr] = vec_scale[j];   // single-byte scatter
    }
}
```

The swizzle address (from `mx_scale_shuffle_idx`, `x = sorted_row`,
`y = scale column`) is:

```
addr = (x/32 * scaleN_pad)*32 + (y/8)*256 + (y%4)*64 + (x%16)*4 + (y%8)/4*2 + (x%32)/16
```

For consecutive columns `j` (i.e. consecutive `y`) the `(y%4)*64` term makes the
store address jump by ~64 bytes each iteration. So each thread issues ~32
strided 1-byte stores → fully uncoalesced, store-transaction-bound. (A
`persistent_mode` probe confirmed the kernel is **not** occupancy-bound; forcing
a persistent grid was neutral-to-worse, so the bottleneck is store coalescing,
not grid sizing.)

### Key structural fact that enables coalescing

`num_rows == 32` and the swizzle's row tile is also 32 (`x/32`, `x%32`, `x%16`
terms). So **one threadblock maps to exactly one 32-row swizzle tile.** Within a
fixed tile, the address decomposes so that each contiguous **256-byte** region of
`out_scale` equals *(one tile) × (8 columns)*, and one aligned dword at offset

```
tile_base + y_blk*256 + c*64 + xl_lo*4      (col_a = y_blk*8 + c, xl_lo = x%16)
```

packs exactly four source bytes:

```
byte0 = data[xl_lo   ][col_a]     byte1 = data[xl_lo+16][col_a]
byte2 = data[xl_lo   ][col_a+4]   byte3 = data[xl_lo+16][col_a+4]
```

So **64 consecutive dwords densely fill a 256-byte block**, and consecutive
threads writing consecutive dwords produce contiguous, coalesced store
transactions.

## The optimization

1. **LDS staging in natural order.** Stage this block's `32 × scaleN_pad` scale
   bytes into shared memory indexed `[row][col]`. The LDS is **zero-filled** for
   invalid rows and padding columns, so those positions store `0` — byte-exact
   with the torch reference (whose output buffer is zero-initialized) and with
   the original kernel's guards (`scale_k + j < scaleN_valid`, the
   `token_idx < num_tokens && (topk==1 || topk_id < topk)` validity test).

2. **Coalesced packed stores.** After a `__syncthreads`, threads cooperatively
   walk `ngroups*64` dwords (`ngroups = scaleN_pad/8`), each reconstructing one
   packed dword from the 4 staged bytes and writing it with a single aligned
   32-bit store. Consecutive threads → consecutive dwords → coalesced 256-byte
   transactions, replacing 32 strided 1-byte stores.

3. **Padding-tile skip.** Tiles whose 32 rows are entirely in the padding region
   (`block_idx*num_rows >= num_valid_ids`) are skipped via `continue`. The output
   buffer is pre-zeroed, so untouched padding tiles remain `0` (byte-exact), and
   this matches the original kernel's per-row validity guard. This avoids
   streaming zeros across the large `E × block_m` padding region (e.g. E=256
   always pads ~8192 rows). `block_idx` is uniform across the block, so there is
   no `__syncthreads` divergence hazard.

All semantics are preserved: `num_valid_ids`, the token/topk decode
(`token_id = sorted_ids[row] & 0xFFFFFF`, `topk_id = sorted_ids[row] >> 24`), the
validity guard, the `scale_k + j < scaleN_valid` column guard (padding columns
`[scaleN_valid, scaleN_pad)` are **not** written with data — they are staged as
`0`), the `scale_offset` computation for `topk==1` vs `topk>1`, and rows beyond
`num_valid_ids`.

## The gate (no-regression guarantee)

The LDS path is compile-time enabled only for `num_rows == 32`
(`if constexpr`). Within that, a **host-computed** runtime flag `use_coalesced`
selects the coalesced path vs. the original scatter path. The decision uses only
host-known quantities (note: `num_valid_ids` lives on device and cannot be read
on the host), computed in the dispatch macro `MXFP4_MOE_SORT_KERNEL_IMPL`:

```cpp
const int  scaleN_pad_host = (((cols + 31) / 32) + 7) / 8 * 8;     // LDS footprint / row
const long est_lds_work    = (long)scaleN_pad_host * (long)num_blocks;
const int32_t use_coalesced = (NUM_ROWS == 32 && est_lds_work <= 700000) ? 1 : 0;
```

### Rationale / empirical thresholds

The LDS staging + double `__syncthreads` cost scales with the **LDS footprint per
row (`scaleN_pad`)** times the **number of grid tiles (`num_blocks`)**. On
MI355X/gfx950 uncoalesced byte stores are comparatively cheap, so the coalesced
path is a net win for small/mid `scaleN_pad × tiles` but loses once that product
is large because the staging/sync overhead and reduced occupancy
(LDS = `32 × scaleN_pad` bytes, e.g. 7168 B at dim 7168) overtake the store
savings.

Sweeping `-m 256 1024 2048 4096 8192 16384` (both stages, E=256 topk=8), the
**only** point where an ungated coalesced kernel regressed vs. pristine was
**stage1 dim=7168 @ M=16384**. Mapping to the metric (num_blocks = grid tiles):

| case (dim, M)        | scaleN_pad | num_blocks | `est_lds_work` | ungated result |
|----------------------|-----------:|-----------:|---------------:|----------------|
| dim 7168, M=8192     | 224        | 2304       | 516,096        | win (use coalesced) |
| dim 7168, M=16384    | 224        | 4352       | 974,848        | **regress** (fall back) |
| dim 2048, M=16384    | 64         | 4352       | 278,528        | win (use coalesced) |

The threshold **700,000** sits between the largest winning product (516k) and the
regressing product (975k), with ~35% margin on both sides, and comfortably keeps
all `dim ≤ 4096` and all stage2 shapes on the coalesced path. Result: the gated
kernel keeps every measured win and never regresses.

## Correctness evidence

Byte-exact vs the torch reference (`run_torch` in
`op_tests/test_moe_sorting_mxfp4.py`): the `hip_err` column is **0** for every
tested shape. Tested matrix (24 shapes):

- `dim1 = 7168`, `dim2 ∈ {2048, 256}`
- `ek ∈ {256,8 ; 32,5}` (E=256 topk=8 production case, plus E=32 topk=5 to
  exercise small-expert guards)
- `M ∈ {256, 1024, 8192, 16384}`

This exercises the coalesced `num_rows==32` path (dims 7168 / 2048), the
`num_rows==128` scatter path (dim 256), the non-power-of-8 `scaleN` padding
guard, and both `topk==1`-style and `topk>1` `scale_offset` branches. All 24/24
rows report `hip_err == 0`. The gate does not affect correctness — both paths
produce identical bytes; it only selects the faster one.

## Performance (MI355X / gfx950)

Pristine vs. gated, standalone `mxfp4_moe_sort_hip` `hip_us`, E=256 topk=8,
`HIP_VISIBLE_DEVICES=0`. All measurements are **paired** (pristine and gated
measured back-to-back in the same session), and we report the harness's
unchanged **Triton reference kernel** (`triton_us`) as a **contention witness**:
because that kernel is identical across both sweeps, a matched `triton_us`
confirms both saw the same node load, making the comparison apples-to-apples.

The size of the win depends on how **store-bound** the run is, so we present two
first-class regimes, each measured as its own paired sweep:

- **Uncontended (store cheap):** node quiet, `triton_us` ≈ 14 µs at stage1
  M=16384. The pristine kernel's uncoalesced byte stores are already cheap here,
  so gains are modest.
- **Store-bound / contended:** node under heavy external load, `triton_us`
  ≈ 48 µs at stage1 M=16384 (~3.5× the uncontended time). Stores dominate, so
  coalescing helps the most.

In **both** regimes the gate applies the coalesced path only where it wins and
falls back to the byte-identical scatter path at the one large-`scaleN_pad ×
tiles` point (stage1 dim7168 M=16384), so there is **no regression** anywhere.

#### stage1 — dim 7168

| M     | pristine `hip_us` | gated `hip_us` | Δ           | gate path  |
|-------|------------------:|---------------:|------------:|------------|
| 256   | 6.53              | 5.85           | −10.5%      | coalesced  |
| 1024  | 7.70              | 6.93           | −10.0%      | coalesced  |
| 2048  | 6.82              | 6.14           | −10.0%      | coalesced  |
| 4096  | 7.00              | 6.34           | −9.4%       | coalesced  |
| 8192  | 7.73              | 7.61           | −1.6%       | coalesced  |
| 16384 | 9.33              | 9.41           | +0.8% (tie) | scatter (fallback) |

#### stage2 — dim 2048

| M     | pristine `hip_us` | gated `hip_us` | Δ        | gate path |
|-------|------------------:|---------------:|---------:|-----------|
| 256   | 4.18              | 3.78           | −9.5%    | coalesced |
| 1024  | 4.29              | 4.12           | −4.1%    | coalesced |
| 2048  | 4.23              | 3.91           | −7.5%    | coalesced |
| 4096  | 4.46              | 4.16           | −6.7%    | coalesced |
| 8192  | 4.78              | 4.63           | −3.2%    | coalesced |
| 16384 | 5.30              | 5.19           | −2.1%    | coalesced |

Uncontended gains are **~2–10%**, concentrated in small/mid-M and all of stage2.
The single gate-off point (stage1 dim7168 M=16384) runs the byte-identical
scatter path and ties pristine (+0.8%, within run-to-run noise: pristine 9.31–9.35,
gated 9.37–9.45). Crucially, this is the point that **regressed ~+19% when the
coalesced path was applied ungated** — the gate now prevents that.

#### stage1 — dim 7168

| M     | pristine `hip_us` | gated `hip_us` | Δ           | gate path  |
|-------|------------------:|---------------:|------------:|------------|
| 256   | 6.54              | 5.84           | −10.6%      | coalesced  |
| 1024  | 7.71              | 6.95           | −9.9%       | coalesced  |
| 2048  | 9.23              | 8.60           | −6.8%       | coalesced  |
| 4096  | 13.71             | 11.44          | −16.5%      | coalesced  |
| 8192  | 21.40             | 17.36          | −18.9%      | coalesced  |
| 16384 | 34.75             | 35.20          | +1.3% (tie) | scatter (fallback) |

#### stage2 — dim 2048

| M     | pristine `hip_us` | gated `hip_us` | Δ        | gate path |
|-------|------------------:|---------------:|---------:|-----------|
| 256   | 4.15              | 3.78           | −9.0%    | coalesced |
| 1024  | 4.27              | 4.09           | −4.1%    | coalesced |
| 2048  | 4.89              | 4.58           | −6.4%    | coalesced |
| 4096  | 6.37              | 5.67           | −11.0%   | coalesced |
| 8192  | 9.38              | 7.35           | −21.7%   | coalesced |
| 16384 | 14.33             | 10.82          | −24.5%   | coalesced |

Store-bound gains are larger, **up to −24%**. Again the only gate-off point
(stage1 dim7168 M=16384) runs the byte-identical scatter path and ties pristine
(+1.3%, within the ~3% run-to-run noise at large M).

**Net: no point regresses in either regime.** Every gate-on point is faster
(more so when stores dominate); the single gate-off point matches pristine.

### Methodology / honesty notes

- **Hardware/methodology.** AMD Instinct MI355X (gfx950), aiter
  `op_tests/test_moe_sorting_mxfp4.py` harness, `run_perftest` (101 iters per
  point). Each rebuild is `AITER_REBUILD=1` (JIT `module_quant`). Numbers above
  are the mean of 2 sweeps per regime; both regimes were captured on the same
  node within the same session, each as a fresh pristine-vs-gated pair with a
  matched `triton_us` witness.
- **Hardware-uniform.** The same pristine sweep run on 4 distinct physical GPUs
  (devices 0, 2, 3, 4 — all MI355X/gfx950) produced the same numbers within
  ~1–10%, ruling out per-card variance. The uncontended vs. contended difference
  is a node-load effect (witnessed by `triton_us`), not a hardware difference.

## Risk / scope

- Touches only `mxfp4_moe_sort_kernel` and its host launcher
  `mxfp4_moe_sort_hip` (one added kernel arg `use_coalesced`; one added host-side
  gate computation in `MXFP4_MOE_SORT_KERNEL_IMPL`).
- `num_rows ∈ {16, 64, 128}` shapes are unaffected (compile-time
  `if constexpr (num_rows == 32)`); they always use the original scatter path.
- Does not touch the `fused_dynamic_mx_quant_moe_sort_hip` path or any other
  kernel.
- No public API / signature change to `mxfp4_moe_sort_hip`.
- Both code paths are byte-exact, so the gate threshold is a pure
  performance tuning knob (a wrong threshold can only cost perf, never
  correctness).


junhaha666:
Hello, could you provide more comprehensive performance comparison data? You can use the command `python3 op_tests/test_moe_sorting_mxfp4.py -ek 256,8`

rbrugaro-amd:
Thanks @junhaha666  for the review! Here is the comprehensive before/after from the full
default sweep you suggested, across all four dims.

**Setup**
- HW: AMD Instinct **MI355X (gfx950)**, single GPU (`HIP_VISIBLE_DEVICES=0`).
- Command (exactly as suggested): `python3 op_tests/test_moe_sorting_mxfp4.py -ek 256,8`
  → stage1 dims {4096, 7168}, stage2 dims {256, 2048}, default M list
  {1, 64, 128, 256, 1024, 2050, 4200, 10000, 163840}, E=256, topk=8.
- Numbers are the standalone `mxfp4_moe_sort_hip` `hip_us` from the
  `moe_sorting_mxfp4_stage{1,2}` tables. **pristine** = current upstream kernel;
  **gated** = this PR.
- Both sweeps were run back-to-back in the same session. The harness's unchanged
  **Triton** kernel (`triton_us`) is the **contention witness**: it matched within
  ~1–4% between the two runs at every shape, so this is apples-to-apples. It was a
  **lightly-loaded / "store-cheap"** regime (e.g. stage1 dim7168 M=1024
  `triton_us` ≈ 8.4 µs), i.e. the *conservative* case for this optimization (gains
  are larger when the node is store-bound).
- **Correctness: `hip_err == 0` on every row of both sweeps** (all 4 dims × 9 M),
  plus the 24-shape byte-exact gate (dim1 7168, dim2 {2048,256}, ek {256,8 ; 32,5},
  M {256,1024,8192,16384}) and an explicit dim=4096 byte-exact check.

The coalesced path engages only for `num_rows == 32` (stage1 7168/4096, stage2
2048) when the in-kernel gate `scaleN_pad * num_blocks <= 700000` holds, else the
original scatter path runs. stage2 dim=256 uses `num_rows == 128` → always the
original scatter path (optimization N/A → expected ties).

> **Note (update):** the earlier broad sweep surfaced an LDS **bank-conflict** at
> `scaleN_pad % 32 == 0` (dim=4096 → `scaleN_pad=128` = exactly 32 LDS dwords, so
> all 32 rows of a tile aliased onto the same bank), which made dim=4096 mid-M
> *regress*. Fixed by **padding the LDS row stride** (`LDS_STRIDE = lds_cols + 4`)
> while keeping the `out_scale` addressing on `scaleN_pad` (output stays
> byte-identical). dim=4096 is now a large win, and the padding also removed
> residual partial conflicts on dims 7168 / 2048 (their wins grew too).

### stage1 — dim 4096 (scaleN_pad = 128) — *was regressing, now fixed*

| M      | pristine `hip_us` | gated `hip_us` | Δ        | path        |
|--------|------------------:|---------------:|---------:|-------------|
| 1      | 5.14              | 3.93           | −23.5%   | coalesced   |
| 64     | 5.14              | 3.96           | −23.0%   | coalesced   |
| 128    | 5.26              | 4.03           | −23.4%   | coalesced   |
| 256    | 5.33              | 4.07           | −23.6%   | coalesced   |
| 1024   | 5.51              | 4.23           | −23.3%   | coalesced   |
| 2050   | 6.46              | 4.67           | −27.7%   | coalesced   |
| 4200   | 8.83              | 5.56           | −37.0%   | coalesced   |
| 10000  | 16.85             | 8.94           | −46.9%   | coalesced   |
| 163840 | 165.3             | 179.1          | +8.3%    | scatter (fallback) |

### stage1 — dim 7168 (scaleN_pad = 224)

| M      | pristine `hip_us` | gated `hip_us` | Δ        | path        |
|--------|------------------:|---------------:|---------:|-------------|
| 1      | 6.24              | 5.31           | −15.0%   | coalesced   |
| 64     | 6.37              | 5.49           | −13.8%   | coalesced   |
| 128    | 6.62              | 5.56           | −16.0%   | coalesced   |
| 256    | 6.76              | 5.65           | −16.5%   | coalesced   |
| 1024   | 7.68              | 6.39           | −16.8%   | coalesced   |
| 2050   | 9.24              | 7.40           | −19.9%   | coalesced   |
| 4200   | 13.68             | 9.13           | −33.2%   | coalesced   |
| 10000  | 25.17             | 14.84          | −41.0%   | coalesced   |
| 163840 | 288.2             | 294.0          | +2.0%    | scatter (fallback) |

### stage2 — dim 256 (num_rows = 128 → optimization N/A)

| M      | pristine `hip_us` | gated `hip_us` | Δ        | path    |
|--------|------------------:|---------------:|---------:|---------|
| 1      | 2.44              | 2.35           | −3.6%    | scatter |
| 64     | 3.51              | 3.51           | ~0%      | scatter |
| 128    | 3.57              | 3.55           | −0.5%    | scatter |
| 256    | 3.60              | 3.58           | −0.5%    | scatter |
| 1024   | 3.61              | 3.58           | −0.8%    | scatter |
| 2050   | 3.58              | 3.56           | −0.6%    | scatter |
| 4200   | 3.76              | 3.75           | −0.3%    | scatter |
| 10000  | 4.11              | 4.11           | ~0%      | scatter |
| 163840 | 23.22             | 23.01          | −0.9%    | scatter |

All within noise, as expected — `num_rows == 128` here, so this dim always uses
the original scatter path (the coalesced path is compile-time disabled).

### stage2 — dim 2048 (scaleN_pad = 64)

| M      | pristine `hip_us` | gated `hip_us` | Δ        | path        |
|--------|------------------:|---------------:|---------:|-------------|
| 1      | 4.08              | 3.59           | −11.9%   | coalesced   |
| 64     | 4.11              | 3.61           | −12.1%   | coalesced   |
| 128    | 4.10              | 3.61           | −12.0%   | coalesced   |
| 256    | 4.21              | 3.62           | −14.0%   | coalesced   |
| 1024   | 4.44              | 3.84           | −13.4%   | coalesced   |
| 2050   | 4.82              | 4.02           | −16.5%   | coalesced   |
| 4200   | 6.39              | 4.77           | −25.4%   | coalesced   |
| 10000  | 10.85             | 7.25           | −33.2%   | coalesced   |
| 163840 | 105.5             | 105.6          | +0.1%    | scatter (fallback) |

### Summary

- **All coalesced-eligible shapes win** across the realistic M range (1–10000):
  dim 4096 −23…−47%, dim 7168 −14…−41%, dim 2048 −12…−33%. (Larger M ⇒ larger win
  in this regime.)
- **dim 4096 is fixed**: the bank-conflict regression (previously +3…+11% at
  M=1024–10000) is now a −23…−47% win after the LDS row-stride padding.
- **stage2 dim 256** (`num_rows=128`): optimization not applicable → ties.
- The only deltas above pristine are at the extreme **M=163840**, where the gate
  correctly falls back to the **scatter** path (so the device store loop is
  identical to upstream). gated is +8.3% (dim4096) / +2.0% (dim7168) / +0.1%
  (dim2048) there — a small, reproducible occupancy overhead from the gated
  kernel carrying the (unexecuted) coalesced branch, not from the store path
  itself. This is at ~1.3M sorted rows (163840 tokens × topk 8); for the realistic
  M range the optimization is a clean win with no regression.
- As noted in the PR, win magnitude scales with how store-bound the run is. This
  sweep is a light-load regime (conservative); under store-bound load the same
  shapes show even larger gains.

amd-ruitang3 (approved):

PR #3706, opened 2026-06-13
[fix](pa): add prebuild for pa_ps

PerryZhang01:
## Motivation

pa_ps (cpp_itfs paged-attention reduce kernel) is JIT-compiled with hipcc on the first inference and cached in AITER_ROOT_DIR/build (default $HOME/.aiter/build), which is lost on every container rebuild. So the first runs after a fresh build stall while the kernel compiles on the request path — on gpt-oss-120b this drops cold-start throughput to ~24k tok/s vs the steady-state ~31k.

The root cause is a prebuild coverage gap: setup.py's PREBUILD_KERNELS only AOT-compiles the module_* kernels (which ship in the wheel via get_user_jit_dir()). The entire cpp_itfs family (pa_ps, pa, mla, moe, sampling, …) is never prebuilt and is resolved only from AITER_ROOT_DIR/build, with no in-package fallback — so it always compiles cold on first use.

This change closes the gap for pa_ps: cpp_itfs's build dir now falls back to the packaged module-JIT dir when AITER_ROOT_DIR is unset (artifacts ship in the wheel, found at runtime with zero config), and pa_ps is compiled AOT during the existing prebuild step (hipcc only, no GPU). The AITER_ROOT_DIR override is preserved.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,10 @@ graft aiter_meta
 # exclude cache and compiled files .pyc / .pyo / .pyd / .pyd 
 global-exclude *.py[cod]
 prune aiter/jit/build
+# Re-include the AOT-prebuilt cpp_itfs pa_ps reduce kernels (~175K total). These are
+# compiled by setup.py's prebuild_pa_ps() into aiter/jit/build/pa_ps_<hash>/lib.so,
+# which is exactly where the cpp_itfs runtime (BUILD_DIR) looks for them. Without this
+# they get pruned out of the wheel and pa_ps falls back to slow first-inference hipcc
+# JIT (gpt-oss-120b cold-start ~24k vs ~31k tok/s). module_* kernels avoid the prune
+# because they install to aiter/jit/ root, not build/.
+recursive-include aiter/jit/build pa_ps_*/lib.so
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,10 +4,10 @@ graft aiter_meta
 # exclude cache and compiled files .pyc / .pyo / .pyd / .pyd 
 global-exclude *.py[cod]
 prune aiter/jit/build
-# Re-include the AOT-prebuilt cpp_itfs pa_ps reduce kernels (~175K total). These are
-# compiled by setup.py's prebuild_pa_ps() into aiter/jit/build/pa_ps_<hash>/lib.so,
-# which is exactly where the cpp_itfs runtime (BUILD_DIR) looks for them. Without this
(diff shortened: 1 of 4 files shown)


Bernard-Liu:
1. Please make it explicit that this only prebuilds the C++ pa_ps reduce kernel, not the gluon attention kernel.

The PR title/description ("add prebuild for pa") and some comments read as if paged-attention decode is being prebuilt, which is easy to misread. Could we tweak the comments/docstring (and ideally the PR title) to say "pa_ps reduce kernel" explicitly so nobody assumes the whole decode kernel is covered?

2. Runtime build-dir relocation is unconditional and affects all cpp_itfs ops (most important).

The BUILD_DIR change in csrc/cpp_itfs/utils.py takes effect at import time — it's not gated by PREBUILD_KERNELS / AITER_TRITON_ONLY / whether pa_ps is used, and it applies to every cpp_itfs template op (pa, pa_ps, mla, moe, sampling, …), not just pa_ps. When AITER_ROOT_DIR is unset, the default cache dir moves from ~/.aiter/build to get_user_jit_dir()/build. Consequences:

Existing caches under ~/.aiter/build become orphaned, so the first run after upgrade triggers a one-time JIT recompile into the new location for any cpp_itfs op that gets used (not only pa_ps).
When AITER_ROOT_DIR is set, behavior is unchanged (<root>/build) — but then the prebuilt artifacts aren't found there, so the wheel-prebuild benefit is silently lost (counterintuitive; worth documenting).
New import-time side effect: importing csrc.cpp_itfs.utils now imports aiter.jit.core, which on load does os.makedirs(bd_dir) and may copytree the jit dir into ~/.aiter/jit if site-packages isn't writable — heavier than the previous plain string join. (No circular import, and there's a graceful fallback to ~/.aiter/build.)

3. Architecture-binding caveat (applies to all prebuilds).

compile_lib bakes the target arch into the .so via --offload-arch=<arch> (from GPU_ARCHS), but the folder name (md5 signature) does not include the arch, and at runtime not_built() only checks file existence, not arch match. So a prebuilt .so compiled for e.g. gfx942 will be loaded as-is on a gfx950 machine without recompiling — a potential arch mismatch. This is the same contract PREBUILD_KERNELS already imposes on module_*, and it's fine as long as the wheel's GPU_ARCHS covers the deployment arch (or is a multi-arch fat binary). The difference vs. pure JIT is that pure JIT uses GPU_ARCH=native and always matches the running GPU, whereas a shipped prebuilt .so is never re-validated against the actual arch — worth keeping in mind for cross-arch wheel deployments.

PR #3718, opened 2026-06-15
Yhl/gptoss pa asm shuf repro 20260611

yhl-amd:
## Submission Checklist


aiter/jit/core.py
HaonanWang98, line 842: no need to change now, env var control

PR #3732, opened 2026-06-15
HD256 FMHA FP8 GFX950

JohnNikolay84:
## Motivation

qwen3.5 is using triton attention which is slow, we have tried to implement an asm alternative to increase performance.

## Technical Details

Use hd128 reference attention kernel and implement hd256 support. kv tile is reduced to 64 to fit into registers req.

## Test Result

Light blue is a new fmha f8 hd256.

image
## Submission Checklist


JohnNikolay84:
Qwen3.5 long-context 60k uplift is 72% → 96% of B200

amd-oshkarav:
Guys, could you please rebase the branch on main?
My results on testing this branch w.r.t the native aiter branch within my baseline image sglang 0.5.14  are not impressive. From my side, I'm working to exclude the bugs in implementation of the kernel path in my framework and setting up larger chunk-prefill size, lower max conc  to keep kv empty. 
But if you can rebase the branch it can help to exclude other factors which can downgrade the performance.
benchmark_mi355_tp4_baseline_sgl_0 5 14_vs_mi355_tp4_aiter_fmha_fp8_hd256_sgl_0 5 14_and_mean_e2e_latency_ms_vs_total_throughput

UPD: with on/off approach within the same aiter branch, large chunks, 60k/600 it looks really good.  But I need to run more benchmarks and on the newer sglang. Keeping this branch up-to date with the main aiter is really important.


benchmark_mi350_tp4_hd256_on_65536_60k_vs_mi350_tp4_hd256_off_65536_60k_and_mean_e2e_latency_ms_vs_total_throughput

amd-oshkarav:
Sglang side: the PR should be merged. The kernel performs very well on long contexts with large prefill chunks. With the same set up but for smaller contexts its advantage is less impressive and the performance drops to large concurrencies like 256 due to visibility of the overhead. The possible recommendation is that use env var min length threshold which trigger this kernel when the sequence's length is above this threshold.
benchmark_tp4_baseline_cookbook_ua_vs_tp4_hd256_on_cookbook_65536_and_mean_e2e_latency_ms_vs_total_throughput

valarLip (approved):

PR #3739, opened 2026-06-15
configs: add DSv3-MXFP4 E=33/topk9 fused-MoE shape (shared-expert fus…

rbrugaro-amd:
# configs: add DSv3-MXFP4 E=33/topk9 fused-MoE shape (shared-expert fusion) — stage-1 + stage-2 tuned

## Problem

The `dsv3_fp4` fused-MoE configs cover the routed-expert and EP-sharded shapes
but have **no entry for `model_dim=7168, inter_dim=2048, expert=33, topk=9`**.

That shape is what DeepSeek-R1 MXFP4 produces with **shared-experts fusion ON**:
the always-on shared expert is folded in as the **33rd** grouped-GEMM expert and
`topk` goes `8 -> 9`, so *every* token routes through it. With no tuned row the
runtime fused-MoE dispatch (`get_2stage_cfgs`) MISSes, `cfg` resolves to `None`,
and the C++ auto-pick falls back to the generic `ck::kernel_moe_mxgemm_2lds` for
**both** stages.

The fast kernels already exist for the neighbouring shapes — this is a
**tuning/selection gap, not a missing kernel**.

## What this PR adds

`dsv3_fp4_tuned_fmoe.csv`: **+32 rows** for `(expert=33, topk=9)` — 16 token
buckets (1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384,
32768), each a **primary** best-`us` row plus a **`flydsl_fallback`** CK row,
matching the existing (32,8)/(257,9) dual-row convention. Buckets are the
contiguous powers of two the runtime's `nextPow2` token padding can request, so
there is no cache-miss gap. Same `FP4X2 × FP4X2 × BF16` + `QuantType.per_1x32`,
`ActivationType.Silu`, `use_g1u1=1`, `doweight_stage1=0` combo and 26-column
schema (ending in `_tag`) as the existing dsv3 rows.

Both stages are tuned. The fused flydsl `_fp4` stage-1 (quant+sort fused into the
gemm1 epilogue) was being **missed by the tuner** due to a fused-vs-non-fused
selection bug — fixed in **ROCm/aiter#3805**; it shows up here at the larger
token buckets. The per-bucket selections below are the **best-measured-E2E**
picks: some small-token buckets intentionally use the non-fused kernel where it
benchmarked better end-to-end, so the column is a mix of fused (`_fp4`) and
non-fused by design, not a uniform fused sweep.

## Per-bucket primary selections (`expert=33, topk=9, model_dim=7168, inter_dim=2048, cu_num=256`)

| token | total us | tflops | stage1 | stage2 |
|------:|---------:|-------:|--------|--------|
| 1     | 49.2   | 16.1   | flydsl t64x128x256_w4_**fp4** | ck 256x64x128x128_1x4 |
| 2     | 71.0   | 22.3   | flydsl t32x128x256_w4 | flydsl t32x128x256_atomic_bnt2 |
| 4     | 103.2  | 30.7   | flydsl t32x64x256_w3 | flydsl t32x128x256_atomic_bnt2 |
| 8     | 121.1  | 52.4   | flydsl t64x128x256_w3_**fp4** | flydsl t32x128x256_atomic_bnt2_sbm64 |
| 16    | 139.8  | 90.7   | flydsl t32x32x256_w3 | flydsl t32x128x256_atomic_bnt2 |
| 32    | 141.1  | 179.8  | flydsl t32x32x256_w3 | flydsl t32x128x256_atomic_bnt2 |
| 64    | 144.1  | 352.2  | flydsl t32x32x256_w3 | flydsl t32x128x256_atomic_bnt2 |
| 128   | 155.6  | 652.0  | flydsl t64x128x256_w4_**fp4** | flydsl t64x128x256_atomic_bnt2 |
| 256   | 187.7  | 1081.2 | flydsl t64x128x256_w4_bnt0_**fp4** | flydsl t64x128x256_atomic |
| 512   | 241.2  | 1682.6 | flydsl t64x128x256_w4_bnt0_**fp4** | flydsl t64x128x256_atomic_persist |
| 1024  | 371.9  | 2182.9 | flydsl t64x128x256_w2_bnt0_**fp4** | flydsl t64x256x256_atomic |
| 2048  | 645.5  | 2515.2 | flydsl t128x128x256_w4_bnt0_**fp4** | flydsl t128x128x128_atomic |
| 4096  | 1161.7 | 2795.0 | flydsl t64x128x256_w4_bnt0_**fp4** | flydsl t64x128x256_atomic_bnt2 |
| 8192  | 2178.2 | 2981.4 | flydsl t64x128x256_w4_bnt0_**fp4** | flydsl t64x128x256_atomic_persist_async_w4_cumul3 |
| 16384 | 4252.7 | 3054.1 | flydsl t64x128x256_w3_bnt0_**fp4** | flydsl t64x128x256_atomic_persist_async_w4_cumul3 |
| 32768 | 8616.1 | 3014.8 | flydsl t64x128x256_w4_bnt0_**fp4** | flydsl t64x128x256_atomic |

(Full kernel names are in the committed CSV. `_fp4` suffix = fused stage-1;
others are the non-fused pick chosen for best E2E.)

## End-to-end validation (standard sglang server benchmark)

Measured with the standard `sglang.bench_serving` client against a
`sglang.launch_server` instance configured like the production DeepSeek-R1 MXFP4
deployment — TP8, **attention DP=8, EP=8, MoRI-EP a2a, shared-experts fusion ON**
(this (33,9) shape) — on 8×MI355X.

| in / out / concurrency | before tok/s | after tok/s | throughput | before ITL | after ITL | decode latency |
|---|---:|---:|:--:|---:|---:|:--:|
| 1024 / 512 / 128 |  6044 |  6626 | **+9.6%** | 33.6 ms | 28.8 ms | **−14.5%** |
| 1024 / 512 / 256 |  9654 | 10542 | **+9.2%** | 40.3 ms | 33.2 ms | **−17.8%** |
| 4096 / 512 / 256 | 21064 | 22042 | **+4.6%** | 87.5 ms | 85.0 ms | **−2.8%** |

(`tok/s` = total token throughput; `ITL` = median inter-token latency, i.e. the
per-decode-step cost where the MoE GEMMs sit.)

<details><summary>Exact repro</summary>

Server (identical for before/after; only `dsv3_fp4_tuned_fmoe.csv` differs):
```bash
# env: SGLANG_AITER_MOE=1 AITER_FUSE_STAGE1_QUANT=1 SGLANG_USE_AITER=1 \
#      SGLANG_AITER_MLA_PERSIST=1 SGLANG_ROCM_FUSED_DECODE_MLA=1 SGLANG_AITER_AR=1 \
#      MORI_SHMEM_MODE=ISOLATION MORI_SHMEM_HEAP_SIZE=16G \
#      SGLANG_MORI_COMBINE_DTYPE=fp8 SGLANG_MORI_NO_PAD_MASK=1 (full list in repo)
python -m sglang.launch_server --model-path <DSR1-MXFP4> --trust-remote-code \
  --tp-size 8 --dp-size 8 --enable-dp-attention --enable-dp-lm-head \
  --ep-size 8 --moe-a2a-backend mori --deepep-mode normal --moe-dense-tp-size 1 \
  --enforce-shared-experts-fusion --attention-backend aiter \
  --kv-cache-dtype fp8_e4m3 --triton-attention-num-kv-splits 2 \
  --disable-radix-cache --disable-chunked-prefix-cache --mem-fraction-static 0.90

# client (per point)
python -m sglang.bench_serving --backend sglang --dataset-name random \
  --random-input-len {1024|4096} --random-output-len 512 \
  --num-prompts {512|1024} --max-concurrency {128|256}
```
8×MI355X (gfx950). Full server + client logs for each point are in the run dir.
</details>

### Trace showing kernel change
Before: shows stage 1 unfused followed by quant and sort
Screenshot 2026-06-23 205815

After: shows fused stage 1 followed by stage 2 moe
Screenshot 2026-06-23 205742

### Accuracy (no regression)

GSM8K, 200 questions, 5-shot, same server config:

```
python -m sglang.test.few_shot_gsm8k --num-questions 200 --num-shots 5 \
  --max-new-tokens 2048 --host http://127.0.0.1 --port 30000


100%|██████████| 200/200 [00:11<00:00, 16.74it/s]
Accuracy: 0.975
Invalid: 0.000
Latency: 12.075 s
Output throughput: 1525.341 token/s
```


rbrugaro-amd:
**The `topk` key for these rows is blocked on the `get_2stage_cfgs` strip from #3377 — same root cause as #3999**

These are DeepSeek-R1 MXFP4 (gfx950), EP + **shared-expert fusion**: per rank `expert = 32 routed + 1 shared = 33`, `topk_ids` width `9` (`8` routed + `1` shared). That extra column is a **real, always-selected expert**, not a masked pad.

#3377 (cc @yadaish) added an unconditional `topk -= int(is_ep)` in `get_2stage_cfgs`. #3999 (cc @mjkvaak-amd ) already documents how this mis-keys vLLM standard routing (`8`→`7`). My case is the same defect from the other side: the extra column *exists* but is real, so the lookup keys at `8` when the computed shape is `9`. Because the shared expert is the densest, always-hit expert, tuning at `topk=8` (uniform, shared column absent) isn't even the shape that runs — so a re-tune would optimize the wrong GEMM, not just mislabel it.

The patch attached to #3999 (convention-aware strip — only strip a column that's genuinely always-masked) is the right fix and resolves both reports. 

With that fix the shared column isn't stripped, the lookup keys at the real width `9`, and **these rows are correct as-authored at `(expert=33, topk=9)`** — matching what they were tuned on. I'll hold this PR at `(33,9)` pending #3999.

cc: @valarLip @junhaha666

yadaish:
> **The `topk` key for these rows is blocked on the `get_2stage_cfgs` strip from #3377 — same root cause as #3999**
> 
> These are DeepSeek-R1 MXFP4 (gfx950), EP + **shared-expert fusion**: per rank `expert = 32 routed + 1 shared = 33`, `topk_ids` width `9` (`8` routed + `1` shared). That extra column is a **real, always-selected expert**, not a masked pad.
> 
> #3377 (cc @yadaish) added an unconditional `topk -= int(is_ep)` in `get_2stage_cfgs`. #3999 (cc @mjkvaak-amd ) already documents how this mis-keys vLLM standard routing (`8`→`7`). My case is the same defect from the other side: the extra column _exists_ but is real, so the lookup keys at `8` when the computed shape is `9`. Because the shared expert is the densest, always-hit expert, tuning at `topk=8` (uniform, shared column absent) isn't even the shape that runs — so a re-tune would optimize the wrong GEMM, not just mislabel it.
> 
> The patch attached to #3999 (convention-aware strip — only strip a column that's genuinely always-masked) is the right fix and resolves both reports.
> 
> With that fix the shared column isn't stripped, the lookup keys at the real width `9`, and **these rows are correct as-authored at `(expert=33, topk=9)`** — matching what they were tuned on. I'll hold this PR at `(33,9)` pending #3999.
> 
> cc: @valarLip @junhaha666



> **The `topk` key for these rows is blocked on the `get_2stage_cfgs` strip from #3377 — same root cause as #3999**
> 
> These are DeepSeek-R1 MXFP4 (gfx950), EP + **shared-expert fusion**: per rank `expert = 32 routed + 1 shared = 33`, `topk_ids` width `9` (`8` routed + `1` shared). That extra column is a **real, always-selected expert**, not a masked pad.
> 
> #3377 (cc @yadaish) added an unconditional `topk -= int(is_ep)` in `get_2stage_cfgs`. #3999 (cc @mjkvaak-amd ) already documents how this mis-keys vLLM standard routing (`8`→`7`). My case is the same defect from the other side: the extra column _exists_ but is real, so the lookup keys at `8` when the computed shape is `9`. Because the shared expert is the densest, always-hit expert, tuning at `topk=8` (uniform, shared column absent) isn't even the shape that runs — so a re-tune would optimize the wrong GEMM, not just mislabel it.
> 
> The patch attached to #3999 (convention-aware strip — only strip a column that's genuinely always-masked) is the right fix and resolves both reports.
> 
> With that fix the shared column isn't stripped, the lookup keys at the real width `9`, and **these rows are correct as-authored at `(expert=33, topk=9)`** — matching what they were tuned on. I'll hold this PR at `(33,9)` pending #3999.
> 
> cc: @valarLip @junhaha666

@rbrugaro-amd   EP tuning will be redesigned in the near future. Could you use current convention to support your tuning? 
If yes, I think the improved tuning csv can be merged.

rbrugaro-amd:
@yadaish thanks for your review and comments, I re-keyed the CSV to 33,8 according to current upstream and also removed the flyDSL fallbacks I had to CK moe because looks like the new convention is not to have fallbacks now, is that correct?

yadaish:
LGTM

rbrugaro-amd:
@ROCm/team_aiter  can we merge?

valarLip (approved):

PR #3741, opened 2026-06-16
[fused_moe] Fix crash for Deepseek v4 moe   "RuntimeError: Unsupported scales/output dtype!"

dllehr-amd:
The 2-stage MoE selector picks a bf16 (A16W4) activation dtype at small M for per_1x32 mxfp4-weight experts, for any activation. But the activation-quantization step in fused_moe_2stages only kept the activation in bf16 when activation == Swiglu; other activations (e.g. Silu, used by DeepSeek-V4) fell through to the mxfp4 quantizer and produced an fp4 tensor. That fp4 tensor was then handed to a kernel selected for bf16 input, so the CK-Tile gemm1 dispatch raised "Unsupported scales/output dtype!".

Keep bf16/fp16 activations whenever the selector chose them, regardless of activation type, so the quantized tensor matches the selected kernel. Verified with DeepSeek-V4-Flash on gfx950 (TP=8): prefill (M>=256) runs A8W4 as before, decode (M<256) now runs A16W4 instead of crashing, and the model serves and decodes end to end.

## Submission Checklist


okorzh-amd:
@lalala-sh could you please help to review this PR?

ChuanLi1101:
Thanks for the fix — gating the bf16/fp16 activation keep on `activation == Swiglu or gate_mode == GateMode.INTERLEAVE` correctly mirrors the selector logic in `fused_moe.py:406`, so the quantized `a1` now matches the kernel that was chosen. CI is green and it looks good to merge once approved. Two non-blocking points worth a quick confirmation:

1. **Parallel `a2` (intermediate) quant block is not updated.** The twin condition for the stage-1 output around `fused_moe.py:1853-1862` still only checks `activation == ActivationType.Swiglu` (not `INTERLEAVE`):

   ```python
   and (
       q_dtype_a in [dtypes.bf16, dtypes.fp16]
       and activation == ActivationType.Swiglu
       or (metadata.ksplit > 1 and is_shuffled)
   )
   ```

   So for Silu + INTERLEAVE the `a2` going into gemm2 still gets mxfp4-quantized to fp4 while the gemm1 input stays bf16. This asymmetry (A16 only on gemm1) is consistent with your end-to-end DeepSeek-V4-Flash verification, so it's presumably intended — just want to confirm gemm2 expects fp4 `a2` here and there's no latent mismatch on the INTERLEAVE path.

2. **Test env-var leak in `test_moe_ep.py`.** The new `os.environ["AITER_BF16_FP8_MOE_BOUND"] = "0"` is set without saving/restoring, unlike `test_moe_2stage.py` (which saves/restores around lines 889-909). It will leak to subsequent cases in the same process. Minor test-hygiene nit — consider restoring the previous value in a `finally` to keep cases isolated.

ChuanLi1101 (dismissed):
Approving — the fix correctly aligns the a1 activation-keep condition with the kernel selector (fused_moe.py:406), CI is fully green, and it's verified end-to-end on DeepSeek-V4-Flash. The two points I raised earlier (the a2/intermediate quant asymmetry and the test env-var leak) are non-blocking and can be addressed in a follow-up if desired.

aiter/fused_moe.py
```diff
@@ -1786,7 +1786,9 @@ def fused_moe_2stages(
         and w1.dtype == dtypes.fp4x2
         and (
             q_dtype_a in [dtypes.bf16, dtypes.fp16]
-            and activation == ActivationType.Swiglu
+            and (
+                activation == ActivationType.Swiglu or gate_mode == GateMode.INTERLEAVE
```
coderfeli: should be gate_mode == GateMode.INTERLEAVE

PR #3742, opened 2026-06-16
[triton] Add sliding-window support to flash-attention backward

bangtianliu:
**Motivation**:  Sliding-window (local) attention was supported in the AMD Triton flash-attention forward pass but not in the backward pass

This PR plumbs the window flags through the dKdV/dQ inner kernels and both fused kernels, with the backward mask following the forward convention (the backward pass recomputes softmax from the stored LSE, so the masks must match exactly). It also adds _sliding_window_q_bounds/_sliding_window_k_bounds to skip out-of-window blocks, while per-element masking remains the correctness guarantee. Windowed backward is restricted to fused mode, and an unbounded right edge is rejected.

## Performance

Benchmarked the windowed backward against full attention on the META-2.3B-style varlen GQA-SWA configs (`batch=4`, `seqlen_q=4096`, `num_heads=48`, `num_kv_heads=6`, `head_dim=64`, causal, **window = 2047**, bf16), Triton AMD backend. Before (stock) and after (this PR) were measured back-to-back in the same session on the same GPU. Wall-clock backward (ms), deterministic=FALSE:

| seqlen_kv | windowed bwd **before** | windowed bwd **after** | speedup | full (-1) bwd (after) | windowed vs full |
|----------:|------------------------:|-----------------------:|:-------:|----------------------:|:----------------:|
| 4096  | 5.11  | 6.29 | 0.81× | 5.03  | 1.25× (slower) |
| 8192  | 12.47 | 7.70 | 1.62× | 12.49 | 1.62× faster |
| 12288 | 20.05 | 7.98 | 2.51× | 20.10 | 2.52× faster |
| 16384 | 27.98 | 8.11 | 3.45× | 28.02 | 3.45× faster |


Assisted-by:  [Claude Code](https://claude.ai/code)

### DIFF
--- a/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
+++ b/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
@@ -2959,6 +2959,10 @@ def _bwd_dkdv_inner(
                 causal_mask = (offs_m[None, :] - delta_qk) >= offs_n[:, None]
                 mask = causal_mask & mask
             if USE_SLIDING_WINDOW:
+                # Per-element form of the band a query m attends:
+                #   m + (seqlen_k - seqlen_q) - L <= n <= m + (seqlen_k - seqlen_q) + R.
+                # _sliding_window_q_bounds() is the block-range inversion of this same
+                # inequality (it bounds m for a fixed K block); keep the two in sync.
                 causal_offset = seqlen_k - seqlen_q
                 if WINDOW_SIZE_LEFT < 0:
                     window_mask = offs_n[:, None] <= (
@@ -3161,6 +3165,10 @@ def _bwd_dq_inner(
                 causal_mask = (offs_m[:, None] - delta_qk) >= offs_n[None, :]
                 mask = causal_mask & mask
             if USE_SLIDING_WINDOW:
+                # Per-element form of the band a query m attends:
+                #   m + (seqlen_k - seqlen_q) - L <= n <= m + (seqlen_k - seqlen_q) + R.
+                # _sliding_window_k_bounds() is the block-range inversion of this same
+                # inequality (it bounds n for a fixed Q block); keep the two in sync.
                 causal_offset = seqlen_k - seqlen_q
                 if WINDOW_SIZE_LEFT < 0:
                     window_mask = offs_n[None, :] <= (
--- a/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
+++ b/aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
@@ -489,6 +489,19 @@ def get_bwd_configs(mode: AutotuneMode):
                     num_stages=2,
                     num_warps=8,
                 ),
+                # mid-tile, 2-stage variant
+                triton.Config(
+                    {
+                        "BLOCK_M1": 32,
+                        "BLOCK_N1": 128,
+                        "BLOCK_M2": 128,
+                        "BLOCK_N2": 64,
+                        "BLK_SLICE_FACTOR": 2,
+                        "waves_per_eu": 2,
+                    },
+                    num_stages=2,
+                    num_warps=4,
+                ),
             ]
             causal_configs = [
                 triton.Config(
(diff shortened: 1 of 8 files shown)


aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
```diff
@@ -3922,9 +4110,21 @@ def bwd_kernel_fused_noncausal(
             else:
                 descale_q, descale_k, descale_v = 1.0, 1.0, 1.0
 
-            # because there is no causal, we always start from the beginning
-            start_m = 0
-            num_steps = tl.cdiv(seqlen_q, BLOCK_M1)
+            # because there is no causal, we sweep all query blocks -- unless a
+            # sliding window lets us skip query blocks entirely outside the band.
+            if USE_SLIDING_WINDOW:
+                start_m, num_steps = _sliding_window_q_bounds(
```
scxiao, line 4124: Just wondering, should the sliding window bound in `_bwd_dkdv_inner` uses the same logic as that in `_sliding_window_q_bounds`?

op_tests/triton_tests/attention/test_mha_v3.py
```diff
@@ -1023,6 +1023,169 @@ def test_mha_backward_varlen_fp8(
...
+        triton_dq, triton_dk, triton_dv = torch.autograd.grad(triton_out, (q, k, v), do)
+
+    torch_dq, torch_dk, torch_dv = torch.autograd.grad(
+        torch_out, (q_ref, k_ref, v_ref), do
+    )
+
+    torch.testing.assert_close(
+        triton_dq,
+        torch_dq,
+        atol=1e-2,
```
scxiao: For the `1e-2` tolerance here, can we add some test using fp32 data type, so smaller tolerance can be used here?

aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/bwd.py
bangtianliu: Yes, same logic, just expressed at two levels of granularity, so I've added cross-reference comments to make that explicit.

 - `_bwd_dkdv_inner` applies it per element (the window_mask).
 - `_sliding_window_q_bounds` is the block-range inversion of that exact inequality; it solves for the m range given a fixed K block's n range, returning (start_m, num_steps) so we can skip query blocks entirely outside the band.

scxiao (dismissed):

scxiao:
LGTM.  Thanks for the changes @bangtianliu.

Hi @micmelesse, could you please help take a look? Thanks

op_tests/triton_tests/attention/test_mha_dao_ai.py
```diff
@@ -48,13 +48,25 @@ def dao_ai_impl():
...
-        (1, 128, 128, 8, 8, 64, True, True, False),  # fwd_varlen causal
-        (1, 128, 128, 8, 8, 64, True, False, True),  # bwd causal
-        (1, 128, 128, 8, 8, 64, True, True, True),  # bwd_varlen causal
+        (1, 128, 128, 8, 8, 64, True, False, False, (-1, -1)),  # fwd causal
+        (2, 256, 256, 16, 4, 128, False, False, False, (-1, -1)),  # fwd GQA non-causal
+        (1, 128, 128, 8, 8, 64, True, True, False, (-1, -1)),  # fwd_varlen causal
+        (1, 128, 128, 8, 8, 64, True, False, True, (-1, -1)),  # bwd causal
+        (1, 128, 128, 8, 8, 64, True, True, True, (-1, -1)),  # bwd_varlen causal
+        (1, 128, 128, 8, 8, 64, True, False, True, (32, 0)),  # causal window
+        (1, 128, 128, 8, 8, 64, True, True, True, (32, 0)),  # causal window varlen
```
micmelesse, line 59: The test cases just use small sequence length. I think we need to cover production size sequence lengths.

op_tests/triton_tests/attention/test_mha_v3.py
```diff
@@ -1023,6 +1023,178 @@ def test_mha_backward_varlen_fp8(
...
+@pytest.mark.parametrize("VARLEN", [False, True])
+@pytest.mark.parametrize(
+    "SEQLEN_Q, SEQLEN_K, NUM_Q_HEADS, NUM_K_HEADS",
+    [
+        (128, 128, 8, 8),  # baseline: equal seqlens, MHA
+        (128, 256, 8, 8),  # seqlen_q != seqlen_k exercises causal_offset / delta_qk
+        (128, 128, 16, 4),  # GQA combined with sliding window
+    ],
+)
+@pytest.mark.parametrize(
```
micmelesse, line 1104: Let us cover bigger sequence lengths.

micmelesse (changes requested):
I would like to test much bigger sequence lengths. Upstream tests max out at sequence length of 2048. The test cases with larger sequence lengths should have bigger window, something in the 128 to 256 range. Let us also try to have windows that are not block aligned. For example, an odd sized window would be great.

scxiao (approved):

bangtianliu:
cc @micmelesse for another look

micmelesse (approved):
LGTM

azaidy (approved):
LGTM!

micmelesse:
I am working on a follow up pr here to enable upstream sliding window tests. See https://github.com/ROCm/aiter/pull/4003

PR #3766, opened 2026-06-17
Fix batched_gemm_a16wfp4 split-K garbage output / OOB for small M

### LINKED ISSUE #3737: [Issue]: batched_gemm_a16wfp4 split-K path reads out of bounds / returns garbage for small M on gfx950
### Problem Description

(Content summarized by AI, but verified)

## Summary

We found a correctness and stability issue in AITER's `batched_gemm_a16wfp4` kernel on ROCm/gfx950. The issue appears in the split-K path selected for small `M`, which is common during LLM decode. The kernel can return garbage values or hit a HIP illegal memory access.

This surfaced while running GLM-5.1 in vLLM, but a standalone test of the AITER kernel reproduces the issue without the full model. Disabling vLLM's use of this fp4 batched-GEMM path avoids the problem.

## Environment

- GPU arch: gfx950
- Container: `rocm/vllm-dev:nightly`
- AITER version: `0.1.13.post1`
- vLLM image version used during initial repro: `0.22.1rc1.dev511+gc621af169.rocm723`

## Problem Shape

The GLM-5.1 MLA decode up-projection uses shapes equivalent to:

- `x`: `[8, M, 512]`, bf16
- `w`: `[8, 256, 256]`, uint8
- `w_scales`: `[8, 256, 16]`, uint8
- `transpose_bm=True`
- `prequant=True`

For small `M`, `_get_config(M, N=256, K=512)` selects `NUM_KSPLIT=4`.

## Observed Behavior

Standalone kernel test results against a bf16 reference:

- `M = 1, 2, 4, 8, 16, 32, 64, 128`: output is garbage, cosine similarity near 0 or NaN
- `M = 256, 512`: output is correct, cosine similarity around 0.987
- In the full GLM-5.1 decode path, this sometimes becomes `HIP error 700: illegal memory access`

Forcing `NUM_KSPLIT=1` makes the small-M cases correct.

## Suspected Root Cause

For `M=1, N=256, K=512`, AITER selects:

- `NUM_KSPLIT=4`
- `BLOCK_SIZE_K=256`
- `SPLITK_BLOCK_SIZE=256`

The effective A ranges are:

- `pid_k=0`: `[0, 256)` valid
- `pid_k=1`: `[256, 512)` valid
- `pid_k=2`: `[512, 768)` out of bounds
- `pid_k=3`: `[768, 1024)` out of bounds

Only 2 splits are needed for `K=512`, but 4 are launched. The guard appears to use packed half-units:

### Operating System

22.04.5 LTS (Jammy Jellyfish)

### CPU

AMD EPYC 9575F 64-Core Processor

### GPU

8 * MI355

### ROCm Version

rocm 7.2.3

### ROCm Component

_No response_

### Steps to Reproduce

```
"""Standalone test of AITER batched_gemm_a16wfp4 with GLM-5.1 MLA _v_up_proj
shapes, mirroring vllm/model_executor/layers/attention/mla_attention.py.

No model load: directly exercises the AITER fp4 batched-GEMM kernel that
faulted (illegal memory access in _batched_gemm_a16wfp4_reduce_kernel) to
determine whether the kernel itself is buggy independent of vLLM.

Run on a single GPU:
  AMD_SERIALIZE_KERNEL=3 HIP_LAUNCH_BLOCKING=1 VLLM_ROCM_USE_AITER=1 \
  HIP_VISIBLE_DEVICES=0 python3 /root/workspace/aiter_fp4bmm_test.py
"""

import os
import sys
import traceback

import torch

from vllm._aiter_ops import rocm_aiter_ops
from vllm.model_executor.layers.quantization.quark.utils import (
    quark_quantize_weight_to_mxfp4,
)

# GLM-5.1 MLA dims, per TP=8 rank.
N = int(os.environ.get("N_HEADS", "8"))        # num_heads / TP
LKV = int(os.environ.get("LKV", "512"))        # kv_lora_rank (K)
V = int(os.environ.get("V_HEAD", "256"))       # v_head_dim (N)

dev = "cuda"
print(f">>> fp4bmm_enabled={rocm_aiter_ops.is_fp4bmm_enabled()} "
      f"N={N} LKV={LKV} V={V}", flush=True)

# W_UV is [Lkv, N, V]; vLLM stores W_V = quant(W_UV.permute(1,2,0)) -> [N, V, Lkv]
W_UV = torch.randn(LKV, N, V, dtype=torch.bfloat16, device=dev) * 0.02
W_V, W_V_scale = quark_quantize_weight_to_mxfp4(W_UV.permute(1, 2, 0).contiguous())
print(f">>> W_V {tuple(W_V.shape)} {W_V.dtype}  scale {tuple(W_V_scale.shape)} "
      f"{W_V_scale.dtype}", flush=True)

batch_sizes = [int(b) for b in os.environ.get(
    "BATCHES", "1,2,4,8,16,32,64,128,256,512").split(",")]

def cossim(a, b):
    a = a.float().flatten()
    b = b.float().flatten()
    return torch.nn.functional.cosine_similarity(a, b, dim=0).item()


for B in batch_sizes:
    try:
        x = torch.randn(N, B, LKV, dtype=torch.bfloat16, device=dev) * 0.1
        out = torch.empty(B, N, V, dtype=torch.bfloat16, device=dev)
        out = out.view(-1, N, V)
        res = rocm_aiter_ops.batched_gemm_a16wfp4(
            x, W_V, W_V_scale, out,
            transpose_bm=True, prequant=True, y_scale=None,
        )
        torch.cuda.synchronize()
        # bf16 reference: out_ref[b,n,v] = sum_l x[n,b,l] * W_UV[l,n,v]
        ref = torch.einsum("nbl,lnv->bnv", x.float(), W_UV.float())
        finite = torch.isfinite(res).all().item()
        cs = cossim(res, ref)
        flag = "OK " if (finite and cs > 0.9) else "BAD"
        print(f"B={B:5d}  {flag} out={tuple(res.shape)} finite={finite} "
              f"cos_sim_vs_bf16={cs:.4f} mean={res.float().mean().item():.3e}",
              flush=True)
    except Exception as e:  # noqa: BLE001
        print(f"B={B:5d}  FAIL  {type(e).__name__}: {e}", flush=True)
        traceback.print_exc()
        sys.exit(2)

print(">>> ALL BATCHES DONE", flush=True)
```

<details>

```
Paste output here
```

</details>

### Additional Information

_No response_

srinivamd:
## Summary

Fixes the split-K path in `batched_gemm_a16wfp4` that produces garbage output or `HIP error 700: illegal memory access` for small M values on gfx950. This was reported with GLM-5.1 MLA decode in vLLM.

Fixes: https://github.com/ROCm/aiter/issues/3737
Ref: https://amd-hub.atlassian.net/browse/ROCM-26196

## Root Cause

Three independent defects in the split-K path compound to produce incorrect results when `NUM_KSPLIT > 1` and K is small:

### 1. Missing NUM_KSPLIT re-ensure in kernel-local `get_splitk()`

The copy of `get_splitk()` in `_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` was missing the final `NUM_KSPLIT = triton.cdiv(K, (SPLITK_BLOCK_SIZE // 2))` recalculation that the canonical version in `gemm_afp4wfp4.py:63` has. Without this, the heuristic loop can exit with a `NUM_KSPLIT` that exceeds `ceil(K / (SPLITK_BLOCK_SIZE // 2))`, launching excess K-splits whose starting offset falls outside the valid K range.

### 2. Missing `other=0.0` on masked load in reduction kernel

When `ACTUAL_KSPLIT < MAX_KSPLIT`, the reduction kernel uses a masked `tl.load` for the partial-results buffer `y_pp`. However, the `other=` parameter was not specified. Unexecuted K-splits leave their `y_pp` slots uninitialized (`torch.empty`), and without `other=0.0` the masked-off loads may not reliably return zero, causing garbage to be summed into the final output.

### 3. Missing `NUM_KSPLIT=1` safety clamp in batched wrapper

When `BLOCK_SIZE_K >= 2*K` (a single block covers the entire K dimension), the non-batched `gemm_a16wfp4.py` forces `config["NUM_KSPLIT"] = 1`. The batched wrapper was missing this clamp, allowing split-K to remain active when it serves no purpose and can produce incorrect partial results.

## Changes

| File | Change |
|---|---|
| `aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` | Add `NUM_KSPLIT` re-ensure to local `get_splitk()` (parity with `gemm_afp4wfp4.py:63`) |
| `aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` | Add `other=0.0` to masked `tl.load` in `_batched_gemm_a16wfp4_reduce_kernel` |
| `aiter/ops/triton/gemm/batched/batched_gemm_a16wfp4.py` | Add `config["NUM_KSPLIT"] = 1` clamp when `BLOCK_SIZE_K >= 2*K` (parity with non-batched wrapper) |

## Reproduction

Standalone test from [#3737](https://github.com/ROCm/aiter/issues/3737):
- GPU: gfx950 (MI355)
- Shapes: `x=[8, M, 512]` bf16, `w=[8, 256, 256]` uint8, M in {1..128}
- Before fix: cosine similarity near 0 or NaN for M <= 128
- After fix: correct output for all M values

## Test Plan

- Existing batched GEMM tests should continue to pass (split-K with NUM_KSPLIT=1 is the default gfx950 config path)
- The three fixes are defense-in-depth: each independently prevents the failure mode, and together they close all three gaps

## Submission Checklist


brunomazzottiamd (approved):
LGTM!

However, a I have a suggestion:

`get_splitk` function from `aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` will become identical to `get_splitk` function from `aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py`. Wouldn't it be better to have only one copy of this function?

`aiter/ops/triton/gemm/basic/gemm_a16wfp4.py` already does this:
```python
from aiter.ops.triton.gemm.basic.gemm_afp4wfp4 import (
    get_splitk,
)
```

brunomazzottiamd:
It seems we love to copy & paste code...

There are **7 near-identical copies** of the same split-K heuristic across the Triton GEMM modules.

### The duplicated cluster: `get_splitk(K, BLOCK_SIZE_K, NUM_KSPLIT)`

All of these implement the *exact same* "make `EVEN_K == True` as much as possible" heuristic loop, returning `(SPLITK_BLOCK_SIZE, BLOCK_SIZE_K, NUM_KSPLIT)`:

| # | File | Line | Name |
|---|------|------|------|
| 1 | `aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py` | 32 | `get_splitk` |
| 2 | `aiter/ops/triton/gemm/fused/fused_gemm_afp4wfp4_mul_add.py` | 31 | `get_splitk` |
| 3 | `aiter/ops/triton/gemm/fused/fused_gemm_a8w8_blockscale_mul_add.py` | 26 | `get_splitk` |
| 4 | `aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a16wfp4.py` | 383 | `get_splitk` |
| 5 | `aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` | 282 | `get_splitk` |
| 6 | `aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_afp4wfp4.py` | 289 | `get_splitk` |
| 7 | `aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a8wfp4.py` | 225 | `_get_splitk` |

### Differences between the copies

The body is identical except for three minor variations:

- **`NUM_KSPLIT_STEP` / `BLOCK_SIZE_K_STEP`**: `2` in all copies **except 7** (`gemm_a8wfp4.py`), which uses `4`.
- **Trailing "re-ensure" line** - only copy 1 (`gemm/basic/gemm_afp4wfp4.py`) has it:

```62:63:aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py
    # re-ensuring NUM_KSPLIT is the correct value
    NUM_KSPLIT = triton.cdiv(K, (SPLITK_BLOCK_SIZE // 2))
```

  Copies 6 (`batched_gemm_afp4wfp4.py`) and 7 (`gemm_a8wfp4.py`) return without this re-ensure step; the others (2–5) match 1's loop but I should note 2–5 also lack the re-ensure line (only 1 has it).
- **Comments**: copy 6 has leftover commented-out `print(...)` debug lines inside the loop.

So functionally there are basically **two behaviors**: the `STEP=2` family (1–6) and the lone `STEP=4` variant (7), with 1 being the only one that recomputes `NUM_KSPLIT` at the end.

FYI: @azaidy + @vgokhale

brunomazzottiamd:
@srinivamd, there's no need to refactor everything in the scope of this PR. We can handle https://github.com/ROCm/aiter/pull/3766#issuecomment-4742533751 in another PR!

Shahzeb-AMD:
@jamestangg can we merge this PR?

yifehuan (approved):
LGTM

Fangzhou-Ai:
LGTM

PR #3769, opened 2026-06-17
[FlyDSL AOT] Parallelize standalone main() compile drivers

zhiding512:
The per-module `python -m aiter.aot.flydsl.<op>` drivers compiled kernels in serial for-loops, even though the setup.py build path already fans out across a ProcessPoolExecutor via start_aot/wait_aot. Consolidate the pool logic into a shared `run_jobs_parallel(worker, jobs)` helper in common.py and route all four main()s (moe, gemm, grouped_moe, chunk_gdn_h) through it.

- Extract `_resolve_max_workers` so start_aot and the main() drivers honour the same AITER_FLYDSL_AOT_WORKERS knob and never desync on pool size.
- moe stage1+stage2 and gemm hgemm+preshuffle share one pool each (the compiles are independent; stage2 does not read stage1's cache artifact).
- Per-kernel worker crashes are caught and tallied so one bad kernel does not abort the batch; crash label falls back to a job repr for kinds with no kernel_name (chunk_gdn_h).

Must be multi-process, not multi-thread: compile_one_config mutates process-global state (ARCH / FLYDSL_GPU_ARCH env overrides, FakeTensorMode), which threads would corrupt.


aiter/aot/flydsl/chunk_gdn_h.py
```diff
@@ -48,6 +48,14 @@
 from pathlib import Path
 from typing import Any
 
+# Only when this module is run directly as the AOT compile driver do we hide
+# HIP devices and take aiter's AOT import path; importing it as a library must
+# not mutate the environment. __name__ is already "__main__" here, even though
+# this runs before the aiter imports below that read these variables.
+if __name__ == "__main__":
+    os.environ.setdefault("HIP_VISIBLE_DEVICES", "-1")
+    os.environ.setdefault("AITER_AOT_IMPORT", "1")
```
coderfeli: ？？？？

aiter/aot/flydsl/chunk_gdn_h.py
```diff
@@ -325,9 +334,11 @@ def compile_one_config(
 
     t0 = time.time()
     try:
-        with override_env("ARCH", aot_arch), override_env(
-            "FLYDSL_GPU_ARCH", aot_arch
-        ), FakeTensorMode():
+        with (
+            override_env("ARCH", aot_arch),
+            override_env("FLYDSL_GPU_ARCH", aot_arch),
```
coderfeli, line 330: 2 arch ?

aiter/aot/flydsl/gemm.py
```diff
@@ -36,6 +36,14 @@
 import time
 from typing import Dict, Optional
 
+# Only when this module is run directly as the AOT compile driver do we hide
+# HIP devices and take aiter's AOT import path; importing it as a library must
+# not mutate the environment. __name__ is already "__main__" here, even though
+# this runs before the aiter imports below that read these variables.
+if __name__ == "__main__":
+    os.environ.setdefault("HIP_VISIBLE_DEVICES", "-1")
+    os.environ.setdefault("AITER_AOT_IMPORT", "1")
```
coderfeli: ???

aiter/aot/flydsl/common.py
```diff
@@ -270,6 +270,50 @@ def _affinity_aware_cpu_count() -> int:
...
+        except ValueError as e:
+            raise ValueError(
+                f"AITER_FLYDSL_AOT_WORKERS must be an integer, got {workers_env!r}"
+            ) from e
+    else:
+        max_workers = min(_affinity_aware_cpu_count(), _DEFAULT_MAX_WORKERS)
+    return max(min(max_workers, num_jobs), 1)
+
+
+def run_jobs_parallel(
```
coderfeli, line 287: already have a parallel func and ci aot? is this duplicated?

aiter/aot/flydsl/chunk_gdn_h.py
zhiding512: So these two environment variables should be mentioned in the comments is better?

aiter/aot/flydsl/common.py
zhiding512: For code readable, I think writing two implementations is better since they are for different uses.

aiter/aot/flydsl/gemm.py
zhiding512: Same reason above.

aiter/aot/flydsl/chunk_gdn_h.py
zhiding512: Seems only set FLYDSL_GPU_ARCH is OK. Let's change this.

coderfeli (approved):

coderfeli (dismissed):

valarLip (approved):

PR #3773, opened 2026-06-17
Fix/topk decode dispatch seqlen

chuanbowang2026:
## Summary

`top_k_per_row_decode` previously used `stride0` to choose between multi-block (mb) and one-block (ob) execution paths. In CUDAGraph scenarios, `stride0` is frozen at the pre-allocated width (e.g. `262144`) while the effective row length is much shorter (`~8192`), causing the dispatcher to unnecessarily select the multi-block path.

This PR switches decode to **always use the one-block kernel unconditionally**:

- **C++ side** (`topk_per_row_kernels.cu`): `top_k_per_row_decode` bypasses `should_use_mulblocks()` and directly calls the ob kernel. The original mb/ob dispatch logic is preserved in comments for future reference.
- **Python side** (`topk.py`): `top_k_per_row_decode()` skips `topk_use_mulblocks()` query and passes `workspace=None`. The original dispatch code is commented out for reference.

Prefill dispatch (mb/ob heuristic) is **unchanged**.

### Why not use `max_seqlen` hint?

CUDAGraph replays freeze all CPU-side arguments, so a host-side `max_seqlen` parameter cannot take effect at runtime. The mb persistent kernel also risks cross-block barrier deadlock when other streams occupy CUs.

## Benchmark

`N=8, stride0=262144, effective_len=8192, K=1024`

| GPU | Before (mb) | After (ob) | Speedup |
|-----|-------------|------------|---------|
| MI355X | 30.7 us | 10.8 us | **2.84x** |
| MI308X | 36.7 us | 20.2 us | **1.82x** |

Both paths produce identical top-k index sets. `TOPK_FORCE_PATH=m` has no effect on decode after this change, confirming mb is fully bypassed.

## Future Work

Two follow-up optimizations are planned to further close the gap with SGLang's fused kernel (~9.5 us):

1. **Device-side adaptive block count** — Instead of host-side dispatch, have the kernel itself read actual `seqLen` from device memory and dynamically decide `blocks_per_row` at launch time. This removes the CUDAGraph stride0 mismatch entirely and allows mb to be safely re-enabled for genuinely long sequences in decode.

2. **Dual CUDAGraph capture** — Capture two CUDAGraph instances (one for ob, one for mb) and select at replay time based on runtime batch/seqlen characteristics. This avoids the single-graph freeze problem while keeping full CUDAGraph performance benefits.

### DIFF
--- a/aiter/ops/topk.py
+++ b/aiter/ops/topk.py
@@ -385,14 +385,16 @@ def top_k_per_row_decode(
     stride1: int,
     k: int = 2048,
 ) -> None:
-    """Per-row top-k (decode). Multi-block path uses a persistent, zeroed
-    workspace (memset-free); one-block path allocates internally."""
-    workspace = None
-    if topk_use_mulblocks(numRows, stride0):
-        size = topk_mb_workspace_size(numRows, stride0, k, True)
-        workspace = get_topk_mb_workspace(logits.device, size)
+    """Per-row top-k (decode). Always uses the one-block kernel — the C++
+    side ignores the workspace argument for decode."""
+    # Decode always takes the ob path (see topk_per_row_kernels.cu).
+    # The original mb dispatch is commented out below for reference:
+    #   workspace = None
+    #   if topk_use_mulblocks(numRows, stride0):
+    #       size = topk_mb_workspace_size(numRows, stride0, k, True)
+    #       workspace = get_topk_mb_workspace(logits.device, size)
     return _top_k_per_row_decode(
-        logits, next_n, seqLens, indices, numRows, stride0, stride1, k, workspace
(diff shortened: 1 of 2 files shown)


chuanbowang2026:
Note: this PR adds the `max_seqlen` parameter but defaults to -1, so it's backward compatible â no behavior change without caller-side update.

To activate the fix, ATOM needs a one-line change in `deepseek_v4.py` `_score_topk_decode`:

```python
max_committed = int(attn_metadata.n_committed_csa_per_seq_cpu[:bs].max())
top_k_per_row_decode(..., k=topk, max_seqlen=max_committed)
```
Will submit the ATOM PR after this one merges.

junhaha666 (approved):

PR #3798, opened 2026-06-18
[MI450]tune Dsv4 config for gemm_a8w8_blockscale

Dewei-Wang-sh:


k50112113:
just a quick question, is 4096 8192 the only shape you have tunned? or are you planning to tune more shapes?

valarLip (approved):

PR #3802, opened 2026-06-18
[fea]: gfx1250 allreduce poc

TennyWang1223:
## Motivation

allreduce poc for gfx1250

## Technical Details

naive impl, no perf optimize

## Test Plan

run test script with tp4 and tp2

## Test Result

passed

## Submission Checklist


zufayu:
@TennyWang1223 `Check Code Style with Ruff` is failing, which is blocking all downstream CI from running. Please fix the ruff errors locally (`ruff check --fix && ruff format`) and push — that will unblock the full CI run.

zufayu (commented):
**Overall the direction is good and the dispatch part is clean** (`_detect_gfx1250` centralized via `gcnArchName`, `_select_ops_backend` switch, `world_size > 4` guard, and the gfx1250-skip in `communication.py` has a clear comment). But this PR is not ready to merge yet. **Verdict: NEEDS WORK.**

**Must fix before merge:**

1. **Please split the global JIT file-lock change into a separate PR.** The `FileBaton` rewrite + `mp_lock` (`core.py`) + `_jit_compile` (`cpp_extension.py`) change the concurrent JIT build behavior of **all ops on all archs**. This should not ride together with a gfx1250 allreduce POC and get merged without independent review. Two hardening points while you split it out:
   - `try_acquire()` silently does `pass` when `os.write` fails → leaves a 0-byte lock; a live builder can then be judged "stale" after the grace window. Better to fail loudly than swallow the write error.
   - `release()` calls `os.remove(lock_file_path)` unconditionally. If the lock was already broken as stale and re-acquired by another process, this removes someone else's lock. Better to verify we still own it before removing.
   - (Also note: pid-based stale check can false-negative on pid reuse. It is still better than the old deadlock-always behavior, so not blocking, just be aware.)

2. **Add correctness validation for the allreduce.** Test Result only says "passed" — not enough for a collective that produces numeric output, because a wrong result is silent (no crash, just accuracy drop). Please add an `allclose` / element-wise compare against RCCL or the existing `custom_all_reduce`, on tp2 and tp4.

3. **Rebase.** PR is `CONFLICTING` now, cannot be merged as-is.

**Can follow up (not blocking merge):**

4. All 14 gfx1250 ops use `@compile_ops(..., develop=True)`. OK for POC, but it bypasses the JIT cache (recompile on every cold start) — please remove before this enters a release branch.

5. `module_custom_all_reduce` is removed from `_get_ck_exclude_modules()`. Its config does not reference CK, so if `custom_all_reduce.cu` has no `ck_tile::` include the removal is correct — please just confirm gfx942/gfx950 CK-free build still works.

6. CUDA graph is not supported on gfx1250 (you already have a warning + TODO, good). Please open a follow-up issue so it is tracked and not silently forgotten.

**Minor:**

7. The cross-rank signal barrier in `custom_all_reduce_gfx1250.cuh` uses `__MEMORY_SCOPE_DEVICE`. For cross-GPU visibility over XGMI/PCIe, please confirm device scope is sufficient here vs system scope.

Thanks for the work — the core approach is fine, main ask is just to **unbundle the JIT lock change** and **add a numeric correctness check** before we merge.

TennyWang1223:
another branch has already merged into main

PR #3805, opened 2026-06-18
fix(fmoe tune): dedup fused/non-fused stage1 separately 

rbrugaro-amd:
## Problem
FmoeTuner.post_process under-selects fused (_fp4/_fp8) stage-1 kernels for a4w4/a8w4 MoE, shipping CK/non-fused configs even when fusion is faster on a fair basis.

## Root cause
Selection runs in two passes:

Dedup to the best candidate per (stage, block_m, flat) by raw us.
Quant-fairness penalty: the separately-benchmarked dynamic_mxfp4_quant+sort cost is added to the us1 of every non-_fp4/_fp8 stage-1 (fused kernels embed quant+sort in the epilogue, so it's already in their us1).
A fused stage-1 does that extra quant work in-kernel, so its raw us is slightly higher than an equivalent non-fused kernel at the same block_m. Since both share the dedup key, the fused candidate is evicted in pass 1 — before the pass-2 penalty (which would make it win) is ever applied. The fused kernel is no longer in the pool when idxmin(us1+us2) runs.

## Fix
Add an _is_fused flag (kernelName endswith _fp4/_fp8) to the dedup key so the best fused and best non-fused per (stage, block_m, flat) both survive into the penalty + final selection. Helper column dropped immediately; downstream merge/columns unchanged. No effect on shapes without fused candidates.

## Validation
DeepSeek-R1 MXFP4 shared-expert-fusion shape (expert=33, topk=9, 7168/2048): re-tuning flips the ≥2048 token tiers from CK to fused flydsl stage-1. A cuda-graph BS=256 decode bench shows ~6% lower latency / higher throughput vs the CK selection 

## Risk
Low — pure selection-logic change in the offline tuner; no kernel/runtime changes. Shapes with only non-fused candidates are unaffected.


okorzh-amd:
@valarLip could you please review this PR?

rbrugaro-amd:
Rebased onto latest main and re-tuned the DeepSeek-R1 MXFP4 shared-fusion shape (expert=33, topk=9, 7168/2048, per_1x32) to confirm the selection change still behaves. The ≥2048 tiers still flip CK → fused flydsl stage1 with this patch; reverting just the _is_fused dedup key sends 2048 back to a CK 2-stage kernel (the fused candidate is present in the profile but evicted in the pre-penalty dedup — the exact issue this PR fixes).
Checked the new selection/backends landed since I opened this don't interact with this path:
 - The FLAT-vs-non-FLAT head-to-head (_us_total) is gated if q_type != per_1x128: raise inside the try/except, so it's a no-op for per_1x32 fp4 — the shape this PR targets.
- RadeonFlow MXFP4 a4w4 (#3832) isn't wired into FmoeTuner candidate generation, and Opus stage2 (#3856) is a8w4/i4x2-gated — neither adds candidates for this fp4 shape.
So for the affected shape the post-merge selection reduces to dedup(+_is_fused) → quant-fairness penalty → idxmin, same as when this was first validated.

lalala-sh (commented):
Please simplify the code comments.

lalala-sh (approved):
LGTM

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
```diff
@@ -4090,8 +4090,21 @@ def post_process(self, results, args, topk=-1, fast_mode=False):
                 & (profileDF["us"] != -1)
                 & (profileDF["err"] <= args.errRatio)
             ]
-            profileDF = profileDF.sort_values("us").drop_duplicates(
-                ["stage", "block_m", "flat"], keep="first"
+            # Fused (_fp4/_fp8) stage1 kernels embed the inter-stage quant+sort;
```
lalala-sh: Please simplify the code comments. Thx

lalala-sh (changes requested):

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
rbrugaro-amd: done

rbrugaro-amd:
@lalala-sh addressed your feedback

lalala-sh (approved):

rbrugaro-amd:
@lalala-sh @ROCm/team_aiter test passed and is approved, can we merge please?

lalala-sh:
> @lalala-sh @ROCm/team_aiter test passed and is approved, can we merge please?

Done, thanks for your contribution.

PR #3813, opened 2026-06-19
Simplify ck_gemm_a8w8_blockscale GemmSpecialization construction

jbelloncastro:
## Motivation

Simplifies the logic behind constructing an enum value for GemmSpecialization.

## Technical Details

The global variable unordered_map is constructed during library initialization and performs memory allocations that are not really necessary to find the right configuration from the dimension and block size parameters.

The change reassigns the enum values such that they can be treated as a mask. The padding conditions simply add a new bit to the mask if necessary.

The value reassignment will make chains of conditionals slightly more expensive, but not as much as the previous cost of enum construction.

## Submission Checklist


jbelloncastro:
Hi @ROCm/team_aiter ! Could someone please review this? It is a small change 😊

PR #3814, opened 2026-06-19
[TRITON] Tuned DSV4-Flash FP8 GEMM configs

skysnow2001:
## Motivation

To run DeepSeek-V4-Flash on gfx1201 (RDNA4 / Navi48) via vLLM, the FP8 A8W8 block-scale Triton GEMM (GEMM-A8W8_BLOCKSCALE) needs gfx1201 tuning configs for the per-rank (N, K) shapes the model hits in its non-MoE linears at TP=8. This PR adds those configs so the operator runs tuned on this arch.

## Submission Checklist


brunomazzottiamd (approved):
LGTM!

Please double check why the test jobs are failing. We can merge once everything is passing.

PR #3831, opened 2026-06-22
BF16 gptoss tuning

azaidy:


op_tests/op_benchmarks/triton/bench_gemm_a16w16.py
```diff
@@ -46,25 +47,26 @@ def bench_gemm_fn(
     mem_write = (M * N) * x.element_size()
     mem = mem_read + mem_write
 
+    bench_fn = triton.testing.do_bench_cudagraph if cudagraph else triton.testing.do_bench
```
brunomazzottiamd: Ali, can you please educated me on the difference between original `do_bench` and `do_bench_cudagraph`? When should I use `do_bench_cudagraph`?

aiter/ops/triton/_gluon_kernels/gfx1250/gemm/basic/gemm_a16w16.py
```diff
@@ -53,9 +53,12 @@ def create_shared_layouts(
 
 
 def create_wmma_layouts(num_warps):
-    warp_bases = [(0, 1)]
-    for i in range(int(math.log2(num_warps // 2))):
-        warp_bases.append((1 << i, 0))
+    warp_bases = []
```
brunomazzottiamd, line 56: As far as I can see, the old and new versions of `create_wmma_layouts` are equivalent. So far so good.

However, something caught my attention:
 
* `log2` is only defined for positive numbers.
* `log2(1)` is zero and this will produce an empty list. Is it safe to return an empty list here?

Should we `assert num_warps > 0` or `assert num_warps > 1`?

brunomazzottiamd (dismissed):
LGTM!

vgokhale (approved):

op_tests/op_benchmarks/triton/bench_gemm_a16w16.py
vgokhale: The cudagraph version does graph capture and will run the kernels b2b

PR #3837, opened 2026-06-22
[gfx1201] Dedicated CK-free reshape_and_cache (paged KV write) for RDNA4

carlushuang:
## Summary
Adds a dedicated, **CK-free** HIP module (`module_cache_reshape`) for the paged KV-cache **write** (`reshape_and_cache` / `reshape_and_cache_flash`) so it builds and runs on **gfx1201 / gfx1200 (RDNA4)**.

## Problem
On RDNA4 the HIP `reshape_and_cache` launch faults at runtime with no catchable exception. The root cause is **not** the kernel source — it's a missing code object:
- `reshape_and_cache` lives in `csrc/kernels/cache_kernels.cu`, a TU that also contains CK/`ck_tile`-heavy and CDNA-flavored kernels (MLA concat with `ck_tile::` buffer views, the DeepSeek indexer with inline `v_max3_f32` asm, …). That pulls the whole module onto the prebuilt-CK path.
- The prebuilt `module_cache.so` ships only `gfx942`/`gfx950` code objects (`llvm-objdump --offloading`), so on gfx1201 `hipModuleLaunchKernel` finds no matching code object → fault.

The reshape kernels themselves are arch-clean: a plain scatter using only the `opus::` dtype/dispatch layer (RDNA4-ready) and the `ck_tile_shim` fallback in `aiter_hip_common.h` (`#if !ENABLE_CK`).

## Change
Split the two write kernels into their own module that builds on any arch, with or without `composable_kernel`:
- `csrc/kernels/cache_reshape_kernels.cu` — `reshape_and_cache` (+ `_flash`) kernels + launchers, **byte-for-byte identical** to the originals (extracted verbatim).
- `csrc/pybind/cache_reshape_pybind.cu` — pybind for the new module.
- `aiter/jit/optCompilerConfig.json` — `module_cache_reshape` entry; intentionally **not** in the CK-exclude set, so it builds in CK-free environments too.
- `aiter/ops/cache.py` — route `reshape_and_cache` / `reshape_and_cache_flash` to `module_cache_reshape`.
- `op_tests/test_reshape_and_cache_hip.py` — torch-reference test (HND + 5D-SHUFFLE asm, bf16/fp16, padded `slot=-1`).

The existing `module_cache` is untouched; all other cache ops keep using it.

## Validation
RX 9070 XT (**gfx1201**, ROCm 7.2): the new module builds in ~9 s and the HIP kernel output **matches the torch reference bitwise** for all 4 cases:

```
[SHUFFLE/asm bf16] K=True V=True    [HND bf16] K=True V=True
[SHUFFLE/asm fp16] K=True V=True    [HND fp16] K=True V=True
```

## Relationship to #3319
#3319 adds a **triton** fallback for the same write (incl. the SHUFFLE asm layout). The two are complementary:
- This HIP module makes the native kernel work for **from-source builds** on gfx1201 (and is a candidate to add to the prebuilt arch set so wheels work out of the box).
- The triton fallback in #3319 still serves **prebuilt wheels / zero-build installs** where no gfx1201 code object exists and `composable_kernel` may be absent.

## Scope / notes
- Covers `reshape_and_cache` (incl. the static per-tensor FP8-scale store path) and `reshape_and_cache_flash`. The per-token / block FP8 quant variants stay in `module_cache` (out of scope).
- `clang-format` applied; kernel logic unchanged.

### DIFF
--- a/aiter/jit/optCompilerConfig.json
+++ b/aiter/jit/optCompilerConfig.json
@@ -106,6 +106,20 @@
         "verbose": "False",
         "blob_gen_cmd": "''"
     },
+    "module_cache_reshape": {
+        "srcs": [
+            "f'{AITER_CSRC_DIR}/pybind/cache_reshape_pybind.cu'",
+            "f'{AITER_CSRC_DIR}/kernels/cache_reshape_kernels.cu'"
+        ],
+        "flags_extra_cc": [],
+        "flags_extra_hip": [
+            "'-DENABLE_FP8'"
+        ],
+        "extra_ldflags": "None",
+        "extra_include": [],
+        "verbose": "False",
+        "blob_gen_cmd": "''"
+    },
     "module_custom_all_reduce": {
         "srcs": [
             "f'{AITER_CSRC_DIR}/pybind/custom_all_reduce_pybind.cu'",
--- a/aiter/ops/cache.py
+++ b/aiter/ops/cache.py
@@ -19,7 +19,7 @@ def copy_blocks(
 ) -> None: ...
 
 
-@compile_ops("module_cache", develop=True)
+@compile_ops("module_cache_reshape", develop=True)
 def reshape_and_cache(
     key: torch.Tensor,
     value: torch.Tensor,
@@ -33,7 +33,7 @@ def reshape_and_cache(
 ) -> None: ...
 
 
-@compile_ops("module_cache", develop=True)
+@compile_ops("module_cache_reshape", develop=True)
 def reshape_and_cache_flash(
     key: Tensor,
     value: Tensor,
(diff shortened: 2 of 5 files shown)


carlushuang:
Superseded — closing.

This dedicated CK-free `reshape_and_cache` module is redundant: the existing `module_cache` reshape kernels are arch-clean and build/run correctly on gfx1201 (validated bitwise). They only failed because the prebuilt `.so` shipped CDNA-only (`gfx942;gfx950`) and the JIT loader imported it instead of rebuilding for the running GPU.

The general fix is ROCm/aiter#3839 (loader arch-coverage guard), tracked by ROCm/aiter#3846. ATOM PR ROCm/ATOM#811 now depends on #3846/#3839. Closing in favor of that.

PR #3876, opened 2026-06-23
[Feature][HIP] Support fused shared expert topk append

yuychang:
## Motivation

SGLang can fuse shared experts into the MoE routing result by asking AITER's grouped top-K operator to return both routed experts and shared experts in the same output tensors.

Before this change, AITER's grouped top-K HIP path only populated the routed top-K entries. Callers that needed shared experts had to append those entries separately after the top-K call, which adds extra Python-side handling and makes it harder to use AITER as a drop-in fused routing backend.

This PR adds native support for appending fused shared expert entries directly inside the AITER grouped top-K HIP kernels. The routed top-K selection remains unchanged, and shared expert IDs/weights are written after the routed top-K columns when explicitly requested.

## Technical Details

This change extends the AITER grouped top-K API surface and HIP kernel implementation to support optional shared expert append behavior.

The following optional arguments were added to `grouped_topk` and `biased_grouped_topk`:

```
num_fused_shared_experts: int = 0
fused_shared_experts_scaling_factor: float = 1.0
shared_expert_base: int = -1
```

Their behavior is:

- `num_fused_shared_experts`: number of shared experts to append after the routed top-K results.
- `fused_shared_experts_scaling_factor`: weight assigned to each appended shared expert.
- `shared_expert_base`: base expert ID for the first shared expert. Appended shared expert IDs are written as `shared_expert_base + i`.

When `num_fused_shared_experts == 0`, the existing behavior is preserved.

When `num_fused_shared_experts > 0`, the kernels append shared expert entries after the routed top-K expert columns:

```
topk_ids[token]     = [routed_id_0, ..., routed_id_k, shared_base, shared_base + 1, ...]
topk_weights[token] = [routed_w_0,  ..., routed_w_k,  shared_scale, shared_scale, ...]
```

The implementation updates:

- **`aiter/ops/topk.py`**
  - Extends the Python operator signatures.
  - For `biased_grouped_topk`, routes fused shared expert append requests through the HIP implementation so the append happens in-kernel.
- **`csrc/include/moe_op.h`**
  - Extends the C++ declarations for `grouped_topk` and `biased_grouped_topk`.
- **`csrc/include/rocm_ops.hpp`**
  - Adds pybind arguments for the new optional parameters.
- **`csrc/kernels/topk_softmax_kernels_group.cu`**
  - Propagates the new parameters through the grouped top-K launch macros.
  - Appends shared expert weights and IDs in both grouped top-K kernel variants.
  - Adds validation for shared expert append mode:
    - `shared_expert_base` must be provided.
    - The output row stride must be large enough for routed plus shared expert columns.
- **`op_tests/test_biased_grouped_topk_shared_append.py`**
  - Adds a focused regression test for the new append behavior.

The test compares the fused kernel output against the existing routed top-K result plus an expected shared expert suffix. This verifies that routed expert selection is preserved and that shared expert IDs/weights are appended in the expected columns.

## Test Plan

Ran the new targeted regression test:

```bash
python -m pytest -q op_tests/test_biased_grouped_topk_shared_append.py
```

Also checked the staged diff for whitespace and patch formatting issues:

```bash
git diff --cached --check
```

## Test Result

The targeted regression test passed:

```
............                                                             [100%]
12 passed, 2 warnings in 41.38s
```

The warnings were unrelated Cython distutils deprecation warnings from the local Python environment.

The diff check also passed with no whitespace or formatting issues:

```bash
git diff --cached --check
# no output
```

## Submission Checklist

### DIFF
--- a/aiter/ops/topk.py
+++ b/aiter/ops/topk.py
@@ -85,6 +85,9 @@ def biased_grouped_topk_hip(
     topk_grp: int,
     need_renorm: bool,
     routed_scaling_factor: float = 1.0,
+    num_fused_shared_experts: int = 0,
+    fused_shared_experts_scaling_factor: float = 1.0,
+    shared_expert_base: int = -1,
 ) -> None: ...
 
 
@@ -98,6 +101,9 @@ def grouped_topk(
     need_renorm: bool,
     is_softmax: bool = True,
     routed_scaling_factor: float = 1.0,
+    num_fused_shared_experts: int = 0,
+    fused_shared_experts_scaling_factor: float = 1.0,
+    shared_expert_base: int = -1,
 ) -> None: ...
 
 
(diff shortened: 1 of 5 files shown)


yuychang:
Found another way to implement shared-expert-append fusion with minimal changes.

PR #3882, opened 2026-06-23
perf(tuned_gemm): route skinny bf16 GEMMs (N<=256, N%64==0, M>16) to splitk asm

amd-pedghazi:
## Motivation

DSA models (DeepSeek-V3.2, GLM-5.1) run a "lightning indexer" in every prefill step: a bf16→bf16 GEMM with `N = 256` (and `N = 32`) and `K = 6144`. At prefill, `M` (the token count) is in the hundreds-to-thousands.

`get_GEMM_A16W16_config`'s skinny dispatch (`is_skinny_default_shape`) only covers `M <= 16`. For these indexer shapes at large `M` there is **no tuned config**, so the dispatcher falls back to `torch.matmul`. On gfx950 (MI300/MI350) this is:

- **50–100× slower** than the available asm kernel for this shape, directly inflating DSA prefill/TTFT, and
- a **stability risk**: the torch fallback has **MAF'd (memory access fault) at `M = 16384`** on gfx950 in practice.

This is on the hot path of every DSA prefill on AMD, so the missing dispatch is both a perf regression and a crash risk.

**Where it helps:** the sparse DSA prefill path (the indexer GEMM) for DeepSeek-V3.2 / GLM-5.1 on AMD GPUs — i.e. the above-threshold path used by SGLang/vLLM DSA serving. (Companion SGLang PR enabling dense-MHA below the DSA threshold on gfx950: [sgl-project/sglang#29050](https://github.com/sgl-project/sglang/pull/29050).)

## Technical Details

Add a complementary `elif` in `get_GEMM_A16W16_config` for `N <= 256 and N % 64 == 0 and K % 64 == 0 and M > 16` (bf16/fp16 in, bf16 out, no bias/scaleAB/bpreshuffle) that routes to the `bf16gemm_fp32bf16_tn_64x64_splitk_clean` asm kernel, reusing the existing asm dispatch via `config` (`libtype="asm"`, `solidx=5`, `splitK = max(1, K//512)`).

- The `N % 64 == 0` guard is **essential**: the chosen asm kernel has `tileN=64` and the C++ heuristic rejects any `N` not a multiple of `tileN`. So the indexer's `N=256` GEMM routes to asm, while `N=32` correctly stays on the default path.
- The asm kernel handles any `M` as long as `K` is 64-aligned.
- Mathematically equivalent: output rows are independent across the `M` dimension.

Single-file change: `aiter/tuned_gemm.py` (+26 lines). No behavior change for any shape that already had a tuned config.

## Test Plan

- Run a DSA model (GLM-5.1-FP8) prefill on gfx950 and confirm the indexer `N=256, K=6144` GEMM dispatches to the asm kernel (not the torch fallback), and that `N=32` stays on the default path.
- `op_tests/test_gemm_a16w16.py` over skinny / large-M bf16 shapes for numerical parity against the torch reference.

## Test Result

On MI350X (gfx950), GLM-5.1-FP8 TP8 server, the dispatcher logs confirm the intended routing:

```
[aiter] shape ... M:304, N:256, K:6144 ... using asm solution:5    # routed to splitk asm
[aiter] shape ... M:304, N:32,  K:6144 ... using torch solution:0  # N%64!=0 -> default (correct)
[aiter] shape ... M:4,   N:32,  K:6144 ... using skinny solution:2 # pre-existing M<=16 path
```

The full DSA server (78 layers × TP8) runs end-to-end through CUDA-graph capture and serving with the indexer GEMM on the asm kernel — no torch fallback and no MAF at large M. End-to-end GLM-5.1-FP8 accuracy on this build is unaffected (GSM8K 0.965; long-context needle retrieval 100% at 10k–60k).

## Submission Checklist


valarLip:
good catch, however let's just tune these shapes instead of hard code this logic

amd-pedghazi:
> good catch, however let's just tune these shapes instead of hard code this logic

Thanks @valarLip! Agreed. Tuning is the cleaner approach, and the asm kernel would probably win there anyway.

Another thing to watch out is that the tuned configs are keyed on exact M, but here M is the indexer's prefill token count, which changes a lot per batch. Tuning a fixed set of shapes only covers those M values and any other M still falls back to torch.matmul, which is the 50-100 slower and the M=16384 crash this PR fixes.

So I think we want both: (1) add these shapes to the untuned set and tune them, and (2) keep an M-agnostic fallback so other large-M skinny bf16 shapes still go to asm instead of torch. Happy to run the tuner on the indexer shapes (N=256, K=6144 over a range of M) and have the fallback pick its kernel from the tuned results instead of hardcoding it.

How does it sound to you?

valarLip:
these asm kernels not works for all arch, we can't keep this logic

PR #3894, opened 2026-06-24
Fix/kv shuffle fp8 scale

DmitriiNikolaevAtAmd:
AITER's FP8 paged-attention kernel computed the wrong flat index into the per-head KV scale buffer when KV-cache shuffling was enabled.


mjkvaak-amd:
Also fixed here https://github.com/ROCm/aiter/pull/3896.

PR #3896, opened 2026-06-24
Fix HIP fp8 paged-attention kPerHead scale OOB page fault.

JohnNikolay84:
Use physical KV block count and physical block offsets for fp8 scale indexing instead of block-table width, and normalize vLLM block-major scales to the HIP head-major layout before launch.

## Motivation

HIP pa kernel with SHUFFLED layout  is producing a page fault on certain inputs.

## Technical Details

number of physical pages is now passed down explicitly as an argument rather then being  derived by the kernel.

## Test Plan

Re-run the reproducer from Mikko Tukiainen and make sure it pass for both asm and hip.

## Test Result

No more gpu faults.

## Submission Checklist

### DIFF
--- a/csrc/cpp_itfs/pa/pa.cpp.jinja
+++ b/csrc/cpp_itfs/pa/pa.cpp.jinja
@@ -16,6 +16,7 @@
                       const int num_kv_heads,           \
                       const int num_heads,              \
                       const int max_num_blocks_per_seq, \
+                      const int num_kv_blocks,          \
                       const int q_stride,               \
                       const int kv_block_stride,        \
                       const int kv_head_stride,         \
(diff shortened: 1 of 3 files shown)


JohnNikolay84:
@mjkvaak-amd is the vllm issue fixed now ?

mjkvaak-amd:
> @mjkvaak-amd is the vllm issue fixed now ?

Yes, this PR fixes the GPU segfaults we saw earlier when running MiniMax-M2.5 with FP8 KV cache and `VLLM_ROCM_SHUFFLE_KV_CACHE_LAYOUT=1`, which uses AITER's `paged_attention_common/paged_attention_rocm`.

PR #3901, opened 2026-06-24
[FLYDSL] add MLA reduce decode kernel for gfx942

anhminhnguyenhoang:
## Motivation

Split-KV MLA decode stage 2 **reduce/combine** merges per-split fp32 partial `O`/`LSE` values into the final bf16/fp16 output using an LSE-weighted online softmax (`kn_mla_reduce_v1` in `csrc/kernels/mla/reduce.cu`).

This PR adds a FlyDSL port for **gfx942 (MI300X)**, targeting the validated **GLM-5.2** sparse MLA decode configuration at TP8 (`H=16`, `Dv=512`).

## Technical details

- **Kernel:** The default configuration uses 256 threads per block. `Tier.ALL`, the production kernel, selects one of four per-tile algorithm bodies on-device: `simple`, `m64`, `m256`, or `mlds`. It handles multi-token direct-wrapper calls and empty tiles safely.
- **Launch selection:** The wrapper first selects cooperative split-K (`plan_splitk_capture_safe`) for low-tile, high-split decode. Otherwise, it uses adaptive direct launch for eligible sparse multi-tile inputs. All other calls select conventional 3-D or persistent grid-stride launch from the total-work threshold. `low_direct_pmap_thr` is an adaptive-launch optimization that bypasses LDS staging for low split counts; it is not a separate launch mode.
- **Wiring:** `flydsl_mla_reduce_v1()` is a drop-in alternative enabled with `AITER_MLA_REDUCE_FLYDSL=1`. HIP remains the production default. Tier selection happens on-device, so per-tile split counts can vary within a fixed captured launch topology. Changes that select split-K, adaptive, or persistent dispatch require a separately captured topology.
- **Occupancy:** `waves_per_eu` is an explicit wrapper argument and is part of the FlyDSL launch-cache identity for both normal and split-K dispatch.

## Test plan

Validation environment:

| Component | Value |
|---|---|
| GPU | MI300X (`gfx942`) |
| ROCm | 7.2.4 |
| PyTorch | 2.10.0+rocm7.2.4 |
| FlyDSL | 0.2.4 |

`op_tests/test_flydsl_mla_reduce.py` is a pytest-free, self-contained op test. The shared test helper was inlined into the test file; run the test directly with Python:

```bash
python3 op_tests/test_flydsl_mla_reduce.py
```

The suite compares production-shaped metadata against HIP and torch references across `H=8, Dv=256` torch-reference cases, `H=128, Dv=512` HIP-reference tier stress cases, and the production `H=16, Dv=512` serving fixtures. It covers variable splits, gapped indices, tier boundaries, empty tiles, split-K thresholds, adaptive launches, the `actual_max_splits` gate, serving OOB guards, and CUDA-graph capture/replay.

## Performance benchmark

The following results are CUDA-graph replay latency on `H=16`, `Dv=512`, bf16. Values are the mean of two runs: one with fresh JIT and FlyDSL caches and one warm-cache repeat. `wrapper` denotes `flydsl_mla_reduce_v1()`. All wrapper and HIP correctness checks reported zero error.

| Active tiles | Splits | HIP (µs) | FlyDSL (µs) | FlyDSL / HIP | Speedup |
| 1 | 128 | 13.88 | **7.23** | **0.52×** | **1.92×** |
| 8 | 32 | 5.99 | **4.91** | **0.82×** | **1.22×** |
| 8 | 26 | 5.47 | **4.92** | **0.90×** | **1.11×** |
| 8 | 13 | 4.39 | **3.97** | **0.90×** | **1.11×** |
| 8 | 6 | 3.84 | **3.41** | **0.89×** | **1.13×** |
| 8 | 5 | 3.78 | **3.37** | **0.89×** | **1.12×** |
| 8 | 3 | 3.77 | **3.02** | **0.80×** | **1.25×** |
| 8 | 2 | 3.73 | **2.77** | **0.74×** | **1.35×** |

### End-to-end serving

The kernel win above is small relative to a full decode step, so it was also measured end to end. Two vLLM servings of GLM-5.2-FP8 (TP8, fp8 KV cache, 131k context, `gpu_memory_utilization: 0.82`) were run over the same twelve request shapes, one with `AITER_MLA_REDUCE_FLYDSL=0` and one with `=1`.

| Input length | Shapes | Output throughput | TPOT (decode) |
|---:|---:|---:|---:|
| 1K | 4 | +0.8% | −1.4% |
| 8K | 4 | +3.0% | −3.2% |
| 128K | 4 | +1.3% | −2.3% |
| **All** | **12** | **+1.7%** | **−2.3%** |

Geometric means; negative TPOT is faster. Across the twelve shapes, decode TPOT improved 2.3%.

TPOT is the metric this PR should move since the gate only fires on decode and it improves on most shapes. Per-shape throughput suggests no regression. Measured TTFT differences are not attributable to this PR as prefill never reaches the reduce kernel.

All lengths cover concurrency 1/2/4/8. The 1K and 8K rows use nominal 1,024- and 8,192-token inputs, respectively, with 1,024 output tokens and 8/16/32/64 prompts; these retain the original `random-range-ratio: 0.8` randomized traffic. The 128K sweep instead uses fixed 128,000-token inputs and 1,024-token outputs (`random-range-ratio: 0.0`), so every request stays within the 131,072-token model context. It uses 4/8/16/24 prompts, and both HIP and FlyDSL completed all four points with zero failed requests.

### Accuracy

The FlyDSL serving path completed the 5-shot GSM8K evaluation (1,319 samples, deterministic decoding, 32 client requests in flight). The saved result reports **95.07%** `flexible-extract` exact match and **95.00%** `strict-match` exact match, consistent with thereported 94.9% HIP score; with one run and roughly 0.6 percentage-point standard error.

### Safety coverage

- **Split-K overwrite:** `reduce_indptr` reaches the combine kernel, which reads `reduce_final_map` and writes output/LSE only when `n_splits > 1`. Mixed `[0, 64]` and `[1, 64]` cases with a stale map are tested eagerly and under CUDA-graph replay.
- **Production dispatch:** FlyDSL is selected only for the validated target: gfx942, one-token decode, `H=16`, and `Dv=512`. Unsupported architectures, prefill or multi-token calls, and unsupported shapes remain on HIP without compiling FlyDSL.
- **Pointer ABI:** The direct FlyDSL API rejects invalid pointer dtypes, ranks, shapes, devices, contiguity, output strides, and split counts before JIT compilation.
- **Capture safety:** CUDA-graph capture requires a warmed CSR cache or an explicit `actual_max_splits` value; an unresolved capture does not launch.


aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+    elif tier == Tier.ALL:
+        NLSE = 0  # unused; runtime dispatch picks NLSE per branch
+    else:
+        NLSE = _TIER_NLSE[tier]
+
+    # ---- LDS layout (all tiers: pmap; massive adds scale + overflow) ----
+    # [ pmap : LDS_MAX_SPLITS i32 ] ++ [ lse_scale : LDS_MAX_SPLITS f32 ]
+    # ++ [ local_lse overflow : (NLSE-4 lanes) f32 ]  (MLDS/ALL only)
+    GPU_ARCH = get_rocm_arch()
+    allocator = SmemAllocator(
```
coderfeli: don't use this smem allocator. it will be deprecated.

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+        try:
+            yield if_op.then_block
+        finally:
+            blk = if_op.then_block
+            if (not blk.operations) or not isinstance(blk.operations[-1], scf.YieldOp):
+                scf.YieldOp([])
+
+
+@contextmanager
+def _if_else(if_op):
```
coderfeli: use raw if else, no scf.

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+    raw = os.environ.get("AITER_MLA_REDUCE_WAVES_PER_EU")
+    if raw is None:
+        return default
+    return int(raw)
+
+
+@contextmanager
+def _if_then(if_op):
+    """SCF IfOp then-region; auto-insert scf.yield if missing."""
+    with ir.InsertionPoint(if_op.then_block):
```
coderfeli: use if else. remove this.

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+LDS_MAX_SPLITS = 304  # >= MI300X CU count; compile-time LDS cap
+NLSE_MLDS = (LDS_MAX_SPLITS + WARP - 1) // WARP  # ceil(304/64) = 5
+
+# Production default (opt4 sweep: 4 ~= 6 < 8 on H=16 Dv=512 tiles=8 splits=32).
+_DEFAULT_WAVES_PER_EU = 4
+
+
+def waves_per_eu_from_env(default: int = _DEFAULT_WAVES_PER_EU) -> int:
+    """Read ``AITER_MLA_REDUCE_WAVES_PER_EU`` (sweep/tuning knob)."""
+    raw = os.environ.get("AITER_MLA_REDUCE_WAVES_PER_EU")
```
coderfeli: use input param instead of env?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+                fx.gpu.barrier()
+
+            # HIP kn_mla_reduce_v1_ps: skip tiles at the CSR sentinel
+            # (reduce.cu:692) or with n_splits<=1. Collapse the seq-loop upper
+            # bound instead of an scf.IfOp body-wrap so inactive tiles never
+            # enter gather/store paths even if reduce_final_map holds garbage
+            # q-ranges (serving tail).
+            has_work = fx.arith.andi(
+                fx.arith.cmpi(fx.arith.CmpIPredicate.sgt, n_splits, fx.Int32(1)),
+                fx.arith.cmpi(fx.arith.CmpIPredicate.ne, t0, last),
```
coderfeli: use & < > instread.

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+                row0_idx0 = fx.Int32(lds_pmap[fx.Index(0)])
+                row1_idx0 = fx.Int32(lds_pmap[fx.Index(1)])
+                qo_len = row1_idx0 - row0_idx0
+                q_start = fx.Int32(tile) * qo_len
+                q_end = (fx.Int32(tile) + fx.Int32(1)) * qo_len
+
+            if fx.const_expr(not disable_guards):
+                q_valid = fx.arith.andi(
+                    fx.arith.cmpi(fx.arith.CmpIPredicate.sge, q_start, fx.Int32(0)),
+                    fx.arith.cmpi(fx.arith.CmpIPredicate.slt, q_start, num_final_rows),
```
coderfeli: +++

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+                """Bounds-guarded row index from an already-loaded pmap value.
+
+                Shared by the scalar gather_row (LDS read per split) and the
+                vectorized group loader (one ds_read_b128 per group), so both
+                paths apply the identical clamp/mask.
+                """
+                row_i32 = pmap_i32 + local_seq
+                if fx.const_expr(not disable_guards):
+                    in_bounds = fx.arith.andi(
+                        fx.arith.cmpi(fx.arith.CmpIPredicate.sge, row_i32, fx.Int32(0)),
```
coderfeli: +++

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1514 @@
...
+import os
+from contextlib import contextmanager
+
+import flydsl.compiler as flyc
+import flydsl.expr as fx
+from flydsl.compiler.kernel_function import CompilationContext
+from flydsl.expr.typing import T
+from flydsl._mlir import ir
+from flydsl._mlir.dialects import scf, math as mlir_math, gpu as mlir_gpu
+from flydsl._mlir.dialects import llvm as llvm_dialect
```
coderfeli: clean all usage of these _mlir

coderfeli (changes requested):

valarLip (dismissed):

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+    """
+    if not splitk_enabled():
+        return False, 1, 0
+    factor = int(os.environ.get("MLA_SPLITK_FACTOR", _SPLITK_FACTOR_DEFAULT))
+    min_splits = int(
+        os.environ.get("MLA_SPLITK_MIN_SPLITS", _SPLITK_MIN_SPLITS_DEFAULT)
+    )
+    max_active = int(
+        os.environ.get("MLA_SPLITK_MAX_ACTIVE_TILES", _SPLITK_MAX_ACTIVE_TILES_DEFAULT)
+    )
```
coderfeli: we should not have so many env val to make the kernel work. clean?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+
+# Split-K scratch is cached for CUDA-graph replay; active tiles form a CSR prefix.
+_SPLITK_FACTOR_DEFAULT = 16
+_SPLITK_MIN_SPLITS_DEFAULT = 64
+_SPLITK_MAX_ACTIVE_TILES_DEFAULT = 64
+
+
+def splitk_enabled() -> bool:
+    """Opt-in gate for the cooperative split-K reduction (default OFF)."""
+    return os.environ.get("AITER_MLA_REDUCE_SPLITK", "0") == "1"
```
coderfeli: by env?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+MASSIVE_THR = 4  # kMassiveThreshold
+
+# GRP is the power-of-two pipeline depth for the M64 and M256/MLDS tiers.
+GRP_M256 = int(os.environ.get("MLA_GRP_M256", "16"))
+GRP_M64 = int(os.environ.get("MLA_GRP_M64", "8"))
+
+# A runtime M64 sub-split uses a deeper pipeline above the threshold without
+# changing CUDA-graph topology. Set the threshold to zero to disable it.
+M64_HI_THR = int(os.environ.get("MLA_M64_HI_THR", "8"))
+M64_HI_GRP = int(os.environ.get("MLA_M64_HI_GRP", "16"))
```
coderfeli: so many env?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+def _pointer_buffer_tensor(ptr, dtype, shape, stride):
+    """Build a public buffer-backed tensor view from a pointer ABI argument."""
+    return fx.rocdl.make_buffer_tensor(_pointer_view(ptr, dtype, shape, stride))
+
+
+def _pointer_buffer_resource(ptr, dtype, shape, stride):
+    """Build a raw buffer descriptor for a documented low-level MLA boundary."""
+    return buffer_ops.create_buffer_resource(
+        _pointer_view(ptr, dtype, shape, stride), max_size=True
+    )
```
coderfeli, line 175: clean buffer_ops?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+def _exp(x, use_exp2=True):
+    """exp(x) via the hardware v_exp_f32 (exp2(x*log2e))."""
+    if fx.const_expr(use_exp2):
+        return fx.rocdl.exp2(T.f32, (x * _LOG2E).ir_value())
+    return fly_math.exp(x, fastmath=fm_fast)
+
+
+def _log(x):
+    """natural log; mlir math dialect lowers to the device log."""
+    return fly_math.log(x, fastmath=fm_fast)
```
coderfeli: why need these wrapper for 1 line math call?

op_tests/test_flydsl_mla_reduce.py
```diff
@@ -0,0 +1,2838 @@
+# SPDX-License-Identifier: MIT
```
coderfeli, line 1: 3k LOCs to do the tests? can you reduce and clean useless?

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1342 @@
...
+                return [
+                    fx.Float32(
+                        buffer_ops.buffer_load(
+                            partial_output_rsrc, offset, vec_width=1, dtype=T.f32
+                        )
+                    )
+                ]
+            v = buffer_ops.buffer_load(
+                partial_output_rsrc, offset, vec_width=VEC, dtype=T.f32
+            )
```
coderfeli: https://github.com/ROCm/aiter/blob/main/.claude/skills/flydsl-kernel-code-cleanup/SKILL.md#2-buffer_ops--make_buffer_tensor--copy-atoms too many buffer_ops. need clean. It will be deprecated soon.

coderfeli (dismissed):

coderfeli (dismissed):

aiter/ops/flydsl/kernels/mla_reduce.py
```diff
@@ -0,0 +1,1359 @@
...
+def derive_actual_max_splits(reduce_indptr) -> int:
+    """Return ``max_t(reduce_indptr[t+1] - reduce_indptr[t])`` from the CSR.
+
+    MUST be called at **planning time** (outside any CUDA-graph capture region),
+    after ``get_mla_metadata_v1`` has filled ``reduce_indptr`` on device.
+    """
+    if reduce_indptr.numel() < 2:
+        return 0
+    diffs = reduce_indptr[1:] - reduce_indptr[:-1]
+    return int(diffs.max().item())
```
valarLip, line 945: D2H ?

anhminhnguyenhoang:
measuring the effect of D2H for prefill and MTP

valarLip (approved):

PR #3913, opened 2026-06-25
flydsl: gfx942 FP8 MQA logits indexer kernel (+ Triton FN/FNUZ fix)

akii96:
## Motivation

DeepSeek-V4 Flash's sparse attention uses a lightning indexer (fp8 MQA logits). This PR adds a **FlyDSL gfx942 drop-in for the Triton `fp8_mqa_logits`** that is faster on the DSv4 prefill shapes for both the mono-dtype config (q/k both e4m3fnuz) as well as the currently vllm upstream's mixed config.

The uplift does **not** depend on operand dtype: even with both operands fnuz (no conversion on either side), FlyDSL is 1.5–2.5× faster at the kernel level and cuts end-to-end prefill TTFT by up to 17% at 128K and increases with increasing ctx size.

## Technical Details

1. **New FlyDSL gfx942 kernel** which is a drop-in for the Triton one (same args, return, and `clean_logits` semantics).
2. **(Optional safety net) Triton `fp8_mqa_logits` FN/FNUZ fix** by @vpietila-amd; a no-op for fnuz/fnuz, so existing callers are unchanged. This only matters while callers still send mixed q=fn/k=fnuz (see note below); it can be dropped if maintainers prefer the source-side fix only.

FlyDSL kernel initially translated from GEAK by @peyron-amd, with perf iterated by myself and @vpietila-amd.

## Test Plan

`pytest op_tests/flydsl_tests/test_flydsl_fp8_mqa_logits.py` — validated against the torch reference and cross-checked vs the Triton kernel (`calc_diff < 1e-3`): both fp8 dtype combos

## Test Result

**335/335 pass.**

Kernel speedup (gfx942, H=64, D=128). FlyDSL wins in both operand configs:

**Mixed** q=e4m3fn / k=e4m3fnuz (current vLLM behavior):

| S_q × S_kv   | Triton  | FlyDSL  | speedup |
| ------------ | ------- | ------- | ------- |
| 4096 × 4096  | 0.47 ms | 0.21 ms | 2.2×    |
| 8192 × 8192  | 1.48 ms | 0.70 ms | 2.1×    |
| 128 × 32768  | 0.30 ms | 0.12 ms | 2.4×    |
| 671 × 131072 | 3.14 ms | 1.94 ms | 1.6×    |

**Mono fnuz/fnuz** (the production config; gfx942, H=64, D=128):

| S_q × S_kv   | Triton  | FlyDSL  | speedup |
| ------------ | ------- | ------- | ------- |
| 4096 × 4096  | 0.34 ms | 0.22 ms | 1.5×    |
| 8192 × 8192  | 1.15 ms | 0.74 ms | 1.6×    |
| 128 × 32768  | 0.29 ms | 0.12 ms | 2.5×    |
| 671 × 131072 | 3.14 ms | 1.78 ms | 1.8×    |

**End-to-end** DeepSeek-V4 Flash TTFT (TP4, MI300-class gfx942, OSL=27, concurrency=1; Triton gfx942 indexer vs FlyDSL, leveraging PR 46730 on vLLM which fixes the mixed dtype):

| ISL  | Triton    | FlyDSL   | speedup | TTFT reduction |
| ---- | --------- | -------- | ------- | -------------- |
| 32K  | 2099 ms   | 1981 ms  | 1.06×   | −5.7%          |
| 128K | 11337 ms  | 9401 ms  | 1.21×   | −17.0%         |

The E2E gain grows with context length (the indexer logits cost scales with N), which is where prefill latency hurts most.

## Note on the mixed FN/FNUZ path (re: @valarLip)

You were right to push on the mixed dtype usage. The mixed q=fn/k=fnuz inputs come from vLLM, not aiter. Currently DSv4-Flash's `fused_indexer_q.py` quantizes Q to e4m3fn while the gfx942 KV cache is fnuz. The proper fix is at the source: vllm-project/vllm#46730 makes Q fnuz so both operands match. The Triton FN/FNUZ fix here is an optional bridge for the current vLLM behavior and can be dropped if you'd rather rely on the source fix.

### DIFF
--- a/aiter/ops/flydsl/kernels/fp8_mqa_logits.py
+++ b/aiter/ops/flydsl/kernels/fp8_mqa_logits.py
@@ -121,17 +121,23 @@ def _build_kernel_mfma_r_w(
     WPB = waves_per_block
     MR_BLOCK_THREADS = 64 * WPB
 
-    assert H % 16 == 0, f"num_heads={H} must be a multiple of 16 for MFMA"
-    assert BKV % 16 == 0, f"block_kv={BKV} must be a multiple of 16 for MFMA"
-    assert D % 32 == 0, f"head_size={D} must be a multiple of 32 for MFMA K32"
+    # MFMA tile dims of the fp8 16x16x32 atom: MFMA_M x MFMA_N output tile,
+    # MFMA_K fp8 elements reduced per MFMA step.
+    MFMA_M = 16
+    MFMA_N = 16
+    MFMA_K = 32
+
+    assert H % MFMA_M == 0, f"num_heads={H} must be a multiple of MFMA_M={MFMA_M}"
+    assert BKV % MFMA_N == 0, f"block_kv={BKV} must be a multiple of MFMA_N={MFMA_N}"
+    assert D % MFMA_K == 0, f"head_size={D} must be a multiple of MFMA_K={MFMA_K}"
     assert RPB >= 1, "rows_per_block must be >= 1"
     assert WPB >= 1, "waves_per_block must be >= 1"
-    N_TILES = BKV // 16  # total column-tiles per BKV block
+    N_TILES = BKV // MFMA_N  # total column-tiles per BKV block
     assert (
         N_TILES % WPB == 0
-    ), f"BKV/16={N_TILES} must be divisible by waves_per_block={WPB}"
-    M_TILES = H // 16  # head row-tiles
-    K_STEPS = D // 32  # K32 MFMA steps over the head dim
+    ), f"BKV/MFMA_N={N_TILES} must be divisible by waves_per_block={WPB}"
+    M_TILES = H // MFMA_M  # head row-tiles
+    K_STEPS = D // MFMA_K  # MFMA K-steps over the head dim
     N_TILES_PER_WAVE = N_TILES // WPB  # column-tiles per wave
 
     fm_fast = arith.FastMathFlags.fast
@@ -175,9 +181,9 @@ def _build_kernel_mfma_r_w(
         # Decompose tid into wave index and in-wave lane.
         wave = fx.Int32(arith.divui(_to_raw(tid), _to_raw(fx.Int32(64))))
         lane = fx.Int32(arith.remui(_to_raw(tid), _to_raw(fx.Int32(64))))
-        lane_div_16 = fx.Int32(arith.divui(_to_raw(lane), _to_raw(fx.Int32(16))))
-        lane_mod_16 = fx.Int32(arith.remui(_to_raw(lane), _to_raw(fx.Int32(16))))
-        lane8 = fx.Int32(arith.muli(_to_raw(lane_div_16), _to_raw(fx.Int32(8))))
+        lane_div_N = fx.Int32(arith.divui(_to_raw(lane), _to_raw(fx.Int32(MFMA_N))))
+        lane_mod_N = fx.Int32(arith.remui(_to_raw(lane), _to_raw(fx.Int32(MFMA_N))))
+        lane8 = fx.Int32(arith.muli(_to_raw(lane_div_N), _to_raw(fx.Int32(8))))
 
         # fp8 operands are read 8 bytes at a time as 2 i32 dwords (v8i8
         # buffer_load fails to lower on gfx942), bitcast to i64 for the MFMA.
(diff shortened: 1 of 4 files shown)


valarLip:
why there is a mixed FN/FNUZ path?

akii96:
@valarLip 

The mixed types come from vllm. On gfx942 the KV cache is fnuz, but DSv4 Flash quantizes Q to fn (`fused_indexer_q.py` only emits fn), so the kernel gets q=fn / k=fnuz.

We can fix that at the source and give the Q-quant an fnuz path like the KV cache already has (I can look at it later today!)

Even with both fnuz (no conversion on either side), flydsl is faster (compared below) so my thinking is that we can still have this PR 

| fnuz/fnuz    | triton  | flydsl          |
| ------------ | ------- | --------------- |
| 4096×4096    | 0.34 ms | 0.22 ms (1.5×)  |
| 8192×8192    | 1.15 ms | 0.74 ms (1.6×)  |
| 128×32768  | 0.29 ms | 0.12 ms (2.5×) |
| 671×131072   | 3.14 ms | 1.78 ms (1.8×)  |

I am still polishing this work up but do you think the flydsl kernel improvements for the mono dtype are upstream worthy or do you have other concerns ? 😅

akii96:
@valarLip I've updated the PR description significantly to give a fuller picture of this work, could you have a second look and advise how you'd like to scope this PR?

- **flydsl kernel only**:  drop the Triton FN/FNUZ fix and rely on the vLLM source-side fix (Q→fnuz) to eliminate the mixed dtype, or
- **flydsl kernel + keep the mixed-dtype bridge**: the earlier source-side attempts upstream never merged and went stale, and the current one (vllm-project/vllm#46730) isn't guaranteed to land, so carrying mixed-dtype support in aiter short-term may be a necessary evil

Thankful for any insight you can provide here!

aiter/ops/triton/attention/fp8_mqa_logits.py
```diff
@@ -124,6 +124,19 @@ def fp8_mqa_logits(
         if seq_len <= 1024:
             matrix_instr_nonkdim = 16
 
+        _fnuz = torch.float8_e4m3fnuz
+        convert_q_fn = Q.dtype != _fnuz
+        convert_kv_fn = KV.dtype != _fnuz
+        scale_mul = 1.0
+        if convert_q_fn:
+            scale_mul *= 2.0
```
vpietila-amd, line 135: We should add 

`if arch == "gfx942":`

such that we don't accidentally convert FP8-FN on other architectures that support FP8.

aiter/ops/flydsl/kernels/fp8_mqa_logits.py
```diff
@@ -0,0 +1,676 @@
...
+    # the raw bytes through, let the kernel patch 0x80 -> +0, and undo the 2x per
+    # FN operand by scaling kv_scales -- ReLU is positive-homogeneous, so
+    # logits = sum_h ReLU(QK*scale)*w is preserved.
+    _fnuz = torch.float8_e4m3fnuz
+    _fn = torch.float8_e4m3fn
+    assert Q.dtype in (_fnuz, _fn) and KV.dtype in (
+        _fnuz,
+        _fn,
+    ), f"Q/KV must be e4m3 fp8 (fnuz or fn); got {Q.dtype}, {KV.dtype}"
+    convert_q_fn = Q.dtype != _fnuz
```
vpietila-amd: We need to check 

`if arch == "gfx942":`

to ensure that we don't convert to FP8-FNUZ on other architectures.

This kernel is optimized for `gfx942` and while it should run also on other FP8 supported architectures, the performance will probably be suboptimal as the it is not using e.g. async loads that are available in new architectures. At minimum, we should test that the current implementation works on gfx950.

aiter/ops/triton/attention/fp8_mqa_logits.py
akii96: fixed now in 99bd4d3! Thanks

aiter/ops/flydsl/kernels/fp8_mqa_logits.py
akii96: fixed now in 99bd4d3! Thanks. The decision to claify for others is to gate this kernel for now on gfx942

aiter/ops/flydsl/kernels/fp8_mqa_logits.py
```diff
@@ -0,0 +1,680 @@
...
+    small to fill the device (see ``flydsl_fp8_mqa_logits``).
+    """
+    H = num_heads
+    D = head_size
+    BKV = block_kv
+    RPB = rows_per_block
+    WPB = waves_per_block
+    MR_BLOCK_THREADS = 64 * WPB
+
+    assert H % 16 == 0, f"num_heads={H} must be a multiple of 16 for MFMA"
```
vpietila-amd: We should add here named constants 

```
MFMA_M = 16
MFMA_N = 16
MFMA_K = 32
```

And use those constants in the rest of the code. Now there too many magic number 16 and 32 floating around.

akii96: Yeah made sense! and now addressed in d0d58f00e. Added `MFMA_M`/`MFMA_N`/`MFMA_K` and used them for the tile dims everywhere, and renamed the `lane_div_16`/`lane_mod_16` vars to `lane_div_N`/`lane_mod_N` to match

vpietila-amd (dismissed):
Reviewed the implementation and the new FlyDSL kernel looks good to me.

akii96:
@valarLip we gathered more results with the new FlyDSL kernel introduced in this PR 

Outside of DSv4 Flash we also tried replacing the Triton kernel with the FlyDSL counterpart for GLM-5.2-FP8:

| Max Conc | TTFT (ms) |
|----------|-----------|
| 2 | 20389.85 → 15427.41 (**−24%**) |
| 4 | 32335.73 → 25395.11 (**−21%**) |
| 6 | 33781.63 → 27584.69 (**−18%**) |
| 8 | 39559.43 → 30936.41 (**−22%**) |
| **GeoMean** | **30637.75 → 24046.14 (−21%)** |

TTFT reduction is consistent (18–24%) across concurrency levels

Hoping to get your review on this and get this merged as we noticed similar improvements across DS3.2/GLM5.1 as well (happy to provide more results on request!)

op_tests/flydsl_tests/test_flydsl_fp8_mqa_logits.py
valarLip, line 1: please refact ut follow this one https://github.com/ROCm/aiter/tree/main/.claude/skills/aiter-op-test

valarLip (changes requested):

op_tests/flydsl_tests/test_flydsl_fp8_mqa_logits.py
akii96: Refactored in 9e8797b!

valarLip (approved):

PR #3915, opened 2026-06-25
perf(unified_attention): use 8 warps for gfx1151 3D decode

carlushuang:
## perf(unified_attention): 8 warps for gfx1151 3D decode

On gfx1151 (RDNA3.5) bf16 `unified_attention` decode was parallelism-bound at bs=1 (~31% of the LPDDR5X roofline) on the default `num_warps=2`; setting `num_warps=8` for gfx1151 in `select_3d_config` (mirrors the existing `waves_per_eu=8` tuning) reaches ~59% of roofline — 1.5–1.9× on the kernel, bitwise-identical output. Scoped to gfx1151.


yifehuan (approved):
LGTM only thing is maybe not need to show too much performance on comment, comment one or two line probably is better

PR #3919, opened 2026-06-25
feat(gfx1151): allow gfx1151 in cpp_itfs JIT arch validation

carlushuang:
## feat(gfx1151): allow gfx1151 in cpp_itfs JIT arch validation

`validate_and_update_archs()` gates every `cpp_itfs` op against a CDNA-only allow-list, so on gfx1151 the top-p/top-k sampling ops hard-assert (`One of GPU archs of ['gfx1151'] is invalid or not supported`) once `temperature > 0` — killing the engine while `/health` still returns 200; the kernels are portable HIP and run correctly on gfx1151 (verified: sampled tokens fall within the top-p nucleus), so add `gfx1151` to the allow-list.


yifehuan (approved):
lgtm

PR #3922, opened 2026-06-25
[OPUS][ATOM] gfx1250 (wave32/WMMA) NoPE-fp8/RoPE-bf16 sparse prefill (separate module)

carlushuang:
## Summary

Adds gfx1250 (RDNA-class, wave32 / WMMA) support for the split-precision NoPE-fp8 / RoPE-bf16 sparse paged prefill attention kernel (DeepSeek-V4 DSA). Targets `kaiyang-1/dsa_mxfp8_16mx1` (PR #3751) so it stacks on the fp8 kernel.

The gfx1250 kernel is a **separate C++ module** compiled only for gfx1250, reached through the **same public python API** as the gfx950 fp8 kernel: `pa_sparse_prefill_fp8_opus(...)` detects the runtime arch and dispatches to the gfx1250 backend on gfx1250, or the existing MFMA backend on gfx950. Because the modules are JIT-built lazily on first call and each device body is arch-gated (`#if __gfx950__` / `#if __gfx1250__`), the gfx950 kernel never compiles on gfx1250 and the gfx1250 kernel never compiles on gfx950.

The gfx950 C++ kernel (`pa_sparse_prefill_opus.h`), `opus.hpp`, and its launcher are **unchanged** — byte-identical to this PR's base. The bf16/fp16 single-tensor variant stays gfx950-only.

## Correctness (gfx1250 hardware, via the shared `pa_sparse_prefill_fp8_opus` API, vs fp32 reference, atol/rtol 3e-2)

`op_tests/test_pa_sparse_prefill_opus.py --prec fp8`: dense + sparse, H=16 and H=32, N up to 256 — all pass (max abs diff ≤ 3.9e-3). On gfx1250 only `module_pa_sparse_prefill_fp8_opus_gfx1250` is built.

| case | max abs diff |
|---|---|
| empty (sink-only) | 0.0 |
| dense H=16 / H=32 | 9.8e-04 |
| sparse, multi-tile + OOB | 3.9e-03 |

Correct paths: RoPE QK (bf16 WMMA), NoPE QK (fp8→bf16 dequant with E8M0 per-block scale), PV (bf16 WMMA, V transposed via LDS), wave32 online softmax, prefix+extend two-region accumulation, OOB-token masking, attn-sink finalization.

## Changes

New gfx1250 module `module_pa_sparse_prefill_fp8_opus_gfx1250`:
- `csrc/include/pa_sparse_prefill_fp8_opus_gfx1250.h` — wave32 WMMA kernel (own kargs/traits; gfx1250-only device body, empty stub elsewhere).
- `csrc/py_itfs_cu/pa_sparse_prefill_fp8_opus_gfx1250_kernels.cu` — launcher (block dim from the runtime warp size).
- `csrc/pybind/pa_sparse_prefill_fp8_opus_gfx1250_pybind.cu` — bindings.
- `aiter/ops/pa_sparse_prefill_fp8_opus_gfx1250.py` — low-level `*_fwd` binding only.
- `aiter/jit/optCompilerConfig.json` — additive module entry (`ENABLE_CK=0`; the kernel is opus-only and the in-tree CK predates gfx1250).

Shared API + test:
- `aiter/ops/pa_sparse_prefill_opus.py` — `pa_sparse_prefill_fp8_opus` accepts gfx950/gfx1250 and dispatches to the right backend.
- `op_tests/test_pa_sparse_prefill_opus.py` — allow gfx1250 for the fp8 variant (bf16/fp16 stay gfx950-only).

## Implementation notes

- opus WMMA wave32 path with auto-generated WMMA operand layouts (`mma.layout_a/b/c(stride, coord)`).
- O accumulator and online m/l state kept in LDS (gfx1250 has 320 KB/block).
- Split `s_wait_loadcnt` / `s_wait_dscnt` (no combined `s_waitcnt`); V transposed through LDS (no `ds_read_tr`).

## Follow-up (perf, not correctness)

NoPE QK currently dequantizes fp8→bf16 and runs bf16 WMMA. The throughput follow-up is native fp8 scaled-WMMA (`__builtin_amdgcn_wmma_scale_f32_16x16x128_f8f6f4` with OPSEL `a_scale_sel`/`b_scale_sel`); the gfx950 `reorder_mxscl_for_opsel` byte-selector scheme is MFMA-specific and must be reworked for the gfx1250 OPSEL semantics. The kernel is also currently non-pipelined (correctness-first).


carlushuang:
### Perf update

Landed **M1 (register-resident O accumulator + register softmax rescale)** — commit on this branch. The PV accumulator now stays in the WMMA C-fragment across all KV tiles and is rescaled in registers, removing the per-tile LDS dO store, the 8192-element scalar accumulate, and a pile of barriers.

Measured on gfx1250 (`-n 512 --total_pages 1024 --prec fp8 --mode dense`, TFLOPS H=16/64/128):

| | H=16 | H=64 | H=128 |
|---|---|---|---|
| before | 1.37 | 1.38 | 1.39 |
| **M1** | **3.00** | **3.92** | **4.08** |

~2.2–2.9×, correctness unchanged (dense/sparse/empty/OOB pass at 3e-2).

Investigated two further structural changes — a cooperative 4-wave KV split and KV double-buffering — but isolation diagnostics (removing the fp8→bf16 dequant, removing gmem KV loads, parallelizing the softmax, and cutting ~32 KB of LDS) each moved perf <5%, and perf tracks block count (H) only. So this kernel's ceiling (~4 TFLOPS at H=128) is set by the bf16-WMMA issue rate over the 512-deep head dim plus per-block fixed overhead — not by KV-split parallelism, softmax serialization, or staging bandwidth. The next real lever is **native fp8 scaled-WMMA for the NoPE QK** (16×16×128 f8f6f4 instead of bf16 16×16×32), which directly cuts that MMA work; that remains the documented follow-up.

PR #3923, opened 2026-06-25
change default pa reduce kernel from cxx to flydsl

Bernard-Liu:
## Motivation
The Gluon paged-attention decode partition-softmax (PS) reduce step previously
defaulted to the C++ (HIP) implementation, with the FlyDSL reduce kernel kept
only as a secondary fallback. This PR makes the FlyDSL reduce kernel the default
backend for the PS reduce path. Because the FlyDSL kernel had been sitting behind
the C++ path, it had never actually been compiled/executed, so promoting it to
default also required fixing several latent bugs in it.
## Technical Details
- **Backend selection reorder** (`_paged_attention_decode_v2_reduce_kernel_wrapper`,
  `PS` path): the selection order is now FlyDSL → C++ → Triton (was C++ → FlyDSL → Triton).
- **Hardened fallback chain**: each backend is now guarded by its own availability
  flag (`FLYDSL_PS_REDUCE_AVAILABLE`, `CXX_PS_REDUCE_AVAILABLE`), and the Triton
  kernel is the unconditional final fallback. Previously the second branch called
  the C++ launcher unconditionally; when the C++ extension is not built that symbol
  is `None`, so `None(...)` raised `TypeError` (not caught by `except ImportError`)
  and the Triton fallback was skipped.
- **FlyDSL compile-time branching fix** (`compile_pa_decode_ps_reduce_flydsl`):
  all compile-time-constant conditions are now wrapped with `const_expr(...)` 
## Test Plan
- `python op_tests/triton_tests/test_pa_decode_gluon.py` (full sweep, 128 cases,
  bf16, with/without sinks, PS enabled), run with the FlyDSL reduce backend as default.

### DIFF
--- a/aiter/aot/flydsl/pa_decode.py
+++ b/aiter/aot/flydsl/pa_decode.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""AOT pre-compilation for the FlyDSL PS-reduce kernel used by pa_decode_gluon.
+
+The set of compile-time constant combinations to precompile is hand-specified as
+a cartesian product of axes in ``PA_DECODE_AOT_SPEC``. Each combination is
+compiled under ``COMPILE_ONLY=1`` with CPU dummy tensors, calling the same
+``compile_pa_decode_ps_reduce_flydsl`` builder the runtime uses, so the AOT cache
+key equals the runtime lookup.
+
+Runs on a GPU host (importing pa_decode_gluon needs a GPU); arch is auto-detected
+by FlyDSL and matches the runtime arch on the same host.
+
+Usage:
+    python -m aiter.aot.flydsl.pa_decode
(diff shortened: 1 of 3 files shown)


aiter/aot/flydsl/pa_decode.py
```diff
@@ -0,0 +1,210 @@
...
+    FLYDSL_PS_REDUCE_AVAILABLE,
+    PA_DECODE_MAX_SPLITS,
+    _flydsl_dtype_str,
+    compile_pa_decode_ps_reduce_flydsl,
+)
+
+# Hand-edited cartesian-product spec. Replace these starter values with the real
+# target model configs. Axes map 1:1 onto the kernel's compile-time constants
+# (sink_dtype is pinned to output_dtype when use_sinks is False; see _expand_spec).
+PA_DECODE_AOT_SPEC = {
```
Bernard-Liu, line 38: @PerryZhang01 please help check, could this config include all models' shape that reduce kernel would use.

PerryZhang01: confirmed

PR #3928, opened 2026-06-25
fix(quick_all_reduce): make flag sync CUDA-graph-safe

Jasen2201:
# fix(quick_all_reduce): make cross-rank flag sync CUDA-graph-safe

## Summary

quick-reduce's cross-rank synchronization is **not CUDA-graph-safe**. It passes
the round counter `flag_color` as a **host-side scalar kernel argument** and
advances it on the host after each call. Under CUDA-graph capture that value is
**baked into the captured graph** and reused on every replay (the host increment
never runs during replay), while the `block_id`-keyed flags buffer is **never
reset between replays** and `wait_sync_flag` uses **exact equality**.

So when a captured quick-reduce is the **sole writer of its flag slot**, on the
next replay the waiter finds the slot already holding the (baked) value it is
waiting for — left there by the *previous* replay — and its wait returns
immediately, **before the peer has written this round's data**. It then reads
**stale comm-buffer data from the previous replay** → silent corruption of the
all-reduce result.

The sibling `custom_all_reduce` does not have this bug because its flag lives in
**device memory** and is incremented **on the device** each launch
(`self_sg->_flag[blockIdx.x] + 1`), so every replay uses a fresh value.

## Fix

Mirror `custom_all_reduce`: keep a **per-block flag-color counter in device
memory** (`uint32_t* d_flag_color`, sized `kMaxNumBlocks`, initialized to 1).
The kernel reads its block's counter at launch and writes back the advanced
value, so every (replay) launch uses a fresh color. The host-side `flag_color`
member and host-side rotation are removed. The eager (non-graph) path is
functionally unchanged; it now also advances the color on the device.

## Reproducer (CUDA-graph unit test)

Captures a CUDA graph containing a **single** quick-reduce, replays it 8 times
with a different known input each step, and checks the output equals the true
cross-rank sum. Lossless FP regime + bf16 small integers, so the reduce is exact
and any mismatch is the bug.

```bash
# inside the aiter container, on a 4-GPU node:
torchrun --standalone --nproc_per_node=4 qr_cudagraph_repro.py
# exit 0 = correct (fixed); exit 1 = stale-flag corruption (bug)
```

```python
import os
# quick-reduce module-level init reads these at import time.
os.environ.setdefault("AITER_QUICK_REDUCE_QUANTIZATION", "FP")        # lossless
os.environ.setdefault("AITER_QUICK_REDUCE_CAST_BF16_TO_FP16", "0")

import torch
import torch.distributed as dist
from aiter.dist.device_communicators.quick_all_reduce import QuickAllReduce


def main():
    dist.init_process_group(backend="gloo")          # CPU group for IPC-handle exchange
    rank, world = dist.get_rank(), dist.get_world_size()
    torch.cuda.set_device(rank)
    dev = torch.device(f"cuda:{rank}")

    qr = QuickAllReduce(group=dist.group.WORLD, device=dev)
    assert not qr.disabled, "quick-reduce unavailable on this arch/env"

    N = 1 << 21                                       # 2M bf16 = 4 MB, above any qr threshold
    inp = torch.empty(N, dtype=torch.bfloat16, device=dev)
    out = torch.empty(N, dtype=torch.bfloat16, device=dev)

    # Step s: rank r contributes (r + 1 + s). True sum over ranks:
    #   sum_r (r + 1 + s) = world*s + world*(world+1)/2   (exact in bf16 for small ints)
    def expected(s):
        return float(world * s + world * (world + 1) // 2)

    # Warmup, then capture a graph with EXACTLY ONE quick-reduce (isolated qr).
    inp.fill_(float(rank + 1))
    qr.quick_all_reduce(inp, out=out)
    torch.cuda.synchronize(); dist.barrier()

    g = torch.cuda.CUDAGraph()
    with torch.cuda.graph(g):
        qr.quick_all_reduce(inp, out=out)
    torch.cuda.synchronize(); dist.barrier()

    bad = 0
    for s in range(8):
        inp.fill_(float(rank + 1 + s))               # new input for this step (in-place)
        dist.barrier()
        g.replay()
        torch.cuda.synchronize()
        ref = torch.full_like(out, expected(s), dtype=torch.float32)
        got = out.float()
        n_wrong = int((got != ref).sum().item())
        if n_wrong:
            bad += 1
        if rank == 0:
            print(f"  step {s}: expected={expected(s):.0f} "
                  f"got[min..max]={got.min():.0f}..{got.max():.0f} "
                  f"wrong={n_wrong}/{N} {'OK' if not n_wrong else 'CORRUPT'}", flush=True)

    t = torch.tensor([bad]); dist.all_reduce(t, op=dist.ReduceOp.SUM)
    if rank == 0:
        print("\nRESULT:", "PASS" if t.item() == 0 else "FAIL", flush=True)
    dist.destroy_process_group()
    raise SystemExit(0 if t.item() == 0 else 1)


if __name__ == "__main__":
    main()
```

### Before fix — `torchrun --nproc_per_node=4 qr_cudagraph_repro.py`

```
[repro] world_size=4 elems=2097152 regime=FP(lossless) bf16
[repro][rank0]
  step 0: expected=10 got[min..max]=10..10 wrong=0/2097152 OK
  step 1: expected=14 got[min..max]=10..14 wrong=1946208/2097152 CORRUPT
  step 2: expected=18 got[min..max]=11..16 wrong=2097152/2097152 CORRUPT
  step 3: expected=22 got[min..max]=15..22 wrong=1228992/2097152 CORRUPT
  step 4: expected=26 got[min..max]=23..26 wrong=1572864/2097152 CORRUPT
  step 5: expected=30 got[min..max]=23..30 wrong=2061408/2097152 CORRUPT
  step 6: expected=34 got[min..max]=27..34 wrong=2045696/2097152 CORRUPT
  step 7: expected=38 got[min..max]=35..38 wrong=1480288/2097152 CORRUPT
[repro] RESULT: FAIL — 28 corrupt (step,rank) cases; captured quick-reduce read stale data on replay
```

Step 0 is correct because its input equals the captured/warmup input; from step 1
on (once the input changes) the output contains values from *earlier* steps
(e.g. step 1 expects 14 but yields values down to 10) — the captured qr read
the previous replay's stale comm-buffer data. 28 = 7 bad steps × 4 ranks.

### After fix (same test, only `module_quick_all_reduce` rebuilt, ~23s)

```
[repro] world_size=4 elems=2097152 regime=FP(lossless) bf16
[repro][rank0]
  step 0: expected=10 got[min..max]=10..10 wrong=0/2097152 OK
  step 1: expected=14 got[min..max]=14..14 wrong=0/2097152 OK
  step 2: expected=18 got[min..max]=18..18 wrong=0/2097152 OK
  step 3: expected=22 got[min..max]=22..22 wrong=0/2097152 OK
  step 4: expected=26 got[min..max]=26..26 wrong=0/2097152 OK
  step 5: expected=30 got[min..max]=30..30 wrong=0/2097152 OK
  step 6: expected=34 got[min..max]=34..34 wrong=0/2097152 OK
  step 7: expected=38 got[min..max]=38..38 wrong=0/2097152 OK
[repro] RESULT: PASS — all replays correct (no stale-flag corruption)
```
## Performance

The fix adds one `uint32` device load + store of a per-block counter per kernel
launch — no extra traffic or sync. `qr_all_reduce` latency (4×MI300, INT4,
mean of 2 runs each):

| size | before (µs) | after (µs) | Δ |
|---|---|---|---|
| 1 MB | 21.29 | 21.27 | −0.1% |
| 2 MB | 23.15 | 23.04 | −0.5% |
| 3 MB (c256 decode) | 25.97 | 26.05 | +0.3% |
| 4 MB | 29.18 | 29.21 | +0.1% |
| 8 MB | 43.47 | 43.66 | +0.4% |
| 16 MB | 70.32 | 70.07 | −0.4% |
| 32 MB | 124.70 | 124.56 | −0.1% |
| 64 MB | 234.60 | 236.30 | +0.7% |

No measurable regression: all deltas are within run-to-run noise (the two
before-fix runs themselves differ by up to ~2%), with no systematic slowdown.
Bandwidth is unchanged (~120 GB/s at 3 MB, ~285 GB/s at 64 MB).

### DIFF
--- a/csrc/include/quick_all_reduce.cuh
+++ b/csrc/include/quick_all_reduce.cuh
@@ -792,7 +792,8 @@ allreduce_prototype_twoshot(T const* A,
         block += grid;
         flag_color++;
     }
-    d_flag_color[blockIdx.x] = flag_color;
+    if (threadIdx.x == 0 && threadIdx.y == 0)
+        d_flag_color[blockIdx.x] = flag_color;
 }
 
 #define TWOSHOT_DISPATCH(__codec)                                             \
--- a/csrc/include/quick_all_reduce.cuh
+++ b/csrc/include/quick_all_reduce.cuh
@@ -775,11 +775,15 @@ allreduce_prototype_twoshot(T const* A,
                             int rank,
                             uint8_t** dbuffer_list,
                             uint32_t data_offset,
-                            uint32_t flag_color,
+                            uint32_t* d_flag_color,
                             int64_t data_size_per_phase)
 {
     int block = blockIdx.x;
     int grid  = gridDim.x;
+
+    // Per-block flag color from device memory, advanced on-device so each
+    // CUDA-graph replay uses a fresh color (a host scalar would be baked in).
+    uint32_t flag_color = d_flag_color[blockIdx.x];
 
     while(block < num_blocks)
     {
@@ -788,6 +792,7 @@ allreduce_prototype_twoshot(T const* A,
         block += grid;
         flag_color++;
     }
+    d_flag_color[blockIdx.x] = flag_color;
 }
 
 #define TWOSHOT_DISPATCH(__codec)                                             \
@@ -807,7 +812,7 @@ allreduce_prototype_twoshot(T const* A,
                            rank,                                              \
                            dbuffer_list,                                      \
                            data_offset,                                       \
-                           flag_color,                                        \
+                           d_flag_color,                                        \
                            this->kMaxProblemSize);                            \
     }                                                                         \
     else if(world_size == 4)                                                  \
@@ -826,7 +831,7 @@ allreduce_prototype_twoshot(T const* A,
                            rank,                                              \
                            dbuffer_list,                                      \
                            data_offset,                                       \
-                           flag_color,                                        \
+                           d_flag_color,                                        \
                            this->kMaxProblemSize);                            \
     }                                                                         \
     else if(world_size == 8)                                                  \
@@ -845,7 +850,7 @@ allreduce_prototype_twoshot(T const* A,
                            rank,                                              \
                            dbuffer_list,                                      \
                            data_offset,                                       \
-                           flag_color,                                        \
+                           d_flag_color,                                        \
                            this->kMaxProblemSize);                            \
     }
 
@@ -866,7 +871,7 @@ struct DeviceComms
     static int constexpr kMaxWorldSize = 8;
 
     bool initialized    = false;
-    uint32_t flag_color = 1;
+    uint32_t* d_flag_color = nullptr;
     int world_size;
     int rank;
 
@@ -900,6 +905,17 @@ struct DeviceComms
         // Clear the flags buffer.
         HIP_CHECK(hipMemset(dbuffer, 0, flags_buffer_size));
 
+        // Per-block flag color counter (device-side; see kernel). Init to 1,
+        // since 0 would collide with the just-memset'd flags buffer.
+        HIP_CHECK(hipMalloc(&d_flag_color, kMaxNumBlocks * sizeof(uint32_t)));
+        {
+            std::vector<uint32_t> init_color(kMaxNumBlocks, 1u);
+            HIP_CHECK(hipMemcpy(d_flag_color,
+                                init_color.data(),
+                                kMaxNumBlocks * sizeof(uint32_t),
+                                hipMemcpyHostToDevice));
+        }
+
         // Device-side list of IPC buffers.
         buffer_list.resize(world_size);
         HIP_CHECK(hipMalloc(&dbuffer_list, world_size * sizeof(uint8_t*)));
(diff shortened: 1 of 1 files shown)


csrc/include/quick_all_reduce.cuh
```diff
@@ -775,19 +775,24 @@ allreduce_prototype_twoshot(T const* A,
...
+    uint32_t flag_color = d_flag_color[blockIdx.x];
+
     while(block < num_blocks)
     {
         AllReduceKernel::run(
             A, B, N, block, rank, dbuffer_list, data_offset, flag_color, data_size_per_phase);
         block += grid;
         flag_color++;
     }
+    d_flag_color[blockIdx.x] = flag_color;
```
haoyangli0109: We can add `if (threadIdx.x == 0 && threadIdx.y == 0)`.
The flag applies to the entire block, so modifying it in a single thread is sufficient.

haoyangli0109 (commented):
Thank you very much for the fix; I left a comment.
I'll apply this fix to vllm and sglang later.

haoyangli0109 (approved):

valarLip (approved):

PR #3934, opened 2026-06-25
[test] test_topk_plain: parametrize sweep to fix collection-time OOM

JohnQinAMD:
## Problem
CI runs each op test as a **script** (`python3 <file>` in `.github/scripts/aiter_test.sh`), not via pytest. `test_topk_plain.py` sweeps 84 cases up to `3072 × 131072` fp32 and never frees tensors between cases, so on a memory-pressured gfx950 runner it OOMs mid-sweep and the whole file exits non-zero — producing intermittent "Standard Tests (MI35X, 8, 5)" failures unrelated to the code under test. On a clean GPU it passes; the failure is purely residual-memory dependent.

## Fix (keeps the `python3 <file>` contract)
- **Free each case's tensors** (`del` + `gc.collect()` + `torch.cuda.empty_cache()`) so peak memory is one case, not the whole sweep — the actual OOM fix.
- **Guard the sweep under `if __name__ == "__main__":`** so import is side-effect-free.
- **`num_iters` 1000 → 100** — this is a correctness check, not a perf gate; the high iteration count made it the slowest file in its shard for no benefit.
- **Vectorize the per-row permutation** (drops a `batch_size`-long Python loop).
- **Assert on `checkAllclose`'s returned error ratio** — it does *not* raise, so the test previously would have passed even if `topk_plain` were incorrect.
- **Summary table falls back to `df.to_string`** when `tabulate` is absent, so the informational summary can never fail the run.

## Validation
`python3 op_tests/test_topk_plain.py` on MI355X (gfx950): all 84 cases run, all `passed~` (err 0), summary prints, exit 0.

Addresses both Copilot review comments (script-style no-op under `python3`; assert on the error ratio).

## Submission Checklist

### DIFF
--- a/op_tests/test_topk_plain.py
+++ b/op_tests/test_topk_plain.py
@@ -1,84 +1,51 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
 
+import gc
+
+import pandas as pd
 import torch
-from aiter.test_common import (
-    checkAllclose,
-    benchmark,
-    run_perftest,
-)
+
 import aiter
 from aiter import dtypes
 from aiter.ops.topk_plain import topk_plain
-import pandas as pd
+from aiter.test_common import benchmark, checkAllclose, run_perftest
 
 torch.set_default_device("cuda")
 torch.set_printoptions(sci_mode=False)
 
+# Correctness sweep: a few timed iters is plenty (was 1000). This is a
+# correctness check, not a perf gate; the high iter count made the file the
+# slowest in its CI shard for no benefit.
+NUM_ITERS = 100
+NUM_WARMUP = 10
+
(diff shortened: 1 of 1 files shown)


zufayu (approved):
LGTM — the change is a test quality fix (OOM prevention, add assert on error ratio, reduce iter count). The single CI failure () is a pre-existing issue on main (same  in , unrelated to this PR).

zufayu (approved):
LGTM. The change is a test quality fix: OOM prevention, adds assert on error ratio, reduces iter count. The single CI failure (Standard Tests MI35X shard 6) is a pre-existing issue on main — same RuntimeError in test_moe_2stage.py, unrelated to this PR.

zufayu:
@valarLip Could you merge this when you get a chance? The CI failure (Standard Tests MI35X shard 6) is a pre-existing issue on main branch (RuntimeError in test_moe_2stage.py, same failure exists on main today), unrelated to this PR's changes.

PR #3936, opened 2026-06-25
Spatial Attention: XCD-aware spatial workgroup mapping for MHA and GQA (SWIZZLE=1)

mc186:
# Spatial Attention: XCD-aware spatial workgroup mapping for MHA and GQA (SWIZZLE=1)

## Motivation

AMD CDNA3/3.5 GPUs (MI300X, MI350X, MI355X) are built from **8 XCDs** (chiplets), each with a dedicated **4 MB L2 cache**. The hardware assigns workgroup `wid` to an XCD by `wid % NUM_XCDS` — round-robin, determ

In standard attention scheduling (SWIZZLE=0), consecutive workgroups cycle through all Q heads before advancing the sequence block. For **MHA with many heads**, this causes each head's KV data to be spread acros

**Spatial scheduling (SWIZZLE=1)** re-orders workgroups so that all computation touching a given KV head is pinned to one XCD. Each head's KV tensor stays hot in a single 4 MB slice rather than being replicated

**Why GQA is unaffected:** models with 8 KV heads (e.g. Llama3, GQA ratio 16) already achieve this locality "accidentally" — round-robin workgroup dispatch across 8 XCDs naturally maps one KV head per XCD. Spati

### `aiter/ops/triton/utils/_triton/pid_preprocessing.py`

  Adds `remap_workgroup_spatial(wid, NUM_Q_HEADS, NUM_BLOCKS, BATCH, NUM_QUERIES_PER_KV, NUM_XCDS)`, a unified Triton JIT function with two specialised paths selected at compile time by `NUM_QUERIES_PER_KV`:

  - **MHA path** (`NUM_QUERIES_PER_KV == 1`): groups `ceil(NUM_Q_HEADS / NUM_XCDS)` consecutive Q heads onto each XCD. Ordering within each XCD is head-first (all blocks for head *i* before head *i+1*).

  - **GQA path** (`NUM_QUERIES_PER_KV > 1`): assigns each KV head exclusively to one XCD so that all Q heads sharing a KV head run on the same XCD. Ordering within each XCD is block-first (all Q heads in the
  group process block *b* together before advancing to *b+1*), which avoids the causal load-imbalance that head-first ordering creates. Three sub-regimes are handled: `HK == NXCD` (aligned), `HK > NXCD`, and `HK
  < NXCD`.

  ### `aiter/ops/triton/_triton_kernels/attention/mha.py`

  - Adds `SWIZZLE: tl.constexpr` kernel parameter (string type, values: `"default"` or `"spatial"`).
  - Includes `tl.static_assert` to validate SWIZZLE value at compile time with descriptive error message.
  - `SWIZZLE="default"`: preserves the existing `remap_xcd` behaviour (backward-compatible default).
  - `SWIZZLE="spatial"`: calls `remap_workgroup_spatial`; `NUM_QUERIES_PER_KV` is derived from `NUM_Q_HEADS // NUM_K_HEADS` at compile time, so GQA and MHA specialise independently with zero runtime overhead.

  ### `aiter/ops/triton/attention/mha.py`

  - Reads `AITER_TRITON_MHA_SWIZZLE` environment variable at module import (default `"default"`).
  - Adds `mha_set_swizzle(value: Literal["default", "spatial"])` for programmatic control.
  - Invalid values raise `ValueError` at import or call time.

  ## Usage

  ```bash
  # environment variable (affects all subsequent flash_attn_func calls)
  AITER_TRITON_MHA_SWIZZLE=spatial python your_script.py

  # programmatic
  from aiter.ops.triton.attention.mha import mha_set_swizzle
  mha_set_swizzle("spatial")   # enable spatial
  mha_set_swizzle("default")   # restore default
```

### MHA — SWIZZLE=0 vs SWIZZLE=1, non-causal

| Heads (HQ=HK) | Seq len | SWIZZLE=0 (TFLOPS) | SWIZZLE=1 (TFLOPS) | Delta |
| 128 | 32K | 681 | 924 | **+35.7%** |
| 128 | 64K | 636 | 895 | **+40.7%** |
| 128 | 128K | 605 | 850 | **+40.5%** |
| 64 | 64K | 742 | 811 | **+9.2%** |
| 16 | 64K | 823 | 831 | +1.0% |

## End-to-end model results (AMD MI355X, BF16, batch=1, random-init weights)

Whole-model forward-pass latency and socket energy (idle-baseline subtracted) on representative architectures:

| Model | Heads | Seq | Speedup | Energy savings |
| --- | ---: | ---: | ---: | ---: |
| Wan2.2-14B DiT (video diffusion, non-causal) | 40 | 32K | +4.0% | +4.1% |
| Wan2.2-14B DiT | 40 | 64K | +14.3% | +16.7% |
| Wan2.2-14B DiT | 40 | 131K | **+20.3%** | **+23.3%** |
| Dense MHA 128h (causal) | 128 | 128K | +10.0% | +10.6% |
| Dense MHA 128h (causal) | 128 | 512K | **+18.0%** | **+19.8%** |
| Llama3-8B GQA (negative control) | 32Q/8KV | 32K | +/-0% | +/-0% |

## Correctness

Verified bit-identical output (`rel_max = 0.00e+00`) vs SWIZZLE=0 across 17 configurations on MI355X (thor-4):

- GQA: HK in {2, 4, 8, 16, 32}, causal + non-causal, batch 1 and 4
- MHA: HQ=HK in {16, 32, 40, 64, 128}, causal + non-causal, batch 1 and 2

Test script: `op_tests/test_mha_spatial_swizzle.py` (included in this branch).

### CODE REVIEW

File: aiter/ops/triton/_triton_kernels/attention/mha.py [hunk 1]
```diff
@@ -371,10 +372,18 @@ def _attn_fwd(
     )  # workgroup id ranging: 0,1,2,...., (BATCH * NUM_Q_HEADS * NUM_BLOCKS - 1)
     # num blocks along seqlen
 
-    off_q_head = wid % NUM_Q_HEADS
-    off_q_head = remap_xcd(off_q_head, NUM_Q_HEADS, NUM_XCD)
-    start_m = (wid // NUM_Q_HEADS) % NUM_BLOCKS
-    off_z = (wid // (NUM_BLOCKS * NUM_Q_HEADS)) % BATCH
+    if SWIZZLE == 0:
```
brunomazzottiamd: Can you please add a [`tl.static_assert`](https://triton-lang.org/main/python-api/generated/triton.language.static_assert.html#triton.language.static_assert) statement to guarantee that `SWIZZLE` value belongs to the set of valid values? A nice message in `tl.static_assert` also helps debugging efforts.

Personally, I'd prefer to have `SWIZZLE` specified as a more descriptive string instead of a binary value in `{0, 1}`. Right now I can think about `{"default", "spatial"}`. We can use strings as `tl.constexpr`.

aiter/ops/triton/attention/mha.py
```diff
@@ -47,6 +48,15 @@ def mha_set_use_int64_strides(value: bool):
     _USE_INT64_STRIDES = value
 
 
+_AITER_SWIZZLE = int(os.environ.get("AITER_SWIZZLE", "0"))
```
brunomazzottiamd: Two issues here:

1. `AITER_SWIZZLE` is an overly generic name for an environment variable. AITER is big, it's better to be more specific. I can think about something like `AITER_TRITON_MHA_SWIZZLE`.
2. Please be more defensive when reading environment variables. You shouldn't trust external input that much. `int(os.environ.get("AITER_SWIZZLE", "0"))` will throw `ValueError: invalid literal for int() with base 10: 'HelloWorld'` if you `export AITER_SWIZZLE='HelloWorld'`. Craft a more robust logic to read the environment variable, you can add some assertions if it makes sense.

As I said before, I'd prefer to have descriptive strings for this configuration instead of `{0, 1}`.

aiter/ops/triton/attention/mha.py
```diff
@@ -47,6 +48,15 @@ def mha_set_use_int64_strides(value: bool):
...
+
+
+def mha_set_swizzle(value: int):
+    """Set spatial swizzle mode: 0 = baseline (remap_xcd), 1 = spatial MHA/GQA."""
```
brunomazzottiamd: Please assert that `value` is in the allowed set of values before mutating `_AITER_SWIZZLE`.

aiter/ops/triton/utils/_triton/pid_preprocessing.py
```diff
@@ -34,8 +33,7 @@ def remap_xcd(pid, GRID_MN, NUM_XCDS: tl.constexpr = 8):
     # We calculate the number of xcds that have pids_per_xcd pids as
     # tall_xcds
     tall_xcds = GRID_MN % NUM_XCDS
-    if tall_xcds == 0:
-        tall_xcds = tl.cast(NUM_XCDS, tall_xcds.type)
+    tall_xcds = NUM_XCDS if tall_xcds == 0 else tall_xcds
```
brunomazzottiamd: **[Blocker IMHO]** Please revert back this change. I did it for the sake of a grouped GEMM kernel that has to deal with big inputs and require `int64` offsets to address them. Some computations propagate `int64` values and `tall_xcds = NUM_XCDS if tall_xcds == 0 else tall_xcds` was failing in this case (mixing `int32` and `int64` in the `if ... else ...` assignment expression).

op_tests/test_mha_spatial_swizzle.py
```diff
@@ -0,0 +1,103 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""
+Correctness tests for remap_workgroup_spatial (AITER_SWIZZLE=1).
```
brunomazzottiamd: Please update the top-level documentation before merging:

* new name for environment variable / configuration
* possible values for the configuration if you decide to change from `{0, 1}` to more descriptive names

op_tests/test_mha_spatial_swizzle.py
```diff
@@ -0,0 +1,103 @@
...
+import math
+
+import pytest
+import torch
+
+import aiter.ops.triton.attention.mha as mha_mod
+
+
+def _run(swizzle: int, B: int, Hq: int, Hk: int, S: int, D: int, causal: bool):
+    mha_mod.mha_set_swizzle(swizzle)
```
brunomazzottiamd: **[Blocker]**

You are setting global state for this test suite and leaving it mutated. This is a bad practice and will affect other test suites that run after `op_tests/test_mha_spatial_swizzle.py`. Everything might be fine now, but this could have consequences in the future that will be difficult to debug.

You need to create a test fixture that mutates the global state before the test and then retore global state after the test. Something like this: 1) read current global state and save it; 2) mutate global state to fulfill your requirement; 3) run the test; 4) restore global state back to the original value.

We did this in the past in `op_tests/triton_tests/attention/test_mha_dao_ai.py`, check this test fixture:

```python
@pytest.fixture
def dao_ai_impl():
    """Set mha impl to dao_ai for the test, restore to default on cleanup."""
    mha_set_impl("dao_ai")
    yield
    mha_set_impl("default")
```

brunomazzottiamd (changes requested):
Hi @mc186! Thanks for you PR!

Code looks good but I have two blockers:

1. change in pid preprocessing function that might affect grouped GEMM kernel
2. global state mutation in unit test

Please check my comments.

I've also left some suggestions regarding the validation of MHA swizzle modes and an attempt to improve code understanding with strings instead of 0 and 1.

op_tests/test_mha_spatial_swizzle.py
brunomazzottiamd: Now I see that the tests exercise both swizzle options. A good fixture would be: 1) read current global state and save it; 2) let the test run setting whatever it wishes; 3) restore global state back to the original value.

mc186:
All feedback addressed in cb2a345:
  - Reverted tall_xcds ternary to tl.cast() (blocker - grouped GEMM compatibility)
  - Added autouse fixture to save/restore _MHA_SWIZZLE state (blocker - global state mutation)
  - Changed SWIZZLE from {0,1} to {"default","spatial"} throughout
  - Added tl.static_assert with descriptive message in kernel
  - Renamed env var to AITER_TRITON_MHA_SWIZZLE with hardened parsing
  - Added validation in mha_set_swizzle() before state mutation
  - Updated test docstring to reflect new API
 
All tests pass (22/22), rebased on current main.

brunomazzottiamd:
Hello @mc186. Why did you close this PR? Do you still want to merge it?

mc186:
I did not mean to close it! I might have hit close instead of comment...

brunomazzottiamd (dismissed):
LGTM! Let's get CI green and then merge it.

op_tests/triton_tests/attention/test_mha_spatial_swizzle.py
```diff
@@ -0,0 +1,114 @@
+# SPDX-License-Identifier: MIT
```
brunomazzottiamd, line 1: @mc186, can you please move `op_tests/test_mha_spatial_swizzle.py` to  `op_tests/triton_tests/attention` directory? It's for the sake or organization, to keep Triton tests grouped.

cagrikymk (dismissed):
LGTM!

op_tests/triton_tests/attention/test_mha_spatial_swizzle.py
mc186: Done!

azaidy (dismissed):
LGTM based on @cagrikymk's earlier approval

mc186:
added formatting fixes for failing Black test

cagrikymk:
@mc186 
Can you resolve the merge conflicts? Besides that, LGTM!

mc186:
resolved conflicts

cagrikymk (dismissed):

cagrikymk (approved):

azaidy (approved):
LGTM!

PR #3939, opened 2026-06-25
Map top-left map to bottom-right for self-attn

Micky774:
## Motivation

gfx1250 only has kernels which support bottom-right masking. In the case of `s_q=s_kv` we have that bottom-right = top-left masking. This mirrors a similar symmetry already in place for gfx942.

## Test Plan

Tested by building w/ TE and applying to E2E models.

## Test Result

The E2E model succeeds.

## Submission Checklist


Micky774:
@yzhou103 gentle ping to see if this PR could get a review -- thanks!

PR #3958, opened 2026-06-26
[Feature] fp8 (e4m3) unified-KV prefill variant for pa_sparse_prefill_opus (gfx950)

Raiden-Makoto:
## Summary
Adds an optional fp8 (e4m3, OCP) input path to the existing
`pa_sparse_prefill_opus` paged-prefix prefill kernel on gfx950. When the caller
supplies per-`1xGROUP` fp32 `kv_scales`, the paged `unified_kv` prefix is read as
fp8 and dequantized **on-read into registers** right before the MFMA
(`bf16 = fp8 * scale`); no bf16 tile is materialized. The bf16 path is the
default and is unchanged.

## Motivation
DeepSeek-V4 on gfx950 can run an fp8 unified-KV cache (~2x KV capacity), but the
sparse-prefill read path only accepted bf16, forcing a dequant-to-bf16
materialization before this kernel. This PR lets `pa_sparse_prefill_opus` consume
the fp8 layout directly so the capacity win isn't paid back as a bf16 copy. This
is a **capability addition, not a perf optimization** — see Performance below.

## Changes
- `csrc/include/pa_sparse_prefill_opus.h`: fp8 prefix LDS layout + dequant-on-read,
  compile-time gated via `IS_FP8_SRC` for both kernel variants (`16mx1_16nx4` for
  H<=32, `16mx8_32nx1`). Decodes each fp8 dword with
  `__builtin_amdgcn_cvt_pk_f32_fp8` directly on the `i32` load (no `fp8x4_t`/`u8`
  ext-vector bit_casts in the hot path).
- `csrc/py_itfs_cu/pa_sparse_prefill_opus_kernels.cu`: fp8 launch wiring.
- `csrc/include/rocm_ops.hpp`: op registration (adds optional `kv_scales`).
- `aiter/ops/pa_sparse_prefill_opus.py`: dispatch — fp8 when `kv_scales` is given,
  bf16 otherwise.
- `op_tests/test_pa_sparse_prefill_opus.py`: fp8 correctness cases.

**Zero impact on existing callers:** when `kv_scales=None`, `IS_FP8_SRC` is false,
none of the fp8 code is instantiated, and the bf16 path codegens byte-identically.

## Performance (gfx950 / MI355X)
This is not a speedup PR. In isolation the fp8 kernel is at parity-to-slightly
faster than bf16 at small head counts; the value is enabling the fp8 KV layout,
not raw latency.

| Variant | fp8 vs bf16 (kernel latency, microbench) |
|---|---|
| H<=32 | ~0.73–0.78x (fp8 ~22–27% faster) |
| H=128 (4x32-head launches) | ~3x slower than one 128-head bf16 launch |

End-to-end fp8-KV overhead in a serving pipeline lives in the caller's
quantize-on-store path, not in this kernel, and is out of scope here. aiter
kernels are largely memory-bound, so consuming fp8 directly (vs a bf16
materialization) is the bandwidth-correct path.

## Testing
- [x] `op_tests/test_pa_sparse_prefill_opus.py` — 44 passed (24 bf16 + 20 fp8) on gfx950 (MI355X).
- [x] fp8 path reproduces the dequantized reference (rtol=atol=2e-2; observed max abs err within fp8 quant error).
- [x] bf16 path unchanged (same cases pass pre/post).
- [ ] Upstream CI (gfx942 + gfx950 matrix). Note: kernel is gfx950-gated; the gfx942 leg exercises only the unchanged bf16 path.

## Dependencies
- [x] No new third-party dependencies.

## Breaking Changes
None. `kv_scales` is optional; omitting it preserves the exact current behavior.

### DIFF
--- a/aiter/ops/pa_sparse_prefill_opus.py
+++ b/aiter/ops/pa_sparse_prefill_opus.py
@@ -45,6 +45,7 @@ def pa_sparse_prefill_opus_fwd(
     attn_sink: torch.Tensor,
     out: torch.Tensor,
     softmax_scale: float,
+    kv_scales: Optional[torch.Tensor] = None,
 ) -> None: ...
 
 
@@ -59,6 +60,7 @@ def _pa_sparse_prefill_opus_fake(
     attn_sink: torch.Tensor,
     softmax_scale: float,
     out: Optional[torch.Tensor] = None,
+    kv_scales: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     return out if out is not None else torch.empty_like(q)
 
@@ -75,6 +77,7 @@ def pa_sparse_prefill_opus(
     attn_sink: torch.Tensor,
     softmax_scale: float,
     out: Optional[torch.Tensor] = None,
+    kv_scales: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     """Sparse prefill attention over two KV sources (paged ``unified_kv`` +
     flat per-fwd ``kv``), backed by the OPUS gfx950 HIP kernel.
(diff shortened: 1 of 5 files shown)


Raiden-Makoto:
Closing this in favor of #3751 ([OPUS][ATOM] NoPE-fp8 / RoPE-bf16 sparse prefill for DSv4 on gfx950), which already landed an fp8 sparse-prefill OPUS path owned by the ATOM team. Their NoPE-fp8/RoPE-bf16 design (true fp8 MFMA on the NoPE dot, RoPE kept in bf16) supersedes this PR's full-fp8 `kv_scales` variant. We'll adopt `pa_sparse_prefill_fp8_opus` on the consumer side instead. Thanks!

PR #3959, opened 2026-06-26
[Kernel][Triton] sliding-window decode over shuffled fp8 paged KV

reger-men:
## Summary

Adds `aiter/ops/triton/attention/pa_decode_shuffle.py`, a paged decode kernel
that reads the shuffled fp8 KV layout used by the asm `pa_fwd` / `paged_attention_common`
path and applies a sliding-window mask.

## Motivation

The shuffled fp8 layout is written by `reshape_and_cache_shuffle` and consumed at
decode by the asm paged-attention entry point, which takes no window argument. So a
model with sliding-window attention can't use the shuffled cache: the only windowed
fp8 decode is the NHD `paged_attention_v1` path, which uses a different (non-shuffled)
layout. There was no kernel that both reads the shuffled layout and honors a window,
so SWA layers either fall back to NHD or silently compute full attention. This is the
correctness gap the kernel closes; the vLLM side that routes to it is a companion PR.

## Changes

- New `pa_decode_shuffle.py` with `paged_attention_decode_shuffle_swa(...)`. It
  mirrors the shuffled indexing exactly (x = 16 // elem_size, = 16 for fp8):

  ```
  K: [num_blocks, num_kv_heads, head_size // x, block_size, x]
  V: [num_blocks, num_kv_heads, block_size // x, head_size, x]
  ```

  dequantizes fp8 KV to fp32 with the per-tensor or per-token K/V scales (A16W8,
  same as the asm path: query stays 16-bit, only KV is fp8), runs an online-softmax
  decode, and masks keys outside `[seq_len - window, seq_len)`. One program per
  (sequence, kv head); decode only (one query token per sequence).
- No change to existing kernels or call sites; this only adds the new file.

## Performance

Not a perf change. This is a correctness/enablement kernel (windowed shuffled-fp8
decode that did not exist before). The split-K performance variant is a separate PR.

## Testing

Standalone numerics harness against an independent fp8 reference, both fp8 dtypes
(`e4m3fn`, `e4m3fnuz`), across the full case matrix: head 128, GQA 32/8 and MQA,
block 16 and 32, windowed and full, per-tensor and per-token scales.

- MI300X (gfx942), ROCm 7.2.1, torch 2.10: ALL_PASS both dtypes. Max abs error
  kernel-vs-fp8-reference 7.7e-3 to 1.81e-2, while the fp8 quantization noise itself
  is 0.19 to 0.35, so the kernel error sits well under the quant floor.
- MI355X (gfx950), ROCm 7.13, torch 2.10: ALL_PASS both dtypes, max abs error
  <= 1.55e-2 against the same 0.18-0.35 quant floor.

The kernel is plain triton and arch-generic; the two arches cover both fp8 encodings.

- [x] Tested on MI300X (gfx942)
- [x] Tested on MI355X (gfx950)
- [ ] Performance benchmarks (n/a, correctness kernel)

## Documentation

Layout and masking are documented in the module docstring. No user-guide change.

## Dependencies

No new third-party dependencies (triton + torch only).

## Breaking Changes

None. New file only.

### DIFF
--- a/aiter/ops/triton/attention/pa_decode_shuffle.py
+++ b/aiter/ops/triton/attention/pa_decode_shuffle.py
@@ -1,0 +1,196 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+import triton
+import triton.language as tl
+
+from aiter.ops.triton.utils.logger import AiterTritonLogger
+
+_LOGGER = AiterTritonLogger()
+
+# Sliding-window paged decode over the shuffled fp8 KV cache. The asm pa_fwd
+# decode path takes no window arg, so SWA layers can't use the shuffled cache;
+# this reads the same layout and masks the window.
+#
+# Shuffled layout (x = 16 // elem_size, = 16 for fp8):
+#   K: [num_blocks, num_kv_heads, head_size // x, block_size, x]
+#   V: [num_blocks, num_kv_heads, block_size // x, head_size, x]
+#   K[d, slot] -> (d // x) * block_size * x + slot * x + (d % x)
+#   V[slot, d] -> (slot // x) * head_size * x + d * x + (slot % x)
+# A16W8: q is 16-bit, fp8 KV dequantized (val * scale) before the dot.
+
+
+@triton.jit
+def _pa_decode_shuffle_swa_kernel(
+    out_ptr,
+    q_ptr,
+    k_cache_ptr,
+    v_cache_ptr,
+    blk_tbl_ptr,
+    seq_lens_ptr,
+    k_scale_ptr,
+    v_scale_ptr,
+    scale,
+    sliding_window,
+    q_s0,
+    q_s1,
+    q_s2,
+    o_s0,
+    o_s1,
+    o_s2,
+    bt_s0,
+    ks_s0,
+    ks_s1,
+    ks_s2,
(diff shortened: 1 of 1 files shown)


reger-men:
The one red check here isn't the kernel, it's prepare_triton_wheel/prepare. It can't resolve the triton-kernels wheel off the AMD pip index, and since that job gates the Standard Tests, test_kvcache gets skipped and never runs, so a branch push won't clear it. @vgokhale, could someone on @ROCm/team_aiter who owns the CI re-run or patch that wheel step so the real tests fire? A review pass after would be great. This also unblocks #3962 (stacked on it) and the vLLM side in #46847.

reger-men:
Still blocked on CI here. The only red is prepare_triton_wheel/prepare, which can't resolve the triton-kernels wheel from the AMD pip index, so the Standard Tests including test_kvcache stay skipped and a branch push won't clear it. Could someone on @ROCm/team_aiter re-run or fix that gate so the tests run, then take a review pass? @vgokhale, #3962 is stacked on this.

reger-men:
@vgokhale this one is just waiting on the prepare_triton_wheel CI gate. Could you re-run it or point me at whats needed to get it green? I can rebase if you want.

reger-men:
@vgokhale following up on this one, still blocked on the prepare_triton_wheel gate rather than the code. Could you re-run it or tell me what to fix?

PR #3962, opened 2026-06-26
[Kernel][Perf] split-K long-context decode for shuffled fp8 SWA path

reger-men:
## Summary

Adds `aiter/ops/triton/attention/pa_decode_shuffle_perf.py`, a flash-decoding
(split-K) variant of the sliding-window shuffled-fp8 paged decode. The one-pass
kernel (companion PR, `pa_decode_shuffle.py`) runs one program per (sequence, kv
head) and walks the whole windowed key range serially, which underfills the device
at long context. This splits that range into `partition_size` chunks across the grid
and reduces the per-partition softmax state in a second pass. It is gated to engage
only where it actually helps and never regresses the short-context path.

## Motivation

At long context the one-pass per-program key walk gets long while the grid
(`num_seqs * num_kv_heads`) is too small to fill the GPU, so most CUs sit idle. The
asm decode path solves this with page-split kernels, but the asm path can't take a
window (the gfx942/950 `pa/*.co` kernels are prebuilt code objects with no `.s`
source or PA asm generator in the tree, and the registry has only a causal-mask flag,
no window). So the realistic lever for the windowed shuffled-cache path is the triton
kernel, parallelising context the same way the asm page-split kernels do.

## Changes

- New `pa_decode_shuffle_perf.py` with
  `paged_attention_decode_shuffle_swa_perf(...)`: a partition kernel + a reduce
  kernel, plus a one-pass fallback that runs the same algorithm as the companion
  one-pass kernel.
- The split is bounded to `[seq_len - window, seq_len)`, so a windowed layer only
  partitions the keys it actually attends. `max_seq_len` is taken as a host int
  (vLLM already tracks it), so no per-step `.item()` device sync.
- fp32 scratch is reused across decode steps (one buffer per device/shape) so the
  split path doesn't allocate every step.
- Gating (`_pick_n_part` + `_arch_allows_split`): the auto path splits only when the
  grid underfills or the serial walk is long (>= a threshold of partitions), AND only
  on arches where the split is measured to win end to end (gfx950 by default;
  override with `AITER_PA_SHUFFLE_SPLIT_ARCH`). Short context, a full grid, or a
  non-listed arch stays one-pass, so it is identical to the one-pass kernel there.
  `force_split` overrides the gating and is used by the tests to exercise both code
  paths on any arch.

## Performance

Decode tok/s, clock-locked (`perf_determinism`), Mistral-family SWA model
(`sliding_window=4096`), `--max-model-len 8192`, fp8 shuffled KV, `ROCM_AITER_FA`.
Before = one-pass kernel (companion PR), After = this split-K kernel.

MI355X (gfx950), ROCm 7.13 (where the split is enabled by default):

| Configuration | Before (one-pass) | After (split-K) | Improvement |
|---|---|---|---|
| batch 32, ctx ~590 (< window, stays one-pass) | 3573 | 3559 | parity (no split) |
| batch 8, ctx ~6700 (> window, n_part ~16) | 382-385 | 762-820 | ~2.0-2.1x |

At long context split-K is ~2.0-2.1x over one-pass and recovers decode to ~84-90% of
the bf16 baseline (909-913 tok/s); at short context it stays one-pass (parity). The
numbers above are from the 0.19.1 image with the 0.19.1 dispatch.

Re-confirmed on vLLM built from source at main d980a3cc
(0.23.1rc1.dev480+gd980a3cc6, same node, clock-locked), routing through the corrected
main dispatch, zephyr-7b-beta: short (batch 32, ctx ~590) bf16 3679 / one-pass 3470 /
split-K 3452 (parity, stays one-pass); long (batch 8, ctx ~6700, n_part ~16) bf16
~743 / one-pass ~387 / split-K 731 and 786 on two warm runs (~1.9-2.0x over one-pass,
recovering to ~parity with bf16). The first windowed split-K call pays a one-time
triton JIT for the partition grid (544 tok/s cold vs ~759 warm); one-pass and bf16
are stable to ~1%.

MI300X (gfx942), ROCm 7.2.1 (kept one-pass by the arch gate; numbers are with the
gate forced off, to show why):

| Configuration | Before (one-pass) | After (forced split-K) | Improvement |
|---|---|---|---|
| batch 32, ctx <window | ~949 | ~1005 | ~1.06x (within run-to-run spread) |
| batch 8, ctx ~6700 (> window) | ~244 | ~228 | ~0.93x (slightly behind) |

Isolated kernel microbenchmark on gfx942 (clock-locked) is a real win: parity at
short window, ~1.3-1.4x at win1024/ctx2048, ~4.6-6.4x at win4096/ctx8192 and full
causal. But on a dense 7B at these shapes the decode-attention kernel is a small slice
of the step next to the QKV/O projections and the MLP, so the isolated win does not
surface end to end and the extra reduce launch makes it parity-to-slightly-behind.
That is why the auto path keeps gfx942 on one-pass. gfx950's bf16 decode is much
faster (909 vs ~276 tok/s at the same long shape), so the underfilled one-pass kernel
is a far larger fraction of the step there and the split has real headroom.

## Testing

Numerics harness against the independent fp8 reference, both fp8 dtypes
(`e4m3fn`, `e4m3fnuz`), both code paths (`force_split` on and off), full case matrix
(head 128, GQA 32/8 and MQA, block 16/32, windowed and full, per-tensor and per-token)
plus long-context cases that exercise 4-16 partitions. Split-K reproduces the one-pass
result within the fp8 tolerance, so the reduce is not a correctness regression. This
is the same case matrix and ALL_PASS result as the companion one-pass kernel PR.

- MI300X (gfx942): ALL_PASS both dtypes, max abs <= 1.81e-2 vs a 0.18-0.35 quant floor.
- MI355X (gfx950): ALL_PASS both dtypes, max abs <= 1.55e-2 vs the same floor.

- [x] Tested on MI300X (gfx942)
- [x] Tested on MI355X (gfx950)
- [x] Performance benchmarks run (tables above)

## Documentation

Layout, split scheme, and gating are documented in the module docstrings.

### DIFF
--- a/aiter/ops/triton/attention/pa_decode_shuffle_perf.py
+++ b/aiter/ops/triton/attention/pa_decode_shuffle_perf.py
@@ -1,0 +1,584 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+import triton
+import triton.language as tl
+
+from aiter.ops.triton.utils.logger import AiterTritonLogger
+
+_LOGGER = AiterTritonLogger()
+
+# Flash-decoding (split-K) sliding-window paged decode over the shuffled fp8 KV
+# layout used by the asm pa_fwd path. Same cache layout and window semantics as
+# pa_decode_shuffle, but the windowed key range is split across PARTITION_SIZE
+# chunks so the long-context case fills the GPU instead of serialising one
+# program per (seq, kv head). A second pass reduces the per-partition softmax
+# state. For a single partition (small window) the host calls the one-pass
+# kernel directly so we don't pay the workspace round-trip.
+#
+# Shuffled layout (x = 16 // elem_size, = 16 for fp8):
+#   K: [num_blocks, num_kv_heads, head_size // x, block_size, x]
+#   V: [num_blocks, num_kv_heads, block_size // x, head_size, x]
+#   K[d, slot] -> (d // x) * block_size * x + slot * x + (d % x)
+#   V[slot, d] -> (slot // x) * head_size * x + d * x + (slot % x)
+
+
+@triton.jit
+def _pa_shuffle_swa_onepass_kernel(
+    out_ptr,
+    q_ptr,
+    k_cache_ptr,
+    v_cache_ptr,
+    blk_tbl_ptr,
+    seq_lens_ptr,
+    k_scale_ptr,
+    v_scale_ptr,
+    scale,
+    sliding_window,
+    q_s0,
+    q_s1,
+    q_s2,
+    o_s0,
+    o_s1,
+    o_s2,
+    bt_s0,
+    ks_s0,
+    ks_s1,
+    ks_s2,
+    k_blk_stride,
+    k_head_stride,
+    v_blk_stride,
+    v_head_stride,
+    HEAD: tl.constexpr,
+    HEAD_POW2: tl.constexpr,
+    BLOCK: tl.constexpr,
+    X: tl.constexpr,
+    GRP: tl.constexpr,
+    GRP_POW2: tl.constexpr,
+    SCALE_SCALAR: tl.constexpr,
+):
+    seq = tl.program_id(0)
+    kvh = tl.program_id(1)
+    seq_len = tl.load(seq_lens_ptr + seq)
+    win_start = 0
+    if sliding_window > 0:
+        ws = seq_len - sliding_window
+        win_start = tl.where(ws > 0, ws, 0)
+
+    grp = tl.arange(0, GRP_POW2)
+    d = tl.arange(0, HEAD_POW2)
+    q_head = kvh * GRP + grp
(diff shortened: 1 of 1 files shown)


reger-men:
Stacked on #3959. Only failure is one shard, MI35X shard 6 (moe_2stage fp4 allclose), a different path from the split-K decode this touches, everything else green. @vgokhale mind kicking off a re-run so we can confirm it's just flaky? I'll land #3959 first, then this.

reger-men:
Following up. This is stacked on #3959, and the only red is Standard Tests MI35X shard 6, the moe_2stage fp4 allclose test, a different suite from the split-K decode path this PR changes. @vgokhale could you re-run CI here? It should go green once #3959's wheel gate is fixed.

reger-men:
@vgokhale same as #3959, blocked on the prepare_triton_wheel gate rather than the code. Can you kick the CI or tell me what to fix?

reger-men:
@vgokhale same for this one, it depends on #3959 and is blocked on the MI35X shard-6 job, not the change itself. Can you kick the CI?

PR #3981, opened 2026-06-29
Support gemma_norm flag for rmsnorm path

Phi-C:
## Motivation

add `gemma_norm` flag support for rmsnorm path

## Submission Checklist


csrc/include/custom_all_reduce.cuh
```diff
@@ -4340,55 +4341,102 @@ void dispatchFusedAllReduceRMSNormQuant(hipStream_t stream,
 #define DISPATCH_AR_FUSION_KERNEL(NGPUS)                                                       \
```
zufayu, line 4426: The `DISPATCH_AR_FUSION_KERNEL` macro grew from ~59 to ~106 lines (+80%) to support the `gemma_norm` template branch. Each of the three launcher paths (1-stage, 2-stage, split) now duplicates its full 13-argument call site inside an `if (gemma_norm) / else` block.

The non-quant `dispatchFusedAllReduceRMSNorm` in this same file already handles the same dispatch with a cleaner double-pointer pattern — `launch_fused_allreduce_rmsnorm(kernel, gemma_kernel)` — where the if/else lives once in an 18-line macro and each call site is a single line. Adopting the same pattern here would keep the macro body flat and make adding future boolean template parameters much cheaper (currently a second bool would double the macro again to ~200 lines across 9 instantiation sites).

aiter/dist/parallel_state.py
```diff
@@ -559,6 +562,7 @@ def fused_allreduce_rmsnorm_quant(
         quant_type: Any = "per_token",
         group_size: int = 128,
         emit_bf16: bool = False,
+        gemma_norm: bool = False,
     ):
         if quant_type == "per_token" and group_size == 128 and not emit_bf16:
```
zufayu, line 632: This fast-path condition compares `quant_type` against the raw string `"per_token"` before any normalization. `communicator_cuda._normalize_fused_ar_rms_quant_type` maps several aliases (`"fp8"`, `"fp8_per_token"`, `"per-token"`, `QuantType.per_Token`, etc.) to `"per_token"`, but that normalization only happens deeper in the stack. Callers using those aliases will silently miss this torch-compile-guarded fast path and fall through to the slower `device_communicator` route.

Suggest calling `_normalize_fused_ar_rms_quant_type` at the top of this method before the fast-path branch (it's already imported via the communicator layer; alternatively expose it from a shared utils module).

aiter/dist/parallel_state.py
```diff
@@ -174,29 +174,32 @@ def fused_allreduce_rmsnorm_quant_fake(
     eps: float,
     group_name: str,
     prefill_support: bool = False,
+    gemma_norm: bool = False,
```
zufayu, line 177: No test covers the new `gemma_norm=True` path. The repo already has `op_tests/multigpu_tests/test_custom_allreduce_fp8.py` exercising the fused AR+RMSNorm+quant pipeline, so there is a ready template to extend.

Per the ROCm [contributing guidelines](https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests): *"New functionality is only merged with new unit tests."* A minimal test could call `tensor_model_parallel_fused_allreduce_rmsnorm_quant` with `gemma_norm=True` and compare the output against a reference `x / rms(x) * (1 + weight)` computation, covering at least the per-token FP8 path across all three dispatch variants (1-stage, 2-stage, split).

zufayu (approved):
LGTM. All three review points addressed: macro refactored to eliminate duplication, quant_type normalization moved before fast-path check, and gemma_norm=True dispatch paths covered by new tests.

valarLip (approved):

PR #3990, opened 2026-06-29
[MLA v4 nm] Align decode ABI with V3 stage1 valid-split slots

liyjiang:
## Summary
- Add `valid_split_count` (nullable i32 `[num_seqs]`) and trailing
  `use_valid_split_count_reduce` scalar to the v4 nm decode entry point so its
  kernarg call shape matches V3's `mla_decode_stage1_asm_fwd`. The shipped v4 nm
  `.co` consumes neither slot — they are passive placeholders enabling
  CUDA-graph-safe fixed-buffer plumbing without a separate codepath.
- `mla.py`: allocate `valid_split_count` up front at fixed `[num_seqs]` size
  (graph-safe), force `USE_VALID_SPLIT_COUNT_REDUCE=0` on gfx950 (uniform
  full-coverage splits → reduce is a no-op), and resolve
  `num_kv_splits`/`split_indptr` via `get_meta_param` when either is unset
  (dropped the redundant `arange` fallback).
- `ops/attention.py`: removed the duplicate `mla_decode_v4_asm` binding (was a
  ruff F811 redefinition); added the two new params to the surviving one.

## Test plan
- [x] `black --check` on changed Python files — unchanged
- [x] `ruff check` — All checks passed (F811 resolved)
- [x] `op_tests/test_mla_v4_nm.py` on gfx950 (mi350) — **5 passed**


amd-ruitang3:
we will merge into main after CI pass.

junhaha666 (approved):

PR #3997, opened 2026-06-29
fix(fused_moe): require KBatch >= 2 for block-fp8 split-k

ZiguanWang:
## Motivation

get_ksplit could pick a split where each K partition owns a single k-tile (KBatch = model_dim / (split * 256) == 1). In that case the CK stage1 kernel still uses atomic-add but skips the output memset, accumulating onto uninitialized memory and producing gibberish at bs=1 (e.g. Qwen3.5-FP8 TP8, K=4096, split=16). Add the KBatch >= 2 condition in get_ksplit, and guard ck_moe_stage1 so splitk from tuned CSV / AITER_KSPLIT cannot bypass it.

## Test Plan

Add op_tests/test_moe_blockscale_ksplit.py: exercises the high-level fused_moe() CK 2-stage path (which get_ksplit feeds, unlike test_moe_blockscale) against a bf16 pre-quant reference, asserting row-0 rel error stays within fp8 block-quant noise.

## Submission Checklist

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -3155,7 +3155,14 @@ def ck_moe_stage1(
     dtype=None,
 ):
     token_num = hidden_states.shape[0]
-    is_splitk = quant_type == QuantType.per_1x128 and splitk > 1
+    # Only enable split-k when each K partition owns >= 2 k-tiles
+    # (KBatch = K / (splitk * KPerBlock), KPerBlock == 256). When KBatch == 1 the
+    # CK kernel uses atomic-add but skips the output memset, accumulating onto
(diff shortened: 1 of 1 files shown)


valarLip (dismissed):

junhaha666:
`((model_dim // i) // tilek >= 2)` in `get_ksplit()`  can it be removed?  I noticed that this issue only exists in the CK backend, and you have already made the fix at the CK interface.

ZiguanWang:
> `((model_dim // i) // tilek >= 2)` in `get_ksplit()` can it be removed? I noticed that this issue only exists in the CK backend, and you have already made the fix at the CK interface.

already removed.

junhaha666 (approved):

PR #4015, opened 2026-06-30
Feat/fused qk norm rope 1way fp8 perhead

LiuYinfeng01:
## Motivation

Z-Image uses a 1way/single-stream Q/K/V layout, while the existing per-head FP8 fused RoPE path targets the 2way layout. This PR adds the matching 1way fused QK norm + RoPE + FP8 per-head quant path so Z-Image can use per-(batch, head) descale tensors and keep the CK FP8 flash-attention fast path instead of falling back to BF16 or per-tensor quantization.

This is the 1way counterpart of the 2way per-head FP8 quant support added in ROCm/aiter#3353.

## Technical Details

- Add `fused_qk_norm_rope_1way_fp8_perhead_quant`, a 1way fused Q/K RMSNorm + RoPE + FP8 quant API that emits per-head descale tensors for Q/K.
- Add `v_1way_per_head_fp8_quant` for matching V FP8 per-head quantization.
- Wire the new ops through the Python wrapper, C++ headers, ROCm op declarations, and HIP kernel implementation.
- Keep the descale layout compatible with CK FP8 flash attention, using per-(batch, head) scale granularity for better accuracy than per-tensor quantization.
- The separate PAIR_VEC_SIZE cos/sin vector fix is tracked in ROCm/aiter#4014.

## Test Plan

- Build AITER with the new 1way FP8 per-head quant kernels.
- Run the Z-Image text-to-image path using the new 1way fused Q/K RoPE + V FP8 per-head quant ops.
- Compare generated output against the baseline path to verify there is no visible quality regression.
- Let ROCm/aiter CI run the existing build and test coverage for affected fused RoPE/quant code paths.

## Test Result

- Verified the Z-Image 1way FP8 per-head quant path locally and compared generated images against the baseline output.
- Existing CI checks will provide build and regression coverage for the ROCm/aiter integration.

## Submission Checklist

### DIFF
--- a/csrc/kernels/fused_qk_norm_rope_cache_quant.cu
+++ b/csrc/kernels/fused_qk_norm_rope_cache_quant.cu
@@ -3525,6 +3525,207 @@ void fused_qk_norm_rope_1way(aiter_tensor_t& q,
     });
 }
 
+// ---------- Z-Image 1-way per-(batch, head) FP8 Q/K RoPE quant ----------
+
+template <typename T, int HEAD_SIZE>
+__global__ void qk_1way_output_partial_amax_kernel(const T* __restrict__ out_q_,
+                                                   const T* __restrict__ out_k_,
+                                                   int num_tokens,
+                                                   int num_heads_q,
+                                                   int num_heads_k,
+                                                   int total_warps,
+                                                   float* __restrict__ q_partial_amax,
+                                                   float* __restrict__ k_partial_amax)
+{
+    using mrope_utils::WARP_SIZE;
+    constexpr int VEC_SIZE        = HEAD_SIZE / WARP_SIZE;
+    const int warp_id             = threadIdx.x / WARP_SIZE;
+    const int lane_id             = threadIdx.x % WARP_SIZE;
+    const int num_warps_per_block = blockDim.x / WARP_SIZE;
+    const int global_warp_id      = blockIdx.x * num_warps_per_block + warp_id;
+    if(global_warp_id >= total_warps)
+    {
+        return;
+    }
+
+    const int batch_id      = blockIdx.y;
+    const int access_id     = lane_id * VEC_SIZE;
+    const int warp_offset_k = num_tokens * num_heads_q;
+    const bool is_q         = global_warp_id < warp_offset_k;
+
+    auto out_q = out_q_ + batch_id * num_tokens * num_heads_q * HEAD_SIZE;
+    auto out_k = out_k_ + batch_id * num_tokens * num_heads_k * HEAD_SIZE;
+
+    int token_id;
+    const T* src;
+    if(is_q)
+    {
+        const int spec = global_warp_id;
+        token_id         = spec / num_heads_q;
+        const int head_id = spec % num_heads_q;
+        src              = out_q + (token_id * num_heads_q + head_id) * HEAD_SIZE;
+    }
+    else
+    {
+        const int spec = global_warp_id - warp_offset_k;
+        token_id         = spec / num_heads_k;
+        const int head_id = spec % num_heads_k;
+        src              = out_k + (token_id * num_heads_k + head_id) * HEAD_SIZE;
+    }
+
+    vec_t<T, VEC_SIZE> x;
+    x.load(src + access_id);
+    float local_max = 0.f;
+#pragma unroll
+    for(int i = 0; i < VEC_SIZE; ++i)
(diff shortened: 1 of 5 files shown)


csrc/kernels/fused_qk_norm_rope_cache_quant.cu
```diff
@@ -3936,6 +4137,126 @@ void v_2way_per_head_fp8_quant(aiter_tensor_t& v0,
...
+    int64_t head_size  = v.size(3);
+    AITER_CHECK(head_size == 128, "v_1way_per_head_fp8_quant currently only supports head_size=128");
+    AITER_CHECK(v_fp8.dtype() == AITER_DTYPE_fp8, "v_fp8 must be fp8");
+    AITER_CHECK(v_descale.dtype() == AITER_DTYPE_fp32, "v_descale must be fp32");
+
+    HipDeviceGuard device_guard(v.device_id);
+    const hipStream_t stream = aiter::getCurrentHIPStream();
+
+    AiterTensor v_amax =
+        AiterTensor::empty({batch_size, num_heads}, AITER_DTYPE_fp32, v.device_id, stream);
```
yzhou103: This maybe AiterTensor::zero()? AiterTensor::empty does not zero-initialize the buffer. Since atomic_fmax_pos semantics are *ptr = max(*ptr, val), any garbage value in v_amax larger than the true amax will silently corrupt the descale, producing wrong FP8 quantization results.

csrc/kernels/fused_qk_norm_rope_cache_quant.cu
```diff
@@ -3936,6 +4137,126 @@ void v_2way_per_head_fp8_quant(aiter_tensor_t& v0,
...
+    {
+        int local_t = idx / HEAD_SIZE;
+        int d       = idx % HEAD_SIZE;
+        int t       = t_start + local_t;
+        int64_t off = ((int64_t)b * num_tokens + t) * slab_h_stride + (int64_t)h * HEAD_SIZE + d;
+        v_fp8_[off] = mrope_utils::fp8e4m3fnuz((float)v_[off] * inv);
+    }
+}
+
+void v_1way_per_head_fp8_quant(aiter_tensor_t& v,
```
yzhou103, line 4208: Why not use the quant op we already have?

csrc/kernels/fused_qk_norm_rope_cache_quant.cu
```diff
@@ -3936,6 +4137,126 @@ void v_2way_per_head_fp8_quant(aiter_tensor_t& v0,
...
+    int slab_h_stride = num_heads * HEAD_SIZE;
+    float inv         = 1.0f / v_descale[b * num_heads + h];
+
+    for(int idx = threadIdx.x; idx < (t_end - t_start) * HEAD_SIZE; idx += BT)
+    {
+        int local_t = idx / HEAD_SIZE;
+        int d       = idx % HEAD_SIZE;
+        int t       = t_start + local_t;
+        int64_t off = ((int64_t)b * num_tokens + t) * slab_h_stride + (int64_t)h * HEAD_SIZE + d;
+        v_fp8_[off] = mrope_utils::fp8e4m3fnuz((float)v_[off] * inv);
```
yzhou103, line 4204: fp8_max = 240.0 is hardcoded (L3442). It's correct for fp8e4m3fnuz (max=240), but would break for OCP e4m3 (max=448). get_dtype_max() derives this from the dtype instead — worth using it here to avoid a silent bug if the fp8 flavor ever changes. Or add check in the entrance of the op

csrc/kernels/fused_qk_norm_rope_cache_quant.cu
LiuYinfeng01: Because our ROCm AITER does not have a general per-head quantization op.

csrc/kernels/fused_qk_norm_rope_cache_quant.cu
LiuYinfeng01: Good point. This 1way Z-Image path is currently validated only on MI308/gfx942, where the fp8 flavor is e4m3fnuz and fp8_max=240. I added runtime checks at the op entry points for both fused_qk_norm_rope_1way_fp8_perhead_quant and v_1way_per_head_fp8_quant, and skipped the corresponding tests on non-gfx942. This avoids silently using the 240 constant on gfx950/OCP e4m3.

yzhou103 (approved):

PR #3974, opened 2026-06-28
[Tune] Add qwen3.5-397B MXFP4 a16w16 GEMM tuning configs

yichiche:
co-author: @Emmanuel0612 

## Motivation

Add a16w16 (bf16) GEMM tuning configs for **Qwen3.5-397B-A17B-MoE-MXFP4** (TP2). In this MXFP4 MoE model the experts run on the fp4 fused-MoE path; the remaining **dense bf16 GEMMs** (attention q/k/v/o, router gate, KV proj, and the MTP/EAGLE draft-layer projections) go through the `a16w16` path and benefit from per-shape kernel tuning.

## Technical Details

Adds two files under `aiter/configs/model_configs/`:

- `qwen3_5_397b_untuned_gemm.csv` — 188 shapes to tune
- `qwen3_5_397b_bf16_tuned_gemm.csv` — 187 tuned shapes (gfx950, cu_num=256)

Shape selection (TP2):
- 7 target bf16 GEMM shapes: `10240×4096`, `8704×4096`, `4096×4096`, `4096×512`, `1024×4096`, `512×4096`, `64×4096`.
- MTP / EAGLE draft-layer additions: q_proj `8192×4096` (output-gated), fc fusion / o_proj `4096×8192` (K=8192), sharded kv `256×4096`.
- Two `4096×512` shapes at M=1024/4096 were omitted because they already exist in `kimi_bf16_tuned_gemm.csv` (the repo enforces one owner per shape across bf16 configs; qwen reuses kimi's entry at runtime on the same gfx950).
- M grid sized for decode + EAGLE verify (draft runs at `1×bs`, target verify at `num_draft_tokens×bs`): `1,2,4,8,16,32,48,64,96,128,192,256,384,512,768,1024` (multiples of the runtime's 16/32 `getPaddedM` lookup granularity), plus a prefill tail `2048,4096,8192`. `16384/32768` intentionally dropped (the `gl=1` fallback caps `padded_M` at 8192 for `N>4096`).

Tuned with `csrc/gemm_a16w16/gemm_a16w16_tune.py`, `--libtype asm,opus,flydsl,triton,torch,skinny --shape_grouped`. **hipBLASLt is excluded** (opt-in `--with-hipblaslt` not used). flydsl wins ~68% of shapes.

## Test Plan

[...]

PR #4017, opened 2026-06-30
[Tune] Re-land qwen3.5-397B a16w16 GEMM configs without DSv4 conflict

yichiche:
## Motivation

Re-land the qwen3.5-397B a16w16 (bf16) GEMM tuning configs from #3974, which was
reverted in #4013 due to a config conflict with DSv4.

`aiter/configs/model_configs/qwen3_5_397b_bf16_tuned_gemm.csv` shared **60
identical shapes** with `aiter/configs/model_configs/dsv4_bf16_tuned_gemm.csv`.
The runtime config merge in `aiter/jit/core.py` concatenates the global
`bf16_tuned_gemm.csv` with every `model_configs/*_bf16_tuned_gemm.csv` and
deduplicates on the shape key:

```
(gfx, cu_num, M, N, K, bias, dtype, outdtype, scaleAB, bpreshuffle)
```

This key excludes `libtype`/`us`, so two independently-tuned winners for the
same shape collide. The merge then raised:

```
RuntimeError: Found 60 duplicate shape entries during merge of 'bf16_tuned_gemm'...
```

which failed `op_tests/test_gemm_a16w16.py`. These 60 shapes are model-agnostic
dense-GEMM shapes (N=64/256/512, K=4096) that both DSv4 and Qwen3.5 happened to
tune independently.

## Modifications

- Re-introduce `qwen3_5_397b_bf16_tuned_gemm.csv` and
  `qwen3_5_397b_untuned_gemm.csv`.
- The tuned config has the 60 shapes already present in
  `dsv4_bf16_tuned_gemm.csv` removed (**187 → 127 rows**), deferring those
  shared shapes to the existing DSv4 config.
- `dsv4_bf16_tuned_gemm.csv` is left untouched, so this PR is self-contained and
  does not modify another model's tuning. The untuned shape spec is restored
  unchanged.

## Test

`python op_tests/test_gemm_a16w16.py` passes (exit 0); all `checkAllclose`
checks pass and the merge no longer raises `RuntimeError`. Verified there are
**0** cross-file duplicate shapes across all `*_bf16_tuned_gemm.csv` configs, and
the test no longer rewrites any CSV as a side effect (the auto-dedup writeback no
longer triggers).


yifehuan (approved):
LGTM — 127 rows, clean power-of-two M bucketing, no hipBLASLt kernel names, and 0 cross-file key conflicts (incl. DSv4), so the re-land is clean.

PR #4011, opened 2026-06-30
perf: add fixed-tile HGEMM candidate

ftyghome:
## Motivation

The bf16 GEMM tuner currently selects among generic FlyDSL HGEMM, `small_m`, ASM, Triton, and Torch implementations on a per-shape basis.

For MLA attention in MoE decode workloads such as DeepSeek-V3 and Kimi-K2, one hot GEMM is the fused `q_a + kv_a` down-projection with `N=2112, K=7168`. This shape is tall-skinny and mostly bandwidth-bound: small `M`, large `K`, and deep split-K. In Kimi-K2.5 serving traces, it sits on the per-layer decode critical path, and the existing candidates are not always the best choice for this shape range.

This PR adds one more FlyDSL HGEMM candidate for the tuner.

## Technical Details

This PR adds `hgemm_fixed_tile`, a FlyDSL kernel family next to the existing `hgemm` and `small_m` kernels. It is ported from a hand-tuned HIP split-K GEMM.

The split-K reduction follows the same general approach as the existing kernel: cross-block split-K results are accumulated into C through packed-bf16 global atomics, guarded by a per-tile signal / semaphore handshake with in-kernel zeroing. No separate reduction workspace is required.

The main difference is that this kernel uses a fixed hand-tuned tile instead of the generic configurable warp shape:

- fixed 4-wave / 256-thread block;
- 2x2 MxN wave output partition;
- 8-block XOR LDS swizzle on both read and write paths to avoid bank conflicts;
- fp32 cshuffle epilogue that reuses the A-tile LDS, without a separate C buffer;
- one `global_atomic_pk_add_bf16` per two output columns.

This fixed tile is mainly targeted at the small/mid-`M`, large-`K`, deep-split-K regime. The kernel is added as a tuner candidate, so it is selected only for shapes where it wins. Existing tuned kernels remain used for the other shapes.

## Test Plan

We evaluated the change with both standalone microbenchmarks and end-to-end Kimi-K2.5 serving profiles.

Test environment:

```text
GPU: MI355X (gfx950)
ROCm: 7.2.3
```

1. **Standalone microbench**

   Each shape is compared against the baseline kernel selected by the existing upstream tuner.

   The benchmark uses `run_perftest` with torch-profiler device time (`self_device_time_total`) on ROCm / roctracer. Inputs are rotated to reduce L2 cache reuse, and each case is measured with `num_iters=101`.

2. **End-to-end profile**

[...]

PR #4023, opened 2026-06-30
feat(prezero): fuse split-K GEMM output zeroing into the preceding AR+RMSNorm

ColorsWind:
## Motivation

The split-K bf16 GEMM added in #4011, along with the generic FlyDSL `hgemm` /
`small_m` kernels, accumulate cross-block partial sums into the output `C`
through packed-bf16 global atomics. This requires `C` to be zero-initialized
before the GEMM runs.

To stay CUDA-graph safe, the zeroing is done inside the GEMM kernel, which costs
two device-wide synchronizations in the launch. In the decode regime this is a
**latency** problem, not a bandwidth one: the GEMM is latency-bound, and those
two in-kernel synchronizations cost on the order of 1-2us of pure latency on the
per-layer decode critical path, largely independent of problem size and paid
every layer, every step.

The key observation is that the synchronization *between* two kernels is free:
the kernel-launch boundary already orders one kernel after the other. So if the
output buffer is zeroed by some kernel that runs *before* the GEMM, the GEMM can
atomic-add straight into it with no zero-init and no handshake — and we pay no
extra synchronization at all.

This PR gives the upper framework that capability: a way to relocate a GEMM's
output-buffer zeroing into a preceding kernel. We realize it for the kernel
that precedes the hot down-projection GEMM in MoE attention decode for
DeepSeek-V4 and Kimi-K2 (the fused `q_a + kv_a` down-projection, N=2112,
K=7168) — the fused allreduce + RMSNorm that produces the GEMM's input. That
kernel launches one CTA per token, so in decode (few tokens) it leaves most CUs
idle; we ride zero-fill blocks along on those idle CUs and get the next GEMM's
output zeroed for free.

> Stacked on [#4011](https://github.com/ROCm/aiter/pull/4011) — please merge that first; the split-K bf16 GEMM kernel it
> introduces is the consumer of the `ZERO_INIT` skip path added here.

## Technical Details

The change has three parts.

**1. Fuse zero-fill into the fused allreduce-RMSNorm (`custom_all_reduce`).**
The fused AR-RMSNorm launch is widened with optional `zero_fill` ride-along
blocks past the RMSNorm grid (both the 1-stage and 2-stage paths), factored
into a shared `fused_zero_fill` device helper. These extra blocks zero a
separate, caller-provided buffer on the CUs the RMSNorm leaves idle, so the
next layer's split-K GEMM gets a pre-zeroed output for free. A host-side
`is_prezero_free(num_tokens, n)` helper reports whether the zero-fill actually
rides for free (AR CTAs + zero CTAs < CU count). `zero_fill` is threaded
through the op, pybind, communicator and dist plumbing (mirroring
`gemma_norm`); the non-custom fallback path just does `zero_fill.zero_()`.

**2. `ZERO_INIT` skip path for the split-K GEMM kernels.**
A `ZERO_INIT` flag (default `True`, preserving today's behavior) is added
across the `small_m` / `hgemm` / 4wave split-K kernels. When
`ZERO_INIT=False`, the kernel skips its in-kernel zero-init and the per-tile
signal / semaphore handshake, and atomic-adds straight into an already-zeroed
`C` (the kernel name gains a `_NOZINIT` suffix). The buffer must be zeroed by
the caller — e.g. by the AR-RMSNorm zero-fill ride-along above.
`compile_hgemm_kernel` / `compile_small_m_hgemm_kernel` /
`compile_splitk_hgemm_4wave` gain the flag; `hgemm_dispatch` and `gemm_kernels`
thread it down to `flydsl_hgemm`, allocating `C` with `torch.zeros` when
`ZERO_INIT=False` and the caller did not supply a pre-zeroed buffer.

**3. Thread `zero_init` + `out=` passthrough through the tuner
(`tuned_gemm`).**
`tgemm.mm` takes a `zero_init` flag and an optional pre-zeroed `out` buffer and
forwards them to the tuner-selected backend, so a caller can hand the
AR-RMSNorm-zeroed buffer straight into the GEMM and request the `_NOZINIT` skip
path. The skip only happens when the tuner picks FlyDSL; for shapes where it
picks a non-FlyDSL backend (e.g. torch at M=1/2) that ignores `out`,
`gemm_a16w16` copies the result back into the caller buffer, so `zero_init=False`
callers always get the correct output regardless of backend.

The skip path is only taken when the prezero rides for free
(`is_prezero_free`); otherwise the kernel keeps its self-zeroing default, so
shapes that do not benefit are unaffected and existing tuned selections are
preserved.

## Test Plan

We evaluated the change with both standalone microbenchmarks and end-to-end
Kimi-K2.5 serving profiles.

Test environment:

- GPU: MI355X (gfx950)
- ROCm: 7.2.3

**Standalone microbench**

All shapes are Kimi-K2.5 TP4 decode shapes. The fused allreduce + RMSNorm runs
on the hidden stream (hidden=7168); the split-K GEMM is the fused `q_a + kv_a`
down-projection (N=2112, K=7168). `M` is the decode token count, swept over the
serving concurrency range.

We measure the two halves of the codesign separately:

1. **AR + RMSNorm zero-fill overhead** — the cost the ride-along zero blocks
   add to the fused allreduce-RMSNorm launch, i.e. (launch with `zero_fill`) vs
   (launch without). Measured for both the **1-stage** and **2-stage** AR+RMS
   paths. While `is_prezero_free` holds (one CTA per token in decode, so most
   CUs are idle) this overhead should be ~0.

2. **split-K GEMM benefit** — the baseline GEMM with in-kernel zero-init
   (`ZERO_INIT=True`) vs the `_NOZINIT` GEMM atomic-adding into an
   already-zeroed buffer, i.e. the latency saved by dropping the two in-kernel
   synchronizations.

Each case is captured into a CUDA graph with 1000 replays of the kernel, and the
per-call time is derived from the total graph runtime. This reflects the
in-graph, latency-bound decode path the change targets.


ftyghome:
## Test Result

Environment: MI355X (gfx950), ROCm 7.2.3, TP4, 1000 graph replays per case.

**1. split-K GEMM benefit** (single GPU, via `tgemm.mm`). `gain = zinit - nozinit`,
i.e. the latency removed by skipping the in-kernel zero-init + handshake:

| gemm  |   M |    N |    K | lib    | zinit_us | nozinit_us | gain_us | gain_% |
|:------|----:|-----:|-----:|:-------|---------:|-----------:|--------:|-------:|
| qkv_a |   4 | 2112 | 7168 | flydsl |    6.763 |      5.052 |   1.711 |  25.3  |
| qkv_a |   8 | 2112 | 7168 | flydsl |    5.283 |      4.175 |   1.108 |  20.97 |
| qkv_a |  16 | 2112 | 7168 | flydsl |    5.631 |      4.523 |   1.109 |  19.69 |
| qkv_a |  32 | 2112 | 7168 | flydsl |    8.638 |      7.388 |   1.250 |  14.48 |
| qkv_a |  64 | 2112 | 7168 | flydsl |   11.897 |     10.849 |   1.048 |   8.81 |
| qkv_a | 128 | 2112 | 7168 | flydsl |   19.743 |     18.403 |   1.340 |   6.79 |
| qkv_a | 256 | 2112 | 7168 | flydsl |   37.709 |     33.987 |   3.722 |   9.87 |
| q_b   |   4 | 3072 | 1536 | flydsl |    3.595 |      2.328 |   1.268 |  35.26 |
| q_b   |   8 | 3072 | 1536 | flydsl |    3.736 |      2.416 |   1.321 |  35.35 |
| q_b   |  16 | 3072 | 1536 | flydsl |    3.884 |      2.579 |   1.305 |  33.59 |
| q_b   |  32 | 3072 | 1536 | flydsl |    4.248 |      3.091 |   1.156 |  27.22 |
| q_b   |  64 | 3072 | 1536 | flydsl |    5.264 |      4.589 |   0.675 |  12.82 |
| q_b   | 128 | 3072 | 1536 | flydsl |    8.138 |      6.525 |   1.612 |  19.82 |
| q_b   | 256 | 3072 | 1536 | flydsl |   16.156 |     16.188 |  -0.032 |  -0.20 |
| gate  |   4 |  256 | 7168 | flydsl |    4.938 |      4.009 |   0.929 |  18.81 |
| gate  |   8 |  256 | 7168 | flydsl |    4.919 |      4.013 |   0.905 |  18.41 |
| gate  |  16 |  256 | 7168 | triton |    6.373 |      6.386 |  -0.014 |  -0.22 |
| gate  |  32 |  256 | 7168 | flydsl |    5.027 |      3.938 |   1.088 |  21.65 |
| gate  |  64 |  256 | 7168 | flydsl |    6.016 |      4.247 |   1.769 |  29.40 |
| gate  | 128 |  256 | 7168 | flydsl |    6.522 |      5.618 |   0.904 |  13.86 |
| gate  | 256 |  256 | 7168 | flydsl |   17.582 |     14.324 |   3.258 |  18.53 |

The NOZINIT path removes a near-constant **~1 us** (14–35 % on the small kernels),
matching the fixed two-synchronization latency claimed above — it is roughly
independent of problem size. `gate M=16` dispatches to Triton (no NOZINIT path),
so it shows no change, as expected.

**2. AR + RMSNorm zero-fill overhead** (TP4, GPU-only). `overhead = with_zf - no_zf`:

| stage  | site         | tokens | zf_n | no_zf_us | with_zf_us | overhead_us | prezero_free |
|:-------|:-------------|-------:|-----:|---------:|-----------:|------------:|:-------------|
| 1stage | input_ln     |      4 | 5184 |   11.417 |     11.392 |      -0.026 | True         |
| 1stage | input_ln     |      8 | 5184 |   11.828 |     11.708 |      -0.120 | True         |
| 1stage | input_ln     |     16 | 5184 |   13.482 |     13.381 |      -0.100 | True         |
| 1stage | input_ln     |     32 | 5184 |   17.309 |     17.318 |       0.010 | True         |
| 1stage | input_ln     |     64 | 5184 |   26.077 |     26.131 |       0.055 | True         |
| 1stage | input_ln     |    128 | 5184 |   42.106 |     44.210 |       2.104 | False        |
| 1stage | post_attn_ln |      4 |  256 |   11.417 |     11.761 |       0.344 | True         |
| 1stage | post_attn_ln |    128 |  256 |   42.106 |     42.082 |      -0.024 | True         |
| 2stage | input_ln     |      4 | 5184 |   13.942 |     13.829 |      -0.113 | True         |
| 2stage | input_ln     |     32 | 5184 |   16.872 |     16.927 |       0.055 | True         |
| 2stage | input_ln     |     64 | 5184 |   21.210 |     21.365 |       0.154 | True         |
| 2stage | input_ln     |    128 | 5184 |   30.196 |     30.283 |       0.086 | False        |
| 2stage | post_attn_ln |    128 |  256 |   30.196 |     30.154 |      -0.043 | True         |

(Rows abbreviated; full sweep M ∈ {4,8,16,32,64,128} × {1stage,2stage} × both
sites is in the script output.)

The zero-fill rides **free** (overhead within ±0.3 us noise) wherever
`is_prezero_free` holds. The one case it does not — 1-stage / input_ln
(zf=5184) at 128 tokens — is exactly the row the heuristic flags `False`, and it
is the only one that pays a real cost (+2.1 us). The 2-stage path hides even that
case (+0.09 us at the same shape), since it launches more CTAs with better
overlap. Outputs match the reference within 1e-2.

**Analysis.** The two tables are the two sides of the trade the codesign makes,
and together they show it is a net win on the decode critical path:

- The split-K GEMM gives back a **fixed ~1 us per GEMM** when it skips the
  in-kernel zero-init + handshake. The saving barely moves with M or N (compare
  qkv_a's ~1.0–1.7 us across M=4→128 with gate's ~0.9–1.8 us), which is the
  signature of a fixed two-synchronization latency rather than saved compute —
  exactly the cost the Motivation attributes to staying CUDA-graph safe. Because
  it is fixed, it is a larger *fraction* of the small, fast kernels (q_b and gate
  at ~20–35 %) that sit on the per-layer decode path.

- The AR+RMSNorm pays **~0 us** to zero those buffers for it, because the
  ride-along blocks land on CUs the RMSNorm leaves idle. That the overhead is the
  same (~0) for zf=5184 and zf=256 — a 20× difference in bytes zeroed — confirms
  it is not GPU zeroing work being timed but truly hidden capacity.

- So per attention layer per decode step, the codesign moves the zeroing of two
  downstream GEMM outputs (qkv_a + q_b after input_layernorm, MoE gate after
  post_attn_layernorm) off the GEMM critical path (≈ 1 us × number of prezero
  GEMMs saved) for ≈ 0 added AR cost — as long as `is_prezero_free` holds. The
  `is_prezero_free` guard is what keeps the trade one-directional: it declines
  exactly the shapes where the ride-along would stop being free (the 128-token /
  5184-wide 1-stage case), so no shape regresses.

The test script can be found here:
[test_prezero_bench.py](https://github.com/user-attachments/files/29534083/test_prezero_bench.py)

ftyghome:
Hi,

I have the merge conflict resolved.

ftyghome:
Hi，

The e2e perf test results are ready and can be found [here](https://github.com/ROCm/ATOM/pull/1421).

Feel free to let me know if there are anything needed from my end. Thanks!

PR #4025, opened 2026-07-01
Remove FlyDSL fallback machinery from MoE tuning (FlyDSL now mandatory)

coderfeli:
## Summary

FlyDSL is now a required dependency for MoE tuning, so the FlyDSL→CK **fallback** machinery is dead weight. Previously the tuner recorded, alongside every winning `flydsl_` config, a companion CK config tagged `flydsl_fallback`, and the runtime swapped it in when `is_flydsl_available()` was `False`. This PR removes that generation + consumption path and the associated `_tag` data.

## Changes

**Tuner — `csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py`** (stops generating fallbacks)
- Removed `FLYDSL_FALLBACK_TAG`, the `self._flydsl_fallbacks` accumulator, the `_non_flydsl_best` computation, and the `best_has_flydsl` block that merged a CK stage1+stage2 fallback row.
- `result_to_csv` no longer writes a `_tag` column or appends fallback rows. It now just drops any legacy `_tag` column/rows read from older tuned files, then writes a plain deduped table.

**Runtime — `aiter/fused_moe.py`** (stops consuming fallbacks)
- Removed `_flydsl_fallback_cache`, `get_flydsl_fallback_cfgs()`, and the runtime block that substituted a fallback cfg (or dropped to default heuristics) when FlyDSL was unavailable.
- The generic `_tag` read-filter in `get_cfg_2stages` is retained as defensive backward-compat for any stale user CSV. The FlyDSL-vs-CK dispatch gates (`is_flydsl1 and is_flydsl_available()`) are untouched — those select the wrapper, not fallback config.

**Config data**
- Removed the obsolete `flydsl_fallback` rows (372 rows) from the 7 tuned configs that carried them: `dsv3_fp4`, `dsv4_fp8fp4`, `gptoss_fp8fp4`, `kimik2_fp4`, `kimik2_fp8fp4`, `minimax_m25_fp4`, `qwen3_5_397b_fp4`. Only the tagged rows were deleted; all other rows and the `_tag` column are left byte-identical to minimize churn.

**Test**
- `op_tests/tuning_tests/test_online_tune.py::test_tag_filter` de-hardcoded its sample tag value (still validates the retained `_tag` read-filter).

## Behavior change

With FlyDSL mandatory, a tuned config that names a `flydsl_` kernel now **requires** FlyDSL to be installed — there is no CK fallback substitution. If FlyDSL is absent, the FlyDSL dispatch gate is skipped as before, so a `flydsl_`-named config will no longer silently degrade to a CK kernel.

## Testing
- `python -m py_compile` on both edited modules — OK.
- `pytest op_tests/tuning_tests/test_online_tune.py -k tag_filter` — passing.
- Repo-wide grep for `flydsl_fallback` — no matches remain.


lalala-sh (dismissed):
LGTM

PR #4031, opened 2026-07-01
[OPUS] Fix layout_shifted dropping compile-time shift N under runtime addition

kaiyang-1:
Operator overloads for usability:

* Implemented `operator+=` and `operator+` for `layout_shifted`, enabling both in-place and returned offset adjustments to the layout.
* Runtime addition must preserve the compile-time shift N. Without this, the inherited base::operator+(index_t) would slice the layout_shifted back to its base and silently drop N, making `u + number<N> + int` differ from `u + int + number<N>`.

### DIFF
--- a/csrc/include/opus/opus.hpp
+++ b/csrc/include/opus/opus.hpp
@@ -643,6 +643,8 @@ struct layout_cached : public remove_cvref_t<Layout> {
     array<index_t, num_issues> offsets;
 };
 
+// Wraps a layout with a *compile-time* offset N. Keeping N in the type lets smem::tr_load lower it into the `offset:` immediate of ds_read_b64_tr_*,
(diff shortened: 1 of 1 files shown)


carlushuang (approved):
LGTM

demonsan:
can you do a simple test for gfx1250? since 320KB LDS will over 64KB immed offset. In my previous experiments, it will make compiler allocate more vgpr for LDS vaddr and lead to vgpr spill.

kaiyang-1:
> can you do a simple test for gfx1250? since 320KB LDS will over 64KB immed offset. In my previous experiments, it will make compiler allocate more vgpr for LDS vaddr and lead to vgpr spill.

The immediate offset of the `ds_read` instruction has a maximum of 65535. If you need to use a larger LDS offset, you can shift the base pointer multiple times and call `make_smem` separately for each offset.

PR #4041, opened 2026-07-01
[fix](moe): fix the accuracy M=1 in qwen3.5

PerryZhang01:
## Motivation

When running GSM8K evaluation with `NUM_CONCURRENT=1`, decode requests can enter the fused MoE path with very small token batch sizes, often `M=1`. The FP8 fused MoE kernels are not safe for `M < 16`, which can corrupt decode outputs and collapse accuracy to zero. Higher concurrency hides the issue by naturally batching requests to larger `M`.
This change pads small MoE decode batches to a safe minimum `M=16` internally and slices the output back to the original token count, restoring correctness for single-concurrency evaluation while preserving the external tensor shape.


junhaha666:
I think this is a problem with the kernel, and it needs to be fixed on the kernel itself.

PR #2994, opened 2026-05-01
[Gluon]: Gluon kernel for mxfp4 quant gfx950

NimitPtl:
## Motivation

Write a gluon kernel for quant_mxfp4 for better efficiency compared to Triton.

## Technical Details

Gluon kernel, with sizes derived from the TTGIR of Triton kernel.

## Test Plan

Test suite as well as benchmark script for testing accuracy and speed of quant kernel.

## Test Result

Marginal improvement of kernel.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4054, opened 2026-07-02
[TRITON] MXFP4 Quantization Gluon implementation for gfx950 and gfx1250

NimitPtl:
## Motivation

Using gluon instead of triton for performance improvement

## Technical Details

Gluon kernel, with sizes derived from the TTGIR of Triton kernel.

## Test Plan

Test suite as well as benchmark script for testing accuracy and speed of quant kernel.

## Test Result

Marginal improvement for gfx950 and around 10% improvement with usage of LDS in gfx1250

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/quant/quant.py
+++ b/aiter/ops/triton/quant/quant.py
@@ -18,7 +18,8 @@ from aiter.ops.triton._triton_kernels.quant.quant import (
     _nvfp4_quant_op,
 )
 from aiter.ops.triton.utils.logger import AiterTritonLogger
-from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon
+from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon_950
+from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon_1250
 from aiter.jit.utils.chip_info import get_gfx
 from aiter.ops.triton.utils.types import e4m3_dtype
 
@@ -160,7 +161,7 @@ def dynamic_mxfp4_quant(
         scaling_mode: The method to calculate MX block scaling.
             - "even" (default): `even_round` in `quark.torch.quantization.utils`.
             - etc.
-        use_gluon: Whether to use Gluon kernel (default: True). Only available on MI350 (gfx950).
+        use_gluon: Whether to use Gluon kernel (default: True). Only available on (gfx950/1250).
                   Falls back to Triton kernel on other GPUs.
     Returns:
         A tuple of (x_fp4, blockscale_e8m0).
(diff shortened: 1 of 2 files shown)


cagrikymk:
@NimitPtl The target for this PR is https://github.com/ROCm/aiter/tree/quant_gluon branch, is that on purpose or you want to merge to main?

aiter/ops/triton/quant/quant.py
```diff
@@ -18,7 +18,8 @@
     _nvfp4_quant_op,
 )
 from aiter.ops.triton.utils.logger import AiterTritonLogger
-from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon
+from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon_950
+from aiter.ops.triton.gluon.quant_mxfp4 import _dynamic_mxfp4_quant_kernel_gluon_1250
 from aiter.jit.utils.chip_info import get_gfx
```
rahulbatra85, line 23: I would use this https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/utils/_triton/arch_info.py which is Triton/Gluon specific

aiter/ops/triton/quant/quant.py
```diff
@@ -160,7 +161,7 @@ def dynamic_mxfp4_quant(
         scaling_mode: The method to calculate MX block scaling.
             - "even" (default): `even_round` in `quark.torch.quantization.utils`.
             - etc.
-        use_gluon: Whether to use Gluon kernel (default: True). Only available on MI350 (gfx950).
+        use_gluon: Whether to use Gluon kernel (default: True). Only available on (gfx950/1250).
```
rahulbatra85, line 164: I think we should use a standardized approach across all kernels and op wrappers. Right now this is all over the place.

My suggestion is that we should add a backend field to each op which can take values "triton", "gluon", or None(default). 

```
if backend is None:
  call the best performing kernel we know of for this op on this arch whether gluon or triton. This is hardcoded by the kernel/op developer or whoever tunes the kernel later on.
elif backend == "gluon":
   call the gluon kernel if available on that arch(some helper in the https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/utils/_triton/arch_info.py) . Otherwise, issue a warning and fall back to Triton kernel
else:
   call the Triton kernel(assumption is there is always a Triton kernel for an op, I think that's a reasonable one to make.)
```

rahulbatra85:
Can you post the benchmark numbers as well for both gfx950 and gfx1250?

aiter/ops/triton/quant/quant.py
NimitPtl: https://github.com/ROCm/aiter/pull/2994/changes
this is the base change. This choses triton/gluon based on the user preference.
let me know if this is a satisfactory approach.

NimitPtl:
> Can you post the benchmark numbers as well for both gfx950 and gfx1250?

have posted the same in Test Result in description.

NimitPtl:
> @NimitPtl The target for this PR is https://github.com/ROCm/aiter/tree/quant_gluon branch, is that on purpose or you want to merge to main?

I think it should be main, however while creating the PR, the older patch was pushed before the repository changes went into affect.
https://github.com/ROCm/aiter/pull/2994/changes is the base change on which the current change is based on. Ultimately both should go into main.

NimitPtl:
closing the moving this to https://github.com/ROCm/aiter/pull/4193
consolidate changes and resolve repo permission issues.

PR #4056, opened 2026-07-02
gfx1250 features compile guard

demonsan:
solve tdm compile error witch clang <=20. Now all gfx1250 features won't be enabled


carlushuang (approved):
LGTM

PR #4065, opened 2026-07-02
feat(attention): head-dim-tiled Triton flash attention for ViT (gfx1151)

carlushuang:
## feat(attention): head-dim-tiled Triton flash attention for ViT

Adds `aiter/ops/triton/attention/vit_flash_attn.py` — non-causal per-image varlen attention for ViT encoders whose `head_dim` is small and not a power of two (e.g. Qwen3.x ViT: 72): instead of padding head_dim to the next pow2 (128) it tiles into 5×16=80 (WMMA k-tile 16) with masked loads for d ≥ head_dim, so the contraction is 80-deep (online softmax, one workgroup per image/head/query-block, head_dim ≤ 80). Portable Triton, validated on gfx1151 where torch SDPA falls back to the math backend; consumed by ROCm/ATOM Qwen3.x VL (~21× over SDPA, relerr < 0.007).


brunomazzottiamd:
https://github.com/ROCm/aiter/pull/3858 is proposing kernel changes targeting vision-transformer head_dim (72) (and other small non-power-of-2 dims). Should we merge both PRs?

PR #4096, opened 2026-07-06
[Perf] mla v4 nm: skip valid_split_count fill

liyjiang:
v4 nm ships only for gfx950, where the stage2 reduce runs with USE_VALID_SPLIT_COUNT_REDUCE=0 so it never reads valid_split_count's content, and the v4 nm decode kernel doesn't write it back either -- only a valid device pointer is needed for the ABI. torch.full launches a vectorized_elementwise_kernel that shows up as an extra slice right before the decode kernel in a per-call perf trace; use torch.empty to drop that fill.


amd-ruitang3:
LGTM. Have tested.

amd-ruitang3 (approved):

PR #4108, opened 2026-07-06
Fix A8W4 CDNA4 scale addressing for padded MoE shapes

xiaohuguo2023:
- Fixes A8W4 MoE GEMM when CDNA4 MX scale swizzle is used with padded weight tensors and unpadded logical dimensions.
- Keeps physical padded `N/K` for swizzled scale addressing, while using logical unpadded `N/K` for dispatch lookup.
- Prevents corrupt W4A8 gpt-oss outputs caused by reading swizzled scale metadata with mismatched dimensions.

before this PR 

```
local-chat-completions ({'model': 'amd/gpt-oss-120b-w-mxfp4-a-fp8',
  'base_url': 'http://0.0.0.0:8894/v1/chat/completions', 'api_key': 'EMPTY',
  'eos_string': '</s>', 'max_retries': 5, 'num_concurrent': 64, 'timeout':
  1800, 'tokenized_requests': False, 'max_length': 9472}), gen_kwargs:
  ({'max_tokens': 5376, 'temperature': 0, 'top_p': 1}), limit: None,
  num_fewshot: None, batch_size: 1
  |Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
  |-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
  |gsm8k|    3.0|flexible-extract|     5|exact_match|↑  |  0.0|±  |   0.0|
  |     |       |strict-match    |     5|exact_match|↑  |    0|±  |     0|
```
after this PR

```
local-chat-completions ({... 'num_concurrent': 64, ... 'max_length': 9472}),
     gen_kwargs: ({'max_tokens': 5376, 'temperature': 0, 'top_p': 1}), limit:
  None, num_fewshot: None, batch_size: 1
  |gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9629|±  |0.0052|
  |     |       |strict-match    |     5|exact_match|↑  |0.9629|±  |0.0052|
```

### DIFF
--- a/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
+++ b/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
@@ -402,7 +402,9 @@ def test_cdna4_swizzled_scales_with_padded_physical_dims(device="cuda"):
         w_scale, arch="gfx950", preshuffle_factor=32, scale_kwidth=8
     )
 
-    x_static_scale = x_physical.abs().max().float() / 448.0
+    x_static_scale = (
+        x_physical.abs().max().float() / torch.finfo(torch.float8_e4m3fn).max
+    )
     x_tri = downcast_to_static_fp8(x_physical, x_static_scale)
     # unpadded_N/K are dispatch keys only for CDNA4 swizzle; the kernel must
     # still launch over the physical padded tensor dimensions.
--- a/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
+++ b/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
@@ -406,18 +406,12 @@ def test_cdna4_swizzled_scales_with_padded_physical_dims(device="cuda"):
         x_physical.abs().max().float() / torch.finfo(torch.float8_e4m3fn).max
     )
     x_tri = downcast_to_static_fp8(x_physical, x_static_scale)
-    # CDNA4 N vs K asymmetry:
-    #   N: the kernel launches over the PHYSICAL padded N (the swizzle is baked
-    #      over it); unpadded_N is a dispatch key only. The extra N columns are
-    #      just sliced off the output (tri_y[:, :logical_n]) — no corruption.
-    #   K: unpadded_K IS the contraction bound the kernel uses (address-safe,
-    #      masked), so the kernel sums only the first logical_k. The reference
-    #      must therefore contract the SAME logical-K slice — physical_k here is
-    #      random (not zero-padded like real vLLM weights), so contracting the
-    #      full 1024 would diverge. logical_k=800 < physical_k=1024 exercises it.
+    # CDNA4: kernel launches over physical padded N (swizzle is baked over it),
+    # so unpadded_N is a dispatch key only; unpadded_K is address-safe (masked).
+    # logical_k=800 < physical_k=1024 exercises the K path.
     ref_y = moe_gemm_torch(
-        x_physical[:, :logical_k],
-        w_ref[:, :logical_k, :],
+        x_physical,
+        w_ref,
         bias_physical[:, :logical_n],
         rdata,
         gindx,
--- a/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
+++ b/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
@@ -402,13 +402,10 @@ def test_cdna4_swizzled_scales_with_padded_physical_dims(device="cuda"):
         w_scale, arch="gfx950", preshuffle_factor=32, scale_kwidth=8
     )
 
-    x_static_scale = (
-        x_physical.abs().max().float() / torch.finfo(torch.float8_e4m3fn).max
-    )
+    x_static_scale = x_physical.abs().max().float() / 448.0
     x_tri = downcast_to_static_fp8(x_physical, x_static_scale)
-    # CDNA4: kernel launches over physical padded N (swizzle is baked over it),
-    # so unpadded_N is a dispatch key only; unpadded_K is address-safe (masked).
-    # logical_k=800 < physical_k=1024 exercises the K path.
+    # unpadded_N/K are dispatch keys only for CDNA4 swizzle; the kernel must
+    # still launch over the physical padded tensor dimensions.
     ref_y = moe_gemm_torch(
         x_physical,
         w_ref,
--- a/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
+++ b/op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
@@ -361,3 +361,79 @@ def test_op(
     if not act_mxfp8 and fused_quant:
         tri_y = (tri_y.float() * quant_static_scale).to(ref_y.dtype)
     assert_close(ref_y, tri_y, maxtol=maxtol, rmstol=rmstol)
+
+
+def test_cdna4_swizzled_scales_with_padded_physical_dims(device="cuda"):
+    if get_arch() != "gfx950":
+        pytest.skip("CDNA4 scale swizzle is only used on gfx950.")
+
+    torch.manual_seed(0)
+
+    m = 16
+    logical_n = 704
+    logical_k = 800
+    physical_n = 1024
+    physical_k = 1024
+    n_expts_tot = 8
+    n_expts_act = 1
+
+    m, rdata, gindx, sindx = init_routing_data(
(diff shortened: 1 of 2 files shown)


xiaohuguo2023:
@lburzawa  Thanks, that is a fair concern. The intent here is not to disable the padding-load optimization globally. The non-CDNA4 path still switches `N/K` to the logical unpadded dims. The special case is only `swizzle_mx_scale == "CDNA4_SCALE"`: for that path the MX scale tensor has already been swizzled over the physical padded `N/K` layout, so using logical `N/K` for scale addressing can read the wrong packed scale block. We still use logical `N/K` for the dispatch/config lookup so the same tuned tile choice is selected for the model shape.

For repro, these are the images I used:

- before / unpatched base: `vllm/vllm-openai-rocm:nightly-34b560b725b401f33477464163244b4744d1c40f`
- after / patched image: `vllm/vllm-openai-rocm:gptoss-a8w4-padded-swizzle-fix`

The patched image is the same nightly base with the AITER site-package patched with this PR. The model path I used on MI355X is:

```bash
/data/amd/gpt-oss-120b-w-mxfp4-a-fp8
```

Perf repro, single TP1/CONC64 point:

```bash
IMAGE=vllm/vllm-openai-rocm:gptoss-a8w4-padded-swizzle-fix
# For A/B, change IMAGE to:
# IMAGE=vllm/vllm-openai-rocm:nightly-34b560b725b401f33477464163244b4744d1c40f

export MODEL=/data/amd/gpt-oss-120b-w-mxfp4-a-fp8
export TP=1
export CONC=64
export ISL=8192
export OSL=1024
export MAX_MODEL_LEN=9472
export RANDOM_RANGE_RATIO=1
export RESULT_FILENAME=gptoss_fp4_mi355x_tp1_conc64.json
export PORT=18894
```

Accuracy repro, same server path plus GSM8K lm-eval:

```bash
IMAGE=vllm/vllm-openai-rocm:gptoss-a8w4-padded-swizzle-fix
# For A/B, change IMAGE to the unpatched nightly above.

export MODEL=/data/amd/gpt-oss-120b-w-mxfp4-a-fp8
export TP=1
export CONC=64
export ISL=8192
export OSL=1024
export MAX_MODEL_LEN=9472
export RANDOM_RANGE_RATIO=1
export RESULT_FILENAME=gptoss_fp4_mi355x_tp1_conc64_eval.json
export PORT=18894
export RUN_EVAL=true
export EVAL_TASKS_DIR=utils/evals/gsm8k.yaml
'
```

For the broader accuracy sweep I used the same patched image/server path and lm-eval concurrency values `4 8 16 32 64 128 256`. The patched run completed all points; examples:

- TP1 CONC64 GSM8K exact match: `0.9628506444`
- TP1 CONC256 GSM8K exact match: `0.9636087945` strict / `0.9628506444` flexible

The original failure mode was zero/near-zero accuracy with visibly corrupted generations on the unpatched A8W4 path at TP1 high concurrency. This PR is meant to restore correct scale addressing for physically padded CDNA4-swizzled scales; I agree perf should be checked because this does make the CDNA4 swizzled path launch over physical padded `N/K` rather than logical `N/K`.

xiaohuguo2023:
@lburzawa please use below instructions

```
 1. Launch the container (no serve)

  docker run -d --name gptoss-repro \
    --device /dev/kfd --device /dev/dri --ipc host --shm-size 16g \
    --group-add video --cap-add SYS_PTRACE --security-opt seccomp=unconfined \
    -v /data:/data -v /tmp:/tmp \
    --entrypoint sleep \
    vllm/vllm-openai-rocm:nightly infinity
  docker exec -it gptoss-repro bash        # work inside

  2. Serve the model (inside the container) — the 3 gotchas matter

  export VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_TRITON_ROPE=1 \
         VLLM_ROCM_QUICK_REDUCE_QUANTIZATION=INT4 AMDGCN_USE_BUFFER_OPS=0

  vllm serve /data/amd/gpt-oss-120b-w-mxfp4-a-fp8 --port 8891 \
    --tensor-parallel-size 8 \
    --compilation-config '{"pass_config":{"fuse_rope_kvcache":true}}' \
    --gpu-memory-utilization 0.95 --max-model-len 16384 \
    --block-size 64 --no-enable-prefix-caching \
    --reasoning-parser openai_gptoss &                 # <-- REQUIRED: else content is empty
  # wait until "Application startup complete" / curl :8891/health returns ok

  3. Run GSM8K with lm-eval directly 

  pip install "lm-eval[api]"
  python3 -m lm_eval --model local-chat-completions --apply_chat_template \
    --tasks gsm8k \
    --model_args
  "model=/data/amd/gpt-oss-120b-w-mxfp4-a-fp8,base_url=http://0.0.0.0:8891/v1/chat/completions,api_key=EMPTY,num_concurrent=64,max_length=16384,tokenized_requests=False" \
    --gen_kwargs "max_tokens=12288,temperature=0,top_p=1" \
    --output_path /tmp/eval --log_samples
  # read the **flexible-extract** exact_match row (NOT strict-match)

  The three non-obvious things InferenceX handles for you — replicate them or the score is meaningless:
  1. --reasoning-parser openai_gptoss on vllm serve — without it chat-completions return empty content (completion_tokens>0 but content=null).
  2. --model local-chat-completions --apply_chat_template against /v1/chat/completions — using plain completions gives 0.0 with a stuck !!!! loop.
  3. Read flexible-extract — gpt-oss answers in prose ("The answer is 4."), so strict #### 4 match under-scores (~0.13 vs ~0.90).

  4. Toggle the aiter PR fix (same container)

  AITER=/usr/local/lib/python3.12/dist-packages/aiter/ops/triton/moe/moe_op_gemm_a8w4.py
  cp -n $AITER $AITER.bak
  # WITHOUT fix (bug): allocate on unpadded N -> OOB
  sed -i 's/^        padded_N,$/        N,/' $AITER
  rm -rf /root/.cache/vllm /root/.triton /tmp/torchinductor_root   # clear compile cache
  # restart vllm serve, re-run lm_eval  -> GSM8K drops
  # restore: cp $AITER.bak $AITER
```

xiaohuguo2023:
@lburzawa , follow your two findings, I did reinvestigation below

> 1) It seems that the accuracy error dont come from unpadded_K, they only come from unpadded_N. With K set to logical/unpadded K accuracy still seems good. Also I found that this zero_padding function in vllm [vllm/vllm/model_executor/layers/quantization/online/fp8.py at b44311b6ef9232d1f345f4b55adef7abc223f0…](https://github.com/vllm-project/vllm/blob/b44311b6ef9232d1f345f4b55adef7abc223f0e7/vllm/model_executor/layers/quantization/online/fp8.py#L58

You're right that it's N, not K — that's the CDNA4 scale-swizzle addressing bug: the swizzled scale layout is built over the
  padded N, but the kernel addressed it with unpadded N, so it read the wrong scale for every weight and corrupted valid outputs. Our aiter PR fix (31e0fd397, fix-a8w4-padded-swizzle) keeps the physical padded N/K for scale addressing and uses
  logical/unpadded only for dispatch lookup.")

>2) zeroes out the weights and bias but not weight scales. Could we try to zero or fill with ones the padding on weight scales in vllm and then check accuracy again?

"Zeroing the weight-scale padding won't fix it: the bug is addressing of the valid scales, not garbage in the padded region —
  and in the quark path gpt-oss actually uses, that padding is already ones/zeros (safe).

op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
```diff
@@ -361,3 +361,77 @@ def test_op(
...
+        device=device,
+    )
+
+    w_tri, w_scale = downcast_to_mxfp(w_physical, torch.uint8, axis=1)
+    w_ref = upcast_from_mxfp(w_tri, w_scale, torch.bfloat16, axis=1)[:, :, :logical_n]
+    w_scale_tri = shuffle_scale_moe(
+        w_scale, arch="gfx950", preshuffle_factor=32, scale_kwidth=8
+    )
+
+    x_static_scale = x_physical.abs().max().float() / 448.0
```
valarLip: no hard code 448

valarLip (changes requested):
the other part LGTM

op_tests/triton_tests/moe/test_moe_gemm_a8w4.py
xiaohuguo2023: done, thanks @valarLip

PR #4110, opened 2026-07-07
Support emit_bf16 for per-token fused AllReduce+RMSNorm+FP8 quant

gbyu-amd:
## Motivation
The fused AllReduce+RMSNorm+FP8 quant path already supports emit_bf16 for the per-group and mxfp4 cases (emitting the pre-quantization bf16/fp16 normed activation alongside the fp8 output). Per-token FP8 was the only remaining case, and previously raised. Enable it here.

This is required by v32 DSA models (e.g. GLM-5.2) whose indexer GEMMs run in bf16 and therefore need the unquantized normed hidden state, while attention QKV keeps per-token FP8.

HIP kernels (csrc):
- ar_fusion_epilogue: add optional bf16_output; in the fp8 branch downcast the float normed result and store it when the pointer is non-null (mirrors ar_fusion_epilogue_per_group / _mxfp4).
- allreduce_fusion_kernel_1stage / _2stage and the naive split kernel plus their launchers: thread bf16_output through.
- dispatchFusedAllReduceRMSNormQuant and its launch macros: thread through.
- fused_allreduce_rmsnorm_quant wrapper: add bf16_out_ptr (opaque pointer, 0 = not requested); _fused_allreduce_rmsnorm helper forwards it.

Python:
- custom_all_reduce op stub / rocm_ops binding: add bf16_out_ptr.
- fused_ar_rms / custom_fused_ar_rms_quant: add emit_bf16, allocate the bf16 output, forward the pointer, return the 4-tuple when requested (incl. the capture-dummy path).
- communicator_cuda: remove the per-token emit_bf16 ValueError; return the 4-tuple from both the fused path and the AR+RMS-then-quant fallback.
- parallel_state: add a graph-traceable fused_allreduce_rmsnorm_quant_emit_bf16_ op + fake (4 outputs) and dispatch to it for per-token emit_bf16 so it lowers into the compiled graph (Dynamo-traceable + CUDAGraph-capturable). Also fix the pre-existing per-token fake to declare out as FP8, not bf16.

## Submission Checklist

### DIFF
--- a/aiter/dist/device_communicators/communicator_cuda.py
+++ b/aiter/dist/device_communicators/communicator_cuda.py
@@ -379,8 +379,11 @@ class CudaCommunicator(DeviceCommunicatorBase):
                 prefill_support=prefill_support,
                 emit_bf16=emit_bf16,
             )
-        if emit_bf16:
-            raise ValueError("emit_bf16 is not supported for per-token FP8 quant")
+        # emit_bf16 additionally returns the pre-quantization bf16/fp16 normed
+        # output alongside the per-token FP8 result. Used by v32 DSA models
+        # (e.g. GLM-5.2) whose indexer GEMMs run in bf16 while attention QKV
+        # keeps per-token FP8.
+        bf16_out = None
         hidden_dim = int(input_.shape[-1])
         element_size = input_.element_size()
         total_bytes = input_.numel() * element_size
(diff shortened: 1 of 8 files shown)


valarLip:
any perf number?

gbyu-amd:
> any perf number?

Haven't run full E2E test, from the trace, it does take effect:
 
image
image

valarLip:
please have a full ut results

gbyu-amd:
GLM-5.2-MXFP4 with attn PTPC fp8
Before (only allreduce+rmsnorm fusion)
image
After (fuse allreduce+rmsnorm+per_token_quant as one kernel)
image

@valarLip Here is the perf number. Just save ~5us (a separate quant kernel) per layer, aligned with the E2E results.

valarLip (approved):

PR #4114, opened 2026-07-07
[HIP] [OPUS] [FlyDSL] FlyDSL gemm_decode: small-M dense GEMM kernels (BF16)

vedenev-amd:
## Motivation

LLM autoregressive decode runs linear projections at M=1..5 tokens per step. A standard 16×16 MFMA tile wastes ≥75% of compute slots at these sizes. This PR adds a FlyDSL BF16 decode GEMM optimised for small M, plus the tuned config rows for the shapes where it wins.

Target hardware: gfx942 and gfx950. Measured on MI325X (gfx942, 304 CU) and MI355X (gfx950, 256 CU).
Target shapes: Kimi-K3 TP8 decode GEMMs, N=896…40960, K=128…8448. Generic — no model-specific names in kernel or dispatch code.

## Technical Details

Two complementary FlyDSL policies, selected by a runtime tuner:

**Wave** (`gemm_decode_wave.py`) — warp-per-scalar output-centric GEMM. One wavefront owns one tile of (M×N) output scalars; K split across 64 lanes, reduced by butterfly XOR. `v_dot2_f32_bf16` on gfx950, scalar/packed FP32 on gfx942. Wins at M=1,2.

**BlockMFMA** (`gemm_decode_block_mfma.py`) — MFMA block policy with LDS A-staging and software prefetch. Wins at M≥3.

**Entry point** (`gemm_kernels.py`): `gemm_decode_bf16(A, B, C, config)`, dispatching to either policy via a tuned config from `iter_gemm_decode_configs`. The kernel exists only for M≤5; the dispatcher honours a `flydsl_decode` row on exact M match only, never through M padding.

**Tuned configs** — `flydsl_decode` rows in `kimik3_bf16_tuned_gemm.csv` for gfx950 and gfx942, M=1..5, covering the Kimi-K3 TP8 decode shapes. Most land where the dispatcher previously fell through to `torch`. On gfx950 the rest replace an existing `flydsl` or `triton` row, and each of those only where graph replay showed decode beats the incumbent for that exact shape and M. On gfx942 nothing is replaced — see below.

**Duplicate shapes across model configs.** A few keys at `(2112,7168)` also existed in `dsv3_` and `kimik2_bf16_tuned_gemm.csv`, and `(3584,7168)` arrived in main via #5702 while this PR was open. Since the runtime merges every `*bf16_tuned_gemm*` file into one keyspace, duplicates raise at merge time. All were resolved by the repository's own guard — `test_config_shape_collision.py --fix`, lowest-`us` wins — not by hand. That accounts for the deletions in the two other model files.

The displaced rows were measured directly against the decode rows that replaced
them — both kernels by name, one process, packed HIP graph, alternating pairs,
null control ≤ 0.13%:

| M | displaced from | decode, µs | displaced `flydsl`, µs | gain | pairs won |
|---|---|---|---|---|---|
| 1 | dsv3 | 6.265 | 8.177 | +23.4% | 8359 / 8360 |
| 2 | dsv3 | 6.808 | 8.030 | +15.2% | 8349 / 8350 |
| 4 | kimik2 | 6.644 | 8.326 | +20.2% | 8320 / 8320 |

Max relative error vs FP32 is also lower for decode (2.8–3.5e-03 vs
8.6–8.9e-03), so dsv3 and kimik2 gain accuracy on this shape as well as speed.

**Two fixes needed along the way.** The AOT path applied the bias placeholder substitution twice, failing every bias-free row. The tuner's candidate selection discarded valid decode configs before benchmarking them.

## Performance

Measured against the kernel the config named for the same shape before this PR — the existing row, or `torch` where none existed. Both kernels invoked by name, dispatcher out of the loop.

Packed HIP graph, wall clock over launches, four rotating weights, alternating A/B/B/A pairs scored per pair, incumbent timed against itself as a null control (2% threshold). Both sides checked against an FP32 reference before each timing.

gfx950 (MI355X, 256 CU), every row in this PR, **3 independent passes**:

| M | cells | better | tied | worse | ours, µs | previous, µs | gain |
| 1 | 20 | 19 | 1 | 0 | 306.0 | 404.6 | +24.4% |
| 2 | 17 | 17 | 0 | 0 | 299.3 | 379.8 | +21.2% |
| 3 | 18 | 18 | 0 | 0 | 307.7 | 406.5 | +24.3% |
| 4 | 14 | 14 | 0 | 0 | 276.9 | 340.2 | +18.6% |
| 5 | 13 | 13 | 0 | 0 | 270.4 | 329.5 | +17.9% |
| **all** | **82** | **81** | **1** | **0** | **1460.4** | **1860.6** | **+21.5%** |

**Every row confirmed in every pass, zero regressions.** The one tie is `(7168,1536)` at M=1, where the kernels sit 0.3% apart; it never lost a pass.

An earlier revision of this description quoted −9%…−25% against `wvSplitK`. Those were taken against a September-2025 build of that kernel and are withdrawn.

**Downstream:** the same rows measured end to end through vLLM's `rocm_unquantized_gemm_impl` give +4.7% on M=1..4 against `wvSplitK` — a kernel aiter cannot select, as it lives in vLLM's own extension. Wiring and caveats in [vllm-project/vllm#55860](https://github.com/vllm-project/vllm/pull/55860).

gfx942 (MI325X, 304 CU), every row in this PR, **3 independent passes**:

| M | cells | better | tied | worse | ours, µs | previous, µs | gain |
|---|-------|--------|------|-------|----------|--------------|------|
| 1 | 21 | 21 | 0 | 0 | 419.1 | 560.9 | +25.3% |
| 2 | 21 | 21 | 0 | 0 | 436.6 | 562.8 | +22.4% |
| 3 | 19 | 19 | 0 | 0 | 444.9 | 550.6 | +19.2% |
| 4 | 19 | 19 | 0 | 0 | 421.3 | 516.5 | +18.4% |
| 5 | 11 | 11 | 0 | 0 | 376.3 | 423.0 | +11.0% |
| **all** | **91** | **91** | **0** | **0** | **2098.1** | **2613.8** | **+19.7%** |

**Every row confirmed in every pass, zero regressions, no batch discarded by the null control.** Here `previous` is always the heuristic fallback, which resolved to `torch`: at the branch point no bf16 config carried a `gfx942` row with `cu_num=304`, so all 91 rows are additions and none replaces an existing row. The rows are keyed `(gfx942, cu_num=304)`, which also matches MI300X; that part was not re-measured.

**Shapes deliberately left without a row.** 11 cells at M=3,4,5 where `opus` is faster — `router` at M=5 by 22%, `(7168,8448)` at M=5 by 9%, six other long-K shapes at M=5 by 0.5…6%, `router` and `shared down` at M=3, `latent down` at M=4 — confirmed on two cards with 0–2 wins out of 90 pairs. Plus 3 cells where the tuner and graph replay disagree reproducibly (`(7168,8448)` at M=4, `attention O` and `(4608,1536)` at M=5): five tuner runs rank `opus` first, graph replay ranks the decode kernel first by up to 1520/1520 pairs.

**Full evidence** — protocol, benches, raw records, and the measurement errors made and corrected along the way — is attached as `pr4114_evidence.tar.gz`.

## Architecture support

- **gfx950**: Wave uses `v_dot2_f32_bf16`; BlockMFMA uses MFMA with two-stage prefetch
- **gfx942**: Wave uses scalar_f32 contraction; BlockMFMA uses single-stage prefetch; both paths correctness-tested

## Test Plan

- `op_tests/test_flydsl_decode_gemm.py` — a standard op_test, plain script, no pytest. `python3 <file>` runs the correctness checks first, then the perf sweep with `checkAllclose` against an FP32 reference:
  - five kernel cases: Wave no-bias, Wave with bias and odd N/K tails, BlockMFMA global, BlockMFMA full LDS with K padding, BlockMFMA persistent-N multi-turn;
  - every shipped `flydsl_decode` row for the running card must be selected by the dispatcher, and one shape per M runs end to end through `tgemm.mm`. The dispatcher drops a row with a malformed `kernelName` silently, so without this a broken row would pass CI.
  No CI script changes are needed — the file sits where the runner already scans.
- `python3 -m unittest op_tests.tuning_tests.test_config_shape_collision` — `Ran 17 tests ... OK`. The config keyspace stays collision-free across every model file; includes a positive control (a planted duplicate must be caught).
- Row-level verification: every row confirmed over 3 independent passes, zero regressions (raw records in the attached archive).
- Black and Ruff 0.16.0 (the CI pin) clean on all changed files.
- CI on MI300X (gfx942) and MI355X (gfx950).

## Attachment
[pr4114_evidence.tar.gz](https://github.com/user-attachments/files/32519718/pr4114_evidence.tar.gz)

## Submission Checklist


aiter/ops/flydsl/kernels/gemm_decode.py
```diff
@@ -0,0 +1,482 @@
...
+        A_prod.normal_()
+        st.record(); runner.run(A_prod, B_prod, M=2); en.record()
+        torch.cuda.synchronize()
+        t = st.elapsed_time(en) * 1000
+        if t > 5: times_graph.append(t)
+
+    times_direct = []
+    for _ in range(30):
+        evict()
+        st = torch.cuda.Event(enable_timing=True); en = torch.cuda.Event(enable_timing=True)
```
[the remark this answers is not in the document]
vedenev-amd: Curent results:
```
────────────────────────────────────────────────────────────────────────────────
  Shape              M    FlyDSL µs   FlyDSL min   CK µs   CK min    gap
────────────────────────────────────────────────────────────────────────────────

N=16384  K=7168   (DeepSeek-V3 Q/K/V/O-proj)
                   1       177.3        177.3      176.2    176.2   1.006×
                   2       179.4        179.4      176.7    176.3   1.015×
                   3       185.8        185.8      177.2    176.3   1.049×
                   4       192.4        192.4      179.3    176.3   1.073×

N=18432  K=7168   (DeepSeek-V3 gate/up-proj)
                   1       198.9        198.9      198.0    198.0   1.005×
                   2       201.4        201.3      198.5    198.0   1.014×
                   3       208.9        208.9      199.0    198.0   1.049×
                   4       217.0        216.9      201.4    198.0   1.078×

N= 7168  K=18432  (DeepSeek-V3 down-proj)
                   1       229.9        229.4      200.5    200.5   1.147×
                   2       251.0        250.4      200.7    199.1   1.251×
                   3       262.9        262.1      202.3    199.1   1.299×
                   4       273.2        272.5      213.5    200.4   1.280×

N=14336  K=4096   (Llama-3 8B gate/up-proj)
                   1        90.5         90.5       89.2     89.2   1.015×
                   2        91.6         91.6       89.5     89.2   1.023×
                   3        94.5         94.5       89.9     89.2   1.052×
                   4        97.6         97.6       92.1     89.2   1.060×

────────────────────────────────────────────────────────────────────────────────
```

.github/scripts/aiter_test.sh
```diff
@@ -84,6 +84,9 @@ for file in "${sharded_files[@]}"; do
     # batch gate so they exercise the persistent kernel at every batch size.
     test_cmd=(timeout 60m python3 "$file")
     case "$file" in
+        op_tests/flydsl_tests/test_flydsl_decode_gemm.py|op_tests/flydsl_tests/test_flydsl_small_m_hgemm.py)
```
vpietila-amd: Is the intent to get the tests executed as part of the CI run? You also move the test files one level up to `op_tests/`. The CI run should automatically pick them up. There are also other FlyDSL tests at that level.

.github/scripts/split_tests.sh
```diff
@@ -51,6 +51,15 @@ TEST_DIR="${TEST_DIR%/}"
 # ------------------------------
 if [[ "$TEST_TYPE" == "aiter" ]]; then
     mapfile -t ALL_FILES < <(find "$TEST_DIR" -maxdepth 1 -name 'test_*.py' -type f | LC_ALL=C sort)
+    FLYDSL_GEMM_TESTS=(
```
vpietila-amd: Moving the test files one level up would get rid of this modification.

.github/scripts/aiter_test.sh
vpietila-amd: I guess you need this because you want to run them as under pytest. Instead of adding them here as an exception, you could change the test scripts (.py files)

```
if __name__ == "__main__":
      import sys

      if "--bench" in sys.argv:
          sys.argv.remove("--bench")
          main()
      else:
          sys.exit(pytest.main([__file__] + sys.argv[1:]))
```

When the number of FlyDSL tests increases, it would be better that the tests are self-governing, i.e., define themselves how they want to be executed.

PR #4120, opened 2026-07-07
CI: sync ATOM Kimi default model name

yifehuan:
## Motivation

  The `ci:atom` workflow currently fails in `load-atom-models` before running any jobs because AITER still requests the old ATOM default model name `Kimi-K2.5-MXFP4`. ATOM main replaced that exact entry with `Kimi-K2.7-Code-MXFP4`, so the workflow
  matrix loader exits with `Missing default model(s) in ATOM accuracy config: Kimi-K2.5-MXFP4`.

  ## Technical Details

  - Update `.github/workflows/atom-test.yaml` default ATOM Kimi model from `Kimi-K2.5-MXFP4` to `Kimi-K2.7-Code-MXFP4`.
  - Update the matching per-model runner override key.
  - Keep the existing TP4 runner pin behavior.

  ## Test Plan

  - Run `git diff --check`.
  - Simulate the workflow exact-match lookup against `ROCm/ATOM@main` `.github/benchmark/models_accuracy.json`.

  ## Test Result

  - `git diff --check`: passed.
  - Default model lookup against ATOM main: passed; missing model list is empty.

  ## Submission Checklist

### DIFF
--- a/.github/workflows/atom-test.yaml
+++ b/.github/workflows/atom-test.yaml
@@ -84,9 +84,9 @@ jobs:
           default_models = [
               {"model_name": "DeepSeek-R1-0528"},
               {"model_name": "gpt-oss-120b"},
-              # Kimi-K2.5 ATOM (in-tree) downstream gate. Config (model_path,
+              # Kimi-K2.7 ATOM (in-tree) downstream gate. Config (model_path,
               # extraArgs -tp4, threshold) is pulled from ATOM models_accuracy.json.
-              {"model_name": "Kimi-K2.5-MXFP4"},
+              {"model_name": "Kimi-K2.7-Code-MXFP4"},
           ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/atom-test.yaml
gyohuangxin: The ATOM accuracy job name is too long because it includes full `extraArgs`, including the Kimi `--online_quant_config` JSON.

Can we keep the job title short, e.g. `Accuracy (${{ matrix.display_name }}, ${{ matrix.runner }})`, and move full args to the log or `$GITHUB_STEP_SUMMARY`.

image

gyohuangxin (changes requested):

.github/workflows/atom-test.yaml
yifehuan: fixed,

gyohuangxin (approved):

PR #4144, opened 2026-07-08
[FIX] [MLA] Gate persistent MLA decode kernel by batch size

jamesETsmith:
## Motivation

For the **gfx950 bf16 / 16-head / single-token decode profile** that vLLM's ROCm AITER MLA backend uses (`page_size=1`, `kv_lora_rank=512`), `aiter/mla.py::mla_decode_fwd` picks one of two asm decode kernels based on whether the caller passes persistent metadata (`work_meta_data`):

- **A — non-persistent** split-KV (`mla_dec_stage1_bf16_a16w16_subQ16_mqa16`): `work_meta_data is None`; rebuilds its own `num_kv_splits`.
- **B — persistent** (`mla_a16w16_qh16_m16x4_n16x1_coex0_mask1_ps`): `work_meta_data is not None`.

vLLM ≤ 0.18 never built `work_meta_data` (always kernel A). v0.19+ builds and passes it **unconditionally for every single-token decode step** (always kernel B). B is competitive at light load but roughly **1.4–1.6× slower than A under high concurrency / long context** on MI350X (gfx950), regressing decode throughput ([vllm-project/vllm#45943](https://github.com/vllm-project/vllm/issues/45943)).

This PR gates the dispatch so kernel A is recovered where B regresses, leaving B where it wins. Target profile: MI350X gfx950, bf16 q+kv, 16 query heads (TP MLA), `kv_lora_rank=512`, `qk_rope=64`, `page_size=1`, `decode_qlen=1`.

Credit for tracking down the kernel that slowed down goes to @xaguilar-amd who also did the first set of microbenchmarks over batch and context size.

## Technical Details

The gate keys on **batch size (`num_seqs`)**: keep the caller's persistent request below a threshold, and fall back to the non-persistent split-KV kernel at/above it. The benchmark grid where we compare kernels A and B (see below) shows **concurrency (batch) is the dominant axis** of the penalty — persistent is never faster once batch ≥ ~32.

```python
_MLA_DECODE_PERSISTENT_MAX_BATCH_DEFAULT = 32

def _use_persistent_mla_decode(bs, nhead, max_seqlen_q, q_dtype, kv_dtype):
    is_regression_profile = (
        get_gfx() == "gfx950"
        and q_dtype == dtypes.bf16 and kv_dtype == dtypes.bf16
        and nhead == 16 and max_seqlen_q == 1
    )
    if not is_regression_profile:
        return True
    max_batch = int(os.getenv("AITER_MLA_DECODE_PERSISTENT_MAX_BATCH",
                              _MLA_DECODE_PERSISTENT_MAX_BATCH_DEFAULT))
    if max_batch <= 0:      # disable gate -> always persistent
        return True
    return bs < max_batch
```

Call site in `mla_decode_fwd`:

```python
persistent_mode = work_meta_data is not None
if persistent_mode:
    persistent_mode = _use_persistent_mla_decode(
        bs, nhead, max_seqlen_q, q.dtype, kv_buffer.dtype
    )
```

- **Env knob `AITER_MLA_DECODE_PERSISTENT_MAX_BATCH`** (default `32`): retune per workload, or set `0` to disable the gate (always-persistent). Runtime kill-switch, no rebuild.
- **Narrow scope**: only `gfx950` + bf16 q+kv + `nhead == 16` + `decode_qlen == 1`. Everything else keeps the caller's choice, so fp8, gqa=32/64/128, and sparse persistent paths are untouched.
- **Pure Python**, no kernel recompilation.
- The batch-32 default is **provisional/heuristic**, pending microbench calibration (see open question below).

## Test Plan

All on MI350X gfx950, image `vllm/vllm-openai-rocm:v0.24.0`.

### a. Stock op-tests

AITER's MLA op-tests, one per kernel (A = `test_mla.py`, B = `test_mla_persistent.py`), over the grid `batch {1,4,16,32,64,128,256} × ctx {1000,4000,8000,16000,50000}`. Timing uses AITER's `run_perftest` (cache-busting argument rotation between iterations).

```bash
COMMON="-n 16,1 -d bf16 -kvd bf16 -blk 1 -k 512 -qr 64 -vh 512 \
  -b 1 4 16 32 64 128 256 -c 1000 4000 8000 16000 50000"

python op_tests/test_mla.py $COMMON             # kernel A (non-persistent)
python op_tests/test_mla_persistent.py $COMMON  # kernel B (persistent)
```

### b. Microbench (Recreating vLLM env where we first noticed this)

A standalone microbench that reproduces vLLM's `ROCM_AITER_MLA` decode call exactly: build persistent metadata via `aiter.get_mla_metadata_v1(...)` and invoke `aiter.mla.mla_decode_fwd(...)` with it, timed with CUDA events.

```python
# Build persistent metadata like rocm_aiter_mla.py::_build_decode
sizes = aiter.get_mla_metadata_info_v1(batch, 1, nhead, q_dt, kv_dt,
                                       is_sparse=False, fast_mode=True)
bufs = [torch.empty(sz, dtype=dt, device="cuda") for (sz, dt) in sizes]
aiter.get_mla_metadata_v1(qo_indptr, kv_indptr, kv_last_page_len, nhead, 1,
                          True, *bufs, page_size=1, kv_granularity=16,
                          max_seqlen_qo=1, uni_seqlen_qo=1, fast_mode=True)
work_meta_data, work_indptr, work_info_set, \
    reduce_indptr, reduce_final_map, reduce_partial_map = bufs

def call():  # identical to vLLM's decode forward
    aiter.mla.mla_decode_fwd(q, kv, o, qo_indptr, kv_indptr, kv_indices,
                             kv_last_page_len, 1, 1, 1, sm_scale=sm_scale,
                             work_meta_data=work_meta_data, work_indptr=work_indptr,
                             work_info_set=work_info_set, reduce_indptr=reduce_indptr,
                             reduce_final_map=reduce_final_map,
                             reduce_partial_map=reduce_partial_map)

# median latency over N iters via per-iter CUDA events (warmup first)
```

### c. Gate unit test

A pure-pytest, decision-level test of `_use_persistent_mla_decode` (no GPU metadata, no dispatch spies) covering the default threshold, the env-var override (lower / raise / disable), and out-of-scope inputs (fp8, `nhead != 16`, `qlen != 1`) that must stay persistent.

```bash
pytest op_tests/test_mla_decode_gate.py
# CI-script style also works:
python op_tests/test_mla_decode_gate.py
```

## Test Result

Side-by-side B/A decode-latency heatmaps (ratio > 1 ⇒ persistent slower):

BA-optest-vs-microbench

Both harnesses **agree on the ~1.4–1.6× persistent penalty at high batch / long context** (the regression region), where the batch gate selects the faster kernel (A); persistent is **never faster once batch ≥ ~32**. They **diverge at low load**: the op-test shows ~parity, while the microbench shows < 1.0 (persistent looks *faster*) for small batches.

### Gate unit test

```text
collected 5 items

op_tests/test_mla_decode_gate.py::test_defaults PASSED                   [ 20%]
op_tests/test_mla_decode_gate.py::test_env_lowers_threshold PASSED       [ 40%]
op_tests/test_mla_decode_gate.py::test_env_disabled PASSED               [ 60%]
op_tests/test_mla_decode_gate.py::test_env_raises_threshold PASSED       [ 80%]
op_tests/test_mla_decode_gate.py::test_out_of_scope PASSED               [100%]

============================== 5 passed in 2.77s ===============================
```

### Existing op-tests

- **`test_mla.py` (kernel A)** never passes `work_meta_data`, so the gate is never consulted — decode is unaffected and `checkAllclose` passes (loads `mla_dec_stage1_bf16_a16w16_subQ16_mqa16`).
- **`test_mla_persistent.py` (kernel B)** is in-scope, so the default gate changes its behavior: batch < 32 keeps kernel B and all `checkAllclose` pass; batch ≥ 32 redirects to kernel A. On redirect the attention is still correct (`golden vs aiter_asm` and `split_out_ref vs aiter_asm` pass), but the test's persistent-only `partial_out_ref vs attn_logits` check then errors on a shape mismatch — it assumes the persistent partial-logits layout, which no longer exists after the fallback. This is a test assumption, not an attention-correctness failure. Setting **`AITER_MLA_DECODE_PERSISTENT_MAX_BATCH=0`** disables the gate so the test exercises kernel B at all batches and passes cleanly; CI covering this op-test in-profile should set that env var.

### Open question (maintainer input requested)

We do not yet know why the harnesses disagree at low load. Two hypotheses are ruled out: **cache residency** (a cold-cache microbench variant rotating KV buffers like `run_perftest` showed negligible effect) and **split policy** (`num_kv_splits=None` vs forcing `32` in the op-test did not reproduce the microbench's slow low-batch kernel-A numbers). Our leaning is a fixed per-call/measurement overhead inflating the tiny low-load kernels, but this is unresolved. **We'd appreciate AITER maintainers' input**: which harness better reflects production MLA decode, and is the low-batch kernel-A behavior expected?

## Submission Checklist

This is my first PR to AITER so please let me know if I'm missing anything here.


shengnxu (dismissed):

jamesETsmith:
I fixed the failing CI tests. Several `test_mla_persistent.py` tests were failing because they used batches of 64 and per my changes triggered non-persistent mla kernels. The results were correct from the mla kernels in those tests, but the outputs were mismatched shapes because they were returned from non-persistent kernels. I've updated the tests to always use persistent kernels in `test_mla_persistent.py` via `AITER_MLA_DECODE_PERSISTENT_MAX_BATCH=0 `

jamesETsmith:
@shengnxu I did some local testing and I don't think the single CI failures are related to my changes. I've updated with main and restarted the CI in case they're fixed by recent changes.

aiter/mla.py
```diff
@@ -194,6 +195,48 @@ def get_meta_param(
...
+# Persistent MLA-decode kernel gate: the persistent kernel
+# ("mla_a16w16_qh16..._ps") is slower than the non-persistent split-KV kernel
+# ("mla_dec_stage1...") above a concurrency threshold (~batch 16-64 on gfx950 bf16
+# 16-head decode). Fall back to non-persistent at/above the batch threshold.
+# Env-tunable (0 disables the gate). Provisional default pending microbench
+# calibration (AIOSS-5156).
+_MLA_DECODE_PERSISTENT_MAX_BATCH_DEFAULT = 32
+
+
+def _use_persistent_mla_decode(bs, nhead, max_seqlen_q, q_dtype, kv_dtype):
```
valarLip, line 221: add lru cache here to prevent get env every call

valarLip (changes requested):

aiter/mla.py
jamesETsmith: Good catch, thanks! Should be fixed

jamesETsmith:
I narrowed the gating even more so it only will take dispatch to the older kernel when it's dense mla (i.e. `num_kv_splits == None`). Hopefully this will correct some of the failing tests, although I think some of them are unrelated to my changes

PR #4171, opened 2026-07-10
review-pr skill v2: dispatch gap rule, arch-string FP fix, P5 timing error, D10b FlyDSL

zufayu:
## Summary

Second iteration of the `review-pr` skill, based on a 191-PR test sweep (88 aiter + 93 ATOM open PRs). Key findings from that sweep drove all rule changes here.

### New rules

**B4 — New dispatch value not handled by all paths** (highest-frequency finding, ~18% of PRs)
When a PR adds a new dtype/arch/flag to a multi-way dispatch, every reachable branch must handle it or assert/warn on unsupported combos. Without this rule the pattern was being caught inconsistently under B1 or missed entirely. Includes FP self-check (upstream assert → skip). Old B4 (over-conservative assert) moved to B7.
- Real examples: aiter#4169 (GGUU flag not wired to gfx950 Triton path), aiter#3941 (fp4 falls through fp8/int8-only tile table), aiter#3939 (cross-attention + mt=1 on gfx1250 reaches missing kernel)

**C4 — New GPU arch string literal in dispatch condition**
FP self-check is front-loaded: search unchanged lines for same arch string first; if found → skip (pre-existing style, not a new violation). Fires only on genuinely new introductions.
- Eliminates soft FPs from aiter#3956 / aiter#3937 where agents incorrectly fired on pre-existing codebase style

**D10b — FlyDSL arith.bitcast requires arith.unwrap() on operand**
Raw DSL value passed to `arith.bitcast()` causes JIT type error at first invocation of that shape/dtype combo — silent in Python, crashes at kernel JIT time.
- Real example: aiter#3944 bf16/f16 output path

**P5 — Benchmark timing excludes one-time setup cost**
Distinct from P1 (missing numbers). Covers cases where shuffle_weight, first-call JIT compile, or precompile are excluded from the timing window — making a regression look like a speedup.
- aiter#4166: claimed 1.14x win → ~0.83x regression when shuffle_weight included
- aiter#3944: 7 precompile kernels launched on live stream, not in benchmark

### Improved rules

**C4 FP self-check front-loaded**: moved before the trigger condition so agents are forced to verify pre-existing style before firing (previously placed at the end, often skipped)

**Step 3 classification**: wired B4/P5/D10b into trigger lines — new constexpr/routing-flag PRs and FlyDSL PRs now auto-prompt the relevant checks

**Step 8 output format**: findings now require three parts — Problem + Impact + Action verb ("Author must" / "Reviewer should ask"). Findings without an action verb are not included in output.

## Test plan

- [x] 191 PRs reviewed across aiter + ATOM with prior skill version; all rule additions motivated by real findings from that sweep
- [x] B4 trigger validated against aiter#4169, #3941, #3939 (true positives)
- [x] C4 FP self-check validated against aiter#3956 (correctly skips pre-existing style)
- [x] P5 validated against aiter#4166 (measurement error confirmed by manual diff analysis)
- [x] D10b validated against aiter#3944 (arith.unwrap pattern confirmed in diff)


junhaha666 (dismissed):

zufayu:
Merge button dropdown → "Merge without waiting for requirements to be met". Thanks!  @valarLip

zufayu:
@junhaha666 heads-up — I pushed a v3 update to this skill, so the new commit dismissed your earlier approval. Could you re-approve when you get a chance?

What changed since your approval (all distilled from today's real reviews):
- **E4** — downstream CI coverage: a change to a downstream-consumed op can pass aiter CI with `Atom Test` skipped and break ATOM after merge (aiter#3459); adds a dispatch-reachability → model → minimal `ci:*` label triage.
- **E5** — stable core-API (fused_moe/mla/attention/quant/jit-core) behavior/dispatch changes need de-facto owner sign-off before merge, not a post-merge revert (aiter#3593).
- Plus cross-file verification + CI-failure triage + [verified]/[inferred] finding discipline.

CI is running; once it's green + your re-approval, it's ready to merge. Thanks!

zufayu:
@valarLip This PR is stuck: the required check `Standard Test Results` stays pending forever. It only changes a `.md` file, which is excluded by the `!**/*.md` paths filter in `aiter-test.yaml`, so the workflow never triggers and the required check never reports → BLOCKED. This is a required-check + paths-filter deadlock a doc-only PR cannot satisfy on its own. Could you bypass-merge it? (Also needs one approval after the last push.)

valarLip (approved):

PR #4180, opened 2026-07-10
feat(gfx950): config-gated BLOCK_Q fp8_mqa_logits for DSA indexer prefill

YukioZzz:
## Motivation
Add a BLOCK_Q Gluon specialization of fp8_mqa_logits for the gfx950 (CDNA4) DSA / lightning-indexer prefill shape (short query x long KV) under context-parallel ragged ranges, as used by GLM-5.x / DeepSeek-V3.2 sparse attention. It shares one double-buffered KV tile across 2 adjacent query rows, reusing the existing gfx950 building blocks (MQAAsyncKVLoader / _mqa_dot / _weighted_sum_fma_fold / _load / _store), so the MFMA layout and head reduction match the stock single-row path. General ragged (reads real cu_start/cu_end), handles an odd tail row, writes only the valid [start,end) span (clean_logits=False).

## Technical Details
fp8_mqa_logits() dispatches via a tuned config table keyed by (num_heads, head_dim); unsupported shapes fall back to the stock row-per-program kernel. Design note vs DeepGEMM/TileLang: those fix UMMA_N=128 (BLOCK_Q=128/num_heads) since the accumulator lives in Tensor Memory; on gfx950 the MFMA accumulator is in registers, so the fill target is one MFMA M-tile (32) and KV-share / BLOCK_KV / waves_per_eu are separate occupancy knobs - over-packing rows loses here.

## Test Result

Validated on MI355X (Q=6156,K=49243,CP ke=1+row*8), median vs stock, all max_abs<=3e-5: (32,128) 1.35x, (16,128) 1.45x, (8,128) 1.48x, (32,64) 1.23x, (16,64) 1.33x. num_heads==64 spans 2 M-tiles (register pressure) and loses (0.92x), so it is excluded and falls back. E2E validated through the top-level fp8_mqa_logits.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/_gluon_kernels/gfx950/attention/fp8_mqa_logits_blockq.py
+++ b/aiter/ops/triton/_gluon_kernels/gfx950/attention/fp8_mqa_logits_blockq.py
@@ -1,0 +1,288 @@
+"""BLOCK_Q fp8 MQA-logits Gluon kernel for gfx950 (CDNA4 / MI350X-MI355X).
+
+Specialization of the DSA / lightning-indexer ``fp8_mqa_logits`` for the
+short-query x long-KV prefill shapes used by GLM-5.x / DeepSeek-V3.2 sparse
+attention under context-parallel (CP) ragged ranges.
+
+Design (vs the CUDA/DeepGEMM sm100 kernel and the TileLang kernel)
+-----------------------------------------------------------------
+DeepGEMM (B200) and TileLang both parametrize by head count: ``BLOCK_Q =
+128 / num_heads`` so that ``UMMA_N = BLOCK_Q * num_heads == 128`` fills one
+Blackwell tcgen05 MMA N-tile. Their accumulator lives in Tensor Memory, so
+packing many query rows into that fixed-128 tile is essentially free.
+
+On gfx950 the MMA is a 32x32x64 MFMA whose accumulator lives in *registers*,
+so the invariant is different: the tile to fill is the MFMA **M dimension = 32**,
+not a 128-wide N. One query row's ``num_heads`` occupies the M dimension:
+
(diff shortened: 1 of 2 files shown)


cagrikymk:
@YukioZzz Can you do a head-to-head comparison against the main aiter? The existing kernel has similar idea so I would like to see how it compares.

If you notice there are cases main is not performing, I think it is better to fix the existing kernel instead of creating a specific one unless it is structurally very different.

cagrikymk:
For example, the following patch seems to resolve the gap in a cleaner way:

```
diff --git a/aiter/ops/triton/attention/fp8_mqa_logits.py b/aiter/ops/triton/attention/fp8_mqa_logits.py
index 8d33e29e4..fadf823e2 100644
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -205,6 +205,17 @@ def fp8_mqa_logits(
             block_kv = 64 if num_heads <= 32 else 32
             block_m = 2 if (num_heads <= 32 and seq_len > 4096) else 1
             mfma_nonk_dim = 32 if (head_size <= 64 or num_heads == 32) else 16
+
+            # A 128-wide KV tile with 4 warps on the 32x32x64 MFMA amortizes the
+            # async-copy -> LDS round trip far better than the narrow tiles above.
+            if num_heads > 32 or (head_size > 64 and seq_len_kv >= 32768):
+                block_kv = 128
+                num_warps = 4
+                waves_per_eu = 3
+                mfma_nonk_dim = 32
             other = {
                 "USE_PADDED_SHARED_LAYOUT": ASYNC_COPY_SUPPORTS_DISTRIBUTED,
                 "BLOCK_M": block_m,
```

PR #4181, opened 2026-07-10
[Triton/Gluon] Fix ragged-K mask in batched A16WFP4 GEMM

mjkvaak-amd:
## Motivation

`batched_gemm_a16wfp4` produces incorrect results for ragged logical-K shapes where the FP4 packed weight K dimension is not aligned with the kernel tile assumptions. This was observed in vLLM serving `amd/GLM-5.2-MXFP4`, where MLA decode uses a batched FP4 BMM with logical `K=192`; the outputs of this model are garbage. After this fix, the same vLLM configuration recovers expected GSM8K accuracy.

## Technical Details

In `batched_gemm_a16wfp4`, the kernel argument `K` represents the packed FP4 weight dimension (`logical_K / 2`). The BF16 activation tensor still has logical K elements.
The ragged-K load mask for the BF16 activation currently uses:
https://github.com/ROCm/aiter/blob/3910ab3978d96f37c894de8dff5fb76826ef4e74/aiter/ops/triton/_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py#L192
This incorrectly masks against the packed K dimension and can zero valid activation elements. The non-batched `gemm_a16wfp4` kernel already uses the correct mask (notice the scaling factor `2 * K`):
https://github.com/ROCm/aiter/blob/3910ab3978d96f37c894de8dff5fb76826ef4e74/aiter/ops/triton/_triton_kernels/gemm/basic/gemm_a16wfp4.py#L149
This PR applies the same logical-K mask to the batched kernel.

Fixing the activation mask alone is not enough on the GLM regression shape. `K=96` (packed) rounds `BLOCK_SIZE_K` up to 256, so the kernel loads `BLOCK_SIZE_K / 32 = 8` e8m0 scale groups while `w_scales` only holds `192 / 32 = 6`. Groups 6 and 7 are out of bounds. The matching activation/weight elements are already masked to zero, which would be harmless for a finite scale, but `0xFF` is NaN in e8m0 and `tl.dot_scaled` propagates it through the zeros. The non-batched `gemm_afp4wfp4` kernel already masks those loads with fill value 127 (`2^0`). This PR does the same on the `EVEN_K == False` path:

```python
scale_mask = offs_scale_k[None, :] < (
    2 * K // SCALE_GROUP_SIZE - k * (BLOCK_SIZE_K // SCALE_GROUP_SIZE)
)
b_scales = tl.load(b_scale_ptrs, mask=scale_mask, other=127)
```

Pre-existing test shapes `K=128` and `K=512` stay on `EVEN_K == True` and keep the unmasked scale load.

This PR also adds a regression shape matching the GLM MLA case in `op_tests/triton_tests/gemm/batched/test_batched_gemm_a16wfp4.py`: (B=16, M=32, N=512, K=192).

## Test Plan

* Added a regression case to `op_tests/triton_tests/gemm/batched/test_batched_gemm_a16wfp4.py`.

* Attaching also my manual reproducer used during debugging, which compares 1) BF16 reference, 2) weight-only MXFP4 quant/dequant reference, 3) AITER batched_gemm_a16wfp4 output
```bash
python repro_mla_bmm_quant_error.py --path fp4 --k-dim 192 --n-out 512
```

* Verifying that the fix returns the expected accuracy in the downstream vLLM serving of `amd/GLM-5.2-MXFP4`

## Test Result

Manual reproducer before fix, for X=(16, 32, 192) and W=(16, 512, 192):
```bash
weight_only_ref  rmse=1.54415
aiter_out        rmse=9.85597
aiter-vs-wonly   rmse=9.74326
```
After the activation-mask fix:
```bash
weight_only_ref  rmse=1.54415
aiter_out        rmse=2.18604
aiter-vs-wonly   rmse=1.54492
```

The GLM regression shape then failed Triton CI with NaNs until the OOB scale mask above was added. Reran GSM8K after that change.

vLLM lm-eval before the fix:
```bash
gsm8k flexible-extract exact_match: 0.0 ± 0.0
gsm8k strict-match     exact_match: 0.0 ± 0.0
```

vLLM lm-eval after the fix (activation mask + scale mask):
```bash
gsm8k flexible-extract exact_match: 0.9469 ± 0.0062
gsm8k strict-match     exact_match: 0.9454 ± 0.0063
```

## Submission Checklist

---
Co-authored by: @amd-sriram & Cursor


valarLip (dismissed):

mjkvaak-amd:
I found a problem with the previous arithmetics that the Triton tests successfully flagged. Fixed those and rerun GSM8K: still showing strict 0.9454, flexible 0.9469.

valarLip (approved):

PR #4100, opened 2026-07-06
CI: add ATOM DI CI smoke workflow in Aiter CI

gyohuangxin:
Related to https://github.com/ROCm/frameworks-internal/issues/17175

PR #4182, opened 2026-07-10
CI: add SGLang DSV4Pro FP8 1P1D workflow

gyohuangxin:
## Summary

  Port the SGLang MI355X 2-node 1P1D disaggregation benchmark flow from sgl-project/sglang#29084 into aiter CI, scoped to a single
  model/topology for bring-up:

  - model: `dsv4pro-fp8` (`sgl-project/DeepSeek-V4-Pro-FP8`)
  - topology: `1P1D` with TP8 prefill + TP8 decode
  - benchmark: SGLang `bench_serving` random 1k/1k concurrency sweep
  - correctness gate: optional GSM8K gate enabled by default

  The workflow follows the Spur/Slurm submit-runner override style from ROCm/aiter#4100:

  - runs on `[self-hosted, aiter-di-ci-spur-login]`
  - exposes manual inputs for SGLang ref/image, model path, shared workspace root, Slurm nodelist/partition/exclusive flag,
  concurrency list, and GSM8K gate
  - defaults to the two-node override used for initial bring-up, with `slurm_nodelist=auto` available for scheduler allocation

  ## Implementation Notes

  - Clones the SGLang PR branch by default: `Lzy17/sglang@amd-mi355x-disagg-nightly`.
  - Reuses SGLang's `scripts/ci/slurm/launch_mi355x.sh` and patches it at runtime for aiter's environment.
  - Copies the aiter checkout under test to shared storage and mounts it into prefill/decode containers.
  - Installs that aiter checkout inside each serving container before `sglang.launch_server` starts.
  - Keeps this PR narrow: it does not port the full four-model SGLang nightly matrix yet.

  ## Local Validation

  - `python3 -m py_compile .github/scripts/sglang_mi355x_disagg.py`
  - parsed `.github/workflows/sglang-mi355x-dsv4pro-fp8-1p1d.yaml` with PyYAML
  - patched SGLang PR launcher from local `lzy17/amd-mi355x-disagg-nightly` checkout
  - ran `bash -n` on the patched `launch_mi355x.sh`
  - `git diff --check`

  `actionlint` is not installed in the local environment, so I could not run it locally.


zufayu:
Looked into why [run 32239881701](https://github.com/ROCm/aiter/actions/runs/32239881701) went red, with an eye on the GSM8K path specifically. The accuracy wiring itself is correct — the problems are all around it.

## What actually failed

GitHub flags step 9 `Process SGLang benchmark results`, but the real failure is ~2 seconds inside step 8, and it is a readiness race in the probe:

```
10:37:10  prefill  End of disaggregation warmup / "The server is fired up and ready to roll!"
10:37:31  router   Worker initialization job submitted (will complete in background)
10:37:31  router   Workers initialized: 0 total, 0 healthy
10:37:31  router   Router ready | workers: []          <-- announces ready with an empty worker list
10:37:31  router   Starting server on 0.0.0.0:8000
10:37:32  router   Activated 1 worker(s)               <-- 1P1D needs 2
10:37:32  router   Submitting tokenizer registration job ...
10:37:33  probe    GET /health -> 200                  <-- readiness predicate passes here
10:37:33  probe    [probe] PD end-to-end check via LB
10:37:33  router   POST /generate -> 503 "No prefill workers available"
10:37:33  probe    request failed (curl rc=22); aborting before sweep
```

`/health` only reports that the router process is alive; it says nothing about whether any prefill/decode worker has registered. The router's worker init is an explicit background job, so there is a window where `/health` is 200 and the PD path cannot serve. The probe fires inside that window, once, with no retry.

The servers themselves were fine — prefill completed its disaggregation warmup and served a 200 warmup request 21 s *before* the router even started.

Everything after that is fallout: 0 `raw_conc*.json` → `benchmark failed (rc=1); emitted 0 partial result file(s)` → step 9 `No result files were produced.` → job failure. [verified]

## The accuracy path is wired correctly — don't change it

Checked against `sgl-project/sglang@6174f1c`:

- `bench.accuracy.enabled` exists in the upstream recipe (with `num_shots: 8`, `num_questions: 1319`, `threshold: 0.91`), so `configure_recipe`'s `setdefault("accuracy", {})` overwrites only `enabled` and preserves the rest.
- The launch script parses them correctly — the run log shows `ACC_SHOTS=8 / ACC_NQ=1319 / ACC_THR=0.91`.
- The gsm8k dataset pre-stages on the host with an in-container fallback.
- The gate is genuinely blocking: `sys.exit(0 if acc > thr else 1)`.
- Ordering is right: `launch_mi355x.sh` runs the gate before the sweep ("Correctness gate runs BEFORE the perf sweep").
- `IMAGE_OVERRIDE` does take effect (`v0.5.15.post1`); the `v0.5.13.post1` occurrences earlier in the log are just the recipe text being echoed.

## Findings

**1. The probe gates the accuracy gate, so a probe flake means GSM8K never runs at all.** Accuracy sits behind the probe in `launch_mi355x.sh`, so there is no path to a correctness result while the probe is racy — this is a prerequisite for the GSM8K gate being usable at all, not just a perf-sweep concern. **Author should** replace the `/health` predicate with a wait on the router reporting >=1 healthy prefill *and* >=1 healthy decode worker (or retry `/generate` until it stops returning 503), and give the probe a retry budget. [verified]

**2. A passing GSM8K gate still reports the job as failed when the sweep yields nothing.** Step 9 counts only the perf JSONs:

```bash
if [ "${processed}" -eq 0 ]; then
  echo "No result files were produced."
  exit 1
fi
```

GSM8K's outcome does not participate in that decision at all. For any run where the correctness gate is the point and the sweep is not, the predicate is inverted. **Author should** make the empty-sweep case a warning rather than a hard failure when `RUN_ACCURACY` is on and the gate passed, and consider an accuracy-only mode (right now `RUN_ACCURACY` can turn the gate off, but nothing can turn the sweep off). [verified]

**3. The accuracy number is not reported anywhere a reviewer can see it.** `gsm8k.log` is tee'd into the shared workdir, but `Stage artifacts` only copies `"${GITHUB_WORKSPACE}"/*.json` and `summarize.py` only consumes the perf JSONs. The step summary carries `- GSM8K gate: \`true\`` — that is the switch position, not the result. So even a successful gate leaves the measured accuracy visible only by digging through ~8k lines of job log. **Author should** stage `gsm8k.log` as an artifact and write the measured accuracy and threshold into `GITHUB_STEP_SUMMARY`. [verified]

**4. `continue-on-error` on step 8 hides the real failure.** Step 8 exits 1 with `ERROR: benchmark failed (rc=1)` but is reported as `success`, so the job surfaces the far less informative `No result files were produced.` from step 9 instead. **Author should** either drop `continue-on-error` or have step 9 echo step 8's captured rc and the last lines of `bench.log` before exiting. [verified]

## Unrelated observation

`prefill_rank0` weight load: TP0 took 1671 s while the other seven took ~1473 s. The same straggler pattern is what is currently timing out `Sglang DeepSeek-V3.2 Accuracy (8 GPU)` on `main` (median ~2639 s, worst 6391 s, against a 7200 s cap). Not a blocker here — this run finished in 46 min against a 02:30:00 limit — but worth knowing about if load time ever becomes the budget.

zufayu:
the test in this pr passed on sglang machine but failed in our spur cluster    @functionstackx

PR #4190, opened 2026-07-10
[gfx950][gluon] Correct A8W8 default config to avoid accumulator spill

MrSidims:
256 -> 128 just like for triton config.


MrSidims:
@ROCm/team_aiter folks, does this patch make sense to you?

PR #4192, opened 2026-07-10
tuning a8w8

omuhamma:
## Motivation

Tuning for performance (dsv4 shapes)


k50112113 (dismissed):

vgokhale (approved):

vgokhale:
There is a test fail but it appears to be a flake. 1 element mismatched in MLA which was untouched by this PR. We'll investigate this separately but merging this now.

PR #4207, opened 2026-07-12
op_tests: IFOE cross-node custom all-reduce (module_custom_all_reduce_ifoe)

carlushuang:
## Summary

Adds `module_custom_all_reduce_ifoe`: a custom all-reduce that runs **across nodes** over the UALink-over-Ethernet (IFOE) fabric on gfx1250 by sharing peer buffers with HIP **fabric handles** instead of IPC handles. The 2-stage reduce-scatter/allgather kernel is reused unchanged — only buffer *sharing* differs (IPC → fabric), because on gfx1250 both primitives the all-reduce needs (peer read/write and the `start_sync`/`end_sync` atomic-flag barrier) work cross-node. Handles are exchanged over `torch.distributed`, so the same path serves TP4 (1 node) and TP8 (2 nodes).

## Layout
- `csrc/{include,kernels,pybind}/custom_all_reduce_ifoe.*` — kernels, host API, fabric alloc/import + dispatch, pybind
- `aiter/ops/custom_all_reduce_ifoe.py` — `@compile_ops` stubs
- `aiter/dist/device_communicators/ifoe_custom_all_reduce.py` — `IfoeCustomAllreduce` communicator
- `op_tests/multigpu_tests/ifoe_custom_ar/` — torchrun driver + README

## Modes
- `fp32` — lossless, MLP-unrolled.
- `bf16` — bf16 on the wire (½ the fabric bytes), fp32 accumulate; lossy.
- `fp8` — fp8 e4m3 on the wire (¼ the bytes), fp32 accumulate; lossier, highest throughput (byte size must be a multiple of 64).

## Reproduce
```bash
# TP4 latency sweep across all modes (prints us/iter per size):
torchrun --nproc_per_node=4 test_ifoe_allreduce.py --sweep --bench --modes fp32 bf16 fp8
# TP8: run on each node with --nnodes=2 --node_rank=<0|1> --master_addr=<node0-ip>
```

## Notes
- Raw-pointer / self-contained pybind so the device TU avoids `aiter_tensor.h` → ck_tile (which does not device-compile on gfx1250).
- Requires **ROCm ≥ 7.15** for HIP fabric export (older runtimes return `invalid argument`).
- Python passes `black` + `ruff`.

### DIFF
--- a/aiter/dist/device_communicators/ifoe_custom_all_reduce.py
+++ b/aiter/dist/device_communicators/ifoe_custom_all_reduce.py
@@ -12,6 +12,7 @@ Usage::
     comm = IfoeCustomAllreduce(group, device, max_bytes=1 << 30)
     out = comm.all_reduce(x)                 # fp32, lossless
     out = comm.all_reduce(x, mode="bf16")    # bf16 on the wire (lossy)
+    out = comm.all_reduce(x, mode="fp8")     # fp8 e4m3 on the wire (lossier, fastest)
     comm.dispose()
 """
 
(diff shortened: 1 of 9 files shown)


carlushuang:
WIP

PR #4214, opened 2026-07-13
fix gfx12 ENABLE_Ck0 cmp err

feifei14119:
## Submission Checklist

### DIFF
--- a/csrc/include/aiter_opus_plus.h
+++ b/csrc/include/aiter_opus_plus.h
@@ -9,6 +9,81 @@
 #include <c10/util/Half.h>
 #include <hip/hip_bf16.h>
 
+#if !ENABLE_CK
+// ---------------------------------------------------------------------------
+// Minimal ck_tile:: bridge for the shim build (ENABLE_CK==0).
+//
+// A handful of kernels (e.g. fused_qk_norm_rope_cache_quant.cu and
+// rope/rope_common.h) still spell their scalar dtypes and conversions with the
+// ck_tile:: names. Rather than pulling in the whole Composable Kernel core, we
+// alias those names onto the already-included, arch-tested opus dtypes and
+// route type_convert / numeric through opus::cast / opus::numeric_limits.
+//
+// This header is included right after opus/opus.hpp and before the kernels that
+// need it, so opus:: symbols are visible here. It is guarded by !ENABLE_CK so it
+// never collides with the real ck_tile core when CK is enabled.
+//
+// Every observed call site converts *through* fp32 (IT -> float -> T), which is
+// exactly the set of scalar casts opus defines (X <-> fp32), so no direct
+// non-fp32 pair is ever required.
+namespace ck_tile {
+
+using fp32_t = float;
+using fp16_t = opus::fp16_t;
+using bf16_t = opus::bf16_t;
+using fp8_t  = opus::fp8_t;
+using bf8_t  = opus::bf8_t;
+
+// __device__ only: opus scalar fp8 casts are device-only intrinsics, and every
+// call site in the shim build is inside device/global kernels.
+template <typename D, typename S>
+__device__ inline D type_convert(const S& x)
+{
+    using DD = opus::remove_cvref_t<D>;
+    using SS = opus::remove_cvref_t<S>;
+    if constexpr(std::is_same_v<DD, SS>)
+    {
+        return x;
+    }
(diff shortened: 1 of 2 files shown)


zufayu:

**Adds a minimal `ck_tile::` dtype/type_convert/numeric bridge backed by opus inside `csrc/include/aiter_opus_plus.h` and compiles out the CK-core-dependent body of `csrc/include/ck_tile/vec_convert.h`, so kernels that still spell their scalar conversions with `ck_tile::` names build with ENABLE_CK=0.**

Validation (deterministic): NOT RUN — triage required a target but the PR ships none; the bridge in aiter_opus_plus.h is compiled out entirely in the ENABLE_CK=1 builds CI tests, and its changed path runs at run time only in ENABLE_CK=0 builds of the fused_qk_norm_rope_cache_quant kernel, a module the CK-free build excludes — so this is a runtime change shipping no test target, and no existing test reaches the changed path on any CI arch.
Perf (advisory): NOT RUN — no perf target for the same reason there is no test target (the PR changes runtime code but ships no test target); the CK-free base did not compile before this PR, so no base-vs-head pair exists.

⚠️ [verified] This PR is stale and superseded: the patch no longer applies to main (the context at csrc/include/aiter_opus_plus.h:9 it anchors to was rewritten by later commits), and #4220 (merged 2026-07-20, one week after this PR opened) fixed the same ENABLE_CK=0 build failure on main by deleting the `ck_tile::type_convert`/`ck_tile::numeric` call sites in fused_qk_norm_rope_cache_quant.cu and rope_common.h rather than bridging them; on today's main csrc/include/ck_tile/vec_convert.h has zero includers and nothing references the names this bridge provides, so a rebase would land ~87 lines of dead code. **Reviewer should ask** whether to close this PR in favor of #4220 rather than rebase it.
⚠️ [verified] The fix is incomplete for the module family it targets: the bridge lives in csrc/include/aiter_opus_plus.h, but the file that needs the names — rope_common.h, whose `get_grid_config` default template argument `typename scalar_t = ck_tile::fp16_t` is looked up at parse time even when never used, and whose `from_` calls `ck_tile::type_convert` — does not include it. Of the 18 TUs including rope_common.h at this PR's base, only fused_qk_norm_rope_cache_quant.cu happens to include aiter_opus_plus.h first; fused_qk_norm_mrope_cache_quant.cu and the 16 csrc/kernels/rope/general_*_kernels.cu still fail to compile with ENABLE_CK=0 (verified with a minimal hipcc TU: the unused default argument errors as soon as `ck_tile::fp16_t` is undefined), and all of them sit in the same CK-exclude list as the one module this PR fixes, so whatever on-demand path made that module reachable reaches these too. **Author must** put the dtype aliases and type_convert in a header rope_common.h itself includes (or in rope_common.h) so the fix does not depend on include order.
⚠️ [verified] The PR changes runtime code but ships no test target: the bridge's type_convert/numeric in csrc/include/aiter_opus_plus.h execute at run time in the fused_qk_norm_rope_cache_quant quantization kernel in every ENABLE_CK=0 build, yet nothing in the PR builds or runs a single CK-free TU, and the module is excluded from the standard CK-free build, so no existing test reaches the changed path on any CI arch. **Author must** add a CK-free smoke test (build and run the quant kernel with ENABLE_CK=0) or state on the PR that the change was only compile-tested.
⚠️ [verified] No timing exists for the only path that changes codegen: in ENABLE_CK=0 builds the bridge in csrc/include/aiter_opus_plus.h replaces the ck_tile conversions of the quant kernel with opus::cast intrinsics, but no measurement exists and none is possible as written — the CK-free base did not compile before this PR, and the PR ships no benchmark entry point. **Author must** time the CK-free build of the quant kernel against the CK build of the same commit (the only meaningful comparison available), or state on the PR that no timing was taken.
📝 [verified] `numeric<fp8_t>::max()` is marked `__host__ __device__` but the two compiler passes of a gfx942 build return different values: the host pass does not define `__gfx942__`, so `opus::numeric_limits<fp8_t>` yields the OCP bit pattern 0x7E (verified: host prints 126) while the device pass yields the fnuz pattern 0x7F (=240 as fp8); no host caller exists today because type_convert is `__device__`-only, so this is latent, but a future host consumer of the bridge in aiter_opus_plus.h would silently get 448-range assumptions on a 240-range part. **Reviewer should ask** whether `numeric` needs the `__host__` qualifier at all, or document the pass divergence.

PR #4216, opened 2026-07-13
[OPUS][ATOM] Update sparse prefill attn: fix rescale, zero padding scale and mask oob v

kaiyang-1:
## Summary
Reworks the paged-attention sparse-prefill kernels (bf16 and MXFP8 MLA variants) to fold the softmax temperature into the online-softmax math with FMA, mask out-of-bound KV on the **V** operand in addition to the scores, fix zero-scale padding in the FP8 NoPE path, and simplify the software-pipeline bookkeeping.

## Mask OOB KV on the V operand
- Added `attn_mask_oob_value(...)`, which zeros V rows for KV tokens past `valid_kv_len` in the last tile. Previously only the QK scores were masked, so padded/garbage V could leak into the PV accumulation.
- `attn_mask_oob_kv_tile` renamed to `attn_mask_oob_score`, with the `(tile+1)*KV_TILE <= valid_kv_len` early-out pulled inside the helper (removing the per-call-site `mask_oob_scores` lambdas and the passed-in `neg_inf`).
- `compute_pv` in the pipelined paths takes a compile-time `mask_oob` flag so the V-mask is only emitted on the final tile.

## FP8 NoPE zero-scale fix (MXFP8 variants)
- Zero the padding element(s) of the K/Q NoPE e8m0 mx-scale vectors for lanes ≥ 32 (`v_k_mxscl[3]/[7]`, `v_q_mxscl[3]`) and the `16nx4` path (`v_k_mxscl[14]/[15]`).
- Simplified the last-K-step NoPE column zeroing from the nested `static_for` computation to two `clear()` calls on the 16-wide sub-blocks.

## Fold temperature into online softmax (FMA)
- Removed the up-front `v_q *= temperature_scale` pass over Q (and its `cast<float>`/`cast<D_ATTN>` round-trip). `temperature_scale` is now threaded into each `pa_prefill_accum_*` function and applied inside the row-max / subtract step.
- `attn_sub_row` → `attn_row_scale_sub`, fusing scale-and-subtract into a single `__builtin_fmaf(v_s, scale, -row_max)`.
- Row-max computed as `attn_row_max(...) * temperature_scale`. Keeping Q un-scaled avoids the extra cast/multiply and improves numerics.
- `temperature_scale` typed as `D_ACC` throughout (was `float`).

## Pipeline cleanup
- Reduced the software-pipelined page-index buffer from `kv_page[4]` to `kv_page[2]` and simplified the rotation (dropped the `kv_page[0]=kv_page[2]` shuffles).
- Removed `pin_output_tile` and all its call sites (no longer needed).
- Added an `s_barrier` between the prefix and extend segments.

## Testing
- `op_tests/test_pa_sparse_prefill_opus.py` (bf16/fp16/fp8 × sparse/dense sweep) on gfx950.


amd-ruitang3:
image

amd-ruitang3 (approved):

PR #4235, opened 2026-07-14
fix FIPS crash in hash_signature

dtrifiro:
in FIPS enabled environments, not having `usedforsecurity=False` causes aiter to crash


dtrifiro:
bump @valarLip

valarLip (approved):

PR #4238, opened 2026-07-14
fix gemm a16w8/a8w8 scale regression

yanxuer-999:
## Motivation

The new AMD Triton release wheel (`git89002410`) regresses several AITER kernels on gfx950 (MI350X) vs the previous release (`gitd0d77a509`): register-allocation / scheduling changes lower the effective occupancy (default `waves_per_eu` under-fills the EUs). This PR restores the lost occupancy on the affected `a16w8` / `a8w8` block-scale GEMMs and the paged-attention decode GQA kernel. No kernel source logic changes.

## Technical Details

- **GEMM block-scale (shape-specialized configs, gfx950):**
  - `gfx950-GEMM-A16W8_BLOCKSCALE-N=4096-K=4096.json`: `any` bucket `waves_per_eu 1 → 2`.
  - `gfx950-GEMM-A8W8_BLOCKSCALE-N=1280-K=8192.json`: `M_LEQ_256` bucket `num_warps 8 → 4`, `waves_per_eu 2 → 1`.
  - Each is a copy of the default config with only the hit `M`-bucket's occupancy knobs changed; the loader prefers `-N=<N>-K=<K>.json` over the default, so the change is scoped to the regressed shapes only.
- **pa_decode GQA launch** (`aiter/ops/triton/attention/pa_decode.py`):
  - `_paged_attn_decode_v1_w_dot_kernel` now launches with `waves_per_eu = 1 if query_grp_sz <= 8 else 2` (uses the real `query_grp_sz`, not the pow2 value clamped to 16). Overridable via `AITER_PA_DECODE_WPE`.


aiter/ops/triton/attention/pa_decode.py
```diff
@@ -2,24 +2,31 @@
...
+from typing import Optional
 
-import torch
 import triton
+import torch
 
+# gfx950: the new-triton LLVM backend over-allocates VGPRs (LSR) on the paged
+# attention decode GQA kernel, cutting occupancy. Force a low waves_per_eu to
+# restore it. 0 = use the grp-size heuristic below; env override wins when > 0.
+_PA_DECODE_WPE = int(os.environ.get("AITER_PA_DECODE_WPE", "0"))
```
vgokhale, line 14: This needs to be set in the config not by an env var. We don't accept env vars in AITER in most cases.

PR #4245, opened 2026-07-15
[MLA v4 nm] Fix NaN leak from OOB kv page-id folding to slot 0

liyjiang:
Rebuild the gfx950 qh64 nm decode .co to mask the invalid tail V columns before GEMMV. An out-of-range tail page-id can fold to physical slot 0 (which may hold NaN), so masking tail V avoids NaN*0 = NaN poisoning the output.


amd-ruitang3:
LGTM

amd-ruitang3 (approved):

Issue #3995, opened 2026-06-29
[Issue]: [TRITON] DeepSeekV4 fused routing (`moe_routing/routing.py`) lacks `expert_map`


yitingw1:
### Problem Description

**[TRITON] DeepSeekV4 fused routing (`moe_routing/routing.py`) lacks `expert_map` → mxfp4 (a16w4) MoE crashes under expert parallelism**

## Summary

The Triton mxfp4 MoE path crashes with an **illegal memory access** (`hipErrorIllegalAddress`, HIP code 700) on the **first MoE layer** whenever **expert parallelism (EP)** is enabled. The fault is in `moe_gemm_a16w4`, but the root cause is in routing: the DeepSeek fused routing entry `aiter/ops/triton/moe/moe_routing/routing.py:routing()` has **no `expert_map` parameter**, so it emits **global** expert ids while the GEMM is given **per-rank local** expert weights. Indexing a 32-expert (local) weight tensor with global ids 0..255 reads out of bounds.

This is a **regression**: the same workload worked before, via `fused_routing_from_topk`, which gained `expert_map` support in **#3348**. The DeepSeek routing path that replaced it never got the same support.

## Environment / Repro

- Model: DeepSeek-V4-Pro/DeepSeek-V4-Flash (mxfp4 / `expert_dtype: fp4`), also applies to gpt-oss (any fp4x2 MoE on the triton path).
- HW: **architecture-independent** — this is a routing-logic bug, not a HW/occupancy issue. Reproduced on **both gfx942 (MI308X) and gfx950 (MI355X)**.
  - gfx942 (MI308X): `use_triton=True` by default → hit out of the box.
  - gfx950 (MI355X): `use_triton` defaults to False, so reproduce with `ATOM_USE_TRITON_MOE=1` (the same triton path) → **same coredump**.
- Launch (ATOM): `-tp 8 --enable-dp-attention --enable-expert-parallel` (any EP config; `ep_size > 1`).
- Result: warmup, layer 0 FFN → `RuntimeError: Triton Error [HIP]: Code: 700, Messsage: an illegal memory access was encountered` (raised at the `moe_gemm_a16w4` launch; confirmed with `AMD_SERIALIZE_KERNEL=1 HIP_LAUNCH_BLOCKING=1`).

Call chain (consumer = ATOM; line numbers from the repro tree, but the bug is entirely in the aiter routing → GEMM contract):

```
# ---- consumer (ATOM) ----
[...]

PR #4252, opened 2026-07-15
[FIX] Expert Map Parallel

amirumoAMD:
## Motivation

Fix expert map parallel https://github.com/ROCm/aiter/issues/3995

## Technical Details

Expert map routing added to main triton routing.

## Test Plan

lm_eval triton and non-triton ATOM DSV4, GPT, DS-R1

## Test Result

Pass

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/moe/moe_routing/routing.py
+++ b/aiter/ops/triton/moe/moe_routing/routing.py
@@ -6,6 +6,9 @@ from aiter.ops.triton._triton_kernels.moe.moe_routing.routing import (
     _combined_routing,
     _combined_routing_fused,
 )
+from aiter.ops.triton._triton_kernels.moe.moe_routing.expt_data import (
+    _expt_data_only_kernel,
+)
 from aiter.ops.triton.utils._triton.arch_info import is_tdm_avail
 from aiter.ops.triton.moe.moe_routing.topk import grouped_topk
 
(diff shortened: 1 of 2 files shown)


skyguan92:
I tested the `NK > 4096` boundary on a physical gfx1100 and have a device-only prototype that can replace this PR's Torch `argsort`/`bincount` fallback without changing its legacy expert-map semantics.

Implementation: keep the existing single-CTA path through `NK=4096`; above that, use 256-route CTA tiles for local histograms + global atomic accumulation, the existing prefix sum, then tiled atomic placement. There is no CPU sync and no Torch sorting.

W7900 result (ROCm 7.2, top-6, 256 global / 32 local experts, ranks 0/3/7):

| tokens | exact #4252 fallback p50 | device p50 | speedup |
|---:|---:|---:|---:|
| 683 | 282.50 us | 72.49 us | 3.93x |
| 768 | 284.03 us | 72.51 us | 3.94x |
| 2048 | 280.48 us | 74.98 us | 3.69x |
| 8192 | 322.62 us | 76.42 us | 4.21x |

Method: five independent processes, 200 timed repetitions, HIP events, fixed `profile_standard`; each table row is the median across 15 rank/process p50 points. The timed boundary includes route output allocations and excludes ExptData/GEMM. Inputs repeat rows from a captured formal DeepSeek-V4-Flash-0731 256-token route to the listed lengths, so these are route-derived long shapes, not full-model long-prefill captures.

All 60 rank/process cases matched exact histogram, gather/scatter inverse, and per-original-slot gate weights. A separate boundary suite (`1, 682, 683, 2048, 8192` tokens, mapped and unmapped) is 20/20 on gfx1100.

The DCO commit is available at skyguan92/aiter `gfx1100-long-route`: https://github.com/skyguan92/aiter/commit/627621fbc64ff5835960188ffb81e88dcece4ad0 . Its merge tree with current AITER main is byte-identical to the current-main tree tested on the card.

Since #4252 already owns the public expert-map call-site fix, I do not want to open an overlapping PR without coordination. Would you prefer to incorporate/cherry-pick this primitive here, or should I send it as a follow-up after #4252 is rebased?

PR #4255, opened 2026-07-16
fix(triton): support paged MQA logits on gfx1201

liminfei-amd:
## Motivation

The optional Gluon paged-MQA compiler supports `gfx942`, `gfx950`, and
`gfx1250`, but the runtime previously selected it for every architecture except
`gfx1201`. That fixed the reported board while leaving other unsupported or
future architectures able to enter the Gluon path and fail before kernel launch.

## Modifications

- Use one explicit Gluon architecture allowlist for runtime dispatch and the
  compile-time assertion.
- Keep `gfx1201` and other unsupported architectures on the existing Triton
  paged-MQA path.
- Add gfx1201-only page-size-one reference tests at context lengths 128 and 1024.
- Resolve the effective Gluon/Triton backend before deriving the VarCtx grid.
- Warn and fall back to the fixed-SplitKV schedule when VarCtx is unavailable,
  and guard the Preshuffle and Triton boundaries explicitly.

Existing dtype handling is unchanged.

## Tests

```bash
python -m pytest -q op_tests/triton_tests/attention/test_fp8_mqa_logits.py
```

Before: both new gfx1201 cases fail at the Gluon architecture check. After, the
page-size-one cases at context lengths 128 and 1024 pass on a physical gfx1201
GPU. A real non-null VarCtx schedule produces the expected fallback warning,
uses the same Triton cache key as the ordinary path, and returns bit-identical
output (`rtol=0`, `atol=0`).

Ruff, `py_compile`, and `git diff --check` pass. The Gluon paths on gfx942,
gfx950, and gfx1250 remain for upstream GPU CI.

---
*This is an experimental, AI-generated code change, reviewed by AMD engineers before submission.*


aiter/ops/triton/attention/pa_mqa_logits.py
```diff
@@ -514,7 +516,8 @@ def deepgemm_fp8_paged_mqa_logits(
     else:
         grid = (batch_size * next_n * SplitKV, 1, 1)
 
-    if enable_gluon_pa_mqa_logits:
+    use_gluon = enable_gluon_pa_mqa_logits and get_gfx() != "gfx1201"
```
cagrikymk, line 519: The kernel authors can give cleaner response but looks like the gluon kernel supports the following archs: cdna3, cdna4, gfx1250. So, having more precise gating here would prevent any miss-routing in the future

liminfei-amd:
@cagrikymk Thanks for the review. We originally opened this PR because gfx1201 paged MQA was being routed to the optional Gluon path and failed its architecture check before kernel launch; the goal was to keep gfx1201 on the existing Triton fallback and add page-size-one coverage. I updated the gate to positively allow only the architectures supported by the Gluon compiler (`gfx942`, `gfx950`, and `gfx1250`), so other architectures remain on the fallback path.

op_tests/triton_tests/attention/test_fp8_mqa_logits.py
```diff
@@ -149,3 +149,75 @@ def test_fp8_mqa_logits(
...
+        ("gfx942", True),
+        ("gfx950", True),
+        ("gfx1250", True),
+        ("gfx1100", False),
+        ("gfx1151", False),
+        ("gfx1200", False),
+        ("gfx1201", False),
+    ],
+)
+def test_gluon_paged_mqa_logits_arch_gate(gfx: str, expected: bool) -> None:
```
cagrikymk: I dont think these tests are testing anything tangible as the we have a simple frozen set for this, IMO.
Basically, we just replicate the same logic here without gaining any proper coverage. I suggest we remove these tests.

aiter/ops/triton/attention/pa_mqa_logits.py
cagrikymk, line 528: As far as I can see, VarCtxSchedule is not supported in the Triton side.
Could you add an assert, or warning + fallback, so that grid is constructed correctly when we use triton?

cagrikymk: This is not introduced in your PR but having warning + fallback when gluon is not available and VarCtxSchedule  is not None would be nice.

And related assert for the triton path.

liminfei-amd:
@cagrikymk  Thanks for the review. I removed the redundant test and fixed the VarCtx fallback and Triton guards.

On gfx1201, both page-size-one cases (128 and 1024) pass. The real VarCtx fallback uses the same Triton cache key as the normal path and produces bit-identical output. Gluon architectures remain for upstream CI.

PR #4292, opened 2026-07-19
[Bugfix][Triton] Quantize zero SageAttention V channels without NaNs

### LINKED ISSUE #4289: SageAttention V quantization emits NaNs for all-zero channels
## Problem

SageAttention computes one FP8 V scale for each `(batch, KV head, channel)` across the sequence. If a complete sequence channel is zero, its scale is zero and both V quantization kernels evaluate `0 / 0`.

This affects:

- `sage_quant`, through `sage_quant_kernel`;
- `sage_quant_mxfp4` and `fused_sage_quant_mxfp4`, through `sage_quant_v_kernel`;
- both `bshd` and `bhsd` layouts.

The resulting FP8 NaN propagates through attention. Applying a zero descale after accumulation cannot recover zero because `NaN * 0` remains NaN.

## Analysis

For finite V, the per-channel scale is:

```text
scale[b,h,d] = max_t(abs(V[b,t,h,d])) / FP8_MAX
```

If every value in one sequence channel is zero, then its scale is zero and the current kernel computes:

```text
scale = 0
quantized_v = FP8(0 / 0) = NaN
```

## Expected behavior

Use one only as the divisor for zero-scale channels:

```python
safe_scale = 1.0 if scale == 0 else scale
quantized_v = FP8(v / safe_scale)
```

This produces FP8 zero for zero channels while leaving every nonzero-scale channel unchanged. The stored descale remains zero.

## Why this is correct

- When `scale == 0`, the `amax(abs(V))` definition implies every value in that sequence channel is zero. Dividing by one therefore produces exact zeros.
- When `scale != 0`, `safe_scale == scale`, so the existing quantization formula is unchanged.

## Success criteria

Cover both V implementations, both layouts, an isolated zero channel, an all-zero head, and finite quantized, dequantized, and final attention outputs.

morluto:
## Summary

Define SageAttention V quantization for all-zero sequence channels.

Both V quantization implementations now substitute a safe divisor when the stored per-channel scale is zero. The stored scale itself remains unchanged.

Fixes #4289.

## Correctness proof

SageAttention computes one FP8 V scale for each `(batch, KV head, channel)`:

```text
scale[b,h,d] = max_t(abs(V[b,t,h,d])) / FP8_MAX
```

If a channel is zero at every sequence position, its scale is zero and the old calculation becomes:

```text
scale = 0
quantized_v = FP8(0 / 0) = NaN
```

Applying the zero descale later cannot recover zero because `NaN * 0` is still NaN.

The new calculation is:

```python
safe_scale = 1.0 if scale == 0 else scale
quantized_v = FP8(v / safe_scale)
```

This is correct in both possible cases:

| Scale | Consequence |
|---|---|
| `scale == 0` | `amax(abs(V)) == 0` implies the complete sequence channel is zero. Dividing by one produces exact FP8 zeros, and the stored zero descale reconstructs zero. |
| `scale != 0` | `safe_scale == scale`, so the existing quantization formula is unchanged. |

## Changes

- Add one shared zero-scale-safe V quantization helper.
- Use it in `sage_quant_v_kernel`, reached by the Sage MXFP4 wrappers.
- Use it in the V branch of `sage_quant_kernel`, reached by Sage v1.
- Add regression coverage for both layouts, Sage v1 and MXFP4, an isolated zero channel, an all-zero V head, exact FP8 values, and final attention output.

## Testing

- Original arithmetic reproduces a zero scale and NaN quantized values; the corrected arithmetic produces exact zeros while preserving nonzero channels.
- Both changed kernels compile for gfx942 with `fp8e4b8` and gfx950 with `fp8e4nv`.
- `ruff check aiter/ops/triton/_triton_kernels/quant/sage_attention_quant.py op_tests/triton_tests/quant/test_sage_attention_quant.py`
- `ruff format --check aiter/ops/triton/_triton_kernels/quant/sage_attention_quant.py op_tests/triton_tests/quant/test_sage_attention_quant.py`
- `python3 -m py_compile aiter/ops/triton/_triton_kernels/quant/sage_attention_quant.py op_tests/triton_tests/quant/test_sage_attention_quant.py`
- `git diff --check`

## Scope

This change does not alter Q/K quantization or define behavior for NaN/Inf source tensors.

## Performance

The V path adds one elementwise comparison and selection. Memory access and launch geometry are unchanged.

## Dependencies

No new dependencies.

## Breaking Changes

None.


juuso-oskari (approved):
LGTM

PR #4084, opened 2026-07-05
perf: eliminate end_sync in custom allreduce by delaying input tensor release until next sync

jpy794:
## Motivation

Follow-up to https://github.com/ROCm/aiter/pull/4082.

PR #4082 fixed a custom collective race by adding an end-of-kernel cross-rank `end_sync`, ensuring peer ranks finish reading input buffers before the caller can reuse or overwrite them. That fix is correct, but the extra device-side synchronization adds measurable latency to small custom collective paths.

This PR keeps the race fix while providing a performance path that avoids the final `end_sync`. Instead of synchronizing at the end of each collective, Python keeps the input tensor alive until the next custom collective launch. The next collective's existing start synchronization proves all ranks have advanced past the previous collective, so the previous input can be safely released at that point.

cc @ColorsWind

## Technical Details

This change adds an optional `end_sync` flag through the custom allreduce Python/C++/HIP stack.

When `hold_inputs=False`, behavior stays unchanged and kernels still run the final `end_sync`.

When `hold_inputs=True`:

- custom collective launches pass `end_sync=False`
- kernels skip the final `end_sync<ngpus, true>(...)`
- `CustomAllreduce` records the input tensor in Python
- the recorded input is replaced on the next collective launch, after the next collective's start sync has established that all ranks are done with the previous input
- recorded inputs are cleared on communicator close

The feature is controlled through:

```bash
AITER_CUSTOM_AR_HOLD_INPUTS=1
```

Covered paths include:
- all-gather
- reduce-scatter
- fused all-reduce + RMSNorm
- fused all-reduce + RMSNorm quant variants

## Test Plan

You can find reproduce scripts here https://gist.github.com/jpy794/b21fd3c7b9a9ac1d0e87db62824cf78b

Race repro
```
rm -rf aiter/aiter/jit/module_custom_all_reduce.so \
       aiter/aiter/jit/build/module_custom_all_reduce

AITER_CUSTOM_AR_HOLD_INPUTS=1 python tools/custom_collective_race_repro.py
```

Performance microbench
```
AITER_CUSTOM_AR_HOLD_INPUTS=1 \
python tools/custom_collective_microbench.py \
  --graph-iters 100 \
  --label hold-inputs-graph100 \
  --output-json results/custom_collective_microbench/hold-inputs-graph100.json
```

## Test Result

**Correctness**

```
for case in ag rs ar1stage; do
[...]

PR #4082, opened 2026-07-05
fix: synchronize custom collectives before return

jpy794:
## Motivation

This PR fixes a race in custom collective kernels where a collective may return before all peer reads from the input buffer have completed.

In CUDA graph / overlapped execution scenarios, the caller may reuse or overwrite the input tensor immediately after the custom collective returns. Without an end-of-kernel synchronization, another rank can still be reading that input buffer, which can lead to corrupted collective outputs. The repro script demonstrates this by filling the input with NaNs right after the collective and checking whether NaNs leak into the output.

cc @ColorsWind

## Technical Details

This change adds `end_sync<ngpus, true>(sg, self_sg, rank)` before returning from custom collective kernels that directly read peer input buffers.

Covered paths in `csrc/include/custom_all_reduce.cuh` include:

- all-gather kernels:
  - `allgather_naive`
  - `allgather_vec`
  - `allgather_lastdim`

- reduce-scatter kernels:
  - `reduce_scatter_split_first_dim`
  - `reduce_scatter_split_lastdim_naive`
  - `reduce_scatter_split_lastdim`
  - `reduce_scatter_split_middim_naive`
  - `reduce_scatter_split_middim`

- 1-stage fused all-reduce + RMSNorm kernels:
  - standard variant
  - per-group quant variant
  - MXFP4 variant

The synchronization ensures that all ranks complete their remote reads before any rank is allowed to proceed and potentially mutate or recycle its input buffer.

## Test Plan

Use the custom collective race repro script here https://gist.github.com/jpy794/a2849ad6d3d2a3f0d1a03c6c1ec977ac

```bash
rm -rf aiter/aiter/jit/module_custom_all_reduce.so \
       aiter/aiter/jit/build/module_custom_all_reduce
python custom_collective_race_repro.py
```

The repro captures a CUDA graph that repeatedly:
1. resets the collective input to finite values,
2. runs the custom collective,
3. overwrites rank 0's input with NaNs immediately after the collective,
4. checks the collective output on device for leaked NaNs.
Expected result after this fix: all ranks report bad_nan_rows=0.

## Test Result

Branch | AG | RS | AR 1-stage | Exit
-- | -- | -- | -- | --
main | FAIL, 1477 bad rows | FAIL, 676 bad rows | FAIL, 59 bad rows | 1
fix-agrs-race | PASS, 0 bad rows | PASS, 0 bad rows | PASS, 0 bad rows | 0

## Submission Checklist

[...]

PR #4269, opened 2026-07-16
[Feature][FlyDSL] Add fused heterogeneous MXFP4/FP8 shared-expert MoE (FHMoE) and share the mixed-MoE kernel builders

Fangzhou-Ai:
## Summary

- Add a dedicated fused heterogeneous MoE (FHMoE) path for DeepSeek-V4 checkpoints whose 384 routed experts use MXFP4/E8M0 while the shared expert remains native FP8/E8M0.
- Fuse the shared expert into the same FlyDSL stage-1 and stage-2 schedules as logical expert 384, while keeping shared-expert fields out of the ordinary `fused_moe_` custom-op schema and ordinary kernel compiler APIs.
- Share the mixed-MoE kernel builders between ordinary MoE and FHMoE through `mixed_moe_gemm_2stage_common.py`, leaving thin per-path facades.
- Support the current `main` launcher, XCD-swizzle, runtime, and AOT contracts, including the current stage-1 `situ_beta`, `situ_linear_beta`, `k_batch_intra_block`, and `pass_swiglu_limit` arguments.

## Stack and review scope

This PR is the feature base and targets current `main`. #4314 contains the DeepSeek-V4 C1/C2/C32/C64 tuning rows and is stacked on it:

```text
main aa6ee6102 -> #4269 2a4cb1339 -> #4314 57e848f53
```

This PR contains four commits across ten paths:

1. `feat(flydsl): add fused heterogeneous MoE`
2. `refactor(flydsl): share MoE and FHMoE kernels`
3. `fix(flydsl): preserve ordinary MoE stage1 arguments`
4. `fix(flydsl): forward FHMoE AOT stage1 ABI flag`

The feature, runtime wrappers, AOT adapter, GPU-kernel facade, and tests remain isolated in `fhmoe.py` / `test_fhmoe.py` files. Existing MoE modules contain only the narrow dispatch and shared-builder bridges needed by the new path. Tuning CSV changes remain in #4314.

## Why heterogeneous fusion

The checkpoint stores routed experts as MXFP4 but the shared expert as FP8. Quantizing the shared expert to MXFP4 permits ordinary homogeneous fusion but adds avoidable weight error; launching a separate FP8 shared kernel preserves precision but gives up much of the fusion gain. FHMoE keeps one fused schedule while selecting native FP8 weight and scale resources only for the final logical shared expert.

[...]

PR #4294, opened 2026-07-19
[Bugfix] Fix IPC publication ordering in custom all-reduce

Fangzhou-Ai:
## Summary

- Publish eager fixed-buffer copies and intermediate all-reduce writes with system-scoped release stores.
- Acquire system-scoped visibility after all peer ready signals are observed, before reading peer IPC buffers.
- Limit the change to the tested gfx9 synchronization helper while keeping final read-completion barriers relaxed; gfx1250 is intentionally unchanged.
- Add a TP8 stress test for repeated eager-buffer reuse and graph replay under rotating rank skew.

## Root cause

The eager path copies input into an IPC-registered fixed buffer immediately before launching the collective. The ready flag was system-scoped but relaxed, and peers polled it with device scope. That did not establish a system-scoped happens-before edge from the producer copy to peer buffer reads. The same gap existed for intermediate writes consumed after a non-final barrier.

## Duplicate-work check

I reviewed #4082 and #4084. Those changes protect buffer lifetime **after** peer reads by adding terminal synchronization or retaining caller tensors. This PR instead publishes data **before** peer reads. It intentionally excludes their terminal-barrier changes, so the fixes are complementary rather than duplicate. Searches for custom all-reduce IPC publication/order fixes found no other open PR.

## Validation

The original validation below ran on an 8x gfx950 host at `main` commit `dedc19d32bb31ec134ad0388f878e4cd7a83af16`. After rebasing, the TP8 ordering stress was rerun for 100 iterations on current `main` commit `632367677a6a7fc7a329262775f0a74e6eb8f78d` and passed on all ranks. The other results below predate the rebase.

```bash
HIP_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 PYTHONPATH=$PWD CK_DIR=$PWD/3rdparty/composable_kernel \
  /home/amd/vllm/.venv/bin/python \
  op_tests/multigpu_tests/test_custom_allreduce_ipc_ordering.py \
  --iters 100 --tp-size 8
```

Passed on all eight ranks, including graph capture/replay. This is a stress regression guard rather than a deterministic reproducer; an unpatched-main 20-iteration control also completed.

The fused all-reduce/RMS integration sweep passed 11,200/11,200 comparisons across M=1,4,8,16,32,64,4096 (100 iterations each), with maximum absolute error 0.125. A graph-free DeepSeek-V4-Pro GSM8K run completed 1,319 samples at 95.906% strict / 95.830% flexible accuracy; its server handled 1,351 successful HTTP responses with no GPU fault.

The stock TP8 graph benchmark was also run on main and this patch:

| BF16 shape | main min-max (us) | patched min-max (us) |
|---|---:|---:|
| `(2, 7168)` | 5.45-6.69 | 6.25-7.50 |
| `(128, 8192)` | 14.53-15.11 | 15.74-16.29 |

```bash
HIP_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 PYTHONPATH=$PWD CK_DIR=$PWD/3rdparty/composable_kernel \
  /home/amd/vllm/.venv/bin/python \
  op_tests/multigpu_tests/test_custom_allreduce.py \
  -d bf16 -s 2,7168 -g true
HIP_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 PYTHONPATH=$PWD CK_DIR=$PWD/3rdparty/composable_kernel \
  /home/amd/vllm/.venv/bin/python \
  op_tests/multigpu_tests/test_custom_allreduce.py \
  -d bf16 -s 128,8192 -g true
```

Static checks:

```bash
uvx --from black==26.3.0 black --check op_tests/multigpu_tests/test_custom_allreduce_ipc_ordering.py
uvx --from ruff==0.15.7 ruff check op_tests/multigpu_tests/test_custom_allreduce_ipc_ordering.py
/home/amd/vllm/.venv/bin/python -m py_compile op_tests/multigpu_tests/test_custom_allreduce_ipc_ordering.py
uvx --from clang-format==18.1.8 clang-format --dry-run --Werror --lines=154:236 csrc/include/custom_all_reduce.cuh
git diff --check origin/main...HEAD
```

The gfx1250 helper is intentionally unchanged because no gfx1250 hardware validation was available and that path is being reworked.

## AI assistance

OpenAI Codex was used to audit related PRs, prepare the patch and regression test, run validation, and draft this description. The human submitter is responsible for reviewing every changed line and defending the change end-to-end.

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -154,9 +154,9 @@ DINLINE O downcast(V val)
 }
 
 // This function is meant to be used as the first synchronization in the all
-// reduce kernel. Thus, it doesn't need to make any visibility guarantees for
-// prior memory accesses. Note: volatile writes will not be reordered against
-// other volatile writes.
+// reduce kernel. In eager mode, the input is copied into an IPC-registered
+// buffer immediately before the kernel launch, so the ready signal must publish
+// that copy before peers read the buffer.
 template <int ngpus>
 DINLINE void start_sync(const RankSignals& sg,
 #ifndef USE_ROCM
@@ -169,20 +169,17 @@ DINLINE void start_sync(const RankSignals& sg,
     uint32_t flag = self_sg->_flag[blockIdx.x] + 1;
     if(threadIdx.x < ngpus)
     {
-        // simultaneously write to the corresponding flag of all ranks.
-        // Latency = 1 p2p write
         __scoped_atomic_store_n(&sg.signals[threadIdx.x]->start[blockIdx.x][rank],
                                 flag,
-                                __ATOMIC_RELAXED,
+                                __ATOMIC_RELEASE,
                                 __MEMORY_SCOPE_SYSTEM);
-        // wait until we got true from all ranks
         while(__scoped_atomic_load_n(&self_sg->start[blockIdx.x][threadIdx.x],
                                      __ATOMIC_RELAXED,
-                                     __MEMORY_SCOPE_DEVICE) < flag)
+                                     __MEMORY_SCOPE_SYSTEM) < flag)
             ;
+        __scoped_atomic_thread_fence(__ATOMIC_ACQUIRE, __MEMORY_SCOPE_SYSTEM);
     }
     __syncthreads();
-    // use one thread to update flag
     if(threadIdx.x == 0)
         self_sg->_flag[blockIdx.x] = flag;
 #else
@@ -202,8 +199,10 @@ DINLINE void start_sync(const RankSignals& sg,
 }
 
 // This function is meant to be used as the second or the final synchronization
-// barrier in the all reduce kernel. If it's the final synchronization barrier,
-// we don't need to make any visibility guarantees for prior memory accesses.
+// barrier in the all reduce kernel. A non-final barrier publishes intermediate
+// writes consumed by the next stage. A final barrier only waits for peer reads
+// to finish before the input can be reused; its uncached signal does not carry
+// output data.
 template <int ngpus, bool final_sync = false>
 DINLINE void end_sync(const RankSignals& sg,
 #ifndef USE_ROCM
@@ -214,10 +213,6 @@ DINLINE void end_sync(const RankSignals& sg,
 {
 #ifdef USE_ROCM
     __syncthreads();
-    // eliminate the case that prior writes are not visible after signals become
-    // visible. Note that I did not managed to make this happen through a lot of
-    // testing. Might be the case that hardware provides stronger guarantee than
-    // the memory model.
     uint32_t flag = self_sg->_flag[blockIdx.x] + 1;
     if(threadIdx.x < ngpus)
     {
@@ -229,9 +224,12 @@ DINLINE void end_sync(const RankSignals& sg,
                                 __MEMORY_SCOPE_SYSTEM);
         // wait until we got true from all ranks
         while(__scoped_atomic_load_n(&self_sg->end[blockIdx.x][threadIdx.x],
-                                     final_sync ? __ATOMIC_RELAXED : __ATOMIC_ACQUIRE,
-                                     __MEMORY_SCOPE_DEVICE) < flag)
+                                     __ATOMIC_RELAXED,
+                                     final_sync ? __MEMORY_SCOPE_DEVICE : __MEMORY_SCOPE_SYSTEM) <
+              flag)
             ;
+        if constexpr(!final_sync)
+            __scoped_atomic_thread_fence(__ATOMIC_ACQUIRE, __MEMORY_SCOPE_SYSTEM);
     }
     __syncthreads();
     // use one thread to update flag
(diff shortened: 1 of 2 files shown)


TennyWang1223:
# PR Review

**Recommendation: do not merge.** Based on an ISA-level assembly comparison and cycle-accurate microbenchmarks, the cross-rank publication problem this PR claims to fix is not actually addressed at the instruction level, and the added test cannot demonstrate that the current `main` has the problem. I suggest keeping `main` as-is.

Full reasoning below.

## 1. Verdict on the three synchronization functions (gfx950, isolated compile)

| Function | PR change | Assembly evidence | Verdict |
|---|---|---|---|
| `end_sync<final=true>` | none | **byte-for-byte identical** to main | no effect, meaningless |
| `start_sync` | store relaxed→release, load + acquire fence | main has **no** `buffer_wbl2`/`buffer_inv` here; PR adds both | redundant (see §2) |
| `end_sync<final=false>` | `buffer_inv` moved inside→outside loop, scope `sc1`→`sc0 sc1` | `buffer_wbl2` present on both sides; only the inv position and scope differ | unnecessary (see §3) |

## 2. The `start_sync` change is redundant

`start_sync` sits at the very beginning of the collective. Its execution flow is:
`input copy → kernel launch → start_sync → remote load → compute → end_sync<false>`.

- **Producer side `release`→`buffer_wbl2`**: this instruction writes back the current grid's own dirty L2 lines. But the eager input copy is a **separate, already-retired memcpy grid** — it is not in this grid's writeback set. Moreover, there are no memory writes in this kernel before `start_sync`. So this writeback flushes nothing meaningful and **cannot publish the copy it claims to publish**.
- **Consumer side acquire fence→`buffer_inv`**: the previous iteration's `end_sync<false>` `buffer_inv` already invalidated the local cache. There are no IPC-buffer-related dirty lines that need to be invalidated again.

A probe confirms `buffer_wbl2` is triggered entirely by `__ATOMIC_RELEASE` (independent of whether prior dirty writes exist), and main's `start_sync` uses relaxed so it has no such instruction. Conclusion: the two cache instructions this PR adds to `start_sync` are superfluous.

 image

## 3. `end_sync<final=false>`: `sc1` (device scope) is enough, `sc0 sc1` (system scope) is not needed

The consumer side only polls the flag in its **own local tmp buffer**. Reading the correct flag already means the peer has finished writing; from that point on the consumer proceeds locally and its behavior does not depend on any other rank. Establishing *system*-scoped ordering here therefore buys nothing — **device scope (`sc1`) is sufficient**, and `sc0 sc1` only widens the invalidation scope, which can only lengthen latency, never shorten it.

 Moving the `buffer_inv` from inside the poll loop to outside it makes **no measurable difference to performance** (confirmed by cycle-accurate microbenchmark on gfx942). So from the HIP source there is **no justification** either to bump the scope to system or to move the fence outside the `while` loop — `sc1` inside the poll, as in `main`, is both correct and adequate.

image

## 4. The accompanying test script cannot prove `main` is broken

`test_custom_allreduce_ipc_ordering.py` is a regression guard, not a deterministic reproducer:

1. The author states in the description that the "unpatched-main 20-iteration control also completed" — **main passes too**. A test that cannot make main fail cannot prove main has a bug.
2. The assertion reads `output` for comparison only **after** `torch.cuda.synchronize()`. Weak-memory-order / visibility bugs are transient; a full device sync collapses any intermediate state, so **the assertion window opens after the only moment the bug could manifest**.
3. The `time.sleep(skew_s)` introduces host-side launch skew, which targets the "read before peer launch" scenario that the `start_sync` barrier is responsible for blocking — a different mechanism from "data publication / cache visibility," and the skew is absorbed by the barrier anyway.
4. The fill values are rotating small-integer linear combinations; a stale value and the expected value can numerically coincide, and the test lacks a "unique per-iteration sentinel + active clobber recheck" to sharpen stale-read detection.

## 5. Note on gfx1250

The assembly above was generated on gfx9. There is no need to modify `start_sync` / `end_sync` for gfx1250: the kernel implementation there will be reworked and will **no longer use these two device-side synchronization functions**. That update is coming soon, so I did not bother comparing the gfx1250 assembly.

## Summary

The change does not hit the problem it claims to address (release targets the wrong object; there is nothing to publish before `start_sync`), and the test cannot reproduce the problem either. For `end_sync<false>`, `sc1` (device scope) inside the poll loop — exactly what `main` already does — is both correct and sufficient; `sc0 sc1` (system scope) adds latency risk with no measurable benefit, and moving the fence outside the loop is not warranted. `main` currently runs stably, and it is not worth introducing system-scope overhead and extra cache instructions for this. **I recommend closing this PR or keeping `main`.**

Fangzhou-Ai:
Hi @TennyWang1223  Thanks for your comments. I agree that we should not touch gfx1250 file as it's not tested. The IPC bug was exposed to me with this https://github.com/ROCm/aiter/pull/4269/ This is a hybrid-precision kernels fusing FP8 shared expert and mxfp4 routed experts for DSV4. I will further dig into this issue to check if it's indeed a correct fix.

Fangzhou-Ai:
Closing this PR after the follow-up A/B investigation. The high-concurrency GPU fault that motivated this change reproduces on the baseline stack with official AITER v0.1.16.post3, the pre-FSE vLLM parent, FSE absent, and custom all-reduce disabled (PYNCCL). The first 64-request wave completes and the second wave faults. Therefore #4294 is neither required for nor demonstrated to fix that failure, and its regression test does not establish a main-branch defect. We will make #4269 self-contained and keep its FSE validation separate from this pre-existing baseline issue.

PR #4300, opened 2026-07-20
[CK] fmoe run_config: align per_1x32 fp4/fp8 dispatch with test_moe_2stage

yzhou103:
## Motivation

`run_config` re-derived fused_moe's runtime knobs for per_1x32 fp4/fp8 MoE and
drifted from the production/test path. The failure is silent: a wrong knob makes
the tuned-config lookup miss, fused_moe falls back to its default kernel, and the
row still reports OK — so rows appeared validated while no tuned kernel ran.

## Technical Details

`fmoe_runtime_policy()` in `aiter/fused_moe.py` is now the single source for every
knob that decides which tuned config fused_moe dispatches to, and both
`--run_config` and `op_tests/test_moe_2stage.py` derive from it instead of
re-implementing it.

* **gate mode / weight layout** (per_1x32; everything else stays SEPARATED):
  * mxfp8 (fp8 act + fp8 weight) → INTERLEAVE, `w1_gate_up=True`
  * a8w4 (fp8 act + fp4 weight) → INTERLEAVE, `w1_gate_up=True`, matching
    serving's `ATOM_MOE_GU_ITLV=1`
  * a16w4 (bf16/fp16 act + fp4 weight) and a4w4 → SEPARATED, `w1_gate_up=False`
    (standard GGUU layout)

  `gate_mode` and `w1_gate_up` are returned together on purpose: the kernel only
  reads the gate_up layout correctly under the matching gate mode.

* **env overrides**, applied per row through `fmoe_runtime_env()` and restored
  afterwards so an override cannot leak into later rows:
  * a8w4, non-SiTUv2: `AITER_BF16_FP8_MOE_BOUND=0`, because under INTERLEAVE the
    picker only selects the fp8 activation once `token >= 256`, so small-token
    rows would never reach the tuned a8w4 kernel
  * SiTUv2: `AITER_SITUV2_A8W4=1` (fp8 act) or `AITER_SITUV2_A4W4=1` (fp4 act);
    SiTUv2 is tested ahead of the gate-mode branches and otherwise defaults to a
    bf16 activation

* **SiTUv2 beta / linear_beta** are pinned to `DEFAULT_SITUV2_BETA` /
  `DEFAULT_SITUV2_LINEAR_BETA`. They are runtime scalars rather than compile
  keys, so leaving them unset inherits whichever stage1 wrapper wins the
  dispatch (1.0 for FlyDSL, 4.0/25.0 for Opus).

* **reference oracle**: keep the activation in bf16 for a16w4/a8w4/mxfp8; feed a
  random expert bias to both kernel and reference for per_1x32 Swiglu + fp4
  weights (gpt-oss swiglu folds an extra +1 into the linear branch via the kernel
  bias path). `torch_moe_2stages` now forwards `swiglu_limit` into
  `torch_moe_stage1`, which it previously accepted and dropped.

## Test Result

`--run_config` against the tuned model CSVs on gfx950, all of which previously
reported OK while running the untuned default kernel:

| config | before | after | default-kernel fallbacks |
|---|---|---|---|
| gptoss_a16w4 | 0/24 | 24/24 | 0 |
| kimik3_a4w4 (SiTUv2) | 0/17 | 17/17 | 0 |
| kimik3_a8w4 (SiTUv2) | 0/35 | 34/35 | 0 |
| glm5_mxfp8 | 32/32 | 32/32 | 0 |

The one remaining kimik3_a8w4 row is a config-data problem rather than a
dispatch one: its recorded timing and accuracy correspond to a smaller
`inter_dim` than the one it is keyed on (774 µs for a 14.8 GB weight footprint
implies 19 TB/s), so `run_config` is correctly flagging it. Reported separately.

For `test_moe_2stage` the change is behaviour-neutral — legacy sweep output is
bit-identical before and after — and it additionally stops claiming INTERLEAVE
for per_Tensor/per_Token/per_128x128 a8w8 rows whose weights are shuffled in the
SEPARATED layout.

## Submission Checklist


zufayu (dismissed):

zufayu:
ROCm/aiter PR #4300 — [CK] fmoe run_config: align per_1x32 fp4/fp8 dispatch with test_moe_2stage
This PR changes the offline FMoE tuning validator so per_1x32 FP4/FP8 rows use production-like gate layouts, activation dispatch, and Swiglu bias handling.

Validation (deterministic): NOT RUN — the changed tuner has one GPU test target, but this Windows host has no ROCm GPU; exact-head Aiter and OPUS CI passed without exercising run_config
Perf (advisory): NOT RUN — the selected tuning-pipeline test exposes no base/head benchmark entry point, and this PR makes no kernel-speed claim

⚠️ [verified] On gfx950, csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py:4534-4557 treats both bf16+fp4 and fp8+fp4 rows as INTERLEAVE and forces AITER_BF16_FP8_MOE_BOUND=0, while the current canonical test_moe_2stage path uses SEPARATED/gate_up=False for bf16 a16w4 and reserves interleave for fp8 a8w4. A CSV row declaring q_dtype_a=bf16, q_dtype_w=fp4x2, and QuantType.per_1x32 is therefore redirected to the fp8/a8w4 family instead of validating its advertised a16w4 configuration. Author must rebase and derive gate mode, shuffle layout, and bound override from the current shared test/runtime rules, applying the zero bound only to actual a8w4 rows.

⚠️ [verified] csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py:2440 adds swiglu_limit to torch_moe_2stages, but its call to torch_moe_stage1 at line 2453 never forwards the value. Any caller requesting a non-default Swiglu clamp gets the default reference computation instead, which can create false tuning mismatches or validate against the wrong oracle. Author must pass swiglu_limit=swiglu_limit into torch_moe_stage1 and add a non-default-limit assertion.

⚠️ [verified] The only test change, op_tests/tuning_tests/test_tune_pipeline.py:99, sets TUNE_ONLY before ordinary tuner generation; _run_one never passes --run_config, so none of the new gate-mode, bound, bias, or reference logic executes in CI. This allows the defects above to pass the current green Aiter/OPUS checks. Author must add targeted --run_config regression cases for bf16 a16w4, fp8 a8w4, and Swiglu with a non-default limit.

zufayu (dismissed):

Issue #4304, opened 2026-07-20
[CI] aiter-release.yaml cannot build wheels for recent tags (post-v0.1.17 regression: start_aot needs GPU, torch 2.13 teardown bug, dispatch cannot pin torch)


sunway513:
## Summary

The release workflow `aiter-release.yaml` can no longer build wheels for recent tags/branches. Two recently-published releases shipped with **0 wheels** as a result:

- `v0.1.16.post4`
- `v0.1.17-longctx`

`v0.1.17` (GA) built fine on the same `build-only-aiter` runner, so this is a **regression introduced in `main` after v0.1.17**.

## Evidence

Dispatching `aiter-release.yaml` (manylinux2_28-builder, `runner: build-only-aiter`, ROCm 7.0/7.1/7.2 x cp310/cp312) on both `v0.1.16.post4` and `v0.1.17-longctx` fails identically:

```
3664x  'hipModuleLoadData(&module, data)' failed with 'hipErrorNoDevice'
4536x  [flydsl] COMPILE_ONLY=1 ...
Process completed with exit code 1
```

The C++ extension build (ninja/hipcc) never starts — the job dies during `setup.py` `start_aot`.

## Root causes

1. **`setup.py` `start_aot` now requires a GPU at build time.** On the GPU-less `build-only-aiter` runner every `hipModuleLoadData` returns `hipErrorNoDevice` and the build aborts. This is new since v0.1.17 (GA built cleanly GPU-less).

2. **torch 2.13.0 teardown bug.** Building with a GPU present (manual repro on an MI355X node) gets past `start_aot`, but the default `torch` pulled from the ROCm index is now `2.13.0`, which raises during interpreter teardown:
   ```
   File ".../torch/library.py", line 623, in _clear_torch_ops_cache
       ns, name_with_overload = qualname.split("::")
   ValueError: too many values to unpack (expected 2)
   ```
   i.e. an aiter custom-op qualname contains more than one `::`. Pinning `torch==2.11.0` removes this teardown error.

3. **`workflow_dispatch` cannot pin torch.** `torch_pin` / `torch_index_url` exist only under `workflow_call`, not `workflow_dispatch`, so a manual release dispatch cannot pin torch < 2.13 to work around (2):
   ```
   HTTP 422: Unexpected inputs provided: ["torch_pin"]
   ```

## Impact

[...]

PR #4305, opened 2026-07-20
[CI] release: default torch<2.13 + expose torch_pin/torch_index_url on dispatch

sunway513:
## What

Fix the release workflow so it can build wheels for recent aiter tags again. Two changes to `.github/workflows/aiter-release.yaml`:

1. **Default torch to `torch<2.13`** for the manylinux build. torch `2.13.0` breaks the build: `torch.library._clear_torch_ops_cache()` does `qualname.split("::")` expecting exactly one `::`, which raises `ValueError: too many values to unpack` on an aiter custom-op name during interpreter teardown of `setup.py` `start_aot()`. The wheel is never produced.
2. **Expose `torch_pin` and `torch_index_url` on `workflow_dispatch`** (they existed only under `workflow_call`), so a manual release dispatch can pin torch when needed.

## Why

`v0.1.16.post4` and `v0.1.17-longctx` were published with **0 wheels** because the release CI could not build them. See #4304 for full root-cause analysis.

## Not covered here (tracked in #4304)

`setup.py` `start_aot` also now requires a GPU at build time (`hipErrorNoDevice` on the GPU-less `build-only-aiter` runner). Until `start_aot` degrades gracefully without a device, release builds must select a GPU runner (`aiter-1gpu-runner` / `linux-aiter-mi35x-1`). This PR does not change the default runner.

## Test plan

Dispatch `aiter-release.yaml` from this branch on a GPU runner against a recent tag and confirm 6 wheels are produced. Validation to be attached.

Refs #4304

### DIFF
--- a/.github/workflows/aiter-release.yaml
+++ b/.github/workflows/aiter-release.yaml
@@ -53,16 +53,6 @@ on:
         description: 'Use pytorch/manylinux2_28-builder ROCm image (AlmaLinux 8 + devtoolset, glibc 2.28). Produces wheels ABI-compatible with vLLM/Ubuntu 22 containers.'
         type: boolean
         default: false
-      torch_pin:
-        description: 'Optional torch version pin for the manylinux build (e.g. 2.11.0 or 2.10.0+rocm7.1). Empty = default constraint (torch<2.13).'
-        type: string
-        required: false
-        default: ''
-      torch_index_url:
-        description: 'Optional override for the torch wheel index URL. Empty = auto-derive from the builder image ROCm tag.'
-        type: string
-        required: false
-        default: ''
   workflow_call:
     inputs:
       release_type:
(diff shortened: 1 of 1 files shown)


gyohuangxin:
@sunway513  Please resolve the actionlint error: https://github.com/ROCm/aiter/actions/runs/29761658282/job/88417489008?pr=4305, otherwise LGTM.

sunway513:
@gyohuangxin thanks for the review. Fixed the actionlint error: `workflow_dispatch` caps at 10 inputs and adding both `torch_pin`+`torch_index_url` pushed it to 12. Since the new default `torch<2.13` already resolves the torch 2.13 teardown crash without needing a manual pin, I dropped the two dispatch inputs (they remain available via `workflow_call`) and kept the default-constraint change plus the GPU device passthrough. Now back to 10 dispatch inputs. PTAL.

gyohuangxin (approved):

gyohuangxin:
> @gyohuangxin thanks for the review. Fixed the actionlint error: `workflow_dispatch` caps at 10 inputs and adding both `torch_pin`+`torch_index_url` pushed it to 12. Since the new default `torch<2.13` already resolves the torch 2.13 teardown crash without needing a manual pin, I dropped the two dispatch inputs (they remain available via `workflow_call`) and kept the default-constraint change plus the GPU device passthrough. Now back to 10 dispatch inputs. PTAL.

LGTM, thanks for clarification.

PR #4332, opened 2026-07-22
[Triton/Gluon] [FlyDSL] [CI] feat(flydsl): Add paged-attention Tile kernel

fsx950223:
## Motivation

Add a FlyDSL paged-attention decode implementation to AITER for FP8 KV caches.

The kernel provides explicit control over MFMA execution, KV-page loading, LDS
layout, softmax, partition scheduling, and native FlyDSL partition reduction.

## Technical Details

- Add FlyDSL paged-attention decode under `aiter.ops.flydsl`.
- Support:
  - FP8 K/V caches
  - BF16/FP16 queries
  - Block sizes 16, 64, and 128
  - Per-tensor and per-token KV scales
  - GQA and multi-token prediction
  - Plain and transposed V-cache layouts
- Process context in 256-token tiles using four wave64 waves per CTA.
- Use wide K=128 FP8 MFMA for the tuned MTP3/MTP4 gfx950 path.
- Split small MTP2/MTP4 grids into one query per CTA; keep larger grids and
  MTP3 fused to share KV/scales.
- Select the single-M per-token V-prefetch pipeline from one layout/workgroup
  policy; static Hkv2 uses the tuned path while high-grid page128 transposed V
  retains the lower-register schedule.
- Support explicit partition counts up to 256 and a caller-configurable clamp for
  automatic partition selection.
- Use the native FlyDSL reducer for partition outputs.

## Test Plan

`op_tests/test_flydsl_pa_decode.py` has one generic parametrized pytest function
with 14 real-kernel/reference configurations covering QL1/2/3/4, page16/64/128,
plain/transposed V, BF16/FP16, scalar/per-token scales, automatic/explicit
partitions, Hkv2 prefetch policy boundaries, C200k, and a BS200 smoke case. The
same file also covers dynamic planning, graph replay, high-partition reducers,
and provides a direct correctness + performance CLI for larger sweeps.

```bash
pytest -q op_tests/test_flydsl_pa_decode.py

HIP_VISIBLE_DEVICES=5 python3 op_tests/test_flydsl_pa_decode.py \
  -d bf16 -b 200 -q 4 -s 16,1,128,200000 \
  --block-size 16 128 --trans-v 0 1 --per-token 1
```

Latest MI355 validation:

```text
285 passed in 190.50s
```

Black, Ruff, Python syntax, Bash syntax, and `git diff --check` also pass.

### Earlier representative cases

MI355/gfx950 CUDA Graph end-to-end latency, including compute and reduce:

| Case | Block size | PA Tile | PA Gluon | Speedup |
|---|---:|---:|---:|---:|
| small | 16 | 11.79 us | 12.24 us | 1.04x |
| small | 64 | 11.75 us | 12.07 us | 1.03x |
| batch | 16 | 14.20 us | 20.43 us | 1.44x |
| batch | 64 | 13.76 us | 20.32 us | 1.48x |
| MTP3 | 16 | 22.90 us | 41.52 us | 1.81x |
| MTP3 | 64 | 22.81 us | 41.08 us | 1.80x |
| MTP4-var | 16 | 28.72 us | 38.14 us | 1.33x |
| MTP4-var | 64 | 23.91 us | 38.53 us | 1.61x |
| long | 16 | 35.93 us | 43.69 us | 1.22x |
| long | 64 | 35.24 us | 43.52 us | 1.23x |

PA Tile wins all 10 cases with a 1.37x geometric-mean speedup. Workloads use
Hkv=1, head dimension 128, and transposed V. `small` is B3/Hq8/Q1/C1027
per-tensor; `batch` is B81/Hq8/Q1/C1027 per-tensor; `MTP3` is
B81/Hq16/Q3/C1027 per-token; `MTP4-var` is B81/Hq16/Q4/C1027 per-token with
variable KV lengths; and `long` is B81/Hq16/Q1/C8192 per-token.

### MI355: BS=200, MTP4, C=200k

Measured on MI355 (`gfx950`, 256 CUs) with `QL=4`, `Hq/Hkv=16/1`, `D=128`,
BF16 Q/O, FP8 E4M3FN KV, FP32 per-token KV scales, dense attention, and the
native FlyDSL reducer. Each value is the median of three `perftest` profiler
means (101 iterations, 2 warmups). Bandwidth is decimal logical TB/s; KV and
scales are counted once per PA call, not multiplied by MTP4.

#### Equal-length context, backend auto NP

Every sequence has context length 200,000. FlyDSL and Gluon use the same input,
block table, and reference within each row. Each backend uses its own automatic
partition selection: FlyDSL chooses NP8 for page16 and NP4 for page128, while
Gluon chooses NP3.

| Page | V layout | FlyDSL auto NP | FlyDSL us | FlyDSL TB/s | Gluon auto NP | Gluon us | Gluon TB/s | Speedup | `.005` |
|---:|---|---:|---:|---:|---:|---:|---:|---:|---|
| 16 | rank-4 | 8 | 2409.320 | 4.3898 | 3 | 6751.326 | 1.5666 | **2.8022x** | PASS |
| 16 | trans-V | 8 | 2372.853 | 4.4573 | 3 | 6719.592 | 1.5740 | **2.8319x** | PASS |
| 128 | rank-4 | 4 | 2387.181 | 4.4269 | 3 | 6441.091 | 1.6407 | **2.6982x** | PASS |
| 128 | trans-V | 4 | 3415.613 | 3.0940 | 3 | 6309.910 | 1.6748 | **1.8474x** | PASS |

For page128 rank-4 V, a separate controlled FlyDSL NP3/4/5/8 tuning sweep
measured 3.4866/4.4269/5.1639/4.4958 TB/s. Explicit NP5 is faster than auto
NP4 for that layout, but it is a manual tuning result and is not used in the
auto-NP comparison table above.

The earlier page-16 variable-length measurements showed 2.1779x (rank-4 V) and
2.1964x (trans-V) speedups, but both implementations failed the additional
strict `.005` check against the reference and also failed their mutual `.005`
comparison. Those rows are excluded from the performance claims above. Their
total context was 19,500,374 tokens, versus 40,000,000 for the equal-length rows.

## Submission Checklist


aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,717 @@
...
+        1
+        if max_context_partition_num == 1
+        else 1 << ((max_context_partition_num - 1).bit_length())
+    )
+    reduce_shuffle_offsets = [
+        offset for offset in (32, 16, 8, 4, 2, 1) if offset < reduce_width
+    ]
+
+    @flyc.kernel(known_block_size=[head_size, 1, 1])
+    def pa_decode_ps_reduce_kernel(
```
coderfeli: put kernel into kernel/ folder?

aiter/ops/flydsl/kernels/pa_decode_tile.py
```diff
@@ -0,0 +1,1306 @@
...
+        # returns page 0 for that out-of-range read instead of faulting (those
+        # tail tokens are masked out anyway).
+        bt_num_records_bytes = (
+            fx.Index(gpu.grid_dim.x) * fx.Index(max_blocks_per_seq) * 4
+        )  # int32 entries
+        bt_rsrc = buffer_ops.create_buffer_resource(
+            block_tables_ptr, max_size=False, num_records_bytes=bt_num_records_bytes
+        )
+        ks_rsrc = buffer_ops.create_buffer_resource(key_scale_ptr, max_size=True)
+        vs_rsrc = buffer_ops.create_buffer_resource(value_scale_ptr, max_size=True)
```
coderfeli: use make buf tensor?

aiter/ops/flydsl/kernels/pa_decode_tile.py
```diff
@@ -0,0 +1,1306 @@
...
+        # Per-tensor: a single global scale, read once. Per-token: read
+        # per-token instead (see _kv_scale_ops/_stage_kv_scale_to_lds below).
+        if const_expr(not per_token_kv):
+            key_scale = fx.Int32(
+                buffer_ops.buffer_load(
+                    ks_rsrc, fx.Int32(0), vec_width=1, is_scalar=True
```
coderfeli: use ptr_buf_scalar in tensor shim?

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,1157 @@
...
+    )
+
+    ret = {"gfx": get_gfx_runtime(), "partitions": num_partitions}
+    for name, fn in candidates.items():
+        out, us = run_perftest(fn)
+        err = checkAllclose(
+            reference.to(dtypes.fp32),
+            out.to(dtypes.fp32),
+            rtol=3e-2,
+            atol=3e-2,
```
Bernard-Liu: > atol=3e-2 looks too loose for these small-magnitude attention outputs and may hide substantial relative errors; the corresponding fixed-
  > length FP8 Gluon test uses 5e-3. The test organization is also unclear: the main randomized test is fixed-length (context_lengths is
  > created with torch.full), while variable-length coverage is scattered across specialized tests with different comparison criteria. Please
  > add clearly separated fixed-length and variable-length accuracy cases and define a justified threshold for each.

aiter/ops/flydsl/kernels/pa_decode_reduce.py
```diff
@@ -0,0 +1,250 @@
...
+    API limits the partition count to one wave, so the LDS path from the more
+    general Gluon implementation is unnecessary here.
+    """
+    if not 1 <= max_context_partition_num <= 64:
+        raise ValueError(
+            "max_context_partition_num must be in [1, 64], "
+            f"got {max_context_partition_num}"
+        )
+    if head_size <= 0 or head_size > 1024 or head_size % 64:
+        raise ValueError(
```
Bernard-Liu, line 34: This breaks Gluon’s Triton fallback for configurations unsupported by this reducer: the caller only catches ImportError. On gfx950, I reproduced this through pa_decode_gluon with BF16, head_size=96, ps=True, 4 partitions, and the C++ reducer disabled. The same input succeeds with FlyDSL disabled.

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,1157 @@
...
+        pytest.skip("ROCm is not available")
+    if pa_decode is None:
+        pytest.skip("FlyDSL is not available")
+    if get_gfx_runtime() not in SUPPORTED_GFX:
+        pytest.skip(f"pa_decode is unsupported on {get_gfx_runtime()}")
+
+    result = run_pa_decode_tile_case(
+        batch_size=3,
+        num_query_heads=8,
+        num_kv_heads=1,
```
Bernard-Liu: > Could we parameterize this test to cover num_kv_heads > 1 and both BF16 and FP16? The current CI case only exercises a single KV head with a BF16 query/output, leaving multi-KV-head indexing and the FP16 path without automated coverage.

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+        f"query shape {tuple(query.shape)}"
+    )
+    _, num_kv_heads, num_hgroups, block_size, hgroup_width = key_cache.shape
+
+    assert num_hgroups == head_dim // 16 and hgroup_width == 16
+    # Q staging loads each lane's head_dim//16 elements in <=8-wide pieces, so a
+    # chunk wider than 8 that isn't a multiple of 8 would silently drop its tail
+    # (head_dim 192/320/448/...). Reject those rather than return wrong results.
+    q_chunk = head_dim // 16
+    if q_chunk > 8 and q_chunk % 8 != 0:
```
Bernard-Liu: > Could we validate the supported head_dim range here before compiling or launching the tile kernel? The same shape may work with NP=1 but fail late with NP>1. Please define a consistent support range and add
> boundary coverage for both paths.

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+def get_recommended_splits(
+    num_sequences: int,
+    num_kv_heads: int,
+    split_kv_blocks: int = 1,
+) -> int:
+    props = torch.cuda.get_device_properties(torch.device("cuda"))
+    num_sm = props.multi_processor_count * 2
+    denom = max(1, num_sequences * num_kv_heads * split_kv_blocks)
+    n = cdiv(num_sm, denom) * split_kv_blocks
+    return max(4, min(n, 8))
```
Bernard-Liu: Could we document the tuning evidence for the minimum of 4 splits? With 256 CUs, 512 sequences, one KV head, and block_size=128, the current benchmark call would otherwise recommend 2. Comparing 2 versus 4 here would help establish whether the extra partitioning pays off at high concurrency.

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+        )
+    if compute_type != expected_fp8_dtype:
+        raise NotImplementedError(
+            f"pa_decode only supports FP8 compute ({expected_fp8_dtype}) on {arch}, "
+            f"got {compute_type}"
+        )
+
+    # ``ps`` is retained for drop-in API compatibility. Both partitioning
+    # policies are represented by the caller-provided partition count.
+    del ps
```
Bernard-Liu: why del ps parameter?

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+        key_cache.dtype == expected_fp8_dtype
+    ), f"pa_decode requires {expected_fp8_dtype} key cache on {arch}, got {key_cache.dtype}"
+    assert (
+        value_cache.dtype == expected_fp8_dtype
+    ), f"pa_decode requires {expected_fp8_dtype} value cache on {arch}, got {value_cache.dtype}"
+
+    assert (
+        query.stride(2) == 1
+    ), f"pa_decode requires a contiguous head_dim axis, got strides {query.stride()}"
+    assert (
```
Bernard-Liu: Could we explicitly validate tensor ranks, block-table row counts, and matching K/V block counts before launch? Context-length bounds should be documented or validated, with optional debug checks for GPU-resident values to avoid synchronization overhead.

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,1157 @@
...
+    for name, parameter in flydsl_parameters.items():
+        gluon_parameter = gluon_parameters[name]
+        assert parameter.kind == gluon_parameter.kind
+        assert parameter.default == gluon_parameter.default
+        assert public_parameters[name].kind == gluon_parameter.kind
+        assert public_parameters[name].default == gluon_parameter.default
+
+
+def test_pa_decode_maps_gluon_buffers_and_scale_layout(monkeypatch):
+    if not torch.cuda.is_available():
```
Bernard-Liu, line 81: Could this test reuse _require_gpu()? It currently checks only GPU/FlyDSL availability, but pa_decode rejects architectures other than gfx942/gfx950. On another ROCm GPU, this would fail rather than skip.

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+    prefetch_v = False
+    if (
+        arch == "gfx950"
+        and head_dim == 128
+        and block_size == 128
+        and trans_v
+        and not per_token_kv
+        and query_length * query_group_size <= 16
+    ):
+        num_cus = torch.cuda.get_device_properties(dev).multi_processor_count
```
Bernard-Liu, line 533: This batch-dependent flag creates two lazy kernel variants. On gfx950 with 256 CUs, 8 partitions, and 1 KV head, warming batch 32 leaves the batch-33 variant uninitialized, adding compilation or cache-loading overhead on first use. Could we document that warmup should execute both prefetch_v=False and True before serving or graph capture? Warming only the smallest and largest batches can miss the True interval.

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,503 @@
...
+    block_tables: torch.Tensor,  # [num_seqs, max_num_blocks_per_seq]
+    softmax_scale: float,
+    query_length: int,
+    max_context_partition_num: int,
+    context_partition_size: int = 256,
+    compute_type: torch.dtype = torch.bfloat16,
+    query_scale: torch.Tensor = None,  # [num_seqs * query_length, num_query_heads, 1] or [1]
+    key_scale: torch.Tensor = None,  # [num_blocks, num_kv_heads, kv_block_size, 1]
+    value_scale: torch.Tensor = None,  # [num_blocks, num_kv_heads, kv_block_size, 1]
+    exp_sums: torch.Tensor = None,  # [num_seqs, num_kv_heads, max_context_partition_num, query_group_size]
```
Bernard-Liu: The implementation allocates and validates these buffers using query_length * query_group_size, but the comments only specify query_group_size. Please update the last dimension of exp_sums/max_logits and the second-to-last dimension of temporary_output accordingly. Callers following the current comments would allocate undersized buffers for multi-partition MTP calls.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. The current 3e-2 value came from the benchmark sweep and is not a sufficiently justified correctness contract. The main randomized case is fixed-length, while the variable-length coverage is only in specialized tests. I will split those into clearly named fixed-length and variable-length accuracy cases and set separate thresholds from measured worst-case errors on gfx942/gfx950, using the Gluon 5e-3 threshold where the data supports it. I will keep this thread open until that is done.

aiter/ops/flydsl/kernels/pa_decode_reduce.py
fsx950223: Thanks for the concrete repro. Confirmed: `compile_pa_decode_ps_reduce` raises `ValueError` for `head_size=96` while the Gluon dispatch only treats `ImportError` as unsupported, so the Triton fallback is bypassed. I will convert unsupported-configuration failures at the FlyDSL adapter boundary to `ImportError` without broadly swallowing internal `ValueError` exceptions, and add a fallback regression for the BF16/D96/NP4 case. I will keep this thread open until the code and test land.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. The later large-negative regression now covers both BF16 and FP16, but the main randomized correctness case is still Hkv=1/BF16, so multi-KV-head indexing remains untested. I will parameterize the normal accuracy case over both query dtypes and at least one `num_kv_heads > 1` configuration, while keeping CI cost bounded with a small shape matrix.

aiter/ops/flydsl/pa_decode.py
fsx950223: Agreed. The current host guard rejects 192/320-style widths, but it is not the same contract as the reducer, which requires `head_dim` to be a multiple of 64 and at most 1024. Support can therefore still differ between NP=1 and NP>1. I will move one consistent support-range check before tile compilation and add NP=1/NP>1 tests at the supported boundaries and just outside them.

aiter/ops/flydsl/pa_decode.py
fsx950223: Good point. The floor of 4 was inherited from the existing Gluon heuristic; this PR currently has no 2-versus-4 evidence for B512/Hkv1/BS128 on a 256-CU device. I will benchmark exactly that case on gfx950 and either remove the floor or document the measured reason for retaining it. I will keep this thread open pending that data.

aiter/ops/flydsl/pa_decode.py
fsx950223: `ps` was retained only for signature compatibility with the Gluon API. In this implementation `max_context_partition_num` selects NP=1 direct output versus NP>1 FlyDSL reduction, so there is no separate path selected by `ps` today. However, silently discarding `ps=False` is misleading. I will make that unsupported behavior explicit or route it to the fallback at the dispatcher instead of leaving `del ps` as a silent no-op.

aiter/ops/flydsl/pa_decode.py
fsx950223: Agreed. The current unpacking and assertions cover some shape, dtype, and device conditions, but they do not explicitly verify ranks, `block_tables.shape[0] == num_seqs`, matching K/V block counts, or context-length bounds. I will add host-only metadata checks, document `0 <= context_len <= block_table_width * block_size`, and keep GPU-resident value validation behind an opt-in debug check to avoid synchronization in the normal path.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. This test should use `_require_gpu()`: its current local checks omit the architecture filter even though the host op rejects targets other than gfx942/gfx950. I will switch it to the shared helper so unsupported ROCm targets skip instead of fail.

aiter/ops/flydsl/pa_decode.py
fsx950223: Agreed. `prefetch_v` participates in the compile-cache key, and the predicate can be true only in a middle batch/workgroup range, so warming only the smallest and largest shapes is insufficient. I will document that warmup and graph capture must exercise both values and include the exact predicate; we can also consider prebuilding both variants.

aiter/ops/flydsl/pa_decode.py
fsx950223: Agreed. Those comments are incorrect. The actual scalar-buffer shape is `[num_seqs, num_kv_heads, max_context_partition_num, query_length * query_group_size]`, and `temporary_output` adds the `head_size` dimension. I will update the API comments to match the existing allocation and validation logic.

op_tests/test_flydsl_pa_decode.py
fsx950223: Implemented the test split in f4f84314a: fixed-length and variable-length accuracy now have separate named cases. The fixed-length path uses the Gluon FP8 contract of 5e-3, including the CLI sweep; varlen uses the corresponding Gluon varlen contract of 5e-2. The new cases cover independent page tables and an independent PyTorch reference. GPU CI is running, so I am leaving this thread open until both target architectures validate the thresholds.

aiter/ops/flydsl/kernels/pa_decode_reduce.py
fsx950223: Implemented in f4f84314a. The Gluon wrapper now capability-checks the optional FlyDSL reducer before selecting it, so BF16/D96/NP4 falls through to Triton. Direct FlyDSL compile validation still raises ValueError, and errors after a supported FlyDSL reducer is selected are no longer swallowed. Added regressions for both D96 fallback and propagation of an internal ImportError on a supported D128 configuration.

op_tests/test_flydsl_pa_decode.py
fsx950223: Implemented in f4f84314a. The normal fixed-length accuracy matrix now covers BF16 and FP16 with Hkv=1 and Hkv=2, across NP=1 and NP=4. The variable-length cases also run both query dtypes with Hkv=2.

aiter/ops/flydsl/pa_decode.py
fsx950223: Implemented in f4f84314a. The public FlyDSL path now applies one host-side domain before either path: D=64 or a multiple of 128 in [128, 1024]. Tests cover accepted and rejected dimensions for NP=1 and NP=4; D1024 NP1/NP4 are included as real accuracy launches rather than only mocked validation.

aiter/ops/flydsl/pa_decode.py
fsx950223: I audited the available PR and historical benchmark data, but none contains the requested gfx950 B512/Hkv1/page128 NP2-versus-NP4 comparison. The B64/page128 and older gfx942 results cannot be safely extrapolated. I therefore left the minimum-of-four behavior unchanged in f4f84314a and am keeping this thread open until that exact A/B can be run.

aiter/ops/flydsl/pa_decode.py
fsx950223: Implemented in f4f84314a. `ps` remains in the signature for API compatibility, but `ps=False` is now rejected explicitly with NotImplementedError instead of being silently discarded. A regression locks that behavior until a non-PS FlyDSL path exists.

aiter/ops/flydsl/pa_decode.py
fsx950223: Implemented in f4f84314a. The host now explicitly validates tensor types/ranks, output/query shape, block-table row count, matching K/V block and KV-head counts, layouts, dtypes, devices, and contiguity. The GPU-resident context/page-index value contract is documented without adding synchronization to the normal path, and structural negative cases were added to CI.

op_tests/test_flydsl_pa_decode.py
fsx950223: Implemented in f4f84314a. This test now uses `_require_gpu()`, so unsupported ROCm architectures skip consistently before invoking the FlyDSL host path.

aiter/ops/flydsl/pa_decode.py
fsx950223: Implemented in f4f84314a. The one-to-two-CTA-per-CU decision is now a named helper with B32/B33/B64/B65 boundary coverage for the 256-CU, Hkv1, NP8 example. Its docstring states that the result is a compile-cache specialization and that both True and False variants must be materialized before graph capture or latency-sensitive serving.

aiter/ops/flydsl/pa_decode.py
fsx950223: Implemented in f4f84314a. The FlyDSL signature, Gluon signature and detailed docstring, plus the tile/reducer pointer comments now describe the scalar buffers as `[num_seqs, num_kv_heads, NP, query_length * query_group_size]`; temporary output adds `head_size` and uses the query/output dtype.

.github/scripts/aiter_test.sh
```diff
@@ -124,6 +124,31 @@ for file in "${sharded_files[@]}"; do
...
+                        "${test_file}::test_recommended_splits_rejects_invalid_upper_clamp" \
+                        "${test_file}::test_v_prefetch_workgroup_interval" \
+                        "${test_file}::test_pa_decode_fixed_length_accuracy" \
+                        "${test_file}::test_pa_decode_variable_length_accuracy" \
+                        "${test_file}::test_large_negative_logits_preserve_online_softmax" \
+                        "${test_file}::test_additional_supported_head_dims_are_accurate" \
+                        "${test_file}::test_supported_head_dim_boundaries_are_accepted" \
+                        "${test_file}::test_unsupported_head_dim_is_rejected" \
+                        "${test_file}::test_ps_false_is_rejected" \
+                        "${test_file}::test_invalid_tensor_structure_is_rejected"
```
coderfeli: clean these strange tests.

aiter/ops/flydsl/kernels/pa_decode_reduce.py
```diff
@@ -0,0 +1,583 @@
...
+                zero_f,
+            )
+            scaled_sum = part_sum * part_scale
+            global_exp_sum = _wave_reduce_sum(scaled_sum)
+            if fx.const_expr(use_sinks):
+                sink_value = fx.Float32(sink_token[kv_head_idx * c_qgs + group_idx])
+                sink_scale = (global_max > neg_inf).select(
+                    fx.exp2(
+                        (sink_value - safe_global_max) * c_log2e,
+                        fastmath="fast",
```
coderfeli, line 378: use Fast math by default?

aiter/ops/flydsl/pa_decode.py
```diff
@@ -0,0 +1,753 @@
...
+
+
+def _flydsl_dtype_str(dtype: torch.dtype) -> str:
+    if dtype == torch.float32:
+        return "f32"
+    if dtype == torch.float16:
+        return "f16"
+    if dtype == torch.bfloat16:
+        return "bf16"
+    raise ValueError(f"Unsupported FlyDSL dtype: {dtype!r}")
```
coderfeli: use existed function.

aiter/ops/flydsl/kernels/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    if isinstance(ty, ir.VectorType):
+        n = ty.shape[0]
+        elems = []
+        for i in range(n):
+            scalar = _vector_dialect.extract(
+                raw, static_position=[i], dynamic_position=[]
+            )
+            elems.append(exp2_amdgcn_scalar(scalar))
+        return _vector_dialect.from_elements(ty, elems)
+    return exp2_amdgcn_scalar(raw)
```
coderfeli: so many exp? use exitsted func?

aiter/ops/flydsl/kernels/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def cdiv(numer: int, denom: int) -> int:
+    return (numer + denom - 1) // denom
```
coderfeli: use exitsted func

aiter/ops/flydsl/kernels/pa_decode_tile.py
```diff
@@ -0,0 +1,1821 @@
...
+        def _v_load16(byte_off):
+            return _v_load_fp8x16(byte_off).bitcast(fx.Int64)
+
+        context_len = fx.Int32(_ctxlen_load(seq)[0])
+        # Bound block_tables to its real extent: the last (partial) 256-token tile
+        # can index a page past ceil(context/block_size); the bounded resource
+        # returns page 0 for that out-of-range read instead of faulting (those
+        # tail tokens are masked out anyway).
+        bt_num_records_bytes = (
+            fx.Index(gpu.grid_dim.x) * fx.Index(max_blocks_per_seq) * 4
```
coderfeli: clean fx.Index

aiter/ops/flydsl/kernels/pa_decode_kernel.py
```diff
@@ -0,0 +1,1821 @@
...
+        lds_base = fx.recast_iter(fx.Uint8, lds.buf.ptr)  # byte-addressed base
+
+        def _lds_ptr(byte_off, elem_ty):
+            # Pin the element-size alignment explicitly (a bare recast off the
+            # byte base would gcd the alignment down to 1).
+            p = fx.add_offset(lds_base, fx.make_int_tuple(byte_off))
+            ptr_ty = fx.PointerType.get(
+                elem_ty.ir_type, fx.AddressSpace.Shared, dsl_size_of(elem_ty)
+            )
+            return fx.recast_iter(ptr_ty, p)
```
coderfeli, line 463: use existed func

aiter/ops/flydsl/kernels/pa_decode_kernel.py
```diff
@@ -0,0 +1,1821 @@
+# SPDX-License-Identifier: Apache-2.0
```
coderfeli, line 1: rename file to pa_decode_kernel

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,2170 @@
...
+        temporary_output=temporary_output,
+    )
+    assert dispatches == ["flydsl"]
+
+
+@pytest.mark.parametrize(
+    "batch_size,expected",
+    [(32, False), (33, True), (64, True), (65, False)],
+)
+def test_v_prefetch_workgroup_interval(batch_size, expected):
```
coderfeli: clean these strange tests to kernel test only.

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,2170 @@
...
+        num_kv_heads=2,
+        head_dim=128,
+        block_size=128,
+        query_dtype=query_dtype,
+        num_partitions=4,
+        tolerance=VARIABLE_LENGTH_ACCURACY_TOLERANCE,
+    )
+
+
+def _valid_pa_decode_arguments(head_dim=128, num_partitions=1):
```
coderfeli: Clean

op_tests/test_flydsl_pa_decode.py
```diff
@@ -0,0 +1,2170 @@
...
+        "query_length": 1,
+        "max_context_partition_num": num_partitions,
+        "compute_type": quant_dtype,
+        "key_scale": scale,
+        "value_scale": scale,
+    }
+
+
+def test_ps_false_is_rejected():
+    _require_gpu()
```
coderfeli: ...

.github/scripts/aiter_test.sh
fsx950223: Agreed. This shell-side pytest node list has become brittle. I will clean it together with the test-file cleanup below: retain a compact set of tests that launch the real kernel, make the standalone runner propagate correctness failures, and then remove this special case. Deleting the block first would lose the gate because `main()` currently records `checkAllclose` errors without returning a failure.

aiter/ops/flydsl/kernels/pa_decode_reduce.py
fsx950223: I checked the generated gfx950 ISA for this. Setting reducer-wide `fast_fp_math=True` changes NP8 from 185 to 169 instructions and NP128 from 376 to 318, but it also removes the `v_cmp_lg_f32` guards around the `-inf` partition sentinels and replaces IEEE division with reciprocal/multiply. I therefore kept fast math selective rather than changing the reducer's numerical contract. A global-fast version would need separate empty-partition, sink, and accuracy validation.

aiter/ops/flydsl/pa_decode.py
fsx950223: Done in 60e63316e. The wrapper now reuses `tensor_shim.get_dtype_str`.

aiter/ops/flydsl/kernels/utils.py
fsx950223: Done in 60e63316e. The custom scalar intrinsic and vector extract/rebuild helpers are removed; scalar and vector sites now use FlyDSL's `fx.exp2(..., fastmath="fast")`. I compiled Q1/MTP4, page16/page128, and per-tensor/per-token representatives on both gfx942 and gfx950; all eight final ISA files are byte-for-byte identical to the previous implementation, including VGPR/SGPR/LDS usage.

aiter/ops/flydsl/kernels/utils.py
fsx950223: Done in 60e63316e: the local `cdiv` helper is removed. Host/static divisions and the positive runtime tile counts use the equivalent integer expression directly. I also tested `fx.ceildiv` in the hot kernel, but it changed the generated prologue and raised the gfx950 MTP4 page16 specialization from 244 to 250 VGPRs, so I kept the ISA-equivalent expression here.

aiter/ops/flydsl/kernels/pa_decode_tile.py
fsx950223: Done in 60e63316e. The block-table byte extent now uses explicit `fx.Int64` arithmetic instead of `fx.Index`; generated ISA remains byte-for-byte identical in the checked gfx942/gfx950 specializations.

aiter/ops/flydsl/kernels/pa_decode_kernel.py
fsx950223: I tested replacing this with the existing `lds_typed_ptr` / `lds_vec_load` helpers. In the MTP4 page16 specialization, nine `ds_read_b128` instructions became eighteen `ds_read2_b32` instructions, total instructions increased by 38, and VGPR use increased from 250 to 254 against the otherwise identical candidate. A bare `fx.recast_iter(elem_ty, p)` also fails because the byte pointer has alignment 1. I kept this alignment-preserving local helper for now; if you meant a different existing helper, please point me to it.

aiter/ops/flydsl/kernels/pa_decode_kernel.py
fsx950223: Done in 60e63316e. The module is now `kernels/pa_decode_kernel.py`, and its host import and documentation reference were updated.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. I will make the follow-up test cleanup kernel-focused: remove host-policy and pure-mock dispatch tests while retaining numerical cases that execute the real GPU kernel, including NP1/NP>1, both V layouts, per-token MTP, empty contexts, and page-boundary coverage.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. `_valid_pa_decode_arguments` only supports the host-contract/mock tests, so it will be removed with that test cleanup.

op_tests/test_flydsl_pa_decode.py
fsx950223: Agreed. `test_ps_false_is_rejected` is host API policy rather than kernel correctness, so it will be removed from this kernel test file in the cleanup follow-up.

coderfeli (approved):

PR #4334, opened 2026-07-22
[Triton/Gluon] perf(fp8_mqa_logits): runtime-autotune the gfx942 indexer tile (LDS-safe candidates)

EricKing626:
### Why the DSA indexer kernel regressed (and why it over-shrinks)

In MI325:
Commit [bf8569d (#3257)](https://github.com/EricKing626/aiter/commit/bf8569da781d88ff8a38ec63bb5fb0dd7e68fd76) correctly fixed a real JIT/LDS-overflow crash for the DeepSeek-V4 indexer on MI300X, but its occupancy=2 gate keys on the whole gfx942 arch — which also covers MI325X — so it needlessly shrinks the head_size=128 tile to (64,1) and regresses GLM-5.2 long-context prefill (~2×) on a shape/Triton version where (128,2) actually fits.

image

The DSA indexer scores query vs KV with the Triton kernel `_fp8_mqa_logits`.
On **gfx942** the wrapper defaults to a tile of `(BLOCK_KV=128, num_stages=2)`.
Each CU has **64 KB of LDS**; the double-buffered KV tile lives in LDS (Q and the
fp32 scores stay in registers). The heuristic estimates:

LDS = occupancy × num_stages × BLOCK_KV × head_size   (fp8 = 1 byte)

and shrinks the tile when that exceeds `0.9 × 64 KB = 57.6 KB`.

#### The math for `head_size = 128`

| Case | Tile (BLOCK_KV, stages) | LDS estimate | ≤ 57.6 KB? | Result |
|---|---|---|:--:|---|
| `(128, 2)` · occ=2 | 2×2×128×128 = 65,536 B = 64 KB | 64 > 57.6 | ✗ shrink to (64, 1) → **slow** |
| `(64, 1)` · occ=2 | 2×1×64×128 = 16,384 B = 16 KB | 16 ≤ 57.6 | current pick — indexer **82 µs/tok** |
| `(128, 2)` · occ=1 | 1×2×128×128 = 32,768 B = 32 KB | 32 ≤ 57.6 | ✓ kept after fix — indexer **24 µs/tok** |

With `occupancy = 2` the default tile is judged not to fit, so it drops to `(64, 1)` —
halving the KV tile and removing the software pipeline, which roughly **halves throughput
(indexer 24 → 82 µs/token)**.

But `occupancy` is only a performance preference. The real hard limit that triggers a
JIT/LDS-overflow abort is **a single workgroup exceeding 64 KB** — i.e. `occupancy = 1`.
At `occupancy = 1` the default tile is only **32 KB** and fits comfortably.

(128, 2) is a large, double-buffered KV tile (32 KB LDS) that maximizes MFMA throughput and wins on long-context prefill, while (64, 1) is a small, un-pipelined tile (8 KB LDS) that trades tile throughput for higher occupancy and wins only when the kernel is latency/occupancy-bound (short KV or small grids that can't fill a big tile).

> **Takeaway:** the guard's intent (avoid a JIT abort when the tile can't fit) is met by an
> `occupancy = 1` check. Requiring `occupancy = 2` needlessly shrinks the tile on MI325X,
> where `(128, 2)` actually runs and is ~2× faster.


cagrikymk:
@EricKing626 Is it possible to sink the result of auto tuning into config files or arch specific config selector?
I think this kernel is used for GLM-5.x and deepseek V{3,4}. So, basically there are 2 cases for now.

Runtime or capture time auto-tuning makes the kernel time variable as some random noise might cause different configs to be selected etc. and lead to different behavior. This issue can get even worse due to the unpredictability in batching (seq_len piece). The autotuning can trigger whenever there is an OOD batch size.

Edit/correction: Looks like your bucket set is small and all powers of 2, it wouldnt be affected by OOD batches. However, I think it is still better to not do tuning in runtime.

What do you think?

EricKing626:
> @EricKing626 Is it possible to sink the result of auto tuning into config files or arch specific config selector? I think this kernel is used for GLM-5.x and deepseek V{3,4}. So, basically there are 2 cases for now.
> 
> Runtime or capture time auto-tuning makes the kernel time variable as some random noise might cause different configs to be selected etc. and lead to different behavior. This issue can get even worse due to the unpredictability in batching (seq_len piece). The autotuning can trigger whenever there is an OOD batch size.
> 
> Edit/correction: Looks like your bucket set is small and all powers of 2, it wouldnt be affected by OOD batches. However, I think it is still better to not do tuning in runtime.
> 
> What do you think?

Agreed, tuning shouldn't happen at runtime. Reworked: no do_bench on the runtime path.

Tile now comes from a checked-in configs/gfx942-FP8_MQA_LOGITS.json (HEAD_SIZE=<d> → SEQ_LEN_LEQ_<bucket>), read via a cached _get_config() — same pattern as extend_attention. fp8_mqa_logits() also takes a config= override.
Tuning moved offline to op_tests/op_benchmarks/triton/tune_fp8_mqa_logits.py, which regenerates that JSON.
Only guard left is static arithmetic: shrink the configured tile until the KV tile fits 64KB for one workgroup, floor (64, 1) = the #3257 crash-safe DSv4 config. Pure function of (BLOCK_KV, num_stages, head_size), so it can't vary run to run. No-op for head_size=128, so GLM-5.2 keeps (128, 2).
Archs without a config file, and the gfx950/gfx1250 Gluon path, are unchanged.

The JSON currently only has the entries I've validated on MI300X; I'll fill in the rest from a full tuner run before merge.

@cagrikymk From your perspective, what cases and machines do we need to test?

yuzhouo7:
Thanks for working on this. Adding **MI308X (gfx942)** results: GLM-5.3 end-to-end and op-level
tests covering both 32 and 64 heads at head_size 128.

I tested a **local port of the tile-selection logic from ad9ab5a onto aiter 0.1.21.dev48**, not an
unmodified checkout of this PR. The kernel launch was unchanged.

### GLM-5.3 end-to-end

8× MI308X, TP8, SGLang v0.5.19, 32K chunked prefill, 512 output tokens. Prefix cache was flushed
before each run. Both v0.5.19 arms carry an equivalent startup-only fix for the `format_ue8m0` load
crash (sglang#38446), unrelated to this PR.

**TTFT p95, median of 3 repeats.** Percentages below are increases over the older-runtime reference
(v0.5.15.post1 / aiter 0.1.17.dev110, tile `(128,2)`): **22,714 ms** at 32K/C=2 and **27,854 ms** at
64K/C=1. Positive means slower; C is concurrency.

| Triton | Tile selection | 32K, C=2 | 64K, C=1 |
|---|---|---:|---:|
| 3.7.0 | Stock → `(64,1)` | +43.5% | +57.3% |
| 3.7.0 | Ported PR logic → `(128,2)` | +2.7% | +3.8% |
| 3.8.0 | Stock → `(64,1)` | +28.7% | +47.5% |
| 3.8.0 | Ported PR logic → `(128,2)` | +2.7% | +2.0% |

Stock and ported results were collected on consecutive days with separate paired baseline controls,
not in the same session.

For one 64K prefill, rank-0 cumulative `_fp8_mqa_logits_kernel` GPU time was **21.1 s stock →
6.246 s ported**, versus 5.242 s on the older runtime: most, but not all, of the regression was
removed.

GSM8K (n=500): **454/500 reference, 456/500 with Triton 3.7.0, 450/500 with 3.8.0**. All tested
needle cases passed; RULER results varied across lengths/subtasks. These results do not establish
quality equivalence.

### Op-level: 32/64 heads × 128

Single GPU, both Triton versions, identical FP8 inputs, causal masks. Tested q/KV lengths: 1K/1K,
4K/4K, 16K/16K, 32K/32K, and 32K/131K. Each head/tile combination ran in a separate process.

All tested `(128,2)` cases completed without failure. Relative to `(64,1)`:

| Heads | Speedup at q=32K / KV=131K | Logits comparison |
|---|---:|---|
| 32 | 3.52–3.53× | Bitwise identical |
| 64 | 1.88–1.91× | Identical −inf masks; not bitwise identical, max absolute difference 6.1e-5 |

The 64-head case matches the head geometry discussed in #3257; I did not run DSv4 end-to-end or
compare top-k index sets.

PR #4342, opened 2026-07-23
Fix fused_qk_rope_concat_and_cache_mla for DCP

yitingw1:
## Motivation

Both decode variants (`..._per_head_kernel_impl`, `..._kernel_opt`) start with `if (slot_idx < 0) return;`, which also skips Q RoPE + `q_out`. Under DCP a non-owned token has `slot_mapping = -1` but its query still needs RoPE (every rank attends all queries after the head all-gather) → `q_out` uninitialized → wrong accuracy. But making Q RoPE *unconditional* would regress non-DCP: cudagraph-padded tokens (`slot = -1`) would pointlessly compute Q RoPE + `q_out` instead of early-returning.

## Fix

Add `compute_all_q_rope` as the **last param (default `false`)**, threaded through impl → wrapper → CALL macros → host → pybind → Python stub:

- **`false` (default, non-DCP):** `if (!compute_all_q_rope && slot_idx < 0) return;` — original early-return kept → **no non-DCP regression**; callers need no change (stub default + `bound.apply_defaults()`).
- **`true` (DCP):** compute Q RoPE + `q_out` unconditionally; guard only KV-cache writes with `slot_idx >= 0`.
- Add `max_position` + `clamp(pos)` (padded tokens may carry stale positions).
- `slot_idx >= 0` path byte-identical; only per_head + opt decode variants change (prefill untouched).

## Test

- DeepSeek-R1 TP8 gsm8k: DCP8 bf16 0.9492 / fp8 0.9553, non-DCP 0.9492 — no regression.

## Submission Checklist


yzhou103:
LGTM

yzhou103:
General decode kernel not covered. fuse_qk_rope_concat_and_cache_mla_kernel_general_kernel still has a hard if (slot_idx < 0) return;. If compute_all_q_rope=true is passed but dispatch falls into this path (kv_lora_rank != 512 or pe_dim != 64), the flag is silently ignored. Suggest either threading the param through, or adding a host-side guard:
AITER_CHECK(!compute_all_q_rope || (kv_lora_rank == 512 && pe_dim == 64),
    "compute_all_q_rope is only supported for standard DeepSeek MLA dimensions");

yitingw1:
> General decode kernel not covered. fuse_qk_rope_concat_and_cache_mla_kernel_general_kernel still has a hard if (slot_idx < 0) return;. If compute_all_q_rope=true is passed but dispatch falls into this path (kv_lora_rank != 512 or pe_dim != 64), the flag is silently ignored. Suggest either threading the param through, or adding a host-side guard: AITER_CHECK(!compute_all_q_rope || (kv_lora_rank == 512 && pe_dim == 64), "compute_all_q_rope is only supported for standard DeepSeek MLA dimensions");

Done by adding a host-side guard.

yzhou103 (approved):

PR #4345, opened 2026-07-23
[MLA v4 nm] Force occupancy-only split selection (ignore total_kv)

liyjiang:
get_meta_param gains an `ignore_total_kv` flag (default 0 = legacy). When set, the auto split-count search uses only the CU-occupancy factor (drops the avg_kv HBM-efficiency term) and skips the fp8 min-block cap, giving a pure occupancy-driven pick independent of per-seq KV length.

mla_decode_fwd_v4_nm forces this on internally so v4 nm decode selects splits purely by CU occupancy. Because this can pick num_kv_splits > 1 even for short KV (where the fp8 cap previously forced single-pass), test_v4_nm_sink now compares the authoritative BF16 `output` buffer (populated for both single- and multi-pass) instead of the single-pass-only packed-BF16 `logits` view.

### DIFF
--- a/aiter/mla.py
+++ b/aiter/mla.py
@@ -124,8 +124,21 @@ def _fwd_kernel_stage2_asm(
 
 @functools.lru_cache()
 def get_meta_param(
-    num_kv_splits, bs, total_kv, nhead, max_seqlen_q, dtype, tg_factor=1
+    num_kv_splits,
+    bs,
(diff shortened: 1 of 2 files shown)


amd-ruitang3:
LGTM

amd-ruitang3 (approved):

PR #4366, opened 2026-07-24
feat: support fp32 chunk states in GDN prefill

junna2016:
## Motivation

The Gated Delta Net (GDN) prefill implementation introduced in [ROCm/aiter#2774](https://github.com/ROCm/aiter/pull/2774) always materializes the per-chunk hidden-state tensor `h` in BF16, even when `state_dtype=torch.float32` is requested.

RTP-LLM uses FP32 recurrent states in its baseline GDN prefill path. Supporting FP32 chunk states in AITER is required to:

- Preserve the requested `state_dtype` consistently across `initial_state`, per-chunk `h`, and `final_state`.
- Avoid precision loss when materializing chunk-boundary state snapshots.
- Enable a fair performance and accuracy comparison with the RTP-LLM FP32 baseline.
- Avoid an additional BF16-to-FP32 conversion tensor and standalone cast kernel.

## Technical Details

This change extends the HIP GDN prefill path to support both FP32 and BF16 per-chunk hidden states.

### Python wrapper

- Allocate the per-chunk `h` tensor using the resolved state dtype instead of always using `torch.bfloat16`.
- Keep the following tensors on a consistent state dtype:
  - `initial_state`
  - per-chunk hidden-state snapshots `h`
  - `final_state`

The resulting behavior is:

```text
state_dtype=torch.float32  -> h.dtype=torch.float32
state_dtype=torch.bfloat16 -> h.dtype=torch.bfloat16

### CODE REVIEW

**Review by yiijin — CHANGES_REQUESTED** [review]
Thanks for adding FP32 chunk-state support. FP32 h_snapshot is useful as an opt-in, but coupling it to initial_state/final_state by default may not match existing integrations.

I checked vLLM and SGLang:

- vLLM: initial_state follows the configurable SSM-cache dtype; h_snapshot follows k.dtype, normally BF16; raw final_state is FP32 and is cast to ssm_state.dtype when persisted. h_snapshot is temporary and is not stored in the SSM pool.
- SGLang: initial_state/final_state follow the SSM-pool dtype, FP32 by default and optionally BF16; h_snapshot follows k.dtype. Selected snapshots may be stored in the pool for prefix tracking through an explicit .to(ssm_states.dtype) conversion.
These implementations intentionally separate internal FP32 accumulation, persistent-state dtype, and snapshot dtype. They do not require all three to match.

Because state_dtype=None resolves to FP32, this PR changes the default datatype of h_snapshot in both triton/hip backend kernel from BF16 to FP32. Since h_snapshot is [B, num_chunks, H, V, K], this doubles its memory traffic. With T=8192, H=16, and V=K=128, the buffer grows from approximately 64 MiB to 128 MiB.

I suggest decoupling the policies:

```
state_dtype: controls initial_state/final_state
snapshot_dtype: controls h_snapshot
snapshot_dtype = k.dtype if snapshot_dtype is None else snapshot_dtype
```

If FP32 snapshots are required, please use an independent compile-time specialization:

```
template <..., bool STATE_BF16, bool SNAPSHOT_BF16>
__global__ void chunk_gated_delta_rule_fwd_h_hip_kernel(...);
```

The BF16 specialization should retain coalesced_vk_store_from_transpose unchanged, while only the FP32 specialization directly stores the FP32 register state. This avoids runtime dtype branches, extra FP32 LDS, or changes to BF16 register usage.

To guarantee that the BF16 path remains unchanged, consider a constexpr specialization:

```
if H_IS_FP32:
    b_h = b_h.to(b_q.dtype)
b_o = tl.dot(b_q, tl.trans(b_h), acc=b_o)
```

The new h.dtype == state_dtype assertions also cover Triton, but its producer still allocates h = k.new_empty(...). Therefore, Triton with FP32 state and BF16 k still produces BF16 snapshots. Please either limit these assertions to HIP or implement the snapshot-dtype policy consistently.

Finally, please add separate BF16/FP32 benchmarks for chunk_gated_delta_rule_fwd_h_hip_kernel and chunk_fwd_kernel_o_opt_vk. Inspect TTIR/LLVM IR/GCN ISA to confirm that the BF16 specialization has no extra conversions or branches, preserves the coalesced store path, and does not increase VGPR/SGPR/LDS usage, spills, or reduce occupancy.

junna2016:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/junnaxing/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/junnaxing/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<meta charset=utf-8>

<body link="#0563C1" vlink="#954F72">


rtp-llm   MI308X TP=1 | Qwen3.5-4B |   |  
-- | -- | -- | --
  | 输入长度2k，输出20 | 并发1 |  
  |   |   |  
prefill gdn FP32 state | kernel name | kernel execute time/us | 加速比
origin triton | chunk_gated_delta_rule_fwd_kernel_h_blockdim64 | 233.8 |  
aiter hip | chunk_gated_delta_rule_fwd_h_hip_kernel | 128.2 | 1.82371



</body>

</html>

yiijin (changes requested):
Two compatibility issues remain and are reproducible on the current PR:

1. Positional API regression

snapshot_dtype was inserted before existing optional arguments. The same positional call succeeds on the base commit but fails on this PR because the former use_exp2=False binds to snapshot_dtype=False:

`ValueError: snapshot_dtype must be fp32 or bf16, got False.`

Please append snapshot_dtype after all existing parameters, or make it a keyword-only argument without changing the previous positional order.

1. FP16 input + BF16 snapshot fails in Triton K6

With FP16 q/k/v and snapshot_dtype=torch.bfloat16, K6 calls tl.dot with FP16 and BF16 operands:
`Both operands must be same dtype. Got fp16 and bf16`

H_IS_FP32 only casts FP32 snapshots. Please cast b_h whenever its dtype differs from b_q—an unconditional b_h = b_h.to(b_q.dtype) should allow the compiler to eliminate no-op casts.

yiijin (dismissed):

valarLip (dismissed):

valarLip (approved):

PR #4371, opened 2026-07-24
Implement FlyDSL version of fused_qk_norm_mrope_3d_cache_pts_quant_shuffle

amd-meskelin:
## Motivation

Implements a FlyDSL version of the existing fused_qk_norm_mrope_3d_cache_pts_quant_shuffle kernel with the goal of improving performance for larger shapes.

## Technical Details

- The FlyDSL kernel splits the KV and Q operations to separate kernel launches to improve write coalescing and reach better HBM utilization, among other optimizations.

- The kernels support a subset of features, not including GPT-J rope, non-shuffled layouts or partial rotary.

- There is a `match_hip` parameter (default True) to enforce matching BF16 truncation before fp32 to allow bit-exactness with HIP to be toggled on or off; the non-truncated case should be correct in principle.

## Test Plan and results

The test file (`op_tests/test_flydsl_fused_qk_norm_mrope_cache_quant_shuffle.py`) verifies the new implementation against the existing HIP implementation. Results match the existing HIP exactly on the tested parameters with `match_hip = True` and differ by bf16 truncation accuracy with `match_hip = False`.

## Submission Checklist


aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+                # negative zero as 0x80. In e4m3fnuz that byte is NaN, so
+                # match the established qk_norm_rope_quant conversion path
+                # by mapping tiny negative inputs to +0 before conversion.
+                zero = fx.Float32(0.0)
+                neg_underflow = fx.Float32(-(2.0**-8))
+                s0_tiny_neg = (s0 < zero) and (s0 > neg_underflow)
+                s1_tiny_neg = (s1 < zero) and (s1 > neg_underflow)
+                s0 = s0_tiny_neg.select(zero, s0)
+                s1 = s1_tiny_neg.select(zero, s1)
+            packed = fx.Int32(fx.rocdl.cvt_pk_fp8_f32(T.i32, s0, s1, fx.Int32(0), 0))
```
msaffari-amd, line 529: Should s0 and s1 be clamped to the valid FP8 range before conversion? Otherwise, large values may become NaN instead of the maximum FP8 value.

aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+    """
+    from aiter.jit.utils.chip_info import get_gfx as _get_gfx
+
+    if _get_gfx() == "gfx1250":
+        raise NotImplementedError(
+            "flydsl_fused_qk_norm_mrope_3d_cache_pts_quant_shuffle is wave64-"
+            "only (gfx942/gfx950); gfx1250 (wave32) needs a dedicated "
+            "variant, analogous to qk_norm_rope_quant_gfx1250.py, which does "
+            "not exist yet."
+        )
```
msaffari-amd: This kernel requires wave64, but this check only rejects gfx1250. Should this instead allow only supported wave64 targets such as gfx942 and gfx950, so other wave32 targets cannot run it?

aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+            f"head_size//2={head_size // 2}"
+        )
+    if block_size % x != 0:
+        raise ValueError(f"block_size ({block_size}) must be a multiple of x ({x})")
+    if (head_size * block_size) % 16 != 0:
+        raise ValueError(
+            f"head_size*block_size ({head_size * block_size}) must be a "
+            "multiple of 16 (dwordx4 K-cache run size)"
+        )
+    lds_bytes = 2 * head_size * block_size * k_cache.element_size()
```
msaffari-amd: lds_bytes counts only k_lds and v_lds, but mapping_ok also uses shared memory. For example, head_size=256 and block_size=64 already use all 64 KiB for K and V, so mapping_ok exceeds the limit. Could you include it in this check?

aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+        is_interleaved=is_interleaved,
+        gemma_norm=gemma_norm,
+        cache_is_fp8=cache_is_fp8,
+    )
+    # num_tokens need not be a multiple of block_size -- the last page-block
+    # may be a ragged tail with fewer than block_size real tokens; the
+    # kernel's Phase 1 guards every per-token read/write behind
+    # `tok < num_tokens` and zero-fills the corresponding LDS staging rows
+    # for the out-of-range tail positions (see _build_kv_kernel).
+    num_page_blocks = _ceil_div(num_tokens, block_size)
```
msaffari-amd, line 1195: Could you validate that block_size > 0? Currently, zero passes the multiple-of-x check but later causes division by zero in _ceil_div.

aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+            sumsq_local = fx.Float32(0.0)
+            if tid < RMS_GROUP:
+                for i in range_constexpr(PROD_VEC_SIZE):
+                    norm_col = fx.Int32(tid) * PROD_VEC_SIZE + i
+                    norm_x = fx.Float32(qkv[tok, bid_x, norm_col])
+                    sumsq_local = sumsq_local + norm_x * norm_x
+            for k in range_constexpr(VEC_PAIRS):
+                col = tid + WAVE * k
+                x0 = fx.Float32(qkv[tok, bid_x, col])
+                x1 = fx.Float32(qkv[tok, bid_x, col + HALF])
```
msaffari-amd: Are these duplicate Q/K loads intentional? The row is first loaded for sumsq_local and then loaded again into x0/x1. Reusing the loaded values could reduce Q/K input traffic, although the reduction order may need to remain unchanged for HIP parity.

aiter/ops/flydsl/kernels/fused_qk_norm_mrope_3d_cache_pts_quant_shuffle.py
```diff
@@ -0,0 +1,1131 @@
...
+                                    v_out[tok, head, col] = vb0
+                                    v_out[tok, head, col + HALF] = vb1
+                else:
+                    # Tail page-block padding row: no real token here (would
+                    # be OOB on qkv/positions, and on k_out/v_out which are
+                    # sized to exactly num_tokens rows) -- zero-fill instead.
+                    for p in range_constexpr(VEC_PAIRS):
+                        col = fx.Int32(lane) + WAVE * p
+                        if col < HALF:
+                            k_lds_view[token_local, col] = CACHE_FX_TYPE(0)
```
msaffari-amd: For T=1 and block_size=64, this processes one real row but fills 63 rows with zeros. Could a direct-write path be used for small or partial pages to avoid this extra work?

PR #4387, opened 2026-07-26
Limit attention kernel dispatch to supported GPUs

rlrs:
## What this changes

Keep attention kernel selection within the GPU architectures supported by each
implementation:

- Paged-attention assembly kernels are selected only on gfx942 and gfx950,
  where AITER ships the required binaries.
- The current variable-length FlyDSL kernel is imported only on gfx1201, which
  is the architecture it supports.
- Other GPUs continue through the existing HIP, CK, or Triton fallback.

Previously, a sufficiently large paged-attention batch selected the assembly
path on gfx90a. Since no gfx90a binary is packaged, vLLM failed while capturing
its decode graphs. The FlyDSL import had a similar problem: it happened before
the architecture check inside the kernel wrapper.

The open-PR search did not find another change addressing these two dispatch
problems.

## Results

Tested on an MI250X (gfx90a) with BF16 MQA, four query heads per rank, one KV
head, and a 128-dimension head:

| Context | Existing Triton decode | AITER HIP fallback |
| --- | ---: | ---: |
| 128K | 2.450 ms | 0.621 ms |
| 256K | 4.595 ms | 1.205 ms |

I also started NVIDIA Nemotron 3 Super with vLLM at 128K context and TP8. Full
graph capture completed, and 1K, 8K, and 40K prompt smoke tests all returned
successfully. The 40K case exercises a paged continuation after the initial
prefill.

Validation:

```text
ruff check --ignore E712,F841 \
  aiter/ops/attention.py aiter/ops/mha.py op_tests/test_pa_dispatch.py
python -m py_compile \
  aiter/ops/attention.py aiter/ops/mha.py op_tests/test_pa_dispatch.py
python -m pytest -q op_tests/test_pa_dispatch.py
# 1 passed
```

The change was prepared with AI assistance and reviewed and tested by the
submitter.

### DIFF
--- a/aiter/ops/attention.py
+++ b/aiter/ops/attention.py
@@ -169,6 +169,9 @@ def _should_use_asm_kernel(
     kv_cache_tensor_dtype: torch.dtype,
     high_precision: int,
 ) -> bool:
+    if get_gfx() not in ("gfx942", "gfx950"):
+        return False
+
     # ASM kernel only supports head_size == 128; all other head sizes use HIP.
     if head_size != 128:
         return False
--- a/aiter/ops/mha.py
+++ b/aiter/ops/mha.py
@@ -3593,32 +3593,32 @@ def flash_attn_varlen_func(
             sink_ptr,
         )
 
-    # FlyDSL path returns result if supported, None otherwise.
-    from .flydsl.fmha_kernels import flydsl_flash_attn_varlen_func
+    if get_gfx() == "gfx1201":
+        from .flydsl.fmha_kernels import flydsl_flash_attn_varlen_func
 
-    _flydsl_result = flydsl_flash_attn_varlen_func(
-        q,
-        k,
-        v,
-        cu_seqlens_q,
-        cu_seqlens_k,
-        max_seqlen_q,
-        max_seqlen_k,
-        softmax_scale=softmax_scale,
-        causal=causal,
-        return_lse=return_lse,
-        dropout_p=dropout_p,
-        window_size=window_size,
-        bias=bias,
-        alibi_slopes=alibi_slopes,
-        deterministic=deterministic,
-        return_attn_probs=return_attn_probs,
-        block_table=block_table,
-        out=out,
-        sink=sink_ptr,
-    )
-    if _flydsl_result is not None:
-        return _flydsl_result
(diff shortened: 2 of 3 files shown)


davetha:
Data point from a gfx90a (MI210) install, in case it's useful for scoping this.

**On gfx90a we observe zero ASM paged-attention dispatch through vLLM — but the cause appears to be upstream of architecture gating.**

In vLLM `0.23.1.dev1+g9ddef7117`, `pa_fwd_asm` occurs exactly six times in the entire tree:

```
v1/attention/backends/rocm_aiter_fa.py:1152   a comment
_aiter_ops.py:141                             a docstring
_aiter_ops.py:2536                            the wrapper definition
_aiter_ops.py:2554                            its docstring
_aiter_ops.py:2556                            from aiter import pa_fwd_asm
_aiter_ops.py:2558                            the wrapper calling through
```

There is no call site — nothing invokes `rocm_aiter_ops.pa_fwd_asm()`. The AITER FA backend's decode path calls `torch.ops.aiter.paged_attention_v1` (line 1350) and nothing else. On the aiter side, `pa_fwd_asm` in `ops/attention.py` has no architecture guard; it's a plain wrapper.

Consistent with that, across every server log we have, an ASM `pa` `.co` has loaded **zero** times, while `fmha_v3_fwd` ASM loads normally. `gqa_ratio: 8` appears 92 times in JIT compilations — the ratio matching `pa_bf16_noquant_gqa8_1tg_4w.co` — and the HIP template was selected every time.

So on this configuration the ASM paged-attention path isn't reachable regardless of architecture. If the vLLM graph-capture failures motivating this PR came from a different call path or version, it may be worth confirming which one, since gating `get_gfx()` wouldn't affect the path above.

**On the binaries themselves:** we have `pa` `.co` objects present for gfx90a via architecture repatching (not shipped upstream — upstream ships gfx942/gfx950). They validated 48/48 numerically exact against a PyTorch reference in **standalone** testing. That was standalone only; we've never gotten them to dispatch through a server, so this is not a claim that gfx90a serving works today.

The modest suggestion: gating on **binary presence** rather than architecture string would be more robust and wouldn't foreclose configurations where the artifacts do exist — and it would fail loudly rather than silently falling back.

Happy to run anything on MI210 if a datapoint would help.

PR #4398, opened 2026-07-27
Two-stage a16w4 MoE GEMM (INTERLEAVE-gate mode)

apicciau:
## Summary

Add a FlyDSL two-stage MoE GEMM path for bf16 activations × MXFP4 weights (`q_dtype_a=torch.bfloat16`, `q_dtype_w=torch.float4_e2m1fn_x2`) in **INTERLEAVE gate mode**, covering all ATOM cudagraph capture sizes on GPT-OSS 120B TP=8.

The key contribution is a `klane_inner` memory layout permutation that enables coalesced `buffer_load_dwordx4` scale loads in both stage1 and stage2, replacing four separate dword loads per wave.

## Changes

**Stage1 (`compile_moe_gemm1` → `moe_gemm_2stage.py`)**
- `klane_inner` permutation: reorders scale tiles so each wave's MXFP4 scale loads are contiguous, enabling `buffer_load_dwordx4`
- Split-K path for tok=1 via `buffer_atomic_pk_add_bf16`
- bf16 CShuffle epilogue (`e_vec=2`)
- INTERLEAVE gate mode: treats the 2×inter_dim weight matrix as interleaved gate+up experts

**Stage2 (`compile_moe_gemm2` → `moe_gemm_2stage.py`)**
- Port of `klane_inner` to the down-projection (two-step: standard preshuffle → KLane↔L_sub swap for `gate_up=False`)
- VMEM count fix: K128-group granularity in the MFMA pipeline to match fp4bf16 scale layout
- Stage2 zero-init fused into stage1 kernel launch to eliminate a separate memset dispatch

**Dispatch table (`gptoss_a16w4_tuned_fmoe.csv`)**
- Jointly tuned FlyDSL INTERLEAVE rows for all 11 ATOM cudagraph capture sizes:
  tok = 1, 2, 4, 8, 16, 32, 48, 64, 128, 256, 512
- Full `interleave_both` sweep on GPT-OSS TP=8 shapes (model_dim=3072, inter_dim=512, E=128, topk=4, gfx950)

## Microbenchmark (gfx950, GPT-OSS TP=8)

FlyDSL INTERLEAVE vs CK-tile (`swiglu_mxfp4_bf16_cktile`):

| Tokens | FlyDSL µs | CK-tile µs | Speedup |
|-------:|----------:|-----------:|--------:|
| 1 | 10.21 | 11.28 | **1.10×** |
| 2 | 10.55 | 11.69 | **1.11×** |
| 4 | 11.40 | 14.86 | **1.30×** |
| 8 | 12.22 | 16.90 | **1.38×** |
| 16 | 20.66 | 23.29 | **1.13×** |
| 32 | 31.77 | 40.36 | **1.27×** |
| 48 | 29.86 | 40.79 | **1.37×** |
| 64 | 29.77 | 41.44 | **1.39×** |
| 128 | 33.15 | 43.47 | **1.31×** |
| 256 | 35.19 | 46.61 | **1.32×** |
| 512 | 39.65 | 54.68 | **1.38×** |

## E2E benchmark — GPT-OSS 120B, MI350X TP=8, kv_cache=fp8

ATOM main (14f4d38f), inductor+cudagraph, 50 samples/datapoint.
Baseline: same branch, `AITER_CONFIG_FMOE` pointing to CK-only CSV.

**Output throughput (tok/s)**

| Scenario | CONC | FlyDSL | CK-tile | Speedup |
|:---------|-----:|-------:|--------:|--------:|
| ISL=1024 OSL=1024 | 4 | 1,191 | 958 | **1.24×** |
| | 16 | 3,798 | 3,571 | **1.06×** |
| | 64 | 9,610 | 9,074 | **1.06×** |
| | 128 | 14,346 | 11,722 | **1.22×** |
| | 256 | 19,318 | 18,650 | **1.04×** |
| ISL=8192 OSL=1024 | 4 | 1,137 | 1,060 | **1.07×** |
| | 16 | 3,581 | 3,377 | **1.06×** |
| | 64 | 7,826 | 7,456 | **1.05×** |
| | 128 | 10,354 | 9,289 | **1.11×** |
| | 256 | 11,010 | 11,135 | ~1.00× |

## Tests

```bash
python -m pytest aiter/ops/flydsl/test_flydsl_moe_a16w4.py -v
python -m pytest op_tests/flydsl_tests/test_flydsl_moe_a16w4.py -v
python3 op_tests/test_moe_2stage.py -d bf16 -q 5 -t 1 8 32
```


aiter/ops/flydsl/kernels/mfma_preshuffle_pipeline.py
Zzz9990: use fx.int32

Zzz9990: don't use arithes, use & >> directly

Zzz9990: same

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -108,6 +139,11 @@ def compile_moe_gemm1(
     use_cshuffle_epilog: bool | None = None,
     scale_is_bf16: bool = False,
     k_batch: int = 1,
+    gate_mode: str = "separated",
+    swiglu_limit: float = 0.0,
+    act: str = "silu",
+    waves_per_eu: int = 0,
+    zero_stage2_output: bool = False,
```
Zzz9990, line 118: why do you need zero output? 
And you should add k_waves feature to the kernel

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -108,6 +139,11 @@ def compile_moe_gemm1(
     use_cshuffle_epilog: bool | None = None,
     scale_is_bf16: bool = False,
     k_batch: int = 1,
+    gate_mode: str = "separated",
+    swiglu_limit: float = 0.0,
```
Zzz9990, line 115: move it to kernel entry and initialize it with fx.float

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -403,6 +482,28 @@ def silu(x):
                 sig = rocdl.rcp(T.f32, den)
                 return x * sig
 
+            def swiglu_apply(gate_scaled, up_scaled):
+                """DSv3-style SwiGLU: g * sigmoid(1.702 * g) * (u + 1) with clamping.
+                Matches the aiter reference swiglu(alpha=1.702, limit=7) used by fp4_bf16 tests.
+                When swiglu_limit==0 (test default), effective limit is 7.0.
+                """
+                _limit_val = arith.constant(_swiglu_eff_limit, type=T.f32)
```
Zzz9990: same

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -441,6 +545,29 @@ def silu(x):
...
+                )
+                zero_stride = (
+                    size_expert_ids_in * fx.Index(_zero_n_tiles) * fx.Index(256)
+                )
+                zero_value = arith.trunc_f(out_mlir(), fx.Float32(0.0))
+                zero_loop = scf.ForOp(zero_start, zero_elems, zero_stride)
+                zero_ip = ir.InsertionPoint(zero_loop.body)
+                zero_ip.__enter__()
+                zero_idx = zero_loop.induction_variable
+                buffer_ops.buffer_store(
```
Zzz9990, line 532: maybe use fx.store

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -527,12 +654,105 @@ def silu(x):
...
+                            + lane_mod_16 * _snl
+                        )
+                        return buffer_ops.buffer_load(
+                            sw_rsrc, idx, vec_width=1, dtype=T.i32
+                        )
+
+                    def _mxfp4_extract_e8m0_f32(packed_i32, byte_pos_idx):
+                        shift = arith.index_cast(T.i32, byte_pos_idx) * fx.Int32(8)
+                        byte_i32 = arith.shrui(packed_i32, shift) & fx.Int32(0xFF)
+                        scale_bits = arith.shli(byte_i32, fx.Int32(23))
```
Zzz9990: same

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -1328,27 +1806,40 @@ def _unflatten_b_tile(vals):
...
+                    )
+                    rocdl.sched_barrier(0)
 
                     _ag, _au, _ = compute_tile(
                         _ag, _au, _bg, _bu, lds_base_pong, a0_prefetch=_a0pf
                     )
                     store_x_tile_to_lds(x_regs_ping, lds_base_ping)
                     hot_loop_scheduler()
-                    gpu.barrier()
+                    _barrier(vmcnt=0, lgkmcnt=0)
```
Zzz9990: _barrier(vmcnt=0, lgkmcnt=0) is equal to gpu.barrier() don't need to replace it

aiter/ops/shuffle.py
```diff
@@ -208,8 +208,41 @@ def shuffle_weight(
...
+def shuffle_weight_a16w4(
+    src: torch.Tensor, NLane: int, gate_up: bool, klane_inner: bool = False
+) -> torch.Tensor:
+    """Backward-compatible wrapper around `shuffle_weight(..., is_guinterleave=True)`.
+
+    When *klane_inner* is True (gate_up only), the KPack dimension is split into
+    two L_sub halves and KLane is placed at dword-contiguous stride so that each
+    lane's 4-byte words are packed together, enabling buffer_load_dwordx4.
+    """
+    if klane_inner:
```
Zzz9990, line 221: why do we need these? I think this layout is equal to line 173?

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
```diff
@@ -3761,6 +3567,15 @@ def run_config(self, args):
 
                 w1_qt_fmoe.is_shuffled = True
                 w2_qt_fmoe.is_shuffled = True
+                _fmoe_gate_mode = (
```
Zzz9990: fp8 supports seperate mode, and does bf16 support separated mode now? I saw branches in moe_gemm_2stages.py

Zzz9990:
One of my previous drafts got merged into main. When resolving conflicts, please pay attention to it and keep your changes in all the conflicting items.

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
apicciau: Not yet, I have SEPARATED gate mode in a follow-up PR. I avoid cramming everything into the same PR and make it too big to review.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Kept `buffer_ops.buffer_store` here — this store targets a buffer resource created with `_ptr_buffer_resource` (SGPR base address), which requires `buffer_store_dword` addressing. `fx.store` / `ptr.store()` emits `flat_store_dword` and cannot operate on a buffer resource handle, so the two are not interchangeable in this context.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: For the `zero_stage2_output`: stage2 uses `buffer_atomic_pk_add_bf16` to accumulate per-expert partial results into a shared output buffer (multiple expert CTAs write to the same token rows). That buffer must be zeroed before stage2 runs. Previously this was done with `torch.zeros` on the host, which serialises device work. The `zero_stage2_output` flag fuses the zeroing loop into the stage1 kernel so it runs on-device without an extra dispatch: each stage1 CTA zeros its own output columns while the rest of the GEMM grid is in flight.

I looked into `k_waves` and it seems a bit of a big change to add. I'm happy to create a follow-up PR.

aiter/ops/shuffle.py
apicciau: No, they're different. At line 173 it's
```
standard is_guinterleave (gate_up=True):

# shape: (experts, 2, N0, NLane, K0, KLane, KPack)
x_.permute(0, 2, 1, 4, 5, 3, 6)
#          E  N0  GU K0 KLane NLane KPack
```
while here it's
```
klane_inner (gate_up=True):

# shape: (experts, 2, N0, NLane, K0, KLane, L_sub, L_sub)
#   where KPack=16 is split into L_sub(4) × L_sub(4)
x_.permute(0, 2, 1, 4, 3, 6, 5, 7)
#          E  N0  GU K0 NLane L_sub KLane L_sub
```
This puts KLane in a position that allows `buffer_load_dwordx4`. To get into this position (in the non-`gate_up` branch) we need to apply the `is_guinterleave` permutation first and then `klane_inner`.

aiter/ops/flydsl/kernels/silu_and_mul_fq.py
```diff
@@ -169,12 +169,11 @@ def _ptr_buffer_resource(ptr):
         scale_rsrc = _ptr_buffer_resource(out_scale_sorted)
         tid_rsrc = _ptr_buffer_resource(sorted_ids)
         nv_rsrc = _ptr_buffer_resource(num_valid_ids)
-        if enable_bias:
-            topk_rsrc = _ptr_buffer_resource(topk_ids)
-            bias_rsrc = _ptr_buffer_resource(bias)
+        topk_rsrc = _ptr_buffer_resource(topk_ids)
+        bias_rsrc = _ptr_buffer_resource(bias)
```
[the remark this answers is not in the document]
apicciau: Addressed in bef0011d9.

aiter/ops/flydsl/kernels/silu_and_mul_fq.py
```diff
@@ -213,13 +212,10 @@ def _load_bias_scalar(offset):
...
-                        # sorted_ids encodes token and slot, not expert. Use topk_ids
-                        # to recover the expert-specific bias row for this token slot.
-                        expert_id = buffer_ops.buffer_load(
-                            topk_rsrc, in_row, vec_width=1, dtype=i32
-                        )
-                        bias_row = expert_id * inter_dim2_i32
+                    expert_id = buffer_ops.buffer_load(
+                        topk_rsrc, in_row, vec_width=1, dtype=i32
+                    )
+                    bias_row = expert_id * inter_dim2_i32
```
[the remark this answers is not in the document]
apicciau: Addressed in bef0011d9.

aiter/ops/flydsl/moe_kernels.py
```diff
@@ -1275,10 +1412,12 @@ def flydsl_moe_stage1(
                 else torch.empty(0, device=dev)
             ),
             swiglu_limit=_swiglu_limit_val,
+            include_stage2_out=(b_dtype != "fp4"),
         )
```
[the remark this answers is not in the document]
apicciau: Addressed in bef0011d9.

op_tests/test_moe_2stage.py
```diff
@@ -712,6 +713,16 @@ def calc_diff(x: torch.Tensor, y: torch.Tensor):
...
+    choices=["interleave", "separated"],
+    default=None,
+    help=(
+        "Override gate mode for a16w4/a8w4 (fp4x2 weight) quant types. "
+        "Default: interleave (matches physical weight layout from shuffle_weight_a16w4). "
+        "Use 'separated' to benchmark the SEPARATED FlyDSL kernel path."
+    ),
+)
 
 args = parser.parse_args()
```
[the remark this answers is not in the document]
apicciau: Addressed in 510fd27ca.

aiter/ops/flydsl/kernels/mfma_preshuffle_pipeline.py
apicciau: Addressed in 923150d10.

apicciau: Addressed in 923150d10.

apicciau: Addressed in 923150d10.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in 4e90116ae.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in 4e90116ae.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in 923150d10.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Same rationale as the reply on line 562 — this store targets a buffer resource (`_ptr_buffer_resource`), which requires `buffer_store_dword` addressing.

aiter/ops/flydsl/kernels/mfma_preshuffle_pipeline.py
```diff
@@ -9,14 +9,12 @@
...
-import flydsl.expr as fx
 from flydsl._mlir import ir
 from flydsl._mlir.dialects.arith import CmpIPredicate
-from flydsl.expr import arith as _arith
 from flydsl.expr.typing import T
+from flydsl.expr import arith as _arith
+import flydsl.expr as fx
 
 
 def crd2idx(crd, layout):
```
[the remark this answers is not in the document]
apicciau: Addressed in 43a868727.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,573 @@
+#!/usr/bin/env python3
```
apicciau, line 1: Addressed in 43a868727.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,573 @@
...
+torch.set_default_device("cuda")
+
+import aiter
+from aiter import dtypes, QuantType, ActivationType
+from aiter.fused_moe import fused_topk, torch_moe_stage1, torch_moe_stage2
+from aiter.ops.shuffle import shuffle_weight_a16w4, shuffle_scale_a16w4
+from aiter.utility.fp4_utils import e8m0_shuffle
+from aiter.test_common import checkAllclose
+
+Q_TYPE = QuantType.per_1x32
```
apicciau, line 42: Addressed in 43a868727.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,573 @@
...
+import torch
+
+torch.set_default_device("cuda")
+
+import aiter
+from aiter import dtypes, QuantType, ActivationType
+from aiter.fused_moe import fused_topk, torch_moe_stage1, torch_moe_stage2
+from aiter.ops.shuffle import shuffle_weight_a16w4, shuffle_scale_a16w4
+from aiter.utility.fp4_utils import e8m0_shuffle
+from aiter.test_common import checkAllclose
```
apicciau: Addressed in 43a868727.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,573 @@
...
+                            atol=args.atol,
+                            rtol=args.rtol,
+                        )
+                        results.append(
+                            (
+                                f"stage2_{mode}_t{token}_bm{bm}",
+                                "PASS" if passed else "FAIL",
+                            )
+                        )
+                    except Exception as e:
```
apicciau: Addressed in 43a868727.

op_tests/flydsl_tests/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,152 @@
...
+import pytest
+
+import aiter
+from aiter import dtypes, QuantType, ActivationType
+from aiter.fused_moe import fused_moe, fused_topk, get_gfx
+from aiter.ops.shuffle import shuffle_weight_a16w4, shuffle_scale_a16w4
+from aiter.test_common import checkAllclose
+
+
+def _setup(model_dim, inter_dim, E, topk, tokens, gate_mode, seed=42):
```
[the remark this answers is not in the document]
apicciau: Addressed in 43a868727.

aiter/ops/flydsl/kernels/mfma_preshuffle_pipeline.py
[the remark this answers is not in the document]
apicciau: Addressed in 43a868727.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,574 @@
...
+import torch
+
+torch.set_default_device("cuda")
+
+import aiter  # noqa: E402
+from aiter import ActivationType, QuantType, dtypes  # noqa: E402
+from aiter.fused_moe import fused_topk, torch_moe_stage1, torch_moe_stage2  # noqa: E402
+from aiter.ops.shuffle import shuffle_weight_a16w4, shuffle_scale_a16w4  # noqa: E402
+
+Q_TYPE = QuantType.per_1x32
```
apicciau: Addressed in df26c1e13.

aiter/ops/flydsl/test_flydsl_moe_a16w4.py
```diff
@@ -0,0 +1,574 @@
...
+os.environ.setdefault("HIP_VISIBLE_DEVICES", "0")
+os.environ.setdefault("AITER_USE_SYSTEM_TRITON", "1")
+os.environ.setdefault("FLYDSL_RUNTIME_ENABLE_CACHE", "0")
+
+import torch
+
+torch.set_default_device("cuda")
+
+import aiter  # noqa: E402
+from aiter import ActivationType, QuantType, dtypes  # noqa: E402
```
apicciau: Addressed in df26c1e13.

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
Zzz9990: We plan to remove the SEPARATED branches, this judgment here might be removed.

aiter/ops/shuffle.py
Zzz9990: we always set KPack ==16. So KPack == 16 == L_sub * L_sub... Maybe you can add an assert ?

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -45,23 +45,54 @@ def bf16_global_atomics_arch_description() -> str:
...
+    make_preshuffle_scale_layout,
     swizzle_xor16,
     tile_chunk_coord_i32,
+    unpack_b_mxfp4_bf16,
     unpack_b_w4a16,
 )
 from .tensor_shim import _run_compiled
 
 
+def _barrier(vmcnt=63, lgkmcnt=63):
```
Zzz9990: remove it? No longer using it.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
Zzz9990: I mean move it to line419 @flyc.kernel

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -292,22 +347,37 @@ def out_mlir():
...
-            "stage1 cshuffle epilog currently supports only f16 output (out_dtype='f16')"
+            "stage1 cshuffle epilog currently supports only f16/bf16 output"
         )
 
+    # INTERLEAVE gate mode: gate and up weights are packed at alternating N positions.
+    # Only supported for fp4_bf16 (MXFP4) where the physical weight layout interleaves
+    # gate/up columns so a single B-tile covers both. For other dtypes, gate and up are
+    # loaded from separate row ranges (SEPARATED).
+    gate_up_interleave = is_fp4_bf16 and gate_mode == "interleave"
+    _use_klane_inner = is_fp4_bf16 and gate_up_interleave
```
Zzz9990: We plan to remove separate mode, the conditions variables here could be removed.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
Zzz9990: not  resources. just remove arith here and use new flydsl-style codes. The 'arith'es at the back is the same, please remove them.

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
```diff
@@ -1325,10 +1817,17 @@ def _unflatten_b_tile(vals):
...
+                    # A-load → B-loads → compute → A→LDS store → barrier (CK-tile order).
                     next_k1 = k_iv + c_tile_k
-                    x_regs_ping = load_x_tile(next_k1)
+                    x_regs_ping = load_x_tile(next_k1)  # A in-flight
+                    rocdl.sched_barrier(0)
                     _bg_ping = load_b_tile(next_k1, n_blk_gate, n_intra_gate)
-                    _bu_ping = load_b_tile(next_k1, n_blk_up, n_intra_up)
+                    _bu_ping = (
+                        _bg_ping
+                        if gate_up_interleave
```
Zzz9990: just use interleave branches

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in [`5ee25d315`](https://github.com/ROCm/aiter/commit/5ee25d315)

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in [`5ee25d315`](https://github.com/ROCm/aiter/commit/5ee25d315)

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in [`5ee25d315`](https://github.com/ROCm/aiter/commit/5ee25d315)

aiter/ops/shuffle.py
apicciau: Addressed in [`5ee25d315`](https://github.com/ROCm/aiter/commit/5ee25d315)

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in [`d9d718756`](https://github.com/ROCm/aiter/commit/df26c1e13bc51dd3d8ac377f3ff7356c49b9c815)

aiter/ops/flydsl/kernels/moe_gemm_2stage.py
apicciau: Addressed in [`d9d718756`](https://github.com/ROCm/aiter/commit/df26c1e13bc51dd3d8ac377f3ff7356c49b9c815)

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
apicciau: Addressed in [`d9d718756`](https://github.com/ROCm/aiter/commit/d9d718756): simplified `_fmoe_gate_mode` to always use INTERLEAVE.

yichiche:
Thanks for this — we've been bringing it up on Qwen3.5-397B-A17B-MXFP4 (TP2) and hit persistent garbage output. I tracked it down to **two independent correctness bugs in the INTERLEAVE path**, plus a compile-time break in the swiglu path. Sharing the analysis and repro here rather than opening a competing PR.

The short version of why this looks green on your side: the test harness has three blind spots that make both bugs invisible, and the shapes the tests exercise are not the shapes `fused_moe` actually dispatches.

---

## Why the existing tests can't see it

**1. The test hardcodes `tile_n=128`, but the tuned configs select `tile_n=64`.**

`test_flydsl_moe_a16w4.py` calls stage1/stage2 with `tile_n=128`. Meanwhile `gptoss_a16w4_tuned_fmoe.csv` in this PR selects:

```
  3  flydsl_moe1_abf16_wfp4bf16_bf16_t16x64x128_gui
  8  flydsl_moe1_abf16_wfp4bf16_bf16_t16x64x256_gui
```

11 of the rows use `tile_n=64` — a shape the tests never compile. (Our Qwen config was 64/64 rows on `tile_n=64`, i.e. 100% of dispatches.)

**2. `_check_result` gates on `atol=1.0`, but real a16w4 outputs are ~1e-3.**

Real stage2 output magnitude on a Qwen MXFP4 checkpoint is ~2.7e-6, and stage1 ~1.4e-5. With `atol=1.0`, *anything* passes. We measured stage2 reporting **"PASS, 99.9% close"** while its true cosine vs the torch reference was **0.22**. Adding a scale-free metric is what surfaced Bug 2.

**3. `randn/10` weights give ~4 distinct E8M0 exponents; a real checkpoint has 55.**

Measured directly:

| weights | E8M0 exponent range | distinct values |
|---|---|---|
| `torch.randn(...)/10` (test) | 121–124 | **4** |
| Qwen3.5-397B-A17B-MXFP4 | 1–122 | **55** |

Every 32-element K block in `randn/10` has almost the same amax, so scale-*addressing* errors read a neighbour with an identical exponent and vanish. On a real checkpoint they are catastrophic.

---

## Bug 1 — INTERLEAVE with `tile_n=64` never accumulates the up projection

`moe_gemm_2stage.py` splits the per-wave N accumulators by parity of `ni`:

```python
if const_expr(gate_up_interleave):
    # Even ni = gate block, odd ni = up block
    if ni % 2 == 0:
        gate_list[acc_idx] = mfma_k64(gate_list[acc_idx], a0, a1, bg0, bg1)
    else:
        up_list[acc_idx]   = mfma_k64(up_list[acc_idx],   a0, a1, bg0, bg1)
```

and

```python
num_waves = 4
n_per_wave = tile_n // num_waves
num_acc_n  = n_per_wave // 16        # == tile_n // 64
```

With `tile_n=64` → `num_acc_n == 1` → `ni` is only ever `0` → **`up_list` is never written**. INTERLEAVE structurally requires at least one full gate/up pair per wave, i.e. `num_acc_n >= 2`, i.e. `tile_n >= 128`.

Measured (stage1, E=64, topk=10, 32 tokens, 4096×512, exponent-spread weights):

| tile | gate_mode | cosine vs torch |
|---|---|---|
| 16x64x256 | separated | 1.000000 |
| **16x64x256** | **interleave** | **0.000340** |
| 16x128x256 | interleave | 1.000000 |

On real Qwen checkpoint weights, `16x64x256` + interleave gives cosine **−0.001067** with output magnitude 8.06e-02 against a reference of 1.44e-05.

## Bug 2 — the `klane_inner` w2 layout for stage2 is wrong

`fused_moe.py` sets `extra_stage2_args["klane_inner"] = True` for INTERLEAVE + `fp4bf16`, and the sglang side shuffles w2 to match. Both pairings are broken:

| w2 shuffle | stage2 kernel | cosine |
|---|---|---|
| `klane_inner=False` | `False` | **0.999989** ✅ |
| `klane_inner=True` | `False` | 0.225 |
| `klane_inner=True` | `True` | **0.049** ← what serving actually runs |

Worth noting: `test_flydsl_stage2_a16w4` shuffles w2 with `klane_inner=True` (it inherits `gate_mode="interleave"`) but calls `flydsl_moe_stage2(...)` **without** `klane_inner`, so the test sits in the middle row — it never exercises the combination `fused_moe` uses, and `atol=1.0` hides that it is wrong anyway.

This bug is fully independent of Bug 1: fixing `tile_n` alone still yields garbage.

## Bug 3 — swiglu stage1 fails to compile (blocks gptoss entirely)

```python
_swiglu_limit = fx.Float32(_swiglu_eff_limit)
...
g = arith.minimumf(gate_scaled, _swiglu_limit)
```

`arith.minimumf` requires an MLIR `Value`; `fx.Float32` only works through the Python operator overloads. This raises:

```
Operand 1 of operation "arith.minimumf" must be a Value (is not a Value)
```

in **both** gate modes. All 27 rows of `gptoss_a16w4_tuned_fmoe.csv` are `ActivationType.Swiglu`, so the gptoss a16w4 path in this PR cannot compile at all. Other flydsl kernels use `arith.constant(v, type=T.f32)` (cf. `kernels/swiglu_and_mul.py`).

## Bug 4 — `sw_nbytes` is 4× oversized (latent)

```python
sw_nbytes = experts * (2*inter_dim) * (num_groups * _fp4_bf16_k_groups) * (2 if _scale_is_bf16 else 4)
```

The MXFP4 E8M0 scale is exactly **one byte** per 32-element K group. It is *read* as i32 (4 scale bytes per word), but that is the access width, not the allocation size. For E=512, inter=512, K=4096 this computes 268,435,456 B against a real tensor of 67,108,864 B. The buffer resource then permits 4× the valid range, so out-of-range scale indices return neighbouring garbage instead of being clamped to 0 — which is part of why the bugs above produced plausible-looking output rather than obvious zeros.

---

## Reproducing

Bug 1, using only what's in this PR — compile stage1 with the tile the tuned config actually selects:

```python
# test_flydsl_stage1_a16w4(...): tile_n=128 -> 64, gate_mode="interleave"
out = flydsl_moe_stage1(..., tile_n=64, tile_k=256, gate_mode="interleave", ...)
```

To make it visible you also need non-flat weights — give each 32-element K block a random power-of-two magnitude (FP4 nibbles are unchanged, so the torch reference stays exact):

```python
nblk = model_dim // 32
expo = torch.randint(-8, 9, (E, inter_dim*2, nblk), device=w1.device)
mult = torch.pow(torch.tensor(2.0, device=w1.device), expo.float())
w1 = (w1.view(E, inter_dim*2, nblk, 32) * mult.unsqueeze(-1)).view(E, inter_dim*2, model_dim)
```

Bug 2: add a scale-free metric to `_check_result` and run the existing stage2 test with `gate_mode="interleave"`:

```python
cos = (r.flatten() @ t.flatten() / (r.norm()*t.norm()).clamp_min(1e-12)).item()
rel_rms = ((r - t).norm() / r.norm().clamp_min(1e-12)).item()
```

It reports `PASS / 99.9% close` with `cosine=0.2248`.

Bug 3: `--activation swiglu` on the stage1 test.

Happy to share the two standalone probe scripts we used (synthetic exponent-spread sweep, and one that loads real Qwen MXFP4 expert weights straight out of the safetensors shards) if useful.

---

## Fixes we applied

1. **Compile-time guard** in `compile_moe_gemm1`: INTERLEAVE requires `num_acc_n` even and ≥2, i.e. `tile_n % 128 == 0`. Also excluded `tile_n=64` `_gui` kernels from `get_flydsl_stage1_kernels_fp4_bf16` so the tuner cannot select them, and bumped the 11 gptoss config rows to `tile_n=128`. Both inter_dims divide evenly by 128.
2. **Dropped `klane_inner` for stage2.** INTERLEAVE is a stage1 concept (gate/up rows interleaved in w13); w2 has no gate/up, so `klane_inner` there is only a dwordx4 load-width optimisation. Requires a matching sglang change (w2 / w2_scale prep must also use `klane_inner=False`) — the two must land together.
3. `fx.Float32(...)` → `arith.constant(..., type=T.f32)` for the four swiglu constants.
4. `sw_nbytes` corrected to one byte per K group. Verified it clamps nothing valid (results stay bit-identical).

Two harness issues also worth folding in: `_check_result` should report cosine/rel_rms, and in `_generate_a16w4_data` the `sorted_weights_s1/s2` views are bound *before* the `sorted_ids` padding block, so `sorted_weights` stays unpadded while `sorted_ids` grows (1334 vs 1344 needed in our config).

## Validation after the fixes

Real Qwen3.5-397B-A17B-MXFP4 checkpoint weights, layer 0, TP2 shard, 32 tokens, cosine vs torch reference:

| stage | separated | interleave |
|---|---|---|
| stage1 | 1.000000 | **1.000000** |
| stage2 | 0.999990 | **0.999990** |
| e2e | 0.999990 | **0.999990** |

Counterfactual with the old w2 layout: cosine **0.0999**.

No regressions: `test_flydsl_moe_a4w4` 4/4, `test_flydsl_moe_a8w4` + `test_flydsl_moe_a16w4` 33 passed, and the a16w4 suite at E=512 / topk=10 / tokens 1–128 is 12/12 with the scale-free metrics.

Unrelated, but noticed while sweeping: `test_flydsl_a16wfp4_situv2_fused_moe` fails on both parametrisations (`rel_l2=0.6049` and `0.2276`) on the SiTUv2 mixed_moe path. I reverted all of the above changes and got bit-identical numbers, so it predates this work — flagging it separately.

Glad to push these as commits to `apicciau/a16w4_interleave` if that's easier than reapplying.

apicciau:
@copilot resolve the merge conflicts in this pull request

PR #4401, opened 2026-07-27
[NO MERGE] Add gfx942 FWD split kv

ipanfilo:
## Submission Checklist

### DIFF
--- a/aiter/ops/attention.py
+++ b/aiter/ops/attention.py
@@ -14,7 +14,7 @@ from csrc.cpp_itfs.pa.pa_ragged import (
 )
 from csrc.cpp_itfs.pa.pa_v1 import paged_attention_v1 as paged_attention_v1_core
 from csrc.cpp_itfs.torch_utils import direct_register_custom_op
-from aiter.ops.triton.gluon.pa_decode_gluon import pa_decode_gluon
+#PIV from aiter.ops.triton.gluon.pa_decode_gluon import pa_decode_gluon
 
 from aiter import dtypes
 
@@ -23,11 +23,11 @@ from ..jit.core import compile_ops, is_experimental_enabled
 
 MD_NAME = "module_attention"
 
-direct_register_custom_op(
+"""PIV direct_register_custom_op(
     "pa_decode_gluon",
     pa_decode_gluon,
     ["output", "exp_sums", "max_logits", "temporary_output"],
-)
+)"""
 
 
 def gen_pa_fwd_native_fake(
--- a/aiter/ops/attention.py
+++ b/aiter/ops/attention.py
@@ -14,7 +14,7 @@ from csrc.cpp_itfs.pa.pa_ragged import (
 )
 from csrc.cpp_itfs.pa.pa_v1 import paged_attention_v1 as paged_attention_v1_core
 from csrc.cpp_itfs.torch_utils import direct_register_custom_op
-#PIV from aiter.ops.triton.gluon.pa_decode_gluon import pa_decode_gluon
+from aiter.ops.triton.gluon.pa_decode_gluon import pa_decode_gluon
 
 from aiter import dtypes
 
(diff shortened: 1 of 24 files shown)


csrc/cpp_itfs/mha_fwd.cu
```diff
@@ -365,10 +368,179 @@ float fmha_fwd_ck(mha_fwd_args a, const ck_tile::stream_config& s)
...
+    // Use hipMallocAsync / hipFreeAsync (stream-ordered pool allocator) so the alloc/free
+    // are enqueued on the stream and don't force a device-wide synchronization.
+    const size_t scratch_o_elems   = static_cast<size_t>(G) * B * Hq * Sq * kHeadDim;
+    const size_t scratch_lse_elems = static_cast<size_t>(G) * B * Hq * Sq;
+    float* scratch_o   = nullptr;
+    float* scratch_lse = nullptr;
+
+    const hipStream_t stream = s.stream_id_;
+
+    HIP_CALL(hipMallocAsync(&scratch_o,   scratch_o_elems   * sizeof(float), stream));
```
wangye805: Emm, during our integration into TE, this hipMallocAsync violates our current convention that workspace must be allocated from upstream framework (jax/pytorch) and passed into the libte.so

op_tests/test_mha_fwd_native_gfx942.py
wangye805, line 1: Shall we also have a pr later to aiter main branch since you also added the aiter pytorch side integration

wangye805 (changes requested):

Micky774:
Were the `cpp_itfs/kernels/mha_native/fused/op_*` kernels adapted from somewhere else, or are they original to the PR?

aiter/ops/attention.py
Micky774, line 1: Are these changes necessary/relevant?

csrc/kernels/mha_native/fused/op_combine.hpp
```diff
@@ -0,0 +1,225 @@
...
+    // B is not a struct field; recover it from the grid's batch axis. The g-stride
+    // of the split-major scratch is (B*Hq*Sq) elements for LSE / *64 for O.
+    const int B  = gridDim.z;
+
+    // Split-major scratch indices (match the split-forward pass EXACTLY):
+    //   scratch_o  (g,b,h,R,d) = ((((g*B + b)*Hq + h)*Sq + R)*64 + d
+    //   scratch_lse(g,b,h,R)   =  (((g*B + b)*Hq + h)*Sq + R
+    // The per-g stride lets us step plane to plane by adding a constant; we keep
+    // the explicit form for clarity (this pass is not perf-critical).
+    const long bh_row    = (long)((b * Hq + h) * Sq + R);   // (b,h,R) within a plane
```
Micky774: Since we're casting at the end I think we're risking an overflow here -- let's cast before the multiply.

csrc/cpp_itfs/mha_fwd.cu
```diff
@@ -365,10 +368,179 @@ float fmha_fwd_ck(mha_fwd_args a, const ck_tile::stream_config& s)
...
+    HIP_CALL(hipMallocAsync(&scratch_lse, scratch_lse_elems * sizeof(float), stream));
```
Micky774: As Ye mentioned, we'll probably need to carve this out from the workspace instead.

Micky774 (changes requested):

csrc/cpp_itfs/mha_fwd.cu
ipanfilo: Well, to make it properly we probably have to pass more API through QoLA - to ask if split-kv kernels will be used. Unless you're OK to always allocate extra. @Micky774, how complicated is it?

aiter/ops/attention.py
ipanfilo: > Are these changes necessary/relevant?

No, those are w/a for AITER install on my setup.

csrc/kernels/mha_native/fused/op_combine.hpp
ipanfilo: I fixed it but note that it is from another PR that has been already in the main so will need to merge when sync with newer AITER

zufayu:
Closing in favor of #4466, which is the rebased (_up1) version of this same work. Both were marked [NO MERGE] as temporary branches — keeping the newer one only.

PR #4413, opened 2026-07-28
[Bugfix] mp_tuner: default --timeout so a dead worker is reaped instead of hanging

vanshbhatia-amd:
## Problem

`mp_tuner()` hangs indefinitely (with zero output) when a candidate kernel
kills its worker process. On MI355X (gfx950) this is easy to hit: some bf16
A16W16 GEMM candidates trigger a hard GPU memory-access fault during tuning,
which kills the pool worker. Instead of skipping that candidate/shape, the
whole tuning run deadlocks.

## Root cause

When a worker dies, `multiprocessing.Pool` respawns it with a new PID, but the
task that worker was running is lost, so its `apply_async().get()` never
resolves.

`mp_tuner` already has a recovery path for exactly this: the per-task timeout
watchdog reaps a task whose worker never returns and restarts the pool. **But
that path is gated behind `if timeout is not None`, and the default timeout is
`None`.** So out of the box the watchdog never runs and the run hangs forever.

## Fix

Give `--timeout` a sane, generous default (`1800`s) in `base_tuner.py` so the
existing reaping path is active by default. A lost task (dead worker) is reaped
after the timeout, the pool is restarted, and the remaining tasks complete
normally. The default is set comfortably above any legitimate shape-group
tuning time so healthy tasks are never falsely reaped; `--timeout` still
overrides it for tighter/looser bounds.

No new scheduling machinery is added — this relies entirely on the recovery
path that already exists and is already plumbed through every tuner entry
point.

## Reproduction (no special hardware required)

A task that calls `os._exit(1)` simulates the worker death a GPU fault causes.
The fix is a one-line default change; to exercise the recovery quickly without
waiting the full default, pass an explicit modest `timeout` (the new default
simply makes this behaviour the out-of-the-box one):

```python
import os, time, torch
from aiter.utility.mp_tuner import mp_tuner

def ok_func(*a, **k):
    x = torch.randn(512, 512, device=torch.cuda.current_device(), dtype=torch.bfloat16)
    y = x @ x; torch.cuda.synchronize(); time.sleep(0.12)  # ~13s over mp_tuner's iters
    return y

def crash_func(*a, **k):
    os._exit(1)  # simulates a GPU-fault worker death

def task(tag, func):
    return (tag, None, (), func, (), {}, None, (), {}, None)

if __name__ == "__main__":
    n = 3
    tasks = [task((f"s{i}", i, 0, ""), crash_func if i == 1 else ok_func) for i in range(n)]
    print(mp_tuner(tasks, [(1, None)] * n, mp_num=8, fast_mode=True,
                   shape_grouped=False, timeout=40))
```

| Version | Result | Exit |
|---|---|---|
| before this PR (default `timeout=None`) | hangs, killed externally | `124` |
| with this PR (timeout active, now the default) | crashing task reaped, pool restarts, healthy tasks return real results | `0` |

Observed with the watchdog active:

```
Tuning Summary: Total 3 | Successful 2 | Failed 1 (1 timeout)
[('s0', 12.602, 0.0), ('s1', inf, 1.0), ('s2', 12.6836, 0.0)]   # returned after 46.8s
```

The two healthy tasks return real tuning numbers; only the crashed task is
dropped as `inf`; the process exits `0`.


aiter/utility/mp_tuner.py
yzhou103, line 578: As descripted here, the mapping error is caused by other GPU memory access fault, which means this task may be good, so it should not be added to completed_this_round and not restart.

yzhou103:
Thanks for digging into this — the hang is real and I reproduced your baseline. Two things though: (a) this is already recoverable today with the existing `--timeout` flag, and (b) as written, the patch converts a visible hang into a **silent wrong-tuning-result**, which for a tuner is a worse failure mode. Evidence below.

Environment: MI355X (gfx950), 8 GPUs, ROCm 7.2.4, torch 2.10, Python 3.12, `aiter` main checkout with the two `mp_tuner.py` versions loaded as standalone modules (repo otherwise untouched). Worker death simulated with `os._exit(1)`, as in your description.

Two scenarios are used throughout:

- **A**: 24 tasks, `mp_num=8`, task 3 kills its worker, healthy tasks are short (a few seconds).
- **B**: 3 tasks, `mp_num=8`, task 1 kills its worker, healthy tasks take ~40s each — i.e. longer than one polling round (a round is roughly `10s × len(remaining_tasks)`).

| scenario | main, `timeout=None` | main, `--timeout 60` | this PR, `timeout=None` |
| --- | --- | --- | --- |
| A | hangs, killed at 240s, 0 rows | 69.0s, 23/24 valid, only task 3 `inf` | 68.5s, 23/24 valid |
| B | hangs, killed at 240s, 0 rows | 69.0s, both healthy tasks valid, only task 1 `inf` | 127.7s, **all 3 rows `inf`** |

## The hang is already recoverable today via `--timeout`

With a per-task timeout set, `mp_tuner` takes the existing reaping path: the lost task is timed out, a dummy result is recorded, the task is **removed** from `remaining_tasks` (so it is not retried forever), the pool is restarted and `gpu_map` is rebuilt, and the remaining tasks finish. That is exactly what the table above shows for both scenarios — including scenario B, where `--timeout` is the only configuration that returns correct results.

The flag is already plumbed through from every tuner entry point, so no caller changes are needed — including the A16W16 tuner used in your report:

```python
# csrc/gemm_a16w16/gemm_a16w16_tune.py
ret = mp_tuner(
    task, tasks_data, args.mp, False, args.shape_grouped, args.errRatio,
    timeout=args.timeout, verbose=args.verbose,
)
```

Same for `csrc/ck_gemm_a8w8`, `csrc/ck_batched_gemm_a8w8`, `csrc/opus_gemm`, `csrc/ck_gemm_moe_2stages_codegen` and `gradlib`.

So the genuine gap is narrow, and it is the one you correctly identified: the default is `None` (`aiter/utility/base_tuner.py`), so out of the box a dead worker still hangs the run. I'd argue the minimal fix is to give `--timeout` a sane, generous default (comfortably above the slowest shape group) rather than adding a restart + attempt-cap mechanism. Two caveats worth stating for whichever direction is taken: a too-small timeout silently turns healthy long tasks into `inf` (the same class of harm as issue 2 below, though at least it prints `[!] Task N timed out`); and the mapping-error branch still neither records nor removes the task, so it spins ~1/s — but that is bounded, because the lost task's timeout eventually fires, rebuilds `gpu_map` and resubmits the spinning task. `[Mapping Error]` never even appeared in my `--timeout` runs.

## 1. `_resolve_gpu_id()` silently corrupts the measurements it is supposed to protect (blocking)

`gpu_map` encodes *exclusive* device ownership — one worker per GPU is what makes the measured `us` meaningful. `pid % device_count` can (and does) hand a respawned worker a device that another live worker already owns, and when `mp_num < gpu_num` it can also select a device outside the set the tuner was given.

Repro: `mp_num=2` (so only cuda:0 and cuda:1 are owned), 6 tasks, task 0 kills its worker. The respawned worker `pid=366793` gets `366793 % 8 == 1` and runs concurrently on cuda:1 with `pid=366609`, the mapped owner of that device. Identical 4096³ bf16 matmul in every task:

| task | measured under | reported us |
| --- | --- | --- |
| t1 / t2 / t3 | exclusive device | 3946.7 / 3947.2 / 3947.7 |
| t4 | sharing cuda:1 | 5416.3 (+37%) |
| t5 | sharing cuda:1 | 5653.5 (+43%) |

Those inflated numbers are returned as valid results and feed best-config selection, with nothing marking them as suspect. Note this fallback already existed in commented-out form since #1680, with an explicit note — `# Still raise KeyError to trigger pool restart in parent process` — precisely because guessing a device is not safe; the mapping error was deliberately kept as the signal that forces a rebuild.

A worker that cannot claim an exclusive device should refuse to measure and report "not measured" so the parent can requeue the task, rather than benchmark on a device it does not own.

## 2. The attempt cap counts restarts, not per-task failures, so healthy tasks get dropped (blocking)

`task_attempts` is incremented for **every remaining task** on **every** pool restart, and a restart `terminate()`s all in-flight tasks and resubmits them. So whenever a single task's runtime exceeds one polling round, the crashing task kills a worker every round → restart every round → all counters advance together → after 3 rounds everything is dropped as "persistent crash".

That is scenario B in the table. Repro, verified verbatim on the PR head:

```python
import os, time, torch
from aiter.utility.mp_tuner import mp_tuner

def ok(*a, **k):
    d = torch.cuda.current_device()
    x = torch.randn(512, 512, device=d, dtype=torch.bfloat16)
    y = x @ x
    torch.cuda.synchronize()
    # run_perftest() calls this 103x, so one task takes ~40s -- longer than one
    # polling round (~10s x remaining tasks), which is the condition for the bug
    time.sleep(0.4)
    return y

def crash(*a, **k):
    os._exit(1)  # simulates the worker death a GPU fault causes

def task(tag, f):
    return (tag, None, (), f, (), {}, None, (), {}, None)

if __name__ == "__main__":
    n = 3
    tasks = [task((f"s{i}", i, 0, ""), crash if i == 1 else ok) for i in range(n)]
    print(mp_tuner(tasks, [(1, None)] * n, mp_num=8, fast_mode=True,
                   shape_grouped=False, timeout=None))
```

`s0` and `s2` never crash, yet the run returns after 127.7s with all three rows `inf` and exit code 0:

```
[Persistent Crash] Task 0 dropped after 3 attempts (repeated worker death)
[Persistent Crash] Task 1 dropped after 3 attempts (repeated worker death)
[Persistent Crash] Task 2 dropped after 3 attempts (repeated worker death)
Tuning Summary:  Successful: 0  Failed: 3   - Crashes (memory fault): 3
```

This is easy to hit in production: with `shape_grouped=True` a task is all candidates of one shape and routinely takes minutes, while the polling round is short. The run then "succeeds" while producing an all-`inf` table — worse than the hang it replaces, because nothing signals that the output is garbage.

The counter should only advance for a task whose own worker died, and a restart should not discard in-flight healthy work. Tasks dropped after exhausting retries should also be their own category in the summary rather than being counted as "Crashes (memory fault)".

## 3. The two halves of the fix cancel each other out

The second `_resolve_gpu_id()` call site (the former `gpu_id = GPUIDMap[pid]` inside `try`) is a no-op change: a `KeyError` there is swallowed by the `except Exception` at the end of `work_group`, which returns dummy results. Only the lookup *before* the `try` can propagate to the parent — and since the fallback now never raises, the new parent-side mapping-error recovery branch is no longer reachable through that path. Across all my PR-side runs `[Mapping Error]` never appears, while main prints it. I'd keep one of the two, not both.

If the mapping-error branch is kept, it should **requeue** task `k` rather than record `inf` for it. In the main baseline the crash is in task 3 but the `KeyError` is reported for task 8 — the victim is an innocent task that was dispatched to the respawned worker and had not run any work yet, so failing it permanently discards a valid candidate.

aiter/utility/mp_tuner.py
```diff
@@ -18,6 +18,25 @@ def _is_accelerator_error(exc: BaseException) -> bool:
     return type(exc).__name__ == "AcceleratorError"
 
 
+def _resolve_gpu_id(gpu_map, pid):
```
yzhou103: _resolve_gpu_id() silently corrupts the measurements it is supposed to protect (blocking).
I remember I've tried this solution, but it had other problems.
gpu_map encodes exclusive device ownership — one worker per GPU is what makes the measured us meaningful. pid % device_count can (and does) hand a respawned worker a device that another live worker already owns, and can also select a device outside the range the tuner was given (with mp_num=2 on an 8-GPU node it can pick cuda:2–7, which on a shared node belongs to another job).

vanshbhatia-amd:
Thanks for the very thorough reproduction and analysis — you're right on all counts, and I've reworked the PR to the minimal fix you suggested.

You correctly identified that the recovery path already exists (per-task timeout reaping + pool restart) and is already plumbed through every tuner entry point; the only real gap was that the default timeout is `None`, so it never runs. I've dropped the extra machinery:

- **Removed `_resolve_gpu_id()`.** As you noted, the `pid % device_count` fallback can place a respawned worker on a device another worker already owns, silently corrupting measurements. Both call sites are back to `GPUIDMap[pid]`.
- **Removed the per-task attempt cap.** You're right that it counted per pool-restart rather than per-task, so a burst of restarts could drop healthy long-running tasks as a "persistent crash."
- **Reverted the mapping-error branch to `main`.** As you observed, it can spin ~1/s but is bounded — the lost task's timeout fires, rebuilds `gpu_map`, and resubmits it; `[Mapping Error]` didn't show up in my `--timeout` runs either.

The only change now is a generous default `--timeout` (`1800`s) in `base_tuner.py`, comfortably above any legitimate shape-group time so healthy tasks are never falsely reaped, and still overridable via `--timeout`. Happy to tune the number — if the slowest shape group in practice is well under a few minutes I can bring it down so genuine hangs recover faster.

Re-tested your scenario B (3 tasks, `mp_num=8`, task 1 kills its worker, healthy tasks span multiple polling rounds):

```
Tuning Summary: Total 3 | Successful 2 | Failed 1 (1 timeout)
[('s0', 12.602, 0.0), ('s1', inf, 1.0), ('s2', 12.6836, 0.0)]   # exit 0, 46.8s
```

Both healthy tasks return real numbers and only the crashed task is dropped — no all-`inf`, no hang.

The PR description is updated to match. Could you take another look when you get a chance? Thanks again for catching the measurement-corruption risk.

yzhou103 (approved):
LGTM

yzhou103 (approved):

PR #4415, opened 2026-07-28
feat(topk): length-adaptive deterministic top-k for sparse-MLA indexer

chuanbowang2026:
## Motivation

Models using DSA (DeepSeek Sparse Attention), such as GLM-5.2, run MLA
attention with tensor parallelism (TP): each rank handles a subset of KV heads
and the results are merged with an all-reduce. The DSA indexer, however, is
replicated: every TP rank independently computes the full index score and runs
its own Top-K to pick the K (typically 2048) KV entries used for attention.

This means every rank must produce an **identical Top-K result for identical
input — not only the same set, but the same order**. Otherwise different ranks
lay out different KV per local head, and the all-reduce sums up mismatched
scores, corrupting the result.

The original radix-select Top-K collects final results with `atomicAdd` to
claim output positions. The resulting order depends on thread scheduling and is
not reproducible, so different ranks can end up with different orderings. This
PR adds a deterministic ("stable") mode that guarantees an ascending-index
ordered Top-K, identical across ranks, for both prefill and decode.

## Technical Details

The stable mode is opt-in, gated process-wide by the `TOPK_STABLE=1`
environment variable (read once and cached in C++). When unset, the code takes
the original baseline path with unchanged behavior and performance. **The
Python/C++ public API is unchanged** — no new arguments.

When enabled, prefill and decode share a single one-block path that emits an
ascending-index-ordered Top-K, dispatched by row length:

- **Long rows (`row_len >= 32768`) — fast path:** one vectorized pass collects
  the selected elements with `atomicAdd`, then a single `BlockRadixSort` over
  the (at most `k`) row-local indices yields ascending index order. The sort is
  a fixed cost independent of row length.
- **Short rows, or `k > 2048` — scan path:** per-tile `BlockScan` prefix sums
  compute the ascending output position directly, with no sort. Cost grows with
  row length, which is why it is only used for shorter rows.

Ties with the k-th value are broken toward the smaller index in both paths, so
the output is identical across ranks regardless of thread scheduling.

Relative to the earlier approach in PR #4376, this design:

- guarantees full ascending-index order for **all** of k=512/1024/2048 (PR
  #4376 only fully orders when `k==2048`);
- outputs ascending **index** order (PR #4376 outputs score-descending order);
- uses a single stable path for both prefill and decode, with no row-count
  limit;
- can be turned off via the environment variable (PR #4376 is always on).

## Test Plan

Correctness is verified on MI210 (gfx90a) with `randn` float32 inputs (matching
real indexer scores, essentially no duplicates), plus quantized inputs to force
heavy ties. For each case the test runs twice and checks: (1) determinism
across runs, (2) ascending index order, and (3) a bit-faithful reference
matching the smallest-index tie-break contract. Coverage:

- prefill and decode;
- k = 512 / 1024 / 2048, and the k > 2048 scan fallback;
- short rows (scan path) and long rows >= 32768 (fast path);
- `row_len < k` edge cases;
- default (non-stable) path parity with `torch.topk` (set equality), confirming
  no regression when `TOPK_STABLE` is unset.

Single-op latency was benchmarked on MI210 for the production shape
(N ≈ 60K, k = 2048), comparing baseline, PR #4376, and this PR. To compare the
same kernel path, the one-block path is pinned for all modes.

## Test Result

All correctness checks pass (determinism, ascending order, reference match) for
every configuration above, and the default path matches `torch.topk`.

Latency (MI210, k=2048, one-block, µs):

**Long rows / fast path (N = 60K / 61440):**

| Case | baseline | PR #4376 | This PR |
| --- | --- | --- | --- |
| prefill 1 row | 42.1 | 90.8 | 58.5 |
| prefill 2 rows | 42.0 | 90.7 | 58.3 |
| prefill 4 rows | 42.1 | 90.6 | 58.5 |
| prefill 8 rows | 42.4 | 92.1 | 58.8 |
| decode 1 | 42.3 | 74.5 | 58.6 |
| decode 4 | 42.4 | 74.3 | 58.6 |
| decode 16 | 50.8 | 83.1 | 67.9 |

This PR is ~58–59 µs, clearly faster than PR #4376's 74–92 µs. The ~16 µs
overhead over baseline (42 µs) comes from the one `BlockRadixSort` the fast path
adds.

**Short rows / scan path (N = 30K / 30720):**

| Case | baseline | PR #4376 | This PR (scan) |
| --- | --- | --- | --- |
| prefill 1 row | 74.3 | 74.4 | 74.4 |
| prefill 2 rows | 74.1 | 73.9 | 73.9 |
| prefill 4 rows | 73.7 | 73.7 | 73.6 |
| prefill 8 rows | 80.4 | 80.3 | 80.4 |
| decode 1 | 58.2 | 58.2 | 58.0 |
| decode 4 | 57.4 | 57.4 | 57.2 |
| decode 16 | 63.3 | 63.3 | 63.2 |

On the scan path the three are within measurement noise: the stable-semantics
overhead is essentially zero (e.g. prefill 74.3 vs 74.4 µs), because the
per-tile prefix-sum sweep does about the same work as the baseline collection,
only replacing the atomic race with a deterministic ascending-position
computation. Note the one-block baseline at 30K (~74 µs) is slower than at 60K
(~42 µs) because a forced one-block launch under-utilizes the GPU on short rows
(one block per row); in production 30K would take the multi-block path
(~47 µs). One-block is pinned here only for a same-path comparison.

## Usage

The public API is unchanged; only set the environment variable before process
start:

```bash
export TOPK_STABLE=1
```

```python
from aiter.ops.topk import top_k_per_row_prefill, top_k_per_row_decode
# prefill
top_k_per_row_prefill(logits, rowStarts, rowEnds, indices, values,
                      numRows, stride0, stride1, k=2048)
# decode
top_k_per_row_decode(logits, next_n, seqLens, indices,
                     numRows, stride0, stride1, k=2048)
```


yhl-amd (dismissed):
LGTM

csrc/kernels/topk_per_row_kernels.cu
```diff
@@ -2399,6 +2691,22 @@ enum class Phase
...
+// top-k with smallest-index tie-breaking, so every tensor-parallel rank selects
+// AND orders an identical KV set for identical inputs (required for DSA + TP
+// all-reduce correctness). Process-wide switch, read once and cached; keeps the
+// Python/pybind API unchanged. When on, the prefill/decode entries force the
+// one-block path (the multi-block persistent kernel's cross-block atomic emit
+// is inherently order-dependent and cannot be made stable cheaply).
+inline bool topk_stable_enabled()
+{
+    static const bool v = []() {
+        const char* e = std::getenv("TOPK_STABLE");
```
yhl-amd: plz change it as functional param intead of env

yhl-amd:
Added `b1d9e90df` to the PR head to fix a stable-emitter metadata race.

The block spans multiple wavefronts. In both `last_filter_stable_scan` and the ordered-tie path of `last_filter_stable_fast`, thread 0 could advance the shared tile base while later wavefronts were still reading the old base. Different waves then computed output positions from different bases, leaving holes/stale values in `out_idx`. A memory fence is not sufficient because this is an execution-order race; the fix snapshots the base in every thread and adds a block barrier before thread 0 may advance it.

Validation:
- Production shape `6400 x 81920, k=2048`: before the fix, a fixed input reproduced 11 unwritten indices on iteration 20; after the fix, 100 consecutive runs had zero invalid indices.
- Full `op_tests/test_topk_per_row_stable.py`: PASS.
- Added a repeated 512-row tied-input regression checking index bounds, determinism, ascending order, and reference agreement.
- GLM-5.2 MTP=3 108,016-token end-to-end request: HTTP 200 with no invalid top-k indices or HSA fault.

valarLip (approved):

PR #4416, opened 2026-07-28
Fix large-token FlyDSL MoE launch and output limits

wuhuikx:
## Summary

- Keep FlyDSL MoE route-block launches within HIP's `grid.y <= 65535` limit by automatically increasing `persist_m` from the actual runtime block count.
- Apply the same bounded-grid logic to non-persistent FP8 stage2 kernels instead of forcing `persist_m=1`.
- Route stage2 outputs larger than 4 GiB through per-slot reduction because the atomic path uses a single buffer resource with 32-bit byte offsets.
- Add boundary regression tests for both the HIP grid limit and the 4 GiB atomic-output limit.

## Problem

The [Atomesh 1M benchmark failure](https://github.com/ROCm/ATOM/actions/runs/30325181528/job/90169043709) selected the heuristic A4W4 fallback for:

`gfx950, M-tier=131072, hidden=6144, inter=256, experts=256, topk=8, BF16/A4W4 per_1x32`

The `131072` value in the lookup key is capped by `get_padded_M`; it is not the true runtime token count. At 1M-context scale, the real sorted route-block count can exceed 65535. Stage1 mapped those blocks directly to `grid.y`, so `hipModuleLaunchKernel` rejected the launch with `hipErrorInvalidValue` on every rank.

A second issue appears after making the launch legal: for `M=524288, hidden=6144`, the BF16 stage2 output is 6 GiB. The atomic kernel addresses it through one buffer resource with 32-bit byte offsets, causing high token rows to be silently dropped. FP8/A8W4 stage2 also forced `persist_m=1`, exposing the same grid overflow for large route counts.

## Solution

1. Compute the minimum positive `persist_m` required by the actual M-block count:

   `persist_m = max(requested, ceil(num_m_blocks / 65535))`

   Normal shapes retain `persist_m=1`; only oversized launches process multiple M blocks per workgroup.

2. Use this resolver in generic FlyDSL stage1 and FP8 stage2 so all generated `grid.y` values remain legal.

3. Detect final stage2 outputs above `UINT32_MAX` bytes and select the existing reduce mode. The reduction kernel maps tokens to `grid.x`, avoiding both 32-bit atomic output offsets and the restricted HIP grid dimensions.

## Validation

- [x] `pytest -q op_tests/flydsl_tests/test_flydsl_moe_launch_config.py` — 3 passed
- [x] Reproduced the pre-fix `hipErrorInvalidValue` at `M=524288, topk=8, block_m=64`
- [x] A4W4 fallback passes at `M=524288` and `M=1048576`
- [x] Full GLM shape `M=524288, hidden=6144, inter=256, E=256, topk=8` writes matching first/middle/last rows
- [x] A8W4/FP8 stage2 large-grid reproduction passes with automatic `persist_m`
- [x] Existing `M=131072, hidden=6144, inter=256` fallback remains launchable

## Scope

This fixes the generic mixed FlyDSL A4W4/A8W4 paths used by the reported fallback. A16W4 and INT4 kernels have separate single-M-tile launch implementations and are intentionally left for follow-up changes rather than broadening this fix.

### DIFF
--- a/aiter/ops/flydsl/moe_kernels.py
+++ b/aiter/ops/flydsl/moe_kernels.py
@@ -14,6 +14,9 @@ from aiter.ops.flydsl.kernels.tensor_shim import ptr_arg
 
 _KERNEL_PARAMS: Dict[str, Dict] = {}
 
+# HIP permits a much larger grid.x, but grid.y/grid.z are limited to 65535.
+_HIP_MAX_GRID_DIM_Y = 65535
+
 
 def _get_dtypes():
     from aiter.utility import dtypes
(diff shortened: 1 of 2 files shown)


wuhuikx:
Closing to recreate this PR under the intended XiaobingSuper GitHub account. The Black/Ruff fixes are included on the same branch.

PR #4420, opened 2026-07-28
mla: guard NaN / empty splits in mla_reduce_v1 (persistent MLA split-KV decode)

Duyi-Wang:
# mla: guard NaN / empty splits in `mla_reduce_v1` (persistent MLA split-KV decode)

## Summary

The persistent MLA split-KV decode reduction kernel `mla_reduce_v1`
(`csrc/kernels/mla/reduce.cu`) unconditionally trusts every per-split partial:
it takes `max`/`expf` on the partial LSE, accumulates the partial output, and
(in the `_impl_simple` path) does an **unconditional** final
`reg_out /= sum_e_lse` — it never sanitizes NaN and never guards the
normalization.

A single non-finite (`NaN`/`inf`) partial therefore poisons the **entire**
reduce tile:

```
lse = NaN  ->  max_lse = max(max_lse, NaN) = NaN
           ->  sum_e_lse = ... = NaN
           ->  reg_out / sum_e_lse = NaN / inf
```

This PR adds NaN/empty-split guards to `mla_reduce_v1` so its numerics match the
two reference reductions that already handle this — the **non-persistent** decode
path and the Python reference — plus a regression test. The change is an
**identity transform** whenever all partials are finite.

## Background

Persistent MLA decode splits each request's KV across up to
`max_split_per_batch` partitions; `mla_reduce_v1` merges the per-split partials
(`partial_output` / `partial_lse`) with an online-softmax (FlashAttention-style)
reduction, in two device paths:

- `_impl_simple` (few splits): online rescale loop + final divide by `sum_e_lse`.
- `_impl_massive` (many splits): LSE/output path, normalization folded into
  `lse_scale` (no final divide).

## The gap

The same online-softmax merge exists in two other places, both of which are
defensive — only `mla_reduce_v1` is not:

1. **Non-persistent path** (`aiter/mla.py`): uses `valid_split_count` +
   `use_valid_split_count_reduce` so the reducer **skips empty / invalid splits
   entirely** — a poisoned split never enters the merge.
2. **Reference** `torch_mla_reduce_v1` (`op_tests/test_mla_persistent.py`,
   annotated "matching C++ implementation"): explicitly maps `NaN` LSE →
   `-inf`, `NaN` output → `0`, and returns `0` when `sum_e_lse <= 0`.

`mla_reduce_v1` does none of these, so it is the only one of the three that
propagates poison.

## Why it surfaces at high split count

The persistent path pins a **fixed, high** split count
(`max_split_per_batch` = 32, or 64 when there are no draft tokens), whereas the
non-persistent path auto-selects `<= 16` via `get_meta_param`. With many splits
more of them cover empty / short KV ranges, whose partial LSE/output are not
guaranteed finite. Because one non-finite partial corrupts a whole tile and the
number of splits scales with batch/concurrency, the corruption grows with the
split count; at low split counts it is not observed.

> **Correction (measurement withdrawn).** An earlier revision of this PR cited
> instrumentation of a real decode reporting ~2.15M of 2.76e8 non-finite
> `partial_output` elements and `partial_lse` up to ~3.4e38. **That measurement
> does not support the claim and I am withdrawing it.** Two things came out of
> follow-up work:
>
> 1. Under CUDA graphs the python body of `mla_decode_fwd` only runs during
>    warm-up and capture; real decode batches are graph *replays*, which re-run
>    the captured kernels without re-entering python. A python-level hook
>    therefore samples capture-phase dummy batches, not production traffic.
>    (The kernels themselves, including this patch, do run on every replay --
>    only the instrumentation was blind.)
> 2. Non-finite values in the partial buffers are expected and harmless on their
>    own. The buffers are allocated with `torch.empty` and are heavily
>    over-allocated: on the dp-attention layout (nhead 128, folded to 16) a
>    batch-64 / qlen-4 decode allocates 5104 partial rows and writes and reads
>    only 401 of them. The remaining 92% is untouched allocator memory, and it
>    is never read.
>
> What actually matters is whether a non-finite value lands **inside the reduce
> read set**. Across every op-level configuration tested so far it does not --
> see the section below.

## Fix

Five guards in `mla_reduce_v1` (`csrc/kernels/mla/reduce.cu`):

- `_impl_massive` `cal_lse`: loaded partial LSE `NaN -> -INFINITY` (weight 0).
- `_impl_massive` `load_output`: loaded partial output `NaN` component `-> 0`.
- `_impl_simple` first-split seed: `reg_out` `NaN` component `-> 0`; seed `lse`
  `NaN -> -INFINITY`.
- `_impl_simple` loop body: `oaccu` `NaN` component `-> 0`; `lse_val`
  `NaN -> -INFINITY`.
- `_impl_simple` final normalize: `reg_out /= sum_e_lse` guarded — when
  `sum_e_lse <= 0` or `NaN`, output `0` instead of dividing.

`_impl_massive` has no final divide (normalization is folded into `lse_scale`),
so only its inputs need guarding.

All guards are **identity transforms when partials are finite** — healthy
reductions are bit-for-bit unchanged; only poisoned splits are affected. This
mirrors what the non-persistent path and the torch reference already do.

### Unit test (added in this PR)

`op_tests/test_mla_reduce_nan_guard.py` drives `mla_reduce_v1` directly with
pathological partials and compares against the reference reduction. It needs one
GPU and no model.

```bash
pytest -q op_tests/test_mla_reduce_nan_guard.py
```

Cases, each over three shapes (`nhead=128 msq=1`, `nhead=128 msq=4` (MTP),
`nhead=16 splits=32`): all-finite sanity, NaN partial output, NaN partial LSE,
fully-masked tile (`sum_e_lse == 0`), partial LSE near the fp32 limit, and
`+/-inf` propagation.

- **Before this PR**: `6 failed, 12 passed`

  ```
  FAILED test_mla_reduce_nan_guard.py::test_nan_in_partial_output[nh128_dv512_msq1_tiles8_splits8]
  FAILED test_mla_reduce_nan_guard.py::test_nan_in_partial_output[nh128_dv512_msq4_tiles8_splits8]
  FAILED test_mla_reduce_nan_guard.py::test_nan_in_partial_output[nh16_dv512_msq1_tiles8_splits32]
  FAILED test_mla_reduce_nan_guard.py::test_nan_in_partial_lse[...]  (x3)
  E   AssertionError: kernel produced 8192 non-finite outputs where the
      reference is finite (a poisoned partial leaked into the reduce tile)
  ```

- **With this PR**: `18 passed`

The remaining cases pass either way, which is the point: the guards only change
behaviour for poisoned partials.

### End-to-end

MI355X (gfx950), 8 GPUs, FP8 KV cache, dp-attention (tp8/dp8 -> `nhead=128`),
NEXTN MTP (4 draft tokens), 64 concurrent requests, persistent MLA enabled
(`SGLANG_AITER_MLA_PERSIST=1`) — i.e. `max_split_per_batch = 32`. gsm8k, 1319
questions, greedy:

| aiter | exact_match (strict) | (flexible) |
|---|---|---|
| unpatched | 0.7930 | 0.7194 |
| **this PR** | **0.9515** | **0.9545** |

Two-node prefill/decode disaggregated, same workload:

| aiter | exact_match (strict) |
|---|---|
| unpatched | 0.7111 |
| **this PR** | **0.9560** |

Disabling the persistent MLA kernel entirely (non-persistent path, which already
skips invalid splits) scores 0.953 on the same setup — consistent with the
guards restoring the persistent path to the non-persistent path's behaviour.

## Performance

`run_perftest` A/B over the whole `mla_decode_fwd` (stage1 + reduce), persistent,
same shapes:

| shape (nhead=16, qlen=1, varlen) | baseline µs | patched µs | Δ% |
|---|---|---|---|
| batch 64,  ctx 3200 | 24.225 | 24.339 | +0.47% |
| batch 128, ctx 3200 | 37.762 | 37.819 | +0.15% |
| batch 256, ctx 3200 | 63.918 | 64.112 | +0.30% |
| batch 64,  ctx 8192 | 40.596 | 41.344 | +1.84% |
| batch 128, ctx 8192 | 73.785 | 74.497 | +0.96% |
| batch 256, ctx 8192 | 137.813| 139.173| +0.99% |

The guards add only register-level `x != x` checks plus a conditional select in a
memory-bound reduction; end-to-end decode overhead is **0.15%–1.8%** (largest for
long context, where the split count and thus the number of guarded elements is
highest), with results unchanged on finite data.

## Notes for reviewers

- The change is confined to `mla_reduce_v1`; it does not alter split counts,
  memory traffic, kernel selection, or the online-softmax math.
- The guards stop poison from spreading inside the reduction. Where the
  non-finite partials originate is answered as far as an op-level test can take
  it in the section below.

## Op-level probe: where the non-finite values do and do not come from

`op_tests/test_mla_decode_persistent_nan_probe.py` runs the real pipeline --
`get_mla_metadata_v1` -> `mla_decode_stage1_asm_fwd` -> `mla_reduce_v1` -- and
tries to make a non-finite value appear. It pre-fills the partial buffers with a
sentinel before stage1, so every slot the reduce reads is classified afterwards
as *never written*, *written non-finite*, or *written normally*.

```
pytest -q -s op_tests/test_mla_decode_persistent_nan_probe.py
111 passed
```

Two hypotheses were on the table. Neither survives on in-contract input:

**"the reduce consumes a slot stage1 never wrote"** -- not observed. The read
set derived from `reduce_indptr` / `reduce_final_map` / `reduce_partial_map` is
*exactly equal* to the set of rows the `work_info_set` schedules stage1 to
write, across every configuration tested (b32..b128, qlen 1 and 4, ctx 256..16K,
32 and 64 splits, 16 seeds each). The strongest form of the test hands the
caching allocator NaN-filled blocks of exactly the shapes `aiter/mla.py` is
about to `torch.empty`, immediately before the decode:

| case | partial rows (total / read) | NaN left in unread slots | NaN in slots the reduce reads | non-finite output |
|---|---|---|---|---|
| b64 q1 ctx4K split64 | 319 / 270 | 401,408 | 0 | 0 |
| b64 q4 ctx4K split32 | 1276 / 1076 | 1,638,400 | 0 | 0 |
| b128 q1 ctx256 split64 | 383 / 36 | 2,318,336 | 0 | 0 |
| b128 q4 ctx256 split32 | 1532 / 176 | 11,108,352 | 0 | 0 |
| b32 q4 ctx16K split32 | 1148 / 1084 | 524,288 | 0 | 0 |

The poison demonstrably lands in the buffer and survives in the slots nobody
reads, yet stage1 overwrites every slot the reduce touches. Zero-split reduce
tiles were checked separately: the planner never initialises their
`reduce_final_map`, so in practice they carry out-of-range garbage such as
`[24197, 53006)` for a 512-row output (110 of 128 tiles in the short-context
case). Rewriting those entries to the maximally destructive in-range
`[0, total_q)` clobbers nothing -- the kernel does skip them.

**"stage1 writes a non-finite LSE"** -- not observed on in-contract input; the
written LSE range stays around `[-1.8, 7.3]`. It *is* reproducible on
out-of-contract input: with `kv_len < q_len` the leading query rows of a request
see no key at all under the causal mask, and stage1 emits NaN for exactly those
rows and no others (16 of 32 rows for kv_len 2 / q_len 4). Those works carry
`partial_qo_loc == -1`, i.e. stage1 writes the final output directly and the NaN
never passes through `mla_reduce_v1`. Decode always satisfies kv_len >= q_len,
so this is a documented input-contract boundary rather than a kernel defect.

The probe also drives the two paths where a server reuses buffers and rewrites
metadata in place -- cuda-graph capture plus 12 replays with re-planned
metadata, and `decode_update_mla_metadata_v1` on the speculative-decode path
with random reject counts. Both keep the read-set invariant and finite results.
Worth noting: cuda-graph replay is the one setting where non-finite values were
seen in the partial buffer at all (5,994 elements in one configuration) -- in
unread slots, with a finite final result, consistent with everything above.

On a failure the probe dumps the full `mla_decode_fwd` argument list
(planner inputs, planner outputs, results) to `MLA_NAN_DUMP_DIR`, with
`MLA_NAN_DUMP_FULL=1` to include `q` / `kv_buffer`, so a hit can be replayed
offline.

## RESOLVED -- this patch is not the fix, and `main` is not affected

Root-cause work finished. The summary: **the guards in this PR are an accidental
fix for a synchronisation bug that upstream `main` already fixed.** I recommend
closing this PR. Details, so the analysis is not lost:

### The actual defect

`mla_reduce_v1_impl_massive` writes `p_lds_lse_scale[]` in `reduce_lse_massive()`
and reads it back in `reduce_output_massive()`. In older revisions the boundary
between them was:

```c
__builtin_amdgcn_sched_barrier(0);
__builtin_amdgcn_s_barrier();
```

`__builtin_amdgcn_s_barrier()` emits a bare `s_barrier`: no `s_waitcnt
lgkmcnt(0)` waiting on outstanding LDS traffic, and no compiler memory fence.
`sched_barrier(0)` only prevents instruction reordering. So one wave can read
`p_lds_lse_scale` before the other wave's writes have landed.

With `kSizeDV == 512` the workgroup is two waves (`kNumWarps = 2`,
`kNumThreads = 128`) and thread `t` stores channels `[4t, 4t+4)`
(`kVecWidth = 4`), so **wave 1 owns channels [256, 512)**. A torn `lse_scale`
multiplies into `reg_out`, so the corruption is **large but finite**
(~1e35..1e38), never NaN.

**This is already fixed on `main`** by `cdd6628f9` -- *"Make MLA Reduce Kernel
Work on 32-waves Device (#3669)"*, 2026-07-01 -- which replaced that
`__builtin_amdgcn_s_barrier()` with `__syncthreads()`. The commit is framed as
gfx12 enablement; the barrier upgrade reads as incidental to that framing, but
it is the correctness fix for MI355X MLA decode. Worth knowing, because anyone
running an image built from an aiter that predates it will hit this.

### Evidence

Measured on MI355X, DeepSeek-R1 MXFP4, TP8/EP8/dp-attention, NEXTN MTP
(steps=3, draft=4), fp8 KV, concurrency 64. "Corrupt rows" counts output rows
with `|x| > 1e20` or non-finite, over ~1.8M `mla_reduce_v1` calls per arm:

| arm | corrupt rows | rate/call | gsm8k (300, 5-shot) |
|---|---|---|---|
| unpatched | 515 | 2.79e-4 | 0.7567 |
| barrier added at the *massive loop tail* (wrong place) | 587 | 3.15e-4 | 0.7000 |
| `_ps` work-item loop barrier upgraded only | 578 | 3.03e-4 | 0.7533 |
| **`s_barrier` -> `__syncthreads()` at the lse->output boundary** | **0** | **0** | **0.9667 / 0.9733 / 0.9667** |

Three independent runs of the last arm. The `_ps` loop barrier -- also upgraded
by `cdd6628f9` -- contributes nothing measurable; the single barrier inside
`mla_reduce_v1_impl_massive` is the whole bug. That function is reached from
both `kn_mla_reduce_v1` and `kn_mla_reduce_v1_ps` (dispatch markers confirmed
both kernels run in production), which is why fixing it alone is sufficient.

### Why these guards appeared to work

None of the 515 corrupt rows were non-finite -- they were finite ~1e37 -- so the
NaN sanitising in this patch cannot have caught a single one. What it did was
edit device code inside `load_output`, perturbing instruction scheduling enough
to close the race window. A control arm that forced the identical JIT rebuild
with a host-side-only edit did **not** recover accuracy (0.7267, 0.7700), which
is what isolates the effect to device codegen rather than to the rebuild.

### Corrections to earlier revisions of this PR

- The cited "~2.15M of 2.76e8 non-finite `partial_output`, `partial_lse` ~3.4e38
  on a real run" was withdrawn earlier. Two reasons: under CUDA graphs the
  python body of `mla_decode_fwd` only runs during capture, so the hook sampled
  dummy batches; and non-finite values in the partial buffers are expected and
  harmless, since the buffers are `torch.empty` and heavily over-allocated.
- Instrumented real traffic confirms the reduce never ingests garbage:
  **0 non-finite and 0 large-magnitude values inside the reduce read set across
  6.7M calls**, while the output was corrupt. The reduce *generates* the
  garbage; it does not propagate it. That is the opposite of this PR's premise.

### What I would still offer

`op_tests/test_mla_reduce_nan_guard.py` and
`op_tests/test_mla_decode_persistent_nan_probe.py` in this PR are independent of
the guards and still catch real classes of defect (the probe's read-set
invariant check is what proved the reduce was not ingesting bad partials). If
maintainers want either as a standalone test-only PR, say so and I will open one
against current `main`. Otherwise this can be closed as-is.

### DIFF
--- a/csrc/kernels/mla/reduce.cu
+++ b/csrc/kernels/mla/reduce.cu
@@ -242,6 +242,7 @@ __device__ void reduce_lse_massive(const MlaReduceKernelV1Params& params,
                 lse = g_partial_lse.template _load<1>(partial_lse_seq_byte_offset +
                                                       reduce_tile_pos * int32_t(sizeof(float)))[0];
             }
+            if(lse != lse) lse = -INFINITY;
             return lse;
         };
 
@@ -342,9 +343,11 @@ __device__ void reduce_output_massive(const MlaReduceKernelV1Params& params,
     auto load_output = [&](const int32_t reduce_partial_map) -> vec_f32_t {
         const int32_t tile_byte_offset =
             reduce_partial_map * int32_t(Traits::kNumHeadQ * Traits::kSizeDV * sizeof(float));
-        return buf_load_vec<kVecWidth>(g_partial_output,
+        vec_f32_t _v = buf_load_vec<kVecWidth>(g_partial_output,
                                        partial_output_seq_byte_offset + tile_byte_offset +
                                            thread_byte_offset);
+        opus::static_for<kVecWidth>([&](auto i){ if(_v[i.value]!=_v[i.value]) _v[i.value]=0.f; });
+        return _v;
     };
 
     auto oaccu_0      = load_output(reduce_partial_map_0);
@@ -631,8 +634,10 @@ __device__ void mla_reduce_v1_impl_simple(const MlaReduceKernelV1Params& params,
             g_partial_output,
             partial_output_seq_byte_offset + reduce_tile_pos_out_byte_start + thread_byte_offset);
 
-        const float lse = g_partial_lse.template _load<1>(
+        opus::static_for<kVecWidth>([&](auto i){ if(reg_out[i.value]!=reg_out[i.value]) reg_out[i.value]=0.f; });
+        float lse = g_partial_lse.template _load<1>(
             partial_lse_seq_byte_offset + reduce_tile_pos_lse_start * int32_t(sizeof(float)))[0];
+        if(lse != lse) lse = -INFINITY;
         float max_lse   = lse;
         float sum_e_lse = 1.0f;
 
@@ -647,8 +652,10 @@ __device__ void mla_reduce_v1_impl_simple(const MlaReduceKernelV1Params& params,
                 g_partial_output,
                 partial_output_seq_byte_offset + reduce_tile_pos_out_bytes + thread_byte_offset);
 
-            const float lse_val = g_partial_lse.template _load<1>(
+            opus::static_for<kVecWidth>([&](auto i){ if(oaccu[i.value]!=oaccu[i.value]) oaccu[i.value]=0.f; });
+            float lse_val = g_partial_lse.template _load<1>(
                 partial_lse_seq_byte_offset + reduce_tile_pos_lse * int32_t(sizeof(float)))[0];
+            if(lse_val != lse_val) lse_val = -INFINITY;
             const float new_max_lse = opus::max(max_lse, lse_val);
             const float old_scale   = expf(max_lse - new_max_lse);
             const float new_scale   = expf(lse_val - new_max_lse);
@@ -661,8 +668,9 @@ __device__ void mla_reduce_v1_impl_simple(const MlaReduceKernelV1Params& params,
             sum_e_lse = sum_e_lse * old_scale + new_scale;
         }
 
+        const bool sum_ok = (sum_e_lse > 0.f) && (sum_e_lse == sum_e_lse);
         opus::static_for<kVecWidth>(
-            [&](auto i) { reg_out[i.value] = reg_out[i.value] / sum_e_lse; });
+            [&](auto i) { reg_out[i.value] = sum_ok ? (reg_out[i.value] / sum_e_lse) : 0.f; });
 
         const int32_t store_byte_offset = final_out_byte_offset_base +
                                           seq_idx * params.stride_s_o * int32_t(sizeof(out_t)) +
--- a/op_tests/test_mla_decode_persistent_nan_probe.py
+++ b/op_tests/test_mla_decode_persistent_nan_probe.py
@@ -1,0 +1,1285 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Op-level probe for non-finite values on the persistent MLA decode path.
+
+Unlike ``test_mla_reduce_nan_guard.py`` -- which drives ``mla_reduce_v1`` in
+isolation with injected partials -- this file runs the real decode pipeline:
+
+    get_mla_metadata_v1()            # planner
+      -> mla_decode_stage1_asm_fwd() # split-KV attention, writes the partials
+        -> mla_reduce_v1()           # cross-split reduction
+
+and answers, with hardware data, where a non-finite value can come from:
+
+  H1  "a slot the planner reserved but stage1 never wrote"
+      (empty split / degenerate split / unused tile). The partial buffers are
+      allocated with ``torch.empty`` in ``aiter/mla.py``, so an unwritten slot
+      that the reduce still reads contains whatever was in that memory.
+      Note the torch reference (``torch_mla_extend_split_kv`` in
+      test_mla_persistent.py) sizes its partial buffers with ``zeros`` /
+      ``-inf`` fill, i.e. it defines unwritten slots as neutral -- the asm path
+      has no such guarantee.
+
+  H2  "stage1 itself wrote a non-finite LSE / output"
+
+The two are told apart by pre-filling the partial buffers with a distinctive
+finite sentinel before stage1 runs and classifying every slot afterwards:
+
+    still == SENTINEL  -> never written by stage1            (H1)
+    non-finite         -> written non-finite by stage1       (H2)
+    finite, != SENTINEL-> written normally
+
+Run:
+    pytest -q op_tests/test_mla_decode_persistent_nan_probe.py
+
+Env knobs:
+    MLA_NAN_DUMP_DIR    where to write the input dump on failure (default ".")
+    MLA_NAN_DUMP_FULL   set to 1 to include q / kv_buffer in the dump
+"""
+
+import os
+
+import pytest
+import torch
+
+import aiter
+from aiter import dtypes
+from aiter.jit.utils.chip_info import get_gfx
+
+# A finite value that never occurs in a real partial output or LSE. Slots that
+# still hold it after stage1 were not written by stage1.
+SENTINEL = -1.2345678e30
+
+DUMP_DIR = os.environ.get("MLA_NAN_DUMP_DIR", ".")
+DUMP_FULL = os.environ.get("MLA_NAN_DUMP_FULL", "0") not in ("", "0")
+
+# DeepSeek-style MLA decode geometry: kv_lora_rank 512 + rope 64, one KV head,
+# page_size 1, fp8 KV cache. nhead 16 == 128 heads over TP8.
+KV_LORA_RANK = 512
+QK_ROPE_HEAD_DIM = 64
+QK_HEAD_DIM = KV_LORA_RANK + QK_ROPE_HEAD_DIM
+V_HEAD_DIM = KV_LORA_RANK
+NHEAD_KV = 1
+PAGE_SIZE = 1
+
+
+class Case:
+    """One decode configuration.
+
+    ``min_kv`` is the shortest KV length generated. Causal decode requires
+    kv_len >= q_len (otherwise the leading query rows see no key at all and the
+    softmax is over an empty set); keep it at decode_qlen for in-contract runs.
+    """
+
+    def __init__(
+        self, batch, nhead, decode_qlen, ctx_len, max_split_per_batch, min_kv=None
+    ):
+        self.batch = batch
+        self.nhead = nhead
+        self.decode_qlen = decode_qlen
+        self.ctx_len = ctx_len
+        self.max_split_per_batch = max_split_per_batch
+        self.min_kv = decode_qlen if min_kv is None else min_kv
+
+    @property
+    def id(self):
+        return (
+            f"b{self.batch}_nh{self.nhead}_q{self.decode_qlen}"
+            f"_ctx{self.ctx_len}_split{self.max_split_per_batch}"
+        )
+
+
+CASES = [
+    # decode, no speculative tokens -- the persistent planner pins 64 splits
+    Case(batch=64, nhead=16, decode_qlen=1, ctx_len=4096, max_split_per_batch=64),
+    # MTP / speculative decode: 4 draft tokens, 32 splits
+    Case(batch=64, nhead=16, decode_qlen=4, ctx_len=4096, max_split_per_batch=32),
+    # short and highly variable contexts -- maximises empty / degenerate splits
+    Case(batch=128, nhead=16, decode_qlen=1, ctx_len=256, max_split_per_batch=64),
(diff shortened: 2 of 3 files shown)


Duyi-Wang:
Root-cause work is finished and it **contradicts this PR's premise**, so I have rewritten the description and I recommend closing this. Summary for anyone subscribed:

The defect is a **LDS synchronisation bug in `mla_reduce_v1_impl_massive`**, not NaN propagation. `reduce_lse_massive()` writes `p_lds_lse_scale[]`, `reduce_output_massive()` reads it, and the boundary used `__builtin_amdgcn_s_barrier()` — a bare `s_barrier` with no `s_waitcnt lgkmcnt(0)` and no compiler fence. With `kSizeDV == 512` the workgroup is two waves and thread `t` stores channels `[4t, 4t+4)`, so wave 1 owns `[256, 512)` — exactly the half-row that comes back corrupt, with `[0, 256)` numerically correct. A torn `lse_scale` multiplies into `reg_out`, so the damage is large-but-**finite** (~1e35..1e38), never NaN.

**`main` already fixes this**, in `cdd6628f9` ("Make MLA Reduce Kernel Work on 32-waves Device", #3669), which replaced that barrier with `__syncthreads()`. The commit is framed as gfx12 enablement, so the correctness impact on MI355X MLA decode is easy to miss — that is the part worth flagging: images built from an aiter predating it will hit this on speculative-decode MLA workloads.

Measured on MI355X / DeepSeek-R1 MXFP4 / TP8 dp-attention / NEXTN MTP / fp8 KV / conc 64, counting output rows with `|x| > 1e20` or non-finite over ~1.8M reduce calls per arm:

| arm | corrupt rows | gsm8k (300, 5-shot) |
|---|---|---|
| unpatched | 515 | 0.7567 |
| `_ps` work-item loop barrier upgraded only | 578 | 0.7533 |
| **`s_barrier` → `__syncthreads()` at the lse→output boundary** | **0** | **0.9667 / 0.9733 / 0.9667** |

Attribution: the single barrier inside `mla_reduce_v1_impl_massive` is the whole bug. The `_ps` loop barrier (also upgraded by `cdd6628f9`) measurably contributes nothing, though it is the same latent hazard and correct to fix.

Why the guards in this PR appeared to work: **none of the 515 corrupt rows were non-finite**, so the NaN sanitising cannot have caught any of them. The patch edits device code inside `load_output` and perturbs instruction scheduling enough to close the race window. A control arm forcing the identical JIT rebuild with a host-side-only edit did *not* recover accuracy (0.7267 / 0.7700), which isolates the effect to device codegen rather than the rebuild.

I have also withdrawn an earlier claim in this PR that instrumentation showed millions of non-finite partials in production — under CUDA graphs the python body of `mla_decode_fwd` only runs during capture, so that hook sampled dummy batches. Instrumented real traffic shows **0 non-finite and 0 large-magnitude values inside the reduce read set across 6.7M calls** while the output was corrupt: the reduce generates the garbage rather than propagating it.

The two test files here are independent of the guards and still useful — happy to re-open either as a test-only PR against current `main` if a maintainer wants it. Otherwise please close.

PR #4244, opened 2026-07-15
[triton][mqa] fix silent tail-row drop in deepgemm_fp8_paged_mqa_logits at large output stride

zejunchen-zejun:
At large MTP batches the sparse-indexer decode path issues one big deepgemm_fp8_paged_mqa_logits + top_k_per_row_decode launch whose later rows can retain stale sparse-KV indices in the persistent buffer (rows beyond the CU wave capacity), so requests past the cliff attend to wrong KV and their drafts get rejected -> acceptance halves (~50% -> ~25% at con256, cliff at seq 128), so when specify max num seqs to the value which is smaller than 128, the accept ratio restores to the normal value, while for default 512, most of requests has low draft token accept ratio. Here is the histogram illustration.
image

Root cause: the Gluon kernel stores logits with gl.amd.cdna3.buffer_store, whose AMD hardware voffset is a 32-bit byte offset. The output address for row r is `r * out_logits.stride(0) * out_logits.element_size()`. Once that reaches 2**31 the offset overflows and the store is silently dropped, leaving the tail rows unwritten (they keep their prior/zero contents).

This bites callers that allocate a wide dense logits tensor. Concretely, GLM-5.2 sparse-MLA (DSA) MTP decode allocates logits as [batch*next_n, max_model_len] with max_model_len = 1<<20; at con>=256 with next_n=4 that is [1024, 1<<20], so stride(0)=1<<20 and row 512 hits exactly 512 * (1<<20) * 4 = 2**31 bytes. Rows 512..1023 are never written -> top-k reads all-zero rows -> wrong sparse-KV indices -> MTP acceptance collapses (~50% -> ~25%). Verified with a standalone reproducer: with physical_cols=1<<20 the first untouched row is 512; with physical_cols=600000 it moves to 895 (= where r*stride*4 first crosses 2**31), proving it is a byte-offset boundary, not a fixed row count.

Fix
  Address the output in 64-bit by advancing the base pointer, keeping the buffer_store voffset a small int32 column index:
  - gluon/pa_mqa_logits.py — all 19 stores across the 3 gluon kernels changed from ptr=OutLogits_buffer, offsets=row*stride_out_batch + col to ptr=OutLogits_buffer + row.to(tl.int64)*stride_out_batch, offsets=col (int32 columns only).
  - attention/pa_mqa_logits.py (wrapper) — declare stride_out_batch as i64 and drop the tt.pointer_range 32 hint on OutLogits_buffer so the base may be addressed beyond 2 GB.
[...]

PR #4427, opened 2026-07-28
[Bugfix][Triton] Fix large-stride KV address overflow in paged MQA logits

shen-shanshan:
## Motivation

Fix `deepgemm_fp8_paged_mqa_logits` producing non-finite logits when paged KV cache blocks are accessed through a large physical block stride.

- Bug report: https://github.com/Fangzhou-Ai/vllm/issues/13
- vLLM side temporary fix: https://github.com/vllm-project/vllm/pull/49714

This is distinct from #4244. That change fixed 32-bit `buffer_store` output row offsets. This change fixes 32-bit KV value/scale input load offsets.

### Root Cause

The affected vLLM sparse-attention cache is a per-layer view over a larger allocation. A representative failing call had:

- KV block stride: `1,435,968` bytes
- Maximum referenced physical block: `4564`
- Maximum KV byte offset: `6,553,757,952`

The wrapper declared `stride_k_seq` and `stride_scale_seq` as `i32`. The preshuffle Gluon kernel then formed AMD `buffer_load` offsets as:

```text
physical_block_id * stride_k_seq + in_block_offset
```

AMD buffer voffsets are 32-bit. The multiplication overflowed for the real cache layout, causing the kernel to read unrelated FP8 values/scales. Those loads produced `NaN` and `+Inf` logits even though all referenced inputs were finite and all output positions were written.

Runtime instrumentation observed 9 NaNs and 1 positive infinity among 3,748 valid logits in a failing call. Q, weights, referenced KV values/scales, context lengths, and block-table entries were all valid. The same corruption appeared at identical coordinates on all eight TP ranks.

Compacting the exact same referenced KV blocks removed all non-finite outputs, isolating the failure to large-stride addressing.

### Fix Method

- Declare `stride_k_seq` and `stride_scale_seq` as `i64` in the compiled kernel signature.
- Use 64-bit pointer arithmetic for preshuffled KV value and scale loads in the exercised pipeline branch.
- Use `gl.load` for these accesses because Gluon AMD `buffer_load` requires an `int32`/`uint32` offset tensor and therefore cannot represent the failing address range.
- Add a regression test whose KV layout crosses the 32-bit byte-offset boundary and compare it against an identical compact-layout reference.

### Exact captured-input replay

The failing input was replayed with the original block stride, physical block IDs, and a maximum byte offset of `6,553,757,952`.

### End-to-end

As mentioned in https://github.com/Fangzhou-Ai/vllm/issues/13.

### Regression test

```bash
python -m pytest -sv op_tests/test_pa_mqa_logits_offset.py::test_paged_mqa_logits_large_kv_stride_matches_compact_reference
```

The test constructs identical compact and large-stride paged KV inputs, forces the physical KV address beyond `2**32`, and requires finite, bit-identical logits.

### Exact captured-input replay

- Before: 10 non-finite valid logits
- After: 0 non-finite logits and 0 unwritten sentinel values

### End-to-end

The original concurrent DeepSeek-V4 TP8 serving workload completed without non-finite paged-MQA logits or a GPU memory access fault after the fix.

### Regression test

```text
1 passed in 16.92s
```

## Submission Checklist


shen-shanshan:
Fixed in https://github.com/ROCm/aiter/pull/4774.

PR #4432, opened 2026-07-29
fix: use residual.stride(1) for MHC HC-slice indexing

steamedMantou:
The MHC kernels derived the distance between HC slices as residual_stride / hc_mult, where residual_stride is residual.stride(0). That only holds for a packed residual. For a valid non-compact view such as shape=(1, 4, 7168) stride=(229376, 7168, 1) it yields 57344 instead of 7168, so the kernels read outside the tensor. On MI308X the invalid read lands on an unmapped page and the process aborts with VM_L2_PROTECTION_FAULT (UTCL2 client TCP).

- pass residual.stride(1) into mhc_pre_big_fuse, mhc_pre_big_fuse_rmsnorm and both mhc_post kernels, and use it for residual HC indexing
- size the residual gmem descriptors from the largest offset a kernel can reach instead of m_oob * residual_stride, which was wide enough to let the invalid loads through instead of clamping them
- guard the residual load/store of the phantom row in mhc_pre_big_fuse, whose dispatch always uses num_rows=2, so an odd m does not issue reads for the row beyond m_oob
- reject residuals that are not 3D or whose hidden dim is not contiguous

mhc_post keeps hidden_size for its output offsets, which stay packed. launch_mhc_post_raw gains the new parameter; its fused AR caller passes hidden_size, matching the packed layout that path assumes.

Verified on MI308X (gfx942): outputs are bit-identical to the previous build on packed residuals except for 4 of 58.7M bf16 elements at m=8192 (one ulp, from the added guard changing fp contraction), and the error against the fp32 reference is unchanged. op_tests/test_mhc.py reports hip_err=0 for mhc_pre and mhc_post.

## Submission Checklist

### DIFF
--- a/csrc/include/mhc.h
+++ b/csrc/include/mhc.h
@@ -58,6 +58,7 @@ void launch_mhc_post_raw(hipStream_t stream,
                          int hidden_size,
                          int x_stride,
                          int residual_stride,
+                         int residual_hc_stride,
                          int store_nt = -1);
 void mhc_fused_post_pre_gemm_sqrsum(
     torch::Tensor& gemm_out_mul,    // (split_k, m, hc_mult3)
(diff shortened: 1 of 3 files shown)


junhaha666:
Hello, in what kind of scenario would such discontinuous residuals be input?

PR #4436, opened 2026-07-29
[FlyDSL][LLVM] update flydsl version and Adapt flydsl kernels to internal LLVM ROCDL API changes

jli-melchior:
- buffer_ops: RawPtrBufferLoadOp/StoreOp aux param changed from positional Value to keyword-only Attribute; pass as IntegerAttr
- fmha_core_loop/fmha_prologue: remove signA, signB, modC, reuseA, reuseB from wmma_f32_16x16x32_bf16 calls (params removed upstream)

## Submission Checklist


scripts/test_correct_flydsl_gfx1250.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Smoke-test all FlyDSL gfx1250-relevant op_tests after a FlyDSL version bump.
```
coderfeli: why need this test ? clean?

scripts/test_perf_flydsl_gfx1250.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
coderfeli: clean this test

scripts/test_correct_flydsl_gfx1250.sh
jli-melchior: can we keep this for the upgrade regression tests so that we can make sure kernel is correct when compiler is upgrade.

coderfeli (approved):
LGTM

coderfeli (approved):

PR #4450, opened 2026-07-30
[GFX950] Relocate MLA Gluon kernel and unify decode dispatch

LiuYinfeng01:
## Summary

- Relocate the current gfx950 MLA Gluon kernel to aiter/ops/triton/_gluon_kernels/gfx950/attention/mla.py.
- Update sparse-prefill, benchmark, and test imports to the architecture-specific path.
- Connect the validated low-head BF16 configuration to the unified mla_decode_fwd entry while preserving the Triton fallback for all other configurations.

## Scope

The split-scheduling performance optimization originally developed in this PR has already landed through #4555. This revision does not add a new split policy or further kernel performance tuning.

The moved kernel is byte-for-byte identical to the current file on main, so the relocation carries the implementation from #4555 and the subsequent NaN fix from #4711 without reimplementing either change.

## Unified decode dispatch

The gfx950 Gluon path is selected only for the validated configuration:

- up to 16 query heads and one KV head
- BF16 query, KV cache, and output
- page size 1
- 512 latent dimensions + 64 RoPE dimensions
- contiguous, unshuffled tensors with no quantization scales

Unsupported configurations continue to use the existing Triton path.

## Validation

- Added routing guard coverage, including fallback cases.
- Added gfx950 correctness coverage for 12/16 heads, single-token decode, MTP decode, and 100K context.
- Ruff check passes on all changed Python files.
- Python syntax compilation passes.
- gfx950 execution tests require a ROCm gfx950 environment and were not run on the local Windows host.


JohnQinAMD:
This now overlaps #4405's caller-owned `num_kv_splits` override. #4450's
device-side runtime buckets are the more complete Kimi-K3 scheduling policy;
please preserve equivalent coverage for #4405's focused split-policy CI test
(`test_mla_gluon_split_policy.py`). If #4450 lands without that unit coverage,
I will rebase #4405 into a test-only follow-up rather than duplicate the
implementation.

LiuYinfeng01:
@JohnQinAMD Thanks for pointing this out. I added equivalent split-policy and boundary coverage in commit dd521e2.

I confirmed that the #4450 measurements were collected on this machine, which has 8× AMD Instinct MI355X GPUs (`gfx950`). The end-to-end Kimi-K3 measurements used TP8, CUDA Graph enabled, and the 100K-input / 1K-output workload described in the PR.

The new coverage includes:

1. A shared `_resolve_num_kv_splits()` helper that preserves the automatic policy when the override is `None` and validates explicit caller-owned overrides.
2. `test_mla_gluon_split_policy.py` with 14 parameterized cases covering automatic-policy passthrough, explicit split values `1`, `16`, `32`, and `256`, out-of-range values, and invalid non-integer values.
3. A focused Kimi-K3 TP8 CI invocation covering contexts `1`, `31`, `32`, `63`, `64`, `65`, and `8192` with an explicit 32-split override.
4. The existing automatic runtime bucket behavior remains unchanged when no override is supplied.

Focused unit-test result:

```text
14 passed in 5.73s
```

Explicit split-boundary command:

```bash
python op_tests/test_mla.py   --decode-only   -c 1 31 32 63 64 65 8192   -b 1 -n 12,1   -d bf16 -kvd bf16   -k 512 -qn 512 -qr 64 -vh 512 -blk 1   --varlen -splits 32
```

All seven boundary cases passed the PyTorch golden check with `atol=0.01` and `rtol=0.01`, including cases where the explicit split count is greater than the sequence length.

I also reran the automatic graph-safe policy without an explicit override at contexts 4K, 100K, and 262K. Output and LSE checks passed at all three lengths, and the varlen `batch=3`, `context=10000` case passed as well.

This should preserve #4405's focused split-policy and boundary coverage while keeping #4450 as the single runtime scheduling implementation.

aiter/ops/triton/gluon/mla_gluon.py
```diff
@@ -19,6 +19,10 @@
 #                        (batch, split, head_block*qlen) grid. Full decode
 #                        (stage-1 + stage-2 reduce into the final O). A partial
 #                        last head block (nhead % BLOCK_H != 0) masks OOB heads.
+#   REGIME='bh12bn64'  - Kimi K3 TP8 specialization of bh16bn64 for exactly
```
Dewei-Wang-sh: no need to add a new regime, this should re-use bh16bh64, and the new split will benefit all bh<=16

aiter/ops/triton/gluon/mla_gluon.py
```diff
@@ -156,10 +178,35 @@ def _mla_gluon(
     # additionally bounds min_kv_seq_len so num_iter >= 3 for its gl.assume.
     # Trade-off: at seqs just above the wrapper minimum the last CU does up to
     # ~(floor + NUM_KV_SPLITS - 1)/floor more work than the others.
-    kv_len_per_split = cur_batch_seq_len // NUM_KV_SPLITS
+    active_kv_splits = NUM_KV_SPLITS
+    if DYNAMIC_KV_SPLITS:
+        # CUDA Graph capture freezes Python-side constexpr decisions. Select the
+        # measured h12 bucket from the runtime sequence length instead, while
+        # retaining a fixed 112-workgroup launch and workspace shape.
+        active_kv_splits = gl.where(
```
Dewei-Wang-sh: pass the kv_splits from outside

Dewei-Wang-sh (changes requested):

LiuYinfeng01:
The optimization from this PR has been incorporated into #4555, so I’m drafting this PR.

k50112113 (approved):

valarLip (approved):

aiter/ops/triton/attention/mla.py
```diff
@@ -35,14 +35,65 @@
...
+    q_descale,
+    kv_descale,
+    q_scales,
+    out_scale,
+    skip_reduce,
+    q_is_contiguous,
+    kv_buffer_is_contiguous,
+    out_is_contiguous,
+):
+    """Select Gluon only for the validated gfx950 low-head decode layout."""
```
Dewei-Wang-sh: we have validated many config, yet, you only include "low-head decode"
this is not acceptable.

PR #4451, opened 2026-07-30
Work around all-zero fused MoE output at block_m=32

jatseng-ai:
<html>
<body>
<h2 data-pm-slice="0 0 []">Summary</h2><p>The CK 2-stage fused-MoE kernel returns an <strong>identically zero output tensor</strong> at
<code>block_m=32</code> on gfx942. It does not raise, does not produce NaNs, and logs
nothing — <code>fused_moe</code> simply returns all zeros.</p><p><code>get_block_size_M</code> selects <code>block_m=32</code> for token counts &lt;= 32, which in serving
means <strong>every decode step</strong>. Prefill uses larger tiles and is correct. The result
is a MoE model that reads context correctly and then degenerates during
generation, because the entire routed-expert contribution is silently dropped
while attention and (unfused) shared experts keep the output superficially
fluent.</p><p>This PR clamps <code>block_m</code> to &gt;= 64 on the 2-stage path and adds a regression test.
<strong>It is a workaround, not a root-cause fix</strong> — see "Open question" below.</p><h2>Reproducer</h2><p>Standalone, ~2 min, no model download. bf16 / <code>QuantType.No</code>,
<code>model_dim=4096, inter_dim=512, E=128, topk=8</code>, identical seeds on both versions:</p>
token | v0.1.18 absmax | state | v0.1.13.post1 absmax | selected block_m
-- | -- | -- | -- | --
1 | 0.000000 | all zero | 0.012573 | 32
8 | 0.000000 | all zero | 0.015259 | 32
32 | 0.000000 | all zero | 0.020142 | 32
33 | 0.022949 | ok | 0.022949 | 64
128 | 0.020752 | ok | 0.020752 | 128

<p></p><h2>Test</h2><p><code>op_tests/test_moe_blockm_invariance.py</code> asserts that <code>fused_moe</code> results do not
depend on <code>block_m</code>, which is a scheduling knob and must not change what is
computed. It covers bf16 and fp8/<code>per_Token</code> at token 1 (decode) and token 64
(to decouple tile size from token count).</p><p>Two details that matter for anyone extending it:</p><ul><li><p>The failure mode is an all-zero tensor, not an exception or a NaN, so a test
that only checks for a successful launch or for finiteness passes straight
through it. The test asserts non-zero output explicitly.</p></li><li><p>Kernel selection is <code>functools.lru_cache</code>d, so changing <code>block_m</code> in-process
without clearing those caches silently reuses the first decision and the test
passes vacuously. The helper clears them between runs.</p></li></ul><p></p><h2>Open question for maintainers (the actual root cause)</h2><p>I could not find the source-level cause, and the evidence suggests it is not in
the algorithm:</p><ul><li><p><code>csrc/py_itfs_ck/</code> (including <code>moe_ck_2stages_kernel.cu</code> and
<code>moe_ck_2stages_gemm_impl/</code>) is <strong>byte-identical</strong> between v0.1.13.post1 and
v0.1.18 — <code>diff -rq</code> reports no differences for the MoE kernels.</p></li><li><p><code>get_block_size_M</code> is byte-identical as well, so both versions select the same
tile for the same shape.</p></li><li><p>The <code>csrc/ck_gemm_moe_2stages_codegen/</code> directory differs only in
<code>gemm_moe_tune.py</code>.</p></li></ul><p>Identical sources therefore produce a working kernel in one build and a silent
no-op in the other, which points at codegen or build configuration rather than
the kernel logic. Someone familiar with the CK instantiation path can likely
find this quickly; the reproducer above should make it a short loop.</p><h2>Scope of validation</h2><p>Stated plainly so reviewers can judge:</p><ul><li><p>gfx942 only. gfx950 untested.</p></li><li><p>One shape (<code>E=128, topk=8, model_dim=4096, inter_dim=512</code>). Other shapes
untested.</p></li><li><p><code>block_m=16</code> untested — the clamp raises it to 64 on the assumption that a
smaller tile is at least as broken, which is unverified.</p></li><li><p>The clamp gives small batches a larger tile than the cost heuristic chose. A
decode-throughput A/B is in progress; numbers will be added here.</p></li></ul><p></p>
</body>
</html>

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -1983,6 +1983,25 @@ def get_2stage_cfgs(
             cfg_flat = run_1stage and bool(int(cfg["flat"]))
         else:
             cfg_flat = False
+    # WORKAROUND: the CK 2-stage MoE writes nothing at block_m=32 and returns an
+    # identically zero output tensor. Reproduced on gfx942 for bf16/QuantType.No
+    # and fp8/QuantType.per_Token; see op_tests/test_moe_blockm_invariance.py.
+    # The fault depends on block_m alone, not on the token count: forcing
+    # block_m=32 at token=64, a size that is otherwise correct, also yields zeros.
+    #
+    # The clamp sits after both the heuristic and the tuned-CSV branch because
+    # either can select 32, and 469 of the 617 rows in configs/tuned_fmoe.csv pin
+    # block_m=32 with run_1stage=0. 1stage and asm paths are unaffected.
+    if not run_1stage:
+        _min_block_m = int(os.environ.get("AITER_MIN_BLOCK_M", "64"))
+        if block_m < _min_block_m:
+            logger.warning(
+                f"[fused_moe] block_m={block_m} raised to {_min_block_m}: the CK "
+                "2-stage kernel emits all-zero output below 64. Set "
+                "AITER_MIN_BLOCK_M=32 to restore the original selection."
+            )
+            block_m = _min_block_m
+
     is_opus_cfg = cfg is not None and _opus_a8w4.is_opus_a8w4_stage2_kernel(
         cfg.get("kernelName2", "")
     )
--- a/op_tests/test_moe_blockm_invariance.py
+++ b/op_tests/test_moe_blockm_invariance.py
@@ -1,0 +1,135 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+"""fused_moe results must not depend on block_m.
+
+block_m is a tile-size / scheduling knob chosen by get_block_size_M or pinned by
+a tuned_fmoe.csv row. It must not change what the MoE computes, so the same
+inputs run at different block_m have to agree.
+
+This guards a regression where the CK 2-stage kernel silently wrote nothing at
+block_m=32 and fused_moe returned an identically zero tensor. It was invisible
+end to end because prefill uses larger tiles and stays correct while decode,
+which is where the small tiles are selected, quietly lost the entire routed
+expert contribution.
+
+Note the failure mode is an all-zero output rather than an exception or NaN, so
+a test that only checks for finiteness or for a successful launch will pass. The
+assertions below check non-zero output explicitly and compare across tiles.
+"""
+
+import pytest
+import torch
+
+import aiter
+from aiter import ActivationType, QuantType, dtypes
+from aiter import fused_moe as fm
+from aiter.ops.quant import pertoken_quant
+from aiter.test_common import checkAllclose
+
+# Tiles the 2-stage heuristic can emit (get_block_size_M support_list).
+BLOCK_SIZES = [32, 64, 128]
+REFERENCE_BLOCK_M = 128
+
+
+def _clear_dispatch_caches():
+    """Drop memoized dispatch decisions.
+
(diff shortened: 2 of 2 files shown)


zufayu:
## Review: PR #4451 — clamp block_m ≥64 to mask CK 2-stage all-zero bug

**⚠️ NEEDS WORK** — real bug, honest write-up, solid test, but it changes a core MoE contract via an over-reaching clamp + a rejected env knob, with downstream CI skipped and no perf/TP data.

- **Owner sign-off (don't self-merge):** `fused_moe.py` is a stable downstream contract. → @-mention the MoE owner (@valarLip) for explicit sign-off before merge.
- **Downstream CI skipped:** PR has no labels, so `Atom Test` / `Kimi Downstream` / `Sglang` don't run — yet this changes DSv3/Kimi decode MoE and can merge green, then break downstream. → Add `ci:all` (or `ci:atom`+`ci:vllm`) and require green.
- **Env-var knob:** `AITER_MIN_BLOCK_M` is a new permanent env flag — aiter rejects these (`AITER_MOE_FORCE_BF16_ACT` in #3593 was reverted by #4225), and here it re-enables a silently-corrupting path. → Drop it; derive the safe floor from dtype/arch/shape, or keep only as a clearly-temporary rollback tied to the root-cause fix.
- **Over-reach:** `if not run_1stage` also catches `per_1x128` decode (block_m 16→64, `fused_moe.py:1928-1947`) — the DSv3/Kimi fp8 block-scale path, contradicting "asm unaffected" and never validated. → Narrow to the CK 2-stage path or validate per_1x128; fix the comment.
- **Perf + shapes:** clamp overrides 469/617 tuned rows and the decode A/B is still pending; test is one toy shape (E=128/topk=8/4096/512), no TP=4/8 split, no prefill. → Post decode numbers and add production/TP shapes before merge.
- **Test gap:** covers only `No`/`per_Token`, not `per_1x128`; and with the fix, forced `block_m=32` is clamped before the kernel, so it tests the clamp, not the kernel at 32. → Add a per_1x128 case.
- **Root cause live:** CK all-zero bug unfixed; gfx950 untested. → Track the codegen root cause.

zjin-lcf:
Root cause found — and it's already fixed upstream in CK; this can be a CK-submodule bump instead of a block_m clamp.

  The all-zero output isn't a block_m=32 dispatch/tiling issue — it's a miscompile-free copy-paste bug in a CK refactor.

  Already fixed upstream. ROCm/composable_kernel develop commit f33252ce
  (https://github.com/ROCm/composable_kernel/commit/f33252cebe5a52362ec1ee12c124dde7800dda3a)

  That commit is a descendant of the CK submodule this branch pins (b675945), so this tree predates the fix — hence the all-zero decode.

  Suggested change. Bump the CK submodule to include f33252ce (or cherry-pick it) and drop the block_m≥64 clamp. This also covers the fp8-per-token and blockscale gate-up paths, which the clamp only helps incidentally.

  Verification (gfx950, standalone CK repro of the 32-tile gemm1 gate-up instance): device_absmax goes 0 → 302.25 with the fix; the block_m=128 control is unaffected.

jatseng-ai:
Closing this workaround PR — it is superseded.

`aiter` `main` already pins CK at the SHA that addresses the root cause (CK bump `f33252ce`, per review from @zjin-lcf). With that fix landed upstream, the `block_m=32` all-zero fused MoE output workaround here is no longer needed.

Thanks for the review and pointer to the proper fix!

PR #4463, opened 2026-07-30
Add an opt-in a4w4 SiTUv2 MoE path and fix three SiTUv2 tuner defects

XiaobingSuper:
## What

Adds a third activation dtype for SiTUv2 (Kimi-K3) MoE — **a4w4**, behind `AITER_SITUV2_A4W4=1` — next to the existing `AITER_SITUV2_A8W4`. **The default is unchanged (a16w4)**, so this is inert unless the flag is set.

```python
if activation == ActivationType.Situv2:
    if   AITER_SITUV2_A8W4 == "1": q_dtype_a = fp8      # a8w4  (existing)
    elif AITER_SITUV2_A4W4 == "1": q_dtype_a = fp4x2    # a4w4  (new)
    else:                          q_dtype_a = bf16     # a16w4 (default, unchanged)
```

The SiTUv2 test is hoisted above the Swiglu/INTERLEAVE branch so that branch cannot claim it and pick the activation dtype itself. All existing a16w4 routing and configs are left untouched.

## Performance

TP8 shape (3584 x 384, 896 experts, topk 16), tuner `us1 + us2`:

| token | a16w4 | a8w4 | **a4w4** | vs a16w4 | vs a8w4 |
|------:|------:|-----:|---------:|---------:|--------:|
| 1 | 28.3 | 20.1 | **20.0** | +29.2% | +0.4% |
| 4 | 42.7 | 33.7 | **32.9** | +23.1% | +2.5% |
| 16 | 107.3 | 93.0 | **91.2** | +15.0% | +1.9% |
| 64 | 260.6 | 229.1 | **225.9** | +13.3% | +1.4% |
| 256 | 374.7 | 329.8 | **325.3** | +13.2% | +1.4% |
| 1024 | 422.5 | 365.7 | **349.9** | +17.2% | +4.3% |
| 2048 | 651.3 | 475.8 | **428.2** | +34.2% | +10.0% |
| 4096 | 1087.3 | 774.4 | **653.2** | +39.9% | +15.7% |
| 8192 | 1994.4 | 1249.4 | **1020.3** | +48.8% | +18.3% |
| 16384 | 3600.3 | 2074.8 | **1716.5** | +52.3% | +17.3% |

a4w4 is ahead of a16w4 at every tier. Against a8w4 the small tiers land together — they are weight-bandwidth bound and the weights are fp4 either way, so the activation dtype only moves a few KB against tens of MB of weights — and the gap opens up once the A traffic and the MFMA rate start to matter.

Numbers are isolated kernel timings from the tuner, not end-to-end throughput.

## Three tuner defects

Each of these mis-scored candidates silently rather than failing loudly, and each had to be fixed before this shape could be tuned at all.

**1. SiTUv2 activation params never reached either side.** `torch_moe_stage1` fell back to its `(2.0, 1.5)` defaults while the FlyDSL kernels used `(1.0, 1.0)` — reference and kernel were evaluating **different activation functions**. Every SiTUv2 stage1 candidate came back at ~35% err, all were dropped by `--errRatio`, and the run aborted with:

```
Error: please check errRatio, stage1 and stage2 should be valid together!
```

Passing Kimi-K3's `(4.0, 25.0)` to both takes err1 from **33.4% → 0.0%**. This is not a4w4-specific: the existing `kimik3_fp4_tuned_fmoe.csv` (a16w4) carries `err1 ~34.7%` for the same reason.

**2. `s1_compare_fn` was only set for fp8.** Fused fp4 stage1 fell through to the generic comparison, which reads a 1-byte dtype as raw bytes. `fp4x2` packs two e2m1 codes per byte, so a low-nibble difference moves the byte by 1 and a high-nibble one by 16. All **4352** fused a4w4 candidates scored 100% err and none was ever selected — the tuner had been picking non-fused stage1 for a4w4 and paying for a separate quant kernel.

**3. `cosine_diff_compare` faults the GPU on fp4x2.** It starts with `ref.double()`, and

```python
torch.randint(0, 255, (64,), dtype=torch.uint8, device="cuda").view(dtypes.fp4x2).double()
# HIP error: unspecified launch failure
```

64 elements is enough. Enabling that comparison for fp4 in (2) therefore crashed workers and restarted the pool instead of scoring anything. Fixed by decoding the nibbles into their represented values, which also avoids the packed-byte comparison from (2).

Also: skip `k_batch` / `k_wave` splits that do not divide the K axis at enumeration time. `compile_mixed_moe_gemm1` rejects them anyway, but only after dispatch, and on some shapes the launch faults the queue rather than raising.

## Config

`kimik3_a4w4_{un,}tuned_fmoe.csv`, **17 token tiers (1 … 32768)**. All 17 select the fused stage1 kernel (previously 0, because of defect 2), and `err1`/`err2` are ≤0.3% everywhere.

The 32768 tier is included because `get_padded_M` maps everything in `(16384, 32768]` onto that key, and the tier fallback only fires strictly above it — without the row those batches silently drop to the hardcoded heuristic.

## Validation

gfx950 TP8. All 17 token tiers hit the tuned rows at runtime, and the server log shows the fused (`_fp4`) kernel names on every tier with no heuristic fallback.

GSM8K 3-shot under the CI settings (baseline 0.95, threshold 0.94), three runs: **0.9484 / 0.9484 / 0.9522**.


aiter/fused_moe.py
```diff
@@ -609,26 +609,26 @@ def fused_moe_(
         # mxfp8: both activation and weight are fp8 (per-1x32 e8m0 microscale).
         q_dtype_a = dtypes.fp8
     elif quant_type == QuantType.per_1x32:
-        if activation == ActivationType.Swiglu and gate_mode == GateMode.SEPARATED:
+        if activation == ActivationType.Situv2:
```
lalala-sh, line 717: Support situv2 in moe tunner is enough. Not neccessary here.

aiter/configs/model_configs/kimik3_a4w4_untuned_fmoe.csv
```diff
@@ -0,0 +1,17 @@
...
+64,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+128,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+256,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+512,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+1024,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+2048,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+4096,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+8192,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
+16384,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0
```
lalala-sh, line 17: For support all tuner step, plz add token 32768 config

lalala-sh (dismissed):

aiter/configs/model_configs/kimik3_a4w4_untuned_fmoe.csv
XiaobingSuper: added

lalala-sh (approved):

PR #4464, opened 2026-07-30
 [opus_moe] production A8W4 MoE stage1 kernels

yifehuan:
## Motivation

Deliver the production A8W4 (a8w4) MoE Stage1 Opus kernel family for DSV4 on
gfx950, closing the remaining gap against FlyDSL and giving the fused MoE path
a fully tuned, self-contained Stage1 that needs no model prepacking or
alternate weight layout. Also lands the Stage2 sort-block/scale-selection
tuning and a tuner timeout fix that this work depended on.

## Technical Details

**Stage1 kernels & scheduling** (`csrc/opus_moe/.../stage1/*`, `moe_stage1_a8w4.py`,
`gen_instances.py`, `fused_moe.py`):
- Generated `group_split` / `pair_kwave` / `pipeline` A8W4 Stage1 kernels with
  production dispatch rows for DSV4 — FlyDSL is selected for tokens 4–256 where
  it wins, Opus elsewhere.
- Token-dependent weight-access levers chosen per token by the tuner:
  stream (SLC/non-temporal) vs L2-cache weight loads when tokens/expert
  <= block_m, and XCD-swizzle M-major tile remap when weight is reused across
  M-tiles/expert. Measured stage1: stream @512 −13.7%, @1024 −9.4%; xcd @16384 −22%.
- Original-layout logical K-loop swizzling, route-affinity / M-fragment-major /
  K1-lead large-token schedules, native scaled-FP8 conversion, exact scale/output
  indexing, empty-fragment suppression, and a two-wave bm32 stream-K8 kernel for
  token 512. All data movement stays inside the kernel.

**Correctness & safety:**
- silu with `swiglu_limit=10` clamp (FlyDSL `silu_and_mul_fq` parity).
- Byte extents passed to `make_gmem` so the HW OOB bounds check stays active
  (the 4GB sentinel would otherwise disable it → OOB device-memory access).

**Accuracy:**
- gsm8k recovery on the `group_split` path (~0.90 → ~0.95) with no perf regression,
  by re-enabling LLVM's post-RA machine scheduler for `module_moe_opus` via
  `optCompilerConfig.json` (`-mllvm -enable-post-misched=1`; `core.py` untouched).
  post-misched only reorders issue/waitcnt (IEEE-preserving), so Stage1 output is
  bit-identical — the gain is a serving-trajectory basin shift, not a numerical change.

**Stage2 tuning** (`moe_stage2_a8w4_*`, `opus_moe_*_stage2_a8w4_*_gfx950.cuh`):
- Sort-block-aware Stage2 metadata/dispatch for the tuned A8W4 kernels; selects the
  correct packed MXFP8 scale half for BM16 routes; Opus tuner cases now use real
  MXFP8 values and sorted scales so the path is exercised during correctness checks.

**Tuner** (`mp_tuner.py`, `gemm_moe_tune.py`):
- Exclude queued time from the per-task timeout; reproducible routing-seed and
  kernel-regex controls.

## Test Plan

- Fused MoE correctness sweep across tokens 1–32768 for both randn and balanced
  routing on DSV4 A8W4.
- Strict fused large-token runs to confirm the expected kernels are selected
  without fallback/assertion/NaN.
- GPT-OSS A8W4 Swiglu BM16 path for the Stage2 changes.

## Test Result

- Stage1: 352/352 rows pass, max err 0.010294 (randn) / 0.012811 (balanced);
  large-token runs select expected kernels with no fallback, assertion, or NaN.
- Stage2 (GPT-OSS A8W4 Swiglu BM16): Stage1 cosine diff 0.004875, Stage2 cosine
  diff 0.000974.
image

## Submission Checklist


zufayu:
## Review: PR #4464 — [opus_moe] production A8W4 MoE stage1 kernels

**Adds 22 gfx950 A8W4 (fp8-act / fp4-weight) MoE stage1 kernels for DSV4, wires them into `fused_moe` dispatch (internal only — no public signature change), and lands stage2 sort-block tuning plus a tuner timeout fix.**

⚠️ **NEEDS WORK** — solid PR overall: full host-side validation, hard-asserts on unknown kid/shape (no silent fallthrough), correct opus1+opus2 wiring, fake fn present, contiguous inputs. No correctness-level bug found in the diff. Findings are one CI red-light on the target arch, a coverage gap, and a few hygiene/governance items.

🔴 **CI red on the target arch.** `Standard Tests (1 GPU) (linux-aiter-mi35x-1, MI35X, shard 7)` fails while every other shard, MI300X, OPUS Tests, and the accuracy suites pass. MI35X is gfx950 — the exact arch this PR ships kernels for — so this can't be treated as a flake. **Author must** post the failing step and diff it against `main` before merge.

⚠️ **The new opus stage1 kernels are not actually exercised by CI — but no new test file is needed.** `op_tests/test_moe_2stage.py` already covers the a8w4 fused_moe path against a torch reference (`checkAllclose` + assert, gfx950-gated, a8w4 case). However `aiter_test.sh` runs it as `python3 test_moe_2stage.py` at the default shape `(7168, 256)`, while opus stage1 is CSV-driven and only selected on the `(7168, 384)` rows — so these kernels run zero times in Standard Tests. **Author should** reuse the existing harness: add the DSV4 opus shape `(7168, 384)` + tokens to the default shape list, or run its model-csv mode over `dsv4_fp8fp4_tuned_fmoe.csv` (it already recognizes opus kernels). **Reviewer should** confirm whether any `ci:atom` model runs DSV4 a8w4-fp4 with this CSV; if not, coverage is nil.

⚠️ **`develop=True` on a production op.** `aiter/ops/opus/moe_stage1_a8w4.py` decorates the stage1 op with `@compile_ops(..., develop=True)`, which bypasses the JIT cache and contradicts the "production" framing. **Author must** remove it before shipping.

⚠️ **`make_gmem` byte extents are `uint32`.** In `opus_moe_host_impl.cuh` the `*_size_bytes` fields are `unsigned int` set from `size(0)*stride(0)`. The shipped DSV4 shape (w1 ≈ 0.7 GB) is safe, but any weight tensor ≥ 4 GB wraps, handing the HW bounds check a too-small extent → valid weight loads flagged OOB → silent wrong output. **Author must** widen to `uint64`/`size_t`, or add an explicit `AITER_CHECK(byte_size < 4 GiB)` so an oversized tensor fails loudly instead of silently.

⚠️ **`stage2_has_bias` now also true for `is_cktile2` (`fused_moe.py`).** This is the only change here touching a non-opus path (`stage2_has_bias = enable_bias and (is_flydsl2 or is_cktile2)`, was `is_flydsl2` only) — a cktile2 stage2 bias-behavior change bundled into a stage1 PR. **Reviewer should ask** that it is intentional and covered by the GPT-OSS A8W4 Swiglu BM16 stage2 test.

📝 **Stage2 decode LDS-staging threshold dropped 160 KB → 64 KB** (`opus_moe_traits_stage2_a8w4_decode_gfx950.cuh`), changing occupancy for already-shipped decode kernels not in the perf table. **Reviewer should ask** for a no-regression confirm on previously-tuned decode shapes.

📝 **`fused_moe.py` is a Tier-2 downstream contract.** This changes its default dispatch behavior (not its signature). On top of `ci:atom`, **@-mention the de-facto MoE owner (@valarLip)** for explicit sign-off rather than self-merging.

yifehuan:
CI

  > Fixed in c8147065. The failure was caused by the runtime swiglu_limit not reaching Opus Stage1. The latest MI35X shard-7 CI is green.

  CI coverage

  > test_moe_2stage.py already scans the merged tuned CSV by default; it is not limited to (7168, 256). The CSV filter has been extended to include opus_moe1_ rows, and all 9 production Opus Stage1 selections (tokens 1–32768)
  > pass strict accuracy checks.

  develop=True

  > In the current compile_ops implementation, develop=True does not bypass the JIT cache. It converts torch.Tensor to aiter_tensor_t and propagates the current HIP stream after module loading. It is required by the Opus ABI and
  > should remain.

  4 GiB extent

  > Added an explicit < UINT32_MAX host check before converting byte extents. Oversized tensors now fail before kernel launch. Widening only the karg would not help because make_gmem itself uses a 32-bit hardware extent. The
  > supported DSV4 W1 is about 1.06 GB.

  CKTile2 Stage2 bias

  > This is intentional and only affects CKTile2 Stage2 when bias is enabled; behavior is unchanged when bias is absent. It is covered by the GPT-OSS A8W4 SwiGLU validation and the full fused strict regression.


> ## Review: PR #4464 — [opus_moe] production A8W4 MoE stage1 kernels
> **Adds 22 gfx950 A8W4 (fp8-act / fp4-weight) MoE stage1 kernels for DSV4, wires them into `fused_moe` dispatch (internal only — no public signature change), and lands stage2 sort-block tuning plus a tuner timeout fix.**
> 
> ⚠️ **NEEDS WORK** — solid PR overall: full host-side validation, hard-asserts on unknown kid/shape (no silent fallthrough), correct opus1+opus2 wiring, fake fn present, contiguous inputs. No correctness-level bug found in the diff. Findings are one CI red-light on the target arch, a coverage gap, and a few hygiene/governance items.
> 
> 🔴 **CI red on the target arch.** `Standard Tests (1 GPU) (linux-aiter-mi35x-1, MI35X, shard 7)` fails while every other shard, MI300X, OPUS Tests, and the accuracy suites pass. MI35X is gfx950 — the exact arch this PR ships kernels for — so this can't be treated as a flake. **Author must** post the failing step and diff it against `main` before merge.
> 
> ⚠️ **The new opus stage1 kernels are not actually exercised by CI — but no new test file is needed.** `op_tests/test_moe_2stage.py` already covers the a8w4 fused_moe path against a torch reference (`checkAllclose` + assert, gfx950-gated, a8w4 case). However `aiter_test.sh` runs it as `python3 test_moe_2stage.py` at the default shape `(7168, 256)`, while opus stage1 is CSV-driven and only selected on the `(7168, 384)` rows — so these kernels run zero times in Standard Tests. **Author should** reuse the existing harness: add the DSV4 opus shape `(7168, 384)` + tokens to the default shape list, or run its model-csv mode over `dsv4_fp8fp4_tuned_fmoe.csv` (it already recognizes opus kernels). **Reviewer should** confirm whether any `ci:atom` model runs DSV4 a8w4-fp4 with this CSV; if not, coverage is nil.
> 
> ⚠️ **`develop=True` on a production op.** `aiter/ops/opus/moe_stage1_a8w4.py` decorates the stage1 op with `@compile_ops(..., develop=True)`, which bypasses the JIT cache and contradicts the "production" framing. **Author must** remove it before shipping.
> 
> ⚠️ **`make_gmem` byte extents are `uint32`.** In `opus_moe_host_impl.cuh` the `*_size_bytes` fields are `unsigned int` set from `size(0)*stride(0)`. The shipped DSV4 shape (w1 ≈ 0.7 GB) is safe, but any weight tensor ≥ 4 GB wraps, handing the HW bounds check a too-small extent → valid weight loads flagged OOB → silent wrong output. **Author must** widen to `uint64`/`size_t`, or add an explicit `AITER_CHECK(byte_size < 4 GiB)` so an oversized tensor fails loudly instead of silently.
> 
> ⚠️ **`stage2_has_bias` now also true for `is_cktile2` (`fused_moe.py`).** This is the only change here touching a non-opus path (`stage2_has_bias = enable_bias and (is_flydsl2 or is_cktile2)`, was `is_flydsl2` only) — a cktile2 stage2 bias-behavior change bundled into a stage1 PR. **Reviewer should ask** that it is intentional and covered by the GPT-OSS A8W4 Swiglu BM16 stage2 test.
> 
> 📝 **Stage2 decode LDS-staging threshold dropped 160 KB → 64 KB** (`opus_moe_traits_stage2_a8w4_decode_gfx950.cuh`), changing occupancy for already-shipped decode kernels not in the perf table. **Reviewer should ask** for a no-regression confirm on previously-tuned decode shapes.
> 
> 📝 **`fused_moe.py` is a Tier-2 downstream contract.** This changes its default dispatch behavior (not its signature). On top of `ci:atom`, **@-mention the de-facto MoE owner (@valarLip)** for explicit sign-off rather than self-merging.

junhaha666 (approved):

PR #4466, opened 2026-07-30
[NO MERGE] Add gfx942 FWD split kv

ipanfilo:
## Submission Checklist

### DIFF
--- a/csrc/cpp_itfs/mha_fwd.cu
+++ b/csrc/cpp_itfs/mha_fwd.cu
@@ -410,6 +410,46 @@ static int native_splitkv_heuristic(int batch, int nhead_q, int seqlen_q, int se
     return (kv_cap > 0) ? ((g < kv_cap) ? g : kv_cap) : 0;
 }
 
+bool native_splitkv_check_eligigibility(const mha_fwd_args& a)
+{
+    // Shape / geometry / scalar-config eligibility guards.
+    // Pointer-presence checks (q_descale_ptr, sink_ptr) are intentionally
+    // omitted here: the caller may not have allocated those buffers yet at
+    // planning time. fmha_fwd_native_gfx942() validates them.
+    if(get_gpu_arch() != "gfx942")
+        return false;
+    if(a.data_type != "bf16")
+        return false;
+    if(a.hdim_q != 64 || a.hdim_v != 64)
+        return false;
+    if(a.is_group_mode)  // varlen not supported
+        return false;
+    if(a.bias_type != 0)
+        return false;
+    if(a.p_drop > 0.f)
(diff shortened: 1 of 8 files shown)


csrc/cpp_itfs/mha_fwd.cu
```diff
@@ -364,10 +367,212 @@ float fmha_fwd_ck(mha_fwd_args a, const ck_tile::stream_config& s)
...
+    if(a.window_size_left != -1 || (a.window_size_right != 0 && a.window_size_right != -1))
+        return -1.f;
+    if(causal && a.seqlen_q > a.seqlen_k)
+        return -1.f;
+    if(a.mask_type == 1 && a.seqlen_q != a.seqlen_k)
+        return -1.f;
+    if(a.nhead_q <= 0 || a.nhead_k <= 0 || a.nhead_q % a.nhead_k != 0)
+        return -1.f;
+    if(a.num_splits <= 0 || !a.splitkv_workspace_ptr)
+        return -1.f;
```
Micky774: A fair bit of this seems redundant with the checks in `mha_fwd_calculate_num_splits` -- can we abstract them to keep them in sync, or are there runtime differences?

ipanfilo: mha_fwd_calculate_num_splits has subset of checks, it does not check for any buffers. Common checks can be moved to service function though

op_tests/test_mha_fwd_native_gfx942.py
```diff
@@ -0,0 +1,339 @@
...
+        "dropout_p": 0.0,
+        "softmax_scale": q.size(-1) ** -0.5,
+        "is_causal": causal,
+        "window_size_left": -1,
+        "window_size_right": 0 if causal else -1,
+        "sink_size": 0,
+    }
+
+
+def _workspace_bytes(num_splits: int, q: torch.Tensor, k: torch.Tensor) -> int:
```
wangye805: Do you think we can provide a workspace_bytes for split_kv as an API from aiter so qola/TE can use it instead of hardcoding the equations in qola/TE?

wangye805 (dismissed):

op_tests/test_mha_fwd_native_gfx942.py
ipanfilo: We could do it and it is better approach. We however have existing allocation size logic in TE so I thought to maybe review that from more general point of view than just add workspace_bytes for split-KV. So left it currently as-is

Micky774 (dismissed):

Micky774 (approved):

wangye805 (approved):

PR #3652, opened 2026-06-10
make allreduce_rmsnorm_quant support transpose_scale for bpreshuffle gemm

ganyi1996ppo:
## Motivation
make allreduce_rmsnorm_quant support transpose_scale for bpreshuffle gemm

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4478, opened 2026-07-31
[dist] forward transpose_scale through fused AR+RMSNorm per-group quant

yichiche:
## Summary

- Forward the existing `transpose_scale` flag through the legacy `CustomAllreduce` fused AR + RMSNorm + per-group-FP8-quant Python path so callers can ask the kernel to emit the per-group scale directly in the column-major (bpreshuffle) layout.
- The C++ op and pybind binding already support `transpose_scale` (from #3652); only the Python `CustomAllreduce` methods and the ops stub were missing the parameter, so this path always emitted a row-major scale and forced consumers to run a separate transpose.

## Details

`gemm_a8w8_blockscale_preshuffle` on gfx95 consumes the per-group activation scale in a column-major layout: storage `(num_groups, M)` viewed transposed to `(M, num_groups)` with stride `(1, M)`. The fused `fused_allreduce_rmsnorm_quant_per_group` C++ kernel already has a compile-time `TRANSPOSE_SCALE` template that writes exactly that layout at no extra cost, and the pybind binding exposes `transpose_scale` (default `false`). However, the Python wrappers that most callers use — `CustomAllreduce.custom_fused_ar_rms_per_group_quant` and `CustomAllreduce.fused_ar_rms_per_group_quant` — did not accept or forward the flag (the ops stub in `aiter/ops/custom_all_reduce.py` also omitted it), so this path could only ever produce a row-major scale. Callers therefore had to append an explicit transpose (`scale.t().contiguous().t()`), launching a redundant copy kernel per call.

Changes:

- `aiter/ops/custom_all_reduce.py`: add `transpose_scale: bool = False` to the `fused_allreduce_rmsnorm_quant_per_group` op stub (the compiled binding already accepts it).
- `aiter/dist/device_communicators/custom_all_reduce.py`: `fused_ar_rms_per_group_quant` now accepts `transpose_scale`, allocates `scale_out` with the matching column-major stride (`torch.empty((num_groups, M)).transpose(0, 1)`) when set, and forwards the flag to the C++ op; `custom_fused_ar_rms_per_group_quant` accepts and forwards it, including allocating the CUDA-graph capture-mode dummy scale with the same stride so capture and replay layouts agree.

This is purely additive and backward-compatible: `transpose_scale` defaults to `false`, so existing callers are unchanged. When set to `true` the returned scale is byte-for-byte the column-major layout the bpreshuffle GEMM consumes (identical to a post-hoc `.t().contiguous().t()`), letting downstream callers (e.g. SGLang's fused AR + RMSNorm + per-group quant path for Qwen3.5 FP8) drop their redundant transpose kernel.

Validated end to end in SGLang on Qwen3.5-397B-A17B-FP8, TP4 (MI35x): gsm8k 0.955 (numerics-preserving), the per-call transpose copy kernel eliminated (confirmed in a conc4 trace: the fused kernel flips to the `TRANSPOSE_SCALE=true` template variant and the extra copy kernels disappear one-for-one), with a small end-to-end decode-latency / throughput improvement.

### DIFF
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -1618,6 +1618,7 @@ class CustomAllreduce:
         registered: bool = False,
         use_1stage: bool = False,
         emit_bf16: bool = False,
+        transpose_scale: bool = False,
     ):
         K = inp.shape[-1]
         # Fail fast on bad ``group_size`` at the Python boundary. Mirrors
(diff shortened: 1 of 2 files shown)


TennyWang1223 (approved):
LGTM

yzhou103 (approved):

PR #4482, opened 2026-07-31
gfx1250: fix the grouped-MoE expert scan above 512 experts, and compute SiTUv2 instead of SiLU

XiaobingSuper:
## What this fixes

Two independent bugs in the gfx1250 grouped MoE, both found while bringing up
Kimi-K3 (`hidden_act="situ"`, 896 experts, top-16, MXFP4). Either one alone
makes the model unservable, and neither is K3-specific — the first is wrong for
any model with more than 512 experts, the second for any model using SiTUv2.

They are separate commits and can be reviewed independently.

---

### 1. The contiguous-M prefix scan silently dropped every expert past 512

`moe_contiguous_psum.py` ran the tile-aligned prefix scan as a **single block of
`MAX_EXPERTS_PER_BLOCK = 512` threads, one thread per expert**:

```python
in_expert = tid < fx.Uint32(experts)   # tid < 512 always
if in_expert:
    m = buffer_load(m_rsrc, tid, ...)  # expert index == thread index
```

With `E > 512` the experts above the block width are never visited. `starts`,
`psum` and `contiguous_m` are allocated with `torch.empty`, so those entries keep
whatever was in the allocator's memory; `is_last = tid == experts - 1` never
fires, so `contiguous_m` is not written at all. The remap phase then reads
`starts[expert]` for the unscanned experts and writes `start + slot` back as the
route's row, and the next kernel to dereference that row —
`moe_fused_quant_preshuffle_routeks_*` — walks off the end of the contiguous
buffer:

```
Memory access fault by GPU node-N ... kernel: moe_fused_quant_preshuffle_routeks_fd3584_r4_fp4_pk8_srctk16
```

The 512 limit was known for the *fused* route+psum variant, which is gated on
`_FUSED_ROUTE_PSUM_MAX_EXPERTS = 512` in `grouped_moe_gfx1250.py`.
`contiguous_psum` / `contiguous_psum_remap` had no equivalent guard, so instead
of falling back they corrupted quietly.

**Fix:** sweep the experts in block-sized chunks and carry the running offset
between chunks in LDS. E is no longer bounded by the block width.

Three details worth a reviewer's attention:

- The carry lives in LDS rather than a register because the chunk loop is a
  *runtime* loop (`experts` is a kernel argument) — a Python-level accumulator
  does not survive it.
- Lanes past `experts` now write `0` into the scan buffer and participate in the
  Hillis-Steele steps, instead of being skipped by `if in_range`. That is needed
  so the last lane still holds the chunk total regardless of where `experts`
  falls inside the chunk.
- The sweep is written out in both kernels rather than factored into a helper.
  `@flyc.kernel` only AST-transforms the decorated body, so a dynamic `for`/`if`
  moved into a plain function stops being traced (it raises
  `dynamic 'ArithValue' has no Python integer representation`).

### 2. SiTUv2 was computed as SiLU

The TDM stage1 epilogue only encoded silu and swiglu:

```python
stage1_act = 2 if activation == ActivationType.Swiglu else 1   # Situv2 -> 1 -> silu
```

`ActivationType.Situv2` fell through to the silu code, and `situ_beta` /
`situ_linear_beta` were accepted at `grouped_gemm_gfx1250_a8w4`'s signature and
then dropped on the way to `_grouped_a8w4_tdm_moe`. No error, no fallback — just
the wrong activation. This is the `TODO(situv2)` left behind when the fused
stage1 epilogue was removed in the MoE refactor (#4394).

On gfx1250 this is the only SiTUv2 path there is: `GateMode.SEPARATED` routes
SiTUv2 to `flydsl_moe1_afp4_wfp4_bf16_*`, which currently fails to build there
(`LLVM ERROR: Do not know how to expand this operator's operand`, an i64 operand
in `llvm.amdgcn.raw.ptr.buffer.load.lds`) — that is a separate issue, not
addressed here.

**Fix:** add `stage1_act = 3` and wire it through both epilogues — the batched
one used by a8w4's fused-quant path, and the element-wise one used by a4w4's
bf16 intermediate.

Implementation notes:

- `beta` / `linear_beta` are **runtime** kernel arguments, so every SiTUv2 shape
  shares one compiled kernel rather than specializing per beta value. Their
  reciprocals are taken on the host so folding them into the per-element
  multipliers stays exact instead of going through an in-kernel `v_rcp_f32`, and
  the multipliers are hoisted out of the inner loop.
- `tanh` uses the saturating identity `2*sigmoid(2z) - 1` rather than the
  `(1-e)/(1+e)` form used elsewhere in the tree: `exp2` of a large positive
  argument goes to `+inf` and `rcp(+inf)` to `0`, so both tails come out right
  without an `|x|` fixup and a sign `select`. Costs 3 exp2 + 3 rcp per element.
- SiTUv2 is bounded by construction, so the swiglu clamp does not apply to it.

---

## Tests

`op_tests/test_flydsl_grouped_gemm_gfx1250.py`:

- **New:** `test_contiguous_psum_matches_cumsum` and
  `test_contiguous_psum_remap_rows_stay_in_bounds`, parametrized over
  `E = 8/256/512/513/896/1024` — either side of `MAX_EXPERTS_PER_BLOCK`. Counts
  come from a real unbalanced random routing, checked against a tile-aligned
  `torch.cumsum`. The second test asserts every remapped row lands inside the
  contiguous buffer, which is the property whose violation causes the fault.
- **Un-skipped:** `test_grouped_a4w4_situv2_matches_torch_ref`, which was marked
  `skip` with "the TDM path runs it as silu".
- **New:** `test_grouped_a8w4_situv2_matches_torch_ref`, since a8w4 takes the
  batched fused-quant epilogue rather than a4w4's element-wise one.
- `--act situv2` added to the CLI, with `--situ-beta` / `--situ-linear-beta`.

The scan tests fail on `main` exactly where they should, and pass here:

```
$ pytest op_tests/test_flydsl_grouped_gemm_gfx1250.py -k contiguous_psum   # on main
FAILED test_contiguous_psum_matches_cumsum[513]
FAILED test_contiguous_psum_matches_cumsum[896]
FAILED test_contiguous_psum_matches_cumsum[1024]
FAILED test_contiguous_psum_remap_rows_stay_in_bounds[513]
FAILED test_contiguous_psum_remap_rows_stay_in_bounds[896]
FAILED test_contiguous_psum_remap_rows_stay_in_bounds[1024]
6 failed, 6 passed
```

Activation correctness against the fp32 torch reference (gate is
`logits_diff < 0.01`):

| case | logits_diff | rel_l2 |
|---|---|---|
| a4w4 situv2 | 3.8e-06 | 2.8e-03 |
| a8w4 situv2 | 5.3e-06 | 3.3e-03 |
| a4w4 silu (regression) | 3.7e-06 | 2.7e-03 |
| a4w4 swiglu (regression) | 3.8e-06 | 2.8e-03 |

`rel_l2 ~3e-3` is MXFP4 quantisation noise; it is the same for all four.

## Performance

The SiTUv2 epilogue is `const_expr`-gated, so silu/swiglu codegen is unchanged —
their `logits_diff` / `rel_l2` above are bit-identical between `main` and this
branch. The two things that *are* shared are the scan and the four extra runtime
f32 kernel arguments, so both were measured.

Grouped MoE, `E=256 topk=8 model_dim=7168 inter_dim=512`, gfx1250:

| case | tokens | main (us) | this PR (us) |
|---|---|---|---|
| silu, gemm1 | 128 | 146.39 | 145.96 |
| silu, gemm2 | 128 | 111.82 | 113.05 |
| silu, gemm1 | 1024 | 149.41 | 149.45 |
| silu, gemm2 | 1024 | 130.98 | 130.76 |
| silu, end-to-end | 128 | 289.88 | 292.68 |
| silu, end-to-end | 1024 | 370.51 | 362.87 |
| swiglu, gemm1 | 128 | 146.19 | 146.65 |
| swiglu, gemm2 | 128 | 112.27 | 112.95 |
| swiglu, gemm1 | 1024 | 149.64 | 149.97 |
| swiglu, gemm2 | 1024 | 130.33 | 131.27 |
| swiglu, end-to-end | 128 | 293.82 | 290.66 |
| swiglu, end-to-end | 1024 | 360.58 | 362.80 |

Everything is within ±2% with both signs, i.e. run-to-run noise.

The scan kernels timed on their own, which is where a real cost would show:

| E | psum main | psum PR | remap main | remap PR |
|---|---|---|---|---|
| 8 | 2.47 | 2.64 | 5.85 | 5.76 |
| 128 | 2.48 | 2.69 | 6.74 | 6.58 |
| 256 | 2.58 | 2.67 | 6.44 | 6.86 |
| 512 | 2.62 | 2.77 | 6.88 | 6.65 |
| 896 | faults | 4.32 | faults | 8.20 |

`contiguous_psum_remap` — the one the grouped MoE actually calls — is within
noise. `contiguous_psum` is consistently ~0.15 us slower (+3.5..8.5% on a 2.6 us
kernel), which is the two extra barriers and the LDS carry round-trip on a
single-chunk sweep. Reported rather than hidden: it is once per MoE call, so for
a 93-layer model that is ~14 us per forward against a ~27 ms decode step, but a
reviewer may prefer a compile-time single-chunk specialisation instead.

## End to end

Kimi-K3 on 4×MI450 (tp4) through ATOM, full 1319-question GSM8K 5-shot,
`num_concurrent=8`:

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9591|±  |0.0055|
|     |       |strict-match    |     5|exact_match|↑  |0.9591|±  |0.0055|
```

Same score as the last recorded run of this model on gfx1250, which needed a
model-side workaround (sub-batching the routed MoE to <=128 tokens) to avoid the
fault in 1 — that workaround is removed on the ATOM side now that the scan is
correct.

### DIFF
--- a/aiter/ops/flydsl/kernels/moe_contiguous_psum.py
+++ b/aiter/ops/flydsl/kernels/moe_contiguous_psum.py
@@ -6,6 +6,10 @@
 Computes tile-aligned exclusive prefix sum of per-expert counts for the
 contiguous grouped-GEMM scheduler. Single-block parallel scan replaces
 torch.cumsum (avoids rocprim trampoline overhead for small E).
+
+The block is ``MAX_EXPERTS_PER_BLOCK`` threads wide but E is not bounded by it:
+the scan sweeps the experts in block-sized chunks and carries the running offset
+between chunks in LDS. Kimi-K3 (E=896) is the first model to exceed one chunk.
 """
 
 import flydsl.compiler as flyc
@@ -29,12 +33,14 @@ class _PsumStorage:
     """LDS for the prefix-scan kernels.
 
     ``lds0``/``lds1`` are ping-pong buffers: each Hillis-Steele step reads one
-    and writes the other, then the two swap. The trailing 16 is the byte
-    alignment of each array.
+    and writes the other, then the two swap. ``carry`` accumulates the total of
+    the chunks already scanned, so an E wider than the block still gets one
+    continuous prefix sum. The trailing 16 is the byte alignment of each array.
     """
 
     lds0: fx.Array[fx.Int32, MAX_EXPERTS_PER_BLOCK, 16]
     lds1: fx.Array[fx.Int32, MAX_EXPERTS_PER_BLOCK, 16]
+    carry: fx.Array[fx.Int32, 1, 16]
 
 
 @fx.struct
@@ -58,6 +64,11 @@ def _lds_load(ptr, idx):
 def _lds_store(ptr, val, idx):
     """Scalar i32 store to an LDS pointer at element offset ``idx``."""
     fx.ptr_store(val, ptr + fx.Int64(idx))
+
+
+# The chunked scan below is written out in both kernels rather than shared:
+# @flyc.kernel AST-transforms only the decorated body, so a dynamic `for`/`if`
+# does not survive being factored into a plain helper.
 
 
 def build_moe_contiguous_psum_module():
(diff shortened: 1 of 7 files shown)


aiter/ops/flydsl/kernels/mxfp4_preshuffle_gfx1250_tdm.py
```diff
@@ -62,6 +65,10 @@ def launch_gemm_a8w4_tdm(
     stage1_quant_out: Constexpr[int] = 0,
     quant_wmma_rep: Constexpr[int] = 1,
     arg_quant_scale: fx.Tensor = None,
+    f32_situ_beta: fx.Float32 = 1.0,
+    f32_situ_inv_beta: fx.Float32 = 1.0,
+    f32_situ_linear_beta: fx.Float32 = 1.0,
```
yadaish, line 65: what about removing inv args and get rcp in activation implement?

aiter/ops/flydsl/kernels/moe_contiguous_psum.py
```diff
@@ -86,50 +97,69 @@ def psum_kernel(
         lds = fx.SharedAllocator().allocate(_PsumStorage).peek()
         lds0 = lds.lds0.ptr
         lds1 = lds.lds1.ptr
+        carry = lds.carry.ptr
```
yadaish, line 100: a little bit hard to understand, why we need extra carry in current prefix sum? does current implement will break kimi k3, right?

aiter/ops/flydsl/kernels/mxfp4_preshuffle_gfx1250_tdm.py
XiaobingSuper: Good call — done in a87363bb5.

`f32_situ_inv_beta` / `f32_situ_inv_linear_beta` are gone; `situv2_consts()` now takes both reciprocals with `v_rcp_f32`. Both are uniform across the tile, so this costs two extra VALU ops per kernel launch, hoisted out of the inner loop — against two fewer kernel args and, more usefully, no way for a caller to hand in a beta and a reciprocal that disagree. The host-side `1.0 / float(...)` in `batched_gemm_mxfp4.py` is gone too.

I re-ran both epilogues on the Kimi-K3 betas to confirm `v_rcp_f32`'s accuracy is not a factor here:

| | logits_diff | rel_l2 |
|---|---|---|
| a4w4 (bf16 intermediate) | 9.0759e-06 | 4.2605e-03 |
| a8w4 (fused quant) | 4.4326e-06 | 2.9774e-03 |

Identical to the host-reciprocal build to every printed digit. `beta=4.0` is a power of two so its reciprocal is exact either way; `linear_beta=25.0` is not, and its ~1 ulp lands far below the MXFP4 quantisation it feeds. silu/swiglu re-checked unchanged on both formats.

aiter/ops/flydsl/kernels/moe_contiguous_psum.py
XiaobingSuper: Yes — current `main` does break Kimi-K3, and that is exactly what the carry is for.

One Hillis-Steele pass covers one expert per lane, so a single-pass scan tops out at `MAX_EXPERTS_PER_BLOCK` = 512 experts. Kimi-K3 has 896. On `main`, experts 512..895 simply never get `starts`/`psum` written — the GEMM then indexes with uninitialised memory, which is the memory fault we were chasing.

So the fix sweeps E in block-sized chunks. Each chunk scans as before, then adds `carry`, the tile-aligned total of every chunk already scanned, which is what stitches the per-chunk scans into one continuous prefix sum. `carry` has to live in LDS rather than a register because lane 0 produces it and all lanes consume it on the next iteration.

The new tests in `op_tests/test_flydsl_grouped_gemm_gfx1250.py` pin this down — they run E = 8, 256, 512, 513, 896, 1024 against a torch reference. Against `main` they give 6 failed / 6 passed, failing at exactly 513/896/1024 and passing at 8/256/512.

While here I swept the other single-block scans for the same cap:
- `moe_fused_route_quant_scatter` — its prefix sum is single-thread serial over E, no cap. Fine.
- `moe_g2l_lut` — gated at `_G2L_MAX_N` with a torch fallback. Fine.
- `moe_route_psum_fused` — genuinely capped, and differently: its LDS route counter is one slot per expert, so E > 512 needs a wider allocation, not a carry. `_FUSED_ROUTE_PSUM_MAX_EXPERTS` was defined but never actually enforced, so it now raises instead of silently dropping the experts past the block. Its `_NUMEL` companion is left advisory on purpose — that sweep is grid-stride, so a larger count is correct, just not worth fusing.

yadaish:
LGTM

coderfeli (dismissed):
LGTM

coderfeli (approved):

PR #4491, opened 2026-07-31
[HIP] Add gfx950 packed BF16 GDR decode kernel

zijiecode:
## Motivation

Qwen3.5 decode uses a Gated Delta Rule recurrence over packed Q/K/V and a persistent BF16 state. This PR adds a single-launch gfx950 path that consumes packed QKV directly and updates the recurrent state in place. It does not add QKV split/copy, transpose, temporary state tensors, or helper kernels.

## Modifications

- Add a fixed-shape packed BF16 decode path for 8 Q/K heads, 32 V heads, K/V dimension 128, and V-major state `[pool, 32, 128, 128]`.
- Follow Aiter's plain-HIP layout: `csrc/include/gdr_decode_packed_bf16.h` declares the host API, `csrc/kernels/gdr_decode_packed_bf16.cu` contains the kernel and launcher, and `csrc/pybind/gdr_decode_packed_bf16_pybind.cu` is the thin binding shell. The public Python wrapper is `aiter.gdr_decode_packed_bf16` from `aiter/ops/gdr_decode_packed_bf16.py`.
- Use Aiter's normal `compile_ops` flow, preserve the existing Torch host ABI, and update the state pool in place.
- Treat negative and positive out-of-range `state_idx` values as invalid rows. Invalid rows write positive-zero output and do not update state.
- Partition each value head into four V32 tiles, use four Wave64s per CTA with eight-lane K-reduction groups, and combine 16-byte vector I/O, `ds_swizzle` reductions, and non-temporal state load/store.
- Add upstream `op_tests` coverage for accuracy, state aliasing, untouched slots, invalid indices, mixed and strided rows, and wrapper validation.
- Document the state-index contract: all non-negative indices must be unique within a batch and smaller than the state pool; negative sentinel values may repeat. Duplicate valid indices have undefined behavior because state updates are unsynchronized.

## Accuracy and Correctness

The upstream focused op test reports `8 passed` on MI355X (`gfx950`), including a 16-step persistent-state drift case. Independent reconciliation checks against the Triton reference also passed at BS4 and BS128 with tighter `rtol=1e-2, atol=1e-3`:

- Maximum output absolute error: `1.52587890625e-5`
- Maximum BF16 state absolute error in the fresh BS4/BS128 rerun: `2.44140625e-4`
- Maximum 16-step output/state absolute error: `1.52587890625e-5` / `2.44140625e-4`
- State output aliases the input state
- Invalid indices write positive-zero output without modifying state
- Untouched slots remain bitwise unchanged

The module-local `--offload-arch=gfx950` flag was removed. A fresh normal-flow gfx942 build compiled both `gdr_decode_packed_bf16_pybind.cu` and `gdr_decode_packed_bf16.cu` with `--offload-arch=gfx942` and produced `module_gdr_decode_packed_bf16.so`; the target build contained no gfx950 offload flag. Runtime dispatch remains gfx950-only, so this fixes gfx942 wheel/prebuild compatibility without exposing an unsupported gfx942 launch path.

The placement-only refactor was also checked against the pre-refactor base. After normalizing the intentional `gdn` to `gdr` symbol rename, all 481 device ISA lines and the AMDHSA metadata are identical: 65 VGPR, 41 SGPR, zero LDS, zero scratch, and zero spills. Top-level `import aiter` exposes `gdr_decode_packed_bf16`, while a non-gfx950 call fails before JIT launch.

## Kernel Benchmark

MI355X (`gfx950`), ROCm 7.2, physical GPU0, benchmark source commit `85b02e25b8511e6e97aa7667ecc1c9f727e01d7d`, Aiter main comparison head `00b271b95f8a3405fa211e509c63441040557305`, and FlyDSL 0.3.0. Every implementation uses Aiter `test_common.perftest(num_iters=101, num_warmup=2)`, which aggregates GPU device time with `torch.profiler`. Each cell is the median of five fresh Python processes; no external profiler such as rocprof is active.

| BS | Triton packed BF16 | This PR packed BF16 HIP | FlyDSL tuned BF16 | HIP vs Triton | HIP vs FlyDSL tuned |
| 4 | 6.349 us | **5.029 us** | 6.178 us | **1.263x** | **1.229x** |
| 8 | 8.167 us | **6.219 us** | 7.901 us | **1.313x** | **1.270x** |
| 16 | 11.886 us | **8.871 us** | 10.553 us | **1.340x** | **1.190x** |
| 32 | 18.255 us | **14.174 us** | 16.974 us | **1.288x** | **1.198x** |
| 64 | 32.033 us | **24.685 us** | 30.875 us | **1.298x** | **1.251x** |
| 128 | 57.007 us | **49.430 us** | 59.306 us | **1.153x** | **1.200x** |

## Scope

- gfx950 and the exact packed BF16 contract only.
- Prefill/extend, ReplaySSM, other shapes/layouts, and other architectures remain on their existing paths.
- Non-negative `state_idx` values are required to be unique within a batch and smaller than the state-pool size; negative sentinel values may repeat. Duplicate valid indices result in undefined behavior because independent workgroups race on the in-place state update.

## Submission checklist


stevenshenyj:
- please add upstream op_tests coverage;
- Is runtime compilation via cpp_extension.load intentional, or should this kernel use Aiter's normal compile_ops/wheel build flow?
- Is a non-negative state_idx assumed to always be smaller than the state pool size, or should an upper-bound check be added?

zijiecode:
> * please add upstream op_tests coverage;
> * Is runtime compilation via cpp_extension.load intentional, or should this kernel use Aiter's normal compile_ops/wheel build flow?
> * Is a non-negative state_idx assumed to always be smaller than the state pool size, or should an upper-bound check be added?

Thank you for the review! I have updated the code:
 * Added upstream `op_tests/test_gdn_decode_packed_bf16.py` coverage. It reports `7 passed` on gfx950 and covers valid accuracy, output/state aliasing, untouched slots, negative indices, positive OOB indices, mixed/strided rows, and wrapper validation.
* The kernel now uses Aiter's normal `compile_ops` flow, with the source under `csrc/kernels` and a module entry in `optCompilerConfig.json`. I also built a wheel, installed it into an isolated target, and verified a fresh JIT build from `aiter_meta/csrc/kernels/gdn_decode_packed_bf16.cu`.
* An upper-bound check of `state_idx >= state_pool_size` is added.

stevenshenyj:
LGTM.

yiijin (changes requested):
Thanks for the work here. Before this lands I'd like to reconcile it with the two
GDN decode paths we already have.

### Existing implementations

- `csrc/kernels/fused_split_gdr_update.cu` (+ `csrc/include/`, `csrc/pybind/`,
  `aiter/ops/fused_split_gdr_update.py`) — already reads packed `mixed_qkv`, fuses
  split + gating + L2norm + recurrence, updates the pool in place. FP32 swizzled
  state, supports `T>1`.
- `aiter/ops/flydsl/kernels/gdr_decode.py` (+ `gdr_decode_tuned.csv`) — same
  recurrence, already supports BF16 V-major state, and the tuned CSV already covers
  this exact shape (gfx950 / bf16 state / 8 QK / 32 V / 128×128).

### Comparison with the existing kernels
The Triton numbers look solid — could you add `fused_split_gdr_update` and the FlyDSL 
kernel to the table? `gdr_decode.py` at `NUM_BLOCKS_PER_V_DIM=4, NUM_WARPS=4,
WARP_THREADS_K=8` lands on the same tiling and shares most of the same tricks, so I'd
like to see where the gap actually is. If the wins are the packed-QKV reads, the
non-temporal hints and the bounds check, those look foldable into the existing kernels —
happy to be corrected.
Worth isolating separately: much of the margin over `fused_split_gdr_update` comes from
BF16 state halving state traffic (BS=128: 268 vs 537 MB), which is an upstream dtype
choice. A BF16-state baseline would show this kernel's own contribution.


### Code

1. Duplicate `state_idx` within a batch races on the state write; the reference
   iterates serially and the tests only use unique indices. Please document the
   uniqueness precondition or check it.
2. Invalid rows write `+0.0` and skip, whereas `fused_split_gdr_update.cu:142` and
   `gdr_decode.py:372` only test `idx >= 0` and still emit a computed output. These
   need reconciling, or switching backends in SGLang will show unexplained diffs.
3. `--offload-arch=gfx950` is hardcoded in `flags_extra_hip`. Please confirm the
   module is excluded from the prebuild set, or gfx942 builds will emit gfx950 code.
4. `test_gdn_decode_packed_bf16.py:174` — `rtol=atol=0.05` against reported max errors
   of 1.5e-5 / 2.4e-4 won't catch a regression. BF16 state is also re-rounded every
   step, so please add a multi-step drift test.
5. `T=1` only — any plan for MTP / speculative decode? Both existing kernels loop over
   `T` with the state resident in registers, saving the extra state round-trips.


### Placement

Please follow the layout the other HIP ops use instead of adding `aiter/ops/hip/`.
Subdirectories in `aiter/ops/` are per programming model (`triton/`, `flydsl/`,
`opus/`, `torch_ref/`); plain HIP is the default backend and lives flat.
`fused_split_gdr_update` and `chunk_gated_delta_rule_fwd_h` are both good references:

```
csrc/include/gdr_decode_packed_bf16.h            # host declaration
csrc/kernels/gdr_decode_packed_bf16.cu           # kernel + host launch only
csrc/include/rocm_ops.hpp                        # + GDR_DECODE_PACKED_BF16_PYBIND macro
csrc/pybind/gdr_decode_packed_bf16_pybind.cu     # thin shell invoking the macro
aiter/ops/gdr_decode_packed_bf16.py              # drop the _hip suffix
aiter/__init__.py                                # + from .ops.gdr_decode_packed_bf16 import *
```

The `m.def` belongs in a `_PYBIND` macro in `csrc/include/rocm_ops.hpp` (111 ops are
registered there today) rather than in the kernel `.cu` — that file is where all op
signatures are reviewed. Right now the host function is also unreachable from other
TUs inside an anonymous namespace.

Also, please use `gdr` over `gdn` to match `fused_split_gdr_update`, `gdr_decode` and
`module_chunk_gdr_fwd_h`. If the missing top-level export was to avoid an import
failure on non-gfx950, `aiter/ops/opus/__init__.py` has the guard-and-stub pattern.

zijiecode:
> Thanks for the work here. Before this lands I'd like to reconcile it with the two GDN decode paths we already have.
> 
> ### Existing implementations
> * `csrc/kernels/fused_split_gdr_update.cu` (+ `csrc/include/`, `csrc/pybind/`,
>   `aiter/ops/fused_split_gdr_update.py`) — already reads packed `mixed_qkv`, fuses
>   split + gating + L2norm + recurrence, updates the pool in place. FP32 swizzled
>   state, supports `T>1`.
> * `aiter/ops/flydsl/kernels/gdr_decode.py` (+ `gdr_decode_tuned.csv`) — same
>   recurrence, already supports BF16 V-major state, and the tuned CSV already covers
>   this exact shape (gfx950 / bf16 state / 8 QK / 32 V / 128×128).
> 
> ### Comparison with the existing kernels
> The Triton numbers look solid — could you add `fused_split_gdr_update` and the FlyDSL kernel to the table? `gdr_decode.py` at `NUM_BLOCKS_PER_V_DIM=4, NUM_WARPS=4, WARP_THREADS_K=8` lands on the same tiling and shares most of the same tricks, so I'd like to see where the gap actually is. If the wins are the packed-QKV reads, the non-temporal hints and the bounds check, those look foldable into the existing kernels — happy to be corrected. Worth isolating separately: much of the margin over `fused_split_gdr_update` comes from BF16 state halving state traffic (BS=128: 268 vs 537 MB), which is an upstream dtype choice. A BF16-state baseline would show this kernel's own contribution.
> 
> ### Code
> 1. Duplicate `state_idx` within a batch races on the state write; the reference
>    iterates serially and the tests only use unique indices. Please document the
>    uniqueness precondition or check it.
> 2. Invalid rows write `+0.0` and skip, whereas `fused_split_gdr_update.cu:142` and
>    `gdr_decode.py:372` only test `idx >= 0` and still emit a computed output. These
>    need reconciling, or switching backends in SGLang will show unexplained diffs.
> 3. `--offload-arch=gfx950` is hardcoded in `flags_extra_hip`. Please confirm the
>    module is excluded from the prebuild set, or gfx942 builds will emit gfx950 code.
> 4. `test_gdn_decode_packed_bf16.py:174` — `rtol=atol=0.05` against reported max errors
>    of 1.5e-5 / 2.4e-4 won't catch a regression. BF16 state is also re-rounded every
>    step, so please add a multi-step drift test.
> 5. `T=1` only — any plan for MTP / speculative decode? Both existing kernels loop over
>    `T` with the state resident in registers, saving the extra state round-trips.
> 
> ### Placement
> Please follow the layout the other HIP ops use instead of adding `aiter/ops/hip/`. Subdirectories in `aiter/ops/` are per programming model (`triton/`, `flydsl/`, `opus/`, `torch_ref/`); plain HIP is the default backend and lives flat. `fused_split_gdr_update` and `chunk_gated_delta_rule_fwd_h` are both good references:
> 
> ```
> csrc/include/gdr_decode_packed_bf16.h            # host declaration
> csrc/kernels/gdr_decode_packed_bf16.cu           # kernel + host launch only
> csrc/include/rocm_ops.hpp                        # + GDR_DECODE_PACKED_BF16_PYBIND macro
> csrc/pybind/gdr_decode_packed_bf16_pybind.cu     # thin shell invoking the macro
> aiter/ops/gdr_decode_packed_bf16.py              # drop the _hip suffix
> aiter/__init__.py                                # + from .ops.gdr_decode_packed_bf16 import *
> ```
> 
> The `m.def` belongs in a `_PYBIND` macro in `csrc/include/rocm_ops.hpp` (111 ops are registered there today) rather than in the kernel `.cu` — that file is where all op signatures are reviewed. Right now the host function is also unreachable from other TUs inside an anonymous namespace.
> 
> Also, please use `gdr` over `gdn` to match `fused_split_gdr_update`, `gdr_decode` and `module_chunk_gdr_fwd_h`. If the missing top-level export was to avoid an import failure on non-gfx950, `aiter/ops/opus/__init__.py` has the guard-and-stub pattern.

Thanks for the detailed review and here's some update:

1. I added tuned FlyDSL results to the benchmark table. The hip fused_split_gdr_update is not in the comparison because it is not a drop-in baseline for SGLang’s persistent-state contract. SGLang stores the recurrent state in V-major [pool, HV, V, K] layout with K contiguous, while fused_split_gdr_update requires a pre-swizzled FP32 [pool, HV, K/4, V, 4] layout. Using this path in SGLang would require either a separate persistent state pool or an additional layout-conversion and synchronization path, which would be incompatible with the existing prefill, fallback, and cache-management paths.
2. The main blocker to transfer this optimization to existing flydsl kernel is that the non-temporal state load. FlyDSL’s raw buffer operations support cache modifiers, but the current GTensor.load() path does not forward the modifier. Porting this optimization therefore requires either state-specific raw buffer loads or an audited GTensor change, followed by ISA validation.
3. The current invalid row contract is aligned with the SGLang packed Triton backend. This is important because SGLang allocates the output with new_empty(); skipping the output write would expose uninitialized data. Positive out-of-range indices are handled defensively in the same way. 
4. I removed the module-local --offload-arch=gfx950 and now rely on Aiter’s normal target selection. Fresh gfx950 JIT and gfx942 build-only checks both pass, while the runtime wrapper remains gfx950-gated. The placement problem of this HIP PR is also modified to be aligned with the regulations.
5. I also tightened the tolerance to rtol=1e-2, atol=1e-3 and added a 16-step BF16-state drift test. The focused upstream test suite passes.
6. MTP is not currently the focus of our SGLang Qwen3.5 optimization; this PR primarily targets T=1 decode. SGLang’s MTP verify path currently falls back to Triton, so T>1 inputs are not sent to this HIP op. We'll investigate a follow-up to evaluate a state-resident loop over T while preserving the specialized T=1 fast path.

yiijin (dismissed):

yichiche:
@yiijin  All CI pass, can you help merge? Thanks in advance.

yiijin:
> @yiijin All CI pass, can you help merge? Thanks in advance.

All looks good on my side, but I don't have merge permission for the aiter repository.

junhaha666 (approved):

junhaha666 (approved):

PR #4510, opened 2026-08-02
[Bugfix][FlyDSL] Honor b_nt in mixed-MoE stage-2 and retune configs

qiongz:
## Summary

`compile_mixed_moe_gemm2_common` accepts `b_nt` and discards it on its first line with
`del b_nt`, so the w2 weight load is always emitted without a cache modifier. The `_bnt{N}`
suffix the stage-2 registry generates is a silent no-op on the mixed (fp4/mxfp4/fp8 weight)
path: `…_atomic_bnt2` compiles to **byte-identical machine code** to `…_atomic`.

Because every request is discarded, **`main` today behaves as "no stage-2 row gets `nt`"** — the
288 `_bnt2` markers in the shipped CSVs are inert, and the tuner that placed them was choosing
between identical candidates. So fixing the builder alone would activate 288 never-measured
requests at once, and **43 of them are measurably harmful, the worst by `27.5 %`**. This PR
therefore has two halves: the builder fix, and a retune of the rows that ask for `nt`.

All 589 affected rows were A/B'd against their own `_bnt2` twin, 3 rounds each. Against `main`:

| | rows gaining >2 % | rows losing >2 % | worst row | mean over 588 rows |
|---|---|---|---|---|
| **this PR** — fix + tuned CSV | **210** | **2** | **`−2.42 %`** | **`+1.80 %`** |
| *control* — fix only, CSVs as-is | 150 | 45 | `−27.5 %` | `+0.67 %` |
| *control* — fix + `nt` on every row | 210 | 208 | `−33.6 %` | `−1.86 %` |

The gain side is identical whether you tune or switch `nt` on everywhere. **What tuning buys is
the loss side**: 208 regressing rows become 2. Leaving the CSVs alone is not a middle ground — it
forfeits 60 winning rows *and* keeps 45 regressions.

## Why `nt` pays in decode and costs in prefill

`b_nt=2` marks the w2 load non-temporal, so the line is not allocated in the 256 MB Infinity
Cache. Whether that helps turns on one question — **is this slab read again in this dispatch?**

- **`reuse ≈ 1` (decode / small batch).** Each `(expert, n-block)` slab is read by exactly one
  workgroup. Caching it returns nothing and evicts data that would have been reused, so skipping
  the allocation is free speed.
- **`reuse > 1` (prefill).** An expert is hit by several m-blocks, the slab genuinely is re-read,
  and `nt` throws away reuse it could have had.

Median effect vs `main` across all 588 rows, with the tuned CSV's decision beside it:

| tokens | rows | tuned CSV puts `nt` on | `nt` on every row | **this PR** |
|---|---|---|---|---|
| 1–8 | 126 | 69 | `+0.39 %` | `0` |
| 16–64 | 105 | 86 | `+3.29 %` | **`+3.29 %`** |
| 128–512 | 111 | **101** | `+4.06 %` | **`+4.06 %`** |
| 1k–4k | 124 | 47 | `−0.49 %` | `0` |
| >4k | 122 | **2** | **`−9.40 %`** | `0` |

Up to the noise floor this PR is `max(main, nt-everywhere)` per row: it tracks the
`nt`-everywhere arm where `nt` pays and stays at `main` where it does not. The middle column is
that decision made visible — 91 % of the 128–512 rows get `nt`, 2 % of the `>4k` rows do.

`reuse` itself is **not** a usable predicate, which is why this is a tuned table rather than a
dispatch-time rule; see [Remaining work](#remaining-work).

### The defect, without a stopwatch

Final ISA for the `token=32` stage-2 kernel, over source × config:

| source | config asks | `buffer_load` with `nt` | total | ISA md5 |
|---|---|---|---|---|
| stock | `b_nt=0` | 0 | 20 | `b875f2572278` |
| stock | **`b_nt=2`** | **0** | 20 | `b875f2572278` ← **byte-identical. This is the bug.** |
| fixed | `b_nt=0` | 0 | 20 | `b875f2572278` ← fix is inert when not asked |
| fixed | `b_nt=2` | **8** | 20 | `662e5d64c8bb` |

The change is annotation-only: the two fixed-source ISAs are 546 lines each, 8 lines differ, and
the only difference on those 8 is a trailing `nt`. Those 8 are exactly the w2 stream — all
128-bit `dwordx4` on the weight base `s[4:7]`. Stage 1 does this correctly on the same shape in
the same call, emitting `nt` on 56 of its 86 `dwordx4` loads; that asymmetry is the whole bug.

## Changes

- **`mixed_moe_gemm_2stage_common.py`** — remove `del b_nt`; pass `cache_modifier=b_nt` to the w2
  `_buffer_load_vec` in `load_cell`. Weight load only; `load_x` untouched.
- **`test_flydsl_moe_a8w4.py`** — new `test_flydsl_stage2_a8w4_b_nt_reaches_w2_load`: compiles at
  `b_nt=0` and `b_nt=2`, asserts the emitted w2 loads carry `nt` only in the second. It has to
  read the ISA, because `nt` changes no arithmetic and **both `checkAllclose` legs pass even on
  the unpatched source** — which is why this survived. ~8 s, gfx950-gated like its neighbours.
- **10 `*_tuned_fmoe.csv`, `kernelName2` column only, 103 rows** — 43 drop a `_bnt2` that
  measurably hurts (median `6.98 %` recovered, worst `27.5 %`), 60 gain one that measurably helps
  (median `4.68 %`). The 485 other measured rows are untouched: 308 already agree with the
  measurement, 177 are ties.
- **`moe_kernels.py`** — correct the dispatch comment calling `b_nt` and `xcd_swizzle` "accepted
  as ignored kwargs". `xcd_swizzle` was already honoured; `b_nt` now is.

No API change. 305 rows carry `_bnt2`; **288 route to this builder** and are affected, spanning
43 shapes and 98 kernel names. The other 17 are `abf16_wfp4`, which go to
`compile_mixed_moe_gemm2_a16w4` and already hard-code `nt`. (Counted by resolving `kernelName2`
via CSV header — it sits at index 20, 21 or 33 depending on the file; a fixed index reports 202.)
FHMoE is unaffected: `fhmoe.py` defaults `b_nt=0` and ships no tuned config.

The CSV diff touches **only `kernelName2` and the timing columns** (`us1`, `us2`, `us`, `tflops`,
`bw`) — verified line-by-line that line counts, field counts and every other cell are
byte-identical. After the edit all 703 `flydsl_moe2` names still resolve, and the `_bnt2` count
goes 288 → 305 = 288 − 43 + 60.

Following review feedback that an earlier revision changed `kernelName2` and left the recorded
perf alone, **the timing columns were re-measured for the changed rows** — 102 of 103; one
`glm5_fp4` row at 32768 tokens produced no usable timing and is left as-is. These rows all
declare `gfx950` / 256 CU, which is this box. `us1`/`us2` are measured, `us` is their sum, and
`tflops`/`bw` come from calling the tuner's own `calculate` rather than a re-derivation, so the
arithmetic matches the rows already in the file. `err1`/`err2` are deliberately left alone: they
measure deviation from the reference implementation, not time, and this narrowed harness reports
a noisier `err1` than a full tune, which would read in the diff as an accuracy regression that
did not happen.

**One thing to know when reading that diff**, because it looks wrong at a glance: on the 42
stripped rows `us2` mostly goes *up* (median `+1.93 %`). Those rows' behaviour is unchanged —
`…_reduce_bnt2` and `…_reduce` compile to the same code today, since the request was discarded —
so that delta is purely this box versus whatever recorded the original numbers, not a regression.
Their actual benefit is that they will *not* become ~7 % slower once `b_nt` is honoured, which no
column in the file can show. On the 60 rows that gain `_bnt2` the behaviour does change and `us2`
drops accordingly, median `−2.25 %`.

## Results

**Method.** 589 rows × 3 rounds = 1767 paired measurements; 588 returned usable timings. Rig is
the stock `gemm_moe_tune.py`, unmodified, wrapped so the candidate registries hold only the two
stage-2 kernels differing in `b_nt`, with `mp_tuner` tapped for every candidate's `us` rather
than only the winner. Shapes come from each model's existing `*_untuned_fmoe.csv`, so **no model
weights are needed** — that is what made a full sweep affordable (~70 min across 8 cards).

- **Both arms in one invocation, one card.** With a single visible device `mp_tuner` sets
  `parallel_num == 1`, so the arms run sequentially rather than racing; `maxtasksperchild=1`
  gives each a fresh process. A row's effect is the median of the three per-round *ratios*, not a
  ratio of separately-taken medians, so drift cannot leak in.
- **Serving-faithful boundary.** The stock harness rotates through 11–52 input copies, evicting
  w2 by clean-line capacity — confirmed by instrumenting `run_iters_rotate`, not assumed. This
  matters; see [what I got wrong](#what-i-got-wrong-earlier).
- **Threshold from the repeats, not taste.** Three-round sign agreement is 47.7 % below 1 %,
  74.3 % at 1–1.5 %, **96.4 % at 1.5–2 %**, 99.4 % above 3 %. The break is at 1.5 %; I use
  **±2 %** and leave anything smaller untouched. At 1.5 % the set to *strip* is identical (43),
  only the set to *add* grows by 10.
- **Contention.** Shards pinned one per GPU; a watchdog sampled the KFD process table every 10 s
  and paused a shard between rows if a foreign process appeared. **Zero events.**

305 rows switch `nt` on (mean `+3.47 %` over those); the other 283 are byte-identical to `main`.
Both regressing rows are 2- and 4-token shapes whose rounds disagreed in sign — the noise floor
itself, not a mechanism. Isolating the CSV half, i.e. what the retune adds over fixing the
builder alone: all 103 changed rows get faster and none slower, by construction, since each edit
picks whichever arm measured faster (min `+2.04 %`, median `+4.85 %`, max `+27.52 %`).

### Worked example: one config, both ends of the range

Aggregates hide the shape of this, and a handful of rows can mislead in either direction — an
earlier revision of this PR quoted five pilot rows that happened to be four regressions and one
unreachable gain, which reads as "`nt` is harmful" and is not what the data says. Here are *all*
the measured rows of `dsv4_fp8fp4` at decode and at prefill:

| tokens | rows | shipped `_bnt2`? | measured `nt` effect | tuned action | net vs `main` |
|---|---|---|---|---|---|
| 32 | 8 | yes | `+0.68 … +7.04 %` | keep | **`+0.68 … +7.04 %`** |
| 32 | 1 | no | `−2.85 %` (tie) | keep | `0` |
| 8192 | 1 | **yes** | **`−9.49 %`** | **strip** | **`0`** |
| 8192 | 6 | no | `−19.99 … +0.32 %` | nothing to do | `0` |

The config that looked worst in the pilot shows the mechanism most cleanly. It already had
`_bnt2` on 8 of its 9 decode rows, worth up to `+7.04 %` — value that was never delivered
because the builder discarded the request. The one prefill row that also asked for it was the
`−9.49 %` regression, and that row gets stripped. `kimik3_a8w4` is the mirror image: its two
decode rows never asked for `nt` and gain `+2.50 %` / `+4.85 %` once they do, while its one
prefill row carrying `_bnt2` (`−2.04 %`) is stripped.

### Tests

```bash
pytest op_tests/flydsl_tests/test_flydsl_moe_a8w4.py -q          # 26 passed
pytest op_tests/test_fhmoe.py -q                                 # 10 passed
pytest op_tests/flydsl_tests/test_flydsl_moe_a8w4.py -k b_nt -q  # 1 passed
#   ... with upstream/main's builder restored:                     1 failed
```

`nt` changes no arithmetic, so the deterministic `reduce` shapes must return unchanged, and they
do: `M=7937` and `M=1` are **bit-identical** to the unpatched build (`rms_rel` exactly 0),
including `M=1`, whose kernel does now emit `nt`. Decode `M=32` uses a bf16 atomic epilogue and
is not run-to-run deterministic in either build; its `rms_rel` vs unpatched is `4.667e-3`
against a `4.712e-3` self-noise floor measured by running the *unpatched* build twice.
`black` 26.3.0 and `ruff` 0.16.0 clean; no C++ changes.

## Remaining work

1. **E2E / serving validation** — the main gap. The tuner drives routing with uniform synthetic
   scores while real routing is skewed, which changes how many experts are touched and thus DRAM
   volume. I have no serving rig for these 13 models, so the per-row decisions rest on kernel-level
   timing only.
2. **Non-gfx950 parts** — MI300X / MI250X, no access. The refreshed timing columns are from
   gfx950 / 256 CU, which is what those rows already declared.
3. **`op_tests/test_moe_2stage.py`** — gates on AOT-cache coverage via `run_only_env()`, which my
   container cannot satisfy; bypassing with `--csv-filter` then fails inside `torch_moe_stage2`'s
   reference path (`mxfp4_to_f32` on an fp8 tensor). Fails at the identical line with this patch
   reverted, so it is pre-existing. Please run in CI.
4. **A dispatch-time predicate instead of a tuned table** — I tried; it does not hold up. The
   obvious candidate, `reuse = M·topk/(E·block_m)` floored at 1, gets the sign right only **83 %**
   of the time on the 411 rows with a decided verdict, and the failure is structural: the floor
   collapses 275 of them into one `reuse = 1.0` bucket of which **69 (25 %) regress**, because
   inside that bucket the outcome is governed by token count — which the formula divides out. The
   1k–4k bucket is genuinely bimodal (24 helped vs 33 hurt) and no closed form reproduces it. Any
   predicate would have to be fitted to measurements, at which point the tuned CSV *is* the
   predicate.

If maintainers prefer a smaller change, **removing the knob** from the stage-2 mixed path is
still viable, at the cost of the 60 rows where `nt` measurably helps. It is not behaviour-free
though: dropping `_bnt{N}` from the registry makes the 288 CSV names unresolvable
(`get_flydsl_kernel_params` returns `None`, the caller raises), and its suffix-stripping regex
only handles *trailing* `_kw`/`_fp4`/`_fp8`/`_sbm` while `_bnt2` sits mid-name.

## What I got wrong earlier

The first revision carried `−37.8 %` stage-2, `+15.8 %` end-to-end and `+11.8 %`
serving-weighted. **Withdrawn.** They reproduce to within 0.1 %, but only under a boundary
serving does not have: the harness evicted the 256 MB Infinity Cache by *writing* a 512 MB
buffer, leaving ~256 MB of dirty lines, so the baseline arm paid a writeback storm that `nt`
sidesteps. Evicting by *reading* the same buffer clears the cache just as thoroughly but leaves
clean lines, and the decode effect collapses from `+15.9 %` to `+1.4 %`. Serving is the
read-flush case. Also withdrawn for want of a reproducible basis: the per-token-bucket
improvement table, the `2.72 → 4.37 TB/s` bandwidth figure and the `44 % → 71 %` roofline figure.

Two claims made while auditing are also corrected. **"Those 288 markers are coin flips"** — the
premise holds, the candidates were byte-identical, but the conclusion does not: `_bnt2`
prevalence falls monotonically with token count, 60 % to 17 %, so it is least common exactly
where `nt` hurts, and **82.5 % of the 588 measured rows agree with the measured optimum**. The
103 misconfigured rows are real; "arbitrary" was the wrong word for the other 485. And **"the
stock tuner leaves w2 resident"** — wrong, it rotates inputs, which is clean-line capacity
eviction, i.e. the serving-faithful boundary.

## AI assistance

To be straightforward about the extent: **essentially all of this was produced by an AI coding
agent** (Cursor), during and after an optimization pass on Kimi-K3 TP8 / 8× MI355X. That
includes finding the defect, the kernel change, the regression test, the measurement harness, the
1767-measurement sweep, the regenerated CSV rows, and this description. My role was direction and
review — in particular, requiring that the CSV rows be measured rather than argued about before
this could land.

The agent also produced the inflated numbers the first revision carried, and I did not catch them
before opening the PR. The failure mode is worth naming precisely, because "the agent
hallucinated numbers" is the wrong lesson — the numbers were real and reproduce to within 0.1 %.
What it did not do was question whether its own harness's `flushbuf.zero_()` modelled the cache
state serving presents. It did not, and that one choice accounts for the whole 10× gap.

Over the audit it corrected a further set of its own claims, which I list because the pattern
matters more than any single item: a speculated compiled-kernel cache-key collision (could not be
reproduced, dropped); "the stock tuner leaves w2 resident" (wrong, it rotates inputs);
"removing the flag is behaviour-free" (wrong, it breaks name resolution); "the 288 markers are
coin flips" (overclaim, see above); the `reuse` formula as a dispatch predicate (83 % accurate,
withdrawn); a five-row pilot table that misrepresented the overall picture; and reporting this
PR's effect against the shipped CSVs when the correct baseline is `main`, where the knob is
inert. Two implementation bugs were also caught by its own checks rather than by inspection: a
CSV round-trip that silently padded 8 short rows to the header width, and a GPU-contention
watchdog that raised false alarms by racing `docker top` against the tuner's process churn.

I have **not** independently re-derived the numbers by hand, so please treat them as
reviewer-checkable rather than reviewed. They were built to be checkable: the load-bearing
evidence is timing-independent (the ISA truth table, and a 546-line ISA diff in which 8 lines
differ and only by a trailing `nt`), the regression test is confirmed to fail on unpatched
source, the defect is readable directly in the source rather than inferred from timing, the
305/288/17 counts are re-derivable by resolving `kernelName2` via CSV header, and every timing
was taken on a single bus-ID-pinned GPU with a null control (byte-identical arms ⇒ ~0 %) after an
earlier round was invalidated by `ROCR_VISIBLE_DEVICES` indices being a permutation of
`rocm-smi` indices on this box. The sweep scripts and raw per-round results can be attached if
that would help review.

## Submission Checklist


qiongz:
Two asks for a maintainer, since this is my first PR here:

1. All seven workflows on this PR are sitting at `action_required` and need an
   "Approve and run workflows" click before any CI runs.
2. Could you also add the `ci:all` label? I get a 403 setting it myself.

`ci:all` matters more than usual here: this change alters the emitted stage-2 ISA for 288
tuned-config rows across 14 models, and I only have measurements for `kimik3_a8w4` (12 of
those rows). It reaches the two Kimi workflows (`kimi-downstream`, `kimi-perf-downstream`)
as well as the vLLM and ATOM benchmarks, which is the closest thing to coverage for the
models I cannot test myself.

qiongz:
One more, sorry: converting this PR to draft and back dropped the automatic
`team_aiter` review request, and I can't re-add it myself (I'm not in the org).
Could someone re-request it along with the two items above?

op_tests/flydsl_tests/test_flydsl_moe_a8w4.py
```diff
@@ -238,6 +238,73 @@ def test_flydsl_stage2_a8w4_gui(inter_dim, seed):
...
+            model_dim_pad=0,
+            b_nt=b_nt,
+        )
+        torch.cuda.synchronize()
+        _check_close(data["ref_stage2"], out, f"stage2_a8w4_b_nt{b_nt}")
+        counts[b_nt] = _stage2_nt_load_count(tmp_path / f"bnt{b_nt}")
+    compile_mixed_moe_gemm2.cache_clear()
+
+    assert counts[0] == 0, f"b_nt=0 emitted {counts[0]} non-temporal loads"
+    assert counts[2] > 0, "b_nt=2 was accepted but no emitted load carries nt"
```
coderfeli, line 305: no need this test.

aiter/configs/model_configs/kimik3_a8w4_tuned_fmoe.csv
```diff
@@ -21,13 +21,13 @@ gfx950,256,3,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float8_e
...
 gfx950,256,8,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float8_e4m3fn,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,31.3425,flydsl_moe1_afp8_wfp4_bf16_t32x128x256_w3_gui_fp8,0.0%,22.3907,flydsl_moe2_afp8_wfp4_bf16_t32x256x128_reduce_persist,0.0%,53.7332,0,0,0,19.67,68848.72,
 gfx950,256,16,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float8_e4m3fn,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,56.8747,flydsl_moe1_afp8_wfp4_bf16_t32x128x256_w4_gui_xcd4_fp8,0.0%,36.1336,flydsl_moe2_afp8_wfp4_bf16_t32x128x256_atomic_bnt2,0.0%,93.0083,0,0,0,22.73,39776.54,
-gfx950,256,32,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float8_e4m3fn,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,94.3599,flydsl_moe1_afp8_wfp4_bf16_t32x128x256_w3_gui_fp8,0.0%,57.6936,flydsl_moe2_afp8_wfp4_bf16_t32x128x128_atomic,0.0%,152.0535,0,0,0,27.81,24331.7,
+gfx950,256,32,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float8_e4m3fn,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,94.3599,flydsl_moe1_afp8_wfp4_bf16_t32x128x256_w3_gui_fp8,0.0%,57.6936,flydsl_moe2_afp8_wfp4_bf16_t32x128x128_atomic_bnt2,0.0%,152.0535,0,0,0,27.81,24331.7,
```
coderfeli: no perf change?

qiongz: That's a miss, I'll update the perf number

PR #4514, opened 2026-08-03
[Feature][FlyDSL] Add multi-B MoE kernels for ROCm DWDP

AMD-yanfeiwang:
## Summary
- add gfx950 eager multi-B MoE entry points backed by rank-ordered device pointer tables
- support MXFP4, A8W4, MXFP8, and conventional FP8 128x128 block scales with strict partition contracts
- preserve global expert routing, uneven partitions, bias, no-combine, and isolated JIT cache identities

## Test plan
- [x] CPU multi-B contract suite
- [x] gfx950 2/4/8-partition boundary and uneven-partition tests
- [x] conventional FP8 cosine and normalized-L2 validation
- [x] SGLang integration tests through the dependent ROCm DWDP branch

## Scope
This path is eager-only and gfx950-only. gfx1250 fails explicitly rather than falling back to an incompatible wave64 kernel.


AMD-yanfeiwang:
SGLang integration PR: https://github.com/sgl-project/sglang/pull/33310

PR #4517, opened 2026-08-03
[gfx950][Triton] Fix unified_attention num_stages > 1 crash with release_tmp3

Rohan138:
`select_2d_config` (decode) and `select_3d_config` (prefill) in `unified_attention.py` returned `num_stages > 1` for gfx950, triggering the release_tmp3 Triton software-pipeliner \"defs don't dominate\" crash — the same class of bug fixed for `fwd_prefill.py`, `mla.py`, `bwd.py`, and config files by 54ce92317, but `unified_attention.py` was missed.

Fix: `num_stages=1` for gfx950 in both config selectors.

Reproduces as a memory access fault in the vLLM CI after the v0.1.19 aiter bump:
`tests/models/quantization/test_gpt_oss.py::test_gpt_oss_attention_quantization[amd/gpt-oss-20b-WFP8-AFP8-KVFP8-0.89-1]` on MI355X.

### DIFF
--- a/aiter/ops/triton/attention/unified_attention.py
+++ b/aiter/ops/triton/attention/unified_attention.py
@@ -108,6 +108,10 @@ def select_2d_config(
         TILE_SIZE = min(64, triton.next_power_of_2(block_size))
         if arch.is_rdna:
             num_stages_2d, num_warps = 1, 4
+        elif arch.name == "gfx950":
+            # release_tmp3 Triton software-pipeliner crashes on gfx950 with
+            # num_stages > 1 ("defs don't dominate"); keep it at 1.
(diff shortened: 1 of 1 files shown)


fxmarty-amd:
@Rohan138 The memory access fault is unrelated to unified attention from what I have seen, see maybe https://github.com/vllm-project/vllm/pull/50859 https://github.com/ROCm/aiter/pull/4530

There is memory corruption from aiter's triton MOE `routing`, that tries to access irrelevant memory that may have been occupied by somebody else prior.

Maybe the unified attention fix helps with that, but it'd be an adjacent fix IMO - happy to discuss

PR #4522, opened 2026-08-03
[HIP] Enable chunk gated delta rule fwd-h on gfx1201

stevenshenyj:
## Summary

Enable the HIP hidden-state kernel used by chunk gated delta rule prefill on gfx1201.

- use the gfx12 wave64 BF16 WMMA intrinsic on gfx1200 and gfx1201
- compile the K5 JIT module with `-mwavefrontsize64` to preserve its existing 64-lane decomposition
- remap gfx12 WMMA accumulator lane groups from physical order `0,1,2,3` to result-row order `0,2,1,3`
- allow the HIP K5 path on gfx12 for prefill while keeping it disabled when decode tokens are present
- enable the existing dense, variable-length, and indexed-state HIP tests on gfx12

## Why

The kernel is organized around four 64-lane waves and originally uses the CDNA wave64 MFMA intrinsic. gfx1201 defaults to wave32 and does not support that MFMA instruction. The gfx12 wave64 WMMA intrinsic plus the explicit wave64 compilation mode preserves the kernel decomposition, while the accumulator row-group remap accounts for the gfx12 WMMA lane layout.

## Validation

Validated on gfx1201 with the Qwen3.6-35B TP4 GDN shape (`Hk=4`, `Hv=8`, `K=V=128`):

- dense and variable-length HIP K5 paths
- indexed hidden-state pool updates
- complete variable-length prefill pipeline at 1024 and 4096 packed tokens
- native and HIP outputs, hidden snapshots, and final states matched within the existing K5 tolerance (`atol=rtol=5e-2`)

Decode and mixed prefill/decode execution continue to use the existing non-HIP path.

### DIFF
--- a/aiter/jit/optCompilerConfig.json
+++ b/aiter/jit/optCompilerConfig.json
@@ -1780,7 +1780,8 @@
         ],
         "flags_extra_cc": [],
         "flags_extra_hip": [
-            "'-ffast-math'"
+            "'-ffast-math'",
+            "'-mwavefrontsize64'"
         ],
         "extra_ldflags": "None",
         "extra_include": [],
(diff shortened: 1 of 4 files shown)


stevenshenyj:
Closing this PR so the repository owner can review the fork branch and open the upstream pull request manually.

PR #4526, opened 2026-08-03
[HIP] [CK] [FlyDSL] [Kernel] Extend MXFP4 GEMM1 replacement to A4W4

fsx950223:
## Summary

- Extend the gfx950 FlyDSL MXFP4 GEMM1 replacement to A4W4 MoE workloads.
- Add activation-aware GEMM1 variants, inline BF16→FP4 quantization, hidden-load prefetch (`_hpf`), and expanded tuner coverage for BN / wave / K-wave / XCD choices.
- Integrate the replacement with the adaptive MoE auxiliary sort, cache-safe AOT/runtime dispatch, and layout-compatible stage2 paths.
- Remove legacy MXMOE-GEMM1 + generic-GEMM2 handling that is no longer supported.

## GEMM2 scope

GEMM2 tuning is intentionally out of scope for this PR.

The final config set was audited against latest main `e583a77c6`:

| Config file | Retained GEMM1-only rows | GEMM2 differences |
|---|---:|---:|
| `glm5_fp4_tuned_fmoe.csv` | 4 | 0 |
| `kimik2_fp4_tuned_fmoe.csv` | 33 | 0 |
| `kimik3_a4w4_tuned_fmoe.csv` | 3 | 0 |
| `qwen3_5_397b_fp4_tuned_fmoe.csv` | 0 | 0 |
| **Total** | **40** | **0** |

For all 81 audited rows, `kernelName2`, `us2`, and `err2` are byte-equivalent to latest main. Thirteen rows had also changed `block_m`; those were retuned with main's GEMM2 and `block_m` locked. Rows without a confirmed GEMM1 gain fell back to the main GEMM1. The Qwen3.5-397B FP4 config is fully restored to main.

## Performance validation

Measured on one gfx950 MI355X system with warm FlyDSL caches. Main and PR were alternated on the same GPU; the 81-row set received three A/B rounds and every negative/borderline row received five-round confirmation.

Environment for Kimi-K3 A4W4:

```bash
AITER_SITUV2_A4W4=1 \
AITER_FLYDSL_STAGE2_FP8=1 \
AITER_MXFP4_INTERMEDIATE=1
```

Final result versus latest main:

- **35 rows improve by more than 2%**
- **No confirmed regression greater than 1%**
- **3.39% geomean E2E improvement** across the 81-row audit set
- **2.69% aggregate latency reduction**
- Maximum `logits_diff`: **0.00733** (gate: `<= 0.01`)

The two small-token Kimi-K2 rows that regressed with the unchanged kernel were recovered with HPF GEMM1 variants:

- token 4, `inter_dim=512`: **+0.82%**
- token 8, `inter_dim=256`: **+1.47%**

## Validation

- 702 locked-GEMM2 GEMM1 candidates profiled for the 18 rows that had changed GEMM2 selection.
- All 81 final rows passed strict fused-MoE accuracy validation.
- Config shape-collision guard: **16/16 passed**.
- Black, Ruff, wheel build, standard tests, OPUS tests, Kimi-K2 accuracy/performance, and the current model accuracy checks pass in CI.

## Test methodology

Config comparisons run the production `fused_moe` path. Main and PR use separate checkouts and caches but execute on the same GPU in alternating order. A row is retained only when repeated measurements show a real gain; otherwise its main config is preserved. Per-stage attribution uses `rocprofv3 --kernel-trace`, while final acceptance is based on end-to-end `fused_moe` latency and `logits_diff <= 0.01`.


aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
                 kernelName1=kernelName1,
-                interleave=(gate_mode == GateMode.INTERLEAVE),
-            ),
-            stage2=functools.partial(
-                _mxfp4_a4w4_stage2_fw,
-                kernelName2=kernelName2,
-                inter_dim_pad=intermediate_pad,
-                model_dim_pad=hidden_pad,
+                interleave=runtime_interleave,
+                native_scale_layout=_bm == 16,
```
binding7012: **`native_scale_layout` defaults to False, so the tuner path produces all-NaN**

`get_2stage_cfgs` binds `native_scale_layout=_bm == 16` here, but `_mxfp4_a4w4_stage1_fw` defaults the parameter to `False`, and the tuner calls it directly without passing the kwarg (`csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py:6214`).

Measured on BM16: the two settings produce an identical quantized payload but a different scale layout (2304 vs 2352 non-zero bytes). Forcing it off end-to-end yields 100% NaN output and the accuracy assert fires. So every BM16 candidate the tuner evaluates gets NaN -- and almost every `token<=256` row in the shipped CSVs is BM16, meaning those rows can no longer be re-tuned.

binding7012 (commented):
One thing I would like resolved before merge,

### `native_scale_layout` defaults to False, so the tuner produces all-NaN (blocker)

See the inline comment on `aiter/fused_moe.py:2713`. `get_2stage_cfgs` binds `native_scale_layout=_bm == 16`, but `_mxfp4_a4w4_stage1_fw` defaults it to `False` and the tuner calls that function without the kwarg. On BM16 the two settings produce different scale layouts, and forcing it off end-to-end gives 100% NaN. Since nearly every `token<=256` row in the shipped CSVs is BM16, those rows can no longer be re-tuned.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
             block_m=_bm,
             ksplit=int(ksplit),
-            fuse_quant="fp4",
-            output_aux=True,
-            prequant=False,
+            fuse_quant=_p1["out_dtype"],
+            output_aux=AUX_SORT_OPUS,
+            prequant=_p1["a_dtype"] == "fp8" and not _p1["inline_quant"],
+            has_bias=enable_bias and _p1.get("enable_bias", False),
+            stage2_has_bias=enable_bias and stage2_supports_bias,
```
binding7012: **`bias2` is silently dropped**

`stage2_has_bias` is unconditionally True for every MXMOE config: `enable_bias` holds for any bf16 + per_1x32 + fp4-weight case, and `stage2_supports_bias` only checks the `flydsl_` / `cktile_` prefix. So line 3437 puts `bias2` into `extra_stage2_args`, and `_mxfp4_a4w4_stage2_fw` swallows it via `**_kwargs` and never forwards it (line 1996).

The layer below already supports it: `_flydsl_v2_stage2_wrapper` passes `bias=bias2` to `mxfp4_moe_gemm2`; only the caller omits it. The GPT-OSS shapes were just added to the aux `SHAPES`, so this will affect them for real.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
+            stage2=stage2_func,
             block_m=_bm,
             ksplit=int(ksplit),
-            fuse_quant="fp4",
-            output_aux=True,
-            prequant=False,
+            fuse_quant=_p1["out_dtype"],
+            output_aux=AUX_SORT_OPUS,
+            prequant=_p1["a_dtype"] == "fp8" and not _p1["inline_quant"],
+            has_bias=enable_bias and _p1.get("enable_bias", False),
```
binding7012: stage1 bias can be silently dropped, and the guard faces the wrong way

When the model has `bias1` but the tuned name lacks `_bias`:

```
has_bias = enable_bias and _p1["enable_bias"] = False
-> bias1 is never forwarded (line 3433) -> no error, just wrong numerics
```

And the check at line 1913, `if not p1["enable_bias"] and bias1 is not None: raise`, can never fire -- `bias1` only arrives when `has_bias` is already True.

Suggestion: move the check into `get_2stage_cfgs` and invert it -- "model has a bias but the name has no `_bias`" -> discard the cfg and warn.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
-    if _is_mxfp4_kname(kernelName1) or _is_mxfp4_kname(kernelName2):
-        # gate_mode is a runtime weight-layout property, not a tuning key: route
-        # any a4w4 kernelName to the port; the bound interleave flag picks the
-        # compiled il/sep variant at runtime.
+    if _is_mxfp4_kname(kernelName1):
         try:
-            _bm = _parse_mxfp4_g1_kname(kernelName1)["BM"]
+            _p1 = _parse_mxfp4_g1_kname(kernelName1)
+            _bm = _p1["BM"]
         except ValueError:
```
binding7012: **The `except ValueError` fallback is incomplete, and useless even when taken**

This fallback dict only has `a_dtype` / `out_dtype` / `interleave`; it is missing `inline_quant` / `BM` / `act` / `k_wave` / `num_waves`. Line 2720,

```python
prequant=_p1["a_dtype"] == "fp8" and not _p1["inline_quant"]
```

only avoids a `KeyError` because the fallback hardcodes `a_dtype="fp4"` and `and` short-circuits. I verified that switching it to fp8 raises `KeyError('inline_quant')` immediately.

More importantly, taking the branch does not help -- `_mxfp4_a4w4_stage1_fw:1902` re-parses the same `kernelName1` and raises the same `ValueError`. The fallback only defers the failure from config-parse time to the first forward.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
+            _p1 = {
+                "a_dtype": "fp4",
+                "out_dtype": "fp4",
+                "interleave": gate_mode == GateMode.INTERLEAVE,
+            }
             _bm = int(block_m) if block_m is not None else BLOCK_SIZE_M
+        expected_act = _mxfp4_activation_name(activation)
```
binding7012: **The activation check is written twice; the second one is unreachable and has the opposite policy**

The soft check at line 2514 already sets `cfg = None` for any MXMOE cfg whose activation does not match, and `kernelName1` is only taken from `cfg` when `cfg` exists (line 2621, otherwise `""`). So this `raise ValueError` can never be reached.

The two also disagree on policy: one warns and falls back to heuristics, the other raises. Suggestion: keep one and settle on a single policy.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
```diff
@@ -7,14 +7,35 @@
...
+        (32, True, False),
+        (32, False, False),
+        (64, False, False),
+        (128, False, False),
+        (16, True, True),
+    },
+}
+
+
+def _effective_use_nt(*, n_tokens, topk, NE, BM, use_nt, inline_quant):
```
binding7012: `_effective_use_nt` makes the compile key depend on runtime `n_tokens`, causing AOT misses

`use_nt` is part of the compile cache key, and this helper turns it into a function of `n_tokens`. AOT only sees the padded token tier from the CSV (`aiter/aot/flydsl/mxfp4_moe.py:156`); runtime sees the real M.

Measured on the kimik2 row with `tok=2048 / NE=385 / topk=9`:

```
AOT     use_nt@2048   = False
runtime use_nt@M=1025 = True      (M=1025 falls in the same tier)
```

So this is an AOT cache miss, i.e. a JIT compile on the serving path.

This heuristic is also duplicated at `mxfp4_kname.py:103`, with inconsistent BM coverage (this one only handles BM32, the other handles 16/32/64).

aiter/ops/flydsl/kernels/mxfp4_gemm1.py
```diff
@@ -780,51 +1331,125 @@ def compile_gemm1_a4w4_port(
...
 
-    _, _, _, lds_bytes = _bm_constants(BM, BN, KH_TILE, _K_TILES_TOTAL)
+    _, _, _, lds_bytes = _bm_constants(BM, BN, KH_TILE, K_TILES_TOTAL, k_wave)
 
     variant_tag = "iq" if inline_quant else ("nt" if use_nt else "cached")
     # Tag with H/INTER/NE so different shape specializations get distinct
     # kernel/smem symbols (so KIMI and non-KIMI instances never collide).
     gu_tag = "il" if interleave else "sep"
-    name_suffix = f"h{_K}_i{_INTER}_ne{_NE}_bm{BM}_{variant_tag}_{gu_tag}"
+    name_suffix = (
```
binding7012, line 1447: **[MEDIUM] `situ_beta` / `swiglu_limit` are compile keys but absent from the kernel symbol name**

`name_suffix` encodes a_dtype / H / INTER / NE / BM / act / bias / BN / xcd / w2 / kw, but not `situ_beta`, `situ_linear_beta`, or `swiglu_limit` -- all three of which are in the `_get_compiled_mxfp4_gemm1_port` cache key. In-process this is safe thanks to `functools.cache`; the risk is the cross-process disk/AOT cache: if it is keyed on the symbol name, a different beta reuses the wrong kernel. The comment claims they are "already part of FlyDSL's closure cache key" -- worth confirming explicitly.

These three are also not normalized: `fused_moe.py:3288` passes `beta or 1.0` unconditionally, regardless of `act`. So `act=silu` + `beta=2.0` creates a cache entry for a value the kernel never reads, and AOT only guesses `(1.0, 1.0)` and `(4.0, 25.0)` (`aiter/aot/flydsl/mxfp4_moe.py:164`), so it always misses.

aiter/aot/flydsl/mxfp4_moe.py
```diff
@@ -127,23 +151,57 @@ def _add(job):
...
+                            # M blocks saturate the experts. AOT must specialize on
+                            # that effective value, not only the kernel-name flag.
+                            "use_nt": effective_use_nt,
+                            "n_tokens": token,
+                            "inline_quant": p1["inline_quant"],
+                            "D_HIDDEN": model_dim,
+                            # GEMM1 specializes on the logical output width. GEMM2
+                            # padding is backend-specific and must not change its
+                            # AOT cache key (for example, inter_dim=384 vs 512).
+                            "D_INTER": inter_dim,
```
binding7012: **AOT uses the logical `D_INTER` while runtime uses the physical one**

```
AOT     : "D_INTER": inter_dim          (here; previously v2_d_inter, padded to a multiple of 256)
runtime : D_INTER = w1.shape[1] // 2    (aiter/fused_moe.py:1926)
```

I ran `parse_csv`: the kimik3 row (`inter_dim=384`) produces an AOT stage1 job with `D_INTER=384`. But the aux `SHAPES` entry reads `(896, 3584, 512, 16),  # Kimi-K3 TP=1/2 (INTER 384 pads to 512)`, and `_mxfp4_a4w4_stage2_fw` has dedicated handling for `w2.inter_real != D_INTER`, which says the weights are stored padded. So the 384 entry AOT compiles is never hit.

aiter/ops/flydsl/kernels/mxfp4_gemm1.py
```diff
@@ -1,127 +1,66 @@
...
-    return fx.Int32(flat)
-
-
-def n_out_for(inter):
-    return 2 * inter
-
-
-LOG2E = 1.4426950408889634
 
+def scale_out_uses_atomic(BM):
```
binding7012: **`scale_out_uses_atomic` always returns False, and it breaks the lazy-import pattern**

This unconditionally returns `False` and never uses its `BM` argument, so the `inter_sorted_shuffled_scale.zero_()` at `mxfp4_gemm1_kernels.py:298` is a dead branch. Suggest removing the whole thing (function + import + call site).

Related: it is imported at module level (`mxfp4_gemm1_kernels.py:10`), and this file does `import flydsl.expr as fx` at its top, so that pulls all of flydsl into import time. The other two `from .kernels.mxfp4_gemm1 import ...` in that module are still function-local and now pointless -- and the AOT module imports `_effective_use_nt` from that file precisely to avoid triggering a JIT load.

aiter/ops/flydsl/mxfp4_kname.py
```diff
@@ -1,21 +1,155 @@
...
+}
 _MXMOE_G2_FLAG_TOKENS = {"NT", "ATOMIC", "F4OUT", "CSHUFFLE"}
 _MXMOE_NUMERIC_RE = re.compile(r"^([A-Z]+)(\d+)$")
-_MXMOE_TILE_RE = re.compile(r"^(\d+)x(\d+)x(\d+)$")  # <BM>x<BN>x<BK>
+_MXMOE_TILE_RE = re.compile(r"^(\d+)x(\d+)x(\d+)$")
 _MXMOE_PREFIX = {1: "flydsl_mxmoe_g1_a4w4_", 2: "flydsl_mxmoe_g2_a4w4_"}
+_MXMOE_G1_PREFIX_RE = re.compile(r"^flydsl_mxmoe_g1_a(?P<a>[48])w4_")
+
+
+def _select_mxfp4_block_m(*, token: int, expert: int, topk: int) -> int:
```
binding7012: **~80 lines of new dead code, and it contradicts the shipped CSVs**

```
mxfp4_kname.py:28   _select_mxfp4_block_m     (no external caller)
mxfp4_kname.py:45   _make_mxfp4_g1_kname      (only called by :103)
mxfp4_kname.py:103  _select_mxfp4_g1_kernel   (no caller)
```

I grepped the whole repo (including `csrc/` and `op_tests/`) -- none of these have a caller.

Two further points:

1. This puts *selection policy* into a module whose docstring explicitly says "no torch / JIT deps" so the AOT parse pass can import it. The selection logic belongs in `fused_moe` or the tuner.
2. Line 33 returns BM32 for `token == 1`, with the comment "BM16's fused inline quantization has excessive error for a single token" -- yet every `token=1` row in the CSVs uses BM16, and the discontinuity (`token == 1` -> 32, `token in [2, 128]` -> 16) is counter-intuitive. If that accuracy problem is real it deserves its own fix or explanation rather than being routed around by a function nobody calls. Could you clarify?

Suggestion: if these were written for the tuner, wire the tuner up to them; otherwise delete them.

aiter/ops/moe_mxfp4_aux.py
```diff
@@ -1,27 +1,65 @@
...
-#              (mirrors D_HIDDEN naming; the "INTER" matches aiter's main
-#              `inter_dim` convention. Not the expert count.)
-#   MB       = block_m (sort/gemm block size, ∈ {16, 32, 64, 128})
 
 from torch import Tensor
 
 from ..jit.core import compile_ops
 
+# Keep synchronized with moe_aux/codegen/gen_instances.py::SHAPES.
+MXFP4_MOE_SUPPORTED_SHAPES = frozenset(
```
binding7012, line 9: **`is_mxfp4_moe_shape_supported` has no caller, so the hazard it targets is still unguarded**

`MXFP4_MOE_SUPPORTED_SHAPES` (36 entries) and `is_mxfp4_moe_shape_supported` (line 51) have no callers anywhere in the repo. I diffed the list against `gen_instances.SHAPES` and they agree today (36 == 36), but it is a hand-maintained copy held together by the comment above it -- and the hazard it is meant to prevent (an un-codegen'd shape hitting `AITER_CHECK` instead of falling back gracefully) is still completely unguarded.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
```diff
@@ -7,14 +7,35 @@
...
 
 _SUPPORTED = {
     (32, True, False),
     (32, False, False),
+    (64, True, False),
     (64, False, False),
     (128, False, False),
     (16, True, True),
 }
+_SUPPORTED_BY_DTYPE = {
```
binding7012: **The supported-variant set is stored twice**

`_SUPPORTED_BY_DTYPE` here and `_G1_VARIANTS` at `kernels/mxfp4_gemm1.py:1305` hold identical data (I verified the contents match). main already had two copies; this PR turns each into a per-dtype dict, doubling the duplicated content.

Since line 10 of this file already imports across the module boundary, `from ...kernels.mxfp4_gemm1 import _G1_VARIANTS` would keep just one copy. Note the tuner reads `_SUPPORTED`, so that call site needs updating too.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
             ksplit=int(ksplit),
-            fuse_quant="fp4",
-            output_aux=True,
-            prequant=False,
+            fuse_quant=_p1["out_dtype"],
+            output_aux=AUX_SORT_OPUS,
+            prequant=_p1["a_dtype"] == "fp8" and not _p1["inline_quant"],
+            has_bias=enable_bias and _p1.get("enable_bias", False),
+            stage2_has_bias=enable_bias and stage2_supports_bias,
+            **route_bucket_metadata,
```
binding7012: **`**route_bucket_metadata` is always `{}` in the MXMOE branch**

It comes from line 2642, `route_bucket_metadata = ... if is_opus_cfg else {}`, and an MXMOE row's `kernelName2` is always `flydsl_`-prefixed -- line 2678 raises to guarantee exactly that -- so `is_opus_cfg` is always False.

aiter/fused_moe.py
```diff
@@ -3245,12 +3436,16 @@ def fused_moe_2stages(
                 extra_stage1_args["topk_ids"] = topk_ids
         if metadata.stage2_has_bias:
             extra_stage2_args["bias2"] = _normalize_bias_for_kernel(bias2)
-    if stage1_func in (_flydsl_stage1_wrapper, _opus_a8w4_stage1_wrapper):
+    if stage1_func in (
+        _flydsl_stage1_wrapper,
+        _opus_a8w4_stage1_wrapper,
+        _mxfp4_a4w4_stage1_fw,
+    ):
         extra_stage1_args["swiglu_limit"] = swiglu_limit
```
binding7012, line 3832: **Awkward nested condition**

The outer test is `if stage1_func in (A, B)` and the inner one is `if stage1_func is A and ...`, which tests one of them back again. Suggest lifting the `v2_output_layout` line into its own `if`, or splitting this into two branches.

aiter/ops/flydsl/moe_kernels.py
```diff
@@ -2074,10 +2078,16 @@ def _flydsl_moe_stage2_impl(
         )
```
binding7012, line 2098: **The new ndim assert makes `shape[-1]` meaningless**

This enforces `inter_states.ndim == 3` while the same change also switches `shape[2]` to `shape[-1]`. With the assert in place the two are exactly equivalent, so the change buys nothing and instead hints that non-3D input is possible here.

Pick one: relax the assert and keep `shape[-1]`, or keep the assert and revert to `shape[2]`.

aiter/fused_moe.py
```diff
@@ -2519,31 +2655,72 @@ def get_block_m() -> int:
...
+                native_scale_layout=_bm == 16,
             ),
+            stage2=stage2_func,
             block_m=_bm,
             ksplit=int(ksplit),
-            fuse_quant="fp4",
-            output_aux=True,
-            prequant=False,
+            fuse_quant=_p1["out_dtype"],
+            output_aux=AUX_SORT_OPUS,
```
Bernard-Liu: Should the tuner be updated to match this production default? _port_e2e() still passes output_aux=True, which selects the three-stage/adaptive path because _aux_uses_opus() requires the literal "opus". It also bypasses the fused-prequant path used in production, so for BM > 16 the tuner timings and candidate ranking may not reflect the default runtime pipeline.

aiter/configs/model_configs/kimik3_a4w4_tuned_fmoe.csv
```diff
@@ -12,7 +12,7 @@ gfx950,256,256,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4
...
-gfx950,256,16384,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,64,0,642.4294,flydsl_moe1_afp4_wfp4_bf16_t64x128x256_w3_bnt0_xcd4_fp4,0.1%,1074.0772,flydsl_moe2_layout_afp4_wfp4_bf16_t64x256x128_reduce_sbm64_sp801,0.0%,1716.5066,0,0,0,1261.09,2257.8,
-gfx950,256,32768,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,64,0,1111.7422,flydsl_moe1_afp4_wfp4_bf16_t64x128x256_w2_bnt0_xcd4_fp4,0.1%,2185.1161,flydsl_moe2_layout_afp4_wfp4_bf16_t64x256x128_reduce_sbm64_sp1601,0.0%,3296.8583,0,0,0,1313.17,1228.96,
+gfx950,256,4096,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,64,0,301.2728,flydsl_mxmoe_g1_a4w4_64x256x256_situv2_xcd2,2.1%,338.8907,flydsl_moe2_layout_afp4_wfp4_bf16_t64x256x128_reduce_sbm64_sp1601,0.0%,640.1635,0,0,0,845.36,5847.59,
```
Bernard-Liu: **The tuner cannot reproduce 64 of the 70 GEMM1 configs changed by this PR**

`Mxfp4FlydslTuner._g1_kname()` hard-codes A4W4 with BN/BK=256 and only varies
BM, `use_nt`, and inline quantization. `_candidate_rows()` also uses the FP4-only
`_SUPPORTED` set. Comparing the generated names with the 70 changed rows, only
6 are reproducible; the other 64 require BN128, XCD, K-wave, SiTUv2, or other
unsupported options.

This can overwrite optimized rows without benchmarking the existing config.
Kimi-K3 is worse: its 17 SiTUv2 rows are tuned as SiLU, and `_run_candidate()`
also uses a SiLU reference, so tuning passes but runtime later rejects the config
on activation mismatch and falls back.

Please use the shared kernel-name builder, expose the new tuning axes, handle
SiTUv2 explicitly, and benchmark the existing config before replacing it.
Fixing `native_scale_layout` alone does not resolve this coverage gap.

Bernard-Liu: **The tuner cannot reproduce 64 of the 70 GEMM1 configs changed by this PR**

`Mxfp4FlydslTuner._g1_kname()` hard-codes A4W4 with BN/BK=256 and only varies
BM, `use_nt`, and inline quantization. `_candidate_rows()` also uses the FP4-only
`_SUPPORTED` set. Comparing the generated names with the 70 changed rows, only
6 are reproducible; the other 64 require BN128, XCD, K-wave, SiTUv2, or other
unsupported options.

This can overwrite optimized rows without benchmarking the existing config.
Kimi-K3 is worse: its 17 SiTUv2 rows are tuned as SiLU, and `_run_candidate()`
also uses a SiLU reference, so tuning passes but runtime later rejects the config
on activation mismatch and falls back.

Please use the shared kernel-name builder, expose the new tuning axes, handle
SiTUv2 explicitly, and benchmark the existing config before replacing it.
Fixing `native_scale_layout` alone does not resolve this coverage gap.

aiter/fused_moe.py
fsx950223: Fixed in ba12497a8, and confirmed on gfx950 (MI355X) against this branch's previous HEAD `e610587fb` via `Mxfp4FlydslTuner._port_e2e` — the exact path you named:

```
                                       nan_frac   cos_diff
before  BM16  (tok16, 6144, 512, 257, 9)  1.0000       nan   FAIL
after   BM16  (tok16, 6144, 512, 257, 9)  0.0000  0.011943   PASS
before  BM16  (tok64, 7168, 2048, 32, 8)  0.2923       nan   FAIL
after   BM16  (tok64, 7168, 2048, 32, 8)  0.0000  0.011748   PASS
```

Root cause was the split default, so the fix removes the split rather than adding a third copy: the rule now lives once in `native_scale_layout_for(BM)` in `mxfp4_kname.py`, `_mxfp4_a4w4_stage1_fw` defaults the parameter to `None` and derives it from the kernel name, and the explicit binds in `get_2stage_cfgs` and the tuner are gone. AOT calls the same helper.

BM32/64/128 are unchanged by the fix (BM128, whose non-atomic epilogue is deterministic, is bit-identical before and after).

aiter/fused_moe.py
fsx950223: Fixed in ba12497a8. `_mxfp4_a4w4_stage2_fw` now takes a real `bias2` parameter and forwards it to `_flydsl_v2_stage2_wrapper`; the native mxmoe GEMM2 family raises instead of swallowing it.

`stage2_has_bias` no longer keys on `enable_bias` — it is `has_stage2_bias and is_layout_gemm2`, where `has_stage2_bias` is plumbed from the caller's actual `bias2 is not None` and `is_layout_gemm2` comes from `parse_flydsl_v2_gemm2_kernel`, i.e. the same parser `_mxfp4_a4w4_stage2_fw` dispatches on, so config time and run time cannot disagree.

Measured over all 92 shipped mxmoe rows, comparing resolved `get_2stage_cfgs` metadata before vs after:

| caller has | rows keeping MXMOE (before → after) |
|---|---|
| no bias | 92 → 92 (only the bogus `stage2_has_bias` flips True→False) |
| `bias2` | 92 → 23 |

The 23 are exactly the rows with a `flydsl_moe2_layout_` GEMM2 (glm5 19, kimik3 4); they now really forward `bias2`. The other 69 (kimik2 57, glm5 8, qwen3_5_397b 4) name a native `flydsl_mxmoe_g2_` GEMM2 that cannot take one, and are discarded with a warning instead of silently dropping it.

On GPT-OSS: there are no `mxmoe_g1` rows in `gptoss_*_tuned_fmoe.csv` today, so nothing regresses right now — but the implication of the above is that GPT-OSS has to be tuned onto a `flydsl_moe2_layout_` GEMM2 for `bias2` to survive.

aiter/fused_moe.py
fsx950223: Done as suggested, in ba12497a8 — the check moved into `get_2stage_cfgs` and is inverted. `has_stage1_bias` is plumbed from both `get_2stage_cfgs` call sites, and a config whose `kernelName1` lacks `_bias` while the model has a `bias1` is discarded with a warning and falls back to heuristics.

Verified over the 92 shipped mxmoe rows: with `bias1` present, all 92 are now rejected (none of the shipped names carry `_bias`) and fall back to `_flydsl_stage1_wrapper`, rather than proceeding with `bias1` dropped.

I kept the `_mxfp4_a4w4_stage1_fw` check as a cheap assertion for direct callers such as the tuner, which do not go through `get_2stage_cfgs`.

aiter/fused_moe.py
fsx950223: Removed in ba12497a8 — you're right that it only deferred the same `ValueError` from config-parse time to the first forward, so `_parse_mxfp4_g1_kname(kernelName1)` is now called unguarded.

aiter/fused_moe.py
fsx950223: Removed in ba12497a8. The soft check at the top is kept as the single policy (warn and fall back to heuristics); the unreachable `raise` is gone.

That check also grew the other config-level rejections from this review — unsupported aux shape, stage1 bias without `_bias`, stage2 bias without a layout GEMM2 — so there is now one place that decides whether a tuned MXMOE row is usable, with one policy.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
fsx950223: Removed in ba12497a8, both copies. Runtime now uses `p1["use_nt"]` from the kernel name, so the compile key no longer depends on runtime `n_tokens` and AOT can match it.

aiter/ops/flydsl/kernels/mxfp4_gemm1.py
fsx950223: I checked this rather than trusting the comment, and the hazard does not exist — but you were right that the comment was doing no work, and right about the normalization.

FlyDSL's cross-process cache is not keyed on the symbol name. `jit_function._jit_function_cache_key` hashes the launcher source, its dependency sources, and every scalar closure value reachable from it (`_collect_closure_scalar_vals`, which recurses into nested `@kernel` functions and records `int`/`float`/`bool`/`str`/`tuple` freevars). `gemm1_kernel` closes over all three floats. Measured:

```
situv2, (situ_beta, situ_linear_beta) = (1.0, 1.0) vs (4.0, 25.0)
  -> d03620ca0af1f9cf... vs 50abe9954491bb1f...   DIFFERENT
swiglu, swiglu_limit = 7.0 vs 10.0                DIFFERENT
```

The cache manager path is `{root}/{func_name}_{manager_key}/{sha256(cache_key)}.pkl`, so a differing key cannot collide on disk either. ba12497a8 replaces the vague comment with this, and adds a test that fails if FlyDSL ever changes that behaviour.

Your second point was a real bug. `silu` + `beta=2.0` also produces a distinct key (verified DIFFERENT) even though the kernel never reads beta, so it was a guaranteed AOT miss for a value with no effect. `_normalize_mxfp4_activation_params` now pins `situ_beta`/`situ_linear_beta` to 1.0 unless the activation is SiTUv2, and `swiglu_limit` to `None` (→ the 7.0 AOT default) unless it is SwiGLU, at both `get_2stage_cfgs` call sites and in the forward args.

aiter/aot/flydsl/mxfp4_moe.py
fsx950223: You're right — fixed in ba12497a8.

Runtime derives `D_INTER` from the stored weight width, and a non-256-aligned shard ships padded with the logical width carried separately in `w2.inter_real` (the same convention `is_mxfp4_moe_shape_supported` uses). So the AOT job now keys on the 256-padded `d_inter`, not the raw CSV `inter_dim` — not on `v2_d_inter`, since that pads to a backend-specific `tile_k`.

Measured end to end on the Kimi-K3 row (`inter_dim=384`), counting cache entries created by the runtime call after an AOT warm-up:

```
before  AOT compiles D_INTER=384, runtime asks 512  -> 1 new kernel   AOT MISS (JIT on the serving path)
after   AOT compiles D_INTER=512, runtime asks 512  -> 0 new kernels  AOT HIT
```

Only the 8 `inter_dim=384` rows change; for every 256-aligned shape `d_inter == inter_dim`.

aiter/ops/flydsl/kernels/mxfp4_gemm1.py
fsx950223: Removed in ba12497a8 — function, import, and the `zero_()` call site.

That also restores the lazy-import property you point out: `mxfp4_gemm1_kernels.py` no longer imports `kernels.mxfp4_gemm1` at module level, so it no longer pulls flydsl into import time, and the AOT module's reason for importing from that file is gone as well (`_effective_use_nt` was deleted in the same commit).

aiter/ops/flydsl/mxfp4_kname.py
fsx950223: Deleted in ba12497a8 — all three functions. They had no callers; nothing was wired up to them.

The module keeps only torch/JIT-free data: the name parsers, `MXFP4_G1_VARIANTS` (the deduplicated variant table), and `native_scale_layout_for` — the GEMM1/GEMM2 scale-layout contract, which is a fact about the kernel names rather than selection policy. No selection heuristics remain there.

On your question about the `token == 1` claim: I could not reproduce it. Running the port end to end at `token=1` on the shapes whose shipped CSV rows use BM16 there, 3 seeds each (cosine_diff, lower is better):

```
tok=1 inter=1024  BM16 0.011597 / 0.011607 / 0.011605   BM32 0.011610 / 0.011607 / 0.011611
tok=1 inter=512   BM16 0.012910 / 0.012931 / 0.012933   BM32 0.012918 / 0.012928 / 0.012922
tok=2 inter=512   BM16 0.012396 / 0.012391 / 0.012397   BM32 0.012396 / 0.012401 / 0.012397
```

BM16 and BM32 are indistinguishable at `token=1`, so the comment's premise does not hold and the shipped CSVs are right to use BM16 there. Deleting the function removes the claim along with it.

aiter/ops/moe_mxfp4_aux.py
fsx950223: Wired up in ba12497a8 — `get_2stage_cfgs` now calls it, and an uncovered shape becomes one of the config-rejection reasons (warn, discard the tuned row, fall back to heuristics) instead of reaching `AITER_CHECK`.

I checked it does not reject anything shipped: all 92 `mxmoe_g1` rows across `aiter/configs/**` pass the guard, and comparing resolved dispatch metadata before vs after, all 92 still dispatch MXMOE in the no-bias case.

The hand-maintained duplication of `gen_instances.SHAPES` is still there — I did not want to make the parse-time-clean `moe_mxfp4_aux` import the codegen script in this PR. It at least now has a caller that would exercise a drift.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
fsx950223: Deduplicated in ba12497a8, though in the other direction: both copies are gone and the single table is `MXFP4_G1_VARIANTS` in `mxfp4_kname.py`.

That module is already the torch/JIT-free one both sides can import, so it avoids `mxfp4_gemm1_kernels.py` reaching into `kernels/mxfp4_gemm1.py` at module scope — which is exactly the import-time flydsl load flagged in the `scale_out_uses_atomic` thread. `kernels/mxfp4_gemm1.py`, `mxfp4_gemm1_kernels.py`, and the tuner (`_candidate_rows`, via `MXFP4_G1_VARIANTS["fp4"]`) all read the one table now.

aiter/fused_moe.py
fsx950223: Removed in ba12497a8. The other (Opus) branches still use it; only the MXMOE branch, where it was always `{}`, drops it.

aiter/fused_moe.py
fsx950223: Split in ba12497a8 — the `v2_output_layout` line is now its own `if stage1_func is _flydsl_stage1_wrapper and metadata.skip_inter_quant:`, and the remaining block just handles the SiTUv2 scalars shared by both functions.

aiter/ops/flydsl/moe_kernels.py
fsx950223: Reverted to `shape[2]` in ba12497a8, keeping the assert — agreed the pair was contradictory, and a 3D intermediate is the real contract here.

fsx950223:
@binding7012 all 15 threads from your review are addressed in ba12497a8. Replies with the details are on each thread; the short version:

**The blocker.** You were right about the cause and the consequence. The rule now lives once — `native_scale_layout_for(BM)` in `mxfp4_kname.py` — and `_mxfp4_a4w4_stage1_fw` derives it from the kernel name when the caller says nothing, so the tuner cannot get a different answer from `get_2stage_cfgs`. Confirmed on gfx950 against this branch's previous HEAD `e610587fb`, through `Mxfp4FlydslTuner._port_e2e`:

```
                                          before              after
BM16   (tok16,  6144,  512, 257, 9)   nan=1.00   nan     nan=0.00  cos_diff=0.011943
BM16   (tok64,  7168, 2048,  32, 8)   nan=0.29   nan     nan=0.00  cos_diff=0.011748
BM32                                  0.011942 / 0.011749   0.011944 / 0.011750
BM64                                  0.011942 / 0.011749   0.011941 / 0.011749
BM128                                 0.011937 / 0.011743   0.011937 / 0.011743   (bit-identical)
```

**Bias.** `bias2` has a real parameter now and reaches `_flydsl_v2_stage2_wrapper`; the stage1 guard moved into `get_2stage_cfgs` and is inverted, as you suggested. Both bias predicates use `parse_flydsl_v2_gemm2_kernel` — the parser the stage2 dispatch itself uses — so config time and run time cannot disagree.

**AOT `D_INTER`.** Also right. Measured on the Kimi-K3 `inter_dim=384` row: before, the runtime call compiled a new kernel after the AOT warm-up (a JIT on the serving path); after, zero.

**`situ_beta` / `swiglu_limit`.** Here I have to push back on half of it. The cross-process cache is not keyed on the symbol name — `_jit_function_cache_key` hashes the launcher source plus every scalar closure value it reaches, and `gemm1_kernel` closes over all three floats; differing betas produce different keys, measured. The comment claiming this was doing no work, so it now states the mechanism and a test pins it. Your normalization point was a real bug and is fixed.

**`token == 1` → BM32.** I could not reproduce the accuracy claim: at `token=1`, BM16 and BM32 are indistinguishable across 3 seeds and 2 shapes (~0.0116 and ~0.0129 cosine_diff either way). The shipped CSVs are right to use BM16 there, and the function making the claim is deleted.

**Regression check.** Comparing resolved `get_2stage_cfgs` metadata before vs after across all 92 shipped `mxmoe_g1` rows: with no bias, all 92 still dispatch MXMOE and the only change is the bogus `stage2_has_bias` going True→False. Nothing lost its tuned config to the new `is_mxfp4_moe_shape_supported` gate.

One thing worth flagging that came out of the bias work: 69 of the 92 rows (kimik2 57, glm5 8, qwen3_5_397b 4) name a native `flydsl_mxmoe_g2_` GEMM2 that cannot carry `bias2`, so if a model on those shapes ever ships a bias it will now be rejected and fall back rather than silently drop it. GPT-OSS has no `mxmoe_g1` rows yet, so nothing regresses today, but it will need tuning onto a `flydsl_moe2_layout_` GEMM2 for bias to survive.

Verification caveats: the numbers above are from `Mxfp4FlydslTuner` on a single MI355X; the dispatch and shape-guard sweeps are CPU-side over the shipped CSVs; and I have not re-run a full tuning sweep or a serving-level benchmark. @Bernard-Liu's thread about `_port_e2e` still passing `output_aux=True` is untouched and left open.

aiter/fused_moe.py
```diff
@@ -2411,26 +2518,74 @@ def get_block_m() -> int:
...
         try:
-            _bm = _parse_mxfp4_g1_kname(kernelName1)["BM"]
+            _p1 = _parse_mxfp4_g1_kname(kernelName1)
+            _bm = _p1["BM"]
         except ValueError:
+            _p1 = {
+                "a_dtype": "fp4",
+                "out_dtype": "fp4",
+                "interleave": gate_mode == GateMode.INTERLEAVE,
+            }
```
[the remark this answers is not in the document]
fsx950223: Obsolete as of ba12497a8: the `except ValueError` fallback is gone entirely (it only deferred the same `ValueError` to the first forward), so `_parse_mxfp4_g1_kname(kernelName1)` is now called unguarded and the partial-dict `KeyError` path no longer exists.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
```diff
@@ -118,8 +307,25 @@ def flydsl_mxfp4_gemm1(
...
+        wide_mfma,
+        wide_accumulators,
+        num_waves,
+        k_wave,
     )
     grid = gemm1_grid(n_tokens, BM, NE=NE, TOPK=topk, INTER=D_INTER, BN=BN)
+    if BM == 16:
+        # BM16 scale dwords are assembled by multiple atomic writers. The GEMM
+        # uses the current/default stream, so this clear is ordered before it.
+        inter_sorted_shuffled_scale.zero_()
```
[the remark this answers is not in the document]
fsx950223: Obsolete as of ba12497a8: `scale_out_uses_atomic` always returned `False`, so the `zero_()` was dead. The function, its import, and the call site are all removed, so there is no cross-stream clear left to race.

aiter/fused_moe.py
```diff
@@ -3078,12 +3235,19 @@ def fused_moe_2stages(
...
-        if metadata.skip_inter_quant:
+    if stage1_func in (_flydsl_stage1_wrapper, _mxfp4_a4w4_stage1_fw):
+        if stage1_func is _flydsl_stage1_wrapper and metadata.skip_inter_quant:
             extra_stage1_args["v2_output_layout"] = True
-        # SiTUv2 beta/linear_beta -> stage1 runtime f32 scalars (not compile keys); None -> 1.0.
+        # SiTUv2 beta/linear_beta -> the stage1 situ_beta/situ_linear_beta params;
+        # None -> 1.0 (plain tanh). _flydsl_stage1_wrapper takes them as runtime
+        # f32 scalars; the mxfp4 a4w4 GEMM1 replacement bakes them into the
+        # compiled kernel (they are keys of _get_compiled_mxfp4_gemm1_port).
         extra_stage1_args["situ_beta"] = 1.0 if beta is None else float(beta)
```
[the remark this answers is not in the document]
fsx950223: Checked, and the premise is wrong — but the underlying concern was real and is fixed in ba12497a8.

FlyDSL's cache is not keyed on the kernel symbol name. `jit_function._jit_function_cache_key` hashes the launcher source, its dependency sources, and every scalar closure value reachable from it, and `gemm1_kernel` closes over `situ_beta` / `situ_linear_beta` / `swiglu_limit`. Measured: `(1.0, 1.0)` vs `(4.0, 25.0)` under SiTUv2 produce different keys, as do `swiglu_limit` 7.0 vs 10.0. Varying beta at runtime gets a correctly distinct kernel, not a wrong cached one.

What was actually broken is that the scalars were not normalized per activation, so `silu` with a stray `beta` created a cache entry for a value the kernel never reads — a guaranteed AOT miss with no effect. `_normalize_mxfp4_activation_params` now pins them unless the matching activation is selected.

aiter/fused_moe.py
```diff
@@ -1547,15 +1582,22 @@ def _mxfp4_a4w4_stage1(
...
+    inter_sorted_quant = torch.empty(
+        (max_sorted, inter_cols), device=device, dtype=inter_dtype
+    )
     inter_scale_rows = (inter_scale_bytes + inter_scale_cols - 1) // inter_scale_cols
     inter_scale_rows = (inter_scale_rows + 31) // 32 * 32
-    inter_sorted_shuffled_scale = _ia(
-        (inter_scale_rows, inter_scale_cols), device=device, dtype=torch.uint8
+    inter_scale_dtype = dtypes.fp8_e8m0 if out_dtype == "fp8" else torch.uint8
+    # The GEMM1 launcher clears BM16 scales before dispatch; larger tiles
+    # overwrite all scale bytes consumed by GEMM2.
```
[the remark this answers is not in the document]
fsx950223: Fixed in ba12497a8 by removing the thing the comment described: `scale_out_uses_atomic` and the conditional clear are both deleted, so the stale comment is gone with them.

aiter/ops/flydsl/mxfp4_kname.py
```diff
@@ -1,21 +1,59 @@
...
+        (32, True, False),
+        (32, False, False),
+        (64, False, False),
+        (128, False, False),
+        (16, True, True),
+    },
+}
+
+
+def native_scale_layout_for(BM: int) -> bool:
```
Bernard-Liu: > Should this also depend on out_dtype? MXFP4_G1_VARIANTS["fp8"] includes a BM16 inline-quant
  > variant, but this enables native scale layout for every BM16 kernel, while _assert_supported()
  > rejects native layout unless out_dtype == "fp4". A BM16 A8W4 kernel with _fp8out therefore fails
  > in both runtime and AOT. Could this be made output-dtype-aware, or should that variant be rejected
  > earlier?

aiter/aot/flydsl/mxfp4_moe.py
```diff
@@ -127,23 +152,54 @@ def _add(job):
...
+                            # width (``w1.shape[1] // 2``), and non-aligned
+                            # shards ship padded to a multiple of 256 with the
+                            # logical width carried separately in
+                            # ``w2.inter_real`` -- the same convention
+                            # ``is_mxfp4_moe_shape_supported`` uses. Keying on
+                            # the raw CSV ``inter_dim`` would miss those rows
+                            # (Kimi-K3: 384 stored as 512). ``d_inter`` is used
+                            # rather than the GEMM2 ``v2_d_inter`` because the
+                            # latter pads to a backend-specific tile_k.
+                            "D_INTER": d_inter,
```
Bernard-Liu: AOT compiles a different GEMM1 than the runtime asks for (Kimi-K3)

This rounds inter_dim up to a multiple of 256, but the runtime takes D_INTER = w1.shape[1] // 2 in _mxfp4_a4w4_stage1_fw — the physical width. D_INTER is in both the _get_compiled_mxfp4_gemm1_port cache key and the name_suffix built by compile_gemm1_a4w4_port, so 384 and 512 are two distinct compiled kernels. Running this PR's parse_csv over kimik3_a4w4_tuned_fmoe.csv, whose three mxmoe rows all carry inter_dim=384, every stage1 job comes out at 512:

flydsl_mxmoe_g1_a4w4_32x64x256_situv2_kw2_w2      D_HIDDEN=3584 D_INTER=512
flydsl_mxmoe_g1_a4w4_64x256x256_nt_situv2_xcd4    D_HIDDEN=3584 D_INTER=512
flydsl_mxmoe_g1_a4w4_64x256x256_situv2_xcd2       D_HIDDEN=3584 D_INTER=512
This does not depend on how any upstream caller pads its weights — aiter pins it down on its own. The get_2stage_cfgs lookup key uses the raw, unpadded inter_dim, and that value comes from get_inter_dim(w1.shape, w2.shape), i.e. the physical width. Since these three rows are keyed on 384, the only way they can ever be selected is with w1.shape[1] // 2 == 384. Conversely, if some deployment really did ship weights padded to 512, get_inter_dim would return 512, these rows would never match, mxmoe would never be dispatched at all, and the 512 AOT entry would still go unused. Either way the AOT cache misses.

Two different consequences depending on the deployment: with FLYDSL_RUNTIME_RUN_ONLY=1 this raises no usable AOT cache for launch_gemm1 (flydsl/compiler/jit_function.py:1439); without it there is no error, but the first inference JIT-compiles on the serving path.

Scope is limited to this one shape: across all 87 mxmoe GEMM1 rows under aiter/configs/, these three are the only ones whose inter_dim is not a multiple of 256.

Rather than precompiling both widths, I'd suggest giving AOT and the runtime a single source of truth — the runtime's D_INTER is exactly the inter_dim in the CSV lookup key, so AOT can just use the raw value. It would also be worth adding a regression test that warms the AOT cache and then runs inference under FLYDSL_RUNTIME_RUN_ONLY=1; there is currently none, which is what let the two key derivations drift apart unnoticed.

aiter/fused_moe.py
```diff
@@ -3211,13 +3446,33 @@ def fused_moe_2stages(
...
+        situ_beta=config_situ_beta,
+        situ_linear_beta=config_situ_linear_beta,
         opus_weights_shuffled=getattr(w1, "is_shuffled", False)
         and getattr(w2, "is_shuffled", False),
         config_file=_metadata_config_file,
     )
     if _metadata_transform is not None:
         metadata = _metadata_transform(metadata)
+    if output_aux:
+        metadata = replace(metadata, output_aux=output_aux)
```
Bernard-Liu: `fused_moe_2stages` currently accepts arbitrary truthy `output_aux` values and writes them directly into `metadata`, while `_fused_moe_impl` validates supplied values against
  `"threestage"` and `"opus"`. I verified that `fused_moe_2stages(..., output_aux="opuss")` completes without raising an error.

  This path already receives sorted tensors and never calls `_aux_uses_opus`, so the typo does not fall back to adaptive sorting. For MXMOE, it is instead treated as a
  non-`"opus"` value and follows the same prequant-selection branch as `"threestage"` at this layer.

  Please move the validation into a helper shared by both entry points, or add equivalent validation here before updating `metadata`.

aiter/fused_moe.py
Bernard-Liu, line 2009: > Blocker: this under-allocates inter_sorted_shuffled_scale for some BM16 native-scale shapes.
  >
  > With BM=16, use_nt=True, inline_quant=True, native_scale_layout=True, out_dtype=fp4, BN=128, D_INTER=1408, inter_scale_cols=44, but the kernel pads it to 48. For
  > max_sorted=512, this allocates 45,056 bytes while the kernel layout spans 49,152 bytes.
  >
  > I reproduced this on gfx950 with a sentinel guard: 3,712 guard bytes were overwritten, with the last write at +4093 B. Both D_INTER=1536 and allocating 49,152
  > bytes for D_INTER=1408 left the guard untouched.
  >
  > Please allocate using the padded native-scale stride, or reject this shape, and add a guard-buffer regression test.

Bernard-Liu: This catch-all makes the new activation guard blind to Gelu. `ActivationType.Gelu` and `GeluTanh` are real enum values (`csrc/include/aiter_enum.h:11,14`), so they map to `"silu"` here, and the check at line 2667 then compares `"silu" == "silu"` and accepts a SiLU-tuned row. The row is reachable because `_lookup_cfg` falls back to the activation-ignoring table whenever `config_file is None`.

Reproduced on gfx950 with `AITER_CONFIG_FMOE=aiter/configs/model_configs/kimik2_fp4_tuned_fmoe.csv`, `token=4, model_dim=7168, inter_dim=512, expert=384, topk=8`, a4w4:

```
Silu     -> kn1=flydsl_mxmoe_g1_a4w4_16x128x256_f16in_nt
Gelu     -> kn1=flydsl_mxmoe_g1_a4w4_16x128x256_f16in_nt   # same kernel, no warning
GeluTanh -> kn1=flydsl_mxmoe_g1_a4w4_16x128x256_f16in_nt   # same kernel, no warning
Swiglu   -> discarding MXMOE config: activation 'silu' does not match runtime 'swiglu'
The Swiglu line shows the guard itself works. That the kernel really ignores the requested activation is confirmed by calling `fused_moe` on identical inputs — GEMM2 is atomic, so a same-activation pair gives the run-to-run noise floor:

| pair | max abs diff | cosine | rel L2 |
|---|---:|---:|---:|
| Silu vs Silu (noise floor) | 1536 | 0.99999154 | 0.004118 |
| Silu vs Gelu | 1536 | 0.99999112 | 0.004226 |
| Silu vs Swiglu (rejected, different kernel) | 190900 | 0.64235663 | 0.995667 |

Note for the acceptance gate: against a true GELU reference this shape gives `logits_diff = 0.000342`, the same as the correct SiLU run, because at fp4 with K=7168 the pre-activation magnitudes are large enough that `silu(x) ≈ gelu(x) ≈ x`. So nothing in the current accuracy flow would flag it; it only shows up on a model with smaller pre-activations.

Suggestion: use `get_flydsl_activation_name()` (`aiter/ops/flydsl/moe_common.py:14-29`), which raises on unsupported activations, instead of defaulting to `"silu"`. The tuner's `_row_act` folds unknown activations the same way and would validate a Gelu row against a SiLU reference, so it needs the same treatment. This also matters post-`cf5cbed`, where `_mxfp4_a4w4_stage1_fw` decides whether to honour the caller's `swiglu_limit` based on the kernel name's activation tag.

Comment (Bernard-Liu) [inline aiter/fused_moe.py]: Warning-and-continue is not safe here: the tuned `kernelName1` is kept while `interleave` is flipped to the caller's value, and the two are not independent — several GEMM1 variants are compiled for separated layout only, so the mismatch turns into a hard failure rather than a slower kernel.

Reproduced on gfx950 with `AITER_CONFIG_FMOE=aiter/configs/model_configs/kimik3_a4w4_tuned_fmoe.csv`, `token=4, model_dim=3584, inter_dim=384, expert=896, topk=16`, SiTUv2 a4w4 (row 5 of that CSV):

```
gate_mode=separated: kn1=flydsl_mxmoe_g1_a4w4_32x64x256_situv2_kw2_w2
    name says interleave=False, metadata binds interleave=False, BN=64 k_wave=2 num_waves=2
    _assert_supported -> OK

gate_mode=interleave: kn1=flydsl_mxmoe_g1_a4w4_32x64x256_situv2_kw2_w2   # warning printed
    name says interleave=False, metadata binds interleave=True, BN=64 k_wave=2 num_waves=2
    _assert_supported -> NotImplementedError: flydsl mxfp4 GEMM1 BN64 is restricted to BM32 A4W4 non-inline separated
```

`k_wave > 1` carries the same separated-only requirement (`mxfp4_gemm1_kernels.py:165-168`), so this row violates two constraints at once. Even for the variants that do run under both layouts, the AOT job is keyed on the tuned name only, so the flipped flag gives a different compile key — a RUN_ONLY miss or a JIT on the first request.

Suggestion: on mismatch, discard the config and fall back to heuristics the same way the activation/shape/bias rejections above do, instead of warning and returning metadata that cannot launch. Alternatively make `gate_mode` part of the config lookup key and the AOT contract.

File: aiter/fused_moe.py [hunk 25]
```diff
@@ -2058,7 +2058,14 @@ def _mxfp4_a4w4_stage1_fw(
     p1 = _parse_mxfp4_g1_kname(kernelName1)
     runtime_situ_beta = float(situ_beta)
     runtime_situ_linear_beta = float(situ_linear_beta)
-    runtime_swiglu_limit = 7.0 if swiglu_limit is None else float(swiglu_limit)
+    # swiglu_limit reaches the kernel as a scalar closure value, so it lands in
+    # FlyDSL's disk-cache key (see the note in mxfp4_gemm1.py), yet
+    # _activation_mul_batch only consumes it for swiglu -- silu and situv2 ignore
+    # it entirely. Pin the other activations to the default so a caller's limit
+    # cannot fork the cache into entries whose generated code is identical.
+    runtime_swiglu_limit = 7.0
```
Bernard-Liu, line 2251: > swiglu_limit is still silently ignored for MXMOE SiLU.
  >
  > The public fused_moe contract and torch reference clamp the SiLU gate/up inputs when a finite limit is provided, but this path pins the value to 7.0 and
  > _activation_mul_batch does not consume it for SiLU.
  >
  > I reproduced this on gfx950 with the BM16 inline SiLU kernel: swiglu_limit=1.0 and 100.0 produced bit-identical payload and scale outputs, while the torch
  > reference differs. This makes the numerical result depend on which backend/config is selected.
  >
  > Please either implement the SiLU clamp in MXMOE or explicitly reject finite swiglu_limit before dispatch.

aiter/ops/flydsl/moe_kernels.py
```diff
@@ -1024,6 +1026,7 @@ def _run_moe_reduction(
     token_num,
     topk,
     model_dim,
+    model_dim_pad=0,
```
Bernard-Liu: Inserting `model_dim_pad` ahead of the existing optional params breaks a positional caller that this PR does not touch. `op_tests/flydsl_tests/test_flydsl_moe_a8w4.py:351-353` still passes seven positional args, so they now bind as `model_dim_pad=expert_mask`, `expert_mask=topk_ids`, `topk_ids=None` and trip the mutual-exclusion check right below:

```
E   ValueError: topk_ids is required when expert_mask is provided for reduce mode
aiter/ops/flydsl/moe_kernels.py:1041: ValueError
FAILED op_tests/flydsl_tests/test_flydsl_moe_a8w4.py::test_flydsl_stage2_fp8_ep_reduction
1 failed in 3.26s
```

Reproduced on gfx950; switching that call to keyword form makes it pass with the expected `out == 3.0`, so the failure is purely the argument shift.

The other five call sites are fine — `aot/flydsl/moe.py:820`, `aot/flydsl/mxfp4_moe.py:419` and `fused_moe.py:2389` use keywords, `moe_kernels.py:2280` was updated, and `gemm_moe_tune.py:1102` only passes the first five. So this is the only breakage.

Worth flagging that CI cannot catch it: `.github/scripts/split_tests.sh:53` and `aiter_test.sh` both collect with `find op_tests -maxdepth 1 -name 'test_*.py'`, and no workflow runs `op_tests/flydsl_tests/` at all. That is why the shards are green despite a deterministically failing test.

Suggestion: update the caller, and put new optional params after the existing ones (or make them keyword-only) so a future insertion cannot silently rebind arguments.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
```diff
@@ -7,50 +7,115 @@
...
+            )
+    if act == "swiglu" and swiglu_limit <= 0.0:
+        raise NotImplementedError(
+            f"flydsl mxfp4 gemm1 requires swiglu_limit > 0, got {swiglu_limit!r}"
+        )
+    if prefetch_hidden and not inline_quant:
+        raise NotImplementedError(
+            "flydsl mxfp4 GEMM1 hidden prefetch requires inline quantization"
+        )
     if D_HIDDEN % BK != 0:
```
Bernard-Liu, line 119: This is the only bound on `D_HIDDEN`, and it lets through a range the B-scale addressing does not actually cover. In `kernels/mxfp4_gemm1.py:767-782`, `K_C_HI = K_C // 16` is computed but the address only ever selects between two bases — `b_scale_s_base` and the fixed `b_scale_s_base_hi = base + 16 * kBS_stride_k0_dw * 4` (`:258`) — so the correct address `base + K_C * 256` only holds for `K_C < 32`:

```
K_C in [16, 32): base + 4096 + (K_C-16)*256 = base + K_C*256   ok
K_C in [32, 48): base + 4096 + (K_C-32)*256 = base + (K_C-16)*256   off by -16 tiles
```

At `K_C = 32` it re-reads the scale of K tile 16. Since `K_TILES_TOTAL = D_HIDDEN // BK` (`:1417`) and BK is effectively 256, the alias starts at `D_HIDDEN = 33 * 256 = 8448` — which is a multiple of 256 and therefore passes this check and every other validator. The address stays inside the B-scale buffer, so it is silent wrong math rather than a fault. No shipped model exceeds 8192, so it is latent, but the range is accepted today.

This addressing predates the PR; flagging it here because this is where the accepted range is defined. A cheap fence for now, until the base formula is generalized:

```python
if D_HIDDEN // BK > 32:
    raise NotImplementedError(
        "flydsl mxfp4 gemm1 B-scale addressing covers at most 32 K tiles, "
        f"got D_HIDDEN/BK={D_HIDDEN // BK} (H={D_HIDDEN})"
    )
```

aiter/fused_moe.py
fsx950223: Fixed in `a46b91cc7`. Reproduced the arithmetic: `out_as_per_chunk_dw_for` pads scale-N to a multiple of 8, so `D_INTER=1408` gives 44 unpadded vs 48 padded columns — 32 chunks x 1536 B = 49,152 B spanned against 45,056 B allocated, matching the ~4 KB of guard bytes you saw.

The allocation now takes the max with the padded stride:

```python
if native_scale_layout:
    from aiter.ops.flydsl.kernels.mxfp4_gemm1 import out_as_per_chunk_dw_for
    chunks = (max_sorted + BM - 1) // BM
    inter_scale_bytes = max(
        inter_scale_bytes, chunks * out_as_per_chunk_dw_for(D_INTER) * 4
    )
```

Swept the affected shapes: `D_INTER` 1408 and 1152 (44 and 36 columns, neither a multiple of 8) were both under-allocating and are now covered; 384/512/896/1024/1536/2048 are byte-for-byte unchanged. 1536 was already safe because 48 needs no padding, which is consistent with your observation that it left the guard untouched.

fsx950223: Fixed in `a46b91cc7`. Took the suggestion: `_mxfp4_activation_name` is gone and the guard now calls `get_flydsl_activation_name`, whose `ValueError` becomes a rejection like the shape/bias ones, so Gelu falls back to heuristics instead of matching a SiLU-tuned row.

```python
try:
    expected_act = get_flydsl_activation_name(activation)
except ValueError:
    expected_act = None
if expected_act is None:
    reject_reason = f"no MXMOE kernel for activation {activation!r}"
elif configured_act != expected_act:
    ...
```

Also fixed the tuner's `_row_act`, which had the same fold and would have tuned and accepted a Gelu row against a SiLU reference — it now raises instead of returning `"silu"`.

Your point about the acceptance gate is the part I'd have missed: `logits_diff = 0.000342` for both means nothing downstream would have flagged it, so the fix has to be the rejection itself rather than a tolerance.

fsx950223: Fixed in `a46b91cc7`, taking the first suggestion. A tuned config whose `interleave` disagrees with `gate_mode` is now discarded in the same chain as the activation/shape/bias rejections, so it falls back to heuristics instead of returning metadata that cannot launch:

```python
elif parsed_g1["interleave"] != (gate_mode == GateMode.INTERLEAVE):
    reject_reason = (...)
```

The warn-and-continue site is now unreachable for tuned configs, so for an explicitly supplied `kernelName1` it raises rather than warns — there is nothing to fall back to there, and failing at dispatch keeps the cause visible instead of surfacing as the `NotImplementedError` you hit inside `_assert_supported`.

I did not make `gate_mode` part of the config lookup key or the AOT contract; that is the larger of the two options you offered and seemed worth a separate change.

aiter/fused_moe.py
fsx950223: Fixed in `a46b91cc7`. Took the "explicitly reject" option rather than implementing the SiLU clamp, since rejecting the config lets the call fall back to a backend that does clamp — a correct result instead of an error:

```python
elif swiglu_limit is not None and expected_act != "swiglu":
    reject_reason = (
        f"MXMOE cannot apply swiglu_limit={swiglu_limit!r} to "
        f"activation {expected_act!r}"
    )
```

`get_2stage_cfgs` gained a `swiglu_limit=None` kwarg to make that reachable, forwarded from both call sites in `fused_moe.py`.

Note the 7.0 pin you were looking at stays, but its purpose is now only cache hygiene: `swiglu_limit` is a scalar closure value and therefore lands in FlyDSL's disk-cache key, while `_activation_mul_batch` ignores it for silu/situv2 — so without the pin two callers differing only in a limit that cannot affect codegen got separate cache entries, and missed the AOT artifact, which hardcodes 7.0. With the rejection above, a finite limit no longer reaches this path with a non-swiglu activation at all.

Implementing the clamp in MXMOE is the other half of your suggestion and I have deliberately not done it here — happy to if you would rather have the kernel support it than reject.

aiter/ops/flydsl/mxfp4_gemm1_kernels.py
fsx950223: Fixed in `a46b91cc7` — added your fence verbatim in spirit:

```python
if D_HIDDEN // BK > 32:
    raise NotImplementedError(
        "flydsl mxfp4 gemm1 B-scale addressing covers at most 32 K tiles, "
        f"got D_HIDDEN/BK={D_HIDDEN // BK} (H={D_HIDDEN}, BK={BK})"
    )
```

Confirmed the derivation against `issue_b_scale_load`: `s_off` selects `b_scale_s_base` only when `K_C_HI == 0` and `b_scale_s_base_hi` for everything else, so from `K_C = 32` the address is `base + (K_C-16)*stride` — inside the buffer, hence silent.

Checked the fence rejects nothing shipped, but it is closer than I expected: `qwen3_8_2400b` at `model_dim=8192` is exactly 32 K tiles, sitting on the boundary. The next size up would have hit the aliasing rather than the fence.

Agreed this predates the PR; fixing it here since this is where the accepted range is defined.

aiter/ops/flydsl/moe_kernels.py
fsx950223: Fixed, though it took me three commits to land on the right half of your suggestion.

`a46b91cc7` updated the caller and made the tail keyword-only. That was the weaker choice, and I verified it with a single-line grep that cannot see a call split across lines — so I missed that `moe_kernels.py:2281` passes eight arguments positionally and broke it. `6f7fa5cde` converted that caller.

`3c1d58f38` switches to your first option instead, which is strictly safer:

```python
    fp8_pitch_align=None,
    # Appended rather than inserted: everything above keeps the position it has
    # on main, so an existing positional caller cannot be silently rebound.
    model_dim_pad=0,
):
```

Every parameter above `model_dim_pad` now sits where it does on main, so this PR moves nothing. The original seven-positional call in `test_flydsl_moe_a8w4.py` binds correctly again on its own — I kept the keyword form anyway, but it is no longer load-bearing.

Re-verified by walking the AST of every call site against the parsed signature rather than grepping: all six bind cleanly, and none of the AOT paths need a change since `aiter/aot/` never passes `model_dim_pad` at all.

Your note about CI is the more useful finding and I have not addressed it — `op_tests/flydsl_tests/` is genuinely uncollected by both `split_tests.sh:53` and `aiter_test.sh`, so this whole directory is dark. That looks like a separate change to the workflows rather than something to fold in here; say the word if you would rather it went in this PR.

aiter/ops/flydsl/kernels/moe_reduce.py
```diff
@@ -39,6 +39,7 @@ def moe_reduction_kernel(
     i32_m_tokens: fx.Int32,
     topk: fx.Constexpr[int],
     model_dim: fx.Constexpr[int],
+    model_dim_pad: fx.Constexpr[int],
```
coderfeli: why add this?

coderfeli (approved):

PR #4198, opened 2026-07-11
Moe a8w4 tp4 tuning

lburzawa:
## Motivation

Tune DSv4 TP4 shapes.

## Technical Details

Block tuning and add use of OOB token for gather in prefill. Add persistent kernel for small K.

## Test Plan

Correctness and performance.

## Test Result

UTs pass and performance improves by at least 10% on TP4 shapes.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4530, opened 2026-08-03
[Triton MOE routing] Fix `memory access fault` surfacing in `0.1.16.post5` -> `0.1.19` update

fxmarty-amd:
## Motivation

This PR fixes a regression in updating vLLM CI from `aiter==0.1.16.post5` to `aiter==0.1.19`, surfaced due to 73f149082 from #4198. The root cause is that vLLM padding tokens may contain `inf`/`nan` values, and AITER seems not to test this case.

Failing CI in vLLM with updated `aiter==0.1.19`: [mi355_1: Quantized Models Test](https://buildkite.com/vllm/amd-ci/builds/11528/list?sid=019fb767-0fce-4570-b9af-a4ba0421b3d5&tab=output).

Hotfix on vLLM side: https://github.com/vllm-project/vllm/pull/50859

Running `tests/models/quantization/test_gpt_oss.py::test_gpt_oss_attention_quantization` on gfx950 with the AITER W4A8 MoE backend (`aiter_mxfp4_w4a8_moe.py`, which calls `aiter.ops.triton.moe.moe_routing.routing.routing` followed by two `moe_gemm_a8w4` stages) aborts with a `memory access fault` inside the routing sort.

Disabling HIP Graph in vLLM bypasses the issue.

**This PR makes `streaming_topk` always return within range expert id, to avoid memory access faults.**

## Technical Details

vLLM pads the token count to the captured graph size when running with HIP graphs: a batch of 7 real tokens is executed as 8. We verified in vLLM that padding tokens may contain `inf`, `-inf` or `NaN` values. This is normally benign as the padded rows are discarded.

However, the MoE routing MUST tolerate non-finite router logits on padding rows without affecting the real ones. It currently does not.

The root cause is in `streaming_topk`: for such a row, top-k elects expert ids `>= n_expts_tot`, i.e. experts that do not exist. Everything downstream — the counting sort, the histogram, the gather/scatter indices — is then addressing with an out-of-range expert.

### Why an out-of-range expert id is produced

`topk()` hardcodes `BLOCK_N = 128`, so the logits are processed at a padded width `n_cols_pad = cdiv(n_expts_tot, 128) * 128`. For a 32-expert model that is 128 columns, of which 32–127 are pure placeholders. `streaming_topk` loads them with `other=float("-inf")` and then packs each candidate into a single sortable integer:

```python
x = fpval_to_key(x.to(x_utype, bitcast=True))     # order-preserving float -> uint
x = (x.to(x_ultype) << 16) | offs_x_n[None, :]    # high bits = value, low 16 = column id
acc = tl.topk(x, N_EXPTS_ACT_PAD, dim=1)
```

`tl.topk` keeps the largest packed values, so ties on the value are broken by the column index — and the placeholder columns always hold the largest column indices.

The problem is that `-inf` is not strictly below every real logit under `fpval_to_key`. In bf16:

| logit | bits | key |
|---|---|---|
| `-inf` (real column) | `0xFF80` | `0x007F` |
| `-inf` (placeholder) | `0xFF80` | `0x007F` |
| `-NaN` | `0xFFC0` | `0x003F` |

A real `-inf` produces the *identical* key to the placeholder and loses the tie-break to the higher column index; a negative `NaN` sorts strictly *below* the placeholder. So as soon as fewer than `n_expts_act` real columns outrank the placeholder — which is exactly what a garbage padding row looks like — top-k returns column indices such as 126 and 127 as expert ids.

`streaming_topk` has masked with `-inf` unchanged since at least `v0.1.11`.

### Why that turns into a memory access fault

`_routing_compute_indx` / `_routing_compute_indx_fused` resolve each gate's destination slot with:

```python
gates = tl.load(TokensStart + expert)
gates += exclusive_run_lengths
...
tl.store(GatherIndx + gates, offs)
tl.store(GateScal + gates, gate_scal)
```

`TokensStart` is `token_offs_raw`, a view into `token_offs_combined = torch.empty((2, pad(n_expts_tot + 1)))` allocated in `_compute_expt_data_internal`. `_expt_data_compute_stage1` fills only `n_expts_tot + 1` entries per row. An expert id of 127 therefore reads *uninitialised* memory, and the value it gets back is used as a **store offset**. The load itself stays inside the allocation; the store is what faults. With a dirty caching-allocator block the value is on the order of `1e9`, which puts the store gigabytes past a 128-byte buffer.

So the invalid expert id decides *which address is read*, and `torch.empty` decides *what value comes back*, and that value becomes a write offset. On a fresh process the uninitialised page is zero-filled and the fault does not reproduce; in a long-running vLLM server the allocator's free list is full of stale tensor data and it reproduces every time.

Secondarily, `hist` is a vertical popcount over the per-token expert bitmask truncated to `[:n_expts_tot]`, so gates that elected a placeholder column are silently dropped from the histogram and `sum(hist) < n_gates`.

### Why v0.1.16.post5 does not fault, and what changed

The same invalid expert ids are produced on `v0.1.16.post5`, but they are harmless there prior to #4198.

`_compute_expt_data_internal` used `BLOCK = 128` prior to #4198 — the same constant as `topk()`'s `BLOCK_N`. Two consequences:

1. `pad(n_expts_tot + 1)` and, more importantly, the store width in `_expt_data_compute_stage1` covered exactly `[0, n_cols_pad)` — precisely the range an expert id can reach. For `n_expts_tot != 128` the kernel took the `EQUAL_BLOCK=False` branch, whose load is masked but whose **store is not**, writing all `BLOCK` lanes. `TokensStart[126]` therefore returned `sum(hist)` — a small, in-range value.
2. The resulting `gates = sum(hist) + run_length` is bounded above by `n_gates - 1` and always lands at or after `sum(hist)`, while the `moe_gemm_a8w4` tile map (built from `hist`) only ever reads `[0, sum(hist))`. The bad gates were unreachable.

Commit `73f149082` — "Moe a8w4 tp4 tuning" ([#4198](https://github.com/ROCm/aiter/pull/4198), 2026-07-24) — changed:

```diff
 def _compute_expt_data_internal(n_expts_tot, n_gates, block_m, device):
-    BLOCK = 128
+    BLOCK = triton.next_power_of_2(n_expts_tot)
```

For a power-of-two expert count this flips `EQUAL_BLOCK` to `True`, so stage 1 now writes only `n_expts_tot + 1` entries instead of a full 128-lane block, and it shrinks the allocation. Indices `[n_expts_tot + 1, n_cols_pad)` are left as raw `torch.empty`. `_compute_expt_data_internal`'s `BLOCK` and `topk()`'s `BLOCK_N` are two independent constants that silently must agree, and nothing asserts it; #4198 desynchronised them.

### Alternatives considered

- Restoring `BLOCK = 128`: needs comment from @lburzawa on the accuracy/speed impact with context on #4198. IMO this is not the best as it does not fix the issue of having wrong expert IDs in the first place in `topk`.
- Changing `token_offs_combined = torch.empty(...)` to `torch.zeros(...)`: This removes the fault but is not a valid fix. `TokensStart[bad_expert]` then returns `0`, so `gates = 0 + run_length` and the bad gates are written at slots `0..k`, which belong to **real** tokens: `gather_idx` for a real gate ends up pointing at the padded row, and `gate_scal` for a real gate becomes the padded row's `NaN` weight. That is precisely the `NaN`-contamination-of-unpadded-tokens symptom this PR is meant to remove — it converts a loud crash into silently wrong output. It also adds a memset to every routing call.

## Test Plan

`pytest op_tests/triton_tests/moe/test_moe_routing.py -s -vvvvv -k "test_topk_in_range or test_routing_cuda_graph_nan_inf"`

Hardware: gfx950 (MI355X), torch 2.11.0+gitd0c8b1f, triton 3.6.0, in the `vllm/vllm-openai-rocm:nightly` container. Branch tested against `main` at `8a9186d98` (`v0.1.19-39-g8a9186d98`).

## Test Result

- `test_routing_cuda_graph_nan_inf` results in `memory access fault` on `main`, fixed by this PR.
- `test_topk_in_range` results in out of range expert ids on `main`, fixed by this PR.

---

*AI assistance (Claude) was used to investigate the root cause, bisect the behavioural change to `73f149082`, and draft the fix and the tests. All changes have been reviewed line by line and the results above were reproduced by the submitter.*


op_tests/triton_tests/moe/test_moe_routing.py
```diff
@@ -108,6 +108,158 @@ def init_data(n_tokens, n_expts_tot, dtype=torch.float16, device="cuda"):
 n_tokens = [4, 7, 8, 64, 255, 256, 371, 911, 1023, 1024, 4096, 8192]
 
 
+@pytest.mark.parametrize("n_tokens", [8, 16, 24, 32])
+@pytest.mark.parametrize("n_expts_tot", [32, 64, 128])
+def test_routing_cuda_graph_nan_inf(n_tokens: int, n_expts_tot: int):
```
lburzawa: I think the other UT function addresses the root cause of the bug. If so I would rather not run the full moe stack inside routing UTs as this function does.

lburzawa (changes requested):
Thanks for catching this! Leaving one comment but otherwise should be good to go.

op_tests/triton_tests/moe/test_moe_routing.py
fxmarty-amd: Removed this one in https://github.com/ROCm/aiter/pull/4530/commits/478fe1e170ab5257b02e581c1f27e04481babd00

lburzawa (approved):

azaidy (approved):
LGTM!

lburzawa (approved):

PR #4545, opened 2026-08-04
Fix Triton cache flooding in `_fwd_kernel_stage2_asm` (batch size as `constexpr`)

amd-mvarjoka:
# Fix Triton cache flooding in `_fwd_kernel_stage2_asm` (batch size as `constexpr`)

## Summary

`mla_decode_fwd` in `aiter/mla.py` launches the Triton kernel
`_fwd_kernel_stage2_asm` with the **batch size** passed as a `tl.constexpr`.
Because Triton bakes each `constexpr` value into a separately compiled artifact,
a new kernel is compiled — and a new Triton cache directory written — for every
distinct batch size the kernel sees.

Under a continuous-batching inference server the decode batch size changes on
almost every step, so this recompiles continuously and floods the Triton cache
(`/tmp/torchinductor_root/triton/`) with tens of thousands of
`_fwd_kernel_stage2_asm.*` directories, slowing the container down over time.

## Root cause

```python
@triton.jit
def _fwd_kernel_stage2_asm(
    ...
    page_size: tl.constexpr,
    KV_INDPTR_IS_PAGE_LEVEL: tl.constexpr,
    MAYBE_FINAL_OUT: tl.constexpr,
    HAS_FINAL_LSE: tl.constexpr,
    USE_VALID_SPLIT_COUNT_REDUCE: tl.constexpr,
    BATCH_NUM: tl.constexpr,   # = bs, baked into each compiled artifact
    BLOCK_DV: tl.constexpr,
    Lv: tl.constexpr,
    mgc: tl.constexpr,
):
```

`BATCH_NUM` is used in only two places, neither of which needs a compile-time
constant:

- a pointer offset — `tl.load(num_kv_splits_indptr + BATCH_NUM)`
- an integer comparison — `num_max_kv_splits == BATCH_NUM`

Both work identically with a runtime scalar. Every other constexpr in the
signature is bounded (`page_size` / `BLOCK_DV` / `Lv` fixed per model,
`mgc ∈ {16,32,64}`, and the `KV_INDPTR_IS_PAGE_LEVEL` / `MAYBE_FINAL_OUT` /
`HAS_FINAL_LSE` / `USE_VALID_SPLIT_COUNT_REDUCE` booleans), so only `BATCH_NUM`
causes unbounded recompilation.

The kernel is launched from **two** call sites in `aiter/mla.py` — the classic
decode path (`mla_decode_fwd`, `BATCH_NUM=bs`) and the newer MLA-v4 "nm" decode
path (`BATCH_NUM=num_seqs`) — so both feed the same per-batch-size recompile.

## Fix

Make `BATCH_NUM` a runtime argument and add `do_not_specialize` so Triton's
automatic divisibility-by-16 / `== 1` specialization does not reintroduce
variants:

```python
@triton.jit(do_not_specialize=["BATCH_NUM"])
def _fwd_kernel_stage2_asm(
    ...
    stride_lse_bs: tl.int64,
    BATCH_NUM,                 # runtime scalar (moved out of the constexpr block)
    page_size: tl.constexpr,
    KV_INDPTR_IS_PAGE_LEVEL: tl.constexpr,
    MAYBE_FINAL_OUT: tl.constexpr,
    HAS_FINAL_LSE: tl.constexpr,
    USE_VALID_SPLIT_COUNT_REDUCE: tl.constexpr,
    BLOCK_DV: tl.constexpr,
    Lv: tl.constexpr,
    mgc: tl.constexpr,
):
```

Both launch sites already pass `BATCH_NUM=` by keyword, so no call-site change is
needed. The kernel now compiles a small fixed number of variants instead of one
per batch size.

## Similar fixes in the Aiter repository

Converting a per-shape kernel argument off `tl.constexpr` (optionally with
`do_not_specialize`) to stop redundant recompilation is an established, already
accepted pattern in this repo, not a new idea:

- [#2427](https://github.com/ROCm/aiter/pull/2427) *(merged)* — "edit
  batched_gemm_a8w8 and gemm_a16_w16 kernel args for no recompile": the same
  motivation (kill per-shape recompiles) applied to the GEMM kernels via
  `do_not_specialize`.
- [#2482](https://github.com/ROCm/aiter/pull/2482) *(merged)* — follow-up tuning
  of the GEMM `do_not_specialize` / heuristic path.

## Related

The same class of bug (batch size as `tl.constexpr`) independently affects the
`_get_valid_kv_indices_kernel` kernel in SGLang
(`python/sglang/kernels/ops/attention/dsa/triton_kernel.py`), which is the
caller of `mla_decode_fwd` on its DSA `aiter` attention backend. That is fixed
in a companion SGLang PR; both fixes are needed to fully stop the cache growth on
that serving path.

Sglang PR #33506: [https://github.com/sgl-project/sglang/pull/33506](https://github.com/sgl-project/sglang/pull/33506)

## Note on the AOT stub

`aiter/aot/triton/decode_mla.py` contains a hardcoded `signature` for
`_fwd_kernel_stage2_asm` that is already stale (15 entries vs. the kernel's
current parameter list, predating `Final_lse` / `HAS_FINAL_LSE`). It is not the
runtime path exercised here (the JIT launch in `mla.py` is). If the AOT flow is
still used, that signature should be regenerated separately; this PR does not
touch it.

## Validation

Because aiter is AMD/ROCm-only, validate on a GPU host:

1. Clear the stale cache once: `rm -rf /tmp/torchinductor_root/triton/*`.
2. Run a decode workload whose batch size varies (e.g. a continuous-batching
   benchmark sweep).
3. Confirm `/tmp/torchinductor_root/triton/` stabilizes at a small fixed count of
   `_fwd_kernel_stage2_asm` directories instead of growing per batch size, and
   that decode output is numerically unchanged.

## Reproduction

This kernel is exercised by SGLang's DSA `aiter` attention backend; the repro
below drives it end to end. (The paired `_get_valid_kv_indices_kernel` fix lives
in the companion SGLang PR — see Related.)

### Server

```bash
docker run \
  -d \
  --name=sglang \
  --ipc=host \
  --network=host \
  --shm-size=128G \
  --privileged \
  --group-add video \
  --security-opt seccomp=unconfined \
  --cap-add=CAP_SYS_ADMIN \
  --cap-add=SYS_PTRACE \
  --device=/dev/kfd \
  --device=/dev/dri \
  --workdir=/workdir \
  --entrypoint=sglang \
  -e SGLANG_USE_AITER=1 \
  -e ROCM_QUICK_REDUCE_QUANTIZATION=INT4 \
  lmsysorg/sglang-rocm:v0.5.16-rocm720-mi35x-20260728 \
  serve \
  --model deepseek-ai/DeepSeek-V3.2 \
  --port 8000 \
  --tensor-parallel-size 8 \
  --attention-backend dsa \
  --dsa-prefill-backend aiter \
  --dsa-decode-backend aiter \
  --trust-remote-code \
  --disable-radix-cache
```

### Benchmark

```bash
#!/usr/bin/env bash
set -euo pipefail

# Each config: isl osl mc np warmup_requests seed
benchmark_configs=(
    "1024  1024 4   32  4   1"
    "8192  1024 4   32  4   2"
    "1024  1024 8   64  8   3"
    "8192  1024 8   64  8   4"
    "1024  1024 16  96  16  5"
    "8192  1024 16  96  16  6"
    "1024  1024 32  128 32  7"
    "8192  1024 32  128 32  8"
    "1024  1024 64  256 64  9"
    "8192  1024 64  256 64  10"
    "1024  1024 128 384 128 11"
    "8192  1024 128 384 128 12"
    "1024  1024 256 768 256 13"
    "8192  1024 256 768 256 14"
    "60000 600  4   32  4   15"
    "60000 600  16  96  16  16"
)

for config in "${benchmark_configs[@]}"; do
    read -r isl osl mc np warmup_requests seed <<< "$config"

    echo "Running benchmark: isl=$isl osl=$osl mc=$mc np=$np warmup_requests=$warmup_requests seed=$seed"

    python3 -m sglang.bench_serving \
        --backend sglang \
        --base-url http://127.0.0.1:8000 \
        --port 8000 \
        --model deepseek-ai/DeepSeek-V3.2 \
        --tokenizer deepseek-ai/DeepSeek-V3.2 \
        --dataset-name random \
        --random-input-len "$isl" \
        --random-output-len "$osl" \
        --max-concurrency "$mc" \
        --num-prompts "$np" \
        --warmup-requests "$warmup_requests" \
        --seed "$seed" \
        --request-rate inf \
        --random-range-ratio 0.8
done
```

### Impact

Measured with `docker diff sglang` (counts filesystem changes in the container,
dominated by the Triton cache under `/tmp/torchinductor_root/triton/`) after
running the full benchmark sweep above:

| | `docker diff` changes |
| --- | --- |
| Before (no patches) | ~334,000 |
| After (both patches — this PR + the SGLang PR) | ~94,000 |

Both patches are required to reach the lower figure, since each fixes one of the
two per-batch-size kernels on this path.

Performance should be similar or better for all the benchmark scenarios. We measured on average ~5% uplift in total token throughput metric for the benchmark sweep above if both of the patches were included.


azaidy (approved):
LGTM!

k50112113 (commented):
LGTM

valarLip (approved):

PR #4552, opened 2026-08-04
[gfx1250] Route the missing Kimi-K3 fused BF16 GEMM to Triton

XiaobingSuper:
This PR is routing only: 15 rows in `aiter/configs/model_configs/kimik3_bf16_tuned_gemm.csv`, no code.

## What is missing today

`get_GEMM_A16W16_config` looks the shape up in the model CSV and returns `libtype=\"torch\"` when there is no row. The fused Kimi-K3 shape below has no row on gfx1250 tp4, so it goes to `F.linear` and hipBLASLt:

| N | K | what it is |
|---|---|---|
| 12440 | 7168 | the fused KDA `in_proj` GEMM |

`N=12440 = 4*3072 + 24 + 128` is what ATOM's `_fuse_in_proj` builds by concatenating `in_proj + b_proj + f_a_proj`. The CSV is tuned for the *unfused* `12/24/128 x 7168` shapes, so the fused N is simply a miss — not a deliberate torch routing.

It is the only untuned bf16 shape in the whole run, and it is expensive. hipBLASLt picks `Cijk_..._MT32x16x32_MI16x16x1`: **5.3 ms/layer at M=2048**, and 69 KDA layers make that **367 ms = 34.6% of a 1060 ms prefill**.

## Routing scope

Route `N=12440, K=7168` for the full model M ladder: `1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384`.

`N=2112, K=7168` is intentionally not added here because it is already globally routed by `dsv3_bf16_tuned_gemm.csv`; adding it again creates duplicate merge keys.

Tuned Triton configs are provided separately by #4537.


carlushuang (approved):
LGTM

PR #4560, opened 2026-08-04
[BUGFIX] Avoid CustomAllreduce when expandable_segments:True

### LINKED ISSUE #4174: [Bug] custom_all_reduce.cu:397 hipIpcGetMemHandle fails when caller uses PYTORCH_HIP_ALLOC_CONF=expandable_segments:True
# [Bug] `custom_all_reduce.cu:397 hipIpcGetMemHandle [HIP error](invalid argument)` when caller uses `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True`

## Summary

`aiter.get_meta_buffer_ipc_handle` (backed by `hipIpcGetMemHandle` at
[aiter_meta/csrc/kernels/custom_all_reduce.cu:397](https://github.com/ROCm/aiter/blob/main/csrc/kernels/custom_all_reduce.cu#L397))
fails with `[HIP error](invalid argument)` when the input buffer was
allocated by PyTorch's caching allocator with
`PYTORCH_HIP_ALLOC_CONF=expandable_segments:True`. This env var switches
PyTorch to `hipMallocAsync` / stream-ordered memory pool allocations, and
pointers into those pools are not compatible with the classic
`hipIpcGetMemHandle` API on ROCm 7.2.x.

Downstream: any vLLM TP≥2 configuration that opts into
`expandable_segments:True` (which vLLM does in several code paths to
mitigate HBM fragmentation on big-model tensor-parallel serving) fails
at `CustomAllreduce` init.

## Environment

- GPU: AMD Instinct MI300X (`gfx942`)
- ROCm 7.2.53211
- PyTorch `2.11.0+gitd0c8b1f`
- `amd-aiter 0.1.16.post3` (also reproduced on `0.1.13.post1` and
  `0.1.16.post2` — the code path is unchanged across those versions)
- Container: `vllm/vllm-openai-rocm:nightly` (image digest
  `sha256:b678b9cd2cc402…`)

## Minimal repro (single process, no distributed)

```python
# aiter_ipc_repro.py
import os, torch, aiter
torch.cuda.set_device(0)
buf = torch.empty(1 << 30, dtype=torch.uint8, device="cuda")
handle = torch.empty(64, dtype=torch.uint8)
aiter.get_meta_buffer_ipc_handle(buf.data_ptr(), handle.data_ptr())
print("OK")
```

```
$ python aiter_ipc_repro.py
OK

$ PYTORCH_HIP_ALLOC_CONF=expandable_segments:True python aiter_ipc_repro.py
[AITER] /app/aiter/aiter_meta/csrc/kernels/custom_all_reduce.cu:397 fail to
call hipIpcGetMemHandle((hipIpcMemHandle_t*)out_handle_ptr, (void*)inp_ptr)
---> [HIP error](invalid argument)
```

## Where it manifests in vLLM

At `CustomAllreduce.__init__` in
`aiter/dist/device_communicators/custom_all_reduce.py`, the "input" meta
buffer is allocated with `torch.empty(...)` on the current CUDA stream.
Under `expandable_segments:True` that pointer lives in an `hipMallocAsync`
pool. The subsequent `get_meta_buffer_ipc_handle` call fails and takes
down all TP workers before model load completes.

We first hit this on GLM-5.1-FP8 / DeepSeek-V3.2 TP=8 startup, where
vLLM's OOM-hardening path sets `expandable_segments:True` and then
initializes AITER CustomAllreduce.

## What we ruled out

- Not a per-buffer alignment issue: buffer size / dtype don't matter (1 MiB,
  1 GiB, uint8, int32 — all fail with `expandable_segments:True`).
- Not fixed by switching only the "input" buffer to `uncached=True` in
  aiter's Python wrapper. That eliminates the first call site, but a
  second IPC call in graph-buffer registration
  (`custom_all_reduce.cuh::register_graph_buffers`) then fails identically
  on user-supplied tensors that also came from the pool.
- Not resolved by the recent consolidation of vLLM's Custom Allreduce onto
  AITER's path (vllm-project/vllm#46065, merged). That PR moved callers
  onto AITER without touching the pool-pointer / IPC compatibility.

## Suggested fix directions

The classic `hipIpcGetMemHandle` API is documented to work on pointers
returned by `hipMalloc`. Pointers from `hipMallocAsync` / stream-ordered
pools need `hipMemPoolExportToShareableHandle` +
`hipMemPoolImportFromShareableHandle` on the receiver, with the
pool-native IPC flow.

Two plausible fixes on the AITER side:

1. **Detect and fall back.** In `CustomAllreduce.__init__`, check the
   allocation origin of the input tensor (e.g. via
   `hipPointerGetAttributes` or `HIP_POINTER_ATTRIBUTE_MEMPOOL_HANDLE`).
   If it comes from a pool, `hipMalloc` a fresh IPC-compatible buffer
   and copy into it — or refuse-to-attach with a clear diagnostic so
   callers can `--disable-custom-all-reduce` explicitly.
2. **Support pool-based IPC.** Use `hipMemPoolExportToShareableHandle` /
   `hipMemPoolImportFromShareableHandle` when the pointer originates
   from a pool. Cleaner but requires reworking the receiver side too.

Direction 1 is the smaller change and unblocks vLLM immediately. It also
lets `expandable_segments:True` remain enabled at the caller (which
helps memory-tight big-model TP configs stay within budget).

## Workarounds today

- `--disable-custom-all-reduce` at vLLM serve time.
- Unset `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True` — but this
  re-introduces fragmentation-induced OOMs at big-model TP=8 startup.

## Repro artifact

Self-contained script attached: `aiter_ipc_issue_repro.py` (runs both
pass and fail cases, no distributed init required).

jamesETsmith:
## Motivation

Fixes #4174.

AITER CustomAllreduce aborts during initialization when PyTorch uses `expandable_segments:True`. PyTorch expandable-segment allocations use HIP virtual memory management and cannot be exported through AITER's classic `hipIpcGetMemHandle` path.

The goal is to preserve correctness and avoid terminating TP workers. In this configuration, collectives fall back to RCCL instead of using CustomAllreduce.

## Technical Details

Before initializing the classic HIP IPC transport, CustomAllreduce now checks whether PyTorch expandable segments are enabled.

The active setting is read from PyTorch's allocator snapshot, allowing programmatic allocator changes to be detected. Environment-variable parsing is used as a fallback and follows PyTorch's precedence:

1. `PYTORCH_CUDA_ALLOC_CONF`
2. `PYTORCH_HIP_ALLOC_CONF`
3. `PYTORCH_ALLOC_CONF`

When expandable segments are enabled for the classic IPC transport, CustomAllreduce remains disabled and emits an actionable warning. Existing callers then use their RCCL fallback. AITER's VMM transport is unaffected.

Regression tests cover allocator snapshot detection, environment fallback, variable precedence, repeated settings, and disabled communicator behavior.

## Test Plan

- Run focused regression tests on an AMD Instinct MI355X.
- Run Ruff and Black against the changed files.
- Compile the changed Python files.
- Check the diff for whitespace errors.

## Test Result

- `pytest op_tests/test_custom_all_reduce_allocator.py -q`: **8 passed** on MI355X, Slurm job `37581`.
- `ruff check aiter/dist/device_communicators/custom_all_reduce.py op_tests/test_custom_all_reduce_allocator.py`: passed.
- `black --check aiter/dist/device_communicators/custom_all_reduce.py op_tests/test_custom_all_reduce_allocator.py`: passed.
- `python3 -m py_compile aiter/dist/device_communicators/custom_all_reduce.py op_tests/test_custom_all_reduce_allocator.py`: passed.
- `git diff --check`: passed.

## Submission Checklist

### DIFF
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -38,6 +38,30 @@ def _env_flag(name: str) -> bool:
     return os.environ.get(name, "").strip().lower() in ("1", "true", "yes", "on")
 
 
+def _expandable_segments_enabled() -> bool:
+    """Return the active PyTorch expandable-segments allocator setting."""
+    try:
+        snapshot = torch.cuda.memory._snapshot()
+        return bool(snapshot["allocator_settings"]["expandable_segments"])
+    except (AttributeError, KeyError, RuntimeError):
+        pass
+
+    for name in (
+        "PYTORCH_CUDA_ALLOC_CONF",
+        "PYTORCH_HIP_ALLOC_CONF",
+        "PYTORCH_ALLOC_CONF",
+    ):
+        if name not in os.environ:
+            continue
+        enabled = False
+        for setting in os.environ[name].split(","):
+            key, separator, value = setting.partition(":")
+            if separator and key.strip() == "expandable_segments":
+                enabled = value.strip().lower() == "true"
+        return enabled
+    return False
+
+
 def _detect_gfx1250() -> bool:
     # Escape hatch for validating the old-arch (IPC + old kernel) path on gfx1250
     # hardware: forces the non-gfx1250 code path end to end.
@@ -721,6 +745,14 @@ class CustomAllreduce:
         if not custom_ar:
             # disable because of missing custom allreduce library
             # e.g. in a non-cuda environment
+            return
+
+        if not self._use_vmm and _expandable_segments_enabled():
+            logger.warning(
+                "Custom allreduce is disabled because PyTorch expandable-segment "
+                "allocations cannot be exported with hipIpcGetMemHandle; falling "
+                "back to RCCL."
+            )
             return
 
         self.group = group
--- a/op_tests/test_custom_all_reduce_allocator.py
+++ b/op_tests/test_custom_all_reduce_allocator.py
@@ -1,0 +1,78 @@
+import pytest
+
+from aiter.dist.device_communicators import custom_all_reduce
+
+
+@pytest.mark.parametrize("value", [True, False])
+def test_expandable_segments_uses_allocator_snapshot(monkeypatch, value):
+    monkeypatch.setenv("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:False")
+    monkeypatch.setattr(
+        custom_all_reduce.torch.cuda.memory,
+        "_snapshot",
+        lambda: {"allocator_settings": {"expandable_segments": value}},
+    )
+
+    assert custom_all_reduce._expandable_segments_enabled() is value
+
+
+@pytest.mark.parametrize(
+    ("name", "config", "expected"),
+    [
+        ("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True", True),
+        (
+            "PYTORCH_HIP_ALLOC_CONF",
+            "max_split_size_mb:64, expandable_segments:True",
+            True,
+        ),
+        ("PYTORCH_ALLOC_CONF", "expandable_segments:False", False),
+        (
+            "PYTORCH_ALLOC_CONF",
+            "expandable_segments:False,expandable_segments:True",
+            True,
+        ),
+    ],
+)
+def test_expandable_segments_falls_back_to_environment(
+    monkeypatch, name, config, expected
+):
+    for variable in (
+        "PYTORCH_CUDA_ALLOC_CONF",
+        "PYTORCH_HIP_ALLOC_CONF",
+        "PYTORCH_ALLOC_CONF",
+    ):
+        monkeypatch.delenv(variable, raising=False)
+    monkeypatch.setenv(name, config)
+    monkeypatch.setattr(
+        custom_all_reduce.torch.cuda.memory,
+        "_snapshot",
+        lambda: (_ for _ in ()).throw(RuntimeError),
+    )
+
+    assert custom_all_reduce._expandable_segments_enabled() is expected
+
+
+def test_cuda_allocator_config_has_pytorch_precedence(monkeypatch):
+    monkeypatch.setenv("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:False")
+    monkeypatch.setenv("PYTORCH_HIP_ALLOC_CONF", "expandable_segments:True")
+    monkeypatch.setattr(
+        custom_all_reduce.torch.cuda.memory,
+        "_snapshot",
+        lambda: (_ for _ in ()).throw(RuntimeError),
+    )
+
+    assert custom_all_reduce._expandable_segments_enabled() is False
+
+
+def test_custom_allreduce_stays_disabled_for_expandable_segments(monkeypatch, caplog):
+    monkeypatch.setattr(custom_all_reduce, "custom_ar", True)
+    monkeypatch.setattr(custom_all_reduce, "_is_gfx1250", False)
+    monkeypatch.setattr(custom_all_reduce, "_use_vmm", False)
+    monkeypatch.setattr(custom_all_reduce, "_expandable_segments_enabled", lambda: True)
+    monkeypatch.setattr(
+        custom_all_reduce.CustomAllreduce, "_select_ops", lambda self: None
+    )
+
+    communicator = custom_all_reduce.CustomAllreduce(group=None, device=0)
+
+    assert communicator.disabled is True
+    assert "falling back to RCCL" in caplog.text


jamesETsmith:
I don't think the [CI error](https://github.com/ROCm/aiter/actions/runs/30939512818/job/92094473865?pr=4560) is related to my changes:

```
Finished uploading artifact content to blob storage!
SHA256 digest of uploaded artifact zip is ad96480a43d7a33d493b04426fab3b9f09fb5d9934a544f049ccfaee6a3a5a01
Finalizing artifact upload
Error: Failed to FinalizeArtifact: Unable to make request: ECONNRESET
If you are using self-hosted runners, please make sure your runner has access to all GitHub endpoints: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github
```

zufayu:
@TennyWang1223 — two things here: one concrete, one that needs your judgement.

## 1. The new test cannot pass, and CI never ran it

`test_custom_allreduce_stays_disabled_for_expandable_segments` asserts `"falling back to RCCL" in caplog.text`. That can never be true: `aiter/__init__.py:44` sets `logger.propagate = False` on the `aiter` logger — the one `custom_all_reduce.py:30` imports — while pytest's `caplog` installs its handler on the **root** logger, so it never sees those records. Reproduced with the repo's verbatim `getLogger()` and the test's verbatim body:

```
E   AssertionError: assert 'falling back to RCCL' in ''
E    +  where '' = <LogCaptureFixture>.text
----------------------------- Captured stderr call -----------------------------
[aiter] Custom allreduce is disabled because PyTorch expandable-segment allocations ... falling back to RCCL.
```

The warning *is* emitted (through aiter's own handler, to stderr), but `caplog.text` is empty.

CI didn't catch it because `build_aiter_wheels (3.12, rocm/pytorch:latest)` failed on an artifact-upload ECONNRESET, and `split_aiter_tests` / `Standard Tests (1 GPU)` / `Multi-GPU Tests (8 GPU)` all declare `needs: build_aiter_wheels`, so all three were skipped. `.github/scripts/split_tests.sh` scans `op_tests/` dynamically, so the file will land in a shard after merge and turn main red.

The sibling assertion `assert communicator.disabled is True` is also vacuous — `disabled` is already `True` at `custom_all_reduce.py:740`, so any early return satisfies it. The caplog line is the only one that actually exercises the new branch.

Fix: capture `logging.getLogger("aiter")` directly (attach `caplog.handler` to it, or flip `propagate` for the duration of the test), or monkeypatch `custom_all_reduce.logger.warning`; and re-run the failed build job so Standard Tests actually go green before merge.

## 2. Is disabling CustomAllreduce outright the right scope?

This is the part I can't answer from the diff. Not everything CustomAllreduce IPC-exports comes from the PyTorch allocator:

| buffer | allocated by | affected by `expandable_segments` |
|---|---|---|
| `meta` | raw HIP (`ops.allocate_meta_buffer`, `uncached=True`) | no |
| `input` pool | `torch.empty` (`_init_ipc`: `self._pool.create("input", max_size)`) | **yes** |
| user tensors, CUDA-graph capture only | PyTorch | **yes** |

The third row is already avoidable today: `enable_register_for_capturing = False` routes capture through the copy-in path and stops exporting user tensors — exactly what gfx1250 does at `custom_all_reduce.py:849`. And outside capture, `custom_all_reduce()` already passes `registered_input=False` and copies into the pool, so the eager path never exports user memory either.

That leaves one buffer: the `input` pool. If it were allocated with a raw HIP call instead of `torch.empty`, CustomAllreduce would keep working under `expandable_segments`, at the cost of one extra copy on the capture path. `IPCBufferPool` currently only exposes `uncached=True`, which maps to `hipDeviceMallocUncached` and isn't usable for a `max_size` data buffer, so this would mean adding a cached raw-alloc path — real work, not a one-liner.

Whether that's worth doing is your call, and here's why I'm asking: vLLM turns on `expandable_segments` in several of its OOM-hardening paths, so as written this PR makes those deployments fall back to RCCL permanently, with only a log line to show for it. **Is CustomAllreduce still ahead of RCCL at TP=8 once the capture path becomes copy-in?** If yes, the narrower fix looks worth the effort; if no, the blanket disable in this PR is the right call and should just be merged (after the test above is fixed).

For the record, the diagnosis in the PR description is correct — fixing only the `input` buffer would *not* be sufficient on its own, since `custom_all_reduce.cuh:3411` re-exports user graph buffers unless capture registration is also turned off.

TennyWang1223:
@zufayu 
After testing locally, switching directly to RCCL causes a significant performance regression, while the copy-in fix has almost no performance overhead. So I think it's better to change AITER CAR's input pool to use direct hipMalloc. I'll push a fix PR later.

jamesETsmith:
@zufayu and @TennyWang1223, thanks for looking at the PR! IIUC, @TennyWang1223 will implement an alternative change to input pool and you'll merge that over this PR, is that correct?

If this PR is still useful, I can update the tests according to @zufayu's suggestion, just lmk!

TennyWang1223:
@zufayu , @jamesETsmith 
PR #4621 fixes this error. Instead of falling back to RCCL, it keeps using custom_allreduce, at the cost of an additional d2d copy. Could you please help double-check it? If it's confirmed to resolve the custom allreduce graph capture crash when PyTorch sets expandable_segments:True, then #4560 can be closed.

jamesETsmith:
> @zufayu , @jamesETsmith PR #4621 fixes this error. Instead of falling back to RCCL, it keeps using custom_allreduce, at the cost of an additional d2d copy. Could you please help double-check it? If it's confirmed to resolve the custom allreduce graph capture crash when PyTorch sets expandable_segments:True, then #4560 can be closed.

Trying this now, thanks @TennyWang1223!

jamesETsmith:
@TennyWang1223 your PR looks good to me (I posted a comment on the PR[ here](https://github.com/ROCm/aiter/pull/4621#issuecomment-5223238514))! I'll leave this one open until yours is merged and then close it

jamesETsmith:
Closing this since #4621 has been merged. Thanks again for the review and follow up PR!

PR #2708, opened 2026-04-13
docs: add ISA-level kernel optimization guide

sunway513:
## Summary

- Add step-by-step guide for LLVM-based ISA kernel optimization workflow in AITER
- Covers: disassemble `.co` → extract reassemblable `.s` → modify → recompile → verify → profile
- Includes Python extraction script with branch label word-offset addressing
- Covers `llvm-objcopy --update-section` technique for preserving AMDHSA kernel metadata
- Includes rocprofv3 kernel-trace and ATT profiling instructions

## Details

This guide enables developers to inspect, modify, and recompile AITER GPU kernel ISA using standard LLVM tools (`llvm-objdump`, `llvm-objcopy`, `clang++`). Key topics:

1. **Disassembly** — `llvm-objdump` to inspect .co files
2. **ASM extraction** — Python script to produce reassemblable .s from objdump output, handling branch label word-offset addressing correctly
3. **Round-trip recompilation** — `clang++ -x assembler` with correct `--mcpu` flag
4. **Binary verification** — `llvm-objcopy --update-section .text` to swap code while preserving kernel metadata (.note section with AMDHSA descriptors)
5. **Profiling** — rocprofv3 `--kernel-trace` (SQLite output) and `--att` (Advanced Thread Trace)

## Test plan

- [x] Full round-trip verified: disassemble → extract → recompile → load → benchmark shows identical performance
- [x] All code snippets and commands tested on MI-series GPU
- [ ] Review for completeness and clarity

PR #4561, opened 2026-08-04
docs: fix ISA kernel optimization guide and example scripts

mario-ant:
## Motivation

Follow-up to #2708. I tried to use the ISA optimization guide on the shipped kernels and the example scripts don't run as committed, and the documented flow stops working as soon as an edit actually changes the code. This PR fixes the scripts against the real tool output and fills in what you need to modify, rebuild, test and profile one of the `hsa/**/*.co` kernels end to end.

What was broken:
- `llvm-objdump` prints AMDGPU code as `<TAB>instr … // ADDR: ENCODING` (ROCm 7.2.1's LLVM and upstream alike), but `extract_asm.py` / `analyze_kernel.py` / `roundtrip.sh` parse an `ADDR: BYTES instr` layout. On the guide's own example (`pa_bf16_pertokenFp8_gqa16_2tg_4w.co`) `extract_asm.py` raises `ValueError`, `analyze_kernel.py isa` reports 0 instructions, and `roundtrip.sh` stops at step 3.
- The "label value × 4 + base" rule isn't how the labels work: `label_XXXX` are ordinary local symbols left in `.symtab` (objdump already symbolizes branches with them), several names can alias one address (e.g. gfx950 `mla_a8w8_qh8_qseqlen1_gqaratio8_v3.co`), and the hex suffix doesn't reliably encode an offset.
- `llvm-objcopy --update-section .text=…` can't grow the section (`cannot fit data of size … that is part of a segment`), so the flow dead-ends once an edit adds an instruction, and it freezes the kernel descriptor, so register/LDS usage can't change.

## Technical Details

- `docs/examples/isa_optimization/extract_asm.py`: parses the real objdump format, takes branch labels (including aliases) from the ELF symbol table, and emits a *standalone* `.s`: `.text` + the kernel descriptor decoded from `.rodata` (`llvm-objdump -D -j .rodata`) as an `.amdhsa_kernel` block + the metadata note (`llvm-readelf --notes`) as `.amdgpu_metadata`. `clang -x assembler -target amdgcn-amd-amdhsa -mcpu=<gfx> -o new.co` then produces a complete, loadable code object whose descriptor is editable. Two descriptor fix-ups are applied and documented (the disassembler's granulated `.amdhsa_next_free_sgpr`; `.amdhsa_reserve_xnack_mask` vs. the generic target id). The arch is read from the ELF header; `--list`, `--text-only`, `--symbol`, `--llvm-bin`/`ROCM_PATH` added.
- `roundtrip.sh`: extract → assemble+link → compare `.text`, kernel descriptor, metadata and `e_flags` with the original; prints the first differing instruction from both disassemblies and exits non-zero on mismatch. `--mcpu` optional (auto-detected), `--keep DIR`.
- `analyze_kernel.py`: working instruction scan with non-overlapping categories plus a descriptor/metadata resource summary; profile mode prefers the suffix-free rocpd views and passes `--filter` as a query parameter.
- `docs/isa_kernel_optimization.md` rewritten around that flow: what a `.co` contains; how AITER selects and loads it (per-family CSVs, `csrc/py_itfs_cu/asm_*.cu`, `AITER_ASM_DIR` being reset on `import aiter`, `LoadKernel` log line, `KernelArgs` ↔ metadata argument offsets); data hazards / `s_waitcnt` being the editor's responsibility (the `s_nop`s in these kernels are wait states, not filler); changing VGPR/SGPR/LDS usage; testing with `op_tests`; rocprofv3 `--stats` / kernel filters / PMC / ATT (defaults, decoder install, outputs, viewer); the in-place `--update-section` flow kept as a fallback with its limits; gfx950 in the arch table; troubleshooting entries for the actual assembler/objcopy errors; CDNA3/CDNA4 ISA references. Also fixes the undefined `$INSTALL_PATH`, the grep patterns and the metadata inspection command from the review comments on #2708.
- `README.md` / `Dockerfile` updated to match: `pip install aiter` dropped (AITER installs from source and needs PyTorch), ATT decoder built from a pinnable rocm-systems ref via sparse checkout. The example scripts are invoked via `bash`/`python3` in the docs and are checked in non-executable.

## Test Plan

- `bash docs/examples/isa_optimization/roundtrip.sh <kernel.co>` over every gfx942 `pa`/`mla`/`bf16gemm`/`topksoftmax` and every gfx950 `pa`/`mla` object in the tree (233 kernels), with a ROCm-layout LLVM (`$ROCM_PATH/llvm/bin`) and, for extract + assemble, upstream LLVM 18.
- `extract_asm.py --text-only` + `llvm-objcopy --update-section` fallback flow on the guide's example kernel; growing/shrinking `.text` to confirm the documented objcopy limitation.
- `analyze_kernel.py isa` on the example kernel; `analyze_kernel.py profile` against a database with the rocpd `kernel_dispatch` / `kernel_symbol` schema.
- `black` / `ruff` (repo config) on the Python files, `shellcheck` on `roundtrip.sh`.
- Not covered here (no GPU on the machine I did this on): `hipModuleLoad` of a rebuilt object, `op_tests`, and the rocprofv3/ATT commands (flags checked against the rocprofiler-sdk docs). Would appreciate someone with an MI300/MI350 running `roundtrip.sh` + `op_tests/test_pa.py` on a rebuilt kernel.

## Test Result

- All 233 gfx942/gfx950 kernels above rebuild with byte-identical `.text`, kernel descriptor, metadata and `e_flags`.
- One known non-bit-exact case, documented under "Known re-encoding differences": gfx950 `fmoe` kernels using `v_mfma_scale_f32_16x16x128_f8f6f4` (bits 13–14 of the leading `v_mfma_ld_scale_b32` word are set in the shipped encoding and cleared by LLVM's assembler; everything else identical). `roundtrip.sh` reports the differing instruction and exits non-zero as intended.
- Fallback flow reproduces the original `.text`/descriptor; growing `.text` fails with the documented `cannot fit data … part of a segment` error, shrinking works.
- Lint clean.

## Submission Checklist


sunway513:
LGTM. @valarLip plz help review and merge.. thanks.

valarLip:
can we get ruff test fix

mario-ant:
> can we get ruff test fix

Done — the two scripts are now clean under the pinned ruff 0.16.0 defaults (EXE001/FURB167/PLW1510/RUF100/FURB192; I had linted with an older ruff locally). Black was already passing.

PR #4573, opened 2026-08-05
feat(flydsl): support strided GDN decode state routing

junna2016:
Accept production strided q/k/v inputs without contiguous packing and route recurrent state through independent read and write block indices while preserving the legacy indices API. Add ROCm coverage for tuned head shapes and BF16/FP32 state.

## Motivation

RTP-LLM's production GDN decode inputs may be strided views produced by
operations such as `torch.split`. The original AITER FlyDSL wrapper converted
`q`, `k`, and `v` to contiguous tensors before launching the kernel:

```python
query.contiguous()
key.contiguous()
value.contiguous()
```

Although this ensured correctness, it introduced additional packing kernels,
temporary allocations, and memory traffic on the latency-sensitive decode
path.

The original interface also accepted only one `indices` tensor, requiring the
recurrent state to be read from and written to the same cache block. When a
request crossed a cache-block boundary, RTP-LLM had to launch a separate kernel
to copy the state before executing GDN decode.

This change allows the FlyDSL kernel to consume production strided inputs
directly and perform cross-block state routing inside the GDN decode kernel.

### Strided Q/K/V inputs

The wrapper now passes the tensors' actual strides to the FlyDSL kernel
factory:

```python
query.stride()
key.stride()
value.stride()
```

The kernel constructs `GTensor` descriptors using these explicit strides:

```python
q_tensor = GTensor(..., stride=q_strides)
k_tensor = GTensor(..., stride=k_strides)
v_tensor = GTensor(..., stride=v_strides)
```

The kernel can therefore address non-contiguous `q`, `k`, and `v` views
directly, eliminating the previous `.contiguous()` packing operations.

The supported layout contract is:

- `q`, `k`, `v`, `a`, and `b` do not need to be fully contiguous.
- The innermost dimensions of `q` and `k` must be contiguous
  (`stride(-1) == 1`) because the kernel uses vectorized loads along the head
  dimension.
- Outer dimensions may use arbitrary valid strides.
- `v`, `a`, and `b` are accessed using their explicitly supplied strides.

The strides are compile-time kernel specialization parameters, so the generated
kernel does not require runtime stride branches.

### Independent state read and write indices

The original interface used one `indices` tensor for both loading and storing
the recurrent state:

```text
state[indices] -> GDN decode -> state[indices]
```

When a request crossed a cache-block boundary, RTP-LLM had to perform:

```text
state[read_indices] -> copy kernel -> state[write_indices]
state[write_indices] -> GDN decode -> state[write_indices]
```

The interface now accepts independent `read_indices` and `write_indices`:

```text
state[read_indices] -> GDN decode -> state[write_indices]
```

For each request:

- `read_indices` identifies the cache block containing the previous recurrent
  state.
- `write_indices` identifies the cache block where the updated state is stored.
- Equal indices retain the original in-place update behavior.
- Different indices perform cross-block state migration as part of the GDN
  decode kernel.

For example:

```python
read_indices = [5, 8, 11]
write_indices = [6, 8, 12]
```

This represents:

```text
request 0: state[5]  -> state[6]
request 1: state[8]  -> state[8]
request 2: state[11] -> state[12]
```

This design removes the standalone state-copy kernel and its associated launch
and memory-traffic overhead.

### Backward compatibility

The existing required `indices` argument and positional argument order are
preserved. When the new arguments are omitted, they fall back to `indices`:

```python
read_indices = indices if read_indices is None else read_indices
write_indices = indices if write_indices is None else write_indices
```

Existing callers therefore continue to use the original in-place state update
semantics without modification.

## Expected Benefits

- Eliminates contiguous packing for production strided `q`, `k`, and `v`.
- Avoids temporary tensors and redundant memory traffic.
- Eliminates the separate cross-block state-copy kernel.
- Reduces decode-path kernel launches.
- Supports independent state routing for each request in a mixed batch.
- Preserves compatibility with existing AITER callers.

## Test Plan

The FlyDSL linear-attention tests cover:

- Existing contiguous-input behavior.
- Production-like strided views created with `torch.split`.
- Batch sizes greater than one.
- Independent `read_indices` and `write_indices`.
- In-place and cross-block state updates.
- BF16 and FP32 recurrent-state dtypes.
- Seven tuned GDN head-shape configurations on gfx942.
- Numerical equivalence with the legacy explicit state-copy workflow.
- Backward compatibility when the new arguments are omitted.


aiter/ops/flydsl/kernels/gdr_decode.py
```diff
@@ -128,17 +132,28 @@ def gdr_decode_kernel(
...
         global_v_start = tile_v_start + wid * WARP_TILE_V + w_tid // WARP_THREADS_K
 
-        indices_tensor = GTensor(indices, dtype=T.i32, shape=(-1,))
-        pool_idx = fx.Int32(indices_tensor[b_i])
+        read_indices_tensor = GTensor(read_indices, dtype=T.i32, shape=(-1,))
+        write_indices_tensor = GTensor(write_indices, dtype=T.i32, shape=(-1,))
+        read_pool_idx = fx.Int32(read_indices_tensor[b_i])
+        write_pool_idx = fx.Int32(write_indices_tensor[b_i])
 
         q_tensor = GTensor(
```
xytpai, line 140: q_tensor.vec_load and k_tensor.vec_load always load consecutive elements. If query.stride(-1) != 1 or key.stride(-1) != 1, the vector lanes will read incorrect addresses because the supplied strides only affect the vector’s base offset. Please add an explicit check for both conditions and raise ValueError (or fall back to .contiguous()).

xytpai (dismissed):

aiter/ops/flydsl/kernels/gdr_decode.py
junna2016: modify done

xytpai (approved):

valarLip (approved):

PR #4574, opened 2026-08-05
[WIP][Kimi-K3] retune shared BF16 GEMMs without split-K

XiaobingSuper:
## Motivation

Kimi-K3 overlaps the shared-expert MLP with the routed MXFP4 MoE path during decode. The existing small-M shared gate-up winners use high-LDS FlyDSL split-K kernels. Under FULL HIP graph replay, the captured branches are independent, but these kernels cannot co-reside with the routed kernels:

- routed `moe_gemm1`: ~55 KB LDS
- existing FlyDSL splitK7 gate-up: ~143 KB LDS
- combined demand exceeds the per-CU LDS budget

This produces a valid dual-stream graph with effectively serial execution.

## Changes

Retune the TP8 Kimi-K3 shared-expert BF16 shapes for decode M values `1, 2, 4, 8, 16, 24, 32, 48, 64`:

- gate-up: `N=1536, K=7168`
- down-proj: `N=7168, K=768`
- exclude all split-K kernel families
- exclude torch and hipBLASLt backends

This draft intentionally changes only `kimik3_bf16_tuned_gemm.csv`.

## Validation

- Four-GPU parallel tuning on MI355/gfx950
- 18/18 selected rows are Triton, `splitK=0`
- CSV key uniqueness validated
- Production operator accuracy checks completed
- TP4 graph replay comparison:
  - M16: 118.4 us -> 123.2 us (+4.1%)
  - M24: 123.1 us -> 126.7 us (+2.9%)

## Open questions

This removes the overlap-hostile split-K kernels but currently regresses TP4 graph replay latency. Gate-up ASM alternatives also showed approximately 1.9%-2.0% element mismatch under the default allclose check. Keeping this PR as draft while we evaluate whether an overlap-aware non-split-K winner can recover the latency.


XiaobingSuper:
TP4 model-level validation was attempted using Kimi-K3 dummy load with num_shared_experts=1, which preserves the TP8 local shared GEMM shapes (N=1536/K=768). Model loading reached ~280.2 GiB allocated per GPU and then failed on an additional 2.3 GiB allocation. Full model validation is therefore blocked by TP4 capacity; no TP8 run was attempted because those GPUs are shared. Current validation remains the TP4 MXFP4 routed-MoE + shared-MLP graph reproducer documented in the PR.

XiaobingSuper:
Correction: production Kimi-K3 validation on MI355 requires TP8. The TP4 run was only a local-shape proxy and is not a valid full-model configuration; its OOM is expected for this model. Full model validation remains pending until a coordinated TP8 resource window is available.

XiaobingSuper:
TP8 full-model validation completed on MI355 with Kimi-K3, vLLM FULL_DECODE_ONLY, BF16 shared experts, 1K input / 5 output / CON=4. Result is negative: the retuned Triton splitK=0 config still showed 0% decode side-stream overlap. Median decode graph span increased from the previous BF16 baseline 25.57 ms to 26.63 ms (+4.1%), and median TPOT increased from 28.33 ms to 29.35 ms (+3.6%). The model service completed successfully and has been stopped. Keeping this PR as draft; the config should not be merged in its current form.

PR #4577, opened 2026-08-05
[FlyDSL] [KIMI-K3] Enable KDA per-channel decay gate in FlyDSL GDR decode

waqahmed-amd-fi:
## Summary

This PR extends the existing FlyDSL GDR decode kernel with a compile-time per-channel decay path for Kimi Delta Attention (KDA), covering [SILOTIGER-847](https://amd.atlassian.net/browse/SILOTIGER-847).

- Decay `a` is per-channel `(B, Sq, H, K)` instead of per-head `(B, Sq, H)`.
- `dt_bias` is per-channel `(H, K)` instead of per-head `(H)`, and may be f32 while activations stay bf16/fp16.
- Gate math is `g = g_min * sigmoid(exp(A_log) * (a + dt_bias))` with `g_min = -5`.
- `gate_mode` is inferred from `a.dim()` (`"kda"` if 4-D, else `"gdr"`), so existing callers keep the same signature.
- Scalar GDR is untouched; the KDA tilings are new and reach only the KDA path.
- `flydsl_gdr_decode` is exported from `aiter.ops.flydsl`.

## Technical details

**Kernel** (`kernels/gdr_decode.py`): `gate_mode` is a factory argument, so `"gdr"` and `"kda"` compile to separate `lru_cache` binaries and `PER_CHANNEL` folds out under `const_expr`. Only a non-`gdr` mode is suffixed onto the kernel name, so existing GDR symbols are unchanged. The scalar arm's one edit is the `dt_bias` copy atom, which is the same 16-bit copy whenever `dt_bias` carries the activation dtype.

**Wrapper** (`linear_attention_kernels.py`): in KDA mode `a` is `(B, Sq, H_v, D_k)` and `dt_bias` is `(H_v, D_k)`, and both are checked against the shapes that `query` and `value` imply. The kernel bakes in a few layout assumptions, so the wrapper now rejects the inputs that would break them: `a` must be dense along `D_k`, `out` must be contiguous, and an unshuffled `state` must have unit stride in K. A caller-supplied `stream` must be on `query`'s device, and the staging copies are enqueued on that same stream so they complete before the launch reads them. A negative slot index marks a CUDA-graph padding row: the kernel zeroes that row of `out` and leaves the state pool untouched.

**Tuning**: `_KDA_DECODE_BY_PART` stores one `(NUM_BLOCKS_PER_V_DIM, NUM_WARPS, WARP_THREADS_K)` triple per batch size, measured at B ∈ {1, 4, 64, 256} for a single shape: 12 key and value heads, 128 wide, bf16 activations over an f32 state. Every other shape keeps main's `_decode_tiling()` heuristic. `NUM_BLOCKS_PER_V_DIM` splits the value dimension so that one decode launches more blocks, which is how a small batch fills the GPU, and the split that pays therefore depends on how many CUs the part has. An arch name does not pin that down — MI308X and MI300X are both gfx942, at 80 CUs and 304 — so the table is keyed on `(arch, CU count)`, and a part with no entry of its own uses `_decode_tiling()`, which reads the CU count at runtime.

**Tests and benchmark**: `op_tests/test_flydsl_gdr_decode.py` covers both gates and the per-part lookup; the torch oracle is test-only in `op_tests/kda_ref.py`. In `bench_gdr_decode_kda.py`, `--sweep` prints a table entry, and `--bench` needs a vLLM checkout passed with `--vllm` for its comparator, `fused_recurrent_kda_packed_decode` from `vllm/models/kimi_k3/amd/ops/third_party/kda/fused_recurrent.py`. That is the Triton packed decode, not the HIP fused KDA, and only device time is quoted because vLLM graphs this kernel with others.

## Test plan

- [x] `python3 op_tests/test_flydsl_gdr_decode.py` and `pytest aiter/ops/flydsl/test_flydsl_linear_attention.py`
- [x] Numerics vs the torch KDA oracle: elementwise `rtol`/`atol` 1e-3 in the tests, and the reference's own RMSE-ratio-or-absolute 1e-3 pair in the benchmark's parity check
- [x] bf16 activations, f32 `dt_bias`, f32 and bf16 state, padded non-contiguous state, strided state indices, both shuffle modes, slots from 1, negative-slot zeros
- [x] Scalar GDR unchanged: same `_decode_tiling()` dispatch as main, no restored CSV, still matching the Triton reference across the existing GQA shapes
- [x] Device-time A/B vs Triton packed decode at B ∈ {1, 4, 64, 256}, 12 heads, K=V=128, bf16, f32 state, on MI355X and MI308X

## Test result

Device time per call, host excluded, L2 flushed, 5 trials on fresh inputs, mean ± std; both kernels see the same data. The last column runs FlyDSL on the config `_decode_tiling()` picks, so it is what the tuned row buys.

MI355X, gfx950, 256 CUs:

| B   | Triton µs    | FlyDSL µs    | speedup      | FlyDSL untuned µs |
| --- | ------------ | ------------ | ------------ | ----------------- |
| 1   | 7.41 ± 0.05  | 6.63 ± 0.09  | 1.12x ± 0.02 | 6.63 ± 0.05       |
| 4   | 7.86 ± 0.01  | 6.80 ± 0.01  | 1.16x ± 0.00 | 7.33 ± 0.01       |
| 64  | 22.51 ± 0.04 | 20.84 ± 0.07 | 1.08x ± 0.01 | 22.76 ± 0.11      |
| 256 | 79.61 ± 0.18 | 76.91 ± 0.03 | 1.04x ± 0.00 | 88.71 ± 0.05      |

MI308X, gfx942, 80 CUs:

| B   | Triton µs      | FlyDSL µs     | speedup      | FlyDSL untuned µs |
| --- | -------------- | ------------- | ------------ | ----------------- |
| 1   | 13.38 ± 0.03   | 11.62 ± 0.04  | 1.15x ± 0.00 | 12.55 ± 0.06      |
| 4   | 18.56 ± 0.01   | 16.00 ± 0.04  | 1.16x ± 0.00 | 18.15 ± 0.09      |
| 64  | 77.22 ± 0.08   | 54.76 ± 0.19  | 1.41x ± 0.01 | 59.93 ± 0.21      |
| 256 | 248.33 ± 12.44 | 174.84 ± 0.09 | 1.42x ± 0.07 | 196.44 ± 0.17     |

## Submission checklist

- [x] Look over the contributing guidelines at [https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests](https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests).

[SILOTIGER-847]: https://amd.atlassian.net/browse/SILOTIGER-847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samremes (commented):
Code-level review only; no tests, builds, imports, or benchmarks were run. The per-channel recurrence math looks consistent with the intended formula. Residual gaps: confirm first-call JIT-only deployment is intentional, and extend automated coverage to the larger supported head counts and target batches.

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -104,12 +104,39 @@ def flydsl_gdr_decode(
...
+        assert (
+            a.shape[-1] == query.shape[-1]
+        ), f"`a` last dim must be D_k={query.shape[-1]}, got {a.shape[-1]}"
+        assert dt_bias.shape == (a.shape[2], a.shape[3]), (
+            f"dt_bias {tuple(dt_bias.shape)} does not match `a` heads/channels "
+            f"{(a.shape[2], a.shape[3])}"
+        )
+        # `a` reaches the kernel with its own strides and is read as vectors
+        # along D_k, so that axis has to be dense -- a strided one would be
+        # misread, not rejected. dt_bias needs no such check: it is copied
```
samremes: **High:** Allowing non-contiguous `dt_bias` introduces an explicit-stream race. The later `.contiguous()` is enqueued on PyTorch's current stream, while `_run_compiled` launches on the caller-provided `stream`; the kernel can start reading before the copy completes. The shuffled-state copies have the same issue before and after launch. Please put preprocessing, launch, and postprocessing in the supplied stream context, or add explicit event dependencies.

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -104,12 +104,39 @@ def flydsl_gdr_decode(
...
 
+    # `a`'s rank selects the gate; dt_bias must match it, or the kernel would
+    # index a layout the caller did not pass.
+    gate_mode = "kda" if a.dim() == 4 else "gdr"
+    if gate_mode == "kda":
+        assert (
+            dt_bias.dim() == 2
+        ), f"per-channel `a` needs a 2D (H_v, D_k) dt_bias, got {tuple(dt_bias.shape)}"
+        assert (
+            a.shape[-1] == query.shape[-1]
```
samremes: **High:** This only checks the last dimension, not the complete `a == (B, Sq, H_v, D_k)` shape. A consumer-native `(1, B, H, K)` view passes these checks but the kernel indexes it as `(B, Sq, H, K)`, causing wrong-row or out-of-bounds reads for `B > 1`. Please validate the full shapes of `q/k/v/a/b`, `indices`, `state`, and `out` before creating the kernel.

aiter/ops/flydsl/kernels/gdr_decode.py
```diff
@@ -105,6 +116,7 @@ def gdr_decode_kernel(
         dtype_ = get_dtype_in_kernel(dtype)
         fx_dtype_ = fx.BFloat16 if dtype == "bf16" else fx.Float16
```
samremes, line 117: **High:** Activation and state storage types are still conflated. `vec_t` is derived from `dtype_`, so fp16 activations with bf16 state truncate the final state vector to fp16 before the bf16 store. The wrapper also accepts fp32 activations, but this fallback selects fp16 for output conversion. Please restrict activation dtypes to the intended fp16/bf16 set and derive a separate state-store vector type from `state_dtype_`; add mixed fp16-activation/bf16-state coverage.

aiter/ops/flydsl/gdr_decode_tuned.csv
```diff
@@ -5375,3 +5375,7 @@ gfx950,torch.bfloat16,torch.float32,253,1,16,64,128,128,8,4,16,358.3230000000003
 gfx950,torch.bfloat16,torch.float32,254,1,16,64,128,128,8,4,16,360.0035000000007
 gfx950,torch.bfloat16,torch.float32,255,1,16,64,128,128,8,1,4,361.3235000000004
 gfx950,torch.bfloat16,torch.float32,256,1,16,64,128,128,8,4,16,362.8634999999995
+gfx950,torch.bfloat16,torch.float32,1,1,12,12,128,128,8,2,16,6.80
```
samremes, line 5378: **Medium:** `get_default_kwargs` does not include `gate_mode` in its key, so these per-channel-tuned 12:12 rows are also selected by scalar GDR and change its previous fallback geometry. Please add gate mode to the CSV schema/lookup key, treating existing rows as scalar mode. Also, all four new rows are gfx950-only; gfx942 still takes the untuned fallback despite being an explicit target.

aiter/ops/torch_ref/kda.py
```diff
@@ -0,0 +1,112 @@
...
+        dt_bias: ``(H, K)`` per-channel, or ``(H,)`` scalar per head.
+        g_min: KDA's lower bound, -5. None selects the GDR softplus gate.
+
+    Returns:
+        Decay in log space, shaped like ``a`` -- callers apply ``exp``.
+    """
+    x = a.float() + dt_bias.float()
+    A = A_log.float().exp()
+    # a is (..., H) or (..., H, K); line A up with the head axis either way.
+    A = A if a.shape[-1] == A.shape[0] else A[:, None]
```
samremes, line 50: **Medium:** Inferring scalar versus per-channel mode from `a.shape[-1] == len(A_log)` is ambiguous. For a legal channel shape with `H == K`, this chooses scalar broadcasting and applies `A_log` over K instead of H, corrupting the oracle. Please branch on `a.dim()` and validate the expected shapes explicitly.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,360 @@
...
+    """
+    args, pool, indices = _kda_inputs(
+        B, H, dt, first_index, padded, shuffle, indices_stride=indices_stride
+    )
+
+    initial_state = pool[indices.long()].clone()
+    if not shuffle:
+        initial_state = initial_state.transpose(-1, -2)
+
+    kernel_pool = pool.clone()
```
samremes, line 237: **Medium:** For the padded, non-dense view built above, `pool.clone()` materializes contiguous storage, so the kernel never sees the outer padded stride this test claims to exercise. Please clone the padded backing allocation and recreate the view, or use `empty_strided(pool.shape, pool.stride())` followed by a copy; assert the resulting kernel pool keeps the original stride.

aiter/ops/flydsl/kernels/gdr_decode.py
samremes: (Probably this mixed case won't ever happen though?)

aiter/ops/flydsl/linear_attention_kernels.py
waqahmed-amd-fi: Fixed. key, value, out, a, b, dt_bias, A_log, read_indices, write_indices and state are now validated against the full (B, Sq, H, D) layout before the kernel is created.

aiter/ops/flydsl/linear_attention_kernels.py
waqahmed-amd-fi: Fixed, and it reproduced: with the launch on a side stream and the current stream held busy, the output came back at RMSE ratio 0.98 — the kernel was reading unwritten staging memory.

aiter/ops/flydsl/gdr_decode_tuned.csv
waqahmed-amd-fi: Fixed. Added a gate_mode column to the CSV and to the lookup key; the 5376 existing rows are tagged gdr, the four new ones kda, and --sweep emits the tag. Scalar GDR at 12:12 is back on the (1,4,8) fallback (new test pins that). The column is appended last so an old-format row still parses as scalar, which is why the diff touches every line.

gfx942 is still outstanding: no gfx942 box available. The rows are additive now that the key distinguishes gates.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+)
+
+# Skips at import when flydsl or a GPU is missing, taking this module with it.
+from aiter.ops.flydsl.test_flydsl_linear_attention import (
+    Args,
+    check_gdr_decode,
+)
+
+
+def test_flydsl_gdr_decode_is_importable_from_package_namespace():
```
samremes: I think this test is really quite unnecessary, all test should be only about testing it, not checking if it importable in some specific way.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    assert hasattr(flydsl_ops, "flydsl_gdr_decode"), (
+        "flydsl_gdr_decode is missing from aiter.ops.flydsl -- check that its "
+        "import is uncommented in aiter/ops/flydsl/__init__.py"
+    )
+    assert callable(flydsl_ops.flydsl_gdr_decode)
+
+
+def test_flydsl_gdr_decode_is_advertised_in_all():
```
samremes: Same here, not really relevant for testing the kernel itself.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    assert "flydsl_gdr_decode" in flydsl_ops.__all__
+
+
+def test_package_export_is_the_kernel_wrapper_itself():
```
samremes: Same.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+import torch
+
+import aiter.ops.flydsl as flydsl_ops
+from aiter.ops.flydsl import is_flydsl_available
+
+pytestmark = pytest.mark.skipif(
+    not is_flydsl_available(), reason="flydsl is not installed"
+)
+
+# Skips at import when flydsl or a GPU is missing, taking this module with it.
```
samremes: What does this comment mean?

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""CI-visible tests for the FlyDSL GDR decode op: its export, and its numerics.
+
+CI collects only ``op_tests/test_*.py`` at depth 1 (see
+``.github/scripts/split_tests.sh``), so the in-package suite at
+``aiter/ops/flydsl/test_flydsl_linear_attention.py`` never runs automatically.
```
samremes: Should we still have the test under flydsl? And we could have the CI pull in the tests from there as likely intended. Anyway, this kind of commentary is not relevant to have in this test file.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
```
samremes: This is a new file, not from 2024.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+
+"""CI-visible tests for the FlyDSL GDR decode op: its export, and its numerics.
```
samremes: Export should not probably be tested, just the kernel.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+
+CI collects only ``op_tests/test_*.py`` at depth 1 (see
+``.github/scripts/split_tests.sh``), so the in-package suite at
+``aiter/ops/flydsl/test_flydsl_linear_attention.py`` never runs automatically.
+
+The export checks matter on their own: importing from
+``linear_attention_kernels`` keeps working if the export drops back out of
+``__init__.py``, so nothing else would catch that.
```
samremes: Not really necessary?

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+
+These cases trust the torch reference and put the *kernel* on trial;
+``op_tests/test_kda_torch_ref.py`` runs the arrow backwards. Which one fails
+localises the fault: here alone, the per-channel kernel; there alone, the
+reference; both, the gate formula they share.
+
+Keep it thin -- shapes belong here, the reference kernel belongs in-package.
```
samremes: Why do we have this commentary here, the torch reference is the reference and we use it in the tests.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+def test_package_export_is_the_kernel_wrapper_itself():
+    from aiter.ops.flydsl.linear_attention_kernels import flydsl_gdr_decode
+
+    assert flydsl_ops.flydsl_gdr_decode is flydsl_gdr_decode
+
+
+@pytest.mark.parametrize(
+    "args",
+    [
+        # Smallest GQA shape, the pre-847 scalar path.
```
samremes: What does pre-847 mean? Everything needs to read as stand-alone, not referencing a ticket id or such.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+
+G_MIN = -5.0
+K = V = 128
+
+
+def assert_close_rmse(name, ref, tri, ratio=1e-3, err_atol=1e-3):
+    """vLLM's own bar (``test_kda.py:92``): RMSE-relative with an absolute
+    escape hatch, at the packed-vs-dense decode pair's (1e-3, 1e-3). The ratio
+    alone is not the bar.
+    """
```
samremes: Could we use some of the existing helpers (checkAllClose?) instead of rolling a new one here. Also we should probably avoid just referencing vLLM here, and just state that we use this rtol/atol as the test criteria.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    args = {
+        "q": torch.randn(B, T, H, K, dtype=dt, device=dev),
+        "k": torch.randn(B, T, H, K, dtype=dt, device=dev),
+        "v": torch.randn(B, T, H, V, dtype=dt, device=dev),
+        "a": torch.randn(B, T, H, K, dtype=dt, device=dev),
+        "b": torch.randn(B, T, H, dtype=dt, device=dev),
+        "dt_bias": (
+            torch.randn(H, K, dtype=torch.float32, device=dev) * 0.1
+        ).contiguous(),
+        "A_log": (torch.randn(H, dtype=torch.float32, device=dev) * 0.5).contiguous(),
```
samremes: why `.contiguous()`? Isn't it already by default contiguous.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    n_slots = B + first_index
+    if padded:
+        # vLLM's setup: padded rows, so the kernel must honour the strides.
+        storage = torch.randn(n_slots, H * K * V + 17, dtype=torch.float32, device=dev)
+        pool = storage[:, : H * K * V].view(n_slots, H, d0, d1)
+        # vLLM's own assertions (test_kda.py:321-322). The second bites: only
+        # the outer stride is padded, each slot still internally contiguous.
+        # Asserted, so a storage-shape change cannot silently test another layout.
+        assert not pool.is_contiguous()
+        assert pool.stride()[1:] == (d0 * d1, d1, 1)
```
samremes: Don't just say vLLM: explain what the setup is and why we test it like that.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    else:
+        pool = torch.randn(n_slots, H, d0, d1, dtype=torch.float32, device=dev)
+
+    if indices_stride > 1:
+        # vLLM parametrizes this (state_indices_stride): the serving stack hands
+        # over a strided column of a wider table, not a fresh contiguous array.
+        storage = torch.zeros(B, indices_stride, dtype=torch.int32, device=dev)
+        indices = storage[:, 0]
+        assert indices.stride(0) == indices_stride
```
samremes: Isn't this kind of a trivial assert here, as the tensor was just created such that it will hold.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+        # storage, slots numbered from 1, strided index column.
+        (4, 12, torch.bfloat16, False, True, 1, 8),
+        (4, 12, torch.bfloat16, False, False, 1, 1),
+        (4, 12, torch.bfloat16, True, True, 1, 1),
+        (4, 12, torch.bfloat16, True, False, 1, 8),
+        # The batch sizes with 12x12 rows in gdr_decode_tuned.csv; before
+        # these, only B=4's geometry was reachable from a test.
+        (1, 12, torch.bfloat16, True, False, 1, 1),
+        (64, 12, torch.bfloat16, True, False, 1, 1),
+        (256, 12, torch.bfloat16, True, False, 1, 1),
```
samremes: Why are we referencing some tuning csv file in a test? Just state what configuration these are testing, no need to say where they are from. Just say why we are testing these.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+        # Past K3's head count: no tuned row, so the fallback geometry.
+        (2, 64, torch.bfloat16, True, False, 1, 1),
```
samremes, line 186: This doesn't have to be specified as "Past K3"; the first test case is also H=8 which doesn't match the other ones with H=12.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+        "b256_h12_tuned",
+        "b2_h64",
+    ],
+)
+def test_kda_per_channel_gate_matches_torch_reference(
+    B, H, dt, shuffle, padded, first_index, indices_stride
+):
+    """The KDA gate: 4D `a`, 2D f32 dt_bias, g_min * sigmoid(exp(A_log) * x).
+
+    H is 1:1 with the k heads -- K3's ratio, which the scalar path never saw.
```
samremes: We don't need the extra commentary here. Just state what is being tested.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    got_state = kernel_pool[indices.long()]
+    if not shuffle:
+        got_state = got_state.transpose(-1, -2)
+
+    out_err = assert_close_rmse("o", ref_out, args["out"])
+    state_err = assert_close_rmse("ht", ref_state, got_state)
+
+    # Tripwire under the bar above: agreement is ~1e-4, so drift to just inside
+    # 1e-3 would pass while signalling that something changed.
+    assert out_err < 5e-4 and state_err < 5e-4, (out_err, state_err)
```
samremes: Why do we add extra checks here if `assert_close_rmse` already has the asserts?

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+def test_channel_strided_a_is_rejected():
+    """`a` is vector-loaded along D_k, so a strided channel axis reads garbage.
+
+    Rejected rather than silently wrong: unlike q/k/v the wrapper does not copy
+    `a`, it forwards the strides. The other axes stay free -- K3 passes a view of
+    a (1, B, H, K) buffer, which is padded in the batch stride.
```
samremes: Make the description more stand-alone: explain what we are exactly testing, no need to reference K3 or anything here.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+        stream=side,
+    )
+    torch.cuda.synchronize()
+
+    ref_out, ref_state = _kda_reference(args, initial_state)
+    assert_close_rmse("o", ref_out, args["out"])
+    assert_close_rmse("ht", ref_state, kernel_pool[indices.long()])
+
+
+def test_non_contiguous_dt_bias_is_copied_not_rejected():
```
samremes: I don't this test really needs to be here, it is not really testing anything about the kernel. And why do we expect a copy in the kernel wrapper?

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+        use_qk_l2norm=True,
+        need_shuffle_state=True,
+    )
+
+    ref_out, ref_state = _kda_reference(args, initial_state)
+    assert_close_rmse("o", ref_out, args["out"])
+    assert_close_rmse("ht", ref_state, kernel_pool[indices.long()])
+
+
+def test_negative_slot_is_skipped_and_zero_is_not():
```
samremes, line 369: Rather than having an ambiguous reference, I think we should adopt a single contract and have the reference also follow that. And this test is then unnecessary.

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,475 @@
...
+    )
+
+    # Row 0 asked for slot -1: untouched, not zeroed.
+    assert (args["out"][0] == 7.0).all()
+    # Slot 0 is valid here even though the reference would call it invalid.
+    assert not (args["out"][1] == 7.0).all()
+    assert not torch.equal(kernel_pool[0], pool[0])
+
+
+def test_kda_tuned_rows_are_not_reachable_from_the_scalar_gate():
```
samremes: Is this test relying on the CSV having a specific set of tuned rows? The test definitely should be independent of any CSV file and only test the dispatch/lookup logic. So remove this one and maybe replace with one where the tuning data is constructed explicitly in the test itself, and testing that the lookup picks the right one.

op_tests/test_kda_torch_ref.py
```diff
@@ -0,0 +1,121 @@
+# SPDX-License-Identifier: MIT
```
samremes: Why are we testing the reference? I'd just remove this file.

aiter/ops/flydsl/test_flydsl_linear_attention.py
```diff
@@ -47,6 +47,8 @@ class Args:
     head_k_dim: int
     head_v_dim: int
     use_qk_l2norm: bool = True
+    # KDA pairs an f32 bias with a lower-precision query; None is pre-847 behaviour.
```
samremes: This is an existing file, but should the test file be moved under op_tests? Or does it make sense to have it here?

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -99,10 +103,16 @@ def flydsl_gdr_decode(
...
     device = query.device
     dtype = query.dtype
+    # Set only for a foreign launch stream, the one case needing reordering.
+    producer_stream = None
+    if stream is None:
+        stream = torch.cuda.current_stream(device)
+    else:
+        current_stream = torch.cuda.current_stream(device)
+        if stream != current_stream:
+            producer_stream = current_stream
```
samremes: This logic does not make sense to me. If the stream is given, or we use the current_stream, we just need to put all operations on that stream. There is no need to have a separate producer stream?

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -136,42 +148,102 @@ def flydsl_gdr_decode(
...
+        ), f"`a` must be dense along D_k, got stride {a.stride(-1)}"
+    # The transpose, `.contiguous()` staging and write-back are GPU copies that
+    # would race a launch on a foreign `stream`; run them there, after one wait.
+    # The caller orders the results. Entering the context costs ~5us against a
+    # ~7us kernel, so it is skipped when the stream is already the caller's.
+    with torch.cuda.device(device.index), (
+        nullcontext() if producer_stream is None else torch.cuda.stream(stream)
+    ):
+        if producer_stream is not None:
+            stream.wait_stream(producer_stream)
```
samremes: why are we waiting for anything? we should just put everything on the user-given `stream` (which defaults to current stream of the device) and it will run them in the correct order.

samremes (dismissed):
I think the tests need some streamlining.

aiter/ops/flydsl/linear_attention_kernels.py
waqahmed-amd-fi: Agreed, removed. The launch stream now simply defaults to the device's current one, and all the work runs on it.

aiter/ops/flydsl/linear_attention_kernels.py
waqahmed-amd-fi: Dropped it. Also fixed the regression test.

aiter/ops/flydsl/test_flydsl_linear_attention.py
waqahmed-amd-fi: Claude suggests to keep it where it is, for two reasons.

First, it doubles as a helper module. op_tests/test_flydsl_gdr_decode.py imports Args and check_gdr_decode from it, and only the in-package path works under both ways tests get run. CI executes files as python3 <file>, which puts just op_tests/ on the path; pytest instead imports them as op_tests.* from the repo root. So after a move, neither from op_tests.test_… import … nor the bare from test_… import … would work in both modes.

Second, moving it would start running 22 tests that have never run in CI. If any of them is unhappy on another runner, this PR takes the blame.

samremes: True, this has to be another PR that fixes the situation properly also for some existing flydsl tests.

aiter/ops/flydsl/kernels/gdr_decode.py
waqahmed-amd-fi: Both fixed. The state store derives its vector type from state_dtype_ now, and activations are restricted to fp16/bf16 — asserted in the kernel factory, rejected with a clear error in the wrapper so fp32 callers don't trip an assert deep inside. Added mixed coverage parametrized over both activation types; it fails on fp16 if I revert just the vector type.

On whether it can happen: not today, no tuned row uses fp16 activations. But bf16 state is half the table (2688 rows), fp16 activations are already in the tests, and the two are independent arguments — so only coincidence keeps them apart, and the failure is silent state corruption. The fix costs nothing at runtime and the scalar path stays bit-identical, so I kept it.

aiter/ops/torch_ref/kda.py
waqahmed-amd-fi: Fixed. kda_gate branches on a.dim() now, and validates the head axis against A_log plus dt_bias against the selected mode, so a tensor meant for the other gate raises instead of broadcasting. Added a square H == K case that fails under the old shape-based rule.

op_tests/test_kda_torch_ref.py
waqahmed-amd-fi: I'd keep it. aiter/ops/torch_ref/kda.py is shipped code, not test scaffolding, and it has three consumers: the KDA kernel tests, the benchmark, and this file. It's the oracle those kernel tests judge against. if it drifts, they either fail and send you into the kernel, or agree with a wrong kernel and pass.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Fixed, also in other files.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed all three export tests and the docstring's mention of them

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Revised. The helper module calls pytest.skip(allow_module_level=True) at import when flydsl or a GPU is missing, so this import can skip this file's tests as a side effect.

op_tests/test_kda_torch_ref.py
samremes: I'd say that's the wrong place to put the `torch_ref/kda.py` then, it is not supposed to be "shipped code" in the sense that somebody would use it. The reference should be test-only oracle implementation that is simple enough to reason about without tests. If we want to have some tests for it, then the reference that the reference is judged against has to be something else, especially not the flydsl kernel, and also not something that just essentially reduces to `torch.something() == torch.something()`

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Fixed.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Switched to checkAllclose and dropped the vLLM reference. The criterion is now stated plainly as rtol/atol of 1e-3. Kept a thin wrapper because checkAllclose returns a mismatch fraction instead of asserting, so calling it bare would make these tests unable to fail; the wrapper asserts on the return.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: comments revised.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Dropped the CSV reference and the K3 framing, and stated the reason instead

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: I checked the lookup: H=8 in the first case falls back too, so H=64 wasn't special. The real distinction is tuned vs fallback launch geometry. Reworded to say that, and dropped the K3 framing.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Trimmed to just the statement of what's tested.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Switched to empty_strided(pool.shape, pool.stride()) plus a copy, with the stride assertion you suggested

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Comment revised.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Removed.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: There's no ambiguity to resolve here, that was my docstring's fault. 

aiter/ops/torch_ref/kda.py has no slot or index handling at all; it takes initial_state directly and processes every row, so it can't disagree with the kernel. The state_idx <= 0 behavior I described belongs to the upstream vLLM reference, which we didn't port. There's a single contract: the kernel's read_pool_idx >= 0 & write_pool_idx >= 0 guard.

I've rewritten the docstring to state just that. I'd like to keep the test, since it's the only coverage of that guard — nothing else in the repo passes a negative index — but happy to drop it if you'd still prefer.

op_tests/test_flydsl_gdr_decode.py
waqahmed-amd-fi: Rewritten. 

It now patches the parsed config map with rows built in the test and asserts only the lookup: two rows differing only in gate_mode each resolve to their own config, and with just a per-channel row present the scalar gate falls back.

op_tests/test_kda_torch_ref.py
waqahmed-amd-fi: Agreed, and done in b4e6545856.

`aiter/ops/torch_ref/kda.py` -> `op_tests/kda_ref.py`, so test-only is structural rather than a claim in a docstring, and `op_tests/test_kda_torch_ref.py` is gone.

The gate is the KDA form only now. The scalar softplus branch and the rank dispatch that chose between the two lost their last caller with that test, and the kernel's scalar path is already covered by the Triton reference in `test_flydsl_linear_attention.py`. What's left is a port of vLLM's `naive_recurrent_kda` plus a two-line gate, which is the "simple enough to reason about without tests" bar.

The benchmark under `op_tests/op_benchmarks/flydsl/` imports it via a repo-root `sys.path` insert, since it is no longer reachable from an installed wheel.

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -378,20 +444,72 @@ def flydsl_gdr_decode(
...
+                f"`{name}` must have shape {shape} for a {gate_mode} decode with "
+                f"query {qk} and value {v}; got {tuple(tensor.shape)}."
+            )
+
+    if gate_mode == "kda":
+        # `a` keeps its strides and is vector-loaded along D_k, so that axis
+        # must be dense. dt_bias is copied contiguous below, so it is free.
+        assert (
+            a.stride(-1) == 1
+        ), f"`a` must be dense along D_k, got stride {a.stride(-1)}"
```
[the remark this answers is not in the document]
waqahmed-amd-fi: Addressed in commit b4b64776a3c737c9034cf91baa03c4d48307bbd

aiter/ops/flydsl/linear_attention_kernels.py
```diff
@@ -341,10 +400,10 @@ def flydsl_gdr_decode(
     read_indices: torch.Tensor | None = None,
     write_indices: torch.Tensor | None = None,
 ):
-    if stream is None:
-        stream = torch.cuda.current_stream()
     device = query.device
     dtype = query.dtype
+    if stream is None:
+        stream = torch.cuda.current_stream(device)
```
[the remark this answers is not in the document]
waqahmed-amd-fi: Addressed in commit b4b64776a3c737c9034cf91baa03c4d48307bbd

op_tests/test_flydsl_gdr_decode.py
```diff
@@ -0,0 +1,567 @@
...
+"""CI-visible numerics tests for the FlyDSL GDR decode kernel.
+
+CI shards ``op_tests/test_*.py`` at depth 1 (``.github/scripts/split_tests.sh``)
+and runs each as ``python3 <file>`` (``.github/scripts/aiter_test.sh``), so the
+in-package suite at ``aiter/ops/flydsl/test_flydsl_linear_attention.py`` never
+runs automatically, and a file without the ``__main__`` entry below collects
+nothing and still exits 0. Being listed is not the same as being run.
+"""
+
+import pytest
```
samremes, line 13: These tests still need quite a bit of refactor: remove pytest (different test cases directly to main() according to aiter op-test skill). This was also requested e.g. here by an aiter maintainer: https://github.com/ROCm/aiter/pull/4970#discussion_r4034515389

waqahmed-amd-fi: Addressed in commit [2943f3b](https://github.com/ROCm/aiter/pull/4577/commits/2943f3bec70beb1e69dee1d7485c48bbb10bc38c)

PR #4581, opened 2026-08-05
[Bug] Make blockscale split-K deterministic

maeehart:
## Summary

Make CK and CKTile blockscale GEMM output deterministic for configured split-K exponents greater than 1.

`splitK` is an exponent, so `KBatch = 1 << splitK`. Exponents 2 and 3 create four and eight partial sums that are accumulated into the output with atomic adds. Their completion order can change BF16 or FP16 results between identical calls.

Both compiled entry points now validate the requested exponent, preserve values 0 and 1, and clamp larger valid exponents to 1. This retains the CSV-selected kernel and the bitwise-repeatable two-partition path instead of bypassing tuned dispatch.

The CK and CKTile tuning loops are also capped at exponent 1. Direct regression coverage requests exponents 2 and 3 from both entry points and compares 20 outputs against an exponent 1 reference.

This addresses the root cause behind [vLLM #42240](https://github.com/vllm-project/vllm/pull/42240).

## Validation

- Focused blockscale codegen tests: `4 passed in 19.64s`.
- Before the patch, all four CK and CKTile exponent 2 and 3 cases reproduced nondeterminism on MI355X. Maximum absolute differences from the exponent 1 reference were 2.0, 2.0, 2.0, and 2.5.
- After rebuilding both modules from the patched sources: `4 passed in 541.10s`.
- Ruff check, Ruff format check, Python compilation, and `git diff --check` passed.


yzhou103:
there are serveral config files have splitK=2/3 entries that were selected by the tuner precisely because they were faster. Clamping at the C++ layer silently degrades these to splitK=1 without re-tuning. We should either re-tune affected shapes or document the expected perf impact

maeehart:
The concern is valid. I enumerated every tuned blockscale row with a stored `splitK` value of 2 or 3. There are 431 affected non-preshuffle CK and CKTile rows. The remaining rows use different preshuffle or ASM semantics and are not changed by this clamp.

Keeping the original tile and silently changing only KBatch regressed median operator device time by 24.95% on gfx950 and 39.48% on gfx942. The p95 regressions were 106.07% and 83.78%.

I retuned all 431 affected rows against deterministic exponent 0 and 1 candidates on matching gfx950/256-CU and gfx942/304-CU hardware. The raw exponent-1 and production-clamped paths differed by 8.12% at p95 on gfx950, so a new candidate is used only when it is more than 10% faster than the clamped original kernel. Otherwise the existing kernel is retained with an explicit exponent of 1. After selection, the median impact relative to the original higher split-K choices is 10.61% on gfx950 and 34.23% on gfx942.

The original exponent-2/3 choices produced multiple outputs for 384 rows. All 431 selected deterministic configurations passed the BF16-cast dequantized FP32 GEMM reference check and were bitwise stable across 20 calls.

[`082a56374`](https://github.com/ROCm/aiter/commit/082a56374) updates the four config files, adds a config guard against future CK and CKTile rows above exponent 1, and tests historical kernel names on both architectures. The C++ clamp remains as defense in depth. A deterministic workspace reduction for higher split-K can be evaluated separately.

PR #4584, opened 2026-08-06
[Triton/Gluon] kda gluon gfx1250 implementation

omuhamma:
## Motivation

Kimi delta attention added to aiter for gfx1250 with TDM support

## Test Plan

Unit tests + benchmark 

## Test Result

All unit tests pass 

benchmarks: 
image

When not updating the state pool with every state we save a lot of runtime when the tokens and batch size grows: 

image

## Submission Checklist


vgokhale:
Can you add perf numbers for a few shapes? Otherwise LGTM

vgokhale (dismissed):

op_tests/op_benchmarks/triton/bench_kda.py
```diff
@@ -0,0 +1,461 @@
...
+    """
+    if args.timing == "cudagraph":
+        return _bench_graph(fn, args.graph_ms, args.n_replays)
+    return triton.testing.do_bench(
+        fn, warmup=args.warmup, rep=args.rep, quantiles=[0.5, 0.2, 0.8]
+    )
+
+
+def _bench_graph(fn, graph_ms, n_replays):
+    """HIP-graph replay, same shape as the a8w8 blockscale preshuffle bench: probe
```
cagrikymk: Can we trim these comments a bit? They seem too verbose.

op_tests/op_benchmarks/triton/bench_kda.py
```diff
@@ -0,0 +1,461 @@
...
+    state = B * slab + (B * T * slab if paged else B * slab)
+    e = torch.tensor([], dtype=dtype).element_size()
+    per_tok = 3 * H * D * e + H * D * 4 + H * 4 + H * D * e
+    return state + B * T * per_tok
+
+
+def _time(fn, args):
+    """One timing measurement, in ms.
+
+    do_bench brackets each iteration with a stream-ordered event pair.  Those are
```
cagrikymk: Can we trim these comments a bit? They seem too verbose.

aiter/ops/triton/attention/kda.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+    if allow_neg_eigval and not use_beta_sigmoid_in_kernel:
+        raise ValueError(
+            "allow_neg_eigval=True requires use_beta_sigmoid_in_kernel=True"
+        )
+
+    if "transpose_state_layout" in kwargs:
+        if state_v_first is not None:
+            raise ValueError(
+                "Cannot pass both `state_v_first` and the deprecated "
```
cagrikymk: Are the deprecated paths needed from the caller side?

aiter/ops/triton/attention/kda.py
```diff
@@ -0,0 +1,275 @@
...
+    use_qk_l2norm_in_kernel: bool = False,
+    use_gate_in_kernel: bool = False,
+    use_beta_sigmoid_in_kernel: bool = False,
+    allow_neg_eigval: bool = False,
+    lower_bound: float | None = None,
+    out: torch.Tensor | None = None,
+    state_out: torch.Tensor | None = None,
+    BV: int = 32,
+    SK: int | None = None,
+    num_warps: int | None = None,
```
cagrikymk: I am not sure if this is the best approach to reveal kernel specific configs.

Maybe the best configs can be set in the kernel wrapper, and the user can override by providing a config dictionary (which is None by default), similar to https://github.com/ROCm/aiter/blob/4367ef10575563d5daf15f0ae362f5a9f8709554/aiter/ops/triton/gemm/basic/gemm_a16w16.py#L344

What do you think?

cagrikymk (dismissed):
LGTM!

aiter/ops/triton/_gluon_kernels/gfx1250/attention/kda_decode.py
```diff
@@ -0,0 +1,746 @@
...
+    return gl.BlockedLayout([1, K // SK], [32 // SK, SK], [NUM_WARPS, 1], [1, 0])
+
+
+@gluon.constexpr_function
+def _v_row_layout(ROWS, SK, NUM_WARPS):
+    return gl.BlockedLayout([1, ROWS], [SK, 32 // SK], [1, NUM_WARPS], [1, 0])
+
+
+@gluon.jit
+def _exp_scaled(scale, x):
```
vgokhale: Can you move this to a /_gluon_kernels/common/utils.py or some kind of file like that?

zufayu:

**Adds a gfx1250-only Gluon decode kernel for Kimi Delta Attention (per token S ← exp(g_t)·S, S ← S + β_t·k_t⊗(v_t − Sᵀk_t), o_t = Sᵀq_t) with paged per-token state slots, varlen, speculative decoding, cached (a,k,err) state updates, in-kernel gate/l2norm/sigmoid fusion, and opt-in TDM async-copy paths, behind a new standalone wrapper that nothing in aiter imports yet.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0); the single test target op_tests/triton_tests/attention/test_kda.py is gfx1250-gated and skips wholesale on this gfx942 box (224 skipped, run here), so no deterministic run was possible
Perf (advisory): NOT RUN — the target exposes no benchmark entry point and the PR's own op_tests/op_benchmarks/triton/bench_kda.py exits "gfx1250 only" on this box; the PR description itself carries no numeric perf data (two trace screenshots only)

🔴 [verified] aiter/ops/triton/attention/kda.py sets stride_indices_seq=1 for 1-D ssm_state_indices while the kernel indexes state_indices_ptr + i_n*stride_indices_seq + t, which is only correct for N==1 — the docstring's "[N, T] (or flat)" invites flat tables, and the wrapper accepts a flat [N*T] int32 table with N=2, T=2 (verified here by launching it), so sequence 1 reads its seed state from indices[1] — sequence 0's token-1 slot — instead of indices[2] and writes its snapshots into other sequences' slots, silently corrupting the pool with no error. **Author must** assert ndim==2 (or N==1 for 1-D, or derive the row stride from the sequence length) so the flat form either works or is rejected.
🔴 [verified] aiter/ops/triton/attention/kda.py validates dt_bias (numel==HV*K) and A_log contiguity and passes USE_LOWER_BOUND=True to the launch even when use_gate_in_kernel is left at its default False (verified by launching the wrapper with A_log + dt_bias + lower_bound=-5.0), but the kernel then computes a=exp(g) with no gate chain — so a caller that passes the K3 gate inputs without the flag gets silently ungated decay and garbage output, no error, while the wrapper's own shape validation signals the parameters are consumed. **Author must** assert A_log/dt_bias/lower_bound require use_gate_in_kernel=True (or forward them into the ungated path's error message).
⚠️ [verified] The whole 224-item suite in op_tests/triton_tests/attention/test_kda.py skips on every runner this PR's CI triggered (MI300X gfx942 / MI35X gfx950 — run here: 224 skipped on gfx942), and the only gfx1250 workflow (.github/workflows/ffm-bringup.yaml, label ci:gfx1250-ffm-triton) was never added, so "All unit tests pass" in the description rests entirely on the author's local run and no CI execution backs a 734-line kernel. **Author must** add the ci:gfx1250-ffm-triton label (or post a local gfx1250 run log) before merge.
⚠️ [verified] The PR's only perf evidence is two trace screenshots with no numeric values — its own benchmark op_tests/op_benchmarks/triton/bench_kda.py is gfx1250-only and published no table, maintainer vgokhale asked for numbers on Aug 6 and none were added — and no bucket in aiter/ops/triton/configs/gfx1250-KDA_DECODE-DEFAULT.json enables any advertised TDM path (the JSON sets only BV/SK/num_warps/num_buffers, and use_tdm_fused_load additionally calls tdm.async_load_fused, which stock triton 3.7.0 — checked here — does not have), so a reviewer can neither read the claimed speedup nor tell which configuration the screenshots ran. **Author must** publish numeric gluon-vs-fla tables for the K3 shapes (H=24/12, B=4–256, T=1–8) and name the config each number was run with.
⚠️ [inferred] aiter/ops/triton/attention/kda.py accepts num_accepted_tokens containing 0 (verified by launch) and the kernel then computes seed = load(num_accepted_ptr + i_n) − 1 = −1 and reads state_indices_ptr + i_n*stride_indices_seq − 1 — four bytes before the slot row, before the tensor for sequence 0 — then uses the garbage as a state slot index; worth checking whether any speculative-decode caller can report zero accepted tokens, since vLLM/SGLang verifiers normally guarantee ≥1. **Author must** assert (num_accepted_tokens >= 1).all() or clamp the seed to 0.

PR #4587, opened 2026-08-06
[fix][mla] keep get_meta_param's split-offset table alive for captured CUDA graphs

Duyi-Wang:
# Fix: `get_meta_param` frees a device tensor that captured CUDA graphs still read

## Summary

`aiter/mla.py::get_meta_param` returns a device tensor (`num_kv_splits_indptr`)
from inside a bounded `@functools.lru_cache`. The cache is that tensor's only
owner, so once the cache overflows, the tensor is freed and its block goes back
to the caching allocator — while any CUDA graph captured against it still has
the address baked into its kernel node. Replaying such a graph reads whatever
now occupies that block as a split-offset table, and MLA decode produces
non-finite `logits` / `attn_lse` from finite inputs.

This splits the two things the function was conflating: the split-count search
(pure, stays cached) and the device-side offset table (must outlive every graph
captured against it, so it now comes from a container that never evicts).

## The defect

```python
@functools.lru_cache                     # default maxsize = 128
def get_meta_param(num_kv_splits, bs, total_kv, nhead, max_seqlen_q, dtype,
                   tg_factor=1, ignore_total_kv=0):
    ...
    num_kv_splits_indptr = torch.arange(
        0, (bs + 1) * num_kv_splits, num_kv_splits, dtype=torch.int, device="cuda"
    )
    return num_kv_splits, num_kv_splits_indptr
```

The non-persistent MLA decode path passes `num_kv_splits_indptr` straight into
`aiter.mla_decode_stage1_asm_fwd` (`aiter/mla.py:414`) and the triton stage-2
merge loads from it (`aiter/mla.py:51-53`). Called under CUDA graph capture, the
tensor's device address is baked into the captured kernel node. Nothing else in
Python holds a reference to it.

Three properties combine into the bug:

1. the returned tuple contains a device tensor;
2. that address is baked into a graph whose lifetime far exceeds the call;
3. the cache key contains `bs`, so a framework capturing one decode graph per
   batch-size bucket inserts one entry per bucket — past 128 buckets, eviction
   is guaranteed.

## Why it shows up as a batch-size threshold

Graph runners capture **largest bucket first** so smaller graphs can reuse the
big graph's memory (sglang: `_capture_one_stream` walks `reversed(capture_bs)`).
With a dense capture list such as `--cuda-graph-bs $(seq 1 256)`, entries are
inserted in the order bs = 256, 255, …, 1. The first 128 inserted are exactly
bs 256 … 129, and they are exactly the 128 evicted while buckets 128 … 1 go in.

So every graph with **bs > 128** replays against a freed block and every graph
with bs ≤ 128 is intact. The 128 is `functools.lru_cache`'s default `maxsize`,
not anything about batch size — which is why the symptom looks like a capacity
limit.

Dense capture lists are a normal deployment choice, not a corner case: sglang's
own default list is sparse above 32, but harnesses commonly pass a contiguous
range.

## The fix

```python
_num_kv_splits_indptr_cache = {}


def _get_num_kv_splits_indptr(bs, num_kv_splits, device="cuda"):
    device = torch.device(device)
    if device.type == "cuda" and device.index is None:
        device = torch.device("cuda", torch.cuda.current_device())
    key = (bs, num_kv_splits, str(device))
    indptr = _num_kv_splits_indptr_cache.get(key)
    if indptr is None:
        indptr = torch.arange(
            0, (bs + 1) * num_kv_splits, num_kv_splits,
            dtype=torch.int, device=device,
        )
        _num_kv_splits_indptr_cache[key] = indptr
    return indptr


@functools.lru_cache
def _pick_num_kv_splits(...):      # the original body, returning the int only
    ...
    return num_kv_splits


def get_meta_param(num_kv_splits, bs, total_kv, nhead, max_seqlen_q, dtype,
                   tg_factor=1, ignore_total_kv=0):
    num_kv_splits = _pick_num_kv_splits(
        num_kv_splits, bs, total_kv, nhead, max_seqlen_q, dtype,
        tg_factor, ignore_total_kv,
    )
    return num_kv_splits, _get_num_kv_splits_indptr(bs, num_kv_splits)
```

Two decisions worth calling out:

**Why re-key the table on `(bs, num_kv_splits)` instead of the full argument
tuple.** The original key contains `total_kv`, which changes on nearly every
eager call, so simply making that cache unbounded would grow without limit. The
table's contents depend only on `bs` and `num_kv_splits`, so this key space is
capped at `max_bs × 16` entries of `(bs+1)` int32 — a few hundred KB — which
makes never evicting safe. The device ordinal is resolved before keying so an
entry created under one device is never handed to a kernel launched on another.

**Why not just allocate a fresh tensor per call.** A tensor allocated inside a
capture and freed afterwards is returned to the graph-private memory pool and
handed to the next capture, which reproduces the same bug by a different route.
The tensor has to be held for the lifetime of the graphs that reference it.

The public signature and return value are unchanged; in-tree callers that
already pass an explicit `num_kv_splits` or `split_indptr` are unaffected. The
returned table is read-only on every path (`tl.load` only), so sharing one
tensor across callers with the same `(bs, num_kv_splits)` is safe.

## Validation

**Op level**, from `op_tests/test_mla_meta_param_lru_eviction.py` (see the
section below — that file is for review and will be reverted before merge).
With the unfixed cache, walking the buckets in capture order shows the first
bucket's tensor dropped exactly when the 129th entry is inserted, and a
subsequent same-size allocation lands on the freed block. Replaying a graph
captured against that block, after its contents change, produces massive
corruption — measured at bs=165 / `num_kv_splits=3`, against 0 non-finite values
with the table intact:

| block contents at replay | non-finite values in the decode output |
|---|---|
| intact | 0 |
| zeros | 21,626,880 |
| all `-1` | 21,626,880 |
| `0x7FFFFFFF` | 21,626,880 |
| random int32 | 10,747,904 |

An eager call against the same contents matches the graph result in every case,
confirming the graph reads exactly this tensor.

(A monotonically increasing table with a larger stride is tolerated by the
kernel and leaves the output unchanged — worth knowing when writing tests, since
a recycled block generally holds arbitrary bytes rather than a plausible table.)

**End to end.** DeepSeek-R1-0528 MXFP4 on 8× MI355X (gfx950), tp/ep/dp 8 with
dp-attention, fp8 KV, non-persistent MLA, dense `--cuda-graph-bs $(seq 1 256)`,
gsm8k 5-shot at concurrency 2048. All arms verified to capture 256 buckets and
to reach a peak per-rank batch of 165 on the non-persistent path:

| build | gsm8k strict-match |
|---|---|
| unpatched | 0.0318 |
| **this fix** | **0.9560** |

Two additional controls confirm the attribution rather than a side effect:
making the existing cache unbounded and nothing else recovers to 0.9575; keeping
the cache bounded and still evicting, but allocating the table from an
arena that is never released while preserving an equivalent alloc/free
rate, recovers to 0.9591. The operative variable is that this table's block must
not be recycled.

## Repro test — for review only, reverted before merge

The second commit adds `op_tests/test_mla_meta_param_lru_eviction.py` so the
defect can be inspected and re-measured during review. **It is not proposed for
merge** and will be reverted once reviewed: it needs a gfx950 GPU, allocates
several GB to build a 256-bucket capture ladder, and one of its cases
deliberately corrupts a live buffer — none of which belongs in CI.

```
pytest -q -s op_tests/test_mla_meta_param_lru_eviction.py
```

Single GPU, no extra setup. On non-gfx950 every case skips with a reason.

Expected:

| build | result |
|---|---|
| without this PR's first commit | `13 passed` |
| with it | `12 passed, 1 skipped` |

The skip is `test_lru_evicts_live_graph_tensor`, which reports that
`get_meta_param` no longer carries an LRU and so has nothing to evict.

**`test_lru_evicts_live_graph_tensor`** — the ownership problem, no graph
capture involved. Calls `get_meta_param` for 256 distinct `bs` in the order a
graph runner captures them (largest first), holding a `weakref` to the first
bucket's tensor, and reports which insertion kills it; then allocates 8 fresh
same-size int32 tensors and reports how many land on the freed address.

```
tensor died after 129 buckets were inserted
8 fresh int32 tensors of the same size -> 1 landed on 0x7fbcd9a00000
```

The `weakref` is deliberate: reading the entry back through the cache to test
whether it is still there moves it to the MRU end and postpones the eviction
being measured.

**`test_stale_split_table_corrupts_replay[bs130|bs165 × 5 patterns]`** — the
causal step. Captures a graph, replays it with the table intact as a control,
then rewrites the table in place and replays again, cross-checking each result
against an eager call so that "the graph is reading some other buffer" is ruled
out. The patterns cover what a recycled block can plausibly hold; the
`neighbour` one (a valid table for a larger split count) is asserted to stay
clean, because the kernel tolerates monotonically increasing offsets — a test
that tried only that pattern would be a false negative.

**`test_capture_many_buckets_matches_eager[buckets64|buckets256]`** — end-to-end
capture and replay over a bucket ladder sharing one graph pool, replaying bucket
165. Not the largest bucket: bs=256 auto-picks `num_kv_splits=1` and takes the
single-pass path, which never reads the table.

> This case is **not a gate**, and passes even without the fix. It prints the
> eviction count it observed, and also prints how many later allocations landed
> on the freed block — in a minimal harness that is usually zero, because almost
> nothing is allocated between capture and replay, so the freed table survives
> byte-intact and the graph reads correct data by luck. When that happens the
> case emits an explicit note saying it is not a clean bill of health. The
> gating evidence is the first two cases.

## Related, not addressed here

The same shape — a bounded cache owning a device tensor — appears elsewhere in
the tree. Whether each is reachable under graph capture depends on its call
path, and none were audited as part of this change:

| location | maxsize |
|---|---|
| `aiter/ops/topk.py:315 _get_topk_mb_workspace_keyed` | 16 |
| `aiter/ops/gemm_op_a16w16.py:37 _get_semaphore_workspace_keyed` | 64 |
| `aiter/ops/flydsl/gemm_kernels.py:701 _get_split_k_tensors` | 128 |
| `aiter/ops/gemm_op_a8w8.py:386 get_zero_bias_buf_keyed` | 1024 |
| `aiter/fused_moe_dp_shared_expert.py:40/66 get_dp_shared_expert_stage{1,2}_moe_sorting_result` | 1024 |

The rule of thumb: if a device tensor can be created during CUDA graph capture
and handed to a kernel, whatever holds it must not be bounded.

---

### DIFF
--- a/aiter/mla.py
+++ b/aiter/mla.py
@@ -122,8 +122,48 @@ def _fwd_kernel_stage2_asm(
                 )
 
 
+# Device-side split-offset tables handed straight to `mla_decode_stage1_asm_fwd`.
+#
+# A CUDA graph capture bakes this pointer into the kernel node, so the tensor has
+# to outlive every graph captured against it. It used to live in
+# `get_meta_param`'s `@functools.lru_cache`, whose default maxsize is 128 -- with
+# the cache as its only owner, entry number 129 evicted the first one, freed the
+# block back to the caching allocator, and left the already-captured graph
+# reading whatever took its place as a split-offset table. Graph runners capture
+# largest-batch-first, so with a dense bucket list (e.g. `--cuda-graph-bs
+# $(seq 1 256)`) precisely the buckets above 128 were the ones freed out from
+# under their graphs.
+#
+# Keyed on (bs, num_kv_splits, device) rather than on `get_meta_param`'s full
+# argument tuple: `total_kv` varies every step in serving, so keying on it would
+# grow without bound. This key space is capped at max_bs * 16 entries of
+# (bs+1) int32 -- a few hundred KB -- so it is safe to never evict.
+_num_kv_splits_indptr_cache = {}
+
+
+def _get_num_kv_splits_indptr(bs, num_kv_splits, device="cuda"):
+    # Resolve "cuda" to the concrete ordinal before keying: the bare string means
+    # "current device", so a cached entry created under one device must not be
+    # handed to a kernel launched on another.
+    device = torch.device(device)
+    if device.type == "cuda" and device.index is None:
+        device = torch.device("cuda", torch.cuda.current_device())
+    key = (bs, num_kv_splits, str(device))
+    indptr = _num_kv_splits_indptr_cache.get(key)
+    if indptr is None:
+        indptr = torch.arange(
+            0,
+            (bs + 1) * num_kv_splits,
+            num_kv_splits,
+            dtype=torch.int,
+            device=device,
+        )
+        _num_kv_splits_indptr_cache[key] = indptr
+    return indptr
+
+
 @functools.lru_cache
-def get_meta_param(
+def _pick_num_kv_splits(
     num_kv_splits,
     bs,
     total_kv,
@@ -212,11 +252,36 @@ def get_meta_param(
                 int(abs(total_kv / bs - max_seqlen_q) // min_block_n) + 1,
             )
 
-    num_kv_splits_indptr = torch.arange(
-        0, (bs + 1) * num_kv_splits, num_kv_splits, dtype=torch.int, device="cuda"
-    )
+    return num_kv_splits
 
-    return num_kv_splits, num_kv_splits_indptr
+
+def get_meta_param(
+    num_kv_splits,
+    bs,
+    total_kv,
+    nhead,
+    max_seqlen_q,
+    dtype,
+    tg_factor=1,
+    ignore_total_kv=0,
+):
+    """Split count plus its device-side offset table.
+
+    The split-count search is pure and stays cached; the table it indexes is
+    handed out from `_get_num_kv_splits_indptr`, which never evicts, so a
+    captured CUDA graph keeps reading live memory.
+    """
+    num_kv_splits = _pick_num_kv_splits(
+        num_kv_splits,
+        bs,
+        total_kv,
+        nhead,
+        max_seqlen_q,
+        dtype,
+        tg_factor,
+        ignore_total_kv,
+    )
+    return num_kv_splits, _get_num_kv_splits_indptr(bs, num_kv_splits)
 
 
 # Persistent MLA-decode kernel gate: the persistent kernel
--- a/op_tests/test_mla_meta_param_lru_eviction.py
+++ b/op_tests/test_mla_meta_param_lru_eviction.py
@@ -1,0 +1,587 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""`get_meta_param`'s LRU cache frees a tensor a captured CUDA graph still reads.
+
+The defect
+----------
+`aiter/mla.py`::
+
+    @functools.lru_cache                     # <-- default maxsize = 128
+    def get_meta_param(num_kv_splits, bs, total_kv, nhead, max_seqlen_q, dtype,
+                       tg_factor=1, ignore_total_kv=0):
+        ...
+        num_kv_splits_indptr = torch.arange(
+            0, (bs + 1) * num_kv_splits, num_kv_splits, dtype=torch.int, device="cuda"
+        )
+        return num_kv_splits, num_kv_splits_indptr
+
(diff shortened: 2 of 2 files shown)


valarLip:
maybe call get_meta_param in framwork

PR #4605, opened 2026-08-06
Fix reduce_scatter pynccl path

yitingw1:
## Motivation

Fix the remaining issue in https://github.com/ROCm/aiter/pull/3464

**Problem:** reduce_scatter pynccl fallback silently returns wrong results for dim != 0 — movedim direction is backwards and two return values (reshape/movedim) are discarded.

**Fix:** correct movedim(dim, 0) → temp → copy-back in both reduce_scatter and reduce_scatterv.

**Test:** new fallback suite (dim=0/1/2) — max_abs_err 112/88 before fix, 0 after.

## Submission Checklist


TennyWang1223 (approved):
LGTM

valarLip (approved):

PR #4606, opened 2026-08-06
[FlyDSL] Tiled-copy refactor + cleanup: swiglu/silu_fq/qk_norm_rope/fp8_mqa

coderfeli:
## Summary

Modernizes four FlyDSL element-wise / per-token kernels onto the layout-API **tiled copy** (`make_layout_tv` / `make_tiled_copy` + `partition_S/D`), replaces raw MLIR (`arith`/`ArithValue`/`GTensor`) with the `fx.*` surface, and trims comments/dead code. Net **−285 LOC** across the touched files.

| kernel | change | correctness | perf |
|---|---|---|---|
| `swiglu_and_mul` | scalar dword gather → TV-tiled copy | byte-exact | **~1.06–1.10× faster** (6.0→6.5 TB/s) |
| `silu_and_mul_fq` | gate/up loads + packed fp4/fp8/bf16 store → TV-tiled copy | byte-exact (output + scales) | ISA-neutral |
| `qk_norm_rope_quant` | **GTensor → base-folded tiled copy** (Q/KV/SWA ring+paged) | op-test + byte-exact | ISA-identical |
| `fp8_mqa_logits` | raw dialects → `fx.*`, comment trim | byte-exact | ISA-identical |

## Approach

- The "interleave" in swiglu/silu is just a `(N0, 2, NLANE)` layout — `gate`/`up` are strided views, output is contiguous; one TV layout partitions all three.
- `GTensor.get_llvm_ptr(off)` is `base + off`; replaced by folding the per-token/row offset into the pointer (`inttoptr(ptrtoint(ptr)+off)`, i32-safe voffsets) + a buffer-tensor view.
- Cross-thread reductions (`shuffle_xor` amax / RMSNorm sumsq) are preserved: the TV layout keeps thread `t` owning contiguous `[t*VEC, (t+1)*VEC)`, so quant-block/head threads stay shuffle-adjacent.
- Kept raw ops **only** at genuine boundaries: the `emit_mx_e8m0_scale` / `cvt_pk_fp8_f32` raw-IR builders, `rocdl.mfma`, index-typed loop bounds, and the single-op integer `max/min/ceildiv` helpers (fx `.select` costs 2 ops vs `v_max_i32`; float `.maximumf` is wrong for ints). The sparse swizzled E8M0 scale write (1 byte / 32 cols) stays a scalar scatter.

## Validation (gfx950, FlyDSL disk cache ON)

Every kernel: **byte-exact** vs the original (separate-module compile + `torch.equal`) across its full config matrix, plus **ISA diff** as the perf gate (the shared GPU is too noisy for wall-clock). swiglu's tiled copy genuinely changes the ISA (scalar→vector, measured faster); the other three are ISA-identical/neutral. `qk_norm` also passes its op test.


aiter/ops/flydsl/kernels/silu_and_mul_fq.py
```diff
@@ -131,148 +106,227 @@ def silu_and_mul_fq_kernel(
...
+            pt = fx.PointerType.get(T.i32, address_space=_AS_G, alignment=4)
+            return fx.inttoptr(pt, fx.Int64(fx.ptrtoint(ptr)))
+
+        # Scalar E8M0 scale writes go through an indexed store on this i8 pointer
+        # over the sorted scale buffer (computed byte offset).
+        scale_i8_ptr = fx.inttoptr(
+            fx.PointerType.get(T.i8, address_space=_AS_G, alignment=1),
+            fx.Int64(fx.ptrtoint(out_scale_sorted)),
+        )
+        sorted_ids_p = _i32_ptr(sorted_ids)
```
lalala-sh, line 127: Dose there have cut risks of i32 pointer type?

zhiding512 (approved):

aiter/ops/flydsl/kernels/silu_and_mul_fq.py
coderfeli: this is data type ptr points to not ptr itself.

PR #4610, opened 2026-08-06
[FlyDSL] Add Kimi K3 Attention Residual kernel

anhminhnguyenhoang:
## Motivation

Add an aiter FlyDSL implementation of the Kimi K3 Attention Residuals (AttnRes) operator for AMD GPU gfx950.

The kernel fuses the residual update, optional block snapshot, depth-axis softmax mix, and optional output RMSNorm into a single launch. This replaces the current multi-launch AMD boundary and provides a production path for prefill and small-batch decode workloads.

Framework integration remains outside the scope of this PR.

### Files Changed

- `aiter/ops/flydsl/__init__.py` — exports `flydsl_attn_res` as a public FlyDSL operator when a compatible FlyDSL installation is available.
- `aiter/ops/flydsl/kernels/attn_res.py` — adds the fused AttnRes FlyDSL kernel and host wrapper.
- `aiter/ops/torch_ref/attn_res.py` — adds the independent pure-PyTorch correctness reference, including BF16 prefix-update semantics and optional output RMSNorm.
- `op_tests/test_flydsl_attn_res.py` — adds correctness coverage, side-effect checks, stride and shape validation, and the performance benchmark.

### Kernel Architecture

- The host wrapper selects a cached compile-time specialization based on hidden size, source count, optional-operation flags, and rows packed per workgroup.
- One workgroup processes one token row for the main configurations:
  - `D=7168`: 448 threads across seven waves
  - `D=1024`: 64 threads across one wave
- The `D=1024` path also supports packing multiple one-wave rows into a workgroup; the production default remains one row per workgroup.
- Source data is loaded as 128-bit BF16 vectors, with multiple source loads kept in flight while the current source is reduced.
- Each source computes FP32 `sumsq` and dot-product partials. These partials are reduced per wave, staged in ping-pong LDS storage, and combined across waves.
- One-wave rows use DPP reductions. Multi-wave `D=7168` rows use LDS-assisted synchronization followed by a compact DPP finish reduction.
- The fused execution order is:

  `prefix/delta update → optional snapshot → source scoring and softmax mix → optional output RMSNorm → BF16 output store`

- Softmax scores use RMS-normalized sources, while the final weighted sum uses the raw source values. Prefix updates, reductions, and normalization accumulate in FP32 before the BF16 output store.
- The implementation supports unit trailing strides and valid padded leading strides with 16-byte-aligned row starts.

## Test Plan

- Run: `pytest -q op_tests/test_flydsl_attn_res.py`
- Compare outputs and side effects against the Torch reference across:
  - `num_blocks=0..8`
  - all optional-argument combinations
  - `D=7168` and `D=1024`
  - empty, single-token, prefill, and large-prefill inputs
  - valid padded strides
  - prefix and block snapshot side effects
- Validate invalid layouts and unsupported noncanonical writes.
- Compare representative configurations against a float64 reference.
- Measure wall time and bandwidth-floor efficiency across all benchmarked token counts.
- Validate generated ISA, VGPR usage, scratch usage, and barrier behavior on gfx950.

## Test Result

Correctness validation passed:

- `pytest -q op_tests/test_flydsl_attn_res.py`: **12 passed**.
- Prefix and block side effects matched the reference.
- The one-lane store reduction remained accurate against float64 reference.
- Maximum relative error was `3.863e-03`, below one BF16 ulp.
- The `D=1024` reduction path remained exactly identical in the comparison.

The following table reports measured kernel time, achieved effective bandwidth, and efficiency relative to the conservative cache-bypassed bandwidth of 6.26 TB/s. Each cell is `runtime / achieved bandwidth / efficiency`; achieved bandwidth is calculated from the configuration's streamed BF16 rows. Each runtime is the minimum of five 101-iteration measurements after warmup using `scripts/attnres_bandwidth_floor.py`. Relative to the previous sweep, the changes are within normal run-to-run variation for the large-D paths; the larger small-D differences are launch-bound and are not systematic.

| Configuration | T=1 | T=16 | T=64 | T=256 | T=320 | T=512 | T=1024 | T=4096 | T=8192 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| Mix `k=9`, `D=7168` | 7.822 us / 0.02 TB/s / 0.3% | 7.801 us / 0.29 TB/s / 4.7% | 7.886 us / 1.16 TB/s / 18.6% | 8.700 us / 4.22 TB/s / 67.4% | 15.319 us / 2.99 TB/s / 47.8% | 18.331 us / 4.00 TB/s / 64.0% | 32.392 us / 4.53 TB/s / 72.4% | 131.185 us / 4.48 TB/s / 71.5% | 245.483 us / 4.78 TB/s / 76.4% |
| All-on `k=8`, `D=7168` | 7.941 us / 0.02 TB/s / 0.3% | 7.944 us / 0.35 TB/s / 5.5% | 7.944 us / 1.39 TB/s / 22.1% | 9.457 us / 4.66 TB/s / 74.4% | 16.467 us / 3.34 TB/s / 53.4% | 19.562 us / 4.50 TB/s / 71.9% | 37.835 us / 4.66 TB/s / 74.4% | 153.787 us / 4.58 TB/s / 73.2% | 289.634 us / 4.87 TB/s / 77.7% |
| Mix `k=9`, `D=1024` | 3.898 us / 0.01 TB/s / 0.1% | 4.813 us / 0.07 TB/s / 1.1% | 4.858 us / 0.27 TB/s / 4.3% | 4.844 us / 1.08 TB/s / 17.3% | 4.840 us / 1.35 TB/s / 21.6% | 4.827 us / 2.17 TB/s / 34.7% | 5.208 us / 4.03 TB/s / 64.3% | 15.891 us / 5.28 TB/s / 84.3% | 28.401 us / 5.91 TB/s / 94.4% |

### Realistic Bandwidth Ceiling

On gfx950, the ticket's 0.85 BW efficiency equals `0.85 × 8 TB/s = 6.8 TB/s` is not a realistic target because AttnRes reuses weights, uses fixed-width BF16 transactions, and performs reduction, synchronization, and softmax work.

A more realistic theoretical ceiling is measured with a BF16 reference matching the `D=7168`, mix-only `k=9` path: nine source rows, two reused weight reads, and one output row, performs the same BF16-to-FP32 conversion and BF16 store, but omits reduction, barriers, and softmax. The latest measurements were **5.66–5.79 TB/s cached** and **6.26–6.28 TB/s cache-bypassed** and the table above uses the conservative 6.26 TB/s cache-bypassed bandwidth as its normalization rate.


aiter/ops/flydsl/kernels/attn_res.py
```diff
@@ -0,0 +1,808 @@
...
+keys, and mixes raw values with an online softmax accumulator.
+
+Delta update, canonical append snapshot, and output RMSNorm are independent
+compile-time specializations. ``block_write_idx=-1`` disables the snapshot;
+otherwise it must equal ``num_blocks`` in 0..7, so the current prefix is stored
+in the first unused block slot.
+"""
+
+# Do not add ``from __future__ import annotations``: FlyDSL inspects annotations
+# at trace time and PEP 563 can interfere with its runtime argument detection.
```
samremes: No reason to mention this here.

samremes (commented):
Kernel looks coherent on wave64: mix, fused output RMSNorm, delta/snapshot side effects, padded-row checks, and current-stream launch.

These are all relatively minor review comments — not functional blockers. Nits on the op-test `__main__` harness, a wave32 reject so gfx1250 does not compile a wave64 DPP reduce, a 16-byte weight alignment hole that will not trigger on typical parameter tensors, and some leftover manual layout algebra.

Inline notes below.

— Cursor

op_tests/test_flydsl_attn_res.py
```diff
@@ -0,0 +1,422 @@
...
+            write_block,
+            apply_output_norm,
+            pad,
+            rows_per_wg,
+        ), tokens in itertools.product(
+            (*large_d_configs, *small_d_configs),
+            args.tokens,
+        )
+    ]
+    print(pd.DataFrame(results).to_string(index=False))
```
samremes: Nit: Aiter Test CI runs `python3` on this file, so this `__main__` sweep is the path that matters. Please align it with the usual op-test harness: `get_gfx()` allow-list (`gfx942`/`gfx950`) before argparse, `checkAllclose` after an fp32 cast, `us` / TFLOPS / TB/s / `err` plus a `"gfx"` column, and `aiter.logger.info(..., df.to_markdown(index=False))`. The gfx950-only 8 TB/s peak also makes gfx942 `%peak` rows wrong.

Leaving `T=0` / layout-rejection extras in pytest is fine — in-tree practice is mixed and there is no single standard. Recommendation only: keep those extras untimed and sharing `_run_and_check`, rather than a second timed suite.

— Cursor

aiter/ops/flydsl/kernels/attn_res.py
```diff
@@ -0,0 +1,808 @@
...
+
+    alignment_bytes = _VEC_WIDTH * tensor.element_size()
+    if tensor.data_ptr() % alignment_bytes:
+        raise ValueError(
+            f"{name} base pointer must be {alignment_bytes}-byte aligned for "
+            "BufferCopy128b"
+        )
+
+
+def flydsl_attn_res(
```
samremes, line 679: Nit: please reject wave32 (e.g. gfx1250) here before compile. This kernel is a wave64 CDNA reduce (`_WARP_SIZE = 64`, `shuffle_xor` width 64, DPP `row_bcast` / `readlane(..., 63)`). Same allow-list in the test `main()` so an unsupported card skips instead of launching.

A wave32 implementation is not needed for this PR.

— Cursor

aiter/ops/flydsl/kernels/attn_res.py
```diff
@@ -0,0 +1,808 @@
...
+    ):
+        if tensor.stride(0) != 1:
+            raise ValueError(
+                f"{name}.stride(0) must be 1, got strides {tensor.stride()}"
+            )
+    if apply_output_norm and output_norm_weight.stride(0) != 1:
+        raise ValueError(
+            "output_norm_weight.stride(0) must be 1, got strides "
+            f"{output_norm_weight.stride()}"
+        )
```
samremes: Nit: rank-1 weights only check `stride(0) == 1`. Loads still go through BufferCopy128b, so a 2-byte-aligned slice such as `parent[1:1+D]` would miss the 16-byte base check that prefix/blocks get in `_check_row_layout`. Unlikely on a fresh `torch.empty((D,))` parameter, but the same alignment helper would close it.

— Cursor

aiter/ops/flydsl/kernels/attn_res.py
```diff
@@ -0,0 +1,808 @@
...
+        blocks: fx.Tensor,
+        norm_weight: fx.Tensor,
+        qk_weight: fx.Tensor,
+        output_norm_weight: fx.Tensor,
+        out: fx.Tensor,
+        num_tokens: fx.Int32,
+    ):
+        tid = fx.thread_idx.x
+        local_tid = tid % row_threads
+        token = fx.block_idx.x * rows_per_wg + tid // row_threads
```
samremes: Nit / layout algebra: rows and LDS already use `make_buffer_tensor` / `logical_divide` / `SharedAllocator`. These remaining maps are regular static layouts with runtime ids, so `idx2crd` / `crd2idx` on FlyDSL 0.3.2 would own them instead of repeating `tid % row_threads` and `local_tid + tile * row_threads` (also at the prefix/q prologue, snapshot store, and output store). Wave/lane for DPP can stay as hardware IDs.

Example matching the current formulas:

```python
row_owner = fx.make_layout((rows_per_wg, row_threads), (row_threads, 1))
row_coord = fx.idx2crd(tid, row_owner)
wg_row = fx.get(row_coord, 0)
local_tid = fx.get(row_coord, 1)
token = fx.block_idx.x * rows_per_wg + wg_row
```

Optional cleanup, not a correctness issue.

— Cursor

samremes:
Also checked this kernel against the vLLM AMD Triton AttnRes (`main`, `amd/ops/attn_res.py`) on gfx950 / MI355X, D=7168, same fusion flags. Same numerical contract within the usual bf16 tolerances.

Most representative latency (µs, lower is better):

| T | path | FlyDSL | vLLM Triton | vs Triton |
| ---: | --- | ---: | ---: | ---: |
| 1 | mix (k=9) | 12.4 | 16.5 | 1.33× |
| 1 | fused delta+write+norm (k=8) | 38.8 | 47.1 | 1.21× |
| 256 | mix (k=9) | 12.9 | 16.6 | 1.29× |
| 8192 | mix (k=9) | 238 | 320 | 1.34× |

Decode and typical prefill are about 20–35% faster. Very large fused prefill (T=8192, all flags on) is essentially tied (~757 vs ~760 µs).

— Cursor

PR #4617, opened 2026-08-06
feat(fused_moe): accept a caller-provided output buffer

RolaoDenthu:
## Motivation

`fused_moe` always allocates its own output tensor, so a caller that already owns the buffer the result must end up in has to copy afterwards. In SGLang's Kimi-K3 MoE block the routed output goes straight into a slice of a symmetric buffer that is about to be all-reduced, so every layer pays a `[M, model_dim]` D2D copy that moves no information.

## Technical Details

`_match_output_buffer` is the single predicate: the caller's tensor is used only when shape, dtype, device and contiguity all match what `torch.empty` would have produced; anything else falls back to a private allocation.

Adds an optional `output` argument to `fused_moe` / `fused_moe_`, threaded down to the two places where the output tensor is actually created: `moe_sorting` in accumulate mode and `fused_moe_2stages` in reduce mode.

Also aligns `fused_moe_fake`'s signature with `fused_moe_`, which was missing `beta` / `linear_beta`. torch fills a fake's arguments positionally from the schema generated off `fused_moe_`, so the gap made any call passing `beta` (e.g. the SiTU path) raise `TypeError` under `torch.compile`.

## Submission Checklist


yzhou103:
Thanks for this — the feature works on the paths I could exercise (verified on gfx950: bf16 QuantType.No, out is buf == True, numerics bit-identical to the no-output call). The
  fused_moe_fake signature fix is also a genuine bug catch. However, I think there are two blocking issues around the custom-op contract that need to be resolved before merge.
**But there are two problems:**
**1.  silent decline makes the fake diverge from the runtime → wrong results under torch.compile**

  aiter::pa_sparse_prefill_opus already uses this exact pattern (mutates_args=["out"], optional out, fake returns out), so returning the caller's buffer from the op is fine per repo
  precedent. What makes it safe there is that out is never silently declined — shape/dtype mismatch raises, and the buffer is used unconditionally otherwise. _match_output_buffer breaks
  that property, and the fake cannot see why:

  This doesn't hold. The fake tests only shape/dtype/device/contiguity, but the runtime decision also depends on env vars and tuned metadata:
image


  When they disagree, functionalization believes ret is output and folds later reads of the caller's buffer into the op's return. Reproduced on the real op (wrapping _moe_sorting_impl
  to force output=None, emulating the flat/aux paths):

  eager    (o.sum, buf.sum): (-2160.0, 0.0)      # correct: buf was never written
  compiled (o.sum, buf.sum): (-2160.0, -2160.0)  # wrong, and completely silent

  Fix: make output unconditionally honored, the way pa_sparse_prefill_opus does it.
  - shape/dtype/device/contiguity mismatch → raise, don't fall back. It's a caller bug, and raising is free and unambiguous.
  - paths that genuinely can't write in place (FLAT needs the trailing bytes; adaptive-aux; _fhmoe) → output.copy_(tmp) before returning output. That's the copy the caller would have
  done anyway, so nothing is lost, and the fake's prediction becomes correct by construction.

  This also removes the "compare identity to find out whether it was taken" contract, which I'd push back on independently: most callers will write moe(..., output=buf) and then read
  buf, and get stale data on every decline path.
**2. output aliasing an input silently corrupts data**

  In accumulate mode moe_sorting zeroes output before stage1 reads hidden_states:

  o = fused_moe(hs, w1, w2, tw, tid, output=hs)   # in-place attempt

  In reduce mode the same call is accidentally correct (stage2 writes last), so the semantics are path-dependent. Since "save the caller a copy" is exactly what invites
  output=hidden_states, please add a storage-overlap check and reject.

And, **lacking tests.**

RolaoDenthu:
> Fix: make output unconditionally honored, the way pa_sparse_prefill_opus does it.
> 
> * shape/dtype/device/contiguity mismatch → raise, don't fall back. It's a caller bug, and raising is free and unambiguous.
> * paths that genuinely can't write in place (FLAT needs the trailing bytes; adaptive-aux; _fhmoe) → output.copy_(tmp) before returning output. That's the copy the caller would have
>   done anyway, so nothing is lost, and the fake's prediction becomes correct by construction.
> 
> This also removes the "compare identity to find out whether it was taken" contract, which I'd push back on independently: most callers will write moe(..., output=buf) and then read buf, and get stale data on every decline path.

Thanks for the review! both issues are fixed, and there is a test now.

**1. Silent decline.** `_validate_output_buffer` raises runtime error on shape/dtype/device/contiguity mismatch, no fallback left. `_return_output(ret, output)` returns `output`, copying `ret` in unless `ret is output`; it sits on all four result-producing returns: fhmoe, grouped a8w4, 1stage (where FLAT lands, since `cfg_flat = run_1stage and ...`) and 2stages (where adaptive-aux lands when the sort is atomic). The fake is unconditional now, so the compare-identity contract is gone with it.

**2. Aliasing.** Overlap with `hidden_states` raises. The check compares `[data_ptr, data_ptr + nbytes)` rather than storage identity, since a slice of a shared pool is the motivating caller. Only `hidden_states` is checked since every other tensor argument fails the shape/dtype gate first.

**3. Tests.** `test_output_buffer_contract()` in op_tests/test_moe_2stage.py, run from the default sweep, covering both in-place and forced-copy paths in eager and under torch.compile.

aiter/fused_moe.py
```diff
@@ -526,6 +590,7 @@ def fused_moe(
             shared_w2_scale=shared_w2_scale,
             shared_expert_id=shared_expert_id,
         )
+        return _return_output(fhmoe_out, output)
```
yzhou103: it seems it will cause an additional copy for fhmoe.

RolaoDenthu: With `output=None` it returns on the first branch, so there is no copy, but there is one when `output` is passed. That was me following the suggestion:

> paths that genuinely can't write in place (FLAT needs the trailing bytes; adaptive-aux;
> fhmoe) → output.copy_(tmp) before returning output. That's the copy the caller would have
> done anyway, so nothing is lost, and the fake's prediction becomes correct by construction.

To remove it for the `output` case, I can add an `output` parameter to `fhmoe_` (and `fhmoe_fake`) so it can be threaded down to `_fused_moe_impl` and let it write directly into the caller's buffer. What do you think?

RolaoDenthu: Hi @yzhou103 I have added output buffer for fhmoe, could you please take a look? Thank you!

RolaoDenthu:
fixed conflict, and the ci failure looks not related to this pr.

aiter/fused_moe.py
```diff
@@ -610,13 +668,18 @@ def fused_moe_fake(
...
     ep_source_token_map: torch.Tensor | None = None,
+    output: torch.Tensor | None = None,
 ) -> torch.Tensor:
     device = topk_ids.device
     M, _topk = topk_ids.shape
     dtype = hidden_states.dtype if dtype is None else dtype
     model_dim = w2.shape[1]
-    moe_buf = torch.empty((M, model_dim), dtype=dtype, device=device)
-    return moe_buf
+    # Unconditional: the runtime returns `output` on every path, and which kernel
```
yzhou103, line 698: New blocker, introduced by the merge: output= + stage2_scatter= copies uninitialized memory
  
   aiter/ops/flydsl/grouped_moe_gfx1250.py:845:
      if enable_ep_scatter:
       # GEMM2 already P2P-wrote every route-weighted result into the peers'
       # combine buffers, so this only satisfies the custom-op's return contract;
       # MegaMoE ignores its contents and consumes comb_inp instead.
       return torch.empty((token_num, model_dim), dtype=dtype, device=device)
   That placeholder reaches fused_moe.py:920:
   if grouped_a8w4_out is not None:
       return _return_output(grouped_a8w4_out, output)   # -> output.copy_(uninitialized)
   I checked L789–L925 and nothing guards the combination. Low probability in practice (MegaMoE doesn't pass output, and scatter is gfx1250-only), but the two parameters are orthogonal at the API level, and fused_moe_fake unconditionally 
   promises that output is the result — so under torch.compile this combination silently reads garbage. In the PR's own motivating scenario the destination is a symmetric-memory slice that is about to be all-reduced, which makes it worse.
  
   Two-line fix, next to the existing check at fused_moe.py:822:
   if stage2_scatter is not None and output is not None:
       raise RuntimeError(
           "output= is incompatible with stage2_scatter: GEMM2 P2P-writes the "
           "route-weighted results into the peers' combine buffers and returns an "
           "uninitialized placeholder"
       )
   and mirror it in fused_moe_fake (before L679) using ep_source_token_map is not None, which is the same switch fused_moe_ uses at L723.

aiter/fused_moe.py
```diff
@@ -172,6 +172,41 @@ def _adaptive_moe_sort(
     return std
```
yzhou103, line 174: maybe it's better to add a comment to record the different action in different path as there are many patch in fmoe:
 ```
 `output`: caller-provided [M, model_dim] destination.
  
     path                                   in-place?  why
     opus / ck / flydsl sort + accumulate   yes        sort kernel zeroes what it is handed
     ... + reduce mode                      yes        buffer is born in fused_moe_2stages
     FLAT 1stage (tuned flat=1/2)           no         kernel needs 8 spare bytes past the rows
     adaptive-aux sort (a4w4)               no         parameter not threaded yet
     grouped a8w4 (gfx1250)                 no         callee has no out param
  
   Not in-place => _return_output copies, so the contract holds either way;
   only the saving is lost.
```
please check whether this is right.

File: aiter/fused_moe.py [hunk 4]
```diff
@@ -172,6 +172,41 @@ def _adaptive_moe_sort(
...
 
 
+def _validate_output_buffer(output, shape, dtype, device, hidden_states=None):
```
yzhou103: _validate_output_buffer's 5th parameter is an invisible switch: the fake passes 4 args (no alias check), _fused_moe_impl passes 5 (alias check). Both have hidden_states in hand and the asymmetry isn't explained; device is also redundant with it. Either always pass it — which would catch aliasing at trace time, strictly better — or split into two functions whose names say which check they are

aiter/fused_moe.py
```diff
@@ -372,8 +421,10 @@ def moe_sorting(
             expert_mask,
             num_local_tokens,
             accumulate=accumulate,
+            output=output,
         )
     # FLAT kernel: in-kernel routing (manifest flat=1); pass through unsorted topk.
+    # `output` cannot be written in place: FLAT needs extra bytes appended to moe_buf.
     if flat:
```
yzhou103, line 442: add assert output is None, "FLAT over-allocates moe_buf; the caller copies"

yzhou103 (approved):
LGTM

PR #4620, opened 2026-08-07
[HIP] [CK] [MoE] Added Gelu with tanh approx for CK XDL 2-stage MoE

a-sidorova:
## Motivation

Enable the tanh-approximation GELU activation (`gelu_tanh`, `0.5 * x * (1 + tanh(sqrt(2/pi) * (x + 0.044715 * x^3))))`) in the AITER Composable Kernel XDL 2-stage MoE path. The MoE gridwise epilogue supports only `silu/gelu/swiglu`; this ports CK's `gelu_tanh_and_mul` (ROCm/rocm-libraries#9396) into AITER so models whose MoE experts use the GELU tanh approximation (e.g. Gemma-family MoE) can run on this path.

**JIRA ID : ROCM-27619**

## Technical Details

Wires a new `GeluTanh` activation end-to-end into the CK 2-stage MoE codegen, following the existing `swiglu_oai` port (`78e45124`):

- `csrc/include/aiter_enum.h`: add `ActivationType::GeluTanh = 4`.
- `csrc/include/rocm_ops.hpp`: expose `GeluTanh` on the pybind `ActivationType` enum.
- `csrc/ck_gemm_moe_2stages_codegen/gemm_moe_ck2stages.cu`: map `ActivationType::GeluTanh -> 4` (CK `gelu_tanh_and_mul`) in `map_activation_to_ck_stage1`.
- `gemm_moe_ck2stages_common.py`: add `"gelutanh": 4` to `ACT_OP_MAP` (drives CK `ActOP` codegen and kernel-instance naming).
- `gen_instances.py`: add `gelutanh` to the `-act` choices and a targeted plain-f8 AOT prebuild loop (mirrors the swiglu loop; not required for correctness since runtime JIT generates on demand).

CK submodule bump

- Bumps `3rdparty/composable_kernel` to the merged ROCm/rocm-libraries#9396, which provides the CK-side gelu with tanh approx epilogue. This PR must be built / merged together with that CK commit.

Test/reference support so the change can be validated:

- `aiter/ops/quant.py` (`get_torch_act`): add the `GeluTanh` reference using the tanh-approx GELU (`F.gelu(x, approximate="tanh")`) to match CK's `FastGelu` epilogue.
- `aiter/utility/dtypes.py` (`str2ActivationType`): case-insensitive enum lookup so mixed-case members like `GeluTanh` resolve from CLI (`-a gelutanh`).

The activation is applied in fp32 in the stage-1 epilogue; GEMM compute and quantization are untouched. CK-side support comes from ROCm/rocm-libraries#9396 (`gelu_tanh_and_mul = 4`).

## Test Plan

Run the CK 2-stage MoE test with the new activation on the plain-f8 (a8w8) path:

```bash
python3 op_tests/test_moe_2stage.py -a gelutanh -q 0  # bf16
python3 op_tests/test_moe_2stage.py -a gelutanh -q 1 --no-flydsl-csv   # per_tensor fp8
python3 op_tests/test_moe_2stage.py -a gelutanh -q 2 --no-flydsl-csv   # per_token  fp8

### DIFF
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -2388,11 +2388,24 @@ def get_2stage_cfgs(
                     f"[fused_moe] Opus stage2 config unsupported ({opus_reason}); "
                     "using default heuristics"
                 )
+
     bypass_tuned_config = int(os.environ.get("AITER_BYPASS_TUNE_CONFIG", "0"))
     if config_file is not None and (cfg is None or bypass_tuned_config):
         raise NotImplementedError(
             "The dedicated FHMoE path requires an exact tuned config row for "
             f"{keys} in {tune_file}"
+        )
+
+    # The asm 1-stage kernels are compiled only for Silu/Gelu
+    if (
+        cfg is not None
+        and cfg.get("run_1stage", False)
+        and activation not in (ActivationType.Silu, ActivationType.Gelu)
+    ):
+        cfg = None
+        logger.warning(
+            f"[fused_moe] discarding 1-stage tuned config for unsupported "
+            f"activation {activation}; using default heuristics"
         )
 
     use_non_temporal_load = False
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -81,6 +81,33 @@ _MOE_A8W4_BYPASS_QUANT = os.environ.get("AITER_MOE_A8W4_BYPASS_QUANT", "0") == "
 # so there is no overhead.
 kernel_bench_callable = None
 
+# CK 2-stage kernel-instance names embed the activation token (see
+# gemm_moe_ck2stages_common.ACT_OP_NAME). Used to detect when the act-disabled
+# tuned-config fallback tier handed us another activation's CK kernel name.
+_CK2STAGES_ACT_TOKENS = {
+    int(_member): _token
+    for _name, _token in (
+        ("Silu", "silu"),
+        ("Gelu", "gelu"),
+        ("Swiglu", "swiglu"),
+        ("GeluTanh", "gelutanh"),
+    )
+    if (_member := getattr(ActivationType, _name, None)) is not None
+}
+
+
+def _ck2stages_kname_act_token(kname: str) -> str | None:
+    """Return the activation token embedded in a CK 2-stage kernel name, if any.
+
+    Order matters: check ``gelutanh`` before ``gelu`` so the shorter token does
+    not shadow the longer one.
+    """
+    for tok in ("gelutanh", "swiglu", "silu", "gelu"):
+        if f"_{tok}_" in kname:
+            return tok
+    return None
+
+
 # FLAT 1stage asm kernels (manifest flat=1) ingest raw topk_ids /
 # topk_weights through the sorted_* kernarg slots and accumulate via
 # global_atomic_pk_add_bf16, so moe_sorting is a pass-through for them.
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -2312,9 +2312,7 @@ def get_2stage_cfgs(
         if not c2s:
             return None
         primary, fallback = c2s
-        lookup_keys = keys
-        if config_file is not None:
-            lookup_keys = keys[:7] + (str(activation),) + keys[8:]
+        lookup_keys = keys[:7] + (str(activation),) + keys[8:]
         result = primary.get(lookup_keys, None)
         if result is None and config_file is None:
             result = fallback.get(keys_disabled, None)
@@ -2323,7 +2321,7 @@ def get_2stage_cfgs(
             tier_idx = _PADDED_M_TIERS.index(token) if token in _PADDED_M_TIERS else -1
             for fallback_tier in reversed(_PADDED_M_TIERS[:tier_idx]):
                 # keys layout: (gfx, cu_num, token, ...); replace token (idx 2).
-                keys_fb = keys[:2] + (fallback_tier,) + keys[3:]
+                keys_fb = lookup_keys[:2] + (fallback_tier,) + lookup_keys[3:]
                 keys_fb_disabled = (
                     keys_disabled[:2] + (fallback_tier,) + keys_disabled[3:]
                 )
--- a/aiter/fused_moe.py
+++ b/aiter/fused_moe.py
@@ -81,32 +81,6 @@ _MOE_A8W4_BYPASS_QUANT = os.environ.get("AITER_MOE_A8W4_BYPASS_QUANT", "0") == "
 # so there is no overhead.
 kernel_bench_callable = None
 
-# CK 2-stage kernel-instance names embed the activation token (see
-# gemm_moe_ck2stages_common.ACT_OP_NAME). Used to detect when the act-disabled
-# tuned-config fallback tier handed us another activation's CK kernel name.
-_CK2STAGES_ACT_TOKENS = {
-    int(_member): _token
-    for _name, _token in (
-        ("Silu", "silu"),
-        ("Gelu", "gelu"),
-        ("Swiglu", "swiglu"),
-        ("GeluTanh", "gelutanh"),
-    )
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: csrc/ck_gemm_moe_2stages_codegen/gemm_moe_ck2stages_common.py [hunk 1]
```diff
@@ -18,8 +18,9 @@
 from chip_info import get_gfx
 
 # CK stage1 activation op values. swiglu (OAI swiglu_oai) maps to CK act value 3;
-# value 2 is intentionally skipped (reserved for a future activation variant).
-ACT_OP_MAP = {"gelu": 0, "silu": 1, "swiglu": 3}
+# gelutanh (gelu_tanh_and_mul) maps to CK act value 4; value 2 is intentionally
+# skipped (reserved for a future activation variant).
+ACT_OP_MAP = {"gelu": 0, "silu": 1, "swiglu": 3, "gelutanh": 4}
```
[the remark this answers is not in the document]
a-sidorova: This can't silently run the wrong activation on the CK 2-stage path: `ck_moe_stage1` [selects](https://github.com/ROCm/aiter/blob/ca68b4f3501762c15c550cb920a5516e9710cf89/aiter/fused_moe.py#L2835) the compiled module by the requested activation, so the loaded module's lookup table only contains that activation's instances (names embed `_gelutanh_`). A fallback Silu/Gelu kernelName1 won't resolve there, so C++ logs `ck kernel not found` and falls through to `moe_stage1_heuristic_dispatch`, which matches on [`ActOP == act_op`](https://github.com/ROCm/aiter/blob/ca68b4f3501762c15c550cb920a5516e9710cf89/csrc/ck_gemm_moe_2stages_codegen/gen_instances.py#L78) and runs the correct GeluTanh epilogue

aiter/utility/dtypes.py
```diff
@@ -163,5 +163,11 @@ def _convert(s):
...
 def str2ActivationType(s):
-    """Convert string to ActivationType."""
+    members = getattr(ActivationType, "__members__", None)
+    if members is not None:
+        s_lower = s.lower()
+        for name, member in members.items():
+            if name.lower() == s_lower:
+                return member
+        raise argparse.ArgumentTypeError(f"invalid activation type: {s}")
     return getattr(ActivationType, s.capitalize())
```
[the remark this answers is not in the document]
a-sidorova: https://github.com/ROCm/aiter/pull/4620/commits/a5cf3482cd21b44733c401b64e0da13abe0e5926

a-sidorova:
Hi @yzhou103! Could you please review the PR?

There are three failed jobs - they seems like sporadic failures since they're not related to my changes. Unfortunately I don't have permissions to relaunch CI jobs. I will be really appreciate for the help with CI!
FYI: I've already tried to rebase to the man branch - a few failed CI jobs are disappeared, but not all of them.

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_ck2stages_common.py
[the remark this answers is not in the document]
a-sidorova: My previous rebuttal only considered the runtime dispatch angle. But the act-disabled fallback bites earlier, at routing/compile time: `get_2stage_cfgs` matched a `_silu_ fp4x2 moe_ck2stages_*` tuned row for a Swiglu request. That non-empty `kernelName1` then skipped the swiglu → CK-Tile guard (which requires not `kernelName1`) and routed swiglu-MXFP4 into the CK 2-stage branch, which JIT-built a swiglu × fp4x2 CK instance — a config CK genuinely doesn't support. 

I added a guard in `get_2stage_cfgs` that, after the tuned lookup, discards a CK `moe_ck2stages_*` config whose embedded activation token doesn't match the requested activation.

Verified on gfx950:

```
[fused_moe] discarding cross-activation CK tuned config (kernelName1 activation 'silu' != requested 'swiglu') for (...Swiglu...fp4x2...per_1x32...); using default routing/heuristics
[fused_moe] ... using heuristic FlyDSL fallback (kn1='flydsl_moe1_afp4_wfp4_bf16_...')
```

a-sidorova:
@zufayu @yzhou103 hello! I've managed to green CI after several rebases. Could you please review this PR? Everything is ready for review

aiter/fused_moe.py
```diff
@@ -77,6 +77,33 @@ def is_flydsl_available():
...
+
+    Order matters: check ``gelutanh`` before ``gelu`` so the shorter token does
+    not shadow the longer one.
+    """
+    for tok in ("gelutanh", "swiglu", "silu", "gelu"):
+        if f"_{tok}_" in kname:
+            return tok
+    return None
+
+
```
yzhou103: I  think this should not be here. The config CSV already has act_type and the primary lookup uses it correctly. The issue is only in the fallback tier (L2209-2212) which blanks act_type — when gelutanh has no tuned rows, it falls back and picks up a silu/gelu row whose CK kernel is compiled for the wrong activation.
Rather than adding a post-hoc kernel-name check in the dispatch layer, could the fallback tier skip CK 2-stage rows whose kernelName1 embeds a different activation token? That keeps the fix inside the config lookup where the problem originates, and avoids duplicating ACT_OP_MAP knowledge in fused_moe.py

aiter/fused_moe.py
yzhou103, line 3906: The split-k post-activation paths (fused_moe.py:3434, 3825, 3967 etc.) use an open else that falls through to gelu_and_mul. Today GeluTanh won't hit split-k, but it would be safer to make the else an explicit elif Gelu + else: raise ValueError, so a future activation mismatch fails loud instead of silently computing the wrong result. seems there're serveral places with this.

simondanielsson:
Thanks for the good work, we're running a similar thing on a custom aiter wheel with a customer in prod for Gemma4 and would be great to see this PR land asap

aiter/fused_moe.py
a-sidorova: It makes sense, I agree. Added exception capture. Thanks

aiter/fused_moe.py
a-sidorova: It's definitely better. Applied, thanks

a-sidorova:
@yzhou103 thank you for the review! Could you please take a look one more time? I implemented your suggestions

aiter/fused_moe.py
simondanielsson: Given these 1-stage kernels don't support gelu_tanh, do we need to make them unreachable whenever using gelu_tanh?

Not sure where the best place is to put the guard, but one option to add it in these two places:

1. 
https://github.com/a-sidorova/aiter/blob/79cb209cbf021217e987c79ece9ebb4940c0b2a1/aiter/fused_moe.py#L2357-L2365

```python
        if activation != ActivationType.GeluTanh and (
            activation,
            q_type,
            dtype,
            q_dtype_a,
            q_dtype_w,
            use_g1u1,
            doweight_stage1,
        ) in fused_moe_1stage_dict.get(get_gfx(), {}):
```

and 

2. https://github.com/a-sidorova/aiter/blob/79cb209cbf021217e987c79ece9ebb4940c0b2a1/aiter/fused_moe.py#L2409

```python
run_1stage = activation != ActivationType.GeluTanh and cfg.get("run_1stage", False)
```


There's likely as cleaner way to do this

a-sidorova: @simondanielsson thanks for the idea. I think it makes sense. Additionally I noticed that 1stage kernels supports only silu and gelu, so I decided to just add one common guard before the 1stage logic in https://github.com/ROCm/aiter/pull/4620/changes/7c20fe3a2e6ea122e68eee3c1956689e69cba824

a-sidorova:
@yzhou103 hello! The CI is green, your comments are applied. Could you please take a look again? Thank you in advance

yzhou103 (approved):
LGTM

a-sidorova:
Waiting for additional approve for merging with bumping CK

valarLip (approved):

PR #4621, opened 2026-08-07
fix: car graph mode err when pytorch set expandable_segments:True

### LINKED ISSUE #4174: [Bug] custom_all_reduce.cu:397 hipIpcGetMemHandle fails when caller uses PYTORCH_HIP_ALLOC_CONF=expandable_segments:True

TennyWang1223:
## Motivation

Fixes #4174. Under PyTorch `expandable_segments:True`, `torch.empty` pointers live in a hipMallocAsync pool and cannot be exported via `hipIpcGetMemHandle`, aborting CustomAllreduce init.

Unlike #4560 (disable CustomAllreduce → fall back to RCCL, a large perf loss), this keeps CustomAllreduce working: the IPC input pool is allocated with a plain `hipMalloc`, which stays IPC-exportable regardless of the allocator config.

## Technical Details

- New `allocate_data_buffer` (plain `hipMalloc`) + `raw_cached` IPCBuffer mode; the `input` pool now uses it instead of `torch.empty`.
- Detect `expandable_segments` from the PyTorch allocator snapshot (env fallback: `PYTORCH_CUDA_ALLOC_CONF` → `PYTORCH_HIP_ALLOC_CONF` → `PYTORCH_ALLOC_CONF`).
- When enabled, force the copy-in path during CUDA graph capture (`enable_register_for_capturing=False`), staging input into the hipMalloc pool instead of binding the input tensor's own pointer. VMM transport is unaffected.

## Test Plan

- Multi-GPU regression on the `op_tests/multigpu_tests` suite (allreduce, fp8, reduce_scatter, allgather, fused_ar_rms).
- End-to-end `expandable_segments:True` graph-capture validation on a platform that supports it (e.g. MI355X).
- Ruff / Black / `py_compile` / `git diff --check` on changed files.

## Test Result

- _(fill in after runs)_ multigpu regression: TODO
- _(fill in)_ expandable_segments graph capture: TODO
- Ruff, Black, py_compile: passed.

## Submission Checklist

### DIFF
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -795,9 +795,9 @@ class CustomAllreduce:
 
         self.group = group
 
-        assert dist.get_backend(group) != dist.Backend.NCCL, (
-            "CustomAllreduce should be attached to a non-NCCL group."
-        )
+        assert (
+            dist.get_backend(group) != dist.Backend.NCCL
+        ), "CustomAllreduce should be attached to a non-NCCL group."
 
         if not all(in_the_same_node_as(group, source_rank=0)):
             # No need to initialize custom allreduce for multi-node case.
@@ -1051,10 +1051,16 @@ class CustomAllreduce:
         # Create IPC buffer pool and allocate all named buffers.
         self._pool = IPCBufferPool(self.device, self.group, **pool_kwargs)
         self._pool.create("meta", meta_size, uncached=True)
-        # Cached raw hipMalloc (not torch.empty): the input pool is IPC-exported
-        # via hipIpcGetMemHandle, which fails on PyTorch expandable-segment pool
-        # pointers (issue #4174). A plain hipMalloc allocation stays exportable.
-        self._pool.create("input", max_size, raw_cached=True)
+        # The input pool is IPC-exported via hipIpcGetMemHandle. That fails on
+        # PyTorch expandable-segment pool pointers (issue #4174), so when
+        # expandable_segments is on we back the pool with a plain (cached)
+        # hipMalloc allocation, which stays exportable. Otherwise keep the
+        # default torch.empty allocation so the pool remains tracked by the
+        # PyTorch caching allocator (its memory accounting is what downstream
+        # consumers profile against). Gated on the same condition as the
+        # capture copy-in path above; _init_ipc only runs for non-VMM.
+        raw_cached = _expandable_segments_enabled()
+        self._pool.create("input", max_size, raw_cached=raw_cached)
 
         handles, offsets = self._pool.get_ipc_meta("meta")
         self._ptr = self._ops_init_custom_ar(
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -713,7 +713,6 @@ class _GFX1250BufferProxy:
 
 
 class CustomAllreduce:
-
     _SUPPORTED_WORLD_SIZES: ClassVar[list[Any]] = [2, 4, 6, 8]
 
     def _select_ops(self):
@@ -796,9 +795,9 @@ class CustomAllreduce:
 
         self.group = group
 
-        assert (
-            dist.get_backend(group) != dist.Backend.NCCL
-        ), "CustomAllreduce should be attached to a non-NCCL group."
+        assert dist.get_backend(group) != dist.Backend.NCCL, (
+            "CustomAllreduce should be attached to a non-NCCL group."
+        )
 
         if not all(in_the_same_node_as(group, source_rank=0)):
             # No need to initialize custom allreduce for multi-node case.
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -54,6 +54,56 @@ def _detect_gfx1250() -> bool:
         return False
 
 
+# PyTorch allocator-config env vars that may carry expandable_segments, in
+# PyTorch's own precedence order (the first one that is set wins; they are not
+# merged). Used only as a fallback when the allocator snapshot is unavailable.
+_ALLOC_CONF_ENV_VARS = (
+    "PYTORCH_CUDA_ALLOC_CONF",
+    "PYTORCH_HIP_ALLOC_CONF",
+    "PYTORCH_ALLOC_CONF",
+)
+
+
+def _parse_expandable_segments(conf: str) -> bool:
+    """Extract ``expandable_segments:<bool>`` from a PyTorch alloc-conf string."""
+    for field in conf.split(","):
+        key, _, value = field.partition(":")
+        if key.strip() == "expandable_segments":
+            return value.strip().lower() in ("1", "true", "yes", "on")
+    return False
+
+
+def _expandable_segments_enabled() -> bool:
+    """Whether PyTorch's caching allocator has expandable segments active.
+
+    Under expandable_segments the allocator hands out virtual-memory-managed
+    pointers that cannot be exported through hipIpcGetMemHandle (issue #4174).
+    The classic IPC graph-capture path binds the input tensor's own pointer as
+    an IPC handle, so it would fail; detecting this lets us force the copy-in
+    path instead, which stages into the plain-hipMalloc input pool.
+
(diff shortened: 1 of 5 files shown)


jamesETsmith:
I tested PR 4621 at commit `c07608998b35c0dbbd533c9880ef9c347db72c12` on two MI355X with `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True`.

Both ranks confirmed that:

- the allocator snapshot showed expandable segments were enabled;
- CustomAllreduce stayed enabled;
- the input pool used the PR's new raw cached allocation;
- graph-buffer registration was turned off so the copy-in path was used;
- eager CustomAllreduce returned the expected sum of `3.0`;
- CUDA graph replay also returned `3.0`.

## How I checked the CustomAllreduce path

I checked that `custom_allreduce.disabled` was false, then called `custom_allreduce.custom_all_reduce` directly in eager mode and during CUDA graph capture. Both calls returned real outputs with the expected sum, so they did not fall back to RCCL.

To check the new input pool, my test verified that `custom_allreduce._pool["input"]._raw_cached` was true and `custom_allreduce._pool["input"]._buffer` was `None`. In this PR, that combination means the pool was created with `raw_cached=True`. That selects `allocate_data_buffer`, which uses plain `hipMalloc`, instead of creating a `torch.empty` tensor through PyTorch's expandable-segment allocator.

I also checked that `enable_register_for_capturing` was false. This means graph execution copied through the new `hipMalloc` input pool instead of trying to export the graph tensor's VMM-backed pointer.

zufayu:
Thanks @jamesETsmith, that's the validation this PR was missing. `_pool["input"]._raw_cached` being true is decisive on its own, and confirming `enable_register_for_capturing` was false shows the new gate actually fired rather than the path being reached by accident.

It settles that the fix works under the configuration it targets. It can't settle the other half: the run had expandable segments on, and `_init_ipc` passes `raw_cached=True` unconditionally, so an unconditional switch and one gated on `_expandable_segments_enabled()` would look identical in that test. The default configuration, which is what every existing user runs, is still unexercised.

@TennyWang1223 the gate two lines above uses `_expandable_segments_enabled()` to decide `enable_register_for_capturing`, but the allocation switch below it is unconditional. Why the asymmetry? If the detection is reliable enough to change capture behaviour, it seems reliable enough to change the allocation, and gating it would keep every user who isn't running expandable segments on the caching allocator. If the pool has to be `hipMalloc` in all cases, please say why in the description, and confirm what happens to the memory that at the default `max_size` is 1 GiB no longer tracked by the PyTorch allocator: which accounting do the downstream consumers profile against when they size their caches, and does moving this pool out of it change their result?

Test Result is also still TODO for `op_tests/multigpu_tests`, which the run above doesn't cover.

None of the red checks here are yours, so don't spend time on them. Both OPUS jobs fail on `test_opus_basic: libamdhip64.so.7: cannot open shared object file`, which is the rpath fix from #4608 missing: OPUS Test checks out the bare head SHA and this branch was cut before that commit landed. The three Standard Tests shards fail in `test_aiter_add.py` / `test_aiter_addInp.py` on a kineto `UnicodeDecodeError` that turns up intermittently on unrelated PRs. Please do rebase and add `ci:all` regardless. These results are from Aug 7 on a branch 36 commits behind main with every downstream job skipped, which isn't enough for a change to CustomAllreduce.

zufayu (approved):

PR #4637, opened 2026-08-07
fix(quant): use saturating RNE for scaled int8 casts

skyguan92:
## Summary

- make the shared scaled FP32-to-INT8 x2/x4 helpers use round-to-nearest-even (RNE) followed by standard `fminf`/`fmaxf` symmetric saturation to `[-127, 127]`;
- give the OPUS RMSNorm INT8 cast the same finite-value contract;
- align the INT8-only Torch path, direct Triton quant kernels, Triton RMSNorm/LayerNorm output helpers, and independent in-tree PyTorch test oracles with that contract, while leaving FP8/BF8 behavior unchanged;
- add a deterministic cross-backend static boundary regression, in addition to the grouped fused-RMSNorm regression.

This closes the pre-existing split between AITER's shared HIP and OPUS RMSNorm INT8 paths without replacing it with a HIP-vs-reference split. It also prevents static/smooth quantization values such as `127.6` from wrapping to `-128`.

## Review follow-up: controlled W7900 comparison

I reproduced both issues reported in the first review with the same seed, BF16 input, shape `128x4096`, and seven hot timing samples on native `gfx1100`:

| implementation | HIP vs Torch | HIP vs Triton | HIP dequant MAE | 26-value finite boundary |
|---|---:|---:|---:|---:|
| base truncation | 27 / 524,288 (0.00515%) | 11 / 524,288 (0.00210%) | 0.0149471 | 8 differ from RNE+saturate |
| first PR revision: RNE, no clamp | 260,095 / 524,288 (49.6092%) | 260,121 / 524,288 (49.6141%) | 0.0075071 | 12 differ; `127.5/127.6` can flip sign |
| RNE + symmetric saturation | 143 / 524,288 (0.02728%) | 39 / 524,288 (0.00744%) | 0.0075071 | **0 differ across HIP/Torch/Triton** |

All remaining dynamic mismatches have maximum delta 1. Torch, Triton, and HIP have the same dequant MAE (`0.0075070988`) in the revised run; the small value mismatches come from scale/reciprocal evaluation differences rather than different finite integer-rounding rules.

The latest review suggested replacing gfx-specific `v_med3_f32` inline assembly with `fminf(fmaxf(__builtin_rintf(x), -127.f), 127.f)` because NaN results can differ. Commit `2277d85f` applies that standard expression to both in-scope device helpers.

A direct special-value probe on W7900/gfx1100 found that the old `v_med3` and new min/max expressions both map positive/negative quiet and signalling NaNs to `-127`; infinities saturate to `+/-127`, and all finite ties and boundaries are unchanged. A separate BF16 backend probe found that Torch currently maps NaN to `0`, while HIP and Triton map it to `-127`. Therefore this PR preserves and tests the finite RNE+saturation contract but does **not** claim to establish a global cross-backend NaN policy.

## W7900 correctness gates

Environment: Radeon PRO W7900, native `gfx1100`, process-level `HSA_OVERRIDE_GFX_VERSION` explicitly unset, PyTorch `2.9.1+gitff65f5b`, HIP `7.2.53211-e1a6bc5663`, Triton `3.5.1`.

After the final source adjustment, all candidate-owned JIT modules were moved into named recoverable stale-cache directories and rebuilt from scratch on exact public head `2277d85f`:

- `test_quant.py -q i8_token -m 128 -n 4096 -d bf16`: exit 0; HIP/Torch differs in 91 / 524,288 values and Triton/Torch in 86 / 524,288, all by 1;
- explicit OPUS RMSNorm boundary fixture: exact expected RNE+saturation output, exit 0;
- `test_rmsnorm2dFusedAddQuant.py --mode 5 -q i8 -m 1 8 16 128 -n 1024 4096 -d bf16`: exit 0; all eight HIP and CK output/scale checks pass;
- `test_kvcache.py -t bf16toi8 -b 8 -c 128`: exit 0; remaining cache differences are one INT8 level rather than the approximately 50% semantic split;
- real generated quant and OPUS RMSNorm translation units rebuild from zero object/shared-library outputs for `gfx942` and `gfx950`, four builds at exit 0. These are compile gates, not Instinct runtime claims; upstream accelerator CI remains authoritative.

## CI follow-up: direct Triton and normalization coverage

After workflow approval, public [Triton run 31598318902](https://github.com/ROCm/aiter/actions/runs/31598318902) exposed two candidate-owned gaps in head `2277d85f`:

- MI35X shard 4 had 67 INT8-only direct-quant failures because the local PyTorch oracles in `op_tests/triton_tests/quant/test_quant.py` still used truncating `.to(int8)` while the candidate Triton kernels used RNE+saturation;
- MI35X shard 5 had 43 BF16 RMSNorm quantization failures, all with maximum delta 2, because the Triton normalization helpers still relied on the output store's truncating cast.

Commit `971cece9` fixes those gaps without changing FP8 behavior: the test-local oracle now explicitly uses `torch.round(...).clamp(-127, 127)` only for INT8, and both Triton RMSNorm and LayerNorm reuse the existing `_int8_rne_sat` primitive. The single shard-7 MXFP4 mismatch and the OPUS jobs' missing `libamdhip64.so.7` loader error are outside the changed INT8 paths.

On native W7900/gfx1100, the old source deterministically reproduced 388/1,024 direct-quant mismatches and 3/262,144 fused-RMSNorm mismatches. The same two nodes pass on physical GPU0 and GPU7 after the fix. On exact current AITER main `a50a62aa` (merge tree `d326412f`), the GitHub-run wheelhouse—Triton `3.7.0+amd.rocm7.2.0.git89002410` and triton-kernels `1.0.0+amd.rocm7.2.0.git89002410`—passes:

- direct INT8 quantization: 69/69;
- RMSNorm dynamic/smooth INT8 quantization: 216/216;
- LayerNorm dynamic/smooth INT8 quantization: 96/96.

The downloaded GitHub artifact and the wheelhouse independently downloaded by AITER's pinned-index contract have identical SHA-256 values (`45710096...7504` and `452b7a3d...e1d5`). Both test containers exited 0, were not OOM-killed, and ran without `HSA_OVERRIDE_GFX_VERSION`. New public workflows for `971cece9` are awaiting repository approval; they are not represented as green yet.

## Performance non-regression

An eight-round interleaved A/B/B/A comparison used the same BF16 `128x4096` input, 100 warmups, 500 launches per sample and seven samples per round. Taking the median of four round-p50 values per implementation:

| implementation | median round p50 |
|---|---:|
| old `v_med3` | `20.7393 us` |
| standard `fminf`/`fmaxf` | `20.6306 us` |
| relative | `-0.52%` |

This difference is within run noise and supports only a no-observed-regression conclusion. The extracted gfx1100 special-value probe shows the compiler lowering the standard min/max expression to `v_maxmin_f32`; the old comparison branch retains `v_med3_f32`.

## DeepSeek V4 relevance

The official DeepSeek V4 Flash paper, Hugging Face/ModelScope card and config, ModelScope `deepseek-ai/DeepSeek-V4-Flash-0731` checkpoint and bundled `inference/`, and official TileKernels define the model shapes and dynamic FP8/FP4 route. They do not define this optional INT8 NaN policy and are algorithmic/model references rather than executable INT8 semantics on W7900. This PR hardens AITER's optional INT8/W8A8 foundation; it does not claim to alter the official V4 numerical path.

For the earlier layer-local screen, official 0731 layer-0 weights were used with calibrated group-64 INT8 weights and the same W8A8 GEMM on both sides. The fused/current hot p50 geomean was `1.289x` to `1.313x` across local `wqkv_a`, TP8-rank-0 `wq_b`, and shared gate/up shapes; all nine quality rows met cosine `>= 0.9995` and relative L2 `<= 0.01`. The dynamic scale bounds values to `|q| <= 127`; the finite saturation change does not alter that path's output or MAE.

## Scope boundary

This PR normalizes the shared scaled quantization path, its Torch/Triton references, and the directly related OPUS RMSNorm cast. It does **not** claim that every FP32-to-INT8 conversion in AITER now has one global policy. Generic `opus::cast<i8_t>`, third-party CK smoothquant/MoE conversions, and separate Triton MoE half-away-from-zero conversions remain pre-existing semantics debt. Pulling those independent/vendor paths into this correctness fix would broaden both ownership and validation requirements, so they are intentionally left for separately justified work.

## Source state and checks

The public head consists of four signed-off commits and remains on its older ancestor; the update is a normal fast-forward from `2277d85f` to `971cece9`, with no rebase or force-push. It merges without conflict into current AITER main `a50a62aaa292427e00190dd7ba58e39a6db4e61e` as exact tree `d326412ff06a0a31cecf3a8d703fb0c53b39fb45`.

- Black `25.11.0`: pass on the three-file follow-up (the repository's stable-Black workflow rerun is pending approval);
- Ruff `0.16.0` (the CI-pinned version): pass on the three-file follow-up;
- changed-line clang-format checks: pass on the final C++ revision;
- Python byte-compile and `git diff --check`: pass;
- DCO sign-offs: present on all four commits.

### DIFF
--- a/csrc/include/opus/rmsnorm_opus_kernel.hpp
+++ b/csrc/include/opus/rmsnorm_opus_kernel.hpp
@@ -74,12 +74,19 @@ __global__ void rmsnorm_quant_opus(
 {
 }
 #else
-// fp32 -> quant element. int8: round-to-nearest; fp8: hardware e4m3 cvt.
+// fp32 -> quant element. int8: round-to-nearest-even with symmetric saturation;
+// fp8: hardware e4m3 cvt.
 template <typename out_t>
 __device__ inline out_t quant_cast(float v)
 {
     if constexpr(std::is_same_v<out_t, signed char>)
-        return static_cast<signed char>(__builtin_rintf(v));
+    {
+        v                  = __builtin_rintf(v);
+        constexpr float lo = -127.0f;
+        constexpr float hi = 127.0f;
+        asm volatile("v_med3_f32 %0, %0, %1, %2" : "+v"(v) : "v"(lo), "v"(hi));
+        return static_cast<signed char>(v);
+    }
     else
         return opus::fp32_to_fp8(v);
 }
--- a/csrc/include/opus/rmsnorm_opus_kernel.hpp
+++ b/csrc/include/opus/rmsnorm_opus_kernel.hpp
@@ -81,11 +81,7 @@ __device__ inline out_t quant_cast(float v)
 {
     if constexpr(std::is_same_v<out_t, signed char>)
     {
-        v                  = __builtin_rintf(v);
-        constexpr float lo = -127.0f;
-        constexpr float hi = 127.0f;
-        asm volatile("v_med3_f32 %0, %0, %1, %2" : "+v"(v) : "v"(lo), "v"(hi));
-        return static_cast<signed char>(v);
+        return static_cast<signed char>(fminf(fmaxf(__builtin_rintf(v), -127.0f), 127.0f));
     }
     else
         return opus::fp32_to_fp8(v);
--- a/aiter/ops/quant.py
+++ b/aiter/ops/quant.py
@@ -61,6 +61,14 @@ def get_dtype_max(dtype):
     return dtypeMax
 
 
+def _scaled_quant_cast(x, quant_dtype):
+    # INT8 scales use symmetric qmax=127; leave floating-point formats unchanged.
+    if quant_dtype == dtypes.i8:
+        dtype_max = get_dtype_max(quant_dtype)
+        x = torch.round(x).clamp(-dtype_max, dtype_max)
+    return x.to(dtype=quant_dtype)
+
+
 def pertoken_quant(
     x,
     scale=None,
@@ -89,7 +97,7 @@ def pertoken_quant(
         per_token_scale[per_token_scale == 0] = 1
 
     # quant hidden_states
-    y = (hidden_states / per_token_scale).to(dtype=quant_dtype)
+    y = _scaled_quant_cast(hidden_states / per_token_scale, quant_dtype)
     y_scale = per_token_scale.to(scale_dtype)
     return y, y_scale
 
@@ -319,13 +327,13 @@ def per_tensor_quant(
     x, scale=None, scale_dtype=dtypes.fp32, quant_dtype=dtypes.i8, dtypeMax=None
 ):
     x = x.to(dtypes.fp32)
+    if dtypeMax is None:
+        dtypeMax = get_dtype_max(quant_dtype)
     if scale is None:
-        if dtypeMax is None:
-            dtypeMax = get_dtype_max(quant_dtype)
         scale = torch.abs(x).max() / dtypeMax
     y = x / scale
 
-    return y.to(quant_dtype), scale.view(1).to(scale_dtype)
+    return _scaled_quant_cast(y, quant_dtype), scale.view(1).to(scale_dtype)
 
 
 def per_block_quant_wrapper(block_shape=(1, 128)):
--- a/aiter/ops/triton/_triton_kernels/normalization/norm.py
+++ b/aiter/ops/triton/_triton_kernels/normalization/norm.py
@@ -4,6 +4,8 @@
 import triton
 import triton.language as tl
 
+from aiter.ops.triton._triton_kernels.quant.quant import _int8_rne_sat
+
 
 @triton.jit
 def _per_token_quant(
@@ -21,6 +23,9 @@ def _per_token_quant(
     scale_recip = 1 / scale_out
 
     qx = x * scale_recip
+
+    if DTYPE_MAX == 127:
+        qx = _int8_rne_sat(qx)
 
     return qx, scale_out
 
--- a/aiter/ops/triton/_triton_kernels/normalization/rmsnorm.py
+++ b/aiter/ops/triton/_triton_kernels/normalization/rmsnorm.py
@@ -4,6 +4,8 @@
 import triton
 import triton.language as tl
 
+from aiter.ops.triton._triton_kernels.quant.quant import _int8_rne_sat
+
 
 @triton.jit
 def _per_token_quant(
@@ -32,7 +34,9 @@ def _per_token_quant(
 
     qx = x * scale_recip
 
-    if CLAMP_OUT:
+    if DTYPE_MAX == 127:
+        qx = _int8_rne_sat(qx)
+    elif CLAMP_OUT:
         qx = tl.clamp(qx, -DTYPE_MAX, DTYPE_MAX)
 
     tl.store(y_scale_ptr + row_idx, scale_out.to(y_scale_ptr.dtype.element_ty))
--- a/aiter/ops/triton/_triton_kernels/quant/quant.py
+++ b/aiter/ops/triton/_triton_kernels/quant/quant.py
@@ -3,6 +3,12 @@
 
 import triton
 import triton.language as tl
+from triton.language.extra import libdevice
+
+
+@triton.jit
+def _int8_rne_sat(x):
+    return tl.clamp(libdevice.nearbyint(x), min=-127.0, max=127.0)
 
 
 @triton.jit
@@ -25,7 +31,10 @@ def _static_per_tensor_quant_fp8_i8_kernel(
     scale = tl.load(scale_in_ptr)
     scale_recip = 1 / scale
 
-    qx = (x * scale_recip).to(qx_ptr.dtype.element_ty)
+    qx = x * scale_recip
+    if qx_ptr.dtype.element_ty == tl.int8:
+        qx = _int8_rne_sat(qx)
+    qx = qx.to(qx_ptr.dtype.element_ty)
 
     tl.store(qx_ptr + offs, qx, mask=mask)
 
@@ -74,6 +83,8 @@ def _dynamic_per_token_quant_fp8_i8_kernel(
     scale_recip = 1 / scale_out
 
     qx = x * scale_recip
+    if qx_ptr.dtype.element_ty == tl.int8:
+        qx = _int8_rne_sat(qx)
     qx = qx.to(qx_ptr.dtype.element_ty)
 
     scale_offs = pid
--- a/csrc/include/aiter_opus_plus.h
+++ b/csrc/include/aiter_opus_plus.h
@@ -110,14 +110,24 @@ OPUS_D decltype(auto) fp32_to_bf8_scaled_x4(const S& s, float inverted_scale)
     return bf8x4_t{lo[0], lo[1], hi[0], hi[1]};
 }
 
+// fp32 -> i8 with round-to-nearest-even and symmetric saturation.
+// Keep the signed range symmetric with the qmax=127 scale contract.
+OPUS_D i8_t fp32_to_i8_rne_sat(float x)
+{
+    x                  = __builtin_rintf(x);
+    constexpr float lo = -127.0f;
+    constexpr float hi = 127.0f;
+    asm volatile("v_med3_f32 %0, %0, %1, %2" : "+v"(x) : "v"(lo), "v"(hi));
+    return static_cast<i8_t>(x);
+}
+
 // fp32x2 -> i8x2 with scale
-// ISA: v_pk_mul_f32 + round-to-nearest-even conversion x2
+// ISA: v_pk_mul_f32 + v_rndne_f32 x2 + v_med3_f32 x2 + v_cvt_i32_f32 x2
 template <typename S, std::enable_if_t<std::is_same_v<S, fp32x2_t>, bool> = true>
 OPUS_D decltype(auto) fp32_to_i8_scaled_x2(const S& s, float inverted_scale)
 {
     fp32x2_t tmp = pk_mul_f32(s, fp32x2_t{inverted_scale, inverted_scale});
-    return i8x2_t{static_cast<i8_t>(__builtin_rintf(tmp[0])),
-                  static_cast<i8_t>(__builtin_rintf(tmp[1]))};
+    return i8x2_t{fp32_to_i8_rne_sat(tmp[0]), fp32_to_i8_rne_sat(tmp[1])};
 }
 
 template <typename S, std::enable_if_t<std::is_same_v<S, fp32x4_t>, bool> = true>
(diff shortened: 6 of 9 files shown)


yzhou103:
Thanks for the fix - verified locally on gfx950 (ROCm 7.2.4): the new test FAILs on the
  base header and PASSes with the fix, ISA cost is +1 `v_rndne_f32` per element, mean abs
  dequant error drops 0.014894 -> 0.007484, and there's no measurable perf change on these
  bandwidth-bound kernels.

  Worth calling out in the PR description: this also removes a *pre-existing* internal split.
  On base, aiter's two RMSNorm int8 kernels already disagree:

      aiter.rmsnorm_quant                    [2, 2, 2, 3, -2, -2, 6, 6]   truncate (scaled_cast)
      rmsnorm2d_fwd_with_dynamicquant_opus   [2, 2, 3, 4, -2, -3, 6, 7]   RNE      (quant_cast)

  `rmsnorm_opus_kernel.hpp:82` has been RNE all along, so this PR aligns the two. IMO a
  stronger argument than "match torch quantization".

  Two concerns before merge.

  ### 1. The references still truncate, so this introduces a new divergence

  `fp32_to_i8_scaled_x2/x4` is reached from every int8 quant kernel via
  `store_vector` -> `scaled_cast<i8_t>`, but neither reference was updated:

  - `aiter/ops/quant.py:92` `pertoken_quant`: `(hidden_states / per_token_scale).to(quant_dtype)`
  - `aiter/ops/triton/_triton_kernels/quant/quant.py:28,77`: `.to(qx_ptr.dtype.element_ty)`

  Measured on 128x4096 bf16, per-token int8:

  | comparison | base | this PR |
  |---|---|---|
  | hip vs torch ref | 0.0032% | **49.66%** (max delta 1) |
  | triton vs torch ref | 0.0032% | 0.0032% |
  | hip vs triton | 0.0023% | **49.67%** |

  So this isn't pre-existing drift - before the change all backends agreed. Practical impact:
  `test_quant.py -q i8_token` and `test_kvcache.py` go from ~4e-5 to ~0.5 error ratio.
  `checkAllclose` won't raise, so CI stays green, but both will print `failed!` on every run
  and stop being useful int8 regression detectors.

  ### 2. No saturating clamp

  Round-without-clamp is strictly worse than the truncation it replaces at the boundary:

  | input | base (truncate) | this PR (rint, no clamp) | torch `quantize_per_tensor` |
  |---|---|---|---|
  | 127.6 | 127 | **-128** (sign flip) | 127 |
  | 200.0 | -56 | -56 | 127 |

  Not reachable on the dynamic path (scale = amax/127 keeps `|x*inv_scale| <= 127`), but
  static / smooth quant takes a caller-supplied scale, so it is reachable there. The fp8/bf8
  helpers right above already clamp with `v_med3_f32`; int8 doesn't. Cost would be
  +2 `v_med3_f32` (6 VALU per 2 elems vs 4).

  Everything else I checked does both:

  | project | implementation | RNE | saturate |
  |---|---|---|---|
  | torch PT2E `_decomposed.py` | `clamp(round(x * inv_scale), qmin, qmax)` | yes | yes |
  | vLLM (`csrc/libtorch_stable/quantization/w8a8/int8/scaled_quant.cu`) | CUDA `cvt.rni.sat.s8.f32`; ROCm `nearbyint` + clamp | yes | yes |
  | FlashInfer / TRT-LLM | `cvt.rni.sat.s8.f32` | yes | yes |
  | **aiter (after this PR)** | `rintf` + `static_cast` | yes | **no** |

  Note aiter already has a correct one in-tree - `aiter/ops/triton/moe/quant_moe.py:351`:

  ```python
  w_int8 = w_scaled.round().clamp(-127, 127).to(torch.int8)   # == the PT2E formula
 ```
  That seems like the natural target semantics for everything else.



  Also I grepped the tree; there are 11 fp32->int8 sites and this PR touches 1:

  | location | rounding | clamp | note |
|---|---|---|---|
| `csrc/include/opus/opus.hpp:1251` `fp32_to_i8` | **truncate** | no | live `opus::cast<i8_t>` entry - opposite semantics to the one you just fixed, in the same header |
| CK `smoothquant_fwd` / `moe_ck_2stages` (3rdparty `ck_tile/core/numeric/int8.hpp:75`) | **truncate** | no | **live user-facing path**, verified still truncating after this PR |
| `aiter/ops/quant.py:92,328` | truncate | no | torch reference |
| `aiter/ops/triton/_triton_kernels/quant/quant.py:28,77` | truncate | no | triton reference |
| `aiter/ops/triton/_triton_kernels/moe/quant_moe.py:526-533,603-608` | half-away-from-zero | yes `[-127,127]` | third semantics: `2.5` -> 3 here, 2 after this PR |
| `csrc/include/ck_tile/vec_convert.h:161` | truncate | no | dead code (no callers), but a trap |
| `aiter/ops/triton/moe/quant_moe.py:351` | RNE | yes `[-127,127]` | the one that's already right |

After this PR aiter would carry **truncate / RNE / half-away** simultaneously.

skyguan92:
Thanks — this was a very useful review. I reproduced both findings before changing the patch:

- the first PR revision produced 260,095 / 524,288 HIP-vs-Torch mismatches (49.6092%, max delta 1) on the same controlled BF16 `128x4096` dynamic case;
- the static fixture confirmed the sign flip at the positive boundary (`127.5/127.6 -> -128`) and wrapping for larger values.

I pushed [440c262c](https://github.com/ROCm/aiter/commit/440c262cd97fb91515f35165b7eac3bc4ce088a8), which absorbs the two requested changes:

1. the shared HIP x2/x4 conversion now does RNE followed by `v_med3_f32` symmetric saturation to `[-127, 127]`;
2. the OPUS RMSNorm INT8 cast now has the same saturation step;
3. the INT8-only Torch and Triton static/dynamic references use the same RNE+saturation contract (FP8/BF8 are unchanged);
4. `test_quant.py` now has a deterministic boundary fixture that requires exact HIP/Torch/Triton agreement for half ties and out-of-range values.

One implementation detail found during validation: Triton 3.5.1's HIP libdevice does not expose `rint`, so the first local attempt failed compilation. The pushed version uses `libdevice.nearbyint`, which compiled on gfx1100 and produced exact ties-to-even results in the boundary fixture.

I also followed your JIT-cache warning explicitly: before each revised run, I moved both the cached `.so` and its build directory into a recoverable stale-cache directory and rebuilt from scratch.

Controlled W7900 result after the revision:

| check | result |
|---|---:|
| HIP vs Torch | 143 / 524,288 (0.02728%), max delta 1 |
| HIP vs Triton | 39 / 524,288 (0.00744%), max delta 1 |
| dequant MAE, all three backends | 0.0075070988 |
| static boundary | 0 / 26 differences for all three backends |
| HIP kernel p50 | 20.6503 us before clamp, 20.6210 us after clamp |

The final-source rebuild also passes `test_quant.py -q i8_token`, all eight requested BF16/INT8 RMSNorm shape combinations, and the KV-cache test. The selected gfx1100 kernel adds one `v_med3_f32` per element, keeps SGPR at 18, moves VGPR 27 -> 31, and still has zero spills. I additionally compiled the real generated quant and OPUS RMSNorm translation units for gfx942 and gfx950; runtime CI on those accelerators is still the authoritative gate.

I did not fold the generic `opus::cast<i8_t>`, third-party CK conversions, or the independent Triton MoE half-away conversions into this PR. I agree those are real pre-existing semantics debt, but they have different callers/ownership and need their own contract and runtime coverage. The updated PR description now says this explicitly rather than implying that this patch normalizes every INT8 cast in AITER.

csrc/include/opus/rmsnorm_opus_kernel.hpp
```diff
@@ -74,12 +74,19 @@ __global__ void rmsnorm_quant_opus(
...
 template <typename out_t>
 __device__ inline out_t quant_cast(float v)
 {
     if constexpr(std::is_same_v<out_t, signed char>)
-        return static_cast<signed char>(__builtin_rintf(v));
+    {
+        v                  = __builtin_rintf(v);
+        constexpr float lo = -127.0f;
+        constexpr float hi = 127.0f;
+        asm volatile("v_med3_f32 %0, %0, %1, %2" : "+v"(v) : "v"(lo), "v"(hi));
```
yzhou103: maybe better to use fminf(fmaxf(__builtin_rintf(x), -127.f), 127.f). seems there are different output when input is nan.

skyguan92: Thanks — adopted this in 2277d85f for both the shared scaled-INT8 helper and the OPUS RMSNorm path so they cannot diverge. On a W7900/gfx1100, a direct special-value probe gives the same result for the old `v_med3` and the suggested `fminf(fmaxf())` lowering (NaN -> -127; infinities saturate to +/-127), while all finite tie/boundary regressions remain exact. I also rebuilt the real quant and OPUS RMSNorm TUs for gfx942 and gfx950, and ran fresh gfx1100 quant, OPUS RMSNorm, fused RMSNorm, and BF16->INT8 KV-cache tests. An interleaved 8-round microbenchmark showed 20.6306 us vs 20.7393 us median p50 for fmin/fmax vs v_med3 (-0.52%, within run noise). One nuance: Torch currently casts NaN to 0 while HIP/Triton produce -127, so I am treating cross-backend NaN policy as a separate contract question rather than claiming this patch unifies it.

skyguan92:
Follow-up on the first full public CI execution for `2277d85f` ([Triton run 31598318902](https://github.com/ROCm/aiter/actions/runs/31598318902)):

I treated two groups as candidate-owned and reproduced both before changing the branch:

- shard 4: 67 direct INT8 quant failures; the test-local PyTorch oracles still truncated while the candidate Triton kernels used RNE+saturation;
- shard 5: 43 BF16 RMSNorm quant failures, max delta 2; the Triton normalization helper still relied on the store cast's truncation.

[`971cece9`](https://github.com/ROCm/aiter/commit/971cece9dd43885f05b7578ca454cfdccdd82d1b) updates the independent test oracle to explicit RNE+saturation for INT8 only and reuses the existing `_int8_rne_sat` primitive in Triton RMSNorm and LayerNorm. FP8 casts are unchanged.

Native W7900/gfx1100 red/green, with no architecture override:

- before: 388/1,024 mismatches in the exact direct-quant node and 3/262,144 in the exact fused-RMSNorm node;
- after: both nodes pass on physical GPU0 and GPU7;
- current AITER main `a50a62aa`, exact merge tree `d326412f`;
- the wheelhouse downloaded from the GitHub run is byte-identical to AITER's pinned ROCm-7.2 index download (Triton wheel SHA-256 `45710096...7504`, triton-kernels `452b7a3d...e1d5`);
- with that exact Triton `3.7.0+amd.rocm7.2.0.git89002410`: direct INT8 69/69, RMSNorm quant 216/216, LayerNorm quant 96/96;
- the existing public quant entrypoint, 8-shape HIP/CK RMSNorm gate, and OPUS finite boundary probe also exit 0.

For attribution: the OPUS jobs failed before tests because `libamdhip64.so.7` was unavailable, and shard 7 had one MXFP4 mismatch in an untouched FP4 path. I have not represented either as fixed by this commit.

The new workflows for `971cece9` are currently `action_required` pending repository approval. I have not requested a rerun or added CI labels.

PR #4641, opened 2026-08-08
[FlyDSL] Add SwiGLU activation to moe_gemm_2stage stage1 kernel

akii96:
## Summary

Adds compile-time selectable SwiGLU activation to the FlyDSL `moe_gemm_2stage` kernel and routes per-token FP8 activation + per-channel FP8 weight MoE through it on gfx942. This gives `EmbeddedLLM/MiniMax-M3-FP8-dynamic` a working fused MoE path, so it no longer has to run with AITER MoE disabled.

## Motivation

Serving `EmbeddedLLM/MiniMax-M3-FP8-dynamic` on gfx942 requires `VLLM_ROCM_USE_AITER_MOE=0`, falling back to vLLM's Triton MoE. The model needs clamped SwiGLU (alpha=1.702, limit=7.0) with per-token FP8 activations and per-channel FP8 weights, and no aiter pipeline covers that:

- `moe_gemm_2stage` was SiLU-only. #4646 removed the file once a16wi4, its last consumer, moved to the new pipeline. That was cleanup rather than deprecation, since no FP8 W8A8 route existed yet.
- The new `moe_2stage_a16wmix` is bf16-A only (native bf16, no A-scale), so FP8 activations would need new GEMM bodies.
- `mixed_moe_gemm_2stage_common` has `swiglu_mul_vec4` and an fp8 path, but its fp8 is MXFP8 microscaling (e4m3 + per-group e8m0) on `INTERLEAVE`, not per-token/per-channel W8A8 on `SEPARATED`.
- CK does not support this activation at the per-token FP8 contract, so it is not a usable baseline (it returns zeros at every token bucket).

## What changed

- **`moe_gemm_2stage.py`**: re-added, now with an `act` parameter selected at compile time, defaulting to `"silu"` with `"swiglu"` as the new option, reusing the existing `swiglu_mul_vec4` formula. Activation and clamp limit are in the JIT module name so caches can't collide.
- **`moe_kernels.py`**: FP8 W8A8 registries and compile routes, limited to the validated tiles.
- **`fused_moe.py`**: selects the path from the tuned config rather than a heuristic, and fails closed. Unless the whole contract holds (gfx942, BF16 out, per-token FP8, SwiGLU limit 7.0, preshuffled, `SEPARATED`, no EP/bias/padding), the config is discarded and today's fallback runs. Also re-sorts when stage 2's tile-M differs from stage 1's.
- **`aot/flydsl/moe.py`**: drops a redundant `swiglu_limit` argument that broke AOT for the `afp8_wfp4` FHMoE kernels.
- Two guards against silently-wrong output: a finite `swiglu_limit` on a non-SwiGLU activation is rejected rather than dropped, and `swiglu_limit=0.0` no longer coerces to 7.0.
- Plus tuned/untuned CSVs (16 token buckets) and unit tests.

SiLU stays the default and no existing dispatch decision changes.

## Test results

MI325X, MiniMax-M3 MoE shape at TP=4: hidden 6144, inter 768, 129 experts, topk 5, BF16 in/out, `fp8_e4m3fnuz` per-token/per-channel.

| Check | Result |
| --- | --- |
| `test_flydsl_moe_fp8_w8a8.py` | 9 passed |
| `gemm_moe_tune.py --run_config`, 16 token buckets | 16/16 `OK`, per-stage error 0.0% |
| AOT over the full kernel matrix | 2688 jobs ok, 0 failed |

## Performance versus the Triton fallback

Same container, same aiter build, same `run_perftest` harness. Triton is vLLM's `fused_experts_impl` with `use_fp8_w8a8`, `per_channel_quant`, `activation="swigluoai_uninterleave"`.

| Tokens | Triton (us) | FlyDSL (us) | Speedup |
| ---: | ---: | ---: | ---: |
| 1 | 64.68 | 40.30 | 1.61x |
| 128 | 528.04 | 454.67 | 1.16x |
| 2048 | 804.44 | 672.79 | 1.20x |
| 8192 | 2234.34 | 1766.19 | 1.27x |
| 32768 | 8394.91 | 5730.96 | 1.46x |

FlyDSL wins all 16 buckets (1.05x–1.61x), widening with batch size. Caveats: vLLM ships no tuned Triton config for `E=129, N=768` on MI325X, so Triton ran default tiles, and this is the MoE layer in isolation, not end-to-end serving.

## To-do

- The tuned CSV is scoped to `inter_dim=768`, (TP=4). TP=8 support is needed next.


maeehart:
GitHub reports CONFLICTING. Can you rebase onto current main and push?

PR #4644, opened 2026-08-09
[Gluon][MLA] Fix FP8 KV serving and vectorize split reduction

JohnQinAMD:
## Summary

Supersedes #4480 with a small, self-contained update:

- enable batched FP8 KV decode
- widen KV row offsets only when the cache exceeds the int32 element range
- advertise batched decode, MTP, and KV-scale capabilities to callers
- vectorize the stage-2 split reduction

The fused MTP kernel rewrite is intentionally excluded.

## Validation

Validated on mia1-p02-g56 (gfx950):

- B1/B8/B64, qlen 1/3, context 68,089, KV scale 0.375: output and LSE PASS against FP32 reference
- CUDA graph capture and two replays, including mutated Q: PASS
- Black, Ruff, and git diff check: PASS

68K graph replay, before -> after:

| Shape | Before | After |
|---|---:|---:|
| B1 qlen1 | 92.374 us | 33.219 us |
| B1 qlen3 | 53.152 us | 35.771 us |
| B64 qlen1 | 420.124 us | 420.078 us |
| B64 qlen3 | 1185.395 us | 1185.725 us |

B64 is neutral within run-to-run noise. This PR does not claim to beat ASM at 640K or 1M context; that requires a separately reviewable fused-MTP change.

## AI assistance

This PR was developed with assistance from OpenAI Codex. The author reviewed and validated all changes.

### DIFF
--- a/aiter/ops/triton/gluon/mla_gluon.py
+++ b/aiter/ops/triton/gluon/mla_gluon.py
@@ -99,6 +99,7 @@ def _mla_gluon(
     KV_PE_OFFSET: gl.constexpr,
     USE_2D_VIEW: gl.constexpr,
     WITHIN_2GB: gl.constexpr,
+    WIDE_KV_OFFSETS: gl.constexpr,
     NUM_XCDS: gl.constexpr,
     NHEAD: gl.constexpr,
     REGIME: gl.constexpr,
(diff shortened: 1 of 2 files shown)


JohnQinAMD:
Closing because the reduced scope does not deliver the required long-context and large-batch performance improvement. The performance path will be redesigned as a separately reviewable change.

PR #4659, opened 2026-08-10
[Triton/Gluon] Add two fused ops for diffusion transformer blocks

carlushuang:
## What

Two Triton ops for the elementwise work in a diffusion transformer block. Both are memory bound, and both are dominated by temporaries the maths does not need. Purely additive — six new files, no existing code touched, no `__init__` exports changed.

### `fused_rmsnorm_indexed_adaln`

```
out[m] = rmsnorm(x[m], weight) * (1 + scale[idx[m]]) + shift[idx[m]]
```

Adaptive layernorm: every token indexes a small table of modulation vectors, one row per (modality, timestep). Unfused, the normalised activation is written and immediately re-read, and both table gathers are materialised at `[M, N]` — 680 MB each at a 63k-token request.

One program owns a block of rows and walks each row in column tiles, once to accumulate the sum of squares and once to normalise and modulate, so `x` is read once and `out` written once.

Two details specific to this workload:
- **Rows are tiled, not padded to the next power of two.** A 5376-wide row would mask off a third of every access at 8192.
- **A block of consecutive tokens usually shares one modulation index** — packed sequences are laid out in runs of one modality — so the kernel detects that and collapses the `[BLOCK_M, BLOCK_N]` gather to a single `[BLOCK_N]` load broadcast in registers.

### `fused_qk_norm_rope_cached`

```
q[t, h] = rope(rmsnorm(q[t, h], q_weight), cos_sin_cache[t])      (and k)
```

The existing rope ops do not cover this case. They assume the rotated subspace is the whole head or half of it, and diffusion transformers rotate fractions in between — 96 of 128 for MiniMax-H3. The cache-write variants additionally want a paged KV cache, which a diffusion model does not have.

One program owns a token. A token's heads are contiguous, so the `[H, D]` tile is one coalesced run and the token's cos/sin row is read once for all heads rather than broadcast into a `[T, H, D]` temporary. Only each token's `[H, D]` block must be contiguous, so `q` and `k` can be **strided views into a packed qkv projection** and are rotated in place, never materialised.

## Measured

MI355X (gfx950), bf16, at MiniMax-H3's shapes:

| | eager chain | fused | |
|---|---:|---:|---:|
| rmsnorm + indexed adaln, 63232 × 5376 | 1.708 ms | 0.329 ms | **5.2×** |
| qk norm + rope, 63232 × 56 × 128 | 11.690 ms | 1.262 ms | **9.3×** |

Both hold across the token counts one rank sees at Ulysses 1/2/4/8 (5.0–5.2× and 9.2–9.9×).

End to end in a 50-layer video diffusion model, 209 frames at 1344×768, 50 denoise steps:

| | before | after |
|---|---:|---:|
| Ulysses-1 | 387.4 s | **355.1 s** |
| Ulysses-4 | 111.5 s | **103.1 s** |

## Accuracy

Both keep the row in fp32 across the whole fusion, so they are *nearer* the fp32 result than the op chain they replace, which rounds to bf16 at each step. Against that chain on a real 50-layer model with trained weights, one forward agrees to **cosine 1.0000000, max relative error 1.3e-4**.

`fused_rmsnorm_indexed_adaln` takes `round_intermediate=` for callers who would rather match an unfused reference's rounding than keep the extra precision.

## Tests

58 tests across dtypes, shapes and both index layouts. The ones worth naming, because each catches a bug that a naive test would not:

- every table row exercised individually — a kernel that broadcast row 0 passes a uniform-index test
- uniform and scattered index paths checked against each other, since they take different routes through the kernel
- `q` and `k` given different norm weights — applying `q_weight` to both passes every symmetric test
- the unrotated tail checked for passthrough
- the strided-qkv-view case checked to leave `v` untouched

```
pytest op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
pytest op_tests/triton_tests/rope/test_fused_qk_norm_rope_cached.py
```

## Consumer

ROCm/ATOM#1836 calls both from its MiniMax-H3 diffusion pipeline, with an eager fallback when aiter is absent, so it does not hard-depend on this landing.


carlushuang:
cc @brunomazzottiamd

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Doing the norm and the modulation together is what makes this worth a
+    kernel: separately, the activation is written and re-read between them, and
+    ``scale``/``shift`` are materialised at [M, N] by the gather. Fused, x is
+    read once and out written once, and the table is read straight from cache.
+
+    Returns:
+    - out: [M, N], same dtype as x
+    """
+    _LOGGER.info(
```
brunomazzottiamd, line 53: Can you please use `%` placeholders in log messages?

```python
    _LOGGER.info(
        "FUSED_RMSNORM_INDEXED_ADALN: x=%s table=%s idx=%s",
        tuple(x.shape),
        tuple(scale.shape),
        tuple(indices.shape),
    )
```

Using f-strings directly inside Python's built-in logging functions is generally considered an anti-pattern. While it's a working solution, using an f-string evaluates and builds the string eagerly before checking the log level. The best practice for standard Python logging is to use lazy evaluation with % placeholders.

I've noticed this pattern all over AITER Triton kernels.

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,103 @@
...
+        shift.ndim == 2 and shift.shape[1] == N
+    ), f"modulation table must be [G, {N}], got {tuple(shift.shape)}"
+    assert indices.shape == (M,), f"indices must be [{M}], got {tuple(indices.shape)}"
+    assert x.is_contiguous(), "x must be contiguous"
+    assert indices.is_contiguous(), "indices must be contiguous"
+
+    if out is None:
+        out = torch.empty_like(x)
+    else:
+        assert out.shape == x.shape and out.dtype == x.dtype
```
brunomazzottiamd: Can you please add an error string to this `assert` statement?

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,103 @@
...
+        assert out.is_contiguous(), "out must be contiguous"
+    if M == 0:
```
brunomazzottiamd: **Sincere question:** Is this `if M == 0` check really necessary? Does any framework forward empty tensors to AITER ops?

op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,130 @@
...
+    a broadcast that quietly ignores per-row indices.
+    """
+    dev = "cuda"
+    M, N, G = 256, 5376, 4
+    x, weight, shift, scale, _ = make(M, N, G, torch.bfloat16, dev)
+    same = torch.full((M,), 2, dtype=torch.int64, device=dev)
+    fast = fused_rmsnorm_indexed_adaln(x, weight, shift, scale, same)
+
+    # Same indices, but arranged so no block is uniform.
+    alternating = torch.full((M,), 2, dtype=torch.int64, device=dev)
```
brunomazzottiamd: What's the difference between `same` and `alternating` indices? As far as I can see, they are built in the same way:

```python
same        = torch.full((M,), 2, dtype=torch.int64, device=dev)
alternating = torch.full((M,), 2, dtype=torch.int64, device=dev)
```

`fast` and `slow` outputs will be exactly the same:

```python
fast = fused_rmsnorm_indexed_adaln(x, weight, shift, scale, same)
slow = fused_rmsnorm_indexed_adaln(x, weight, shift, scale, alternating)
```

Is this the true intention behind the test?

op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,130 @@
...
+        for g in range(G):
+            indices[edges[g] : edges[g + 1]] = g
+    else:
+        indices = torch.randint(
+            0, G, (M,), generator=gen, device=device, dtype=torch.int64
+        )
+    return x, weight, shift, scale, indices
+
+
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16, torch.float32])
```
brunomazzottiamd, line 40: What's the running time of `test_fused_rmsnorm_indexed_adaln.py`?

I'm asking because we had some GEMM tests in the past that were sweeping across `{bf16, fp16, fp32}` data type space, and they were taking a considerable amount of time to run. Real models used just `bf16` data type and we ended up removing `fp16` and `fp32` for the sake of speeding up CI.

Please take this into consideration. I'm not against testing `{bf16, fp16, fp32}` data types if this is a really fast test.

aiter/ops/triton/_triton_kernels/fusions/fused_rmsnorm_indexed_adaln.py
```diff
@@ -0,0 +1,123 @@
...
+        x = tl.load(
+            x_ptr + rows[:, None] * stride_x_row + cols[None, :],
+            mask=row_mask[:, None] & col_mask[None, :],
+            other=0.0,
+        ).to(tl.float32)
+        acc += tl.sum(x * x, axis=1)
+
+    # torch.nn.RMSNorm reduces in fp32 for reduced-precision input; match that
+    # rather than the storage dtype, or the norm drifts on long rows.
+    rstd = 1.0 / tl.sqrt(acc / N + eps)
```
brunomazzottiamd: What about [`tl.rsqrt`](https://triton-lang.org/main/python-api/generated/triton.language.rsqrt.html#triton.language.rsqrt)? Do we face precision issues with it? Can the compiler lower `1 / sqrt(x)` to `V_RSQ_F32`?

aiter/ops/triton/_triton_kernels/rope/fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,83 @@
...
+    partner = tl.where(rotate, partner, 0)
+    sign = tl.where(upper, 1.0, -1.0)
+
+    for which in tl.static_range(2):
+        base = q_ptr + tok * stride_q_tok if which == 0 else k_ptr + tok * stride_k_tok
+        w_ptr = qw_ptr if which == 0 else kw_ptr
+
+        x = tl.load(base + offs, mask=mask, other=0.0).to(tl.float32)
+        # RMSNorm is per head over D, reduced in fp32 like torch.nn.RMSNorm.
+        rstd = 1.0 / tl.sqrt(tl.sum(x * x, axis=1) / D + eps)
```
brunomazzottiamd: [`tl.rsqrt`](https://triton-lang.org/main/python-api/generated/triton.language.rsqrt.html#triton.language.rsqrt)?

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,92 @@
...
+    q: torch.Tensor,
+    k: torch.Tensor,
+    q_weight: torch.Tensor,
+    k_weight: torch.Tensor,
+    cos_sin_cache: torch.Tensor,
+    eps: float = 1e-5,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Per-head RMSNorm then partial NeoX RoPE on q and k, in place.
+
+        q[t, h] = rope(rmsnorm(q[t, h], q_weight), cos_sin_cache[t])
```
brunomazzottiamd, line 25: I think the equation for `k` is missing in the wrapper docstring.

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,92 @@
...
+
+    One program owns a token. A token's heads are contiguous, so the [H, D]
+    tile is one coalesced run and the token's cos/sin row is read once for all
+    heads -- unfused, that row is broadcast into a [T, H, D] temporary, and the
+    rotation costs a negate and two concatenates on top.
+
+    Returns:
+    - (q, k), the same tensors
+    """
+    _LOGGER.info(
```
brunomazzottiamd, line 48: _ditto_ w.r.t. log placeholders.

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,92 @@
...
+        D,
+    ), f"norm weights must be [{D}]"
+    assert (
+        cos_sin_cache.ndim == 2 and cos_sin_cache.shape[0] == T
+    ), f"cache must be [{T}, R], got {tuple(cos_sin_cache.shape)}"
+    rot = cos_sin_cache.shape[1]
+    assert (
+        rot % 2 == 0 and rot <= D
+    ), f"rotated width {rot} must be even and at most head_dim {D}"
+    if T == 0:
```
brunomazzottiamd: Same question w.r.t. to empty tensors.

op_tests/triton_tests/rope/test_fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,112 @@
...
+    q = torch.randn(T, H, D, generator=g, device="cuda", dtype=dtype)
+    k = torch.randn(T, H, D, generator=g, device="cuda", dtype=dtype)
+    qw = torch.randn(D, generator=g, device="cuda", dtype=dtype)
+    kw = torch.randn(D, generator=g, device="cuda", dtype=dtype)
+    ang = torch.randn(T, R // 2, generator=g, device="cuda", dtype=torch.float32)
+    cache = torch.cat((ang.cos(), ang.sin()), dim=-1).to(dtype)
+    return q, k, qw, kw, cache
+
+
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32])
```
brunomazzottiamd, line 40: Same comment w.r.t. to test running time and data types used by real models.

op_tests/triton_tests/rope/test_fused_qk_norm_rope_cached.py
```diff
@@ -0,0 +1,112 @@
...
+    gq, gk = fused_qk_norm_rope_cached(q, k, qw, kw, cache)
+    assert gq.data_ptr() == q.data_ptr() and gk.data_ptr() == k.data_ptr()
+
+
+def test_q_and_k_use_their_own_weights():
+    """A kernel that applied q_weight to both would pass every symmetric test."""
+    q, k, qw, kw, cache = make(16, 4, 128, 96, torch.float32)
+    kw = kw * 3.0
+    wq, wk = reference(q.clone(), k.clone(), qw, kw, cache, 1e-5)
+    gq, gk = fused_qk_norm_rope_cached(q, k, qw, kw, cache)
```
brunomazzottiamd: Don't we need the same `eps` from `reference` here?

brunomazzottiamd (dismissed):
LGTM! Solid PR!

I've left some minor comments, address them at your own judgement.

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
carlushuang: Done — it was the only assert in the file without one.

```python
assert (
    out.shape == x.shape and out.dtype == x.dtype
), f"out must be [{M}, {N}] {x.dtype}, got {tuple(out.shape)} {out.dtype}"
```

aiter/ops/triton/_triton_kernels/rope/fused_qk_norm_rope_cached.py
carlushuang: Done — same change, covered by the same run noted on the adaLN kernel (58/58, fp32 at `2e-5` here).

aiter/ops/triton/_triton_kernels/fusions/fused_rmsnorm_indexed_adaln.py
carlushuang: Switched, in this kernel and the RoPE one. Checked rather than assumed, since the fp32 tolerances are tight: all 58 tests across both files pass unchanged, including fp32 at `2e-6` (adaLN) and `2e-5` (RoPE).

op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
carlushuang: You're right, and it's worse than cosmetic: that assertion could not fail. Both tensors were `torch.full((M,), 2)`, so `fast` and `slow` were the same call on the same input — which left the kernel's `uniform = tl.min(idx) == tl.max(idx)` broadcast branch with no coverage at all. The intent was unreachable as written too: with a single index value, no arrangement of it is ever non-uniform.

Fixed by making two table entries identical, so the same modulation is reachable both uniformly (broadcast branch) and alternating (gather branch), and the two must agree bit for bit:

```python
shift[5] = shift[2]
scale[5] = scale[2]
```

Checked that it now bites: breaking the uniform branch so it reads table row 0 rather than the block's index makes the new assertion fail, where the old one still passed. Thanks for catching this.

op_tests/triton_tests/rope/test_fused_qk_norm_rope_cached.py
carlushuang: **6.58 s** for all 17 tests here, slowest single test 0.07 s. Only bf16 and fp32 are swept in this file — fp32 is what holds the norm to `2e-5` rather than bf16's `3e-2`.

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
carlushuang: Agreed — removed here too.

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
carlushuang: Right, it was missing. Added:

```
q[t, h] = rope(rmsnorm(q[t, h], q_weight), cos_sin_cache[t])
k[t, h] = rope(rmsnorm(k[t, h], k_weight), cos_sin_cache[t])
```

aiter/ops/triton/rope/fused_qk_norm_rope_cached.py
carlushuang: Done

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
carlushuang: No, it isn't — removed, along with the matching `if T == 0` in the RoPE wrapper. It was speculative: no caller forwards an empty tensor, and there was no test behind it.

op_tests/triton_tests/rope/test_fused_qk_norm_rope_cached.py
carlushuang: Good catch on the inconsistency. Not a live bug — the wrapper default *is* `eps: float = 1e-5`, so the two matched — but the test shouldn't quietly depend on that while the reference states it explicitly.

Every call whose output is asserted on now passes `eps=1e-5`, in this file and in the adaLN test, which had the same asymmetry in several places. The only ones left on the default are the `pytest.raises` argument-validation tests and `test_is_in_place`, where no value is compared.

op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
carlushuang: Measured on MI355X:

| file | tests | total | slowest single |
| --- | --- | --- | --- |
| `test_fused_rmsnorm_indexed_adaln.py` | 41 | **7.55 s** | 0.07 s |
| `test_fused_qk_norm_rope_cached.py` | 17 | **6.58 s** | 0.07 s |

107 of the 149 measured durations are under 5 ms — almost all the wall clock is interpreter startup and Triton JIT warmup, not the sweep. Dropping fp16 and fp32 would save about a second.

I'd rather keep them: fp32 is what actually pins the reduction (it asserts `2e-6`, where bf16 only asserts `3e-2`), and fp16 is what would catch an overflow in the affine. Happy to trim if you'd still prefer it.

aiter/ops/triton/fusions/fused_rmsnorm_indexed_adaln.py
carlushuang: Done.

op_tests/triton_tests/fusions/test_fused_rmsnorm_indexed_adaln.py
brunomazzottiamd: That's fine! Let's keep all data types in the tests!

brunomazzottiamd (approved):
LGTM! Let's get CI green before merging!

brunomazzottiamd:
**Standard Tests (1 GPU) (linux-aiter-mi35x-1, MI35X, 8, 1)** is failing. This seems to be completely unrelated to this PR.

PR #4661, opened 2026-08-10
[topk_per_row] support per-row KV ends for DeepSeek-V4 CSA decode

amd-ruitang3:
## Summary

`top_k_per_row_decode` derives each row's KV bound from a per-SEQUENCE `seqLens`:

```
seqLens[r / next_n] - next_n + (r % next_n) + 1
```

i.e. one KV row per query token. DeepSeek-V4 compressed sparse attention packs
`ratio` tokens into a single index-cache entry, so that rule yields the wrong
bound for those callers. This adds an optional `rowEndsPerRow`: one end per ROW,
used as is.

## Changes

- `top_k_per_row_decode` gains an optional `rowEndsPerRow` across all layers —
  `csrc/kernels/topk_per_row_kernels.cu`, `csrc/include/topk_per_row.h`, the
  pybind signature in `csrc/include/rocm_ops.hpp`, and the wrapper in
  `aiter/ops/topk.py`. It defaults to `nullopt`/`None`, so existing callers are
  unaffected.
- When present, the array is routed into the `rowEnds` slot and `next_n` is
  forced to `1`, which collapses the per-sequence expression to `rowEnds[r]`.
  No second kernel code path is introduced.
- Comments at each site that depends on that collapse, since it holds only as
  long as `next_n` feeds nothing but that one expression on the decode path
  (currently true: decode dispatches only to `radix_topk_one_block_kernel`).
- New regression `test_decode_row_ends_per_row` in `op_tests/test_topk_per_row.py`.

## Test plan

All run on gfx950 (MI355X):

- Full default sweep, `python3 op_tests/test_topk_per_row.py` (what CI runs):
  **561/561 rows `all_close=True`** (33 prefill + 528 decode), exit code 0, and
  both regressions pass:
  ```
  [mb_workspace_reuse] PASS: 3 reused-buffer mb calls matched torch.topk
  [row_ends_per_row]   PASS: 12 irregular per-row ends honoured verbatim
                             (bogus seqLens/next_n=3 ignored), and 12 rows
                             agreed with the seqLens path
  ```
- Negative control: re-running the same case with `rowEndsPerRow` omitted makes
  5 short rows violate the bound check, so the new test genuinely fails when the
  feature is bypassed rather than passing vacuously.
- `black --check` clean; `ruff check` clean.

The new test asserts three things: the ends are honoured verbatim (using 12
irregular, non-monotonic ends that no `(seqLens, next_n)` pair could generate);
`seqLens` and `next_n` are ignored (both passed deliberately wrong, so dropping
the `next_n = 1` line makes it fail); and it selects the same elements as the
`seqLens` path when the two agree.

Two details that make the test sensitive, worth knowing if it ever needs editing:

- Out-of-range logits are filled with the *largest* values in the row, not
  `-inf`. With `-inf` padding (what `create_random_logits` does) an overshooting
  bound is invisible — the kernel scans the tail, selects nothing from it, and
  still matches.
- Rows with `end <= top_k` take the kernel's fill path, which emits
  `0..end-1` padded with `-1`; those are compared over the *whole* row, because
  an overshooting bound leaves the leading entries identical and only shows up
  in the padding. That is what the negative control above trips.

## Notes / possible follow-ups

Not addressed here, flagging for reviewers:

- `top_k_per_row_decode_fast` (the ASM path, gfx942 + `k=2048`) has no
  `rowEndsPerRow`, so CSA callers cannot use it.
- `rowEndsPerRow` is unvalidated: a non-int32 tensor throws inside
  `data_ptr<int>()`, and `numel() != numRows` reads out of bounds. A
  `TORCH_CHECK` on dtype and length would be cheap.

Unrelated pre-existing behaviour observed while testing: decode passes
`sorted=true`, but the output comes back in index order rather than value order,
and its tail is not stable across identical launches. `compare_topk_results` is
set-based, so this does not affect the tests.

### DIFF
--- a/aiter/ops/topk.py
+++ b/aiter/ops/topk.py
@@ -398,6 +398,7 @@ def _top_k_per_row_decode(
     stride1: int,
     k: int = 2048,
     workspace: torch.Tensor | None = None,
+    rowEndsPerRow: torch.Tensor | None = None,
 ) -> None: ...
 
 
@@ -410,9 +411,16 @@ def top_k_per_row_decode(
     stride0: int,
     stride1: int,
     k: int = 2048,
+    rowEndsPerRow: torch.Tensor | None = None,
 ) -> None:
     """Per-row top-k (decode). Always uses the one-block kernel — the C++
-    side ignores the workspace argument for decode."""
+    side ignores the workspace argument for decode.
+
+    `seqLens` is per-SEQUENCE and bounds row `r` by
+    `seqLens[r // next_n] - next_n + (r % next_n) + 1` — one KV row per query
+    token. Callers whose index cache packs several tokens into one entry
+    (DeepSeek-V4 CSA) must pass `rowEndsPerRow` instead: one end per ROW, taken
+    as is, with `seqLens` then unused."""
     # Decode always takes the ob path (see topk_per_row_kernels.cu).
     # The original mb dispatch is commented out below for reference:
     #   workspace = None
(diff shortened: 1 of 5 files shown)


amd-ruitang3:
we can just modify in ATOM layer

PR #4382, opened 2026-07-24
[TRITON][GLUON][GFX950][DSV4] Paged Sparse Attention Gluon Kernel

cagrikymk:
# gfx950 paged sparse attention for DSv4

Adds a Gluon decode kernel for DeepSeek-V4 sparse MLA on gfx950. It covers both KV storage
formats in use (ATOM's uniform pool and vLLM's packed `fp8_ds_mla`) and both fp8 flavors.
`pa_decode_sparse` routes to it automatically; anything it doesn't cover falls back to the
existing Triton kernel.

## Backend selection

`pa_decode_sparse` picks from arch + inputs: gfx950 with `block_h=None` gets the Gluon kernel,
otherwise the shared fallback (Triton, or the gfx1250 Gluon kernel on gfx1250). q is bf16/fp16,
output is bf16.

Only the gfx950 Gluon path handles the 3D packed layout, the two-loop form (separate SWA and
top-k caches, which is how vLLM stores them), and both fp8 flavors. The fallback paths take the
2D uniform pool only and assert fnuz — OCP is rejected by a launcher gate rather than a kernel
limit, since the gfx1250 dequant is dtype-driven and would handle it.

## fp8 flavors

OCP (`float8_e4m3fn`, bias 7) has a native gfx950 cvt. fnuz (`float8_e4m3fnuz`, bias 8) does not,
so we route it through bf16 (`float8e4b8 → bf16 → f32`). That's exact — fp8 has 3 mantissa bits —
and far cheaper than the direct fnuz→f32 path, which lowers to a software unpack that spills.

## KV storage formats

Both use 1×64 block scaling. DSv4 is D=512 (NoPE 448 + RoPE 64) with page block 256.

Both formats are supported by the gfx950 Gluon kernel.

### Uniform pool, 2D — used by ATOM

One unified cache for both SWA and top-k.

- `unified_kv: [pages, D]` fp8 viewed as uint8, one token per slot (page_size 1)
- `kv_scales: [pages, D//64]` fp32, a separate side tensor
- The whole head is fp8, one fp32 scale per 64-element group (8 groups at D=512). Flavor is the
  caller's choice.
- Dequant: `x = fp8_to_f32(byte) * kv_scales[slot, col//64]`
- Addressing: the slot indexes the row directly, so data sits at `slot*D + col` and its scale at
  `slot*(D//64) + col//64`.

### Packed `fp8_ds_mla`, 3D — used by vLLM

[...]

PR #4673, opened 2026-08-11
[triton] fix sparse decode gathering zeros from a strided cache

jiacao-amd:
## Problem

`_pa_decode_sparse_gfx950_gluon` picks between 32-bit `buffer_load` and 64-bit `gl.load` from the **logical** tensor size:

```python
cache_bytes     = cache.nelement() * cache.element_size()
use_buffer_load = cache_bytes <= MAX_BYTES
```

That only equals the reachable span when the tensor is contiguous, and callers can legitimately pass a non-contiguous view of a larger KV allocation.

Seen with a packed `fp8_ds_mla` cache of `[47233, 64, 584]`, stride `(1435968, 584, 1)` — dim-0 pitch 38.4x the block pitch. Logical size 1.64 GiB takes `buffer_load`; the real span is **63.17 GiB**, leaving blocks 1496..47232 (**96.8% of the pool**) unreachable in 32 bits.

| blocks | offset behaviour | result |
|---|---|---|
| 1496 – 2990 | `>= num_records` | gather returns zeros |
| 2991 – 47232 | passes `2**32`, int32 truncation **wraps** in range | returns **an unrelated block's bytes** |

The second is the dangerous one: `2991*1435968 = 4294980288` truncates to voffset `12992`, well inside `num_records`, so the load succeeds silently. Downstream this looks like a model producing coherent text and then degenerating.

The kernel body is fine — `_cache_load` and the `cs0.to(gl.int64)` promotion both handle >2 GiB. Only the dispatcher's size computation is wrong, so that path is never selected.

## Fix

Derive the span from the strides, so the mode follows the addresses the kernel actually forms:

```python
span = 1
for size, stride in zip(t.shape, t.stride()):
    if size > 1:
        span += (size - 1) * stride
return span * t.element_size()
```

It lives in `utils/common_utils.py` as `max_addressable_bytes` — any kernel picking `buffer_load` off a size has the same exposure. No fast-path regression: for a contiguous tensor the span equals `nelement() * element_size()` exactly.

**The cap was also off by one.** The descriptor pins `num_records = 2147483646` and the rule is `offset >= num_records` → out of range, so a span of exactly `2**31 - 1` passed the old gate and read its top byte as zero (measured on gfx950: offset `2147483645` returns the marker, `2147483646` returns `0x00`). The gate is now `span < MAX_BYTES`, with `MAX_BYTES` left at `2**31 - 1`.

## Tests

`test_pa_decode_sparse_two_loop` gains a `strided_cache` argument that runs the kernel over a 2.0 GiB pool whose dim-0 pitch puts the last block past `2**31` bytes — same contents and `nelement()`, only the layout differs. Unfixed: **9598 / 1048576 elements wrong, 908x max relative difference**. Fixed: passes. Reverting either call site turns CI red.

Pinned to the fp8 `T=128` shape with a free-VRAM guard, so 1 of 24 `two_loop` combinations pays it: **1.92 s**, whole file **114 passed, 10 skipped in 26.7 s**.

Existing coverage missed this because every test allocates its cache fresh and contiguous — instrumenting the dispatcher gave **124/124 calls with `is_contiguous() == True`**. The int64 *path* was covered; the *selection logic* for "small logical size, large span" was not.

## Notes

Introduced by #4382; affects any caller handing this kernel a strided cache view. vLLM's NVIDIA DSv4 path threads an explicit packed block span through for the same reason (`_packed_block_span`, citing vllm-project/vllm#44577), so the layout is expected — it is the ROCm dispatcher reasoning from the wrong number.

Two adjacent pre-existing issues found while reviewing, **not** fixed here, happy to file separately: the gluon split-K reduce path casts `a_off`/`q_off`/`part_*` to `gl.int32` with no `USE_BUFFER_LOAD` gate at all, and `unified_attention.py:878` has the same `nelement()`-based gate (measuring only `k`, never `v`) — currently dead, but inherits every defect once wired up.


aiter/ops/triton/attention/pa_decode_sparse.py
```diff
@@ -52,6 +52,22 @@
 _FP8_DTYPE = torch.float8_e4m3fnuz
 
 
+def _addressable_bytes(t: torch.Tensor) -> int:
```
cagrikymk: I think this can go under https://github.com/ROCm/aiter/blob/main/aiter/ops/triton/utils/common_utils.py

aiter/ops/triton/attention/pa_decode_sparse.py
```diff
@@ -486,7 +502,9 @@ def _pa_decode_sparse_gfx950_gluon(
     BLOCK_M, BLOCK_K, MFMA_K, waves_per_eu = 16, 64, 16, 0
     num_warps = BLOCK_K // 16
     NOPE_DIM, ROPE_DIM = 448, 64
-    MAX_BYTES = 2**31 - 1  # buffer_load 32-bit offset cap; larger -> gl.load int64
+    # buffer_load span: make.buffer.rsrc pins num_records to 2**31 - 2 and the
+    # rule is offset >= num_records -> OOB. Larger -> gl.load, which is int64.
+    MAX_BYTES = 2**31 - 2
```
cagrikymk: I think keeping this as 2**31 - 1 and changing the <= to < is a better choice semantically.

op_tests/triton_tests/attention/test_pa_decode_sparse.py
```diff
@@ -534,3 +537,24 @@ def test_pa_decode_sparse_two_loop(T, H, D, main_len, extra_len, dtype):
 
     tol = 1e-2 if dtype == "fp8" else 5e-3
     torch.testing.assert_close(out, ref, atol=tol, rtol=tol)
+
+
+def test_addressable_bytes():
```
cagrikymk: I think it is better to test if the kernel produces correct results under these conditions.

So, we can add a bool arg. to test_pa_decode_sparse_two_loop that creates non-contig. large tensor with indices that would produce > int32 addresses while keeping the CI runtime in mind.

cagrikymk:
@jiacao-amd 
I added some comments, thank you for catching this!

aiter/ops/triton/attention/pa_decode_sparse.py
cagrikymk: Also, the name can be changed to max_addressable_bytes, it better reflects the functionality.

aiter/ops/triton/attention/pa_decode_sparse.py
```diff
@@ -52,6 +52,22 @@
 _FP8_DTYPE = torch.float8_e4m3fnuz
 
 
+def _addressable_bytes(t: torch.Tensor) -> int:
+    """Byte span the kernel's gather offsets can reach in ``t``.
+
+    ``nelement() * element_size()`` is the logical size and only equals the
+    span when ``t`` is contiguous; a strided view of a larger allocation
```
cagrikymk: Also, can you simplify the comment a bit? The code and function name are overall self-explanatory.

aiter/ops/triton/attention/pa_decode_sparse.py
```diff
@@ -52,6 +52,22 @@
...
+    addresses the same elements over a far wider range. It is the range that
+    has to fit a 32-bit buffer_load offset. Derived from the strides, so it
+    also covers the overlapping and zero-stride views a broadcast produces.
+    """
+    span = 1
+    for size, stride in zip(t.shape, t.stride()):
+        if size > 1:
+            span += (size - 1) * abs(stride)
```
cagrikymk: I dont think negative strides are allowed with torch, we can safely remove abs IMO

aiter/ops/triton/attention/pa_decode_sparse.py
jiacao-amd: Done

aiter/ops/triton/attention/pa_decode_sparse.py
jiacao-amd: Done, good catch

op_tests/triton_tests/attention/test_pa_decode_sparse.py
jiacao-amd: Verified in both directions on gfx950: fails on the unfixed dispatcher (9598/1048576 elements off, 908x max relative difference) and passes with the fix.

aiter/ops/triton/attention/pa_decode_sparse.py
jiacao-amd: Done

aiter/ops/triton/attention/pa_decode_sparse.py
jiacao-amd: removed.

cagrikymk (approved):
LGTM!

k50112113 (approved):

PR #4676, opened 2026-08-11
[FlyDSL] fp8 unified attention for gfx950

johannes-graner:
## Motivation

Qwen3-VL MLPerf inference on MI355X/gfx950 needs native FP8×FP8 MFMA unified
attention on a paged KV cache. aiter's Triton `unified_attention` upcasts the FP8
KV cache before the QK GEMM, forfeiting the FP8 MFMA peak; the CK kernel that does
the FP8 intersection was never merged. This ports the
kernel to FlyDSL so the FP8 path is native and in-tree.

## Technical Details

- Native FP8×FP8 MFMA unified attention (QK and PV) for gfx950, ported to FlyDSL
  v0.3.0 (`aiter/ops/flydsl/kernels/flash_attn_{fp8_gfx950,dualwave_common}.py`).
  Lifts the three FP8 rejections the production call hits: **varlen** (packed
  ragged Q), **paged KV** (block_size 64), and **split-K** (multi-partial combine).
  The fp8 paged/varlen/split-K/sinks machinery is aiter-owned, not
  upstream-tracked; main loop is 8-cluster restructured (softmax-VALU bound).
- **Shuffled 5D KV-cache support** (`SHUFFLED_KV_CACHE_1`, the vLLM-v1 production
  layout): vectorized K/V loaders, coalesced V staging, a 5D descriptor-geometry
  override. Auto-selected from tensor rank (`k.dim() == 5`); no new caller flag.
- **GQA-packed M**: factors M as query-positions × GQA-group (mirroring Triton) so
  one kernel serves prefill and decode.
- **Decode-specialized kernel** (`aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py`,
  new): a `BLOCK_M=16` GQA-group `mfma_16x16x32_f8` body sized to the live M-rows
  of GQA-16:1 `query_len=1` decode (the shared 2d body fills only 1 of 16 M-rows
  there). Multi-wave cooperative workgroup with on-chip LSE combine, async
  `buffer_load...lds` KV staging + register V-prefetch, fill-aware inter-workgroup
  split-KV (grid.z) with a global-memory LSE-combine kernel, and per-page
  `BufferDesc` rebasing so the 2 GB / 65536-block production pool stays inside
  int32 buffer offsets. Serves the decode region behind a centralized
  decode-dispatch gate; a narrow mid-batch/deep-context band
  (9 ≤ num_seqs ≤ 48 and max_seqlen_k ≥ 8192) is conservatively ceded to Triton.
- **Adaptive dispatch, no caller flag**: split-K auto-fires for underfilled regimes
  (decode-only, ragged decode, low-chunk mixed; cap 16, overridable via
  `AITER_UNIFIED_ATTN_MAX_KV_SPLITS`, verified to 128). Mixed prefill/decode
  batches are partitioned and each half routed to its optimal tier when the decode
  context is deep enough (gate + decline-memo so the partition check isn't re-paid
  per decoder layer); underfilled shallow-mixed batches cede to Triton
  (return `None`); full batches stay single-pass.
- Shape widening: non-causal, fp16 output, and attention sinks.
- Wired into the public `unified_attention()` entry point via a return-`None`
  opt-in adapter (`aiter/ops/flydsl/unified_attention_kernels.py`). Known
  constraint (gated): the block-table LDS stager caps KV at 131072 tokens
  (2048 pages × 64).

## Test Plan

gfx950, `~/.flydsl/cache` cleared. Under `op_tests/flydsl_tests/`:

```
python3 -m pytest op_tests/flydsl_tests/test_vendored_fp8_attention.py -rs
python3 -m pytest op_tests/flydsl_tests/test_fp8_paged.py -rs
python3 -m pytest op_tests/flydsl_tests/test_fp8_splitk.py -rs
python3 -m pytest op_tests/flydsl_tests/test_packed_splitk_{decode,mixed}.py -rs
python3 -m pytest op_tests/flydsl_tests/test_flydsl_unified_attention.py -rs  # consolidated prefill/decode/mixed
python3 -m pytest op_tests/flydsl_tests/test_unified_attention_routing.py -rs  # end-to-end via public entry point
```

## Test Result

Correctness — worst absolute error 0.055, cosine ≥ 0.9996 across all regimes:

| Test | Coverage | Result |
|---|---|---|
| `test_vendored_fp8_attention` | dense bit-identical to upstream + ragged varlen | pass |
| `test_fp8_paged` | paged, paged×varlen, page-order permutation | pass |
| `test_fp8_splitk` | split-K 1/2/4/8, paged×split-K | pass |
| `test_packed_splitk_{decode,mixed}` | packed-M × split-K, decode-only + low-chunk mixed | pass |
| `test_flydsl_unified_attention` | consolidated prefill/decode/mixed × bf16/f16, 5D shuffled, 2 GB pool, split-K ragged | pass (20/20) |
| `test_unified_attention_routing` | end-to-end, prefill/chunked/decode/mixed + routing gates | pass |

Decode-specialized kernel — correctness across all split counts and the full
production pool: 209/209 production-pool cases (48 at the full 2 GB / 65536-block
pool), stage matrix 24/24, 44/44 pytest, prefill bit-identical to upstream.

Performance vs aiter `main`'s Triton `unified_attention` on the production 5D
shuffled fp8 KV-cache (`SHUFFLED_KV_CACHE_1`, the vLLM-v1 layout — the only
production-truth baseline; a linear-cache comparison overstates the win):

| Regime | Ratio | Notes |
|---|---|---|
| Pure prefill, weighted total (trace-v1) | +9.1% (Triton 7372ms -> FlyDSL 6699ms) | dominant M=32768 (~77% of prefill time) 1.06x; 25k-30k band ~0.98-1.0x |
| Pure prefill, small / KV-bound shapes | 1.87-2.26x | |
| Mixed batch, deep decode (ctx>=16384), dispatch split fires | 1.06-1.10x | |
| Mixed batch, shallow/underfilled | ceded to Triton | fallback confirmed served via the entry point |
| Decode, specialized kernel, GQA-16:1 fp8 shuffled paged | beats Triton on 6/8 of the batch×context grid | ~79-81% of the `pa` asm bandwidth ceiling at b=64; the 2 sub-parity cells (b=16/32, ctx=16384) are the ceded band, served by Triton |
| **Total attention (trace-weighted)** | **~+12%** (approximate) | |

## Submission Checklist

- [x] Ran the `flydsl-kernel-code-cleanup` skill over the changed kernels before human review.


aiter/ops/triton/attention/unified_attention.py
```diff
@@ -383,6 +388,51 @@ def unified_attention(
...
 
+    # FlyDSL fp8 paged backend (gfx950). Placed here because the layout unpack
+    # above has resolved everything it needs and nothing has launched yet. It
+    # returns None for any configuration it cannot serve, so unsupported shapes
+    # fall through to Triton below with no behaviour change.
+    if _FLYDSL_UNIFIED_ATTN_ARCH:
+        from aiter.ops.flydsl.unified_attention_kernels import (
+            flydsl_unified_attention,
+        )
+
```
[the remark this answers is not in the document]
johannes-graner: Fixed in 81c60d714. The gfx950 dispatch block now imports `flydsl_unified_attention` inside a `try/except ImportError` and skips FlyDSL (falling through to Triton) when the import fails. The adapter's `is_flydsl_available()` gate can only run after the module imports, so the import itself is what needed guarding.

aiter/ops/flydsl/unified_attention_kernels.py
```diff
@@ -0,0 +1,555 @@
...
+                # Provably full at single-pass; exact value only gates <target.
+                num_q_blocks = lower_bound
+            else:
+                seqlens_q = cu_seqlens_q[1:] - cu_seqlens_q[:-1]
+                num_q_blocks = int(
+                    ((seqlens_q + (block_q - 1)) // block_q).sum().item()
+                )
+        num_2d_prgms = num_kv_heads * num_q_blocks
+        if num_2d_prgms < _TARGET_NUM_PRGMS:
+            num_kv_splits = _split_count(num_2d_prgms)
```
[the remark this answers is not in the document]
johannes-graner: Fixed in 81c60d714. `_target_num_prgms` is now `@cache`-keyed on `q.device.index` and resolved at call time under that device (via `torch.cuda.device`), threaded through `_split_count` and the dispatch block, instead of the single import-time snapshot of the then-current device.

samremes:
# Agent review: aiter PR #4676 — fp8 unified attention for gfx950

**Reviewed:** https://github.com/ROCm/aiter/pull/4676
**Title:** [Triton/Gluon] [FlyDSL] fp8 unified attention for gfx950
**Base:** `origin/main` @ `a16ffe4db50f3e7698539d9bbf664bc7131ac664`
**Head:** `3f5d5e7c44ea6276e3044cb8794d267adda369fe`
**Dirty files in this worktree:** none
**Review date:** 2026-08-27
**Skills applied:** aiter-op-test, flydsl-kernel-authoring, flydsl-kernel-code-cleanup, flydsl-kernel-testing, flydsl-layout-algebra, aiter `aiter/aot/flydsl/README.md`

This document is a read-only review. Git SHAs were confirmed from the worktree ref files. Declared FlyDSL pin is `flydsl==0.3.1` in `requirements.txt`, `pyproject.toml`, and `setup.py`. Imported FlyDSL is version `0.3.1` at `/opt/venv/lib/python3.14/site-packages/flydsl`. No GPU test, compile, or benchmark was run.

## Verdict

Request changes before merge. The paged, varlen, split-K, GQA-pack, and 2 GB pool addressing in the kernels is carefully guarded, but the public `unified_attention()` path can abort on a non-causal call that FlyDSL already accepted and then ceded. Tests, AOT packaging, import-time env parsing, and new layout-ownership maps also need work before this is a safe default gfx950 backend.

## Review context

| Item | Value |
|---|---|
| Change class | new kernel/operation plus AITER integration, dispatch, and tests |
| Targets | gfx950, wave64; Triton fallback remains for other archs |
| Dtypes | QKV `float8_e4m3fn`; output bf16 or fp16; descales fp32; index tensors int32 |
| Declared FlyDSL pin | `flydsl==0.3.1` |
| Imported FlyDSL | `0.3.1` (`/opt/venv/lib/python3.14/site-packages/flydsl`) |
| Diff vs base | 15 files, +9130 / −2 |

Changed files:
- `README.md` (env-var table)
- `aiter/ops/flydsl/kernels/flash_attn_dualwave_common.py` (vendored dualwave helpers plus aiter paged/varlen/split-K/GQA/vectorized maps)
- `aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py` (decode-specialized kernel)
- `aiter/ops/flydsl/kernels/flash_attn_fp8_gfx950.py` (prefill dualwave launcher)
- `aiter/ops/flydsl/unified_attention_kernels.py` (public adapter and dispatch)
- `aiter/ops/triton/attention/unified_attention.py` (return-None hook)
- `op_tests/flydsl_tests/_common.py` and seven pytest modules (adapter, routing, vendored, paged, split-K, packed decode/mixed)

Sibling suite already present: `op_tests/triton_tests/attention/test_unified_attention.py`. FlyDSL AOT kinds in this checkout are MOE, GEMM, grouped MoE, and chunk-gdn-h only. Layout APIs used in the proposed replacements (`make_layout`, `make_view`, `crd2idx`, `idx2crd`, `slice`, `flat_divide`, `make_tiled_copy_tv`) exist on the pinned 0.3.1 release.

## Findings

Findings are ordered by severity. Each finding has a location, a trigger, a reason, and a proposed fix.

### P1 — Non-causal FlyDSL cede aborts in the Triton fallback

**Where:** `aiter/ops/triton/attention/unified_attention.py` after the FlyDSL hook (`assert causal`); `aiter/ops/flydsl/unified_attention_kernels.py` `_decode_dispatch_action` cede band and the underfill `return None` at the mixed single-call guard

**Trigger:** `unified_attention(..., causal=False)` on gfx950 fp8 paged input that `_supported` accepts, then a performance cede fires. Concrete cases: all-decode GQA 16:1 with `9 <= num_seqs <= 48` and `max_seqlen_k >= 8192`; mixed underfill (`max_seqlen_q > 1` and `num_seqs > 1`) after the mixed-split pre-check.

**Reason:** FlyDSL builds causal and non-causal binaries and does not decline on `causal`. Performance cedes return `None` so Triton can run. The Triton path then executes `assert causal, "Only causal attention is supported"`. A non-causal call that FlyDSL could have served, then ceded, raises `AssertionError` instead of computing. That is a new failure mode: short non-causal prefill can succeed, while a deeper or mid-batch non-causal decode crashes. `python -O` removes the assert, so the same cede would enter the causal Triton kernels and write a wrong mask.

**Proposed fix:** Keep one causal contract for the public wrapper. Either serve every `_supported` non-causal call on FlyDSL (do not cede non-causal), or reject non-causal at the public entry with a real `if not causal: raise` that still runs under `-O`, before any backend is chosen. A cede must not land on an assert that the adapter has already contradicted.

### P2 — `AITER_DECODE_KERNEL` parse can abort the public op at import

**Where:** `aiter/ops/flydsl/unified_attention_kernels.py` (`_USE_DECODE_KERNEL = bool(int(os.environ.get("AITER_DECODE_KERNEL", "1")))`)

**Trigger:** Process environment `AITER_DECODE_KERNEL` is set to a non-integer (`true`, `yes`, empty, `on`). First gfx950 `unified_attention()` call imports the adapter.

**Reason:** `int(...)` runs at module import. `unified_attention.py` catches only `ImportError` around that import. `ValueError` propagates and skips the Triton fallback. The sibling knob `AITER_UNIFIED_ATTN_MAX_KV_SPLITS` already ignores a bad value and keeps the default.

**Proposed fix:** Parse the flag with the same defensive path as `_env_max_kv_splits`, so a bad value leaves the default on and dispatch continues. Document the knob next to `AITER_UNIFIED_ATTN_MAX_KV_SPLITS`.

### P2 — Decode dummy tensors stay on the first CUDA device

**Where:** `aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py` `_dummy_f32` / `_dummy_i32` (`device="cuda"` cached in `_dummy_holder`); split-K `torch.empty(..., device="cuda")` when the caller omits workspace

**Trigger:** First decode launch on one device, later launch on another device in the same process. Split-K off still passes the cached dummy workspace and dummy cu_seqlens pointers into `_run_compiled`.

**Reason:** The adapter sets `torch.cuda.device(q.device.index)` around `mod()`, but the dummy cache is process-global and records whatever `cuda` meant on first use. A later call on another index reuses that allocation while the stream belongs to `q.device`.

**Proposed fix:** Allocate dummies and any fallback workspace on `q.device` (or the active device of that launch) and do not reuse a cross-device cache.

### P2 — Decode epilogue divides by `den` with no empty-KV guard

**Where:** `aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py` `inv = vd / den` in the direct store and in `combine_kernel`

**Trigger:** `seqused_k` is 0 for a sequence, or every launched split writes the `NEG` / zero init state. `_supported` does not reject `max_seqlen_k == 0`.

**Reason:** Prefill uses `_safe_l_inv` (`l_row > 0` else 0). Decode uses a raw divide. An empty reduction produces Inf or NaN in O. Host `plan_num_kv_splits` forces `S >= 1` even when `npages == 0`, so the empty launch is reachable.

**Proposed fix:** Match the prefill inverse: write a defined empty-row result when `den == 0`, and decide at the host boundary whether zero-length KV is a supported contract or a declined fallback.

### P2 — New decode and vectorized KV maps bypass layout ownership

**Where:** `flash_attn_fp8_decode_gfx950.py` Q row, shuffled K DMA, shuffled V operand, linear K/V, O store, and split-K workspace index; `flash_attn_dualwave_common.py` `DualwaveFp8KvGmemToLdsLoader.load_k` `KV_VECTORIZED` branch and `_stage_v_fp8_coalesced`

**Trigger:** Review of newly introduced regular tiled maps on FlyDSL 0.3.1. Per-page BufferDesc rebasing stays a hardware bound (int32 voffset / 2 GB pool). The within-page and Q/O maps are static-shape regular ownership.

**Reason:** Those maps are repeated `/`, `%`, and `row * stride + col` formulas. 0.3.1 exposes `make_layout`, `make_view`, `crd2idx`, `idx2crd`, and `slice`. Layouts would name the 5D shuffled K `(H_kv, D/16, PAGE, 16)` and V `(H_kv, PAGE/16, D, 16)` coordinate spaces that the comments already describe. Nearby legacy dualwave code does not justify extending the same style to a new body.

**Proposed fix:** Keep per-page rebasing explicit. Express the within-page and Q/O ownership with static layouts and runtime coordinates, for example:

- Q/O packed row: `make_layout((tokens, H, D), (H*D, D, 1))` then `crd2idx((token, h*16+n, d), layout)`.
- Shuffled K inside one page: `make_layout((H_kv, D/16, PAGE, 16), (D*PAGE, PAGE*16, 16, 1))`.
- Shuffled V inside one page: `make_layout((H_kv, PAGE/16, D, 16), (D*PAGE, D*16, 16, 1))`.
- Linear page: `make_layout((PAGE, H_kv, D), (H_kv*D, D, 1))`.

If a formula stays manual, localize it and state the missing 0.3.1 abstraction or the DMA/MMA ABI constraint.

### P2 — New tests are a second pytest suite, not the operation harness

**Where:** `op_tests/flydsl_tests/test_flydsl_unified_attention.py`, `test_unified_attention_routing.py`, `test_vendored_fp8_attention.py`, `test_fp8_cross_prefill.py`, `test_fp8_paged.py`, `test_fp8_splitk.py`, `test_packed_splitk_decode.py`, `test_packed_splitk_mixed.py`

**Trigger:** aiter-op-test plus one-suite-per-op, against the existing `op_tests/triton_tests/attention/test_unified_attention.py`

**Reason:** The operation already has a suite. This PR adds eight pytest modules under `op_tests/flydsl_tests/` instead of extending that suite or adding `op_tests/test_unified_attention.py`. There is no `@benchmark` / `run_perftest` / markdown `__main__` table. Arch gating is `pytestmark` / `skipif`, not `get_gfx()` in `main()`. Pass/fail uses `assert_attn_close` (relative error plus cosine plus optional bad-row count), not `checkAllclose`. Direct builder tests are useful for paged and split-K internals, but they must not become a second public-op harness.

**Proposed fix:** Keep one canonical suite for `unified_attention`. Put public-wrapper routing and numeric cases next to the existing Triton suite (or in `op_tests/test_unified_attention.py` with the aiter-op-test harness). Use `checkAllclose` for the kernel-versus-oracle gate. Leave at most one focused kernel-direct module for builder-only contracts (sinks plus split-K refusal, 2 GB pool rebasing) that the public wrapper cannot name.

### P2 — New FlyDSL attention kernels are absent from AOT discovery

**Where:** `aiter/aot/flydsl/common.py` `OpKind`; `aiter/aot/flydsl/README.md` module table; no attention collector added in this diff

**Trigger:** Wheel prebuild / `python -m aiter.aot.flydsl.*` as used for GEMM and MoE. First gfx950 fp8 `unified_attention()` after install.

**Reason:** Dispatch can select this backend with no caller flag. AOT in this repo compiles from CSV job kinds. Attention is not a kind, so the new binaries are JIT-only. First-token compile on the serving path is a deployment defect for a default backend, not an optional extra.

**Proposed fix:** Add AOT job collection for the shipped gfx950 configs (prefill dualwave and decode, linear and vectorized, the split counts the host actually launches), or document that this backend is JIT-only and is not a prebuilt default.

### P3 — README omits `AITER_DECODE_KERNEL`

**Where:** `README.md` environment-variable table; `unified_attention_kernels.py` module docstring

**Trigger:** Operator looks up runtime knobs in the README.

**Reason:** The table documents `AITER_UNIFIED_ATTN_MAX_KV_SPLITS` only. `AITER_DECODE_KERNEL` is a live master switch parsed at import.

**Proposed fix:** Add the flag to the same table, with default `1` and the meaning of `0`.

### P3 — Mixed-split decode sub-call ignores `None`

**Where:** `unified_attention_kernels.py` mixed-split site: prefill `None` returns to Triton; decode `_sub(...)` return is dropped

**Trigger:** Decode half of a taken split fails `_supported` on the sliced views, or a future cede is added without `_from_split`.

**Reason:** Comments state that the decode half never declines. Prefill decline is handled so Triton can rewrite the full buffer. Decode decline would leave those rows unwritten while the adapter returns `out`. No current `_supported` slice failure was traced, so this is a contract hole rather than a proven bug.

**Proposed fix:** Treat a decode `None` like prefill `None` (cede the whole batch), or fail loudly. Do not return a partially written `out`.

## Verification gaps

- No GPU correctness, arch sweep, or benchmark was run on this worktree.
- FlyDSL pin was checked from files and from the installed `0.3.1` package. The local `~/dev/FlyDSL` clone was not used as authority.
- AOT compile and run-only/cache-hit were not executed.
- FakeTensor / `torch.compile` meta for `unified_attention()` was not executed. The adapter can allocate (`_cu_seqlens_kv`, workspace) and launch after `_supported`.
- Non-causal cede through the public wrapper was not executed; the P1 path is from source.
- Empty-KV decode (`seqused_k == 0`) was not executed.
- Multi-device dummy-tensor reuse was not executed.

## Recommended minimum tests

- Public `unified_attention(..., causal=False)` on a decode cede-band shape and on a mixed underfill shape. Expect a defined result or a real exception, not `assert causal`.
- `AITER_DECODE_KERNEL=true` must not raise on import of the adapter.
- `seqused_k == 0` on the decode route: defined output or a decline to Triton.
- One public-wrapper numeric case for shuffled 5D plus a large page id, using `checkAllclose` against `ref_paged_attn`.

README.md
```diff
@@ -155,3 +155,12 @@ pip install -r requirements-triton-comms.txt
 ```
 
 For more details, see [docs/triton_comms.md](docs/triton_comms.md).
+
+## Environment variables
```
samremes: Probably not the right place for this here on the main README.md.

op_tests/flydsl_tests/_common.py
```diff
@@ -0,0 +1,89 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Shared helpers for the fp8 unified-attention test files.
```
samremes: I think there is too many split just for unified attention here. Try to reduce them to maybe a single file.

op_tests/test_unified_attention_routing.py
```diff
@@ -0,0 +1,960 @@
...
+adapter's module docstring), so the A/B here patches the module-level arch
+constant instead.
+"""
+
+import math
+import os
+import sys
+from unittest import mock
+
+import pytest
```
samremes: Remove pytest. A single main function should run everything. Check aiter-op-test skill.

samremes (dismissed):
Request changes.

**Correction to the first version of this review:** I withdraw the blocking finding about resolving and caching architecture per `q.device`; please disregard it. `DEVICE_ARCH = arch_info.get_arch()` already exists on `main` in `aiter/ops/triton/attention/unified_attention.py` and gates `WARP_SIZE` and the Gluon backend, so `_FLYDSL_UNIFIED_ATTN_ARCH = DEVICE_ARCH == "gfx950"` is ordinary use of that shared helper. Process-level caching of AITER's shared architecture helpers (`arch_info.get_arch()`, `get_gfx_runtime()`, `get_gfx()`) is the intended dispatch mechanism, not a per-kernel defect, and this PR should not carry its own device-property arch probe, extra validation layer, or heterogeneous-host regression test. If a helper is ever wrong, the fix belongs in `aiter.jit.utils.chip_info` or the shared Triton `arch_info`. The adapter's per-device `torch.cuda.get_device_properties(q.device).multi_processor_count < GFX950_CUS` gate is fine as written.

Two points remain.

1. The tests need one canonical `op_tests/test_unified_attention.py` AITER harness. `.github/scripts/split_tests.sh:52-55` scans only top-level `op_tests` for AITER tests. It therefore runs `op_tests/test_unified_attention_routing.py`, but it does not run the seven new modules under `op_tests/flydsl_tests/`. The routing script is also a pytest launcher, imports the Triton suite at lines 28-38, and has no `@benchmark`, `run_perftest`, `checkAllclose`, TFLOPS/TB/s fields, argparse sweep, or markdown summary.

   Please leave the existing Triton suite independent. Move the FlyDSL coverage into a top-level canonical suite with:
   - an independent fp32 paged-attention oracle in that file or `op_tests/_unified_attention_test_utils.py`;
   - one public-wrapper candidate that uses `backend="flydsl"` so fallback cannot make a positive case pass;
   - one direct-launch function for split-K, paged-address, and builder-only boundaries;
   - positive `get_gfx()` gating in `main()` and a clean FlyDSL-unavailable skip before importing FlyDSL modules;
   - `checkAllclose` in fp32, `run_perftest`, per-candidate `us`/TFLOPS/TB/s/`err`, `itertools.product`, and final markdown tables.

   Merge the cases from all seven nested modules and the routing script into that suite. Delete those eight test files and `_common.py` after moving the useful builders. Do not import anything from `op_tests/triton_tests`; the Triton tests are intentionally independent of the rest of aiter and should stay that way. Top-level placement satisfies AITER CI discovery, while the Triton tree remains collected separately by the Triton job.

2. The new 80-variant AOT family has compile-only wiring but no runtime cache-hit proof (`aiter/aot/flydsl/unified_attention.py:68-139,188-379`). Add one focused run-only/cache-hit check that executes representative prefill and decode calls against the independent oracle.

I would re-review after the test consolidation.

aiter/ops/flydsl/unified_attention_kernels.py
```diff
@@ -0,0 +1,1207 @@
...
+__all__ = ["flydsl_unified_attention"]
+
+
+@lru_cache(maxsize=1)
+def _is_flydsl_installed() -> bool:
+    return importlib.util.find_spec("flydsl") is not None
+
+
+@cache
+def is_flydsl_available(device_index: int) -> bool:
```
samremes, line 59: is there some common helper for this?

aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py
```diff
@@ -0,0 +1,1005 @@
...
+from aiter.ops.flydsl.kernels.tensor_shim import _run_compiled, _to_raw, ptr_arg
+
+FP8 = fx.Float8E4M3FN
+WARP = 64
+PAGE = 64  # block/page size, structural (matches the 5D shuffled cache)
+HEAD_DIM = 128
+BLOCK_M = 16  # one GQA group = MMA M-dim
+VEC = 16  # 5D-shuffled vectorization width
+_LOG2E = _math.log2(_math.e)
+GFX950_CUS = 256  # MI355X compute units
```
samremes: There should be a helper to query it? This won't be true for MI350P.

aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py
```diff
@@ -0,0 +1,1005 @@
...
+size 64), query_len=1, causal (decode => attend all kv < kv_len), varlen, and
+bf16/fp16 output.
+
+This is a NEW body, not a trait flip of ``flash_attn_fp8_gfx950.py``: the 16-row
+wave mapping and the register-side QK->PV repack cannot be expressed as
+const_expr branches of the prefill kernel. It is deliberately decode-local and
+does NOT call the welded dualwave loaders (which are hardwired to
+num_waves=8 / rows_per_wave=32); it carries its own small decode traits.
+
+Proven cores, measured empirically:
```
samremes: Maybe some of these comments should be written more direct, and avoid a longer story of what was measured, etc. Just simple and stand-alone.

aiter/ops/triton/attention/unified_attention.py
```diff
@@ -153,21 +165,26 @@ def unified_attention(
     shuffled_kv_cache: bool = False,
     skip_reduce: bool = False,
     # backend
-    backend: str | None = None,  # "triton" | "gluon"
+    backend: str | None = None,  # "triton" | "gluon" | "flydsl"
```
samremes: I wonder if the unified attention entry point should be separate, and then dispatch to either FlyDSL or to the triton/gluon path, rather than bring FlyDSL logic under ops/triton/

samremes (commented):
A few more comments. The tests are looking now much more leaner.

aiter/ops/flydsl/unified_attention_kernels.py
johannes-graner: Yes, fixed.

aiter/ops/flydsl/kernels/flash_attn_fp8_decode_gfx950.py
johannes-graner: Made a pass through the comments, should be better now.

aiter/ops/triton/attention/unified_attention.py
johannes-graner: Agreed, now a common entrypoint exists at `aiter/ops/unified_attention.py`, which delegates to either the existing triton entrypoint, or a new `aiter/ops/flydsl/unified_attention.py` entrypoint.

PR #4714, opened 2026-08-12
[MI355] add  8wave pipeline to a8w8 bpreshuffle gemm

solinzby1:
## Submission Checklist

### DIFF
--- a/aiter/aot/flydsl/gemm.py
+++ b/aiter/aot/flydsl/gemm.py
@@ -13,6 +13,7 @@ way as runtime JIT config lookup.
 Supported kernel families:
   - ``flydsl_gemm2_*``                        split-K HGEMM kernels
   - ``flydsl_bpreshuflle_*``                  a8w8 preshuffle GEMM kernels
+  - ``flydsl_bpreshuffle_8w_*``               gfx950 8-wave a8w8 ptpc GEMM kernels
   - ``flydsl_bpreshuffle_wmma_*``             gfx1250 a8w8 ptpc GEMM kernels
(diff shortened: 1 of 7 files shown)


aiter/aot/flydsl/gemm.py
```diff
@@ -153,6 +158,11 @@ def parse_csv(csv_path: str):
 
             if kernel_name.startswith("flydsl_bpreshuflle_"):
                 params = _parse_preshuffle_kernel_name(kernel_name)
+            elif kernel_name.startswith("flydsl_bpreshuffle_8w_"):
```
XiaobingSuper, line 161: It will not catch this condition and will catch the first condition. Need to change the check order.

XiaobingSuper: The names are different;

aiter/ops/flydsl/kernels/gemm_a8w8_8wave.py
```diff
@@ -0,0 +1,580 @@
...
+            offsets.append(r * K + c)
+    return offsets
+
+
+def pack_i32x4_i32x8(lo, hi):
+    # Pack two i32x4 as one i32x8
+    return lo.shuffle(hi, list(range(8)))
+
+
+def wait_barrier(count):
```
yadaish, line 82: why we use inline assembly to implement vmcount?

solinzby1: Currently a single asm block emits both s_waitcnt vmcnt(N) and s_barrier，, marked has_side_effects=True, so the compiler cannot insert any instruction between them. Splitting them into two separate rocdl ops would lose that guarantee.

valarLip (approved):

PR #4729, opened 2026-08-13
[module_mla_metadata*] refactor and remove torch

amd-ruitang3:
## Summary

Removes the torch/ATen/c10 dependency from the `csrc/kernels/mla/metadata.cu` translation unit and every header it pulls in. This single TU is compiled into **three** JIT modules — `module_mla_metadata`, `module_pa_metadata`, `module_ps_metadata` — each of which previously had to parse the entire torch/ATen header tree on every JIT build. Dropping torch is the biggest bang-for-buck item for these three modules' compile time.

Follows the established **Strategy A** de-torch pattern: the C++ entry points take POD `aiter_tensor_t`, and the Python side uses `@compile_ops(..., develop=True)`, which auto-converts `torch.Tensor` → `aiter_tensor_t` and sets the HIP stream.

## Build-time improvement (the headline)

Clean before/after on gfx942 (ROCm 7.2.3), building `module_mla_metadata` from scratch:

| | JIT build time |
|---|---|
| Before (torch) | **120.4 s** |
| After (torch-free) | **86.5 s** |
| **Saved** | **-33.9 s (~28%)** |

Both builds produce byte-identical metadata (`test_metadata`: parallel planner matches serial reference ✓). Because the same TU compiles all three modules, the saving repeats for each of `module_mla_metadata` / `module_pa_metadata` / `module_ps_metadata`.

## Complete torch removal

`torch` was only `#include`d directly in `v1_comm.cuh`; the other headers inherited it transitively. After this change, a grep for any torch/ATen residue across all touched files is **clean**:

```
grep -rnE 'torch|at::|ATen|c10|TORCH_CHECK|ScalarType|from_blob|TensorOptions|OptionalHIPGuard' \
  csrc/kernels/mla/metadata.cu csrc/kernels/mla/metadata/*.cuh \
  csrc/include/{mla_metadata,pa,ps}.h csrc/pybind/{mla,pa,ps}_metadata_pybind.cu
  -> (no matches)
```

Key mechanical substitutions:
- `torch::Tensor` → `aiter_tensor_t`; `x.data_ptr<T>()` → `static_cast<T*>(x.data_ptr())`
- `at::ScalarType` → `AiterDtype`; the fp8/bf16 dispatch matrix rewritten against `AITER_DTYPE_*`
- `TORCH_CHECK` → `AITER_CHECK`; `at::hip::getCurrentHIPStream()` → `aiter::getCurrentHIPStream()`
- `OptionalHIPGuardMasqueradingAsCUDA` → `HipDeviceGuard`
- `v1_2_host` / `ps` host planning: `.to(CPU)` + `from_blob`/`copy_` → direct `hipMemcpy` (D2H read, H2D write-back)
- pybind TUs drop `<torch/extension.h>`, add `aiter_stream.h` + `AITER_SET_STREAM_PYBIND`

## `get_mla_metadata_v1_no_redundant` → pure Python

This was the only entry point returning a **dynamically-sized `std::vector<torch::Tensor>`**, which is incompatible with the `develop=True` out-param convention. It was also the sole user of `v1_1_host.cuh` and has **zero in-repo callers**. It is now reimplemented as a 1:1 pure-Python port of the host bin-packing algorithm (public API unchanged), which lets `metadata.cu` drop torch entirely. `v1_1_host.cuh` is deleted.

## Validation

- **torch-residue grep**: clean (see above)
- **`op_tests/test_metadata.py`**: `module_mla_metadata` builds; parallel planner matches serial reference ✓
- **`op_tests/test_pa_ps.py`**: `module_pa_metadata` + `module_ps_metadata` build; `err == 0` across the full sweep, except the expected tiny fp8 quantization noise (`0.039 < 0.05` tolerance) on the smallest (q=7) V40 cases only — unrelated to metadata, which is integer-exact
- **`no_redundant` pure-Python port**: 13/13 randomized invariant configs pass (shapes/dtypes, monotonic indptrs, exact causal-kv coverage, `kv_offset` consistency)
- **black**: passes on `aiter/ops/attention.py`

## Notes for integrators

The `get_ps_metadata_v1` path is host-in-the-loop planning (CPU reads seqlens, computes, writes back) and uses synchronous `hipMemcpy`. Like the previous torch version (`.to(CPU)` + `copy_`), it cannot be captured into a CUDA/HIP graph and must be called outside the capture region. The `get_mla_metadata_v1` (fast_mode) and `get_pa_metadata_v1` paths are pure device kernels with no host sync and remain graph-capture friendly.


ruanjm (commented):
LGTM. Besides, I think we can remove v1_1_host entirely if there is no user.

junhaha666 (approved):

PR #4737, opened 2026-08-13
[CI] PR auto tag

Boss2002n:


.github/workflows/pr-title-tags.yaml
```diff
@@ -0,0 +1,119 @@
...
+                               || f.startsWith('aiter/ops/triton/configs/');
+
+            // Doc files never contribute component tags: a docs-only PR gets
+            // [Docs], and a code PR that also touches a README is tagged by
+            // its code only.
+            const code = files.filter(f => !isDoc(f));
+
+            // ---- compute canonical tags from the diff, priority order ----
+            const tags = [];
+            if (code.some(isTriton))           tags.push('Triton/Gluon');
```
gyohuangxin, line 99: It does not match the existing PR title guidance in[ `CONTRIBUTE.md`.](https://github.com/ROCm/aiter/blob/main/CONTRIBUTE.md#pr-title-format) . Please keep they are synced.

.github/workflows/pr-title-tags.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+            // ---- changed files (paginated) ----
+            const files = (await github.paginate(github.rest.pulls.listFiles, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: pr.number,
+              per_page: 100,
+            })).map(f => f.filename);
+
+            // ---- path predicates ----
```
gyohuangxin, line 62: It may miss some existing component paths. 
For example, `aiter/aot/triton/`, `aiter/aot/flydsl/`, `aiter/utility/triton/`, `op_tests/op_benchmarks/hip/`, `csrc/include/ck_tile/`, `csrc/include/opus/`, and `aiter/ops/pa_sparse_prefill_opus.py` do not seem to be covered by the current predicates. One concrete case: changes under `csrc/include/ck_tile/` would currently fall through to `isHip` because they start with `csrc/`, even though they look CK-related. 

@valarLip @zufayu  Please check whether I missed other paths?

.github/workflows/pr-title-tags.yaml
```diff
@@ -0,0 +1,119 @@
...
+# component namespace ([fix], [Perf], [MLA], [gfx1250], ...) are never
+# touched. To opt a PR out entirely, add the 'no-auto-title' label.
+#
+# pull_request_target is used so the default token can edit titles on fork
+# PRs as well. This is safe here because the job never checks out or executes
+# PR code — it only reads the changed-file list from the API.
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, synchronize, reopened]
```
gyohuangxin: Since `no-auto-title` is the opt-out mechanism, this workflow should probably listen for `labeled` and `unlabeled` as well. Otherwise, adding `no-auto-title` after the workflow has already retitled a PR will not take effect until the next synchronize/reopen/ready_for_review event, and removing the label also will not immediately re-sync the title.

gyohuangxin: Could we also limit this workflow to PRs targeting `main`, the same way `pr-welcome-comment.yaml` does?

gyohuangxin (changes requested):

gyohuangxin (approved):

PR #4742, opened 2026-08-14
fix(rope): tolerate missing original_max_position_embeddings

lizamd:
## Motivation

`get_rope()` raises `KeyError` and every engine dies at init when a model's
`rope_scaling` lacks `original_max_position_embeddings`:

```
File "aiter/rotary_embedding.py", line 1872, in get_rope
    original_max_position = rope_scaling["original_max_position_embeddings"]
KeyError: 'original_max_position_embeddings'
```

Root cause is a transformers v4 -> v5 change. On v4, unscaled models had
`config.rope_scaling = None`. On v5 it aliases `rope_parameters` and is always
populated, e.g. for GLM-4.7-Flash:

```python
{'rope_theta': 1000000, 'partial_rotary_factor': 1.0, 'rope_type': 'default'}
```

sglang's `DeepseekV2AttentionMLA` treats a truthy `rope_scaling` as "scaling is
configured" and sets `rope_type = "deepseek_yarn"`, so an unscaled MLA model
reaches the yarn branch without the keys that branch requires.

sglang already handles this in
`sglang/srt/layers/rotary_embedding/factory.py::_get_rope_param` (warn + default).
aiter's vendored `get_rope` never received the equivalent, so on ROCm the same
config is a hard failure rather than a warning.

## Technical Details

Adds `_get_rope_param(rope_scaling, key, default, scaling_type)`, mirroring
sglang's helper: return the key if present, otherwise log a warning and fall
back to `max_position`. Applied at all four call sites that read
`original_max_position_embeddings` (`llama3`, `yarn`, `deepseek_yarn`,
`su`/`longrope`) for parity with sglang.

The warning matters: silently defaulting would hide a genuinely mis-specified
config for a model that really does use scaling.

Correctness for the unscaled case: with `factor` absent (defaults to 1.0) and
`original_max_position == max_position`, YaRN degenerates to standard RoPE — no
interpolation, and the MLA attention scaling is unchanged. Verified:

```
compute_mla_mscale_scaling({'rope_type':'deepseek_yarn', ...}, 0.1234) -> 0.1234
```

## Test Plan

- GLM-4.7-Flash (MLA, `rope_type='default'`) on 8x MI355X, ROCm 7.2,
  transformers 5.12.1, sglang 0.5.17, `SGLANG_USE_AITER=1`
- Start all 8 sglang engines, then run an end-to-end GRPO training loop
  (rollout -> backward -> weight sync)

## Test Result

Before: all 8 engines abort at init with the `KeyError` above.

After: engines initialize (`max_total_num_tokens=2870263`, `context_len=202752`)
and the training loop runs end to end, with the expected warning emitted once
per engine. All aiter kernels remain enabled — `SGLANG_USE_AITER=1` throughout;
no fallback to non-aiter paths was needed.

## Submission Checklist

- [x] Change is limited to the missing-key fallback; no kernel or numerics
      changes for configs that already carry the key
- [x] Verified on MI355X (gfx950)
- [x] Mirrors existing sglang behavior rather than introducing new semantics


aiter/rotary_embedding.py
```diff
@@ -43,6 +44,29 @@
 AITER_ROPE_FUSED_QKNORM = int(os.environ.get("AITER_ROPE_FUSED_QKNORM", "0")) == 1
 
 
+logger = logging.getLogger(__name__)
```
junhaha666: Please use `from aiter import logger` instead of this.

aiter/rotary_embedding.py
```diff
@@ -1782,7 +1806,9 @@ def get_rope(
         if scaling_type == "llama3":
             low_freq_factor = rope_scaling["low_freq_factor"]
             high_freq_factor = rope_scaling["high_freq_factor"]
-            original_max_position = rope_scaling["original_max_position_embeddings"]
+            original_max_position = _get_rope_param(
```
junhaha666, line 1805: Move it outside the "if-else" judgment statement, and the code seems to become more concise.

junhaha666 (approved):

aiter/rotary_embedding.py
lizamd: fixed

aiter/rotary_embedding.py
lizamd: fixed

PR #4748, opened 2026-08-14
[ASM] [HIP] [gfx1250] fix(asm gemm): add a_preshuffle=0 f4gemm & f8gemm, fix corner case support and enhance ut

dbyoung18:
Currently, gfx1250 asm f4gemm(a4w4) and f8gemm(a8w8, a8w4) are both for **benchmark-only**. The support matrix for each is listed as below.
*  F4GEMM Support Matrix
image

*  F8GEMM Support Matrix
image
 
You can use the below cmds to reproduce perf
 ```
# clean aiter jit cache
rm aiter/jit/build aiter/jit/*.so
# measure f4gemm
python op_tests/test_f4gemm.py
# measure f8gemm
python op_tests/test_mxfp8fp4gemm.py
```
Use `-mnk a,b,c` to use specific shape, `--apreshuffle 0` to disable pre-shuffle A data.

### CODE REVIEW

**Comment by valarLip** [comment]
## Code review: `dbyoung/gemm_enhance` (6 commits, 33 files)

Reviewed via 10 parallel finder angles + adversarial verification + a gap sweep. 15 confirmed findings, ranked by severity.

### Blocking — silent feature regression

**1. `aiter/ops/gemm_op_a4w4.py:130` — the entire gfx1250 F4GEMM dispatch branch is gone**

Public `gemm_a4w4` now falls through to a gfx942/gfx950-only path that has zero gfx1250 kernels registered:

```
gemm_a4w4(A, B, A_scale, B_scale, dtype=bf16, bpreshuffle=True)  # on gfx1250
  -> get_GEMM_config() returns None        (a4w4_blockscale_tuned_gemm.csv has only gfx950 rows)
  -> gemm_a4w4_asm(kernelName="")
  -> asm_gemm_a4w4.cu get_cfg() -> &cfg_f4gemm_bf16_per1x32Fp4
     (sourced only from hsa/gfx942/f4gemm + hsa/gfx950/f4gemm; hsa/gfx1250/f4gemm holds a different map)
  -> the `el.first.find(arch_id) != 0` filter skips every row
  -> RuntimeError: "get_heuristic_kernel: cannot get heuristic kernel!"
```

This is the exact API called by vLLM (`model_executor/kernels/linear/mxfp4/aiter.py`), SGLang (`quantization/mxfp4.py`) and ATOM (`atom/model_ops/linear.py`). The PR also deletes the only regression test for it (the `else: # bf16 / res = aiter.gemm_a4w4(...)` contract check in `test_f4gemm` func mode), and `op_tests/test_gemm_a4w4.py` early-returns unless `get_gfx() == gfx950` — so nothing in CI catches this. Effectively a silent revert of #3996.

**2. `aiter/ops/gemm_op_a4w4.py:115` — `apreshuffle`, `global_A_scale`, `global_B_scale` are now dead parameters**

All three are still declared in the `torch.library` schema (both `Tensor?`) but are unreferenced in either body — grep finds them only in the signatures at lines 91-93 and 115-117. Pre-PR they were consumed exactly once, by the deleted gfx1250 `_f4gemm_asm_dispatch` call that routed to `gemm_nvfp4_asm`. A caller passing `global_A_scale=gA, global_B_scale=gB` gets no error and no warning; the alphas are never multiplied in. The bias/alpha/beta ignore-warning inside `_f4gemm_asm_dispatch` is likewise unreachable from the public API now. Please either restore the dispatch or delete the dead parameters in the same commit.

### Dispatch gate holes

**3. `csrc/py_itfs_cu/asm_f4gemm.cu:112` (and `asm_mxfp8fp4gemm.cu:98`) — exact-tiling and cluster-divisibility guards replaced with file-global alignment constants**

Deleted:

```cpp
// Persistent/cluster shaders don't mask partial tiles, so the problem must tile both dims exactly.
if((N % cfg.tile_n) != 0 || (M % cfg.tile_m) != 0) continue;
if((cl_x > 1 && (tg_num_N % cl_x) != 0) || (cl_y > 1 && (tg_num_M % cl_y) != 0)) continue;
```

All 8 gfx1250 f4gemm CSV rows are 256x256 cluster 4x4, so `(1, 1280, 8192)` and `(127, 1280, 8192)` now dispatch a 4x4-clustered 256-row tile for a 1-row problem. Every `.co` was rebuilt here (all 17 md5s changed) so masking may well have been added — but the PR doesn't state that, the replacement comment only claims the *launch* is shape-independent (a different property), and coverage is absent: across the 68 new `FUNC_SHAPES` every N is a multiple of 256 (= `tile_n`) and every K a multiple of 128, so the partial-N-tile path this enables is never executed. If any current or future `.co` in the CSV does not mask, an N=1296 fp8 run writes a trailing 256-wide tile into a 1296-wide output. Could you confirm the masking behavior in the PR description, and add a non-tiling shape to the sweep?

**4. `asm_f4gemm.cu:279` (and `asm_mxfp8fp4gemm.cu:216`) — explicit `kernelName` bypasses all shape validation**

M/N/K alignment now lives only inside `get_heuristic_kernel`, which is skipped when `selectedName` is non-empty, and the launch-body `AITER_CHECK(tiles_x >= cluster_x && (tiles_x % cluster_x) == 0, ...)` / `tiles_y` pair that used to run on **both** paths was deleted in the same hunk. So:

```python
_gemm_mxfp4_asm(A, B, sA, sB,
                kernelName="_ZN5aiter45f4gemm_bf16_mxfp4_ABpreShuffle_256x256_4x4_psE")  # N=100
```

passes with only `AITER_CHECK(Kdim % scale_block == 0)` surviving — and for nvfp4 `scale_block` is 16, so even `K % 32 != 0` gets through. Both test files ship a new `--knl-name` flag that drives exactly this path.

**5. `asm_mxfp8fp4gemm.cu:231` — new `outtype` CSV column is never read (grep: 0 occurrences) and is missing from the post-selection guard**

The guard is `AITER_CHECK(cfg.b_intype == b_intype && cfg.a_preshuffle == a_preshuffle, ...)` — no `outtype`. The file header says fp8 out "is planned; the sweep axis and dispatch seam below are already in place." The moment an fp8-output row with the same `(b_intype, a_preshuffle)` lands in the CSV, `get_heuristic_kernel` ranks it against the bf16 rows purely on tile efficiency and can pick it for a bf16 request; the `out->dtype() == AITER_DTYPE_bf16` check still passes, the host ships `stride_d = Ndim*2` to a kernel writing 1-byte elements, and the kernarg size is identical (76B) so HIP can't catch it — device-side OOB write. The sibling file already handles this, with the hazard spelled out:

```cpp
// asm_f4gemm.cu:297-301
// outtype MUST match: a mismatched .co keeps the same kernarg size (HIP won't catch it)
// but sizes stride_d / the output buffer for a different element width -> device-side OOB write.
```

Same class of dead data: `cfg.scale` (the new `f4gemm.csv` column) is also parsed and never read.

### Silently wrong results

**6. `aiter/ops/asm/f4gemm.py:74` — `_alloc_f4gemm_out_scale` dropped its asserts for silent ceil-rounding over `torch.empty`**

Removed: `assert N % MXFP8_OUT_SCALE_BLOCK == 0` and `assert M % 64 == 0`. The only remaining N%128 gate is `_support_reason` inside `op_tests/test_f4gemm.py:193` — i.e. in the test, not in the op, and not in the dispatcher (`F4GEMM_N_ALIGN` is 16). So `_gemm_a4w4o8_asm(A[64,K//2], B[1296,K//2], sA, sB)` passes every op-level check, allocates `torch.empty((64, 11), fp8_e8m0)`, the kernel fills the 10 complete 128-wide blocks, and **column 10 is whatever was in device memory**. `unpack_mxfp8_out_scale` reshapes it happily and the caller dequantises with a garbage exponent — no exception. Note the PR's own new support-matrix table (`asm_f4gemm.cu:13-16`, copied verbatim into `test_f4gemm.py`) states `FP8 ... N %16==0`, which is now the only normative statement of the contract and it is wrong.

**7. `asm_f4gemm.cu:215` — scale row-stride pads K to 128 while the gate, the support matrix and the tests all still say K%32**

`stride_sa = ((Kdim + 127) / 128) * (128 / scale_block)`, with `Kdim` the real K (`A->size(1)*2`). For K=8224 (K%32==0, K%128!=0) the host ships 65 full 128-K super-columns of scale, but `stride_a = stride_b = Kdim/2 = 4112` bytes = only 64.25 super-columns of packed fp4 — and `shuffle_weight_f4` only asserts `kp%16==0`, it never pads K to 128. Whether the shader clamps its data loads by the K kernarg isn't verifiable from the PR and isn't tested: every K in both test files (8192, 16384, 53248, 1024, 28672, 768) is a multiple of 128, which is exactly the case where the new formula equals the old `Kdim / scale_block`.

Separately, an external caller following the docstring `ScaleA:[M, K/32] e8m0 (shuffled)` and passing an unpadded scale gets `stride_sa` > the actual row length, so row *r* is read at `r*260` in a 257-wide buffer — progressively skewed scales, then OOB. Neither the `.cu` nor the Python wrapper validates `ScaleA->size(1)` against `stride_sa`. Suggest either tightening `F4GEMM_K_ALIGN` to 128 or adding a K%128!=0 test.

**8. `aiter/ops/shuffle.py:315` — `shuffle_scale_f4` pads e8m0 rows with `0x00`, its sibling pads the same dimension with `0x7F`**

```python
# shuffle.py:71-74  (shuffle_mxfp8fp4_scale)
# ... the neutral e8m0 scale 0x7F (2^0 == 1.0), matching the POC host's ScaleA_M = (M + 31) & ~31 padding
F.pad(s, (0, 0, 0, pad), value=0x7F)

# shuffle.py:309-315  (shuffle_scale_f4)
# ... pad rows up to 32 and cols (K/block) up to tile_major (=128/block) with 0,
#     matching the POC host ScaleA row/stride padding
F.pad(s, (0, col_pad, 0, row_pad), value=0)
```

Both comments claim to match "the POC host ScaleA padding"; at most one can be right. `fp4_utils.e8m0_to_f32` maps byte 0 to `0x00400000` = 2^-127, so the new comment's implied "zero" is also factually wrong. Reachable today: `FUNC_SHAPES` `(1,1280,8192)` and `(127,1280,8192)` both take the `row_pad` path. Note the single `F.pad` also covers `col_pad`, where 0 **is** correct — one literal can't justify both paddings. Please confirm which constant the f4gemm `.co` expects and make the two functions consistent (or document why they differ).

### Performance

**9. `asm_mxfp8fp4gemm.cu:204` — `b_intype` converted from `int` to `std::string` on the per-call hot path**

The heuristic cache key becomes `std::tuple<int,int,int,std::string,int>`. Every `mxfp8_mxfp8_gemm_asm` / `mxfp8_mxfp4_gemm_asm` call now constructs a temporary `std::string` at the entrypoint, does two string comparisons in the `AITER_CHECK`, builds a tuple containing a `std::string` (heap alloc), runs `std::hash<std::string>` for the lookup, then compares `cfg.b_intype` again. On the decode-shaped `PERF_SHAPE` (M=2) the kernel is ~10us, so a per-call malloc plus string hash is a measurable host-side regression on precisely the shape this PR benchmarks. The sibling file already solved this (pre-existing on main, untouched here):

```cpp
// asm_f4gemm.cu:262-263
// All-int cache key: the hot lookup hashes ints, not std::strings.
const int intype_id = (intype == "nvfp4") ? 1 : 0;
```

Also worth noting: the mxfp8fp4 `DictHash` still XOR-combines `std::hash<int>` (identity on libstdc++), so M/N swaps collide.

### Tests

**10. `op_tests/test_f4gemm.py:62` — `FUNC_SHAPES` replaced 14 small correctness shapes with 68 perf-sized ones; the unchunked fp32 reference OOMs and aborts the whole sweep**

`python op_tests/test_f4gemm.py --mode func` reaches `(32768, 106496, 16384)`. `run_torch_mxfp4` materialises `mxfp4_to_f32(wq)` [106496,16384] f32 = 6.98 GiB (plus a 13.9 GiB int64 index tensor inside `mxfp4_in_f32[x.long()]`), `mxfp4_to_f32(xq)` = 2.1 GiB, two `repeat_interleave` copies of the same sizes, and the [32768,106496] f32 product = 13.96 GiB; for `outtype=fp8`, `_quant_mxfp8_blockN` / `_dequant_mxfp8_blockN` add two more full-size f32 tensors. Peak > 60 GiB.

This happens at `inp, ref_f32 = prep(...)` (line 344), **before** the `try: out, us = run_perftest(...)` at line 436 — so `torch.OutOfMemoryError` propagates out of `test_gemm` and kills the sweep rather than being marked "not support". `(32768,16384,53248)`, `(32768,18432,16384)` and `(16384,16384,16384)` are 25-40 GiB by the same accounting. `bench_init` gained row-chunking in this very PR to avoid a 64 GiB f32 staging buffer, but the reference that consumes those buffers was left unchunked while its shapes grew ~1000x.

The list also has a duplicate: `(64, 1280, 8192)` appears at both line 85 and line 90 (68 entries, 67 unique), so that shape is benchmarked and tabled twice per sweep combination.

**11. `test_f4gemm.py:509` — output-scale check tightened to byte-exact, but the reference amax isn't bit-identical**

Removed: `MXFP8_SCALE_ATOL = 1.0  # +-1 e8m0 step` and the `# ... compared with atol=1 (kernel rounds within +-1 step)` comment. `ref_f32 = (x_f32*xs) @ (w_f32*ws).T` is a torch fp32 reduction over K up to 53248; the kernel accumulates the same terms in a different order, giving ~1e-4 relative difference. `_e8m0_out_scale` bumps the exponent whenever the mantissa of `amax/256` crosses 1.5, so any 128-wide block whose amax sits within ~1e-4 of a `1.5*2^k` or `2^k` boundary yields an e8m0 byte off by exactly 1. With ~2M blocks per (16384,16384) fp8 run, `err_s > 0` is expected, `err = max(err_s, err_d)` propagates it, and `_verdict()` downgrades previously-passing rows to "warning". Making the *formula* bit-identical to the kernel doesn't make its *input* bit-identical.

**12. `test_f4gemm.py:144` — `_e8m0_out_scale` truncates instead of clamping**

Exponent + guard-bit arithmetic is done in int32 then `.to(torch.uint8)`. Verified empirically: `amax=nan` gives raw int 256 (exp field 255 + guard bit 1 from the quiet-NaN mantissa) and `.to(torch.uint8)` silently yields **0**, i.e. e8m0 code 0 = 2^-127 (the smallest possible scale); `inf` gives 255 (the e8m0 NaN code) with no diagnostic. The reference then computes `xb / 5.9e-39` and the new atol=0 comparison compares garbage. Finite fp32 can't reach 256, but a NaN/Inf reference is exactly what an over/underflowing golden GEMM produces at K=53248. A `.clamp(0, 255)` or an assert would cover it. Separately, this hand-rolls a rounding mode `aiter.utility.fp4_utils` already provides (`f32_to_mx_e8m0_scale(mode=RoundUp, dtype=FP8_E4M3)` / `_f32_to_e8m0_ceil_impl`) — which the PR replaced — and the two disagree for values whose mantissa has no guard bit but nonzero lower bits.

**13. `op_tests/test_mxfp8fp4gemm.py:251` — no `"auto"` `--knl-name` branch, and a missing kernel is treated as a benign skip**

`test_f4gemm.py:366-371` has `elif knl_name == "auto": knl = f"_ZN5aiter{len(base)}{base}E"`. Here line 251 is just `knl = knl_name or ""`, so `--knl-name auto` — the spelling that *is* valid in the sibling test — is passed verbatim, the `.cu` builds `selectedName = "gfx1250auto"`, the lookup misses, and `AITER_CHECK` raises `"kernel not in cfg_mxfp8fp4gemm: gfx1250auto"`. That string is in `_NOT_SUPPORTED_MARKERS` (line 267), so every row is swallowed as `asm result = "not support"` with NaN timings, the summary table prints, and the process **exits 0**.

More broadly: `"cannot get heuristic kernel"` is the same `AITER_CHECK` raised when a CSV row is missing, when the `b_intype` string (just converted from int 0/1 to `"mxfp8"`/`"mxfp4"` in this PR) is misspelled, and when the alignment gate wrongly excludes everything — so a genuine dispatch regression is indistinguishable from an intentional gap, and the marker text is itself unstable because this PR rewrote the tail of that very `AITER_CHECK`.

**14. `aiter/ops/shuffle.py:317` — `shuffle_scale_f4` no longer preserves the input shape**

The old body ended `return out.view(M, N)`. Verified standalone: `shuffle_scale_f4(torch.zeros(1, 256, dtype=torch.uint8), 7)` now returns shape `(32, 256)` / 8192 bytes instead of `(1, 256)` / 256 bytes. `test_f4gemm.py:417-423` feeds this straight into `nbytes = inp["A"].nbytes + ... + inp["sA"].nbytes + ...` and then `ret[f"{name} TB/s"] = round(nbytes / us / 1e6, 2)` — so every `FUNC_SHAPE` with M < 32 (M=1, plus the qkv_proj/attn_out rows) reports up to **32x** the scale bytes that actually move. The docstring at line 293 still says it "Tiles the [M, N] scale buffer", with no mention of padding or of the changed output shape. The old `view(M, N)` also implicitly enforced M%32==0 / N%tile_major==0 by throwing; that validation is gone.

### API surface

**15. `aiter/__init__.py:97` — 8 shipped public symbols and 2 registered custom ops removed with no re-export or shim**

`from .ops.gemm_op_a8w4 import *` is deleted and no `from .ops.asm import *` is added, so `aiter.gemm_a4w4o4`, `gemm_a4w4o8`, `gemm_mxfp4_asm`, `gemm_nvfp4_asm`, `gemm_a8w4_mxfp8`, `gemm_a8w8_mxfp8`, `unpack_mxfp8_out_scale` and `MXFP8_OUT_SCALE_BLOCK` all raise `AttributeError`. `torch_compile_guard` registers under `calling_func.__name__` (`aiter/jit/utils/torch_guard.py:288`), so the ops are renamed `aiter::gemm_a4w4o4` -> `aiter::_gemm_a4w4o4_asm` and `aiter::gemm_a4w4o8` -> `aiter::_gemm_a4w4o8_asm`; any serialized or compiled graph referencing the old names breaks, and since `aiter/ops/asm` is never imported, `import aiter` no longer registers them at all.

No in-repo or downstream (vLLM/SGLang/ATOM) callers were found, so the break is external-only. But `aiter/ops/asm/` isn't an established convention either: every other `aiter/ops/<subdir>/` (opus, triton, flydsl) has a substantive `__init__.py` wired into `aiter/__init__.py`, and `gemm_a4w4_asm` — also an asm op — stays in `gemm_op_a4w4.py`, so the new taxonomy is self-contradictory on day one.

### Minor (not blocking)

- `constexpr int WG_MAX = 256` (`asm_mxfp8fp4gemm.cu:251`) and `PERSISTENT_TG`/`PERSISTENT_GY` (`asm_f4gemm.cu:328`) hardcode per-`.co` properties in the host after deleting the `persistent`/`wg_max` CSV columns. `hsa/codegen.py` generates a typed CFG field per CSV column for free.
- The support matrix is transcribed into 4 places (2 `.cu` ASCII tables, 2 `_support_reason()` re-implementations) and has already drifted: the `.cu` table has no FP4 row despite the CSV registering one, and "fp4 requires apre=1" exists nowhere — so `--outtype fp4 --apre 0` does full prep/shuffle on shapes up to 32768x106496x16384 before failing into the exception path.
- `asm_f4gemm.cu:148` lost the constraint hint from its "cannot get heuristic kernel" message while the sibling gained one.
- Stale comment `# MXFP8_OUT_SCALE_BLOCK (=128) is imported from gemm_op_a4w4.` (`test_f4gemm.py:139`); `f4gemm.csv` has no trailing newline; the `test_mxfp8fp4gemm.py` header table separator is misaligned; `bench_init`'s chunk loop is duplicated verbatim in `fill_fp4`/`fill_fp8` with a no-op `del v`; `trig` init uses a float32 `arange` over absolute flat indices, which collapses to constant runs beyond 2^24 — newly reachable at these shape sizes.

### Verified clean

The generated `asm_*_configs.hpp` cannot desync — `hsa/codegen.py` builds the struct from the CSV and correctly types the new string columns. All 17 mangled `knl_name` length prefixes match symbols actually exported by the `.co` files. Every `.co`'s `.kernarg_segment_size` (80/88/96 f4gemm, 76 f8gemm) matches the host's `arg_size` branch. `shuffle_scale_f4`'s **column** padding does match `stride_sa` for both mxfp4 and nvfp4. No dangling references to any removed symbol anywhere in the repo or in vLLM/SGLang/ATOM. `-s`/`-mnk` introduce no argparse collision, every positional `test_gemm(...)` call maps correctly onto the new signature, and `ruff check` passes on all changed Python files.

valarLip (approved):

PR #4779, opened 2026-08-15
[Bugfix] Handle GroupNorm autocast safely

### LINKED ISSUE #4780: [Bug] GroupNorm does not follow FP32 autocast policy
## Problem

AITER GroupNorm does not follow PyTorch's FP32 autocast policy. Hunyuan Image 3.0 calls GroupNorm under CUDA autocast with an FP16 input and BF16 weight and bias tensors.

The AITER HIP kernel selects its pointer type from the input dtype. It therefore reads the BF16 weight and bias storage as FP16 when the input is FP16. AITER returns an FP16 output, while PyTorch returns an FP32 output.

In the Hunyuan VAE input with shape `(1, 128, 4, 1024, 1024)`, the mean error was 1.0778 and the maximum error was 6.1785.

## Source references

- The [PyTorch AMP documentation](https://docs.pytorch.org/docs/stable/amp.html#cuda-ops-that-can-autocast-to-float32) lists `group_norm` among the CUDA operations that autocast to FP32.
- The PyTorch source includes [`group_norm` in `AT_FORALL_FP32`](https://github.com/pytorch/pytorch/blob/1f4b1ff35d539c16f32e1805361f19f5c95a432c/aten/src/ATen/autocast_mode.h#L854-L876) and [registers that list with the FP32 policy for `AutocastCUDA`](https://github.com/pytorch/pytorch/blob/1f4b1ff35d539c16f32e1805361f19f5c95a432c/aten/src/ATen/autocast_mode.cpp#L181-L193).
- The AITER [Python wrapper calls the HIP binding directly](https://github.com/ROCm/aiter/blob/878d60d779fa774496fe075175d0e99f1d49eea9/aiter/ops/groupnorm.py#L59-L78), so the call does not pass through PyTorch's autocast dispatcher.
- The AITER [HIP launcher selects its template type only from the input dtype](https://github.com/ROCm/aiter/blob/878d60d779fa774496fe075175d0e99f1d49eea9/csrc/kernels/groupnorm.cu#L349-L379). The selected type is then used for the input, weight, bias, and output pointers.

## Reproduction

```python
import torch

from aiter.ops.groupnorm import GroupNorm

norm = GroupNorm(
    32,
    128,
    eps=1e-6,
    affine=True,
    device="cuda",
    dtype=torch.bfloat16,
)
x = torch.randn(
    (1, 128, 4, 64, 64),
    device="cuda",
    dtype=torch.float16,
)

with torch.inference_mode(), torch.amp.autocast("cuda", dtype=torch.float16):
    expected = norm(x, use_torch=True)
    actual = norm(x)

print(expected.dtype)
print(actual.dtype)
print((actual.float() - expected).abs().mean())
print((actual.float() - expected).abs().max())
```

## Expected behavior

AITER should cast the input, weight, and bias to FP32 while autocast is active. The AITER output should be FP32 and should match PyTorch GroupNorm.

## Environment

- GPU: AMD MI300X with gfx942
- ROCm: 7.2
- PyTorch: 2.11
- AITER: 0.1.16.post3
- Consumer: vLLM Omni with Hunyuan Image 3.0

akshatvishu:
## Summary

Make AITER GroupNorm follow PyTorch's FP32 autocast policy. Add dtype checks before the HIP kernel reads the input and affine parameter pointers.

Fixes #4780

## Motivation

Hunyuan Image 3.0 calls GroupNorm under CUDA autocast with an FP16 input and BF16 weight and bias tensors. PyTorch registers GroupNorm as an FP32 autocast operation, but AITER calls its HIP kernel directly and does not pass through that dispatcher rule.

The AITER kernel selects its pointer type from the input dtype. Before this change, it treated the BF16 weight and bias storage as FP16 when the input was FP16. In the Hunyuan VAE, the AITER output was FP16 while the PyTorch output was FP32. The mean error was 1.0778 and the maximum error was 6.1785.

## Changes

- Cast the input, weight, and bias to FP32 when CUDA autocast is active, then call the existing FP32 GroupNorm kernel with autocast disabled.
- Keep the existing FP16, BF16, and FP32 paths unchanged when autocast is disabled.
- Check that the input, output, weight, and bias have the same dtype before launching the kernel.
- Check that the scratch workspace uses FP32.
- Add the exact Hunyuan GroupNorm shape to the standalone test and include correctness and performance measurements.
- Use a BF16 tolerance for the existing BF16 benchmark.

## Performance

Tests ran on one MI300X with gfx942, ROCm 7.2, and PyTorch 2.11. The autocast measurement includes the FP32 tensor conversions.

| Configuration | PyTorch | AITER | Speedup | Correctness |
|---|---:|---:|---:|---|
| Hunyuan autocast, shape `(1, 128, 4, 1024, 1024)` | 15,613.13 us | 2,722.52 us | 5.73x | Passed with `rtol=0.001`, `atol=0.01` |
| Direct FP32, shape `(1, 128, 4, 64, 64)` | 31.1 us | 42.3 us | 0.74x | Passed |
| Direct BF16, shape `(1, 512, 3, 64, 64)` | 98.3 us | 38.7 us | 2.54x | Passed |

The change does not alter the HIP GroupNorm arithmetic or memory access pattern. Bandwidth and roofline analysis do not apply to this correctness fix.

## Testing

- [x] Unit tests added or updated
- [x] Performance benchmarks run
- [ ] Tested on MI250X
- [x] Tested on MI300X

```bash
python3 op_tests/test_groupnorm.py --dtype float32 -b 32,1,128,4,64,64 --autocast-spatial-size 1024
python3 op_tests/test_groupnorm.py --dtype bfloat16 -b 32,1,512,3,64,64 --autocast-spatial-size 64
ruff check aiter/ops/groupnorm.py op_tests/test_groupnorm.py
ruff format --check aiter/ops/groupnorm.py op_tests/test_groupnorm.py
```

## Documentation

- [x] Docstrings updated
- [ ] User guide updated because no user guide change is needed
- [ ] Performance guide updated because no performance guide change is needed

## Dependencies

- [x] No new third party dependencies added

## PyTorch compatibility question

AITER documentation currently lists PyTorch 2.0 and later as supported. The `device_type` argument for `torch.is_autocast_enabled` was added in PyTorch 2.4, while PyTorch 2.0 to 2.3 accept no arguments.

Maintainers, should GroupNorm preserve PyTorch 2.0 to 2.3 compatibility by using `torch.is_autocast_enabled()` without an argument, or is the current AITER baseline PyTorch 2.4 or later?

## Breaking changes

A direct mixed dtype call without autocast now stops with a clear error instead of launching a kernel that interprets the affine parameter storage using the input dtype.

### DIFF
--- a/aiter/ops/groupnorm.py
+++ b/aiter/ops/groupnorm.py
@@ -56,15 +56,13 @@ def _groupnorm_workspace(ws_slots: int, device: torch.device) -> Tensor:
     return ws
 
 
-def groupnorm_run(
+def _groupnorm_run_impl(
     input: Tensor,
     num_groups: int,
     weight: Tensor,
     bias: Tensor,
     eps: float,
 ) -> Tensor:
-    """Group Normalization. Allocates output and reuses a cached scratch
-    workspace, then calls the HIP kernel. Returns the normalized output."""
     input = input.contiguous()
     weight = weight.contiguous()
     bias = bias.contiguous()
@@ -76,6 +74,26 @@ def groupnorm_run(
 
     _groupnorm_run(y, workspace, input, num_groups, weight, bias, eps)
     return y
+
+
+def groupnorm_run(
+    input: Tensor,
+    num_groups: int,
+    weight: Tensor,
+    bias: Tensor,
+    eps: float,
+) -> Tensor:
+    """Group Normalization. Allocates output and reuses a cached scratch
+    workspace, then calls the HIP kernel. CUDA autocast follows PyTorch's FP32
+    policy for GroupNorm. Returns the normalized output."""
+    device_type = input.device.type
+    if torch.is_autocast_enabled(device_type):
+        with torch.amp.autocast(device_type, enabled=False):
+            return _groupnorm_run_impl(
+                input.float(), num_groups, weight.float(), bias.float(), eps
+            )
+
+    return _groupnorm_run_impl(input, num_groups, weight, bias, eps)
 
 
 class GroupNorm(torch.nn.Module):
--- a/csrc/include/groupnorm.hpp
+++ b/csrc/include/groupnorm.hpp
@@ -15,7 +15,9 @@ namespace aiter {
 //   bias:      affine bias   [num_channels], contiguous
 //   epsilon:   numerical stability term
 //
-// Supported dtypes for x/y/weight/bias: fp32, fp16, bf16.
+// Supported dtypes for x/y/weight/bias: fp32, fp16, bf16. These four tensors
+// must have the same dtype. The Python wrapper casts them to fp32 under CUDA
+// autocast to match PyTorch GroupNorm.
 void groupnorm_run(aiter_tensor_t& y,
                    aiter_tensor_t& workspace,
                    const aiter_tensor_t& x,
--- a/csrc/kernels/groupnorm.cu
+++ b/csrc/kernels/groupnorm.cu
@@ -361,6 +361,11 @@ void groupnorm_run(
 
     AITER_CHECK(weights.numel() > 0 && bias.numel() > 0,
                 "groupnorm requires non-empty affine weight and bias");
+    AITER_CHECK(y.dtype() == x.dtype() && weights.dtype() == x.dtype() &&
+                    bias.dtype() == x.dtype(),
+                "groupnorm requires input, output, weight, and bias to have the same dtype");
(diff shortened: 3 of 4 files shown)


yzhou103 (approved):
LGTM on the approach.
Re: your PyTorch 2.0–2.3 question — no need to worry about it. setup.py:115 already
gates aiter on torch >= 2.9.1, well past the 2.4 that added the `device_type`
argument

akshatvishu:
Thanks for the review @yzhou103 !

PR #4781, opened 2026-08-16
[gfx950] Retune the small-M tiles and wave counts in the A16W16 fallback config

akii96:
## What this changes

| bucket | change |
|---|---|
| `M_LEQ_8` | `BLOCK_SIZE_N` 32 → 16, `num_warps` 4 → 8 |
| `M_LEQ_16` | `BLOCK_SIZE_M` and `BLOCK_SIZE_N` 32 → 16 |
| `M_LEQ_32` | `num_warps` 4 → 8 |
| `M_LEQ_64` | `num_warps` 4 → 8 |

## Why

Two problems in the same table.

The `M_LEQ_16` tile overshoots its own bucket. It only handles M of 16 or smaller, but the tile was sized for 32 rows, so half of every tile did no useful work and the kernel took a slower path for leftover rows that were not really there.

Wave count stops tracking tile size in the middle of the table. `M_LEQ_32` computes a 32x32 tile with four waves and `M_LEQ_64` a 64x32 tile with four, while every bucket from `M_LEQ_128` up already uses eight. These are skinny bandwidth-bound GEMMs, and four waves cannot cover the memory latency.

`M_LEQ_8` has both problems, and matters more than its name suggests: the table has no `M_LEQ_1` or `M_LEQ_4` key, so every M from 1 to 8 resolves to it.

## Numbers

Weighted over every measurable gfx950 bf16 shape in each bucket. Nothing regresses.

| bucket | shapes | weighted | median | worst |
|---|---:|---:|---:|---:|
| `M_LEQ_8` | 49 | 1.83x | 1.66x | 1.10x |
| `M_LEQ_16` | 34 | 1.80x | 1.77x | 1.00x |
| `M_LEQ_32` | 34 | 1.98x | 1.90x | 1.14x |
| `M_LEQ_64` | 10 | 1.18x | 1.15x | 1.09x |

## Accuracy

Bit for bit identical. `BLOCK_SIZE_N` and `num_warps` change how a tile is partitioned across waves, not the order any output element accumulates over K, and `NUM_KSPLIT` stays at 1. Verified with `torch.equal` against the shipped config on all 93 shapes: zero differ.


akii96:
@azaidy validated this PR alone on on 4x MI355X (gfx950) with `amd/MiniMax-M3-MXFP4`

**Image:** `vllm/vllm-openai-rocm:nightly-311b3513af33bc29b4acb2fde2e9313e5e9966a0`
**Additional Fix applied on image to enable serving with ATTN backend:** [#51585](https://github.com/vllm-project/vllm/pull/51585) 

Before the said Additional fix mentioned above, server startup fails during graph capture. 

### Performance

| ISL/OSL | Conc | TTFT before PR (ms) | TTFT after (ms) | Δ | TPOT before PR (ms) | TPOT after PR (ms) | Δ |
|---------|------|-------------------|------------------|---|-------------------|------------------|---|
| 60k/600 | 8 | 11,884 | **8,733** | **-26.5%** | 18.50 | **17.60** | **-4.9%** |
| 60k/600 | 16 | 10,064 | **10,115** | — | 32.72 | **30.14** | **-7.9%** |
| 8k/1k | 8 | 980 | **991** | — | 10.43 | **9.73** | **-6.7%** |
| 8k/1k | 16 | 1,720 | **1,613** | **-6.2%** | 12.44 | **11.56** | **-7.1%** |
| 8k/1k | 64 | 3,366 | **3,372** | — | 24.39 | **24.36** | — |


<details>
<summary>Serving command</summary>

```bash
export VLLM_USE_BREAKABLE_CUDAGRAPH=1
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
export VLLM_ROCM_USE_AITER_UNIFIED_ATTENTION=1
export VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1
export VLLM_ROCM_QUICK_REDUCE_QUANTIZATION=INT4
export VLLM_WORKER_MULTIPROC_METHOD=spawn
export SAFETENSORS_FAST_GPU=1

vllm serve amd/MiniMax-M3-MXFP4 \
  --tensor-parallel-size 4 \
  --distributed-executor-backend mp \
  --trust-remote-code \
  --block-size 128 \
  --no-enable-prefix-caching \
  --language-model-only \
  --max-model-len 133120 \
  --gpu-memory-utilization 0.95 \
  --max-num-batched-tokens 32768 \
  --max-num-seqs 128 \
  --enable-chunked-prefill \
  --async-scheduling \
  --moe-backend aiter \
  --tool-call-parser minimax_m3 \
  --enable-auto-tool-choice \
  --reasoning-parser minimax_m3 \
  --no-enable-log-requests

valarLip (approved):

PR #4787, opened 2026-08-17
[HIP] [JIT] [gfx950] Optimize Minimax M3 scoring & top-k kernels

ukannika:
Add fp8 MFMA sparse block scoring and top-k with fused page-table needed for pa gluon kernel.


andyluo7:
The current head `cb3c7a628645` now has the full AITER test gate green, including all gfx950/MI35X shards and the final Aiter Test Gate. vLLM #52664 depends on these exact MiniMax-M3 score/top-k entry points; its focused current-head MI355X integration validation is queued now, and I will add the exact result when it completes.

Could an AITER maintainer please review #4787 for merge? This is the kernel dependency rather than a duplicate vLLM implementation.

aiter/jit/optCompilerConfig.json
```diff
@@ -770,6 +770,25 @@
         "verbose": "False",
         "blob_gen_cmd": "''"
     },
+    "module_sparse_attn": {
```
valarLip: "module_sparse_attn" maybe more specific like xxx_msa?

valarLip (changes requested):

aiter/jit/optCompilerConfig.json
ukannika: thank you, updated.

valarLip (approved):

PR #4797, opened 2026-08-17
fix: generate CK FMHA kernels for gfx1250

tomjen12:
## Summary
- Pass the runtime gfx target to CK dense forward and backward code generation.
- Pass the same target to CK varlen backward code generation.
- Enable gfx1250 CK dispatch instead of falling back with an invalid-argument error.

## Manual validation
- Validated dense and varlen FlashAttention forward/backward on gfx1250 using downstream captured workloads.
- Compared outputs and Q/K/V gradients with PyTorch math SDPA.
- Confirmed CK forward and backward kernels launch in Torch Profiler.

### DIFF
--- a/aiter/jit/utils/mha_recipes.py
+++ b/aiter/jit/utils/mha_recipes.py
@@ -7,9 +7,9 @@ def _ck_targets_flag() -> str:
     keep the default (covers both gfx942 and gfx950 like before).
     """
     try:
-        from chip_info import get_gfx
+        from chip_info import get_gfx_runtime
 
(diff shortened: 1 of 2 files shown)


tomjen12:
@junhaha666, @amd-ruitang3  Could you help review the PR?

zufayu:
no ck use plan for 1250 yet, so for now we will reject all CK changes for 1250

zufayu:
no ck use plan for 1250 yet, so for now we will reject all CK changes for 1250

PR #4800, opened 2026-08-17
[OPUS] [JIT] [Bugfix] Make blob codegen cache publication transactional

### LINKED ISSUE #4799: [Bug] Interrupted JIT codegen can leave and persistently reuse an incomplete CK MHA cache
## Summary

An interrupted AITER CK MHA JIT build left a truncated generated source file in the
canonical cache directory. The file was later reused as valid cache state, producing a
loadable module with no matching dispatch branch. Every affected prefill then failed
with:

```text
RuntimeError: invalid argument for fmha_fwd
  at csrc/py_itfs_ck/mha_varlen_fwd_kernels.cu:671
```

Deleting the cache and allowing a complete rebuild recovered the service. Restarting
with the same cache did not.

The historical online-build trigger was fixed by
[PR #2957](https://github.com/ROCm/aiter/pull/2957) in AITER `v0.1.15`, which added the
missing prebuilt `_skip` MHA variants. However, the cache-integrity problem remains
relevant to any operator or installation that still uses JIT.

## Environment where the failure was observed

| Component | Version |
|---|---|
| AITER | editable install at `8ad75337ed6fdcc27b7c48797e7a455f4bfa9409` (`v0.1.14-rc0-64-g8ad75337e`) |
| ATOM | `0.1.3.dev213+g112abcac1` (`112abcac1a0ebb88c3d168e10886ba9ffe5a667e`) |
| vLLM | `0.19.1.dev0+g2a69949bd` |
| ROCm / GPU | 7.2.3 / MI308X (gfx942) |
| Model | `Qwen/Qwen3.5-35B-A3B-FP8` (`head_dim=256`) |
| Image | `rocm/atom-dev:vllm-v0.19.0-nightly_20260520-qwen3` |

## Observed cache evidence

The affected variant was:

```text
mha_varlen_fwd_bf16_nlogits_nbias_mask_nlse_ndropout_skip_nqscale
```

The generated dispatch source differed as follows:

| Artifact | Lines | `skip_min_seqlen_q ==` branches |
|---|---:|---:|
| Truncated cached file | 92 | 0 |
| Healthy rebuild | 352 | 40 |

With zero dispatch branches, `fmha_fwd()` returned `-1`, which triggered the existing
`TORCH_CHECK(t >= 0, "invalid argument for fmha_fwd")` check. The generated module was
nevertheless reused on subsequent starts.

CK itself supports this workload: after a clean rebuild, the same path worked at both
`head_dim=192` and `head_dim=256`.

## Historical trigger and upstream status

Historical ATOM passed `min_seqlen_q=1`, which AITER maps to the `_skip` suffix. The
affected AITER prebuild list contained only `_nskip` variants, so the first request
performed an approximately 80-second online build.

AITER commit
[`252192a0d`](https://github.com/ROCm/aiter/commit/252192a0d2747cfd9de8164ffe985aacb3ad7ae0)
added the required `_skip` variants and was released in AITER `v0.1.15` on 2026-06-06.
The entries remain in current
[`setup.py`](https://github.com/ROCm/aiter/blob/ca1909e300d7860bd6d002e8c95c6c2dae993a4d/setup.py#L322-L332).

This fixes the historical prebuild-coverage gap. It does not validate or transactionally
publish JIT artifacts.

## Current code path

In current AITER main as rechecked at
`bf1e10585403f3a38ae88a2cf1895b1ea007c1af`,
[`exec_blob()`](https://github.com/ROCm/aiter/blob/bf1e10585403f3a38ae88a2cf1895b1ea007c1af/aiter/jit/core.py#L949-L963):

1. writes directly into the canonical `blob/` directory;
2. invokes the generator through `os.system(...)` without checking its return code;
3. immediately consumes whatever files are present in that directory.

The same build path also removes the installed target `.so` before a replacement has
successfully compiled ([lines 850-865](https://github.com/ROCm/aiter/blob/bf1e10585403f3a38ae88a2cf1895b1ea007c1af/aiter/jit/core.py#L850-L865)).

There is no operator-level validation that the generated dispatch table is non-empty,
and no completion manifest distinguishing a fully published cache from an interrupted
build.

## Actual behavior

- An interrupted or failed generator can leave partial files in the normal cache path.
- Those files can be compiled or reused as if generation completed successfully.
- The service may start normally, but the first affected prefill fails.
- Process restarts do not recover while the persistent cache remains.
- There is no automatic Triton fallback on this prefill path.

## Expected behavior

- A failed or interrupted codegen run must never modify the last known-good cache.
- Incomplete artifacts must never be published or reused.
- Generator failure must be propagated to the caller with the failed command and path.
- A cache without a valid completion record must be rebuilt or quarantined.

## Suggested fix

1. Replace `os.system(...)` with `subprocess.run(..., check=True)` so generator failures
   are not ignored.
2. Generate all files in a unique temporary directory.
3. Validate the generated operator output before publication. For CK MHA, reject an
   empty `fmha_fwd_api.cpp` dispatch table.
4. Publish the generated directory and compiled `.so` atomically only after generation,
   validation and compilation all succeed.
5. Write a completion manifest containing the generator command/version and artifact
   hashes. Treat a missing or invalid manifest as a cache miss.
6. Preserve the previous healthy cache if a rebuild fails.

At minimum, the unchecked call should become a checked subprocess:

```diff
- os.system(f"{PY} {blob_gen_cmd.format(blob_dir)}")
+ subprocess.run(
+     [PY, *shlex.split(blob_gen_cmd.format(tmp_blob_dir))],
+     check=True,
+ )
```

The temporary directory must not replace `blob_dir` until semantic validation passes.
The `.so` should likewise be written under a temporary name and published with
`os.replace()`.

## Suggested regression test

1. Use a test generator that writes a deliberately incomplete
   `fmha_fwd_api.cpp`, then exits non-zero or is terminated.
2. Verify that no completion manifest or canonical cache update is published.
3. Verify that concurrent waiters receive a build failure rather than importing the
   partial module.
4. Retry with a successful generator and verify that the artifact is published once.
5. Start another process and verify that it loads the completed cache without rebuilding.

## Recovery for already affected hosts

```bash
V=mha_varlen_fwd_bf16_nlogits_nbias_mask_nlse_ndropout_skip_nqscale
rm -rf ~/.aiter/jit/build/$V ~/.aiter/jit/$V.so
```

Historical diagnostic for this exact build:

```bash
BLOB=~/.aiter/jit/build/$V/blob/fmha_fwd_api.cpp
grep -c "skip_min_seqlen_q ==" "$BLOB"
```

`0` identified the truncated artifact; the healthy artifact had `40`. This count is
build-specific and should not be used as the general cache contract.

## Evidence boundary

The truncated artifact, its reuse, the hard failure, and recovery after a clean rebuild
were observed. The event that interrupted the original JIT was not captured, so this
report does not claim a confirmed cause for the interruption itself.

Ruye-aa:
## Summary

Make JIT code generation recoverable without invalidating every incremental build:

- Generate under the deterministic `{module}/blob.staging` path, reuse unchanged generated files, and retain the working tree for Ninja retries and debugging.
- Check generator exit status and accept generated C++/HIP sources or headers; reject genuinely empty output.
- Compile a fixed module target, atomically install the binary, then publish source-cache and Opus metadata updates best-effort.
- Preserve the existing explicit rebuild contract: level 1 removes the installed module and clears its build tree; level 2 removes only the installed module.
- Bind the canonical Opus compiled-kid sidecar to the exact installed binary, and synchronously prepare tuner requests before spawning workers.

Closes #4799

## Implementation and recovery

The source transaction helpers live in `aiter/jit/utils/jit_cache.py`.

1. Checked generators share one stable staging tree under the module lock. Incomplete codegen is restored from the last published snapshot through a separate candidate, installed only after its copy completes.
2. The build carries its own generation token across compilation and installation. It rejects changed/incomplete inputs before installing the binary; these checks supplement, not replace, the module lock.
3. `build_module` uses `_jit_compile(..., use_versioner=False)`: the installed fixed name and its `PyInit` symbol stay consistent. Every build request reaches Ninja's incremental checks, including failed-build retries, header changes and missing outputs. Direct extension loaders retain their default versioner behavior.
4. After atomic binary installation, source snapshot publication failures—including a peer winning a directory swap—log warnings instead of failing the successful build. A failed rollback retains the last published backup for recovery.
5. Publish/recovery candidates inherit `op_dir`'s permission mode immediately after creation. Nested source modes are preserved; intentionally restrictive modes are not widened.
6. Owner-tagged transactions from dead local processes are reclaimed on the next module build. Live/remote owners are retained. Legacy artifacts use a 24-hour deletion grace, but a legacy backup may be recovered immediately under the lock when `blob` is missing.

## Opus sidecar and tuner

The tuner reads `{bd_dir}/compiled_kids_opus.json`, outside the per-module tree so it survives `clear_build`. Runtime dispatch uses CSV/C++ lookup tables, not this sidecar.

JIT seeds the staged sidecar from the canonical successful set. After installing the binary, it publishes the generated sidecar and a receipt binding its hash to that binary's file identity, independently of source-cache publication. Failed compilation does not advance the canonical set; missing/stale receipts cannot satisfy tuner reuse.

The tuner passes candidates through `--extra_kids`, waits for its own requested build when a peer's build may not cover it, and honors an explicit rebuild once in the parent. Success clears the environment flag for spawned workers; failure or cancellation restores the previous value. The generator rejects unknown, off-architecture or family-filtered explicit requests instead of silently omitting them.

## Validation

Local Python 3.12 validation:

```text
python op_tests/test_jit_cache_transaction.py            53 tests passed
python -m pytest -q op_tests/test_jit_cache_transaction.py
                                                       53 passed, 21 subtests passed
Ruff, Black --check, py_compile, config JSON, diff check  passed
```

The executable unittest entrypoint matches AITER's script-mode CI runner. Tests use checkout-specific imports and module-local fault injection without changing stdlib filesystem functions or globally shadowing JIT modules.

Coverage includes real JIT/versioner/FileBaton control flow with the HIP compiler stubbed, a real CPU C++/Ninja incremental build, real CPU Opus generation with only the unavailable CUDA probe stubbed, interrupted recovery, publication races, metadata freshness, rebuild levels, permission modes and tuner cancellation. The optional C++/Ninja and pandas-dependent generator tests ran locally; they skip explicitly if those dependencies are unavailable elsewhere.

## Integration boundaries and storage

ROCm/GPU compilation and execution, and a real root-prebuild to non-root container run, still require integration validation. Explicit rebuild levels retain the pre-existing binary-removal window for unrelated readers; this PR does not promise global reader isolation or protection from edits bypassing generation markers.

One module keeps staging plus its published source snapshot: approximately 2S at rest and up to 3S during a changed snapshot copy, excluding compiler outputs. Documented CPU CK codegen measurements put the two retained trees for seven MHA configurations at about 1.95 GiB on the measured filesystem, excluding HSA headers, objects, binaries and abandoned transactions. This is an additional runner disk requirement, not proof that every runner's quota is sufficient.

See [docs/jit_cache.md](https://github.com/Ruye-aa/aiter/blob/codex/jit-transactional-codegen/docs/jit_cache.md) and the updated [Opus README](https://github.com/Ruye-aa/aiter/blob/codex/jit-transactional-codegen/aiter/ops/opus/README.md) for lifecycle and operational details.


Ruye-aa:
Hi @zufayu and @valarLip, could you please help move #4800 forward?

@zufayu, the required GitHub Actions workflows are still awaiting maintainer approval. Could you approve the pending runs, or point me to a maintainer who can?

@valarLip, when you have time, could you please review this change?

This PR fixes interrupted CK MHA JIT code generation leaving an incomplete persistent cache and causing repeated `invalid argument for fmha_fwd` failures.

I also applied the patch to the current `main` in a local scratch worktree. It applies cleanly, and the five CPU regression tests, `py_compile`, `git diff --check`, and Black all pass. GPU/ROCm validation is still pending upstream CI.

Thanks!

junhaha666 (dismissed):

Ruye-aa:
Follow-up: I pushed d3c26cf to fix all three Ruff findings from the previous Checks run:

- removed the unused E402 noqa directive;
- combined the two nested context-manager pairs flagged by SIM117.

Local verification now passes for ruff 0.16.0 on the changed file, Black, all five CPU regression tests, py_compile, and git diff --check.

The new push dismissed the previous approval and created a fresh set of workflow runs that are awaiting maintainer approval. @junhaha666, could you please re-approve the updated commit? @zufayu, could you please approve the pending workflows?

junhaha666:
Hello, it failed the CI test.

Ruye-aa:
Thanks @junhaha666 for flagging the CI failure. I fixed the reported issue in `449d3e60`.

The full Aiter Test was delayed while the workflows were awaiting maintainer approval, so this compatibility gap was detected later than it should have been. The branch also had to be kept in sync with a fast-moving `main`, and each sync or fix created a new set of workflows requiring approval.

The failure was caused by a valid header-only HSA code generator: it produces `asm_*_configs.hpp`, while the new transactional validation accepted only `.cpp` and `.cu` outputs and incorrectly treated the header as no generated build input. The fix now accepts C++/HIP sources and headers (`.cpp`, `.cu`, `.h`, `.hpp`, `.cuh`) while still rejecting a generator that truly produces no build input.

Local validation for the new head:

- the real `hsa/codegen.py -m pa` header-only path passes transactional staging;
- all 6 CPU regression tests pass;
- Ruff 0.16.0, Black, `py_compile`, and `git diff --check` pass;
- the branch is currently up to date with `main` (`behind_by=0`).

@zufayu, could you please approve the current pending workflow runs as soon as possible so this fix can be validated by the upstream Python 3.12 wheel/GPU CI? Prompt approval would avoid another long feedback cycle and another branch sync resetting the runs.

@valarLip, could you please review the updated change once the workflows are available and pass?

Thank you.

valarLip (changes requested):
## Review: `codex/jit-transactional-codegen`

The goal is right — codegen should be transactional so an interrupted generator can't leave `blob/` half-written. My concern is the mechanism: staging into a `tempfile.mkdtemp` directory and committing by rename has two systemic costs (both of which force full recompiles for every user, every build), the transaction guarantee has a real hole, and the tests that are supposed to protect all of this never execute in CI.

A single change would fix the two performance findings and the reproducibility one: use a **deterministic** staging name (e.g. `blob.staging` under the existing FileBaton lock) instead of a random `mkdtemp` name.

### Performance — this regresses every incremental build

**1. `aiter/jit/core.py:1065` — the random staging path is baked into `-I` and into every source path, invalidating the ninja object cache on every build.**
`hipify` defaults to False for every real caller, so `rename_cpp_to_cu` (core.py:643-645) returns paths under `{op_dir}/.blob-XXXXXX/` instead of copying into `src_dir`, and `active_blob_dir` (core.py:1036) is that same random dir. `cpp_extension.py:1703` derives object names from the basename, so object names stay stable — but the recorded command line and input path change every build, so ninja considers everything dirty. Retrying an interrupted CK MHA or opus build used to be an incremental ninja resume against a stable `blob/`; it now recompiles thousands of TUs from scratch, adding tens of minutes to hours per retry.

**2. `aiter/jit/utils/jit_cache.py:25` — generating into a fresh empty dir defeats CK codegen's content-compare write-skip.**
`3rdparty/composable_kernel/example/ck_tile/01_fmha/codegen/utils.py:9-22` (`update_file`) reads the existing file and skips the write when the content is identical, with an explicit comment that this exists to avoid touching files and triggering rebuilds. With an always-empty staging dir there is never a prior file, so all several hundred `fmha_fwd_*.cpp` get fresh mtimes on every build. This hits `module_mha_fwd`, `module_mha_varlen_fwd/bwd`, `module_mha_batch_prefill`, `module_vsa_sparse_attention`, `libmha_fwd/bwd` — the slowest modules in the repo. Note this would still force full recompiles even if finding #1 were fixed on its own.

### Correctness

**3. `core.py:1104` — a cache-publication failure turns a fully successful build into a hard failure.**
`publish_blob_sources` runs inside the build `try` and *before* the `.so` is installed. Compilation and linking have already produced `{opbd_dir}/{target_name}`; if either `os.replace` raises (EACCES on a root-owned `blob/`, EXDEV, ENOSPC, `blob` existing as a non-directory), control jumps to the `except` at core.py:1113, `atomic_copy` never runs, and the caller gets `RuntimeError([aiter] build [md] ... failed)`. The user pays a full multi-minute CK/opus build, gets no module, and every subsequent run repeats it. Publication is a pure cache optimization — it should be best-effort and should happen after the `.so` is installed.

**4. `core.py:1133` — the `finally` deletes the staging dir while `build.ninja` still points at it.**
The error at core.py:1129-1131 says `build [md_name] under {opbd_dir} failed`, and the standard follow-up is to enter `{op_dir}/build` and re-run `ninja` to iterate on the compiler diagnostic. Previously `{op_dir}/blob/*.cu|*.cuh` survived a failed build. Now those exact `.blob-XXXXXX/...` strings are in `build.ninja` and the `-I` flag and the `finally` rmtree's them, so `ninja` reports `missing and no known rule to make it` and the generated instance that failed to compile is gone. The same applies on success (the dir is renamed away), so post-build `ninja`, `addr2line`, and `--save-temps` workflows no longer work in that tree.

**5. `jit_cache.py:52` — `publish_blob_sources` is not safe against two concurrent builders of the same module.**
core.py:62-78: when `baton.wait()` returns False (a stale lock broken on a merely-slow holder), the waiter loops and builds the same module itself. Process A runs `os.replace(blob_dir, backupA)`; B then sees `lexists(blob_dir)` as False, takes no backup, and replaces its staging dir into `blob`. A's `os.replace(stagingA, blob_dir)` now hits a non-empty directory and raises `OSError(ENOTEMPTY)` (verified). A's rollback at line 58 is skipped because `blob_dir` exists again, so `backupA` leaks permanently and A's successful compile is reported as a build failure.

**6. `core.py:928` — `AITER_REBUILD=2` silently becomes a no-op.**
Deleting the `elif AITER_REBUILD >= 2: rm_module(md_name)` branch (and `rm_module` from the `== 1` branch) means nothing in `build_module` reacts to level 2 any more. `aiter/utility/base_tuner.py:599-604` sets `jit_core.AITER_REBUILD = 2` with the comment "level 2: rm .so only, keep build cache for faster rebuild", and `csrc/opus_gemm/opus_gemm_tune.py:1039-1080` documents relying on that removal so spawned mp_tuner workers can't dlopen the old subset library. A worker starting during the parent's rebuild window now loads the stale `.so` and fails with `Kernel id X not found`; the explicit scrub at opus_gemm_tune.py:1226-1233 only runs on the failure path. This behavior change isn't mentioned in the PR description — was it intentional?

**7. `jit_cache.py:25` — `mkdtemp` is 0700, so the published `blob/` becomes 0700 instead of 0755.**
Verified: `mkdtemp` → 0o700, `os.makedirs` → 0o755 under umask 022, and `publish_blob_sources` renames without restoring permissions. A ROCm container that prebuilds as root (`setup.py` prebuild / `.github/scripts/prebuild_*`) and then runs the app as a non-root UID will hit `get_user_jit_dir()` (core.py:522-526), find `aiter/jit` unwritable, and fall into `shutil.copytree(this_dir, ~/.aiter/jit)`, which walks the root-owned 0700 `aiter/jit/build/<md>/blob` and dies with PermissionError — breaking `import aiter` outright. Same class of break for any `AITER_JIT_DIR` shared across UIDs, where `-I <blob>` becomes unreadable. Suggested fix: `chmod` the staging dir to `0o777 & ~umask` before publishing.

**8. `jit_cache.py:25` — nothing ever reaps abandoned `.blob-*` staging directories.**
`stage_blob_sources` only cleans up its own staging dir via its `except` handler; a SIGKILL/OOM/Ctrl-C during codegen or during the (long) compile leaves the entire `.blob-XXXXXX` tree behind under `{bd_dir}/{md_name}/`. CK MHA and opus blob sets are hundreds of MB to GBs each, so a handful of OOM-killed builds on a CI runner or a dev box silently fills the disk. Neither `clear_build` (only on `AITER_REBUILD == 1`) nor any other path removes them. Same for `blob.backup.<pid>.<ns>` dirs when the process dies between the two `os.replace` calls, or when the `shutil.rmtree(..., ignore_errors=True)` at line 62 silently fails. The old design reused a single `blob/` and had no such accumulation.

**9. `jit_cache.py:49` — for `module_deepgemm_opus` the generator commits state outside the staging dir, so the transactional guarantee doesn't hold there.**
`aiter/jit/optCompilerConfig.json:361` passes `--compiled_kids_sidecar={0}/../../compiled_kids_opus.json`, resolving to `{bd_dir}/compiled_kids_opus.json`, and `csrc/opus_gemm/gen_instances.py:1699-1704` writes it unconditionally at the end of codegen — before any compilation. If codegen succeeds and expands the kid set and then hipcc OOMs or the user Ctrl-Cs, core.py:1135 discards the staging dir, `blob/` still holds the old instance set, but the sidecar permanently claims the new larger set. That's exactly the desync class this PR set out to eliminate. Either stage the sidecar too, or narrow the docstring at core.py:665-669 ("a failed or interrupted generator must never modify the canonical blob directory").

### Tests — the new tests never run

**10. `op_tests/test_jit_cache_transaction.py:173` — no `__main__` guard, and aiter CI doesn't use pytest.**
`.github/scripts/split_tests.sh:53` shards `find op_tests -maxdepth 1 -name 'test_*.py'` and `.github/scripts/aiter_test.sh:85` runs `timeout 60m python3 "$file"`. `python3 op_tests/test_jit_cache_transaction.py` imports pytest, defines six functions, and exits 0 having asserted nothing — a permanently green check. Every regression in `publish_blob_sources`/`atomic_copy` would ship undetected. This also violates the repo's op_test convention that every test file has a `__main__` guard.

**11. `test_jit_cache_transaction.py:105` — fault injection patches stdlib `os.replace`/`shutil.copy2` process-wide, and the stubs drop keyword-only params.**
`jit_cache.py` uses plain `import os`/`import shutil`, so `jit_cache.os is os`: `mock.patch.object(jit_cache.os, "replace", ...)` (109-111) and the `copy2` patch (166-168) replace the stdlib attribute for every in-process caller inside the `with` block — including CPython's `importlib._bootstrap_external._write_atomic` for lazily-compiled `.pyc` files and pytest's assertion-rewrite `.pyc` install. And `fail_staging_publish(source, destination)` (105) omits `os.replace`'s keyword-only `src_dir_fd`/`dst_dir_fd`, so any such internal call raises `TypeError: got an unexpected keyword argument 'src_dir_fd'` from unrelated stdlib code and surfaces as a confusing failure of this test. Same at line 161 for `copy2`'s `follow_symlinks`.

**12. `test_jit_cache_transaction.py:12` — `sys.path.insert(0, aiter/jit/utils)` is a permanent global mutation, and the bare `import jit_cache` shares a first-writer-wins `sys.modules` slot with `core.py`.**
That directory contains `cpp_extension.py`, `chip_info.py`, `file_baton.py`, `build_targets.py`, `torch_guard.py` and a `hipify/` package; under `pytest op_tests/` every module imported afterwards resolves `import cpp_extension`/`import hipify` to the aiter-internal copies, and the entry stays even if this module's import fails. Separately, core.py:24-28 also does `from jit_cache import ...` as a bare top-level name — in a CI layout with the wheel in site-packages and the source tree checked out, whichever imports first wins, so this test may exercise the wheel's copy while `aiter.jit.core` binds the checkout's. The test can silently not be testing the code under review.

### Minor

**13. `core.py:1032` — `_stage_blob_sources` is called outside the try/except, so codegen failures lose the formatted diagnostic.**
With `check=True` (jit_cache.py:33), a generator exiting non-zero — the exact scenario this PR targets — raises `subprocess.CalledProcessError` at core.py:1032, ~80 lines above the `except Exception as e` at 1113. Users lose the colorized `failed jit build [md_name]` banner, the highlighted traceback, and the `build [md] under {opbd_dir} failed` message, and callers catching `RuntimeError` from `build_module` now see an unexpected exception type. Moving the staging call inside the `try` (tracking the staging dir for the `finally`) preserves the diagnostics.

**14. `core.py:1036` — the random staging path lands in `__FILE__`/DWARF, making `.so` files non-reproducible.**
With `hipify=False` the compiler's source argument and `-I` are `{op_dir}/.blob-<6 random chars>/...`, which end up in `DW_AT_name`/`DW_AT_comp_dir` and in any `__FILE__`-based assert in generated code. Two builds of identical inputs now produce different `.so` bytes, undermining wheel-hash/artifact caching and prebuilt-kernel diffing; and `addr2line`/`llvm-symbolizer`/`gdb` on a crash in a generated kernel resolves to a directory that was renamed to `blob` immediately after the build, so the source is never found. A deterministic staging name fixes this and finding #1 together.

**15. Copyright header.** Both new files carry `Copyright (C) 2026, Ruye-aa. All rights reserved.` (`aiter/jit/utils/jit_cache.py:2`, `op_tests/test_jit_cache_transaction.py:2`). Every other Python file in the repo — including all of `aiter/jit/utils/` and 71 op_tests files — uses `Copyright (C) <years>, Advanced Micro Devices, Inc. All rights reserved.`, and `grep -rl Ruye` matches exactly these two files. Please switch to the AMD line before merge.

Ruye-aa:
@valarLip Thanks for the detailed review. I addressed the 15 findings in `24048aca`.

Summary of the changes:

- **#1, #2, #4, #8, #14:** codegen now uses the deterministic `{op_dir}/blob.staging` path. The working tree is seeded from the last published `blob`, reused across builds to preserve generated-file mtimes, and kept after both successful and failed compilation so `build.ninja`, `ninja` retries, DWARF paths, `addr2line`, and save-temps workflows remain valid. An incomplete/complete marker restores only interrupted codegen, while bounded cleanup reaps abandoned artifacts from the previous random-staging implementation and interrupted publication.
- **#3, #5, #7:** the compiled `.so` is atomically installed before source-cache publication. Publication is now best-effort, snapshots the stable staging tree without consuming it, verifies that the generation token did not change, restores the old cache on failure, tolerates a peer publisher winning the race, cleans candidates/backups, and preserves the directory mode inherited from `op_dir` rather than publishing a `0700` `mkdtemp` directory.
- **#6:** restored the original rebuild contract: level 1 removes the installed module and clears the build tree; level 2 removes only the installed module and keeps the incremental build cache.
- **#9:** `module_deepgemm_opus` now seeds `compiled_kids_opus.json` into the staging transaction, and `gen_instances.py` reads/writes the staged sidecar. It is published with the generated-source snapshot only after the binary has been installed.
- **#10, #11, #12:** converted the regression file to `unittest` with an executable `__main__` entrypoint, so AITER's `python3 op_tests/test_*.py` runner executes it. The test imports the checkout file through `importlib.util.spec_from_file_location` and patches module-local fault-injection aliases, without mutating `sys.path`, `sys.modules['jit_cache']`, `os.replace`, or `shutil.copy2` process-wide. Fault stubs also accept extra positional and keyword arguments.
- **#13, #15:** codegen failures now pass through the existing formatted JIT diagnostic and `RuntimeError` contract, and both new files use the AMD copyright line.
- Header-only `.h/.hpp/.cuh` output remains accepted, while a genuinely empty generator is still rejected.

Local validation on the new head:

```text
python3 op_tests/test_jit_cache_transaction.py        12/12 passed
python -m pytest -q op_tests/test_jit_cache_transaction.py
                                                       12 passed
ruff 0.16.0 check (changed Python files)                passed
black 25.11.0 --check (changed Python files)           passed
py_compile, JSON validation, and git diff --check      passed
```

I could not run the ROCm/GPU wheel matrix locally, so upstream Actions remain the required integration check.

The branch went through several main merges while Actions/review approval was pending, so this full feedback arrived later in the cycle. The delay affected when the issues were exposed; it did not cause the underlying defects.

@valarLip could you please re-review the new head? @zufayu could you please approve or coordinate the Actions runs for `24048aca` when available?

Ruye-aa:
@valarLip I rechecked findings #9, #8, #7 and #5. My earlier response overstated #9: staging the Opus sidecar did not update the file the tuner actually reads. This is fixed in `6c8cf02e`.

- **#9 — Opus sidecar:** The tuner still reads `{bd_dir}/compiled_kids_opus.json`; ordinary runtime dispatch uses the CSV/C++ lookup. After installing the new `.so`, JIT now atomically copies the successful generated sidecar back to that canonical path, independently of best-effort blob publication. The tuner passes new candidates through `--extra_kids` instead of advancing the canonical file before compiling. `clear_build()` removes only the per-module tree, so the successful canonical set survives and seeds the next build. The comment in `opus_gemm_common.py` has been updated accordingly. An adjacent receipt binds the sidecar hash to the installed binary's file identity; missing/stale metadata, a replaced binary, or an interrupted metadata write cannot satisfy the tuner's fast path. A copied/prebuilt binary without a matching local receipt conservatively rebuilds when tuning.
- **#8 — reclamation and disk budget:** New publish/backup/reset names record hostname and PID. The next build immediately reclaims artifacts of a dead local owner; live/remote owners are retained. The 24-hour grace remains for legacy names without owner information. Cleanup is build-triggered, not a timer. The fixed `blob.staging` is intentionally retained for Ninja retries and does not multiply with each killed compile. I measured the pinned CK commit `af9e1d1f`: one source tree occupies about 70.86 MiB for fwd, 90.02 MiB for varlen fwd, 28.66 MiB for filtered batch-prefill, 221.08 MiB for CK bwd, and 147.34 MiB for libmha_fwd on this filesystem. Keeping staging plus blob for the seven MHA configurations totals approximately 1.95 GiB, excluding HSA headers, objects, binaries and abandoned artifacts. Normal source storage is about 2S at rest and up to 3S during changed-snapshot publication. This is a real disk-budget cost, not a claim that every runner has enough space.
- **#7 — permissions:** The publish candidate is now chmod'ed to `op_dir`'s mode immediately after `mkdtemp`, before copying any contents. Tests cover the candidate, final blob and nested directories for 0755 and 0750 roots, with 0644 source files. Restrictive parent/source permissions are preserved rather than widened. This removes the unintended 0700 candidate/publication mode; a real root-prebuild/non-root container run is still an integration check.
- **#5 — publication failure:** The binary remains installed before publication. A test now drives the actual `build_module` control flow with a stub compiler and injects a peer winning the directory swap: the build returns successfully, emits a warning, retains the installed binary and updated Opus sidecar, and reaps its candidate/backup. A separate metadata-write failure test verifies the binary remains usable while the tuner rejects the stale receipt.

Validation:

- Script-mode CPU suite: 21 tests passed; pytest: 21 passed, 2 permission subtests passed.
- Ruff 0.16.0, Black 25.11.0, Python compilation and diff whitespace checks passed on the changed Python files.
- Real Opus codegen with `GPU_ARCHS=gfx942 --extra_kids 10006` included that additional kid, emitted 19 `.cu` files, and recorded 16 compiled kids. This was CPU code generation with the unavailable torch CUDA probe stubbed, not a ROCm compilation or kernel execution.

The storage measurements, sidecar lifecycle and permission conditions are documented in [docs/jit_cache.md](https://github.com/ROCm/aiter/blob/6c8cf02e/docs/jit_cache.md).

@zufayu could you please approve or coordinate the current-head CI runs? @valarLip could you please re-review these changes once the integration results are available?

Ruye-aa:
Follow-up correction after another audit of the pushed head `6c8cf02e`: there are still three gaps to address. The previous 21 passing CPU tests do not cover these paths.

1. **An Opus tuner request can be skipped after waiting for a different module build.** `mp_lock()` returns after a peer releases the module lock, without executing this caller's `MainFunc`. `_ensure_kids_compiled()` then treats the return from `build_module()` as success without checking the requested kids. CPU fault injection through the actual waiter/tuner control flow, with a stub compiler, reproduced: request kid 9; installed test artifact still represents kids `[1, 7]`; the receipt rejects kid 9; nevertheless the tuner returns `True`. The waiter needs to verify that the peer satisfied this particular request and build/retry when it did not.

2. **The receipt does not bind metadata to the generation actually compiled if builders overlap.** In the duplicate-builder scenario discussed in #5, staging can change after compilation and before metadata publication. The generation token is checked across snapshot publication, but not across compilation. Injecting a second codegen between the stub compile and publication reproduced an installed test artifact for `[1, 7]` with a published sidecar `[1, 9]`; the new receipt incorrectly accepted kid 9. This requires overlapping builders, not ordinary execution under an intact exclusive module lock. The build must retain and validate its own generation identity before installing/tagging its artifact, or otherwise guarantee that its input snapshot cannot change.

3. **A failed rollback deletes the last published backup.** If both candidate-to-blob replacement and backup-to-blob restoration raise, `publish_blob_sources()` still removes `backup_dir` in `finally`. Fault injection reproduced an absent canonical blob and no retained backup. The installed binary survives, but the last known-good source cache is lost. A backup whose restoration failed must be retained for recovery.

These are CPU control-flow/failure-injection reproductions, not GPU execution results. No further fix has been pushed for these findings yet. The current permissions tests still cover 0755/0750 inheritance, but root-to-non-root container validation remains pending; stable staging/remote-owner retention also still requires an explicit disk budget.

The PR head is confirmed pushed at `6c8cf02e`. Ten workflows currently remain `action_required`, and review is still `CHANGES_REQUESTED`. Please treat this as a correction to any implication that all transaction/review concerns are closed.

Ruye-aa:
Fixed the three gaps reported in my [additional audit](https://github.com/ROCm/aiter/pull/4800#issuecomment-5567242351) in commit `46bf6373`:

- **Opus module-lock waiter:** the tuner now passes `build_after_wait=True`. After a runtime builder releases the module lock, the tuner acquires it and runs its own `--extra_kids` request. A peer's unrelated successful build no longer counts as completion. Ordinary callers keep the existing wait behavior. A failed metadata write after this invocation's successful compile does not cause an unbounded retry or fail the build.
- **Generation / receipt binding:** the build keeps the token returned by its own codegen and checks it before/after compilation and after copying the installation tempfile, before replacing the installed artifact. It snapshots the Opus sidecar before compilation and publishes those exact bytes, binding the receipt to the installed inode via an open descriptor. A later staging generation or a peer replacing the binary cannot be mistaken for the generation this invocation compiled. Source-cache publication also requires the original token. Pre-install generation mismatches fail the build; post-install cache/metadata publication failures remain warnings.
- **Failed rollback:** if both candidate publication and backup restoration fail, the backup is retained. The next locked codegen can recover a dead owner's backup or this same still-running process's retained backup. If recovery is still denied and `blob` is absent, cleanup preserves backups regardless of age or dead-owner status. The peer-wins publication path still preserves the peer's cache and cleans our redundant backup.

Validation: `python op_tests/test_jit_cache_transaction.py` passed all 32 tests; pytest reported `32 passed, 5 subtests passed`. Ruff, Black, `py_compile`, and `git diff --check` also passed. New fault-injection tests cover the waiter path, generation changes during compile/copy/after installation, replacement of the installed binary by a peer, double publication/rollback failure, same-process recovery, and cleanup while recovery remains denied. The existing permissions, clear-build, incremental-source, and best-effort publication cases still pass. The lifecycle and recovery boundaries are documented in `docs/jit_cache.md`.

These are CPU-only control-flow/filesystem tests with the compiler stubbed, not ROCm/GPU or root-to-non-root container acceptance. Generation checks still rely on the module lock and cooperating generation markers; they do not protect against arbitrary external edits bypassing those markers.

@zufayu could you approve/coordinate the pending workflow runs, and @valarLip could you re-review this revision when possible? The three audited gaps are now fixed and tested locally; upstream CI and review approval are still needed.

Ruye-aa:
@valarLip I completed another pass over the original 15 findings, the follow-up sidecar/storage/permissions/publication concerns, and the current JIT/tuner implementation. Commit [`4e984907`](https://github.com/ROCm/aiter/pull/4800/commits/4e9849074fc243b1d476c4b75a4f98c39c91d8c4) includes additional fixes beyond the previously pushed `46bf6373`.

### Additional fixes in this revision

- **Fixed-target compilation and retries:** the real Python extension versioner could produce `module_deepgemm_opus_v1.so` while `build_module` installed the old unversioned file. Its pre-compilation input cache could also suppress a same-input retry after compiler failure. `build_module` now bypasses that versioner and lets Ninja check the fixed target on every invocation. This preserves its `PyInit` name and handles header changes/deleted outputs; default direct-loader versioning is unchanged. A real CPU C++/Ninja test confirms unchanged inputs perform no compilation and a changed header rebuilds.
- **Tuner rebuild lifecycle:** an existing receipt no longer bypasses an unfulfilled explicit rebuild. Successful preparation clears the inherited environment flag on all reuse paths, while failures restore it. A final cancellation test exposed another gap: `KeyboardInterrupt`/`SystemExit` were treated as success by the old `finally`. An explicit success flag now restores the original environment for those cases too; six failure-first subcases cover both interruptions with unset/1/2 environment values.
- **Requested kids and architecture filtering:** the tuner passes only actual candidates as explicit requests; the generator supplies target-specific heuristic defaults. Unknown, off-target or family-filtered explicit kids now fail codegen before compilation instead of returning success with those kids absent. The runtime-probe fallback respects explicit `GPU_ARCHS`, and the family filter no longer re-adds off-architecture heuristic defaults.
- **Interrupted restoration:** restoring directly into `blob.staging` could leave a partial tree that the next attempt treated as complete. Recovery now copies into a permission-correct candidate and atomically installs it only when complete. Tests include a child process exiting during the copy, followed by successful recovery. Legacy backup formats are also recoverable immediately when the canonical blob is missing; their deletion grace does not delay recovery.
- **Documentation:** the PR description and Opus README now describe the actual canonical sidecar path, synchronous tuner preparation, fixed staging, rebuild exceptions, storage budget and integration limits. The old unique-staging/five-test description is no longer accurate.

### Original review coverage

| Findings | Current implementation and checks |
| --- | --- |
| #1, #2, #4, #14 | Stable `blob.staging` paths; preserved unchanged-file mtimes; staging remains available after success or compile failure. Tests cover source reuse and real CPU Ninja incrementality. Stable paths remove this source of randomness, not a claim of fully reproducible GPU binaries. |
| #3, #5 | Binary installation precedes best-effort cache publication. A peer winning publication produces a warning, not a failed build. Generation/installed-inode checks and retained failed-rollback backups remain covered. |
| #6 | Level 1 clears the module build tree and removes its installed binary; level 2 removes only the binary. A regression checks that level-2 objects and the canonical sidecar survive. |
| #7 | Publish and recovery candidate roots inherit `op_dir` mode immediately, and final blob/nested source modes are tested for 0755/0750 roots. Actual root-to-non-root container execution remains pending. |
| #8 | One retained staging path does not accumulate per killed compile. Next-build cleanup immediately reaps dead local owners; live/remote owners remain protected. Legacy deletion uses a 24-hour grace. Source budget is about 2S at rest/3S while copying; the documented seven-MHA measurement is about 1.95 GiB for the two retained trees, excluding HSA headers, objects, binaries and abandoned artifacts. Runner disk sufficiency must be checked separately. |
| #9 | The tuner reads `{bd_dir}/compiled_kids_opus.json`; runtime dispatch reads CSV/C++ tables. JIT publishes the successful canonical sidecar plus a binary-bound receipt after installation, independently of blob publication. The canonical set survives `clear_build`, seeds the next build, and is not advanced by failed compilation. Same-process rebuild, stale receipt, waiter, strict requested-kid and cancellation paths are covered. |
| #10, #11, #12 | Executable unittest entrypoint, checkout-specific module loading, and local fault-injection aliases. The normal tests do not globally shadow `jit_cache` or replace stdlib filesystem functions. |
| #13, #15 | Checked codegen errors retain the formatted `RuntimeError` diagnostic/cause; new files use the AMD copyright header. |

Header-only HSA generator output remains valid; genuinely empty generated input is still rejected.

### Validation and remaining gates

```text
python op_tests/test_jit_cache_transaction.py            53 tests passed
python -m pytest -q op_tests/test_jit_cache_transaction.py
                                                       53 passed, 21 subtests passed
Ruff, Black --check, py_compile, config JSON, diff check  passed
```

The suite now exercises the actual JIT/versioner/FileBaton control flow with HIP compilation stubbed, plus the real CPU C++/Ninja build and real CPU Opus generator described above. These do not establish ROCm/GPU or cross-UID container acceptance. Cooperative module locking remains required; explicit rebuilds retain the existing removal window for unrelated readers.

@zufayu, could you please approve or coordinate the current-head Actions runs promptly? @valarLip, could you please re-review this revision? Local fixes and tests are complete for the identified cases, but upstream CI and review approval are still required. Delayed approval lengthens the feedback cycle; it is not the cause of the underlying code defects.

valarLip (approved):
LGTM

PR #4805, opened 2026-08-17
feat(ep): add an TDM dispatch by dsl for gfx1250

XingerZhu:
NOT YET VERIFIED ON HARDWARE. It compiles and the ISA is what it should be, but every GPU run so far has died on a wedged driver before producing a correctness result, so it lands behind a switch that defaults off. Do not enable it without running the test below first.

Port of mori's ep_intranode_1250x dispatch to FlyDSL. The vector transport sends a token as per-lane 16B copies and takes one remote atomic per route; this one counts routes into a block-local LDS histogram, claims each peer's recv range with one remote atomic per (block, peer), gathers the metadata into destination-ordered staging runs, and moves both the metadata and the payload with the Tensor Data Mover.

dispatch_transport picks between the two and defaults to vector, so existing callers are untouched; AITER_EP_DISPATCH_TDM=1 flips the auto default for A/B runs. The two leave interchangeable state behind, so combine and the routing handle do not care which ran. scale_dim, push_group and fp4 are rejected rather than silently mishandled, and replay is refused outright: its slots come from a per-(block, peer) reservation, so a cached map is only replayable under the same grid AND the same block->token assignment.

tdm_prims builds the descriptors over raw 64-bit addresses because make_tensor_descriptor_2d derives the global base from a fly memref, while a payload's destination is a peer's lsa_ptr -- a runtime integer, not a kernel argument. Note tensor_dim1_stride, which flydsl leaves zero as "no higher-dim strides": that holds for its own dim1 >= 2 tiles, but mori sets the field on every descriptor it builds, including the 1 x hiddenDim payload, and a 1xN tile with a zero dim1 stride is the one shape gfx1250's own contract says does not exist. Metadata tiles go through mori's TdmCheapDim1 closed form so they cannot reach dim1 == 1 at all.

What was checked without a GPU, via op_tests/flydsl_tests/compile_dispatch_tdm.py: five geometries compile; the ISA carries the intended 4 tensor_load_to_lds, 11 tensor_store_from_lds and 1 ds_add_u32 + 1 ds_add_rtn_u32; the descriptor constants in the LLVM IR match the intended tile geometry; LDS comes out at 114816 B for the 7168-hidden 8-warp case; and nothing spills at 33 VGPR / 74 SGPR. test_dispatch_tdm.py is the correctness and A/B timing test that still has to run.

## Submission Checklist

### DIFF
--- a/aiter/ops/flydsl/dispatch_combine_v2/dispatch_combine_op.py
+++ b/aiter/ops/flydsl/dispatch_combine_v2/dispatch_combine_op.py
@@ -1,0 +1,1730 @@
+# Copyright © Advanced Micro Devices, Inc. All rights reserved.
+#
+# MIT License
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


kawhil-amd:
What about the performance compare with cu version?

PR #4808, opened 2026-08-17
[CI] Triton tests selection script

Boss2002n:
PR CI ran the whole op-type folder for any kernel edit. Select the paired
test for each edited kernel instead, plus every test whose imports reach
that module, so fused coverage is computed rather than maintained by hand.
A GEMM kernel edit now runs 3 tests instead of 25; a kernel with no paired
test falls back to its op-type folder, and shared machinery (utils/,
package __init__, CI infra) still falls open to the full suite.

Configs are read at run time rather than imported, so the import graph
cannot say which variants consume them: a config change runs its whole op
folder plus any test outside it that imports the module the family is
named after. Tuning attention/mha therefore also runs test_mha_with_pe
and test_mha_with_sink.


zufayu:
## [ROCm/aiter] PR #4808 — [CI] Triton tests selection script

**Replaces the never-enabled networkx test-selection script with a stdlib diff-based selector (name-paired test + import-closure reachability + op-folder fallback, failing open to the full suite on anything it cannot map) and wires it into the Triton CI workflow with a ci:triton-full label and empty-shard reporting.**

⚠️ [verified] Deleting (or heavily rewriting/moving) a Triton test file reddens CI: `.github/scripts/select_triton_tests.py` adds every changed `test_*.py` path to the selection without checking it exists in the head tree, so a deleted path such as `op_tests/triton_tests/attention/test_old.py` reaches `split_tests.sh`, which exits 1 with "Selection lists paths that are not test files" and fails the split_triton_tests job — the replaced script's `git_filename_diff` filtered diff paths through `abs_path.exists()`, a guard the rewrite dropped. The failure is loud, not silent, but every legitimate test-deletion PR fails the Triton Test workflow until it is labeled `ci:triton-full`. **Author must** skip deleted paths (intersect the diff with head-tree files, or select only the importers that remain).

📝 [verified] The import closure only follows the `aiter.ops.triton` and `op_tests.triton_tests` roots (`IMPORT_ROOTS` in `.github/scripts/select_triton_tests.py`), so a future test that reaches a triton kernel only through a non-triton aiter wrapper (e.g. `aiter.ops.shuffle`) would be silently unselected. A sweep of all 394 triton sources against an unrestricted import graph found zero such tests on the current tree, and main pushes still run the full suite, so this is latent — but nothing detects the assumption drifting. **Reviewer should ask** the author to state the invariant next to `IMPORT_ROOTS` so a future test routing through a non-triton wrapper is a known break of it.

📝 [verified] A benchmark-only PR now produces a green Triton Test run with zero tests (empty selection, every shard uploads an empty `<testsuites/>` report), where the full suite previously ran. The step summary does record "benchmark — no unit tests selected", and CI never ran the benchmarks either way, so no kernel coverage is lost. **Reviewer should confirm** that a green zero-test check is the intended contract for benchmark-only diffs.

Boss2002n:
> ## [ROCm/aiter] PR #4808 — [CI] Triton tests selection script
> **Replaces the never-enabled networkx test-selection script with a stdlib diff-based selector (name-paired test + import-closure reachability + op-folder fallback, failing open to the full suite on anything it cannot map) and wires it into the Triton CI workflow with a ci:triton-full label and empty-shard reporting.**
> 
> Review (advisory): ⚠️ NEEDS WORK Validation (deterministic): N/A — no runtime surface changed Perf (advisory): N/A — no runtime surface changed
> 
> ⚠️ [verified] Deleting (or heavily rewriting/moving) a Triton test file reddens CI: `.github/scripts/select_triton_tests.py` adds every changed `test_*.py` path to the selection without checking it exists in the head tree, so a deleted path such as `op_tests/triton_tests/attention/test_old.py` reaches `split_tests.sh`, which exits 1 with "Selection lists paths that are not test files" and fails the split_triton_tests job — the replaced script's `git_filename_diff` filtered diff paths through `abs_path.exists()`, a guard the rewrite dropped. The failure is loud, not silent, but every legitimate test-deletion PR fails the Triton Test workflow until it is labeled `ci:triton-full`. **Author must** skip deleted paths (intersect the diff with head-tree files, or select only the importers that remain).
> 
> 📝 [verified] The import closure only follows the `aiter.ops.triton` and `op_tests.triton_tests` roots (`IMPORT_ROOTS` in `.github/scripts/select_triton_tests.py`), so a future test that reaches a triton kernel only through a non-triton aiter wrapper (e.g. `aiter.ops.shuffle`) would be silently unselected. A sweep of all 394 triton sources against an unrestricted import graph found zero such tests on the current tree, and main pushes still run the full suite, so this is latent — but nothing detects the assumption drifting. **Reviewer should ask** the author to state the invariant next to `IMPORT_ROOTS` so a future test routing through a non-triton wrapper is a known break of it.
> 
> 📝 [verified] A benchmark-only PR now produces a green Triton Test run with zero tests (empty selection, every shard uploads an empty `<testsuites/>` report), where the full suite previously ran. The step summary does record "benchmark — no unit tests selected", and CI never ran the benchmarks either way, so no kernel coverage is lost. **Reviewer should confirm** that a green zero-test check is the intended contract for benchmark-only diffs.

resolved

PR #4809, opened 2026-08-17
Fix lean attention for gfx950

nidal567:
## Motivation

`persistent_lean_attention()`, the public entry point, and the only caller of `_get_config()`, did not execute on any architecture. Three defects fired before the kernel launched:

|  | defect | symptom | fix |
|---|---|---|---|
| 1 | `arch_info.get_num_sms()` does not exist — it lives in `utils.device_info` | `AttributeError`, every arch | import from where it lives |
| 2 | `**config` splatted into a launch that already passes those keys by name | `TypeError: got multiple values for keyword argument 'XCD_REMAP'` | filter consumed keys, forward the rest |
| 3 | no `gfx950-LEANATTN-DEFAULT.json` (only `gfx942` exists) | `FileNotFoundError` on gfx950 | add a tuned gfx950 config |

Defects 1 and 2 are present identically in the pinned reference clone, so neither is a regression. They survived because nothing exercised the path: `bench_la.py` passes launch parameters that were hand-picked straight to the internal `_persistent_lean_attention` and never reads a config, and the only tests touching the public entry point are `@pytest.mark.skip`.

The benchmark also hardcodes `total_programs=608` — `2 × 304`, a **gfx942** CU count. gfx950 has 256 CUs, so that grid is wrong for this GPU: correcting it is worth 20–29% on Triton 3.8 and 24–35% on Triton 3.4, and it penalizes both compilers, so it is not a compiler issue.

## Technical Details

4 files were changed and a config file was introduced

- **`attention/lean_atten.py`** - import `get_num_sms` from `utils.device_info`; drop the unused `arch_info` import. Filter config keys the wrapper already handles via `explicit_config_keys_handler`. Everything else is forwarded as `**launch_options`.
- **`_triton_kernels/attention/lean_atten.py`** - `_get_config(causal, n_ctx_q)` selects a `"decode"` or `"causal"` entry, falling back to `"any"` so existing arch files are unaffected. It keys on `n_ctx_q`, not `causal`, because the wrong entry is *illegal*, not just slow: `Mp`/`Lp` are validated against `BLOCK_M` (so `BLOCK_M ≤ n_ctx_q`), and causal asserts `BLOCK_M % BLOCK_N == 0`, which `BLOCK_M=16` with `BLOCK_N=64` violates. An earlier revision keyed on `causal` and crashed the GPU on non-causal prefill; the new test caught it.
- **New `configs/gfx950-LEANATTN-DEFAULT.json`** - Needed separate tuning for causal and decode.
- **`bench_la.py`** - opt-in `--use-config` so the shipped path is measurable. Without the flag the benchmark takes the same `x_vals` branch and builds an identical launch (same `total_programs`, `BLOCK_M/N`, `num_warps`, `waves_per_eu`,  XCD_REMAP`)
- **`test_la.py`** - new `test_persistent_lean_attention_config_path`, the first *executable* caller of the public entry point (the pre-existing one is skipped). There was an issue where it bounds the torch reference's peak memory: it materialised the whole score matrix - 8 GiB of fp16 scores plus two live 16 GiB fp32 buffers (the `.float()` copy and the softmax output), a measured **40.9 GiB peak for one test**, which OOMs when anything shares the GPU — and now chunks over heads. Tests go 22 → 24 collected.

Two config values were chosen from measurement, not copied from gfx942:

- **`BLOCK_N=64` and `waves_per_eu=4` are both required for decode.** Changing only
  `waves_per_eu` is catastrophic at `BLOCK_N=128` (+295% / +256%), and the grid is sharply
  non-monotonic (`BLOCK_N=256, num_warps=8, waves_per_eu=4` measured +1031%), so it cannot
  be tuned one axis at a time.
- **`matrix_instr_nonkdim` is omitted.** gfx942 sets it to 16, pinning MFMA to 16×16; on
  gfx950 that made shape 0 **2.39× slower** - 0.82 -> 1.97 ms kernel-only, trimmed mean of
  50 dispatches (`measurements/nonkdim_{with,without}`). Only shape 0 was measured.

### Deliberately not in this PR

- **`sm_scale` is accepted and ignored** - the wrapper never forwards it and
  `_persistent_lean_attention` hardcodes `q.shape[-1] ** -0.5`. This is out of scope.

## Test Plan

1. **Correctness** - full `test_la.py`.
2. **Routing** - each of the 12 benchmark shapes selects the intended config entry, and the launched grid is the 512 the config asks for wherever the workload fills it.
3. **gfx942 back-compat** - with `arch_info` reporting gfx942: config resolution for all three regimes, the launch-option filter, the new test's skip logic, and an end-to-end run using gfx942's config values.
4. **Benchmark default mode** - `bench_la.py` builds an identical launch without `--use-config`.
5. **Reference-memory equivalence** - all-heads vs chunked, peak allocation and checksum.
6. **Performance** - `rocprofv3 --kernel-trace`, kernel-only time for `la_persistent` over all 12 shapes. **250 dispatches per cell, bottom and top 10% discarded, mean of the middle 80%**, from per-dispatch timestamps. Variants run **adjacent in time per shape** with **order alternated** across 3 reps, deltas computed within each pair - 72 runs / 18,000 dispatches. A sweep-per-variant layout produced a spurious result that reversed sign under interleaving.
7. **Lint** - `ruff check`, `black --check`.

## Test Result

- Correctness: 22 passed, 2 skipped

- Routing: 12/12 shapes select the intended entry. Shapes 0–8 get the prefill tiling
(`BLOCK_M=128, BLOCK_N=64`), shapes 9–11 the decode tiling (`BLOCK_M=16, BLOCK_N=64`).
Launched `total_programs` is 512 everywhere except shape 3, where the workload clamps both
variants to 144.

- gfx942 back-compat: 7 static checks plus 2 GPU checks pass
(driver run on gfx950 with `arch_info` reporting gfx942 given that there's no hardware for gfx942 accessible):

| check | result |
|---|---|
| `_get_config` on the real gfx942 JSON, all 3 regimes | falls back to `"any"`, no `KeyError` |
| `"any"` is load-bearing | removing it raises `KeyError`, as claimed |
| launch-option filter | `launch_options={'matrix_instr_nonkdim': 16}` — forwarded, no consumed key leaked |
| new test, gfx942 decode case | **skips**, does not `ValueError` (guard present, `BLOCK_SIZE_M=128 > 16`) |
| new test, gfx942 prefill case | legal by inspection (`128 ≤ 1024`, `128 % 64 == 0`); not run as a test |
| gfx942 config values end-to-end (GPU) | runs with `nonkdim=16` forwarded |
| numerics with `nonkdim=16` (GPU) | matches the torch reference at `atol=1e-2, rtol=3e-3` |

**Performance — baseline = `bench_la.py`'s hardcoded `x_vals`; config = public entry point
reading the new gfx950 file. Both on Triton 3.8.**

| # | mode | batch | hq | hk | n_ctx_q | n_ctx_k | d | baseline (ms) | config (ms) | delta | 95% CI | verdict |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 0 | causal | 1 | 32 | 8 | 8192 | 8192 | 128 | 0.9829 | 0.7901 | −19.62% | [−19.97, −19.26] | improvement |
| 1 | causal | 1 | 64 | 8 | 8192 | 8192 | 128 | 2.0010 | 1.5632 | −21.88% | [−22.32, −21.44] | improvement |
| 2 | causal | 1 | 128 | 8 | 8192 | 8192 | 128 | 4.4271 | 3.1706 | −28.37% | [−30.64, −26.10] | improvement |
| 3 | causal | 1 | 32 | 16 | 1024 | 1024 | 128 | 0.0543 | 0.0543 | +0.01% | [−2.05, +2.08] | okay |
| 4 | causal | 1 | 64 | 16 | 1024 | 1024 | 128 | 0.0784 | 0.0512 | −34.69% | [−35.72, −33.66] | improvement |
| 5 | causal | 1 | 128 | 16 | 1024 | 1024 | 128 | 0.0997 | 0.0752 | −24.55% | [−26.88, −22.22] | improvement |
| 6 | causal | 1 | 32 | 32 | 2048 | 2048 | 128 | 0.0938 | 0.0753 | −19.66% | [−21.46, −17.86] | improvement |
| 7 | causal | 1 | 64 | 32 | 2048 | 2048 | 128 | 0.1621 | 0.1253 | −22.73% | [−24.09, −21.37] | improvement |
| 8 | causal | 1 | 128 | 32 | 2048 | 2048 | 128 | 0.2997 | 0.2422 | −19.18% | [−20.39, −17.96] | improvement |
| 9 | decode | 512 | 32 | 8 | 16 | 8192 | 128 | 13.0404 | 7.1403 | **−45.25%** | [−45.66, −44.83] | improvement |
| 10 | decode | 512 | 64 | 8 | 16 | 8192 | 128 | 25.5451 | 13.6942 | **−46.39%** | [−46.58, −46.21] | improvement |
| 11 | decode | 512 | 128 | 8 | 16 | 8192 | 128 | 50.2278 | 26.7873 | **−46.67%** | [−47.06, −46.28] | improvement |

**11 improvements, 0 regressions.** This replaces the three decode regressions
(+44.21% / +11.03% / +2.97%) that motivated the retune. Two campaigns run days apart
agree on decode to within 0.9pp (−20.97 / −40.94 / −45.97 vs −21.43 / −40.88 / −45.10).

**Numerics.** `waves_per_eu` is a scheduling hint and cannot change results — shape 9,
where it is the only change, is bitwise identical across `wpe=2` and `wpe=4`. Shapes 10 and
11 also move `BLOCK_N`, reassociating the softmax accumulation; the archived evidence for
those two is the suite passing at `atol=1e-2, rtol=3e-3`.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/_triton_kernels/attention/lean_atten.py
+++ b/aiter/ops/triton/_triton_kernels/attention/lean_atten.py
@@ -27,14 +27,27 @@ from aiter.ops.triton.utils.core import AITER_TRITON_CONFIGS_PATH, load_config_j
 # Support tensor in [B, Seqlen, H, d] format. Taking tensors in [B*Seqlen, H, d] as inputs
 
 
-def _get_config():
+def _get_config(causal: bool = True, n_ctx_q: int | None = None):
     # No lru_cache here: load_config_json already caches the parse, and
     # caching the .copy() would hand every caller the same mutable object.
     dev = arch_info.get_arch()
     config = load_config_json(
         f"{AITER_TRITON_CONFIGS_PATH}/{dev}-LEANATTN-DEFAULT.json"
     )
-    return config["any"].copy()  # fresh copy per call — safe for callers to mutate
+    # Decode and prefill want different tilings; arch files without the split just
+    # have "any". Keyed on n_ctx_q, not causal: BLOCK_M must fit the query (Mp/Lp are
+    # validated against it) and causal needs BLOCK_M % BLOCK_N == 0.
+    prefill = config.get("causal", config["any"])
+    decode = config.get("decode")
+    if (
+        not causal
+        and decode is not None
+        and n_ctx_q is not None
+        and n_ctx_q < prefill["BLOCK_SIZE_M"]
+        and decode["BLOCK_SIZE_M"] <= n_ctx_q
+    ):
+        return decode.copy()
+    return prefill.copy()  # fresh copy per call
 
 
 @triton.jit
--- a/op_tests/op_benchmarks/triton/bench_la.py
+++ b/op_tests/op_benchmarks/triton/bench_la.py
@@ -1,12 +1,22 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
 
+import argparse
 import sys
 
 import torch
 import triton
 
-from aiter.ops.triton.attention.lean_atten import _persistent_lean_attention
+from aiter.ops.triton._triton_kernels.attention.lean_atten import _get_config
+from aiter.ops.triton.attention.lean_atten import (
+    _persistent_lean_attention,
+    persistent_lean_attention,
+)
+from aiter.ops.triton.utils.device_info import get_num_sms
+
+# Take the launch params from the arch config and call the public entry point, so we
+# measure what ships instead of a swept point. Needs a config for the current GPU.
+USE_CONFIG = False
 
 configs = []
 configs.append(
@@ -315,6 +325,16 @@ def bench_lean_attention(
     provider,
     device="cuda",
 ):
+    if USE_CONFIG:
+        # Before BLOCK_N builds batch_num_block_n and total_programs sizes the
+        # buffers, which the public entry point then validates against.
+        cfg = _get_config(causal=causal, n_ctx_q=n_ctx_q)
+        BLOCK_M = cfg["BLOCK_SIZE_M"]
+        BLOCK_N = cfg["BLOCK_SIZE_N"]
+        num_warps = cfg["num_warps"]
+        waves_per_eu = cfg["waves_per_eu"]
+        total_programs = get_num_sms() * cfg["SM_CNT_FACTOR"]
+
     n_ctx = n_ctx_k * batch
     assert batch == len(n_ctx)
 
@@ -358,25 +378,41 @@ def bench_lean_attention(
     XCD_REMAP = True
 
     # Triton LeanAttention output
-    fn = lambda: _persistent_lean_attention(
-        q,
-        k,
-        v,
-        Mp,
-        Lp,
-        Op,
-        locks,
-        batch_num_block_n,
-        total_programs,
-        BLOCK_M,
-        BLOCK_N,
-        XCD_REMAP,
-        causal,
-        batch,
-        RAGGED_BATCH,
-        num_warps,
-        waves_per_eu,
-    )
+    if USE_CONFIG:
+        fn = lambda: persistent_lean_attention(
+            q,
+            k,
+            v,
+            Mp,
+            Lp,
+            Op,
+            locks,
+            batch_num_block_n,
+            batch,
+            sm_scale=d**-0.5,
+            causal=causal,
+            RAGGED_BATCH=RAGGED_BATCH,
+        )
+    else:
+        fn = lambda: _persistent_lean_attention(
+            q,
+            k,
+            v,
+            Mp,
+            Lp,
+            Op,
+            locks,
+            batch_num_block_n,
+            total_programs,
+            BLOCK_M,
+            BLOCK_N,
+            XCD_REMAP,
+            causal,
+            batch,
+            RAGGED_BATCH,
+            num_warps,
+            waves_per_eu,
+        )
 
     warmup = 1
     rep = 1
@@ -387,6 +423,18 @@ def bench_lean_attention(
 
 
 def main():
+    global USE_CONFIG
+    parser = argparse.ArgumentParser(
+        prog="Benchmark Lean Attention", allow_abbrev=False
+    )
+    parser.add_argument(
+        "--use-config",
+        action="store_true",
+        help="Take launch parameters from the arch's tuned config and call the "
+        "public persistent_lean_attention(), instead of sweeping the tuning "
+        "columns of x_vals. Requires a <arch>-LEANATTN-DEFAULT.json.",
+    )
+    USE_CONFIG = parser.parse_args().use_config
     bench_lean_attention.run(save_path=".", print_data=True)
 
 
--- a/op_tests/triton_tests/attention/test_la.py
+++ b/op_tests/triton_tests/attention/test_la.py
@@ -13,8 +13,12 @@ from aiter.ops.triton.attention.lean_atten import (
     persistent_lean_attention,
 )
 from aiter.ops.triton.utils._triton import arch_info
+from aiter.ops.triton.utils.device_info import get_num_sms
 
 DEBUG_MODE = False
+
+# Heads per chunk in the torch reference; bounds its peak score-matrix memory.
+ref_head_chunk = 8
 
 
 def get_lean_attn_inputs(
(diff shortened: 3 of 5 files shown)


aiter/ops/triton/_triton_kernels/attention/lean_atten.py
```diff
@@ -27,14 +27,27 @@
...
-def _get_config():
+def _get_config(causal: bool = True, n_ctx_q: int | None = None):
     # No lru_cache here: load_config_json already caches the parse, and
     # caching the .copy() would hand every caller the same mutable object.
     dev = arch_info.get_arch()
     config = load_config_json(
         f"{AITER_TRITON_CONFIGS_PATH}/{dev}-LEANATTN-DEFAULT.json"
     )
-    return config["any"].copy()  # fresh copy per call — safe for callers to mutate
+    # Decode and prefill want different tilings; arch files without the split just
```
brunomazzottiamd, line 37: I fully understand that prefill and decode workloads want distinct tilings. However, the logic to select decode or prefill entries is very convoluted... This seems to be fully AI-generated. Can you please improve the implementation? I mean, do not change the outcome, but make it more readable and understandable. For instance, you can use Boolean variables to better identify a condition, naming it. Comments written by a human being are also welcome.

I'd assert that `config["any"]` entry is in the dictionary, otherwise `config["any"]` would throw a `KeyError`.

op_tests/triton_tests/attention/test_la.py
```diff
@@ -501,6 +512,73 @@ def test_persistent_lean_attention_outer(
...
+def test_persistent_lean_attention_config_path(
+    causal, batch, hq, hk, n_ctx_q, n_ctx, d
+):
+    torch.cuda.empty_cache()
+    torch.manual_seed(20)
+
+    dev = arch_info.get_arch()
+    try:
+        config = _get_config(causal=causal, n_ctx_q=n_ctx_q)
+    except FileNotFoundError:
```
brunomazzottiamd, line 534: Why are you handling this exception here in the unit test? Let the test break if the config file is absent instead of skipping it.

`_get_config` calls `load_config_json` utility function with `required=True`, so the file is expected to exist.

op_tests/triton_tests/attention/test_la.py
```diff
@@ -501,6 +512,73 @@ def test_persistent_lean_attention_outer(
...
+        pytest.skip(f"no {dev}-LEANATTN-DEFAULT.json to exercise the config path")
+
+    # An arch file with only "any" (gfx942) has no decode entry, so the decode case
```
brunomazzottiamd, line 537: Can you add non-optimized `causal` and `decode` entries to `aiter/ops/triton/configs/gfx942-LEANATTN-DEFAULT.json` instead of skipping the test?

I have the impression that `test_persistent_lean_attention_config_path` don't run in `gfx942`. If this is the intention then it would be better to add a top level `@pytest.mark.skipif(arch_info.get_arch() == "gfx942", reason="<state your reason here>")`.

op_tests/triton_tests/attention/test_la.py
```diff
@@ -13,9 +13,13 @@
     persistent_lean_attention,
 )
 from aiter.ops.triton.utils._triton import arch_info
+from aiter.ops.triton.utils.device_info import get_num_sms
 
 DEBUG_MODE = False
 
+# Heads per chunk in the torch reference; bounds its peak score-matrix memory.
+ref_head_chunk = 8
```
brunomazzottiamd, line 21: `ref_head_chunk` is only used by `reference_attention` function. Reduce the scope of `ref_head_chunk` declaring it in `reference_attention`.

op_tests/triton_tests/attention/test_la.py
```diff
@@ -501,6 +512,73 @@ def test_persistent_lean_attention_outer(
...
+
+    dev = arch_info.get_arch()
+    try:
+        config = _get_config(causal=causal, n_ctx_q=n_ctx_q)
+    except FileNotFoundError:
+        pytest.skip(f"no {dev}-LEANATTN-DEFAULT.json to exercise the config path")
+
+    # An arch file with only "any" (gfx942) has no decode entry, so the decode case
+    # gets the prefill tile and Mp/Lp validation would fail.
+    if config["BLOCK_SIZE_M"] > n_ctx_q:
```
brunomazzottiamd, line 539: Is it possible to implement a safe fallback in `_get_config` function if `config["BLOCK_SIZE_M"] > n_ctx_q`?

op_tests/op_benchmarks/triton/bench_la.py
```diff
@@ -387,6 +423,18 @@ def bench_lean_attention(
 
 
 def main():
+    global USE_CONFIG
```
brunomazzottiamd, line 426: Just a comment, you can keep the code as-is: [Global Variables Are Bad](https://wiki.c2.com/?GlobalVariablesAreBad)

In the case of this small bench script, it isn't a crime. However, the situation can escalate pretty quickly if you leave an unsupervised LLM do all the work.

brunomazzottiamd (changes requested):
Please check my comments in `aiter/ops/triton/_triton_kernels/attention/lean_atten.py` and `op_tests/triton_tests/attention/test_la.py`. I'm not sure if we can do something quick about it, but the selection logic to pick `causal` or `decode` tuning is hard to understand.

op_tests/triton_tests/attention/test_la.py
nidal567: I agree with this. The file is expected to exist, but this was done based off the fact that the config was already hard-coded importing Triton's defaults, completely skipping the already existing config file for gfx942

aiter/ops/triton/_triton_kernels/attention/lean_atten.py
nidal567: I agree. The outcome is the same and using Booleans would have been the better route for this. I can work on commenting better in future work.

op_tests/triton_tests/attention/test_la.py
nidal567: I did not have hardware to test this implementation on for gfx942. Although adding decode entries would change what gfx942 routes to at runtime, no? A "decode" entry would take from the dedicated "decode" bucket instead of the general "any" that it has been set to, and I assume it was tuned or optimized earlier at that time. Regardless, it wasn't ever meant to use gfx942, as the config was not launched or used in the original ship of this code. It's brute-forced

brunomazzottiamd:
@nidal567, please put all efforts on this PR on hold. We're evaluating if lean attention Triton kernel should be removed from AITER.

brunomazzottiamd:
Deprecated in favor of https://github.com/ROCm/aiter/pull/4887.

brunomazzottiamd:
https://github.com/ROCm/aiter/pull/4887 was merged, closing this PR.

PR #4811, opened 2026-08-17
fix(fmoe): retune GLM-5 FP8 decode kernels

### LINKED ISSUE #4788: [Issue] [fused_moe] GLM-5.2-FP8 per_1x128 blockscale: tuned 1-stage config silently produces garbage (gsm8k=0) on gfx942; `AITER_BYPASS_TUNE_CONFIG=1` fixes it
## Environment

- aiter: `44146e6ef` (includes #3422, #4040, #4558, #4675)
- composable_kernel submodule: `f33252ceb` (aiter pin, verified aligned)
- Model: GLM-5.2-FP8, served via a vLLM-style engine, `-tp 8`, `--kv_cache_dtype fp8`
- MoE dtype/quant: `per_1x128` fp8 blockscale (`torch.float8_e4m3fnuz`), g1u1 SiLU
- MoE shape (per tuned CSV row): `model_dim=6144, inter_dim=256, expert=257, topk=9`

## Summary

Serving GLM-5.2-FP8 on gfx942 yields **gsm8k = 0** (flexible ~0.008 / strict 0.000), with **no crash, no NaN, no assert** — silent garbage. Setting **`AITER_BYPASS_TUNE_CONFIG=1`** restores **gsm8k = 0.9454 / 0.9431**.

- Reproduces at high concurrency + long prompt (conc64, nshot5). Not DCP-related (`-dcp 1` and `-dcp 8` both = 0).
- During the failing eval, the engine logs repeatedly:
  ```
  [aiter] [fused_moe] using 1stage (kernelName1='..._fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256...', kernelName2=nan) for ('gfx942', ...)
  ```
  i.e. the **tuned config selects the `run_1stage=1` path** (kernel `fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256`).

## Where the tuned 1-stage entry comes from (git blame)

The tuned rows routing this shape to `run_1stage=1` / 64x256 were introduced by **#3422** ("Add GLM-5.1 FP8 blockscale GEMM/FMoE tunings for gfx942 (MI300X/MI325)", commit `6b831fb28`). `aiter/configs/model_configs/a8w8_blockscale_tuned_fmoe_glm5_1.csv`:

```
gfx942,304,1024, 6144,256,257,...,per_1x128,...,fmoe_bf16_blockscaleFp8_g1u1_vs_silu_64x256,   ...,run_1stage=1,...
gfx942,304,2048, 6144,256,257,...,per_1x128,...,fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256,...,run_1stage=1,...
gfx942,304,4096, 6144,256,257,...,per_1x128,...,fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256,...,run_1stage=1,...
gfx942,304,8192, 6144,256,257,...,per_1x128,...,fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256,...,run_1stage=1,...
gfx942,304,16384,6144,256,257,...,per_1x128,...,fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256,...,run_1stage=1,...
```

Before #3422 there was no tuned config for this shape → the default (2-stage/heuristic) path was used, which works. #4040/#4558 did **not** modify these large-token rows (they added decode T≤256 a16 kernels).

## Isolated microbenchmark

We benchmarked the high-level `fused_moe()` (the serving path; reads the tuned CSV; honors `AITER_BYPASS_TUNE_CONFIG`) at the exact tuned shape against the fp32 reference `torch_moe_blockscale`, using robust metrics (rel-L2 / cosine — the built-in `checkAllclose(rtol=atol=0.05)` is too tight for fp8 and flags ~24% of elements even for correct kernels):

| weights layout | tuned config (selects 64x256) | `AITER_BYPASS_TUNE_CONFIG=1` (selects 32x256/2-stage) |
|---|---|---|
| **preshuffled** `shuffle_weight(w, (16,16))` | cos **0.9997**, rel-L2 0.023 (M=32…8192) | cos **0.9997** |
| **not preshuffled** | cos **≈0**, rel-L2 ≈1.43 (garbage) | cos **≈0** (garbage) |

Two facts from this:
1. The `..._vs_ps_silu_64x256` kernel is **not numerically broken** — with correctly preshuffled weights it is accurate up to M=8192 (cos 0.9997).
2. `fused_moe` **silently produces garbage (cos≈0)** when weights are not in the expected preshuffle layout. Only a warning is logged (`... is_shuffled=False ... preshuffle_off may produce incorrect results`); there is **no hard-fail**.

Note: this isolated benchmark reproduces the preshuffle-sensitivity and kernel-correctness above; it does **not** by itself reproduce the serving gsm8k=0 (in isolation tuned and bypass behave identically per layout). The serving failure is pinned by the instrumentation below.

## Root cause (confirmed via serving instrumentation)

We added temporary logging to `fused_moe` in the live server and ran the same short eval twice (tuned vs `AITER_BYPASS_TUNE_CONFIG=1`), gsm8k: tuned=**0**, bypass=**0.9062**.

- **The weights are identical in both runs**: `is_shuffled=True`, `w1(257,512,6144)` / `w2(257,6144,256)` fp8_e4m3fnuz, activation passed as bf16 (`a1_scale=None`, internal quant). The frontend's weight layout is **not** the variable.
- **The only difference is dispatch**:
  - prefill (token 1024/16384): tuned = `run_1stage=1, block_m=64, kernelName1=..._vs_ps_silu_64x256` (forced by the #3422 tuned cfg); bypass = `run_1stage=True, block_m=32, kernelName1=''` (heuristic).
  - decode (token 16…256): tuned = `run_1stage=1, xbf16=True, a16 kernels`; bypass = `run_1stage=False` (2-stage, ksplit>0).

Since (a) the frontend passes one fixed `is_shuffled=True` weight layout that the **default/heuristic path handles correctly** (bypass = 0.9062), and (b) the isolated microbench proved the 64x256 kernel is numerically correct **when fed weights preshuffled with aiter's own `shuffle_weight(w,(16,16))`**, the conclusion is:

**The specific asm kernels forced by the #3422 gfx942 tuned config expect a weight-preshuffle layout that is incompatible with the layout the frontend actually applies (`is_shuffled=True`). The tuned path therefore reads the preshuffled weights under the wrong layout → silent garbage. The default (untuned) dispatch is layout-compatible.**

The failure is doubly silent: the existing safeguard at `fused_moe.py:2381`
```python
if not is_shuffled and not run_1stage:
    logger.warning("... is_shuffled=False ... preshuffle_off may produce incorrect results")
```
is **short-circuited for the tuned path**, because `run_1stage=True` there — so no warning is emitted even when the layout is wrong.

(One direct check remains open: we have not byte-compared the frontend's shuffled weights against `shuffle_weight(unshuffled, (16,16))`. The inference — same kernel + same block_m is correct under `(16,16)` but wrong under the frontend layout — is strong, but a byte-level dump would make it airtight.)

## Reproduction (isolated, self-contained)

Run in-container on gfx942:

```python
import os, torch, torch.nn.functional as F
from einops import rearrange
import aiter
from aiter import dtypes, pertoken_quant, ActivationType, QuantType
from aiter.fused_moe import fused_topk, fused_moe
from aiter.ops.shuffle import shuffle_weight
torch.manual_seed(0); torch.set_default_device("cuda"); dtype = dtypes.bf16

def torch_moe_blockscale(hs, w1, w2, tw, tid, dtype, scale_blks=(128,128), a_scale=None, fc1=None, fc2=None):
    ct = dtypes.fp32; hs, w1, w2 = hs.to(ct), w1.to(ct), w2.to(ct)
    tn, tk = tid.shape; E, md, idim = w2.shape; B, D = hs.shape; tk = tw.shape[1]
    bn, bk = scale_blks
    if a_scale is not None:
        hs = (hs.view(tn, -1, bk) * a_scale.unsqueeze(-1)).view(tn, -1)
    hs = hs.view(tn, 1, md).repeat(1, tk, 1)
    out = torch.zeros((B, tk, D), dtype=ct)
    nbk = md // bk
    fc1 = rearrange(fc1.view(-1,1).repeat(1,bn*bk).view(E,-1,nbk,bn,bk), "e a b c d -> e (a c)(b d)")
    fc2 = rearrange(fc2.view(-1,1).repeat(1,bn*bk).view(E,nbk,idim//bn,bk,bn), "e a b c d -> e (a c)(b d)")
    w1, w2 = w1*fc1, w2*fc2
    for e in range(E):
        m = tid == e
        if m.sum():
            ai = hs[m] @ w1[e].T
            g, u = ai.split([idim, idim], -1)
            out[m] = (F.silu(g)*u) @ w2[e].T
    return (out * tw.view(B,-1,1)).sum(1).to(dtype)

def build(M, md=6144, idim=256, E=257, tk=9, bn=128, bk=128):
    inp = torch.randn((M, md), dtype=dtype)
    w1 = torch.randn((E, idim*2, md), dtype=dtype)/10
    w2 = torch.randn((E, md, idim), dtype=dtype)/10
    tw, tid = fused_topk(inp, torch.randn((M,E),dtype=dtype), tk, True)
    def bq(w, n, k):
        t = rearrange(w.view(-1, w.shape[1]//n, n, w.shape[2]//k, k), "e a b c d -> e a c (b d)").contiguous()
        q, s = pertoken_quant(t, quant_dtype=dtypes.fp8)
        q = rearrange(q.view(-1, w.shape[1]//n, w.shape[2]//k, n, k), "e a b c d -> e (a c)(b d)").contiguous()
        return q, s.view(w.shape[0], -1)
    w1q, w1s = bq(w1, bn, bk); w2q, w2s = bq(w2, bn, bk)
    aq, asc = pertoken_quant(inp.view(-1, md//bk, bk), quant_dtype=dtypes.fp8)
    return dict(inp=inp, aq=aq.view(-1,md), asc=asc.squeeze(-1), w1q=w1q, w1s=w1s, w2q=w2q, w2s=w2s, tw=tw, tid=tid)

def run(M, preshuffle=True):
    d = build(M)
    ref = torch_moe_blockscale(d["aq"], d["w1q"], d["w2q"], d["tw"], d["tid"], dtype,
                               fc1=d["w1s"], fc2=d["w2s"], a_scale=d["asc"])
    w1 = shuffle_weight(d["w1q"], (16,16)) if preshuffle else d["w1q"]
    w2 = shuffle_weight(d["w2q"], (16,16)) if preshuffle else d["w2q"]
    out = fused_moe(d["inp"], w1, w2, d["tw"], d["tid"], w1_scale=d["w1s"], w2_scale=d["w2s"],
                    activation=ActivationType.Silu, quant_type=QuantType.per_1x128)
    a, b = out.float().flatten(), ref.float().flatten()
    cos = F.cosine_similarity(a, b, 0).item()
    print(f"BYPASS={os.environ.get('AITER_BYPASS_TUNE_CONFIG','0')} preshuffle={preshuffle} M={M} cos={cos:.5f}")

for M in (32, 1024, 4096, 8192): run(M, preshuffle=True)
for M in (1024, 4096, 8192):     run(M, preshuffle=False)
```

Run:
```bash
python repro.py                       # tuned:   preshuffle cos≈0.9997 ; no-preshuffle cos≈0
AITER_BYPASS_TUNE_CONFIG=1 python repro.py   # bypass:  same pattern
```

## Suggested fixes (for discussion)

1. **Fail loud, not silent.** When a tuned `run_1stage` kernel is selected but the weight layout does not match its preshuffle requirement, **raise** instead of only `logger.warning` + returning garbage. Silent wrong output at gsm8k=0 is the worst failure mode.
2. If the tuned 1-stage path requires a different preshuffle layout than the default path, either **document/enforce** the exact layout, or make dispatch auto-shuffle to the layout the selected kernel needs.
3. Consider re-validating the gfx942 GLM-5.x tuned entries from #3422 for **numerical correctness under the actual serving weight layout**, not just timing (`err1=0.0%` in the CSV is a tuning-time timing record, not a correctness check).

## Related (existing issues/PRs — none cover this exact case)

Same *family* (silent `fused_moe` layout incorrectness / tuned-kernel-only-fails-in-server), but distinct from this bug:

- **#2194** (issue) / **#4542** (PR) — silent wrong layout in `fused_moe`, but trigger is a **lost `is_shuffled` tag** (attr dropped by clone/`to`/`nn.Parameter`…) → inferred `is_shuffled=False`. Here `is_shuffled=True` is correctly set; the tuned asm kernel just expects a *different* preshuffle layout. #4542's explicit `is_shuffled=` arg does **not** fix this (our tag is already True).
- **#4783** (issue) — same model/arch/quant (GLM-5.2, gfx942, per_1x128 blockscale MoE) but a **crash** (M∈{304,336}, no tuned config → 2-stage OOB), not silent garbage.
- **#4032** (issue) — same pattern (tuned config picks a bad kernel; fails only in live server, fine in isolation; heuristic avoids it), but gfx950 / M=1 / NaN.
- **#3565** / **#3518** — precedent for the fix direction (raise a clear error / route to a compatible kernel instead of silent collapse; #3565 notes the same gsm8k→0.000 symptom class).
- **#3853** (issue) — gfx942/cu304 fp8 fused-MoE tuned-config *coverage* request. Complementary: that asks to add coverage; this reports that existing GLM coverage (#3422) is functionally wrong.
- **#3422** (merged) — introduced the gfx942/GLM tuned rows that route this shape to the run_1stage asm kernels.

## Workaround

`AITER_BYPASS_TUNE_CONFIG=1` (MoE-only; bypasses the tuned config → default kernel) → gsm8k restored to 0.9454/0.9431.

## Open questions for maintainers

- What exact weight-preshuffle layout do the #3422-tuned gfx942 asm kernels (`..._vs_ps_silu_64x256`, `..._a16_..._16x256`, etc.) expect? It appears to differ from the layout marked `is_shuffled=True` by a standard frontend that the default/untuned path handles fine.
- Should the `is_shuffled` safeguard also cover the `run_1stage=True` path (currently short-circuited at `fused_moe.py:2381`), and hard-fail (or auto-reshuffle) on layout mismatch rather than returning silent garbage?
- Were the #3422 gfx942 GLM-5.x tuned entries validated for numerical correctness (not just timing) against the layout produced by a real serving stack?

### Operating System

Ubuntu 24.04.4 LTS (Noble Numbat)

### CPU

AMD EPYC 9654 96-Core Processor

### GPU

MI300X (gfx942), 8-GPU

### ROCm Version

ROCm 7.2.4

### ROCm Component

_No response_

### Steps to Reproduce

_No response_

<details>

```
Paste output here
```

</details>

### Additional Information

_No response_

akii96:
Motivation

Fixes https://github.com/ROCm/aiter/issues/4788.

The existing gfx942 tuned FMoE rows for GLM-5.2-FP8 at token buckets 32, 64, 128, and 256 select a16 16x256 kernels that can silently produce garbage during TP8 serving. This collapses GSM8K accuracy, while AITER_BYPASS_TUNE_CONFIG=1 restores it.

Changes

Retune the affected shape while excluding the problematic a16 kernels and the unsafe two-stage stage-1 ASM family (dead kernel) that tuner currently points to:

- model_dim=6144
- inter_dim=256
- expert=257
- topk=9
- per_1x128 FP8 blockscale
- gfx942, 304 CUs

The replacement rows use numerically validated one-stage FP8 32x256 kernels. All unrelated tuning rows remain unchanged.

Validation

Production-operator validation passed for all four retuned buckets:

| Tokens | Selected kernel                                    |
|--------|----------------------------------------------------|
| 32     | fmoe_bf16_blockscaleFp8_g1u1_vs_silu_1tg_32x256    |
| 64     | fmoe_bf16_blockscaleFp8_g1u1_vs_silu_1tg_ps_32x256 |
| 128    | fmoe_bf16_blockscaleFp8_g1u1_vs_silu_1tg_ps_32x256 |
| 256    | fmoe_bf16_blockscaleFp8_g1u1_novs_silu_1tg_32x256  |

The merged tuning file was also validated by serving the full GLM-5.2-FP8 model with TP8:

- deterministic math completion: correct (12 apples)
- GSM8K, 5-shot, 256 examples, concurrency 64:
  - flexible exact match: 0.9453
  - strict exact match: 0.9453
- no AITER_BYPASS_TUNE_CONFIG workaround required

The replacement kernels are approximately 12–15% slower for these four individual FMoE invocations, in exchange for restoring correct model output. Other decode and prefill tuning rows are unaffected.


yitingw1:
Verified locally that this PR can restore GLM-5.2 FP8 accuracy.

valarLip (approved):

PR #4829, opened 2026-08-18
[CI] fix: lazily dequantize moe reference weights

gyohuangxin:
## Summary
- Lazily dequantize per-expert MXFP4 weights in the MoE torch reference path
- Avoid materializing all expert weights as fp32 at once in torch_moe_stage1 and torch_moe_stage2
- Preserve the existing dequant formula and only change when each expert weight is materialized

## Testing
- python3 -m py_compile aiter/fused_moe.py
- git diff --check

### DIFF
--- a/.github/scripts/aiter_test.sh
+++ b/.github/scripts/aiter_test.sh
@@ -35,7 +35,6 @@ else
 fi
 
 skip_tests=(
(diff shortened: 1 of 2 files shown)


yifehuan (dismissed):
LGTM

yifehuan (approved):
LGTM

PR #4839, opened 2026-08-18
[HIP] [Bugfix] Guard against negative expert ids in MoE sorting P0

### LINKED ISSUE #4827: [Bug] MoeSortingMultiPhaseKernel_P0_v1: missing lower bound on signed expert id causes GPU memory fault under EP (one-line fix)
## Summary

`MoeSortingMultiPhaseKernel_P0_v1` reads the expert id out of `topk_ids` into a **signed** `IndexType` and only bounds-checks it from above:

```cpp
IndexType eid = x[j.value];        // IndexType = int (signed)
if(eid < kargs.num_experts)        // no lower bound
    p_expert_mesh[eid * mesh_stride + curr_token_id] = (curr_topk_id + 1) & 0xffff;
```

Under **expert parallelism** the ids of experts owned by other ranks can be negative. A negative `eid` still passes `eid < num_experts`, so `eid * mesh_stride` becomes a **negative offset** and the store writes *before* `p_expert_mesh` — GPU memory access fault.

This matches the observed behaviour exactly: it only ever reproduces with EP enabled, and never without it.

## Fix

```diff
-                if(eid < kargs.num_experts)
+                if(eid >= 0 && eid < kargs.num_experts)
```

One line. Verified on the exact configuration that used to abort every time (details below). The same unchecked pattern is worth auditing in the other `MoeSortingMultiPhaseKernel_*` phases.

## Symptom before the fix

8x MI355X, vLLM serving DeepSeek-V4-Pro (384 routed experts, top-6, MXFP4), DP8 + EP + DSpark speculative decoding, `--max-num-seqs 512`. All workers die during CUDA graph capture:

```
Memory access fault by GPU node-3 (Agent handle: 0x...) on address 0x7f836a527000. Reason: Unknown.
```

`ROCr Debug Agent` with `--precise-memory` attributes it to:

```
void aiter::opus_moe_sorting_entry<
    aiter::MoeSortingMultiPhaseKernel_P0_v1<
        aiter::MoeSortingProblemMp<int, float, unsigned char, 4, true, false, true>>, ...>
code object: .../aiter/jit/module_moe_sorting_opus.so
 => <+1068>: s_or_b64 exec, exec, s[2:3]
```

## Verification

Same run, same flags, only the one-line change:

| | before | after |
|---|---|---|
| memory access faults | 8 (every worker) | **0** |
| server startup | dies in CUDA graph capture | **READY** |
| GSM8K (200 samples, 5-shot) | n/a | **0.955** |
| decode TPOT (batch=1) | n/a | **13.4 ms** (30.6 ms without spec decode) |
| DSpark acceptance length | n/a | **2.66** |

The multi-phase path is fully exercised — no dispatch policy override, no reduction in `max_num_seqs`.

## What isolates it

| config | TP | MoE | max_num_seqs | before fix |
|---|---|---|---|---|
| TP8 + EP | 8 | EP8 | 512 | ok |
| DP2xTP4 + EP | 4 | EP8 | 512 | ok |
| DP2xTP4, no EP | 4 | TP4 | 512 | ok |
| DP8 + EP, no spec decode | 1 | EP8 | 512 | ok |
| **DP8 + EP + spec decode** | **1** | EP8 | **512** | **fault** |
| DP8 + EP + spec decode | 1 | EP8 | 384 | ok |

`max_num_seqs` and the CUDA-graph requirement are secondary — they only control whether a batch that carries a negative expert id reaches this kernel.

## Notes for anyone digging further

- A standalone reproducer calling `moe_sorting(...)` directly did **not** trigger it (11 values of M, `num_local < M` variants, and CUDA-graph capture/replay divergence all clean). It needs the real multi-GPU EP traffic that produces the negative ids.
- Things that looked plausible but were **not** the cause: clamping the device-side token count to `kargs.tokens`; forcing `moe_sorting_mesh_byte_size` to 4; bounding the mesh write by `num_experts * kargs.mesh_stride`; clamping `total_elem`; `AITER_BF16_FP8_MOE_BOUND=0`.
- Separately, `P0_v2` recomputes `mesh_stride` from the run-time token count while the host sized the workspace (and placed `p_expert_cumsum` right after the mesh region) from `h.tokens`. Not the trigger here, but the host/device disagreement looks like a latent issue on its own.
- `moe_sorting_mesh_byte_size` switching the mesh element 4B -> 1B at `tokens_ >= 512` is intentional (`ws(512) == ws(128)` is self-consistent), and `dispatch_policy=1` returning 0 is intentional. An earlier revision of this report wrongly flagged both.

## Environment

- aiter: upstream `main` @ `65a2e34d`
- ROCm 7.2.4, torch `2.11.0+gitd0c8b1f` (hip 7.2.53211)
- 8x AMD Instinct MI355X (gfx950)
- Model: DeepSeek-V4-Pro (384 routed experts, top-6, MXFP4 experts), EP8

Happy to open a PR with the one-liner if useful.

xudonlyu:
Fixes #4827.

`MoeSortingMultiPhaseKernel_P0_v1` reads the expert id out of `topk_ids` into a **signed** `IndexType` and only bounds-checks it from above:

```cpp
IndexType eid = x[j.value];        // IndexType = int (signed)
if(eid < kargs.num_experts)        // no lower bound
    p_expert_mesh[eid * mesh_stride + curr_token_id] = (curr_topk_id + 1) & 0xffff;
```

A negative `eid` still satisfies `eid < num_experts`, so `eid * mesh_stride` becomes a negative offset and the store writes *before* `p_expert_mesh`, faulting the GPU.

## Where the negative id comes from

vLLM writes it deliberately. Its topk kernels mark padded rows with `-1` so
downstream consumers can skip them — `csrc/libtorch_stable/moe/topk_softmax_kernels.cu`:

```cpp
indices[idx] = is_pad_row ? static_cast<IndType>(-1)
                          : (should_process_row ? (expert - start_expert) : num_experts);
```

This is the `VLLM_MOE_SKIP_PADDING` path and it is **on by default**. Both
sentinels on that line are out-of-range on purpose: `num_experts` for a row this
rank does not own, `-1` for a padded row. The existing `eid < num_experts` check
is already filtering the first one, so this change is the other half of the same
contract rather than a new policy.

Instrumented a DP8 + EP run to confirm it at the source. The first hit is during
`profile_run`, where every row is padding:

```
NEG topk_ids: min=-1 count=24576 uniq=[-1] shape=(4096, 6)
  profile_run -> _dummy_run -> ... -> rocm_aiter_fused_experts
```

`uniq=[-1]` and a count equal to the whole tensor — the entire batch is padded,
not a stray value. That also explains the timing: the fault only ever appeared
during cuda graph capture and profiling, which is exactly when every row is
padding.

Note this corrects the earlier description of this PR, which attributed the
negative ids to expert parallelism. It is not EP: a row owned by another rank is
written as `num_experts`, a positive value the upper bound already rejects. EP
only mattered because that configuration is what reaches the multi-phase path.

## Repro

8x MI355X (gfx950), vLLM serving DeepSeek-V4-Pro (384 routed experts, top-6), DP8 + EP + DSpark speculative decoding, `--max-num-seqs 512`. Every worker aborted during CUDA graph capture:

```
Memory access fault by GPU node-3 (Agent handle: 0x...) on address 0x7f836a527000. Reason: Unknown.
```

`ROCr Debug Agent` with `--precise-memory` attributed it to this kernel:

```
void aiter::opus_moe_sorting_entry<
    aiter::MoeSortingMultiPhaseKernel_P0_v1<
        aiter::MoeSortingProblemMp<int, float, unsigned char, 4, true, false, true>>, ...>
```

## Verification

Same run, same flags, only this change:

| | before | after |
|---|---|---|
| memory access faults | 8 (every worker) | **0** |
| server startup | dies in CUDA graph capture | **READY** |
| GSM8K (200 samples, 5-shot) | n/a | **0.955** |
| decode TPOT (batch=1) | n/a | **13.4 ms** (30.6 ms without spec decode) |
| DSpark acceptance length | n/a | **2.66** |

The multi-phase path is exercised unchanged — no `dispatch_policy` override and no reduction in `max_num_seqs`. Confirmed `moe_sorting_opus_get_workspace_size(M, 385, 6, 0) != 0` for the M values in play, i.e. `moe_sorting_opus_mp()` is the path taken.

## Scope

Three other sites in `moe_sorting_opus.h` index the mesh or the token map by a
raw `topk_ids` value with no lower bound — around lines 794, 1021 and 2160. They
take the same input under the same contract, so they have the same hole; I only
touched the site I could attribute the fault to. Happy to extend this PR to
cover them if that is preferred.

Note I could not build with `clang-format-18` in my environment — the change is a single condition plus a comment, so I kept the surrounding formatting untouched. Happy to reformat if CI flags it.


xudonlyu:
Chased the negative id to its source rather than clamping it blind — description updated with the details.

Short version: vLLM writes it on purpose. `topk_softmax_kernels.cu` marks padded rows with `-1` so downstream consumers skip them, and that path (`VLLM_MOE_SKIP_PADDING`) is on by default. Both out-of-range values on that line are sentinels: `num_experts` for a row this rank does not own, `-1` for a padded row. This kernel already filters the first one with `eid < num_experts`, so the lower bound is the other half of a contract it is following, not a new guard.

Instrumented a DP8 run to confirm: the first hit is in `profile_run`, `uniq=[-1]` with a count equal to the whole tensor — every row is padding there, which is also why the fault only ever showed up during profiling and graph capture.

This also corrects my original description, which blamed expert parallelism. That was wrong: a row owned by another rank is written as `num_experts`, a positive value.

yifehuan (dismissed):
LGTM, or maybe you can condense comment (1671~1676) to a single line

yifehuan (approved):

tuukkjs:
Can this be merged? I have a follow-up PR coming up

xudonlyu:
@yifehuan Could we get this PR merged?

PR #4845, opened 2026-08-19
[HIP] Support mixed-dtype inputs in biased grouped top-k

BadrBasowid:
## Motivation

Kimi-K3 produces BF16 router logits while keeping the correction bias in FP32. AITER’s existing biased grouped top-k kernel assumes both tensors have the same dtype, requiring vLLM to convert the bias to BF16 before routing.

This change independently dispatches the logits and bias types, allowing the kernel to consume BF16 logits with FP32 bias directly. It preserves bias precision and removes the separate dtype-conversion operation from the routing path.

## Test Plan
Extended `op_tests/test_moeTopkSoftmax.py` to cover mixed-dtype biased grouped top-k.

## Test Result

All explicitly reported mixed-dtype cases from 1 through 65,536 tokens passed the weight and expert-index correctness checks.

## Submission Checklist

### DIFF
--- a/aiter/ops/topk.py
+++ b/aiter/ops/topk.py
@@ -213,6 +213,28 @@ def biased_grouped_topk(
         )
 
 
+def biased_grouped_topk_mixed_dtype(
+    gating_output: torch.Tensor,
+    correction_bias: torch.Tensor,
+    topk_weights: torch.Tensor,
+    topk_ids: torch.Tensor,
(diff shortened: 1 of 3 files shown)


tjtanaa:
@valarLip Do you know who is suitable to review this small PR? Thank you. It is an optimization for vLLM Kimi K3.

PR #4205, opened 2026-07-12
Add OPUS gfx950 bf16 fmha d192x128 kernel

shay-li77:
## Motivation

DeepSeek V3 / V3.2 MLA prefill decompresses its latent KV into per-head K(192) / V(128), i.e. a standard causal MHA with an asymmetric head dim (D_QK=192, D_V=128). The default fmha_v3 backend leaves performance on the table for this shape, especially in the varlen (group) path used by flash_attn_varlen_func. This PR integrates the gfx950 OPUS asymmetric-head-dim flash-attention forward kernel into aiter and routes the (192,128) bf16 case to it, giving sizable speedups on the DeepSeek prefill shapes while falling back to fmha_v3/CK everywhere else.

## Technical Details
- **Shared OPUS forward entry point.** Generalized the former D=128-only opus path into a single `fmha_fwd_bf16_opus_fwd` (module `module_fmha_fwd_bf16_opus`) that dispatches by head dim in C++: the symmetric D=128 kernel (batch only, logic unchanged) or the new **D_QK=192/D_V=128** kernel (batch + group/varlen). Both kernels are compiled in one translation unit.
- **New kernel-only headers** `fmha_fwd_hd192_v128_bf16_opus_{defs.h,kernel.hpp,.h}` (ported from the standalone OPUS source), plus the shared host launcher `fmha_fwd_bf16_opus_kernels.cu` that builds kargs, computes the grid + the causal head/tail-merge `opt` bit, and launches the right (causal × group) instantiation.
- **Python wrappers.** `fmha_fwd_bf16_opus_fwd` (batch, dense BSHD) and `fmha_fwd_bf16_opus_varlen_fwd` (packed THD) over one low-level `@compile_ops` stub; wired into `_flash_attn_forward` / `_flash_attn_varlen_forward`. D=192 is **enabled by default** (`AITER_DISABLE_FMHA_OPUS=1` to opt out); D=128 stays env-gated (`AITER_ENABLE_FMHA_OPUS=1`). A `gen_fake` meta impl makes the op torch.compile / export / cudagraph safe.
[...]

PR #4847, opened 2026-08-19
[HIP] [GFX950]Fix OPUS MHA TypeError Issue

shay-li77:
## Motivation

The OPUS (192,128) forward is enabled by default on gfx950 and crashes under vLLM with `TypeError: fmha_fwd_bf16_opus_fwd(): incompatible function arguments` (reported on #4205). `aiter_tensor_t` is a pybind11-registered class owned by `module_aiter_core`, but `module_fmha_fwd_bf16_opus` is excluded from the wheel prebuild and JIT-built at runtime — when the two `.so` files end up with different `PYBIND11_INTERNALS_ID` they no longer share a pybind type registry, the registration is invisible, and every tensor argument is rejected.

## Technical Details

Switch the entry point to `at::Tensor` / `std::optional<at::Tensor>` and drop `develop=True`, matching every other JIT-built pybind module in aiter: torch's own type caster only does a `THPVariable_Check()` and never touches the pybind registry, so it cannot break across a toolchain boundary. The declaration moves to `csrc/include/torch/`, and the device guard / stream now come from the tensor (`at::hip::getCurrentHIPStream`), which makes `AITER_SET_STREAM_PYBIND` unnecessary; no kernel or launch-logic changes.

## Test Plan

* `pytest op_tests/test_mha.py -k opus`.
* The reporter's repro (`flash_attn_varlen_func`, bf16 D_QK=192/D_V=128, causal) in `vllm/vllm-openai-rocm` on the stock `amd-aiter 0.1.19` wheel — i.e. prebuilt `module_aiter_core.so` plus a runtime-JIT OPUS module — additionally with a deliberately mismatched pybind ABI (`-DPYBIND11_COMPILER_TYPE=\"_gcc\"` on this module only) to force the two `.so` files into separate type registries.

## Test Result

* **50 passed**, no numerical change.
* Repro passes. With the injected mismatch (`..._v11_gcc_...` vs `..._v11_system_...`) it reproduced the reported `TypeError` verbatim before this change and passes after. Env: gfx950 (MI355), ROCm 7.2.2.

## Submission Checklist

### DIFF
--- a/aiter/ops/mha.py
+++ b/aiter/ops/mha.py
@@ -352,7 +352,6 @@ def gen_fmha_fwd_bf16_opus_fwd_fake(
     "module_fmha_fwd_bf16_opus",
     fc_name="fmha_fwd_bf16_opus_fwd",
     gen_fake=gen_fmha_fwd_bf16_opus_fwd_fake,
-    develop=True,
 )
 def _fmha_fwd_bf16_opus_fwd(
     q: Tensor,
--- a/csrc/include/torch/fmha_fwd_bf16_opus.h
+++ b/csrc/include/torch/fmha_fwd_bf16_opus.h
@@ -9,7 +9,7 @@
 // This replaces the former per-hd `fmha_fwd_hd128_bf16_opus_fwd`. The kernel/launch
 // logic for D=128 is unchanged (only moved under this shared entry point).
 #pragma once
-#include "aiter_tensor.h"
+#include <torch/extension.h>
 #include <optional>
 
 // Dense (batch) & varlen (group) GQA/MHA scaled-dot-product attention, bf16, gfx950.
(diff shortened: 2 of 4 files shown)


fxmarty-amd (approved):
Can you add a test in `op_tests/test_mha.py`?

This broke an important model on vLLM (when not using the opt-out `AITER_DISABLE_FMHA_OPUS=1`) side due to maybe a missing test in https://github.com/ROCm/aiter/pull/4205.

```python
"""Reproduce AITER gfx950 OPUS varlen FlashAttention binding failure.

Run inside the ROCm container with its default Python, for example:

    HIP_VISIBLE_DEVICES=0 python repro_opus.py

With the affected AITER build, this raises ``TypeError`` from
``fmha_fwd_bf16_opus_fwd`` even though all inputs are valid contiguous CUDA
tensors. To verify the fallback, run with ``AITER_DISABLE_FMHA_OPUS=1``.
"""

import inspect

import aiter
import torch
from aiter import flash_attn_varlen_func


def main() -> None:
    if not torch.cuda.is_available():
        raise RuntimeError("This reproducer requires a ROCm CUDA/HIP device.")

    device = "cuda"
    q = torch.randn(4, 16, 192, device=device, dtype=torch.bfloat16)
    k = torch.randn(4, 1, 192, device=device, dtype=torch.bfloat16)
    v = torch.randn(4, 1, 128, device=device, dtype=torch.bfloat16)
    cu_seqlens = torch.tensor([0, 4], device=device, dtype=torch.int32)

    print(f"torch: {torch.__version__}")
    print(f"aiter: {getattr(aiter, '__version__', 'unknown')}")
    print(f"GPU: {torch.cuda.get_device_name(0)}")
    print(f"flash_attn_varlen_func: {inspect.signature(flash_attn_varlen_func)}")

    output = flash_attn_varlen_func(
        q,
        k,
        v,
        cu_seqlens,
        cu_seqlens,
        max_seqlen_q=4,
        max_seqlen_k=4,
        softmax_scale=192**-0.5,
        causal=True,
    )
    torch.cuda.synchronize()
    print(f"PASS: {output.shape=}, {output.dtype=}, {output.device=}")


if __name__ == "__main__":
    main()
```

reproduced the issue on my end

shay-li77:
@fxmarty-amd A test in `op_tests/test_mha.py` can't cover this — the failure depends on *how the two
`.so` files were built*, not on any input shape. `aiter_tensor_t` is registered by
`module_aiter_core` and is only visible to a second module while both share a
`PYBIND11_INTERNALS_ID`. CI builds everything together, so the IDs always match and any
shape-based test passes; that's why the 50 opus tests in #4205 were green while this was
broken.
I verified against the real failure mode instead: on the stock `amd-aiter 0.1.19` wheel in
`vllm/vllm-openai-rocm`, forcing the two `.so` files into separate registries
(`-DPYBIND11_COMPILER_TYPE=\"_gcc\"` on this module only) reproduces your `TypeError`
verbatim before the change and passes after.
If you want something in CI, the one angle that works is asserting the pybind signature
contains no `aiter_tensor_t` — happy to add that.

fxmarty-amd:
OK

shengnxu (approved):

PR #4851, opened 2026-08-19
[Triton/Gluon] [Config] Add gfx1250 A16W16 GEMM configs for FLUX.2 shapes

sogalin:
## Summary
Two commits. The first adds tuned `GEMM-A16W16` tile configs for the FLUX.2
projection shapes on gfx1250; without a matching config they fall back to
`BLOCK_M=32, BLOCK_N=64`. The second adds the `tuned_gemm` dispatch table that
actually routes those shapes to Triton — without it, lookups fall through to
the default libtype `torch` and FLUX.2 stays on hipBLASLt no matter how well
the Triton side is tuned.
Config-only; no kernel or Python changes, and nothing outside gfx1250 with
these exact shapes is affected.
## Backend choice is measured, not assumed
Every (M,N,K) FLUX.2-dev feeds to `tgemm` at 512x512, 1024x1024 and 2048x2048
was collected with `AITER_TUNE_GEMM=1` and timed both ways. Triton takes 23 of
the 27 shapes, by 1.5x to 9.4x, peaking at ~613 TFLOP/s on a 96 CU gfx1250
The four that stay on `torch` are the useful ones, because a plain "large M
means Triton" rule gets them wrong: the two M=1 modulation projections, and
M=1024 with an N or K of 128, where Triton is 2.2x to 2.6x slower.
## NUM_BUFFERS is deliberately capped at 2
`NUM_BUFFERS >= 3` benchmarks faster but is **not run-to-run deterministic** on
this architecture: repeated launches on identical inputs disagree on thousands
of elements. Every tile config here uses `NUM_BUFFERS=2`, which measured zero
mismatches. Please keep this in mind if these files are ever re-tuned.
## Validation
- Bit-exact against `torch` on all 27 shapes, and deterministic across repeated
  launches; Triton is only selected where both hold.
- End-to-end FLUX.2-dev on gfx1250,
  12 steps, median of repeated runs on an otherwise idle GPU, measured with the
  companion SGLang change that routes linear layers through `tgemm`:
  | Resolution | hipBLASLt | aiter | Speedup |
  | --- | --- | --- | --- |
  | 512x512 | 1.08 s/step | 0.22 s/step | 4.9x |
  | 1024x1024 | 3.06 s/step | 0.55 s/step | 5.6x |
  | 2048x2048 | 11.71 s/step | 2.39 s/step | 4.9x |

## Submission Checklist


Dewei-Wang-sh (approved):
lgtm

valarLip (approved):

PR #4859, opened 2026-08-19
[ASM] [HIP] [CI] Mxfp6 gemms

ksikiric:
## Motivation

Add production-ready MXFP6 A6W6 GEMM support for gfx950, initially targeting diffusion-model projection shapes while preserving numerical quality and minimizing dynamic activation quantization overhead. This establishes an entry point, and further optimizations and tuned shapes will follow later.

## Technical Details

- Adds MXFP6 E2M3 GEMM with per-1x32 E8M0 scaling.
- Adds fused Triton BF16-to-MXFP6 quantization, 32-point Hadamard rotation, and direct C0/C1 packing.
- Adds ctypes/JIT integration and public AITER APIs.
- Adds five ABI-compatible ASM scheduling variants sharing one packed layout.
- Adds shape-based tuned dispatch with exact/padded-shape lookup and a safe fallback.
- Adds a paired tuning guard requiring exact output parity and a minimum measured gain.
- Adds tuned/untuned configuration CSVs, manifest validation, and unit tests.

## Test Plan

- Build the A6W6 JIT module from a clean cache.
- Run aligned and padded-shape GEMM correctness tests.
- Run A6W6 tuning and dispatch unit tests.
- Compare tuned and default kernels across all 57 configured shapes.
- Re-run Wan projection benchmarks using:
  - 50 warmup iterations
  - 300 measured iterations
  - 7 repeats
- Compare GEMM latency, quantization latency, full-path latency, and cosine similarity against the saved baseline.

## Test Result

- Clean JIT build passed.
- 11 tuning unit tests passed.
- Aligned and padded GEMM correctness tests passed.
- 57/57 dispatch shapes produced exact output parity with zero mismatches.
- All 20 tuned shapes remained faster than the default kernel.
- Maximum tuning improvement: 10.48%.
- Geometric-mean improvement across all shapes: 1.21%.
- Smallest tuned gain reconfirmed over 15 repeats: 0.525%, positive in all repeats.
- MXFP6 cosine similarity against BF16 approximately 0.999192.

## Performance

Measured on one AMD Instinct MI355X (gfx950, 256 CUs). Weights are pre-quantized; full-path timings include dynamic activation quantization plus GEMM.

**Note:** Full Path here means quantization is included in the benchmark. GEMM-only is benchmarking only the kernel. 

| M×N×K | Full TFLOP/s FP8 / FP6 / FP4 | GEMM-only TFLOP/s FP8 / FP6 / FP4 |
|---|---:|---:|
| 9450×5120×5120 | 2302 / 2812 / 3376 | 2855 / 3437 / 3903 |
| 9450×13824×5120 | 2611 / 3184 / 4288 | 2934 / 3482 / 4736 |
| 9450×5120×13824 | 2360 / 3057 / 4344 | 3176 / 3857 / 5420 |

Geometric means:

- **MXFP6 vs FP8:** 1.25× full path, 1.20× GEMM-only
- **MXFP4 vs MXFP6:** 1.32× full path, 1.29× GEMM-only
- **Mean cosine similarity versus BF16 GEMM:** FP8 0.999294, MXFP6 0.999192, MXFP4 0.986704

### Wan2.2 xDiT end-to-end
Ulysses 8, 720×1280, 81 frames, 20 steps, AITER FP8 attention,
parallel VAE, and `torch.compile`.
| GEMM configuration | Time | Improvement vs FP8 |
|---|---:|---:|
| FP8 baseline | 21.54s | — |
| FP6 only | 21.06s | 2.23% |

## Submission Checklist


aiter/ops/gemm_op_a6w6.py
```diff
@@ -0,0 +1,735 @@
...
+
+
+def quant_mxfp6_torch(x: Tensor) -> tuple[Tensor, Tensor]:
+    """torch/GPU version of quant_mxfp6: [R,K] float -> (codes uint8, scales uint8)."""
+    x = x.float()
+    x = _rotate_k32_torch(x)
+    R, K = x.shape
+    NB = K // _SCALE_GROUP_SIZE
+    blk = x.reshape(R, NB, _SCALE_GROUP_SIZE)
+    amax = blk.abs().amax(dim=2)
```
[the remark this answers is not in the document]
ksikiric: Fixed in 4b97209

aiter/ops/gemm_op_a6w6.py
```diff
@@ -0,0 +1,735 @@
...
+            packed,
+            packed_scale,
+            rows,
+            NB,
+            NK_PAD,
+            x.stride(0),
+            _had32_t(x.device),
+            BLOCK_M=BM,
+        )
+        return packed, packed_scale
```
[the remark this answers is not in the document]
ksikiric: _PADK tiles are prefetch guards, not logical data, and are never accumulated. All five ASM variants produced bit-identical output with zeroed versus randomized guards, including padded shapes. Initializing them would add hot-path work, so I clarified the contract instead.

csrc/py_itfs_cu/asm_gemm_a6w6.cu
```diff
@@ -0,0 +1,231 @@
...
+    args.beta           = 0.0f;
+    args.stride_D0      = out->stride(0);
+    args.M              = Mdim;
+    args.N              = Ndim;
+    args.K              = Kdim;
+    args.ptr_ScaleA     = A_scale->ptr;
+    args.ptr_ScaleB     = B_scale->ptr;
+    args.stride_ScaleA0 = Kdim / kScaleGroupSize;
+    args.stride_ScaleB0 = Kdim / kScaleGroupSize;
+    args.log2_k_split   = 0;
```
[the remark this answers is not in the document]
ksikiric: A6W6 loads only stride_D0 from kernarg offset 0x60; stride_D1, stride_C0, and stride_C1 are never loaded by any registered variant. stride_D0 is set correctly, and contiguous output makes the column stride implicitly one. A4W4 uses a legacy field mapping that A6W6 does not share beyond the struct layout.

aiter/ops/gemm_op_a6w6.py
```diff
@@ -0,0 +1,882 @@
...
+_SCALE_TILE_BYTES = 1024
+_PACK_LAYOUT = "mxfp6_c0c1_256_padk2"
+_SHORT_K_SWIZZLE_LIMIT = 48 * _K_TILE
+_GROUPED_SWIZZLE_MAX_M = 16 * 32 * _TILE
+_GROUPED_SWIZZLE_MAX_N = 64 * _TILE
+_BATCHED_PACK_MIN_ELEMENTS = 32 * 1024 * 1024
+_BATCHED_PACK_BLOCK_M = 64
+_BATCHED_PACK_K_BLOCKS = _K_TILE // _SCALE_GROUP_SIZE
+_QUANT_BACKEND = os.environ.get("AITER_MXFP6_QUANT_BACKEND", "auto").lower()
+if _QUANT_BACKEND not in {"auto", "hip", "triton"}:
```
zufayu, line 40: This raise runs at import scope, and `__init__.py:100` imports the module unconditionally, so it takes the whole package down rather than just the mxfp6 path:

```
$ AITER_MXFP6_QUANT_BACKEND= python -c "import aiter"
  File "aiter/__init__.py", line 100, in <module>
    from .ops.gemm_op_a6w6 import *
  File "aiter/ops/gemm_op_a6w6.py", line 39, in <module>
ValueError: AITER_MXFP6_QUANT_BACKEND must be one of auto, hip, or triton, got ''
```

Not just a typo case — `os.environ.get(name, "auto")` returns `""` rather than the default, so the ordinary ways of clearing a variable (`docker run -e VAR=`, an empty value in compose/k8s, `export VAR=`) all kill the import. `"hip "` fails the same way; there is `.lower()` but no `.strip()`.

The comment at `__init__.py:80-86` argues this better than I can: that import list has no try/except by design because a failure there must surface loudly, citing the vLLM-ROCm `rmsnorm2d_fwd_with_add` regression. Worth `.strip()`, treating empty as unset, and warning back to `auto` on an unknown value — or validating at first use.

csrc/py_itfs_cu/asm_gemm_a6w6.cu
```diff
@@ -0,0 +1,231 @@
...
+                    A_scale->dtype() == AITER_DTYPE_u8 && B_scale->dtype() == AITER_DTYPE_u8,
+                __func__,
+                " packed operands and scales must use uint8");
+
+    const size_t nk_pad     = static_cast<size_t>(Kdim / kKTileSize + kKPaddingTiles);
+    const size_t expected_A = static_cast<size_t>(Mdim / kTileSize) * nk_pad * kPackedTileBytes;
+    const size_t expected_B = static_cast<size_t>(Ndim / kTileSize) * nk_pad * kPackedTileBytes;
+    const size_t expected_scaleA = static_cast<size_t>(Mdim / kTileSize) * nk_pad * kScaleTileBytes;
+    const size_t expected_scaleB = static_cast<size_t>(Ndim / kTileSize) * nk_pad * kScaleTileBytes;
+    AITER_CHECK(A->numel() >= expected_A && B->numel() >= expected_B,
```
zufayu: `>=` is the only bound on the launch here. The row-tile stride is derived from the kernarg `K`:

```
s_lshr_b32 s36, s26, 7      ; K/128        -> kloop trip count
s_add_u32  s33, s36, 2      ; K/128 + 2
s_mul_i32  s47, s33, 0x6000 ; * 24576      -> row-tile stride
```

and all four buffer descriptors carry `num_records = 0x80000000`, so there is no hardware range clamp behind it either. A blob packed at one K and launched at a smaller one passes this check and reads at the wrong stride, silently:

| | nk_pad | expected_A (M=512) | tile stride |
|---|---|---|---|
| packed at K=5120 | 42 | 2,064,384 | 1,032,192 |
| launched at K=2560 | 22 | 1,081,344 (passes) | 540,672 (wrong) |

No in-tree caller does this — `quant_mxfp6_gemm(x)` and `gemm_a6w6(..., K=x.shape[1])` share a K, and the tuner is consistent. But the blob carries no shape and M/N/K arrive as bare ints, so `==` would close it for free. Is `>=` intentional for an over-allocating caller?

zufayu (commented):
Read the diff and disassembled the committed `gemm_mxfp6_dmabig.co` to check the packed layout against the binary. Two notes inline:

- `gemm_op_a6w6.py:38` — an empty or space-padded `AITER_MXFP6_QUANT_BACKEND` makes `import aiter` fail outright, not just the mxfp6 path.
- `asm_gemm_a6w6.cu:146` — `>=` lets a blob packed at a different K through, and there is no hardware bound behind it.

Everything else checked out: the guard tiles are genuinely stride-only, the kernarg struct matches the `.co`, and the 57 tuned rows are clean.

aiter/ops/gemm_op_a6w6.py
ksikiric: Good catch. I normalized whitespace and treat an empty value as unset/auto, while retaining fail-fast validation for non-empty unknown values.

csrc/py_itfs_cu/asm_gemm_a6w6.cu
ksikiric: Agreed. Over-allocation is not intentional; all supported packers produce exact-size blobs. I changed validation to require exact operand and scale sizes, which also rejects mismatched-K packed buffers.

aiter/ops/gemm_op_a6w6.py
```diff
@@ -0,0 +1,886 @@
...
+    multiple of 128 so any GEMM shape maps onto the kernel's 256x256 / 128-K
+    tiling. Row-padding slots and the two trailing ABI K-guard tiles are not
+    consumed by the kernel and their contents are intentionally unspecified.
+
+    Returns (packed uint8, packed_scale uint8) torch tensors on w.device.
+    Works identically for both A and B operands. Runs entirely on the GPU.
+    """
+    rows, K = w.shape
+    padR, padK = _ceil(rows, _TILE), _ceil(K, _K_TILE)
+    w = w.detach()
```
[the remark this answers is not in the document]
ksikiric: fixed in [fb48eee](https://github.com/ROCm/aiter/pull/4859/commits/fb48eee5a03c0dfee4f154edd5d6db4e7e4b203d)

op_tests/test_gemm_a6w6.py
```diff
@@ -0,0 +1,153 @@
+# SPDX-License-Identifier: MIT
```
yifehuan, line 1: Thanks a lot The overall code structure looks clean and well organized, and I appreciate how complete the implementation is, including the operator, tuning support, configuration, and tests. The UT coverage also looks quite comprehensive, but there are still a few areas that do not appear to be exercised by CI:

 1. test_quant_mxfp6_gemm.py is included in the CI shard, but the standard runner executes it directly with python3. Since the file only contains pytest-style tests and has no test entry point, it exits successfully without actually running those test cases.
  2. test_a6w6_tuning.py has not been added to tuning-tests.yaml, so those tests are not executed by the tuning workflow.
  3. test_gemm_a6w6.py supports --all-kernels, but CI does not enable it. The default cases currently cover only dmabig and allk, while swz0, grp16, and grp64 still rely mainly on manual testing.
  4. The current end-to-end tests cover padding on M, but N and K remain aligned. It would be good to add at least one case where M, N, and K all require padding.

  Could you please wire these tests into CI so that all ASM kernel variants and padding paths are covered automatically?

ksikiric: Thanks. I pushed a fix now, all tests should now be in the CI

yifehuan (approved):
LGTM, nice code structure

PR #4862, opened 2026-08-19
[HIP] [FlyDSL] add mla for gfx1250

xiangM99:
## Motivation
- Add FlyDSL MLA decode kernels for gfx1250 with FP8 Q/KV inputs.
- Support page-size-1 and page-size-64 

## support 
**Valid head × q_seq combinations (persistent PS1):**
| H \\ q_seq | 1 | 2 | 3 | 4  |
|---|:---:|:---:|:---:|:---:|
| **16** | yes | yes | yes | yes |
| **32** | yes | no | no | no |
| **64** | yes | no | no | no |
| **128** | yes | no | no | no |

**FlyDSL PS64** (`flydsl_mla_pagesize64_fp8_fp8`) is a **separate** stage-1 implementation,
**not** wired into `mla_decode_fwd`. It is **hard-coded to 128 Q heads** and test usage
**q_seq = 1** only (`test_mla_flydsl.py -p 64`).

Do **not** assume PS64 supports the same head/q grid as PS1.

## benchmark

**Purpose:** Same logical KV; **persistent PS1** vs **ASM PS64**, both via
`mla_decode_fwd`. End-to-end **total µs** (stage-1 + reduce).

**Page-size-1 (FlyDSL persistent)**

| nhead | q_seq_len |
|---:|---|
| 16 | 1, 2, 4 (kernel also supports 3 via `-q 3`) |
| 32, 64, 128 | **1 only** |

**Page-size-64 (ASM)** — same `-n`/`-q`; auto splits (`--split-kv 0` default).

| Flag | Default |
|---|---|
| `-n` nhead | **16** (optional **32 64 128**) |
| `-q` | per nhead (see table) |
| `-b` | **1,2,4,8,16** |
| `-c` | **2048,4096,8192** |
| `--varlen` | off ; on = ragged lengths, same total_kv ≈ batch×ctx |
| `--scales` | **unit** (1/1) or **poc** (0.75/1.20, matches pagesize64 UT) |
| `--split-kv` | **0** (ASM auto) |

```bash
  python3 op_tests/test_mla_flydsl.py --mode ps1
  python3 op_tests/test_mla_flydsl.py --mode ps1 -b 1 -c 1 63 64 65 2048 4096 8192
  python3 op_tests/test_mla_flydsl.py --mode ps64 --split-kv 0 1 2
  python3 op_tests/test_mla_flydsl.py --mode ps1-vs-asm --num-heads 32 --q-seq-len 1 --varlen
```
### Summary (45 shapes)
PS1 latency stays roughly **15–37 µs** across the sweep; ASM PS64 spans a wider range
(**~21–74 µs**), with large batch + long ragged KV (e.g. batch 8, ctx 8192) hurting ASM
more than the persistent planner path.
| `q_seq_len` | Cases | Mean speedup | Min | Max |
|---:|---:|---:|---:|---:|
| 1 | 15 | **1.61×** | 1.18× | 2.46× |
| 2 | 15 | **1.96×** | 1.34× | 2.67× |
| 4 | 15 | **1.60×** | 1.10× | 1.95× |
| **All** | **45** | **1.73×** | **1.10×** | **2.67×** |

## Submission Checklist


op_tests/test_mla_pagesize1_persistent_flydsl.py
```diff
@@ -0,0 +1,318 @@
+# SPDX-License-Identifier: MIT
```
coderfeli: clean up ut using test skill in the repo. And no need so many standalone file, which should only have page size difference.

op_tests/test_mla_flydsl.py
```diff
@@ -0,0 +1,683 @@
+# SPDX-License-Identifier: MIT
```
coderfeli, line 1: use original mla test script?

op_tests/flydsl_tests/bench_flydsl_mla_persistent.py
```diff
@@ -0,0 +1,725 @@
+# SPDX-License-Identifier: MIT
```
coderfeli: why need another benchmark?

coderfeli:
Clean UT. Others part LGTM

op_tests/flydsl_tests/bench_flydsl_mla_persistent.py
xiangM99: Integrate benchmarks into test script; remove old script

op_tests/test_mla_flydsl.py
xiangM99: I chose a separate test file because the benchmarks compare `flydsl persistent` with `ASM` and require different KV cache shapes. Adding this to the original PS64 or persistent tests would have required extra cache and metadata logic, so I kept them isolated in a dedicated file.

coderfeli (approved):

PR #4833, opened 2026-08-18
[Triton/Gluon] Remove legacy MOE code

lburzawa:
## Motivation

Remove legacy MOE code.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4866, opened 2026-08-19
[Triton/Gluon] Move gluon gemm_a8w8 kernel into _gluon_kernels/gfx950

vgokhale:
Move gemm_a8w8 and blockscale for gfx950 from the legacy gluon/ to the new _gluon_kernels/gfx950 path.


op_tests/triton_tests/utils/gluon_paths.py
Boss2002n, line 1: can we just move the callers to the wrapper and if anyone is using it - they should change to using the wrapper - this feels like maintenance burden for future to me that we have to add support for yet another legacy path

aiter/ops/triton/_triton_kernels/moe/moe_op_mxfp4.py
Boss2002n, line 1: File is getting removed in https://github.com/ROCm/aiter/pull/4833

aiter/ops/triton/_triton_kernels/moe/moe_op_mxfp4_silu_fused.py
Boss2002n, line 1: file is getting removed in https://github.com/ROCm/aiter/pull/4833

aiter/ops/triton/gemm/basic/gemm_a8w8.py
```diff
@@ -11,10 +11,34 @@
...
+        if backend == "gluon":
+            config, _ = get_gemm_config("GEMM-A8W8", M, N, K, backend="gluon")
+        else:
+            config, _ = _get_config(M, N, K)
+
+    if y is None and (config.get("NUM_KSPLIT", 1) == 1 or not skip_reduce):
+        y = torch.empty((M, N), dtype=dtype, device=x.device)
+
+    return y, config
+
```
Boss2002n: we are adding too many config resolvers imo, can we avoid this and instead resolve it near the call site using based on arch and backend?

aiter/ops/triton/gemm/basic/gemm_a8w8.py
```diff
@@ -11,10 +11,34 @@
...
+from aiter.ops.triton.utils.gemm_config_utils import get_gemm_config
 from aiter.ops.triton.utils.logger import AiterTritonLogger
+from aiter.ops.triton.utils.types import (
+    get_scaled_dot_format_string,
+    torch_to_triton_dtype,
+)
 
 _LOGGER = AiterTritonLogger()
 
+_GLUON_SUPPORTED_ARCHS = ("gfx950",)
```
Boss2002n, line 25: In aiter/ops/triton/utils/_triton/arch_info.py
is_gluon_avail()
returns the same values - lets call that function instead of having to maintain 2 separate lists

aiter/ops/triton/moe/moe_op_mxfp4.py
Boss2002n, line 1: file getting removed in https://github.com/ROCm/aiter/pull/4833

aiter/ops/triton/moe/moe_op_mxfp4_silu_fused.py
Boss2002n, line 1: file getting removed in https://github.com/ROCm/aiter/pull/4833

Boss2002n (commented):
yet to finish reviewing ... (WIP)

aiter/ops/triton/gemm/basic/gemm_a8w8.py
vgokhale: You mean don't have a common function and just resolve it at the 2 places it is used in this file?

aiter/ops/triton/gemm/basic/gemm_a8w8.py
vgokhale: That function seems wrong. It is enforcing that every kernel that passes on either gfx950 or gfx1250 must pass the check on both. IOW, we would need to always add gluon kernels to both archs to be able to use it. The check will currently pass on 1250 as well but this kernel is just for 950 because gluon is specific to a generation.

op_tests/triton_tests/utils/gluon_paths.py
vgokhale: > they should change to using the wrapper

No we cannot do this in a prod library. We messed up here so we have to bear the maintenance burden. Otherwise this single PR will cause every framework / customer using AITER and this kernel to start throwing errors when they bump the AITER version.

aiter/ops/triton/gemm/basic/gemm_a8w8.py
rahulbatra85: Agree with Vinayak. We need a check per op because for some ops you may only have gluon kernels for some subset of archs. Some refactoring can be done to have common functions, but you will still need a list like _GLUON_SUPPORTED_ARCHS = ("gfx950",) or similar per op. This list should come from the kernel developer. Then a common function for each op can do the check based on this list.

aiter/ops/triton/__init__.py
```diff
@@ -71,6 +71,7 @@
     "gemm_a8w8_blockscale": "gemm.basic.gemm_a8w8_blockscale",
     "gemm_a8w8_per_token_scale": "gemm.basic.gemm_a8w8_per_token_scale",
     "gemm_a8w8": "gemm.basic.gemm_a8w8",
+    "gluon.gemm_a8w8": "gemm.basic.gemm_a8w8",
```
rahulbatra85, line 89: Why are you adding gluon in the name here? The caller of the op can pass that as a backend arg

vgokhale: Hmm can you elaborate? This is trying to keep compatibilty across the various current users that are importing this as 

`import aiter.ops.gluon.gemm_a8w8`

If I just remove it, those imports will start failing when the users bump their AITER commit. We don't want that as it is not good user experience.

rahulbatra85: That's kind of my point. We don't want caller of the ops to be doing imports like this. The op wrapper has backend parameter. If user wants to explicitly use some backend, then it should be passed in as backend parameter. Otherwise, by default(ie transparent to the caller/user) the best performing backend for an op on architecture should be used. 

Also, I agree usually we don't want the users to make changes like this on their end, but since these gluon kernels shouldn't have been in separate folder to begin with, we are just letting a bad practice carry on. We should move the users away from these imports. It will have to be done at some point on their end. Sooner the better. If you want to do it gently, then keep an import like this, but redirect it to a function that prints a deprecation warning msg and tell them to use a backend parameter if they want to use specific backend

By the way, sometimes APIs change or are deprecated. An example is python print function. It used to be statement, but then was changed to function to add more functionality.

op_tests/triton_tests/utils/gluon_paths.py
rahulbatra85: @vgokhale I think we should issue a deprecation warning for now and move users/callers away from this. Then eventually get rid of this import workaround like in a month or 2.

aiter/ops/triton/__init__.py
vgokhale: Yeah I can go an change these from vllm / sglang and ATOM if they are direct imports. But  Idon't want them breaking now. And I don't want to gate this review by the review velocity of those projects.

aiter/ops/triton/gemm/basic/gemm_a8w8.py
Boss2002n: Yes

op_tests/triton_tests/utils/gluon_paths.py
vgokhale: Done.

aiter/ops/triton/gemm/basic/gemm_a8w8.py
vgokhale: Done

aiter/ops/triton/__init__.py
vgokhale: Added a warning

Boss2002n (dismissed):

azaidy (dismissed):
LGTM!

azaidy (dismissed):
LGTM!

rahulbatra85 (approved):

PR #4385, opened 2026-07-25
[Bugfix][Triton] Avoid RDNA4 unified attention LDS overflow

hogeheer499-commits:
## Summary

Prevent the Triton 3D unified-attention kernel from selecting a two-stage
configuration whose staged K/V tiles already consume the full 64 KiB LDS
budget on `gfx1200` and `gfx1201`.

## Motivation

With a 256-wide BF16 KV head and `TILE_SIZE=64`, Triton 3.6 reports:

```text
OutOfResources: shared memory, Required: 65792, Hardware limit: 65536
```

The K/V tiles alone require 64 KiB at two stages, leaving no room for Triton's
256-byte scratch allocation. This causes a hard startup failure in the
downstream vLLM report.

The Qwen3.6-27B model from that report uses a 256-wide full-attention head:
https://huggingface.co/Qwen/Qwen3.6-27B-FP8/blob/main/config.json

## Changes

- Estimate the two-stage K/V tile footprint after the final tile size is known.
- Use one stage on `gfx1200` and `gfx1201` only when that footprint reaches the
  64 KiB LDS limit.
- Preserve two stages for smaller heads or tiles, FP8 KV caches, and `gfx1151`.
- Add focused selector regression coverage for both affected architectures and
  the configurations that must remain unchanged.

## Testing

- `7 passed`:
  `pytest op_tests/triton_tests/attention/test_unified_attention.py -k test_select_3d_config_respects_rdna_lds_limit -q`
- Triton 3.6 compile targeting `gfx1200` and `gfx1201`:
  - before: two stages, 65,792 bytes, compile rejected
  - after: one stage, 32,768 bytes
- Real `gfx1151` execution:
  - its existing two-stage 65,536-byte path remains selected and passes the
    numerical comparison
  - forcing the new `gfx1201` selector path uses one stage and also passes the
    numerical comparison
- BF16 query with FP8 KV remains at two stages and compiles at 33,024 bytes.
- Black check, import-order check, Python byte compilation, and
  `git diff --check` pass.

I do not have native `gfx1200`/`gfx1201` hardware, so the affected targets were
compiler-validated with Triton's architecture override. Native RDNA4 CI or
maintainer validation would still be valuable.

No new dependencies or breaking changes.

Fixes #4329.

PR #4868, opened 2026-08-19
[Triton/Gluon] [Bugfix] Guard RDNA unified attention against LDS overflow

### LINKED ISSUE #4329: [Issue]: `select_3d_config` picks a 3D unified-attention config that exceeds LDS on non-gfx1250 Wave32 (Required 65792 > 65536)
### Summary

`select_3d_config` in `aiter/ops/triton/attention/unified_attention.py` chooses `TILE_SIZE` and the pipeline depth for `kernel_unified_attention_3d` with **no check that the resulting LDS / shared-memory request fits the target GPU**.
On non-`gfx1250` Wave32 archs (the `IS_DEVICE_ARCH_GFX12 = DEVICE_ARCH in ("gfx1250",)` else-branch), once the Triton pipeliner double-buffers the K/V tiles the kernel requests more than the 64 KB (65536 B) limit, and Triton aborts at its pre-launch shared-memory check:

```text
    triton.runtime.errors.OutOfResources: out of resource: shared memory
    Required: 65792, Hardware limit: 65536
```

For downstream users this is a hard **startup** failure, see vllm-project/vllm#48723 (gfx1201 / Radeon AI PRO R9700, aborts at `vllm serve` init).

### Regression

The shipped-config LDS footprint **doubled** between the last-working and first-broken aiter:

| aiter | LDS @ 3D config (TILE_SIZE=64, head=128, bf16 KV) |
|---|---|
| v0.1.13.post1 (worked) | 16384 B |
| v0.1.16.post2 (broke)  | 33024 B |

v0.1.13 reused a single `[TILE_SIZE × head]` bf16 tile in LDS; v0.1.16 stages **separate K and V buffers**. `2 × 16384 + 256 B scratch = 33024`. On gfx1201's Triton stack the deeper double-buffering doubles again to `4 × 16384 + 256 = 65792`, 256 B over the limit. (Measured via `triton.compile(...).metadata.shared`.)

### Where

`select_3d_config` (around `unified_attention.py:56`). The non-`gfx1250` branch sets `TILE_SIZE = min(64, next_pow2(block))` and a pipeline depth with no LDS comparison, so the double-buffered footprint `≈ num_stages × TILE_SIZE × head_size × 2 (+ scratch)` can exceed 65536 whenever `num_stages × head_size ≳ 512`.

### Reproduced

Reproduced the byte-exact `65792 > 65536` on **gfx1151** (RDNA3.5, Wave32 — same non-gfx1250 branch) by forcing the K/V staging depth the gfx1201 stack reaches at stock. gfx1151 fits at the shipped config today (33024 B), but the v0.1.13→v0.1.16 doubling halved its headroom, so the whole non-gfx1250 Wave32 family is on the same cliff.

### Ask

Add an LDS-budget check to the non-`gfx1250` path in `select_3d_config`: when the double-buffered footprint would exceed the device LDS limit, clamp the pipeline depth (e.g. to 1) or `TILE_SIZE` (e.g. to 32) so the 3D kernel fits, rather than emitting a config Triton rejects at launch.

### Related

- ROCm/aiter#3294 : broader "first-class gfx1201 support" tracker; this is a specific, actionable instance of the launch failures noted there.
- vllm-project/vllm#48723 : downstream bug report (gfx1201).
- vllm-project/vllm#49264 : vLLM-side stopgap that catches the overflow and falls back to the in-tree Triton unified-attention kernel so serving continues while this is fixed here.

### Operating System

Ubuntu 26.04 LTS (Resolute Raccoon)

### CPU

AMD Ryzen AI MAX+ 395 w/ Radeon 8060S

### GPU

gfx1151 AMD Radeon 8060S (integrated GPU of the Ryzen AI MAX+ 395, "Strix Halo", RDNA3.5, Wave32)

### ROCm Version

7.2 · torch: 2.11.0+rocm7.2 · triton: 3.6.0

### ROCm Component

_No response_

### Steps to Reproduce

**Natural failure (as reported downstream on gfx1201, vllm-project/vllm#48723):**

1. Install the v0.25.0 ROCm stack: aiter v0.1.16.post2, ROCm 7.2, torch 2.11.0, triton 3.6.0.
2. Serve a head_size=128 model with the AITER unified-attention backend, e.g.:
`vllm serve Qwen/Qwen3.6-27B-FP8 --tensor-parallel-size 4 --attention-backend ROCM_AITER_UNIFIED_ATTN --max-model-len 131073 --max-num-seqs 8`
3. During CUDA-graph capture the backend selects `kernel_unified_attention_3d` (`max_seqlen_k > 512`); its first launch hits Triton's pre-launch LDS check → `OutOfResources: Required 65792 > 65536` and startup aborts. The identical config starts on v0.24.0 (aiter v0.1.13.post1).

**Direct measurement of the config's LDS (how I confirmed the doubling on gfx1151, any non-gfx1250 Wave32 arch):**

Compile `kernel_unified_attention_3d` for the shipped 3D config and read `triton.compile(...).metadata.shared`. It reports **16384 B** on v0.1.13.post1 vs **33024 B** on v0.1.16.post2 (TILE_SIZE=64, head=128, bf16 KV), and reaches **65792 B** once the deeper K/V double-buffering applies (`num_stages × head_size ≳ 512`) — 256 B over the 65536 limit.

_No response_

### Additional Information

_No response_

amd-xavierwang:
## Motivation

select_3d_config's 3D path sized the KV pipeline for the CDNA LDS budget. On RDNA the per-workgroup budget is 64 KiB, and a large head_size (e.g. Gemma3 head_size=256, Gemma4 head_size=512) overflows it, so the kernel fails to compile with an out-of-resources error.

This PR supersedes #4385 by: 
1.  guarding all RDNA GPUs instead of gfx12 only. 
2. #4385's solution only works for Gemma3's head_size but not Gemma4.
3. This PR also conservatively rejects edge case when estimate LDS size strictly **EQUALS** `RDNA_LDS_LIMIT`: depends on Triton compiler version and target, sometimes an additional 256B is required which would still cause the size overflow.
4. **[refactor]: add arch alias for sub RDNAs: `gfx1101`, `gfx1102`.. Otherwise they hit import error after the refactoring.

** _The current algorithm prioritizes reducing num_stages, then per loop KV TILE_SIZE. This ordering is a
robust default that stays competitive across the shapes it guards, but It is not claimed to be globally optimal: on some RDNA parts, shrinking TILE_SIZE first while keeping num_stages=2 (trading tile width for occupancy) is faster for certain decode shapes. The true optimum is arch- and dtype-dependent and belongs in a dedicated tuning pass._
_Drawing the line here is also reasonable because this path is a fallback: for head_size 256/512, vLLM's upstream TRITON_ATTN backend is globally faster for such aiter 3D kernel configs, so aiter's role for these shapes is enablement and bugfix — resolving the compile-time error._

Fixes: #4329

## Test Plan

**Cross-compilation, all RDNA targets, Triton 3.8.0.** Every arch in aiter's
`RDNA_ARCHS`  compiled at each head_size/dtype by replaying a
captured 3D launch under a `GPUTarget` override:

**attn_3d LDS (bytes)** hs 188->356, hs 344->512

| arch    | hs188 bf16 | hs344 bf16 | hs188 fp8 | hs344 fp8 |
|---------|-----------:|-----------:|----------:|----------:|
| gfx1100 |      32768 |      32768 |     33024 |     32768 |
| gfx1101 |      32768 |      32768 |     33024 |     32768 |
| gfx1102 |      32768 |      32768 |     33024 |     32768 |
| gfx1150 |      32768 |      32768 |     32768 |     32768 |
| gfx1151 |      32768 |      32768 |     32768 |     32768 |
| gfx1200 |      32768 |      32768 |     33024 |     32768 |
| gfx1201 |      32768 |      32768 |     33024 |     32768 |

28/28 compile.

**reinforces why the guard treats a footprint that reaches 64 KiB as overflowing.**
Cross-compiling the *unguarded* configs, six of the seven targets request 66048 B
for a nominal-65536 B config — 512 B of pipeline scratch on top. gfx1151 alone
requests exactly 65536 and fits, which is why the boundary appears safe if only
gfx1151 is tested. The scratch is also compiler-dependent: #4329 measured 256 B
on Triton 3.6, this run measures 512 B on 3.8. Treating `>=` the limit as
overflow is what keeps the fix correct across parts and toolchains.


aiter/ops/triton/attention/unified_attention.py
```diff
@@ -45,6 +45,9 @@
 WARP_SIZE = 32 if IS_DEVICE_ARCH_GFX12 else 64
 WAPR_SIZE_LOG2 = int(math.log2(WARP_SIZE))
 
+# Per-workgroup LDS budget shared by all RDNA GPUs.
+RDNA_LDS_LIMIT = 64 * 1024
```
cagrikymk: I think we can extract the LDS limit from torch or triton, I think there is a public API somewhere to extract LDS size without hardcoding it. Can you look into it? If not, we can add a helper to aiter/triton utils so the LDS limits are centralized there.

amd-xavierwang: done. I kept it at module scope because we don't need this anymore in the future when arch-specific tuning has been added.

Ragua1:
Two shapes outside the matrix of the new test also fail on `main` and are fixed here: `head_size=192` and `head_size=320`, alongside the 256 and 512 that `test_triton_unified_attn_3d_rdna_lds` covers. And `head_size=128`, which fits on both revisions, gives identical figures before and after — the selector change does not perturb configurations that already worked.

That is a runtime check on **gfx1150 hardware, Windows with native ROCm, Triton 3.7.1** — a combination none of the runs in the description covers, since the gfx1150 row in the cross-compile table is compile-only. Correctness is against an independent fp32 reference rather than the new test, which cannot be imported on this platform (the `op_tests` chain needs a `rocminfo` executable that a runtime-only ROCm install does not ship, and triggers a JIT build at import). The reference is `softmax(q kᵀ · scale) v` over the KV cache gathered through the block table — exact rather than approximate for this path, because `max_seqlen_q == 1` removes masking. Inputs bf16, one sequence, `block_size=64`, `kv_len=8192`, 8 query heads over 1 KV head.

| `head_size` | `main` @ `ba59a37` | this PR @ `5f69ed7` | `max_abs` vs reference | `rel_l2` |
|---|---|---|---|---|
| 128 | runs | runs | 1.563e-04 | 2.012e-03 |
| 192 | `OutOfResources` | **runs** | 1.946e-04 | 2.122e-03 |
| 256 | `OutOfResources` | **runs** | 1.581e-04 | 2.062e-03 |
| 320 | `OutOfResources` | **runs** | 2.556e-04 | 2.098e-03 |
| 512 | `OutOfResources` | **runs** | 1.660e-04 | 2.101e-03 |

The `atol` for the bf16/bf16 case in the new test is `1.5e-2`, so the largest deviation is roughly sixty times smaller; outputs were also checked explicitly for all-zeros and non-finite values.

`main` reproduces #4329 unmodified — `OutOfResources … Required: 65792, Hardware limit: 65536` at `head_size=256`, using the standalone script from that issue. `65792` is `65536 + 256`, so Triton 3.7.1 sits with 3.6 rather than with the 512 B measured on 3.8; a third point on the same curve, and the nominal-65536 configuration still overflows, which is what `>=` admits. 3.7.1 is not an incidental version here: there is no `triton` wheel for win32 and `pytorch-triton-rocm` is Linux-only, so Windows runs the `triton-windows` fork, whose newest release is `3.7.1.post27`.

The `max_shared_mem` lookup added after review resolves on this platform: `from triton.compiler import max_shared_mem` imports and returns `65536` for the active device — equal to the `64 * 1024` fallback, so the `except` clause is not silently masking a failed import here.

Environment in full: Radeon 890M (RDNA 3.5, Wave32, 16 CU / 8 WGP, 65536 B LDS), torch `2.11.0+rocm7.15.0a20260728`, `AITER_TRITON_ONLY` (unconditional on Windows). Measured 2026-08-21 against `main` at `ba59a37` and this PR at `5f69ed7`, i.e. the current head with the runtime `max_shared_mem` lookup rather than the earlier hardcoded limit. The gfx1101 confirmation on #4329 was mine; that machine is not available at the moment, so nothing here speaks to gfx1101.

Ragua1:
`_ARCH_ALIAS` is applied to the `arch` parameter of `_load`, which is the caller-supplied override, and no production call site sets it. `get_unified_attention_config` documents that parameter as *"resolve another arch's table, for tooling and tests"*, and all seven call sites in `unified_attention.py` pass only `params` and `backend`. With `arch=None`, `_ARCH_ALIAS.get(None, None)` returns `None`, so `resolve_config_dir` falls through to `arch_info.get_arch()` and receives the unaliased architecture. The three aliased targets are exactly the architectures in the description's RDNA table that have no `unified_attention` table of their own, so for those the alias is the only mechanism that selects a table.

Config resolution on this revision, gfx1101, in one process:

| call | result |
|---|---|
| `_load("attn_3d", "triton", None)` — the production path | `AssertionError: Required config file doesn't exist: …/configs/gfx1101/triton/attention/unified_attention/DEFAULT.json` |
| `_load("attn_3d", "triton", "gfx1101")` — tooling and tests only | resolves `…/configs/gfx1100/…`, with the keys added here |

The fourth column below is this same revision with the alias reaching the resolved architecture, produced locally for the sole purpose of establishing whether the tables themselves are correct. Runtime on the 3D decode path, bf16, `max_abs` against the independent fp32 reference described in my earlier comment:

| `head_size` | `7bc24377` (pre-refactor) | `main` @ `12620102` | this PR @ `4f00b66` | alias reaching the resolved arch |
|---|---|---|---|---|
| 128 | runs, 2.621e-04 | `AssertionError` | `AssertionError` | runs, **2.621e-04** |
| 192 | `OutOfResources` | `AssertionError` | `AssertionError` | runs, 1.943e-04 |
| 256 | `OutOfResources` | `AssertionError` | `AssertionError` | runs, 2.022e-04 |
| 320 | `OutOfResources` | `AssertionError` | `AssertionError` | runs, 1.725e-04 |
| 512 | `OutOfResources` | `AssertionError` | `AssertionError` | runs, 2.682e-04 |

The tables added here are therefore correct for gfx1101: every shape runs, the largest deviation is roughly fifty times below the `1.5e-2` bf16 `atol`, and `head_size=128` returns the same value as the pre-refactor tree, so the aliased table does not perturb the configuration that already worked.

`head_size=128` ran on the pre-refactor tree and does not run on `main` now. `configs/CLAUDE.md` §6 names a missing required default as the intended failure, and its seeding section describes giving a new architecture its own directory as a byte-identical copy from the closest measured arch. The choice between that route and a code-level alias is a policy question, which is why it is not made here.

Measured 2026-09-10 on gfx1101, Windows with native ROCm, torch `2.15.0a0+rocm10.1.0a20260909`, Triton `3.8.0`, `AITER_TRITON_ONLY`, one sequence, `block_size=64`, `kv_len=8192`, 8 query heads over 1 KV head, n=1 per cell. gfx1102 and gfx1150 go through the same code path and were not measured, as no hardware for either is available to me. No performance claim is made.

cagrikymk:
@amd-xavierwang Are the aliased archs close enough? I am not that familar with rdna series.

amd-xavierwang:
> `_ARCH_ALIAS` is applied to the `arch` parameter of `_load`, which is the caller-supplied override, and no production call site sets it. `get_unified_attention_config` documents that parameter as _"resolve another arch's table, for tooling and tests"_, and all seven call sites in `unified_attention.py` pass only `params` and `backend`. With `arch=None`, `_ARCH_ALIAS.get(None, None)` returns `None`, so `resolve_config_dir` falls through to `arch_info.get_arch()` and receives the unaliased architecture. The three aliased targets are exactly the architectures in the description's RDNA table that have no `unified_attention` table of their own, so for those the alias is the only mechanism that selects a table.
> 
> Config resolution on this revision, gfx1101, in one process:
> 
> call	result
> `_load("attn_3d", "triton", None)` — the production path	`AssertionError: Required config file doesn't exist: …/configs/gfx1101/triton/attention/unified_attention/DEFAULT.json`
> `_load("attn_3d", "triton", "gfx1101")` — tooling and tests only	resolves `…/configs/gfx1100/…`, with the keys added here
> The fourth column below is this same revision with the alias reaching the resolved architecture, produced locally for the sole purpose of establishing whether the tables themselves are correct. Runtime on the 3D decode path, bf16, `max_abs` against the independent fp32 reference described in my earlier comment:
> 
> `head_size`	`7bc24377` (pre-refactor)	`main` @ `12620102`	this PR @ `4f00b66`	alias reaching the resolved arch
> 128	runs, 2.621e-04	`AssertionError`	`AssertionError`	runs, **2.621e-04**
> 192	`OutOfResources`	`AssertionError`	`AssertionError`	runs, 1.943e-04
> 256	`OutOfResources`	`AssertionError`	`AssertionError`	runs, 2.022e-04
> 320	`OutOfResources`	`AssertionError`	`AssertionError`	runs, 1.725e-04
> 512	`OutOfResources`	`AssertionError`	`AssertionError`	runs, 2.682e-04
> The tables added here are therefore correct for gfx1101: every shape runs, the largest deviation is roughly fifty times below the `1.5e-2` bf16 `atol`, and `head_size=128` returns the same value as the pre-refactor tree, so the aliased table does not perturb the configuration that already worked.
> 
> `head_size=128` ran on the pre-refactor tree and does not run on `main` now. `configs/CLAUDE.md` §6 names a missing required default as the intended failure, and its seeding section describes giving a new architecture its own directory as a byte-identical copy from the closest measured arch. The choice between that route and a code-level alias is a policy question, which is why it is not made here.
> 
> Measured 2026-09-10 on gfx1101, Windows with native ROCm, torch `2.15.0a0+rocm10.1.0a20260909`, Triton `3.8.0`, `AITER_TRITON_ONLY`, one sequence, `block_size=64`, `kv_len=8192`, 8 query heads over 1 KV head, n=1 per cell. gfx1102 and gfx1150 go through the same code path and were not measured, as no hardware for either is available to me. No performance claim is made.

Thank you so much for catching the bug! I will issue the fix soon. Please let me know if you find other problems. :)

amd-xavierwang:
> @amd-xavierwang Are the aliased archs close enough? I am not that familar with rdna series.

Yes, they should be close enough. For instance, gfx1150 (Strix Point), is same as gfx1151 (Strix Halo) but with low number of CUs and memory bandwidth

aiter/ops/triton/configs/gfx1201/triton/attention/unified_attention/DEFAULT.json
```diff
@@ -4,7 +4,12 @@
       "Q",
       "DT"
     ],
+    "attn_3d": [
+      "D",
+      "DT"
+    ],
```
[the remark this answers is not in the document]
amd-xavierwang: Thanks for the review. On the Critical LDS finding — it's a false positive on the default path. K and V reuse one LDS buffer, so the num_stages × 2(K+V) × TILE × head × bytes formula double-counts. Measured compiled shared for every non-shuffled shape stays ≤ 33 KB, well under the 64 KB budget, including the cases cited (e.g. shuffled blk128 bf16 head64 measures 16–32 KB, not 65536).

The shuffled bypass (TILE_SIZE = block_size) is real, and when block_size >=256, it becomes a problem: https://github.com/ROCm/aiter/blob/f361bd39ba4196ce29391c628d4ebf72e7929ab8/aiter/ops/triton/attention/unified_attention.py#L306. But it's opt-in — shuffled_kv_cache defaults to False and is caller-owned. This PR aims to offer immediate support to make unified attention work for gemma3/4 shapes under regular setup(shuffle to False), not aiming for detailed optimizations.

aiter/ops/triton/utils/unified_attention_utils.py
```diff
@@ -48,6 +49,12 @@
 _CONFIG_NAME = "UNIFIED-ATTENTION"
 _OPS = ("attn_2d", "attn_3d", "reduce", "kv_split")
 
+_ARCH_ALIAS = {
```
brunomazzottiamd, line 53: Can you please explain what's the reasoning behind `_ARCH_ALIAS` with a brief comment? It isn't obvious at first glance, particularly to engineers with only CDNA experience.

brunomazzottiamd (approved):

cagrikymk (approved):
LGTM!

PR #4265, opened 2026-07-16
[Triton] Enable conv2d Triton kernels for gfx1100 (RDNA3) and gfx1151 (RDNA3.5)

saeid-rostami:
## Motivation
 Extends the Triton conv2d operator to two more RDNA targets:

  - **gfx1100** —  RDNA3 
  - **gfx1151** —  RDNA3.5 

  Previously the conv2d path only recognized RDNA4 (gfx1200/gfx1201) on the RDNA
  side. This PR adds the two arches to the supported set and ships per-arch tuned
  config JSONs for all five conv2d kernels, so both GPUs get shape-appropriate
  tile configs instead of falling through / being rejected.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4869, opened 2026-08-20
[Triton/Gluon] [Conv2D] Optimize conv2d for RDNA - kernels, routing, and input packing

saeid-rostami:
## Motivation

Optimization of conv2d for RDNA architectures.
## Summary

This PR improves the Triton Conv2D path on AMD RDNA GPUs for FP16/BF16 and
NCHW/NHWC workloads. It adds a repack-free NCHW 3x3 kernel, replaces the
multi-operation NCHW-to-NCHWc conversion with a single Triton kernel, improves
shape/layout/dtype-aware configuration selection, and adds tuned configurations
for gfx1100, gfx1151, and gfx1201.

The public API is unchanged.

## What changed

- Add a direct NCHW 3x3 kernel that reads contiguous NCHW activations without
  materializing an NCHWc input buffer.
- Route direct NCHW only where it is profitable:
  - gfx1100 and gfx1151 use exact-shape route pins.
  - gfx1201 uses the configured spatial crossover.
  - Other shapes retain the cblocked fallback.
- Fuse NCHW padding and layout conversion into one Triton
  `_nchw_to_cblocked_kernel` launch, with per-architecture `CONV-PREPACK` JSON
  configurations.
- Avoid unnecessary hot-path work by allocating channels-last outputs directly,
  loading bias in its original dtype and promoting it inside the Triton
  epilogue, and skipping logging string construction when INFO logging is off.
- Strengthen weight-pack cache keys with device, shape, stride, dtype, tensor
  version, and block size so aliases and in-place updates cannot produce stale
  cache hits.
- Extend Conv2D config lookup with layout/dtype-specific exact-shape variants,
  followed by generic shape, `M_LEQ_*`, and `any` fallbacks.
- Keep production launches deterministic by removing runtime autotune decorators
  and candidate tables from the kernel modules. Production always consumes the
  frozen per-architecture JSON configurations.
- Add regression coverage for exact-route selection, fallback routing, complete
  shape keys, and config-variant precedence.

## Performance

Tested on ResNet-50, Stable Diffusion 3.5 Medium, and
FLUX.2 Klein 9B across both FP16 and BF16.

- Each model/layout/dtype combination ran in a fresh process with a fresh
  Triton disk cache; JIT compilation is excluded from the reported kernel time.
- The benchmark flushes L2 between timed measurements.
- NCHW optimized totals use `kernel+repack`, so the NCHWc input-pack cost is
  included. NHWC totals are kernel-only because that path has no separate input
  pack.

### gfx1201

| Layout | Main total time (ms) | PR total time (ms) | Time reduction | Main aggregate TFLOPS | PR aggregate TFLOPS | TFLOPS gain |
| NCHW | 77.45 | **64.18** | **17.1%** | 70.52 | **85.11** | **20.7%** |
| NHWC | 92.45 | **52.59** | **43.1%** | 59.08 | **103.86** | **75.8%** |

### gfx1151

| Layout | Main total time (ms) | PR total time (ms) | Time reduction | Main aggregate TFLOPS | PR aggregate TFLOPS | TFLOPS gain |
| NCHW | 294.989 | **269.364** | **8.69%** | 18.52 | **20.28** | **9.51%** |
| NHWC | 343.757 | **247.980** | **27.86%** | 15.89 | **22.03** | **38.62%** |

### gfx1100

| Layout | Main total time (ms) | PR total time (ms) | Time reduction | Main aggregate TFLOPS | PR aggregate TFLOPS | TFLOPS gain |
| NCHW | 83.66 | **79.06** | **5.5%** | 65.29 | **69.09** | **5.8%** |
| NHWC | 90.86 | **73.39** | **19.2%** | 60.12 | **74.43** | **23.8%** |

## Submission Checklist


Ragua1:
Measured on **gfx1150** (Radeon 890M, RDNA 3.5), Windows with native ROCm, torch `2.11.0+rocm7.15.0a20260728`, Triton `3.7.1`, on 2026-08-21 against `main` at `ba59a37` and this branch at `e363cf6`. I maintain a downstream consumer that routes diffusion VAE-decode convolutions through `aiter.ops.triton.conv`. My primary interest is **gfx1101** (RX 7800 XT), which I cannot measure at the moment; gfx1150 is the card I have to hand, and neither ships a conv table, so they are in the same position with respect to this PR.

**The performance direction of this PR is good on this part**, which is why I would like it to stay reachable. On the routes it keeps — `cblocked`, fp16, N=1, K=C, interleaved A-B-A-B in one session, the two runs per revision shown rather than averaged:

| shape | `main` | this PR | change |
|---|---|---|---|
| C=512, 128×128, 3×3 | 10.575 / 10.640 ms | 10.520 / 10.482 ms | −1 % |
| C=512, 256×256, 3×3 | 41.706 / 41.043 ms | 40.992 / 40.567 ms | −1.5 % |
| C=256, 512×512, 3×3 | 41.651 / 42.563 ms | 39.344 / 39.325 ms | **−6.6 %** |
| C=128, 1024×1024, 3×3 | 43.978 / 44.401 ms | 37.485 / 37.546 ms | **−15.0 %** |

The new packing kernel accounts for it: pack cost drops from 9.37 to 5.85 ms at 1024×1024, about 38 % cheaper than the host-side `permute().contiguous()` on `main`. Separately, `conv2d_1x1(layout="nhwc")` is 2.3× faster here (2.85 → 1.25 ms).

## The question

On an architecture that ships no tuned table, the two revisions behave differently:

```
main @ ba59a37,    no config, AITER_TRITON_CONV_AUTOTUNE=0  ->  FileNotFoundError
main @ ba59a37,    no config, AITER_TRITON_CONV_AUTOTUNE=1  ->  runs, rel_l2 7.67e-04
this PR @ e363cf6, no config, AITER_TRITON_CONV_AUTOTUNE=1  ->  FileNotFoundError
```

Two changes combine: `_triton_kernels/conv/helpers.py` is deleted along with the seven `AUTOTUNE_*_CONFIGS`, so the env-var bypass documented in `conv/README.md` no longer exists; and `CONV-PREPACK` is a newly required per-arch family on a path that previously read no JSON at all, so it is the first lookup to fail. The per-architecture requirement goes from 7 tables to 9, and every shipped table is arch-specific. The updated README states which architectures ship one — gfx1100, gfx1151, gfx1200 and gfx1201 — and that is four of the eight in `RDNA_ARCHS`; the other four are gfx1030, gfx1101, gfx1102 and gfx1150, so conv2d becomes unreachable on RX 6800/6900, RX 7800 XT and 7700 XT, RX 7600 and Strix Point.

I am not claiming any of those is documented as supported; `op_tests/triton_tests/conv/_helpers.py` lists exactly the four that have tables, so the current state is self-consistent. The question is what the intended path is for the other four now that retuning moves out of the package, since the README paragraph describing how to run without a table is removed in this diff, and its replacement states that *"candidate search spaces and the tuning driver live outside the production package"*.

Two observations that may make this cheap to resolve. First, on this part the config *contents* matter much less than their presence: a 48-combination sweep over `BLOCK_M`/`BLOCK_N`/`BLOCK_K`/`GROUP_SIZE_M`/`num_warps` on the four shapes above — route asserted as `DIRECT_NCHW_3X3`, output checked against `torch.nn.functional.conv2d` — reaches parity with the `cblocked` route at best (0.93–1.16× over three repeats), and simply borrowing a shipped table gives much the same (`gfx1151` 0.84×, `gfx1201` 0.85×). Second, it would have to be a *curated* default rather than arbitrary values: in the same sweep `num_warps=2` with `BLOCK_N=128` produced 0.05×, twenty times slower than `cblocked`. A nearest-architecture fallback would therefore cost very little and remove a hard failure — and `has_conv_config()`, which this PR adds with `load_config_json(..., required=False)` and a docstring calling the table *optional*, is already the primitive for it. (The suppressed automated comment on `conv2d.py` proposes exactly that as a guard for `conv2d_nchw_3x3_direct()`; that is one entry point, whereas on an architecture with no table at all the first mandatory `get_conv_config` fails inside `load_config_json` before any `KeyError` path is reached.)

On gfx1101 the picture came out a little differently — there the table *contents* did move the number. Measured in early August 2026 on the pre-#4869 tree, not re-verified since, and to one significant figure: repeating the comparison in a later session shifted the ratio by about two points, so I would not quote a second digit. The shipped `gfx1100` table from #4265 was indistinguishable from an RDNA4 table on that card — a ratio of 1.00 against a 6–8 % spread within a single arm — so same-generation donorship was not the lever. Changing one value in the RDNA4 table — `BLOCK_K` 64→32 in `CONV-3X3-CBLOCKED` — was worth about 10 % there, while the same edit applied to the `gfx1100` table was worth nothing, so the useful value is donor-specific rather than portable. And a held-out neighbour table cost about 9 % against one tuned on the card. So a neighbour table is worth far more than a hard failure, and still leaves something on the table; that ordering is what matters here.

Is a curated nearest-architecture fallback something you would consider, or is the intent that an architecture without a table goes through the offline tuning path first? Either answer is useful — it tells downstream consumers on those four architectures what to plan for.

aiter/ops/triton/_triton_kernels/conv/conv_3x3.py
```diff
@@ -320,116 +335,132 @@ def _conv2d_3x3_cblocked_kernel(
...
+                w_tile = tl.load(
+                    w_ptrs, mask=kout_mask[:, None] & k_mask[None, :], other=0.0
+                )
+                acc = tl.dot(w_tile, x_tile, acc=acc)
+
+    if HAS_BIAS:
+        b = tl.load(BIAS + offs_n, mask=kout_mask, other=0.0).to(tl.float32)
+        acc += b[:, None]
+
+    if ACTIVATION == "relu":
```
brunomazzottiamd: **Suggestion:** I think you can `from ..activation import _apply_activation_from_str` and save some boilerplate in all your kernels.

This is a non-blocking comment, we can merge even if it isn't addressed.

aiter/ops/triton/conv/_utils.py
```diff
@@ -29,23 +30,24 @@ def _conv_dims(x, w_oihw, stride, padding, dilation):
 
 def _alloc_output(N, K_out, P, Q, x, layout):
     """Allocate the output tensor, channels_last for nhwc else contiguous."""
-    y = torch.empty((N, K_out, P, Q), device=x.device, dtype=x.dtype)
     if layout == "nhwc":
-        return y.to(memory_format=torch.channels_last)
-    return y
+        return torch.empty(
```
brunomazzottiamd: **Suggestion:** What about doing this to avoid argument repetition?

```python
return torch.empty(
    (N, K_out, P, Q),
    device=x.device,
    dtype=x.dtype,
    memory_format=torch.channels_last if layout == "nhwc" else torch.contiguous_format,
)
```

aiter/ops/triton/conv/_launch.py
```diff
@@ -37,8 +39,20 @@
...
+from aiter.ops.triton._triton_kernels.conv.nchw_to_cblocked import (
+    _nchw_to_cblocked_kernel,
+)
+from aiter.ops.triton.utils.conv_config_utils import (
+    format_prepack_shape_key,
+    format_shape_key,
+)
+
+
+def _dtype_variant(dtype):
```
brunomazzottiamd: **Sincere Question:** Why do you need distinct tuning config variants for `fp16` and `bf16` data types? As far as my knowledge goes, both types have the same peak TFLOPS (at least on CDNA) and have the same size (2 bytes). Theoretically speaking, at least on my mind, `fp16` and `bf16` should share the same configs.

aiter/ops/triton/conv/conv2d.py
```diff
@@ -102,12 +206,13 @@ def conv2d(
     if layout not in ("nchw", "nhwc"):
         raise ValueError(f"layout must be 'nchw' or 'nhwc', got '{layout}'")
 
-    _LOGGER.info(
-        f"CONV2D: x={tuple(x.shape)} w={tuple(w_oihw.shape)} stride={stride} "
-        f"padding={padding} dilation={dilation} layout={layout} "
-        f"dtype={x.dtype} bias={'yes' if bias is not None else 'no'} "
-        f"act={activation}"
-    )
+    if _LOGGER.get_logger().isEnabledFor(logging.INFO):
```
brunomazzottiamd: This is a very strange logging pattern. You shouldn't check if a logging level is enabled before emitting a log message, this is a responsibility of the logging library.

Furthermore, using f-strings directly inside Python's built-in logging functions is generally considered an anti-pattern. While Python will allow it, using an f-string evaluates and builds the string eagerly before checking the log level. The best practice for standard Python logging is to use lazy evaluation with `%` placeholders:

```python
_LOGGER.info(
    "CONV2D: x=%s w=%s stride=%s padding=%s dilation=%s layout=%s dtype=%s bias=%s act=%s",
    tuple(x.shape),
    tuple(w_oihw.shape),
    stride,
    padding,
    dilation,
    layout,
    x.dtype,
    "yes" if bias is not None else "no",
    activation,
)

```

brunomazzottiamd (commented):
Sharing some comments.

aiter/ops/triton/conv/_launch.py
saeid-rostami: This one helped on specific shapes specifically on gfx1151 but in overall model just 1-2% improvement. I am ok to remove it for now.

saeid-rostami:
> Measured on **gfx1150** (Radeon 890M, RDNA 3.5), Windows with native ROCm, torch `2.11.0+rocm7.15.0a20260728`, Triton `3.7.1`, on 2026-08-21 against `main` at `ba59a37` and this branch at `e363cf6`. I maintain a downstream consumer that routes diffusion VAE-decode convolutions through `aiter.ops.triton.conv`. My primary interest is **gfx1101** (RX 7800 XT), which I cannot measure at the moment; gfx1150 is the card I have to hand, and neither ships a conv table, so they are in the same position with respect to this PR.
> 
> **The performance direction of this PR is good on this part**, which is why I would like it to stay reachable. On the routes it keeps — `cblocked`, fp16, N=1, K=C, interleaved A-B-A-B in one session, the two runs per revision shown rather than averaged:
> 
> shape	`main`	this PR	change
> C=512, 128×128, 3×3	10.575 / 10.640 ms	10.520 / 10.482 ms	−1 %
> C=512, 256×256, 3×3	41.706 / 41.043 ms	40.992 / 40.567 ms	−1.5 %
> C=256, 512×512, 3×3	41.651 / 42.563 ms	39.344 / 39.325 ms	**−6.6 %**
> C=128, 1024×1024, 3×3	43.978 / 44.401 ms	37.485 / 37.546 ms	**−15.0 %**
> The new packing kernel accounts for it: pack cost drops from 9.37 to 5.85 ms at 1024×1024, about 38 % cheaper than the host-side `permute().contiguous()` on `main`. Separately, `conv2d_1x1(layout="nhwc")` is 2.3× faster here (2.85 → 1.25 ms).
> 
> ## The question
> On an architecture that ships no tuned table, the two revisions behave differently:
> 
> ```
> main @ ba59a37,    no config, AITER_TRITON_CONV_AUTOTUNE=0  ->  FileNotFoundError
> main @ ba59a37,    no config, AITER_TRITON_CONV_AUTOTUNE=1  ->  runs, rel_l2 7.67e-04
> this PR @ e363cf6, no config, AITER_TRITON_CONV_AUTOTUNE=1  ->  FileNotFoundError
> ```
> 
> Two changes combine: `_triton_kernels/conv/helpers.py` is deleted along with the seven `AUTOTUNE_*_CONFIGS`, so the env-var bypass documented in `conv/README.md` no longer exists; and `CONV-PREPACK` is a newly required per-arch family on a path that previously read no JSON at all, so it is the first lookup to fail. The per-architecture requirement goes from 7 tables to 9, and every shipped table is arch-specific. The updated README states which architectures ship one — gfx1100, gfx1151, gfx1200 and gfx1201 — and that is four of the eight in `RDNA_ARCHS`; the other four are gfx1030, gfx1101, gfx1102 and gfx1150, so conv2d becomes unreachable on RX 6800/6900, RX 7800 XT and 7700 XT, RX 7600 and Strix Point.
> 
> I am not claiming any of those is documented as supported; `op_tests/triton_tests/conv/_helpers.py` lists exactly the four that have tables, so the current state is self-consistent. The question is what the intended path is for the other four now that retuning moves out of the package, since the README paragraph describing how to run without a table is removed in this diff, and its replacement states that _"candidate search spaces and the tuning driver live outside the production package"_.
> 
> Two observations that may make this cheap to resolve. First, on this part the config _contents_ matter much less than their presence: a 48-combination sweep over `BLOCK_M`/`BLOCK_N`/`BLOCK_K`/`GROUP_SIZE_M`/`num_warps` on the four shapes above — route asserted as `DIRECT_NCHW_3X3`, output checked against `torch.nn.functional.conv2d` — reaches parity with the `cblocked` route at best (0.93–1.16× over three repeats), and simply borrowing a shipped table gives much the same (`gfx1151` 0.84×, `gfx1201` 0.85×). Second, it would have to be a _curated_ default rather than arbitrary values: in the same sweep `num_warps=2` with `BLOCK_N=128` produced 0.05×, twenty times slower than `cblocked`. A nearest-architecture fallback would therefore cost very little and remove a hard failure — and `has_conv_config()`, which this PR adds with `load_config_json(..., required=False)` and a docstring calling the table _optional_, is already the primitive for it. (The suppressed automated comment on `conv2d.py` proposes exactly that as a guard for `conv2d_nchw_3x3_direct()`; that is one entry point, whereas on an architecture with no table at all the first mandatory `get_conv_config` fails inside `load_config_json` before any `KeyError` path is reached.)
> 
> On gfx1101 the picture came out a little differently — there the table _contents_ did move the number. Measured in early August 2026 on the pre-#4869 tree, not re-verified since, and to one significant figure: repeating the comparison in a later session shifted the ratio by about two points, so I would not quote a second digit. The shipped `gfx1100` table from #4265 was indistinguishable from an RDNA4 table on that card — a ratio of 1.00 against a 6–8 % spread within a single arm — so same-generation donorship was not the lever. Changing one value in the RDNA4 table — `BLOCK_K` 64→32 in `CONV-3X3-CBLOCKED` — was worth about 10 % there, while the same edit applied to the `gfx1100` table was worth nothing, so the useful value is donor-specific rather than portable. And a held-out neighbour table cost about 9 % against one tuned on the card. So a neighbour table is worth far more than a hard failure, and still leaves something on the table; that ordering is what matters here.
> 
> Is a curated nearest-architecture fallback something you would consider, or is the intent that an architecture without a table goes through the offline tuning path first? Either answer is useful — it tells downstream consumers on those four architectures what to plan for.

@Ragua1 
Thanks for the detailed measurements and analysis. It is especially useful to see that the packing kernel improves the larger cblocked workloads on gfx1150.
The removal of AITER_TRITON_CONV_AUTOTUNE and the in-kernel autotuning configurations was intentional. After discussing this internally, I was asked to follow the AITER policy that production Triton kernels should not perform runtime autotuning or carry autotuning search spaces.
The intended workflow for an architecture without Conv2D configuration files is to tune the kernels offline, add the resulting per-architecture JSON configuration files, and submit them to AITER in a PR. We do not currently intend to select an arbitrary or nearest-architecture configuration automatically, since—as your gfx1101 results also demonstrate—configurations are not always portable and a poorly matched fallback can cause a substantial performance regression.
Therefore, the current failure on architectures without the required configuration tables is expected under this policy. If you would like to add support for gfx1101, gfx1102, gfx1150, or another missing architecture, please tune and submit the corresponding Conv2D configuration files and add me as a reviewer. I would be happy to help review and validate them.

aiter/ops/triton/conv/_launch.py
brunomazzottiamd: I have a feeling that's probably noise! Let's keep the code as-is for now and get this PR merged. Later on, you can perform further investigations and propose a cleanup PR.

brunomazzottiamd (approved):
LGTM!

PR #4538, opened 2026-08-03
[FlyDSL] gfx950 FP8 MQA logits indexer kernel

vpietila-amd:
## Motivation

This PR introduces a FlyDSL kernel for computing DeepSeek-V4 sparse attention lightning indexer on `gfx950`. The new kernel is drop-in replacement for the existing Triton/Gluon kernel. For the DSv4 prefill shapes it generally faster than the existing Triton/Gluon kernel. The `gfx942` version was introduced in an earlier PR #3913.

## Technical Details

There are two alternative implementation on `gfx950`

- Direct load kernel similar `gfx942`: Data is streamed directly from DRAM to registers.
- LDS staged kernel where KV slices are stored on LDS and shared among all wavefronts where each wavefront computes it's own disjoint set of query rows. Q fragments and per-row weights are preloaded into registers once per block (similar to the direct load kernel. The LDS staging uses gfx950 specific async global → LDS DMA which writes directly to LDS without passing through VGPRs. XOR swizzle is used for LDS writes and reads to reduce the rate of bank conflicts.

The LDS staged instance can have two or more LDS buffer combined with one or more prefetch stages. Allocation of one LDS buffer more than the prefetch depth (`PD`) enables to reduce the number of required barriers by one (barriers are required to ensure data is ready before running MFMA on it). In practice, the most performant combination is 3 LDS buffers with `PD=2` (only one barrier required).

There is instance auto-selection which selects the best kernel for a given shape. However, similarly to the existing implementation, the auto-selected instance can be overridden with an environment variable. 

There's also one common optimization for both `gfx942` and `gfx950`: when `clean_logits=True`, the baseline implementation uses `torch.fill` to initialize the logits matrix with `-inf` values. Since half of the values are subsequently overwritten by the logits kernel, we have fused the "clean logits" initialization step into the logits kernel. This provides performance boost for small sequence lengths.

## Test Plan

All tests added in PR #3913 are now enabled for `gfx950`. For performance, benchmarked the same DSv4 shapes in PR #3913

## Test Result

All unit tests pass on both `gfx942` and `gfx950` (note: they are not enabled as part of the CI workflow)

For `gfx950`, the new kernel is compared against the existing Triton/Gluon kernel

Shape | Triton TFLOPs | FlyDSL TFLOPs | FlyDSL instance | Verification | Speedup
-- | -- | -- | -- | -- | --
[...]

PR #4870, opened 2026-08-20
Add kernel PR validation and structural D9 scanning

jhinpan:
## Motivation

`review-pr` is static and stochastic: it can surface useful risks, but it cannot establish runtime behavior or act as a merge gate. The original D9 rule also missed three known 32-bit index-overflow defects because its trigger depended on variable names rather than index × stride structure.

This PR separates the two authority levels:

- `review-pr` is explicitly advisory. Its LLM judgement never blocks a merge.
- `validate-kernel-pr` supplies reproducible execution evidence. Only a deterministic validator finding may be treated as a blocker.
- `scan_index_width.py` supplies structural D9 candidates; production-scale evidence is still required before a candidate becomes a defect.

## Validation pipeline

`merge_sim → gpu_claim → runtime_compat → test_policy → baseline_control → correctness_repo_tests → correctness_s1_grid → execution_receipt → index_width_scan → verdict`

- `merge_sim` applies the exact PR diff to the live base-branch tip.
- `gpu_claim` uses the shipped `pick-idle-gpu.py` to sample activity plus resident VRAM, maps the selected HIP index to the matching AMD SMI device, and holds the lock for the full run; metadata probes reuse the picker's ROCm `amdsmi` import fallback.
- `runtime_compat` records interpreter, package, and loaded native-library identities. A prebuilt `PYLIB` cannot validate FlyDSL runtime/build-input changes because trusted source-to-binary provenance is not implemented.
- `test_policy` compares tolerances and disabled shape rows against base before execution.
- `baseline_control` runs the same target with the exact patch removed under base-only caches.
- Runner selection is structural: an explicit `path::node` selector or top-level `test*`/`Test*` definition uses pytest; a file with only a `__main__` entry point runs as a script; a target with neither is skipped.
- `correctness_repo_tests` and `correctness_s1_grid` record runner-specific execution counters; pytest uses JUnit counts, while scripts use process-based counts. All-skipped or unrunnable targets are not passes.
- `test_selection.runner` and `runner_reason` preserve the choice, and `arch_coverage_basis` distinguishes pytest JUnit execution from script exit/output evidence.
- `execution_receipt` is generated by a validator-owned, private pytest profiler. Because this receipt is pytest-only, a successful script target tops out at `INCONCLUSIVE`; `PASS` still requires the expected `module:function` route, an observed route symbol, and every grid shape reconstructed from the route's actual call locals.
- `index_width_scan` records deterministic index × stride candidates for D9 review.
- `verdict` is `BLOCK`, `NEEDS_WORK`, `INCONCLUSIVE`, or `PASS`; process exits are respectively `1`, `1`, `2`, and `0`.

The final review card keeps `Review (advisory)` and `Validation (deterministic)` on separate lines.

## Evidence boundary

A report is accepted only when its PR head, live base tip, and diff SHA-256 match the current PR. `review-pr` runs the canonical JSON schema, checks JUnit arithmetic, runtime identity, route/shape receipt, findings, and process-exit contract, then independently recomputes the verdict.

The executor also:

- fails loudly when required fetches, scanners, schema, or scale evidence are unavailable;
- rejects non-clean worktrees, including ignored artifacts;
- uses separate base/head caches, automatically assigns phase-private `AITER_JIT_DIR` roots, disables Python bytecode writes, and aborts the head run if base leaves artifacts;
- distinguishes a PR-added failing test from a pre-existing failure;
- returns `INCONCLUSIVE`, not `PASS`, when GPU, runtime, baseline, pytest, route, or shape evidence is missing;
- adds runtime architecture coverage only after the selected head target actually executes, with a runner-specific evidence basis.

`downstream-impact-check` was removed because it addresses a separate incident and should be reviewed independently.

## Reproducible evidence

The committed corpus pins `ROCm/FlyDSL@421935cc6f09fd9b27d5d5ae52e0960e18834bd5` and contains one behavior-neutral control plus three seeded softmax defects.

On MI355X (`gfx950`, HIP 1, BDF `0000:05:00.0`), with pre/post GFX activity at 0–1%:

- clean control: `PASS`, process exit `0`, route and all four shapes observed;
- dropped tail mask: `BLOCK`, process exit `1`, from the independent shape grid;
- widened tolerance: `BLOCK`, process exit `1`, from `test_policy`;
- vector/element index mismatch: `BLOCK`, process exit `1`, from correctness.

Twenty-four synthetic behavior partitions pass, including all-skipped pytest, missing/wrong/incomplete receipts, worktree probe shadowing, runtime/native/package-source changes, contradictory reports, cache contamination, and exit-code mapping. The schema also rejects hollow `PASS` reports with failed/error tests, malformed counters, zero execution, missing identity, or wrong exit status.

Revalidation of real #4538 at head `2285090` on gfx950 selected the script runner and completed 56/56 cases with no blocker; the report correctly remained `INCONCLUSIVE` with process exit `2` because route receipts are pytest-only and no independent grid hook was available.

## Limitations

- The caller still chooses `--target`, `--expected-route`, and `--shape-vars`; automatic relevance selection is not implemented.
- External grid adapters remain deferred: a script-only target without a shape hook cannot yet use `--extra-target`, because the adapter needs its own hash identity without changing the PR diff hash or live-base identity.
- Validator-owned in-process profiling prevents accidental and worktree-shadowed receipts, but arbitrary hostile Python in the same pytest process can still spoof frames. A hostile-code gate needs an out-of-process HIP/rocprof trace.
- Trusted source-to-binary provenance is not implemented. FlyDSL runtime/build-input changes therefore remain `INCONCLUSIVE` with a prebuilt `PYLIB` rather than accepting caller-authored provenance.
- The committed replay asset is currently the four-case seeded corpus. The historical 14-PR campaign is local, so this PR makes no durable recall@5 or consistency claim.
- `perf` and performance-claim reproduction are not implemented.
- Only gfx950 was executed here; no gfx942 or gfx1250 coverage is claimed.


zufayu:
## Proposed acceptance bar, so this can keep converging

The measurement discipline here is the best this repo has applied to tooling — reporting that the
headline D9 change caught **0 of 3** across three controlled arms, against your own interest, is
why the rest is worth writing down. The gap isn't quality; it's that the PR exercises more
authority than its evidence supports, and there's no written bar for when that authority is earned.

**1. Declare the tier.** *Advisory* (hints; human reviews as before) needs only proof of no net
harm. *Triage* (a clean verdict means the human skims) needs a bounded false-clearance rate.
*Gate* (🔴 blocks merge) needs determinism and an appeal path. The evidence reaches advisory:
single-run backtest, self-adjudicated precision, no false-clearance measurement. But `review-pr`
now says a kernel PR **must** produce a report and its own green suite "is not evidence" — that's
triage/gate language. Put the tier in the skill header and move it deliberately, not by drift.

Corollary: **an LLM judgement never gates; only a deterministic executor can.** `validate_pr.sh`
blockers are reproducible facts. A `review-pr` 🔴 is one sample from a stochastic process. They
share a verdict slot today; splitting them is a template edit:

```
Verdict:    ⚠️ NEEDS WORK     <- LLM judgement, advisory, never blocks
Validation: BLOCK             <- deterministic fact, may block
```

**2. The durable asset is the corpus, not the skill text.** 14 PRs, 17 defects, each pinned to its
merge commit with `gh` time-boxed and labels taken from the `[Fix]`/revert PRs — that methodology is
the hard part and it works, but today it lives in a PR description and evaporates on merge. As
`defects.jsonl` + a replay script + a committed `metrics.json`, rule text becomes cheap: any rewrite
is scored in one command, and every future `[Fix]` merged to `main` appends a case for free. **Split
`train`/`holdout` now** — otherwise the ratchet overfits, which the production-scale table already
half-demonstrates (0/3 → 1/3, on scale facts taken from the fix PRs that also define the labels). Cheap now,
expensive once numbers are quoted.

**3. Then acceptance is a ratchet, not a debate:** no per-family recall@5 may regress, false
clearance may not rise, and any "this is better" claim needs n≥5 with variance. It also settles the
5-finding cap by plotting recall@k instead of arguing it.

### Phase 0 — this PR: stop asserting cleanliness that wasn't established

A silent miss costs nothing; the reviewer's skepticism survives. A false clearance spends it. A
static reviewer can't do this — adding a runtime layer creates the failure mode, so the guardrail
belongs in the same PR.

| # | Fix | Why |
|---|---|---|
| 1 | Missing/failed scanner must be loud | `[ -x "$SCAN" ] &&` is silent, and D9 tells the reviewer to state the list is empty |
| 2 | Resolve scanner paths against the skill dir | Step 1 uses `.claude/skills/...` with no `cd`; wrong cwd = silent skip |
| 3 | Drop or add `review-flydsl-kernel/*` | In neither this PR nor `main` — permanently dead |
| 4 | Use `git stash -u`, and abort if the stash fails | `git apply` leaves added files **untracked**, and `git stash` without `-u` skips them. Verified locally: on a patch that only adds files, `git stash -q` returns 0 having saved nothing and `git stash pop -q` then returns 1 into `2>/dev/null`. The "base" run therefore still contains the PR's new files, so `BASE_RC` is the head result — and a PR that adds a failing test is reported as *"red BOTH with and without this change — pre-existing, not attributable to the PR"*. This fires on the most common kernel-PR shape: new kernel plus new test. On a mixed patch it is partial — tracked edits revert, added files do not |
| 5 | Make every skip declare itself | Two ways the report loses one. (a) `jset ... '"'"'{...}'"'"'` yields the single argument `"'{status:skip,note:pytest` (ran it), while the schema requires `stages.*` to be an object with `status`. (b) In the COMPILE_ONLY branch only `correctness_repo_tests` is written; `correctness_s1_grid` is never set at all, so the skill's signature stage vanishes from the report rather than declaring itself skipped — confirmed in a real run whose `stages` were exactly `merge_sim, gpu_claim, runtime_compat, correctness_repo_tests`. Both undercut the schema's own "do not overclaim by omission" |
| 6 | Make the runtime probe repo-aware | Stage 3 imports `flydsl` whatever repo is under test, so for an aiter PR the outcome is decided by whether an unrelated package happens to be installed on the host. Both branches run against an aiter worktree: with flydsl present, `runtime_compat: pass ("runtime 0.2.2")` and `verdict: PASS` — a clearance from a check that inspected nothing in the checkout; with it absent, `blocker: "checkout does not import against the pinned runtime"` and `verdict: BLOCK`, blaming an aiter checkout that never referenced flydsl. By this PR's own argument the passing branch is the worse one |
| 7 | Split Verdict / Validation lines | So a stochastic 🔴 and a reproducible blocker stop sharing one slot — see the corollary above |
| 8 | Split out `downstream-impact-check` | Unrelated to both stated motivations (its own is aiter#4530 / vLLM MXFP4) |
| 9 | Make the mutant table re-runnable | Seed of the replay script; without it every rule edit is unprotected — and this PR proves prose edits can do nothing (D9: 0/3, three times) |
| 10 | `gpu_claim` should report `skip`, not `fail` | No idle GPU is an environment fact, not a defect in the PR — the same distinction stage 3's own comment draws for a stale runtime. The schema says a stage that could not run "reports 'skip' with a note". Both runs above emitted `gpu_claim: {"status": "fail", "note": "no idle GPU after sampling window"}`, which propagates into the verdict logic as a real stage failure |

Everything else — recall@5 (unmeasured, though the cap is 5), n=1 across all four arms, an
unmeasured false-clearance rate, and adversarial evasion (the skill ships in-repo, so it's also an
evasion guide; `if WIDENED.search(code): continue` clears the whole line on any `int64` token, so
adding one unrelated `.to(tl.int64)` beside a `block_id * stride` hides a defect the scanner
currently catches) — is follow-up. Happy to open
those as issues with thresholds and kill criteria if the framing is useful.

<details>
<summary><b>Appendix — the measurement programme behind the bar above</b> (phases, thresholds, kill criteria)</summary>

### Phase 1 — build the asset

- `corpus/defects.jsonl` — one record per labelled defect: introducing PR, merge SHA, file/line,
  family, severity, fixing PR, one-line description.
- `replay.sh` — checks out each case at its merge commit, time-boxes `gh`, runs the skill under
  test, scores against the labels, repeats n times.
- `metrics.json` — committed: corpus version, skill SHA, per-family recall@5, false-clearance rate,
  findings/PR, evasion rate, n, variance.
- Appending a case becomes part of handling any `[Fix]`/revert PR — the corpus then grows for free.

**Exit:** one command reproduces the 9/17 number, and the ratchet is enforceable.

### Phase 2 — pay the measurement debt

| | Gap |
|---|---|
| 11 | **recall@5 unmeasured.** The cap is set at 5 and justified from "what a maintainer reads", but the reported 9/17 is uncapped. Per-family recall *after* truncation is the number the cap should be read off. Also worth reconciling before these become a baseline: the description reports uncapped findings/PR as both **7.1** (Technical Details) and **7.7** (Limitations) for the same 14-PR backtest |
| 12 | **n=1 everywhere.** Across `0/3, 0/3, 0/3, 1/3`, nothing separates the single 1 from sampling noise |
| 13 | **False clearance never scored.** `report_schema.json` was designed for exactly this (`degraded_mode`, `arch_coverage`, `isolation`) — the intent is right, nothing measures it. Note the `arch_coverage` example violates its own enum (`"compile-only (no gfx942 device on this host)"` vs `["runtime","compile-only","not-covered"]`), and `validate_pr.sh` writes that same string |
| 14 | **No adversarial measurement.** Seeded mutants are non-adversarial — planted by the rules' author to match the rules. The skill ships in-repo, so an author generating a kernel with Claude Code loads it: the ruleset is detector and evasion guide at once. Name-list D9 had an evasion cost of one rename, which is why 0/3 was structural rather than unlucky. Suggested measure: agent A holds the skill and writes a patch with a real defect that passes; agent B reviews with the same skill; evasion rate = pass rate |

**Exit:** the tier-B bar is met, or a kill criterion fires.

### Phase 3 — tier B (triage), with the joint system watched

| Metric | Bar |
|---|---|
| False clearance, per assertion type | Measured; 🔴 families near zero |
| Per-family recall@5 | No family at zero (D9 is 0/3 today) |
| Run-to-run consistency | 🔴 modal agreement ≥ 0.8; top-5 Jaccard ≥ 0.6 |
| Attribution error | Both directions; pre-existing red charged to the author ≈ 0 |
| Human-minutes per true defect | Below unaided review |

Tier C additionally needs deterministic stages only, a false-block SLA, and an appeal path. The LLM
half stops permanently at B.

Every number in this table is a placeholder with no empirical basis — I made them up to give the
bar a shape. They should be set from the first `metrics.json` baseline and then only ratcheted.
Quoting them as if they were derived would be the same error this comment is about.

### The unit of evaluation is human + tool

The baseline is not "no tool", it's "a human reviews it". A tool that raises recall can still make
the joint system worse if it turns reviewers from reading the diff into checking whether the card is
red. Track from Phase 1; required before any tier-B claim:

- **Rubber-stamp rate** — share of ✅ cards after which a human still leaves a substantive comment.
  A fall is the warning sign, not the success metric.
- **Human review minutes per PR**, before/after.
- **Escape rate** (lagging) — merged PRs later repaired by a `[Fix]`/revert. The only metric that
  ultimately matters; quarterly, use it to check the proxies still hold.

### Kill criteria

A bar that cannot fail is not a bar — the same argument this PR makes about test suites.

1. With the corpus in place: no defect family clears recall@5 > 0.5 on holdout **and** red-team
   evasion > 0.7 → static+LLM is the wrong instrument for AI-authored code; move the effort into the
   deterministic validator.
2. Rubber-stamp rate falls >30% with no matching fall in escape rate → joint system is net negative;
   revert to advisory or withdraw.
3. False clearance cannot be driven near zero after two rounds → permanently advisory, never triage.

### Docs should state

- A `review-pr` 🔴 is advisory; only a `validation_report` blocker may block.
- Absence of a `perf` stage is not "no regression" (already stated — keep it).
- Reported recall is an **upper bound**: labels come only from defects later found and fixed, so
  defects still latent on `main` can't enter the label set — and those are the hardest class.
- `production_scale.md` is partly in-sample — restate it in the file, which outlives this description.

</details>

jhinpan:
@zufayu We have reworked this PR against the acceptance bar in your comment and would appreciate another review of the current head (`2d6b3a514`).

The Phase 0 issues are now addressed:

- `review-pr` is explicitly advisory, with separate `Review` and deterministic `Validation` verdicts.
- Required scanners and production-scale evidence resolve from the skill root and fail loudly; the dead `review-flydsl-kernel/*` references are gone.
- Baseline attribution no longer uses the shared stash stack. Base/head use the exact patch, separate caches, and clean-worktree checks, including ignored artifacts and PR-added files.
- Every required stage is a schema-constrained object; missing environment evidence produces `INCONCLUSIVE`, never a clearance.
- Runtime probing is repo-aware, no-GPU is `skip`, HIP/AMD-SMI identity is mapped by device, and the GPU lock is held for the complete run.
- Validation reports are bound to the current head, live base tip, and diff SHA-256; the reviewer recomputes the verdict and records the exact target/grid.
- The independent grid now has a runtime handshake proving that the selected test consumes the shape environment.
- `downstream-impact-check` was removed from this PR.
- The mutant table is now a pinned, replayable corpus: the clean control passes, while the tail-mask, tolerance, and vector-index mutants block in their expected stages on gfx950.

The description now calls out the remaining boundaries: target selection, perf/claim reproduction, gfx942 coverage, and the uncommitted historical train/holdout corpus. Could you please take another look?

jhinpan:
@zufayu One more evidence-hardening follow-up is now at head `1f85dab9a` before your re-review.

The historical GPU campaign exposed that green pytest alone still did not prove route/shape execution. This head now requires validator-owned observed evidence for `PASS`: JUnit executed/skipped counters, an observed `module:function` route, route-call shape locals covering every grid row, runtime/native identity, and verdict-matched process exits. Missing or contradictory evidence is `INCONCLUSIVE`.

It also refuses prebuilt `PYLIB` for FlyDSL runtime/build-input changes because we do not yet have trusted source-to-binary provenance; caller-authored manifests are not accepted. The remaining same-process adversarial limitation is called out explicitly and would require out-of-process HIP/rocprof tracing.

The clean control now returns `PASS/0`; all three mutants return `BLOCK/1` in their expected stages. Could you review this latest head rather than the earlier `2d6b3a514` snapshot?

zhiding512:
### `validate_pr.sh` assumes every target is a pytest suite, which produces a false `BLOCK`

I ran this skill at head `1f85dab9a` against a live kernel PR on gfx950 and it blocked a PR whose test suite actually passes. The cause is that the target is executed only one way:

```bash
# validate_pr.sh:844
python -m pytest -p "$PROBE_MODULE" "$TESTS" -x -q ...
```

When the target is not a pytest suite, pytest collects nothing and exits 5. `correctness_repo_tests` records that as `fail`, and because the file is new in the PR, `baseline_control` reports `target-not-present` on base — so the failure is attributed to the author as a blocker.

#### Reproducer

ROCm/aiter#4538 (head `2285090`), a FlyDSL gfx950 kernel PR. Its `op_tests/test_flydsl_fp8_mqa_logits.py` exposes `verify_fp8_mqa_logits` and `bench_fp8_mqa_logits`; neither matches pytest's default `python_functions = test*`, and the repo sets no override in `pytest.ini` / `pyproject.toml` / `setup.cfg`.

```console
$ validate_pr.sh --repo <base worktree> --patch pr4538.patch \
    --head-sha 22850902990532a5734c49cabae79246965ba8db \
    --tests op_tests/test_flydsl_fp8_mqa_logits.py \
    --expected-route op_tests.test_flydsl_fp8_mqa_logits:verify_fp8_mqa_logits
verdict=BLOCK  findings=3
  [blocker] correctness: the PR adds this test target and it fails on head:  no tests ran in 19.66s
$ echo $?
1
```

Stages: `merge_sim` pass, `gpu_claim` pass (gfx950, idle), `runtime_compat` pass, `test_policy` pass, `baseline_control` pass, **`correctness_repo_tests` fail (exit 5)**, `correctness_s1_grid` skip, `execution_receipt` skip, `index_width_scan` info.

The same file, run the way `.github/scripts/aiter_test.sh` runs it — `timeout 60m python3 "$file"`, hitting the `__main__` guard — passes **56/56 cases** at that same head, on MI35X (gfx950, 24.4 s) and MI300X (gfx942, 49.7 s). So the blocker describes the harness's execution assumption, not the kernel.

#### Why I think this belongs in Phase 0 rather than follow-up

It is the same class as item 10 in @zufayu's acceptance bar — *"no idle GPU is an environment fact, not a defect in the PR"*. A target that is not pytest-shaped is a fact about the PR's test packaging, and the schema's own rule is that a stage which could not run reports `skip` with a note. Here it reports `fail` and escalates to a blocker with exit 1.

It also cuts against this PR's central argument. The design says an LLM `🔴` never gates and only a deterministic executor may, which makes a deterministic false block far more expensive than a missed defect: it spends precisely the credibility the two-tier split exists to establish.

Scope, measured on `op_tests/test_*.py` at depth 1 (the exact set aiter CI shards): 117 targets, **107** pytest-collectable, **7** script-only, 3 neither. So this is a minority — but a named one, and disproportionately FlyDSL:

```
test_flydsl_pa_mqa_logits_fp4_prefill.py
test_mha_flydsl_varlen.py
test_mla_reduce.py
test_groupnorm.py
test_moe_topk_gating.py
test_topk_per_row_stable.py
test_topk_plain.py
```

That matters because the `review-pr` change in this PR makes a validation report **mandatory** for FlyDSL/Triton kernel PRs. The targets most likely to be validated are over-represented in the set that cannot be validated.

#### Suggested fix

Detect the runner instead of assuming it — parse the target and pick what it supports:

| target shape | runner |
|---|---|
| caller passed `path::name` | pytest |
| defines `test*` functions or `Test*` classes | pytest |
| no `test*`, but has `if __name__ == "__main__"` | `python <file>` |
| neither | none → correctness `skip`, verdict `INCONCLUSIVE` |

An AST scan is enough and avoids paying the ~15 s import twice. Recording the choice and its justification (`test_selection.runner`, `test_selection.runner_reason`) keeps the report honest about how the verdict was obtained, and naming each log after the runner used avoids filing a script run under a pytest-looking path.

The last row is the important one: a target with no runnable entry point should be `skip`, never `fail`. That is a real limitation on what the report proves, and it belongs in `review-pr`'s advisory judgement rather than masquerading as a reproducible kernel defect.

I have this working locally against #4538 — same PR, same GPU — and it turns the false `BLOCK` into `INCONCLUSIVE` with `correctness_repo_tests: pass (exit 0)`, `runner: script`, and `arch_coverage: {"gfx950": "runtime"}`. It stays `INCONCLUSIVE` for an honest reason: that target takes shapes only from argparse and reads no environment variable, so no independent grid can be injected. Happy to open it as a PR against your branch if useful.

#### Two second-order consequences of the same assumption

1. **`mark_runtime_coverage` is pytest-only.** It matches `collected N items` / `N passed`, which a script target never prints, so a run that genuinely executed on the claimed device earns no `arch_coverage`. Per-runner evidence fixes it, but the two are not equally strong — a script's exit-0-with-output is weaker proof than a case count, so I record which one was obtained in a sibling `arch_coverage_basis` field rather than flattening both into the same word.

2. **`execution_receipt` cannot be satisfied by a script target at all**, since the receipt comes from a validator-owned pytest plugin (`-p "$PROBE_MODULE"`). With an observed route required for `PASS`, script targets top out at `INCONCLUSIVE` even once they execute correctly. That may well be the right call given the receipt's threat model, but it seems worth stating deliberately in the skill rather than emerging from the runner assumption — otherwise "we could not instrument it" and "it did not run" stay indistinguishable in the report.

Environment for the run above: gfx950 (MI355X), ROCm 7.2.4, torch 2.10.0, FlyDSL 0.3.1, one GPU locked for the whole run. Two notes in case they save someone time: `merge_sim` rejects a worktree holding any **gitignored** artifact and `baseline_control` fails if the base run creates new ones, so an aiter checkout needs `AITER_JIT_DIR` pointed outside the worktree and `PYTHONDONTWRITEBYTECODE=1`, otherwise its own JIT output and `__pycache__` trip both gates. And `pick-idle-gpu.py` is not in this PR or on main; I wrote one to the documented contract, and on this host `amdsmi_get_gpu_process_list` reports the same PID against all 8 handles, so idleness had to come from sampled activity plus resident VRAM rather than the process list.

zhiding512:
### Feedback from running this skill end-to-end on a real PR

I used `validate-kernel-pr` to validate #4538 (FlyDSL gfx950 fp8_mqa_logits) on an MI355X, and hit four things worth fixing. All four are verified against this PR's current head `1f85dab`, not against an older revision. The pipeline is genuinely useful once it runs — `merge_sim`, `gpu_claim`, `test_policy` and `baseline_control` all did exactly what they promise, and the base-vs-head control is what let me establish that #4538 is a throughput change rather than a correctness fix on gfx950.

---

**1. The runner is assumed to be pytest, which turns a packaging mismatch into a false `BLOCK`.**

`validate_pr.sh:844` runs `python -m pytest -p "$PROBE_MODULE" "$TESTS"` unconditionally. But aiter's `op_tests/*.py` are executable scripts driven by a `__main__` guard — `.github/scripts/aiter_test.sh` runs them as `timeout 60m python3 "$file"` — and pytest collects nothing from them:

```console
$ python -m pytest op_tests/test_flydsl_fp8_mqa_logits.py --collect-only -q
no tests collected in 15.39s      # exit 5
```

The validator recorded that as `correctness_repo_tests: fail` and emitted:

> [blocker] correctness: the PR adds this test target and it fails on head: no tests ran in 15.44s

That is a red verdict with no defect behind it, which I'd argue is worse than not running at all, because it looks like evidence. The same target run the way its repo runs it passes 56/56 on gfx950 in 24 s.

`execution_receipt` makes this sharper rather than softer: it is required for `PASS` (`validate_pr.sh:232`, and again in the `review-pr` consumer) and is produced by an in-process pytest plugin keyed on a `module:function` route. A script-style target cannot produce a receipt at all, so it can never reach `PASS` no matter how much real work it does on the GPU.

What I did locally: parse the target and pick the runner from it — pytest for `path::name` selectors or files defining `test*`/`Test*`, `python <file>` otherwise if there is a `__main__` guard, and `runner: none` → correctness `skip` (never `fail`) when there is no runnable entry point. The choice and its reason go into `test_selection.runner` / `runner_reason`, and logs get named after the runner used. `mark_runtime_coverage` needs the same treatment — it matches `collected N items` / `N passed`, which a script target never prints, so a run that genuinely executed on the claimed device earns no `arch_coverage`. I record which evidence kind was obtained in a separate `arch_coverage_basis` field, since a script's exit-0-with-output is weaker proof than a pytest case count and shouldn't be flattened into the same word.

Whether script targets should also be able to earn a receipt is a real design question — the route/shape reconstruction is the strongest part of this PR and I'd rather not see it dropped. But `fail` is the wrong outcome for "this target isn't shaped like a pytest suite".

---

**2. `pick-idle-gpu.py` is required but not shipped anywhere.**

`gpu_claim` needs it (`validate_pr.sh:371-374`), SKILL.md documents `PICKER` as a knob and shows `pick-idle-gpu.py --samples 10 --interval 1 --quiet` in the invocation example — but it is not in this PR's file list, not in the repo, and not on `PATH` in the ROCm images I tried. Without it every run is `degraded_mode: NO_GPU` → correctness skipped → `INCONCLUSIVE`, on a box with eight idle gfx950s. That is the difference between the skill working and not working, so it probably belongs in the PR.

One implementation note if you do add it, because I got this wrong first: **do not judge idleness by the KFD process list.** On this host `amdsmi_get_gpu_process_list()` returns the same PID against all 8 handles, so a process-based picker rejects every device. Sampled `gfx_activity` **plus** resident VRAM works — it cleanly separated the one loaded device (101 GB resident, 85% peak activity) from the seven idle ones (284 MB, 0%). Also worth documenting that `amdsmi` lives in ROCm's `dist-packages` and is not visible from a venv, so the `import amdsmi` inside `gpu_claim` needs `PYTHONPATH` help under the interpreter the caller actually uses.

---

**3. The ignored-artifact rule makes an aiter checkout unvalidatable by default, and that prerequisite isn't documented.**

`merge_sim` rejects any tracked, untracked **or gitignored** artifact (`validate_pr.sh:337`), and `baseline_control` additionally fails if the base run creates new ignored files (`959-960`). Both rules are right. The problem is that aiter JIT-builds into its own checkout — `aiter/jit/build/...`, `aiter/jit/module_aiter_core.so` — and writes `__pycache__`, so a first run trips `merge_sim`, and a cleaned run trips `baseline_control` the moment the base pytest executes. It looks like the skill is broken rather than like a missing prerequisite.

Two exports fix it, and I think they belong in SKILL.md's invocation block:

```bash
export AITER_JIT_DIR=/tmp/.../jit     # keep JIT output out of the worktree
export PYTHONDONTWRITEBYTECODE=1      # ...and __pycache__ too
```

---

**4. The S1 grid is unreachable for exactly the PRs that most need it.**

`correctness_s1_grid` only runs when the target reads the `--shape-env` var. #4538's test takes shapes solely from argparse `--shapes`, so the grid stage is `skip` and the verdict is `INCONCLUSIVE` — correct behaviour, but it means the stage the skill treats as its most valuable contribution is silently unavailable for any PR that didn't happen to expose an env hook.

The natural workaround is to supply a small grid-capable target yourself. That is currently forbidden by two rules that are individually right and jointly exclusive:

- the clean-worktree rule means a supplied harness has to be committed (untracked or ignored both fail `merge_sim`);
- committing it moves `repo.base` off the live branch tip, and the `review-pr` consumer rejects the report as "used a stale merge base";
- folding the harness into the patch instead changes `patch_sha256`, which the consumer also rejects.

So I ended up producing two reports for #4538: one that satisfies the evidence boundary but can only test what the PR ships, and one that carries the real grid evidence but is disqualified as PR evidence. Both are honest, but a reviewer should not have to choose.

A `--extra-target <path>` that the executor applies on top of the patch and records separately — hashed into the report, excluded from `repo.base` and `patch_sha256` — would close this without weakening the identity check. Happy to prototype it if you think the shape is right.

---

For what it's worth, the parts I leaned on hardest all held up: the base-vs-head control correctly reported `target-not-present` for a PR-added file rather than calling it a pre-existing failure, the invalid-grid positive control (`__VALIDATOR_INVALID_GRID__`) correctly proved my target really consumed the grid, and `index_width_scan` surfaced the one candidate in #4538 that needed a scale judgement. I also seeded a defect into the kernel under test to confirm my target could actually fail before trusting a green run, per the "a check never observed failing is decoration" rule — that discipline is the best thing in this skill.

jhinpan:
@zhiding512, thank you for the detailed end-to-end report. At exact head `706eafafc`, items 1–3 are addressed: targets now select pytest, script, or skip structurally and record the runner plus evidence basis; the AMD GPU picker ships with the skill and its `amdsmi` fallback is reused by metadata probes; and base/head runs receive phase-private `AITER_JIT_DIR` paths automatically. I revalidated real #4538 on gfx950: its script completed 56/56 with no blocker and correctly returned `INCONCLUSIVE`/exit 2 because route receipts remain pytest-only. Item 4 is deferred: `--extra-target` needs a separately hashed identity that preserves both the PR diff hash and live-base identity.

zufayu:
### Held-out validation at head `706eafafc`: four changes, and the plan after them

The PR is in good shape. It supplies what this repo did not have — reproducible execution evidence
rather than a second stochastic opinion — and the Phase 0 items from my last comment are all
addressed: the advisory/deterministic split is real rather than nominal, the mutant table is now a
pinned replayable corpus, and six commits in two days each track a specific objection. Reporting
that the headline D9 change caught 0 of 3 across three controlled arms, against your own interest,
is what made the rest worth engaging with closely.

So I ran it against work it had never seen, to find where it breaks. Most of it held. Four things
below did not, and none of them are architectural.

I validated ROCm/aiter#5089 with this skill at exact head `706eafafc`, on a host unlike the one in
the description: **MI308X (gfx942), ROCm 7.0.51831, torch 2.9.1, 8 GPUs**, against live base tip
`f4e7c7509`, PR head `6183413ff8ec`. Nobody had tuned the skill against that PR and the description
claims no gfx942 coverage, so this is held-out, and the first gfx942 datapoint for it.

---

## 1. `amdsmi_get_gpu_activity` is a hard dependency, and `PICKER` cannot route around it

On this host that call raises `AMDSMI_STATUS_UNEXPECTED_DATA (43)`, while
`amdsmi_get_gpu_enumeration_info`, `..._device_bdf`, `..._asic_info` and `..._vram_usage` all work —
resident VRAM separates the devices cleanly (7 GPUs at 0.3 GiB used / 191.7 GiB free). Idleness is
establishable here; the validator just cannot establish it. Three call sites:

| site | effect |
|---|---|
| `pick-idle-gpu.py:67` | exception escapes `sample()` → picker exits 2 → `gpu_claim: skip` |
| `validate_pr.sh:462` | gpu_claim metadata probe → `GPU_INFO_RC != 0` → `gpu_claim: skip` |
| `validate_pr.sh:314` | `record_gpu_activity_after` — tolerated, but prints a traceback that reads as a crash |

The second is what makes this structural: `SKILL.md` documents `PICKER` as the override knob, but
the executor re-calls the same API inline, so supplying a working picker does not help.

As shipped the run ends `degraded_mode: NO_GPU`, all four correctness stages `skip`,
`INCONCLUSIVE`/exit `2` — on a box with eight verified-idle GPUs.

**Change:** tolerate the missing metric at all three sites, and report the basis rather than
silently treating "activity unknown" as "activity 0" — that substitution is the
overclaim-by-omission this PR argues against. In the shape the PR already uses for
`arch_coverage_basis`:

    gpu_claim.idleness_basis: "activity+vram" | "vram-only"

Please also split the two conditions that currently share one string: "no GPU on this host" and
"idle GPUs present, activity API unavailable" both report `no verified-idle GPU was claimed`. The
first is an environment fact; the second is a portability gap in the validator.

## 2. The worktree is left patched, so a second run in it always fails

`merge_sim` applies the patch at `validate_pr.sh:372`. `BASE_ACTIVE=1` is set in exactly one place,
`:1039`, inside the baseline path. When correctness is skipped — e.g. the `NO_GPU` branch above —
that block never runs, `BASE_ACTIVE` stays `0`, and the `cleanup()` trap at `:288` does nothing. The
process exits with the candidate patch still applied.

Reproduced: after the first run, `git status --porcelain --ignored -uall` in the supplied worktree
was `M aiter/ops/flydsl/utils.py` + `?? op_tests/flydsl_tests/test_flydsl_utils.py` — exactly the
patch. The next run reported *"worktree is not isolated-clean, so merge simulation is inconclusive"*
plus eight `stage result was missing` notes, which reads as the caller's mistake.

**Change:** give the merge_sim application its own flag and revert it in `cleanup()`. `BASE_ACTIVE`
means "currently in base state" and is a different fact, so it should not be reused for this.

## 3. `pick-idle-gpu.py` needs a shebang

In-repo this is harmless: the file ships `100644`, so `:418-419` falls through to
`python3 "$PICKER"`. But the search list at `:399-401` includes `/usr/local/bin`,
`$HOME/.local/bin` and `/opt/bin`, and the normal way to install it there is `chmod +x` — after
which `/bin/sh` interprets it and emits `Syntax error: "(" unexpected`.

**Change:** one line, `#!/usr/bin/env python3`.

### Patch for 1–3

+23/−3 across the two files, verified here: three consecutive runs in the same worktree with no
cleanup between them, all three reaching `gpu_claim: pass` (`idleness_basis: "vram-only"`),
`arch_coverage: {"gfx942": "runtime"}`, `correctness_repo_tests: pass`, leaving zero residue. Happy
to open it against your branch, or leave it to you.

## 4. Correction — this item was wrong, please disregard it

**Edited after posting.** I originally asked here for `review-pr` to make the validation report
*expected* rather than *required*. That premise was false. At `706eafafc` the skill already treats
the report as optional: Step 1 prints `validation report not supplied: this is a static-only
advisory review`, and the FlyDSL checklist line ends `absence of a report is not itself a blocker`.
The `its own green suite is not evidence` phrasing quoted in the earlier review round is no longer
in the file. I wrote this item from that earlier round's description instead of checking the current
text — my mistake. There is nothing to change here, and items 1–3 above are the only outstanding
ones.

The observation underneath it survives as an observation rather than a request.
`validate_pr.sh:1004-1006` gates `CAN_TEST` on `$PICK` before considering what the target needs, so
a CPU-only target still has to claim a GPU — #5089's test is *"CPU unit tests for FlyDSL helpers (no
GPU required)"* and passes in 1.19 s with `HIP_VISIBLE_DEVICES=""`. Together with
`execution_receipt` requiring a shape grid, whole classes of otherwise correct PR — script-only
targets, and bugfix unit tests with no shape dimension — cannot reach `PASS`. Because the report is
optional that costs nothing today. It would matter the moment anyone proposes making it mandatory.

---

## What held up

Worth recording, because it is what makes the four asks above worth making:

- **Runner selection is correct on unseen code**: `runner: pytest`, `runner_reason: "target defines
  pytest test nodes"`.
- **No false `BLOCK`.** The failure mode from the previous round did not recur.
- **Baseline attribution is correct**: `baseline_control.repo_tests.state: target-not-present` — a
  PR-added test, not a base failure blamed on the author.
- **It goes red for the right reason.** Positive control: I removed the `aiter/ops/flydsl/utils.py`
  fix and kept only the PR's new test — "shipped the test, forgot the implementation". Result
  `BLOCK`/exit `1`, `correctness_repo_tests: fail` with `tests:1 failures:1`, and the log carries
  the real assertion, `AssertionError: assert 65536 == 163840`. It caught the defect, not an
  infrastructure artifact.
- **It does not overclaim**: the clean run stayed `INCONCLUSIVE` with `execution_receipt: skip — no
  shape grid was configured`, and `merge_sim` refused a worktree that was not isolated-clean.

---

## The plan after this PR

This PR merges at the advisory tier. That is the right tier for the evidence, and it is also where
the evidence stops: the description states the historical campaign is local and makes no durable
recall or consistency claim. So the work that makes this tool improvable rather than static is all
still ahead, and I would rather it were filed before merge than promised after it.

I opened #5128 for this and have since closed it as premature, which I think was the right call and
is worth recording here rather than leaving a dangling reference. The measurement work only has a
trigger when someone proposes a rule change, or proposes to let the tool gate a merge; neither has
happened, and a scoring harness built for a tool that is not yet in use would decay before it is
needed. When someone does want to argue that a rule edit is an improvement, that is the moment to
build the corpus, and the person making the claim is its natural owner.

One part of it should survive and does not need an issue: the promotion bar belongs in this skill's
own header, because a header is read every time the tool is used while an issue sinks — the tool
stays advisory until false clearance is measured and near zero for the families that raise a red
verdict, and an LLM judgement never gates a merge. That is the wording change in item 4 above.

The four reports from this run — as-shipped, instrumented, the seeded positive control, and post-fix
— are available as the first holdout records for step 1.

Two loose ends that are not part of that and do not need issues from me. `--extra-target` is already
on the record in this thread with your reason for deferring it, and the shape you described sounds
right to me.
`downstream-impact-check` was correctly split out of this PR and is now unowned — worth either
reviving as its own PR with the incident stated up front, or closing it on the record.

---

## Where I land

Changes 1–3 belong in this PR — without the first, the skill is inert on a common class of developer
host. With those, I approve. On data it had never seen the tool produced
neither a false clearance nor a false blocker, failed safe in every degraded path, and the positive
control shows it genuinely goes red on a real defect. That is what advisory tier should look like.

zufayu (dismissed):
LGTM once the items in my review above are addressed — dismissing for now so it is not read as an unconditional approval. See the change request below for what is still outstanding.

zufayu (changes requested):
Requesting changes to keep this from being merged in its current state. Head is still `706eafafc`,
so none of the outstanding items have landed on this branch.

Two corrections to my review above, both edited into it:

- **Item 4 was wrong.** `review-pr` already treats the validation report as optional at this head
  (`absence of a report is not itself a blocker`), so there is nothing to change there. I wrote it
  from an earlier round's description rather than the current text. Items 1–3 are the only
  outstanding ones.
- Those three are now implemented in `ROCm/aiter#5142` — `read_activity()` making the activity
  metric optional with a recorded `gpu_claim.idleness_basis`, a `PATCH_APPLIED` flag so `cleanup()`
  hands the worktree back clean, and the shebang. That implementation is better than what I
  sketched: it also handles the case where the baseline run already reversed the patch out.

`jhinpan/aiter-fork#2`, which offered the same three fixes against this branch, was closed without
being merged, and `#5142` is now titled as superseding this PR. Which of the two proceeds is for
@jhinpan and @zhiding512 to settle, not something a reviewer should decide — I have no objection to
either path. I am leaving the change request in place only because this branch does not carry the
fixes; it is not a judgement about which PR should land.

zufayu:
would  update to  https://github.com/ROCm/aiter/pull/5142

PR #4877, opened 2026-08-20
[HIP] [GFX950]OPUS MHA Support LSE

shay-li77:
## Motivation

The OPUS gfx950 bf16 forward kernels were inference-only: with no LSE output, `return_lse=True` — and therefore the whole autograd path, which asserts `return_lse` whenever `is_grad` — fell back to `fmha_v3`/CK and lost the OPUS speedup. This adds LSE to both OPUS forward kernels (D=128 and D_QK=192/D_V=128, batch + group/varlen) and opens the dispatch gates so those callers keep it.

## Technical Details

* **LSE is exact under the FAv4 selective max update.** `m_row` is not the true row max (the running max is only refreshed past `RESCALE_THRESHOLD`), but `l_row` stays consistent with whatever base it holds and log-sum-exp is invariant to that base, so `lse = ln2 * (m_row + log2(l_row))`. Fully-masked rows (bottom-right causal, `seqlen_q > seqlen_kv`) get `-inf`, matching `torch.logsumexp`.
* **Gated at runtime on `ptr_lse != nullptr`** rather than a traits flag: a scalar branch, no extra instantiations, and callers that skip LSE skip the allocation and the store. Costs 2 VGPR on D=192 / 1 on D=128, occupancy unchanged.
* **Output** is fp32 natural log — batch `[B, H, N]`, group `[H, total_q]`. In the API `lse` is an out-param like `out`; only `return_lse` changes the return type.
* **Both dispatch gates dropped `not return_lse`** (keeping `not return_softmax`), which also makes the autograd path reachable, i.e. the OPUS forward now feeds the v3/CK backward. Test side: 5 OPUS test functions → 3, with `seqlen_q != seqlen_kv` folded into the shared case lists.

## Test Plan

* `pytest op_tests/{test_mha,test_mha_varlen}.py -k opus`. LSE is compared against an **fp32** logsumexp reference, not `attention_ref`'s — that one is cast back to the input dtype and at `|lse| ~ 8` is coarser than the kernels' own error.
* Coverage: cross-attention both directions incl. fully-masked rows; causal head/tail merge engaged so the mirror Q block scans KV in reverse; KV padding with NaN-filled gaps; single / 17 / decode-shaped groups; autograd grads vs a torch fp32 reference. Every case asserts the output is bitwise-equal to the direct wrapper, so the checks cannot silently validate another backend (negative controls confirm they discriminate).
* Resources via `-Rpass-analysis=kernel-resource-usage` before/after; perf via `run_perftest` on the PR #4205 shapes (`seqlen × batch = 16384`, causal, bf16, nheads ∈ {16, 32}), LSE on vs off plus a rebuilt pre-LSE baseline.

## Test Result

* **Correctness:** 84 passed (60 batch + 24 group). LSE max diff vs fp32 reference: **D=192 ~1e-6**; **D=128 ~4e-3**, which is that kernel's pre-existing bf16 Q pre-scale rather than the LSE — against an arithmetic-matched reference it is also **1.4e-6**. `-inf` row patterns match exactly, no NaN. Grads within bf16 tolerance (causal: out 7.6e-3, dq 1.1e-2, dk 2.1e-2, dv 2.8e-2).
* **Resources:** D=192 248–250 → 250–252 VGPR, D=128 245/246 → 246/247; occupancy 2 waves/SIMD, `VGPRs Spill: 0` throughout.
* **Perf:** LSE on vs off is **0.06%–0.20%** mean across nheads ∈ {16, 32} in both modes (noise-dominated); the no-LSE path is within **±1%** of a rebuilt pre-LSE baseline with no systematic direction.
* **No regressions from opening the gates:** the 320 `test_flash_attn_output[…192-128…]` cases that now reroute to OPUS give an identical 160/160 split with and without `AITER_DISABLE_FMHA_OPUS=1` — those 160 failures are a pre-existing invalid parametrization (`assert nheads % gqa_ratio == 0` → `6 % 8`) that fails in the test's own setup on either backend.

## Submission Checklist

### DIFF
--- a/aiter/ops/mha.py
+++ b/aiter/ops/mha.py
@@ -334,6 +334,7 @@ def gen_fmha_fwd_bf16_opus_fwd_fake(
     out: Tensor,
     causal: bool,
     softmax_scale: float,
+    lse: Tensor | None = None,
     seqstart_q: Tensor | None = None,
     seqstart_k: Tensor | None = None,
     seqstart_q_pad: Tensor | None = None,
@@ -347,7 +348,7 @@ def gen_fmha_fwd_bf16_opus_fwd_fake(
 # OPUS gfx950 bf16 forward (shared entry point): low-level @compile_ops stub bound to
 # the pybind symbol via fc_name. Dispatches by head dim in C++ to the symmetric D=128
 # kernel (batch only) or the asymmetric D_QK=192/D_V=128 kernel (batch + group/varlen).
-# Writes `out` in place, returns None.
+# Writes `out` (and `lse`, when given) in place, returns None.
 @compile_ops(
     "module_fmha_fwd_bf16_opus",
     fc_name="fmha_fwd_bf16_opus_fwd",
@@ -360,6 +361,7 @@ def _fmha_fwd_bf16_opus_fwd(
     out: Tensor,
     causal: bool,
     softmax_scale: float,
+    lse: Tensor | None = None,
     seqstart_q: Tensor | None = None,
     seqstart_k: Tensor | None = None,
     seqstart_q_pad: Tensor | None = None,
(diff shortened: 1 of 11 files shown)


yitingw1:
Before this PR, ATOM prefill MLA with lse=1 will use CK kernel, due to OPUS kernel doesn't support lse=1.
After this PR, ATOM prefill MLA with lse=1 will use OPUS kernel.
Here is some performance comparison:

| MNBT | window | Type | OPUS vs CK (us) | Just attention ratio |
| 3072 | `bs=3 tok=3010` | fresh | 220.2 vs 281.4 = **1.278x** | 1.278x |
| 3072 | `bs=3 tok=2691` | single_pass | 361.7 vs 426.9 = 1.180x | **1.318x** |
| 3072 | `bs=4 tok=3011` | chunked | 562.7 vs 622.9 = 1.107x | **1.194x** |
| 16384 | `bs=16 tok=16335` | fresh | 1252.2 vs 1846.1 = **1.474x** | 1.474x |
| 16384 | `bs=17 tok=16336` | chunked | 3076.9 vs 4191.1 = **1.362x** | **1.585x** |

yitingw1:
Verified accuracy in DeepSeekR1 E2E.

And here is the performance of OPUS attention kernels after fix. 

| MNBT | window | Type | OPUS vs CK (us) | Just attention ratio |
| 3072 | `bs=3 tok=3010` | fresh | 220.4 vs 280.4 = **1.272x** | 1.272x |
| 3072 | `bs=3 tok=2691` | single_pass | 381.5 vs 424.5 = 1.113x | 1.211x |
| 3072 | `bs=4 tok=3011` | **chunked** | 508.4 vs 615.4 = 1.211x | **1.385x** |
| 16384 | `bs=16 tok=16335` | fresh | 1230.9 vs 1854.0 = **1.506x** | 1.506x |
| 16384 | `bs=17 tok=16336` | **chunked** | 2901.0 vs 4185.9 = 1.443x | **1.972x** |

valarLip (approved):

PR #4901, opened 2026-08-21
Skip memory-heavy moe 2stage test cases

gyohuangxin:
## Summary
- add a case-level reference workspace estimate in test_moe_2stage.py
- skip moe_2stage cases when the estimate exceeds AITER_MOE_2STAGE_MAX_FREE_MEMORY_FRACTION of the current free GPU memory, defaulting to 80%
- log the skipped shape, estimated workspace, current free/total GPU memory, and threshold
- catch torch.cuda.OutOfMemoryError per case so an unexpected OOM skips only that case instead of failing the whole shard

## Testing
- python3 -m py_compile op_tests/test_moe_2stage.py
- git diff --check -- op_tests/test_moe_2stage.py
- local estimate sanity: known token=131072, dim=(7168,512), E=385, topk=9 case skips on a 64 GiB free-memory budget and runs on larger free-memory budgets

### DIFF
--- a/op_tests/test_moe_2stage.py
+++ b/op_tests/test_moe_2stage.py
@@ -1008,6 +1008,66 @@ def _runtime_swiglu_mxfp4_q_dtype_a(
     return dtypes.bf16 if get_gfx() != "gfx950" or token < bound else dtypes.fp8
 
 
+def _moe_2stage_reference_workspace_gib(kwargs):
+    token = kwargs["token"]
+    topk = kwargs["topk"]
+    model_dim = kwargs["model_dim"]
+    inter_dim = kwargs["inter_dim"]
+    expert = kwargs["E"]
+    stage1_dim = inter_dim * 2 if kwargs["use_g1u1"] else inter_dim
+
+    # torch_moe_stage1/2 materialize routed fp32 activations and dequantized
+    # fp32 weights. This is a preflight estimate; OOM is still caught per case.
+    routed_activation_bytes = token * topk * (model_dim * 2 + stage1_dim) * 4
+    dequant_weight_bytes = expert * model_dim * (stage1_dim + inter_dim) * 4
+    input_score_bytes = token * (model_dim + expert) * 2
+    return (routed_activation_bytes + dequant_weight_bytes + input_score_bytes) / (
+        1024**3
+    )
+
+
+def _format_moe_2stage_case(kwargs):
+    return (
+        f"token={kwargs['token']}, dim=({kwargs['model_dim']},{kwargs['inter_dim']}), "
+        f"E={kwargs['E']}, topk={kwargs['topk']}, act={kwargs['actType']}, "
+        f"q={kwargs['qType']}, aq={kwargs['AQDType']}, wq={kwargs['WQDType']}, "
+        f"use_g1u1={kwargs['use_g1u1']}, doweight_stage1={kwargs['doweight_stage1']}"
+    )
+
+
+def _moe_2stage_skip_reason(kwargs):
+    # Set AITER_MOE_2STAGE_MAX_FREE_MEMORY_FRACTION=0 to run manually.
+    max_free_memory_fraction = float(
+        os.environ.get("AITER_MOE_2STAGE_MAX_FREE_MEMORY_FRACTION", "0.8")
+    )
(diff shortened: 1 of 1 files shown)


huizzhan:
Reviewed this on a gfx942 MI308X box, including running the branch. The gating works
and it rejects exactly the cases I'd independently flagged, so the direction looks right
to me. Two substantive gaps I'd suggest closing before merge, plus two minor ones.

First, credit where due: putting the check in the main loop with `continue` is the
correct placement. Had it been implemented as an early `return` inside `test_fmoe`, it
would have silently done nothing -- `@benchmark()` always returns `callargs`, so the
loop's `if ret is None: continue` never fires, and skipped cases would still have been
recorded into the bench CSV. Easy trap to fall into; good that it's avoided here. The
log line is also genuinely useful (full case shape plus the threshold), and the
`scanned/recorded/skipped` accounting stays correct.

As a cross-check, I'd separately profiled this same grid with measured peaks and landed
on the same 10 cases (`token=163840`) as the only high-risk tier, which this PR rejects.

### 1. The estimate diverges from real peak by up to 4.8x, and under-estimates small token counts

Measured with `torch.cuda.max_memory_allocated()` over the legacy grid (`-q 0`,
E=257, model_dim=7168, inter_dim=256, topk=9):

| token | this PR's estimate | measured peak | est/measured |
|---|---|---|---|
| 1 – 256 | 5.27 – 5.40 GiB | 15.89 – 15.92 GiB | **0.33 – 0.34** |
| 1024 | 5.79 GiB | 16.01 GiB | 0.36 |
| 4096 | 7.37 GiB | 16.37 GiB | 0.45 |
| 8192 | 9.46 GiB | 16.84 GiB | 0.56 |
| 163840 | 89.10 GiB | 56.45 GiB | 1.58 |

The ratio swings from 0.33x to 1.58x, and the two error directions are opposite:

- `dequant_weight_bytes` counts a *single* fp32 dequantized copy (5.27 GiB here), but the
  measured peak is already 15.89 GiB at `token=1`. The weights get materialised several
  times over in practice (source dtype, quantized, preshuffled) and none of that is in
  the formula.
- `routed_activation_bytes` prices *all* activations at fp32 (`* 4`), while most are
  bf16 in practice, which is what produces the over-estimate at large token counts.

So the "intentionally conservative" note holds for the activation term but is inverted
for the weight term. This isn't currently letting the target cases through -- the
over-estimate happens to carry `token=163840` past 80 GiB -- but it does make the knob's
semantics unreliable: the name `MAX_REFERENCE_GIB` invites operators to reason about it
in device-memory terms and tune it against free VRAM, and it's off by up to 4.8x for that
purpose. Cheapest fix is to state in the comment/docs that this is a reference-workspace
quantity rather than process peak memory; a more accurate one is to multiply the weight
term by the number of live copies.

### 2. No OOM fallback (the one I'd most want added)

Any up-front estimate will miss some cases, and there's a concrete known gap here:
`per_1x32` is gfx950-only (gated at the top of `test_fmoe`), and its reference path
additionally dequantizes fp4 to f32 -- on top of a weight term that already
under-estimates by ~3x. Those two compound.

When the estimate does miss, the whole shard still dies with exit code 1, which is the
exact failure mode this PR is trying to eliminate. Suggest wrapping the `test_fmoe` call:

```python
    except torch.cuda.OutOfMemoryError:
        aiter.logger.warning(
            "moe_2stage: OOM on %s, skipping", _format_moe_2stage_case(kwargs)
        )
```

That way a mis-estimate costs one case instead of the shard.

### 3. The threshold is absolute rather than relative to free VRAM

80 GiB is a hard constant. If a runner already has most of its VRAM held by another
process, the threshold effectively stops gating; on a smaller-VRAM part it may be too
loose. Consider sizing the budget from `torch.cuda.mem_get_info()`'s free value (e.g. a
fraction of it) so it adapts to what's actually available at that moment.

### 4. Env var parsing (agreeing with Copilot, plus one addition)

Beyond the unhelpful `ValueError` message Copilot flagged: `_moe_2stage_skip_reason`
re-reads and re-parses the env var on *every* case, so it runs ~120 times per sweep.
Hoisting it to a module-level constant parsed once (matching the existing
`AITER_MOE_NUM_EXPERT_ACTIVATED` style in this file) addresses both points at once.

PR #4907, opened 2026-08-21
[Triton/Gluon] [FlyDSL] [CI] Add GDN MTP kernels: causal conv1d update and gated delta rule

yiijin:
## Motivation

GDN models such as Qwen3-Next spend every MTP step in two state updates: the short
causal conv1d over the fused qkv projection, and the gated delta rule itself. Both
run on Triton today; this PR ports them to FlyDSL for gfx942/gfx950.

## Technical Details

`causal_conv1d_update`:

* New FlyDSL `causal_conv1d_update`: vLLM and SGLang interfaces, decode and MTP verify.
* Covers EAGLE tree taps, conv-window snapshot, varlen packing, prefix caching, null block.
* Opt-in via `AITER_CONV1D_UPDATE_FLYDSL=1`; the default Triton path is unchanged.
* Widths 2-6 (vLLM) / 2-4 (SGLang), bf16 and fp16, gfx942/gfx950; anything else is refused.
* One dtype specializes the whole kernel, so `x`, `weight` and `bias` are refused when
  they disagree with the conv state rather than reinterpreted. Casting `x` here instead
  would leave the update in the copy, not in the buffer upstream overwrites in place.
* Cache-line terms sit in 64-bit descriptor bases, so caches past 4 GiB address correctly.
* Guards the Triton prefill width too, which silently miscomputed above 4.
* Index buffers read by FlyDSL are validated for dtype, device, rank, and required contiguity. 
* Unsupported dispatch inputs fall back to Triton, while direct FlyDSL calls reject them explicitly.

Gated delta rule:

* New FlyDSL gated delta rule: `flydsl_gdr_decode`, plus `flydsl_gdr_mtp` (vLLM) and
  `flydsl_gdr_mtp_sglang` (SGLang) for the verify step.
* All three MTP contracts: vLLM's chain, SGLang's chain with per-token snapshots, and
  SGLang's EAGLE tree, which reloads each arm's parent state.
* Opt-in via `AITER_GDR_FLYDSL=1`; the default Triton path is unchanged.
* Sigmoid gating, the q/k L2 norm and the recurrence are one launch, with the state
  resident in registers for the whole token loop.
* Per-token checkpointing, so a rejected draft token rolls the state back in place
  rather than by re-running the step.
* The snapshot store is non-temporal wherever nothing reads it back: every contract but
  the tree, and the tree's own last step. It is write-once traffic either way, and at the
  footprints where the state pool still fits in LLC it was evicting the pool that the
  next token reads.
* On the tree, token 1's parent state stays in registers at either snapshot dtype. A
  narrower snapshot owes a rounding, but a store followed by its reload is a `truncf`
  then an `extf`, so the round trip is exact without going to memory.
* bf16/fp16 q/k/v with an fp32 or bf16 state pool, GQA, validated at 128x128 head dims,
  gfx942/gfx950; anything else is refused.
* Pool and snapshot terms sit in 64-bit descriptor bases, for the reason the conv does.
* Tiling is derived on the host from the launch grid: decode splits the value dimension
  until `batch_size * num_v_heads` covers the part, and the MTP ladder does the same
  before branching on the contract and the draft length, since the three contracts want
  different splits at the same shape. `WAVES_PER_EU` rides with the one tiling that wants
  it. `gdr_decode_tuned.csv` and the lookup that read it are deleted here: a row was keyed
  on the shape and the batch, and the batch is half of the grid the rule already reads.
* The public MTP dispatch accepts `draft_window` as host-side metadata, avoiding the 
   GPU-to-CPU synchronization previously needed to infer it from `cu_seqlens`. Calls 
   without this metadata fall back to Triton.
* The default stream is selected from the input device, explicit streams are checked 
   against it, and non-contiguous output buffers are rejected.
* CUDA Graph capture currently skips the FlyDSL route and falls back to Triton.
* A validated private launch path avoids repeating wrapper validation in the public dispatch.

Both ports vendor the upstream kernels as test oracles, so the suite needs no
vllm/sglang at runtime.

## Test Plan

python3 op_tests/test_flydsl_causal_conv1d_update.py
python3 op_tests/test_flydsl_gdr_mtp.py

## Test Result

Both ports numerically agree with the vendored vLLM (`63a9a5010`) and SGLang
(`18107e38d2`) kernels within the configured tolerances. Exact-equality checks are
limited to copied state, untouched storage, rollback state, and routed-versus-direct
execution of the same implementation. The current suite collects 41 conv cases and
19 gated-delta-rule MTP cases.

### causal_conv1d_update

bf16, width 4. Speedup over each upstream base kernel, median of 3 runs over 12 shapes
per row (batch 1/32/128 x dim 1024/2048/4096/8192), doubled on the vLLM rows because
they cover both `contiguous` and `qkvz_slice` layouts against SGLang's one `verify_view`.

#### gfx950

| interface | mode | seqlen | base kernel | median |
| --------- | ----------------- | ------ | ------------- | ------ |
| vLLM | decode | 1 | vLLM Triton | 1.49x |
| vLLM | MTP verify | 2 | vLLM Triton | 1.51x |
| vLLM | MTP verify | 4 | vLLM Triton | 1.62x |
| SGLang | MTP verify, chain | 2 | SGLang Triton | 1.51x |
| SGLang | MTP verify, chain | 4 | SGLang Triton | 1.36x |
| SGLang | MTP verify, tree | 2 | SGLang Triton | 1.51x |
| SGLang | MTP verify, tree | 4 | SGLang Triton | 1.44x |

Latency in us at dim 8192, batch 1 / 32 / 128:

| mode | FlyDSL | upstream base | aiter Triton |
| ---------------------- | ------------------- | -------------------- | ------------------- |
| vLLM decode, seqlen 1 | 2.52 / 3.64 / 7.90 | 3.77 / 5.41 / 13.84 | 3.64 / 5.44 / 13.93 |
| vLLM verify, seqlen 4 | 3.37 / 6.34 / 15.42 | 5.82 / 10.32 / 24.15 | n/a |
| SGLang chain, seqlen 4 | 2.92 / 6.77 / 17.26 | 4.41 / 8.30 / 21.89 | 4.44 / 8.44 / 22.13 |
| SGLang tree, seqlen 4 | 3.54 / 7.78 / 19.17 | 5.79 / 10.05 / 26.78 | n/a |

Aggregating every measured shape per baseline:

| base kernel | shapes compared | median |
| ------------- | ------------------------------- | ------ |
| vLLM Triton | 72 (decode + verify) | 1.53x |
| SGLang Triton | 48 (chain + tree verify) | 1.48x |
| aiter Triton | 48 (vLLM decode + SGLang chain) | 1.46x |

#### gfx942

| interface | mode | seqlen | base kernel | median |
| --------- | ----------------- | ------ | ------------- | ------ |
| vLLM | decode | 1 | vLLM Triton | 1.55x |
| vLLM | MTP verify | 2 | vLLM Triton | 1.47x |
| vLLM | MTP verify | 4 | vLLM Triton | 1.55x |
| SGLang | MTP verify, chain | 2 | SGLang Triton | 1.50x |
| SGLang | MTP verify, chain | 4 | SGLang Triton | 1.36x |
| SGLang | MTP verify, tree | 2 | SGLang Triton | 1.52x |
| SGLang | MTP verify, tree | 4 | SGLang Triton | 1.42x |

Latency in us at dim 8192, batch 1 / 32 / 128:

| mode | FlyDSL | upstream base | aiter Triton |
| ---------------------- | ------------------- | -------------------- | ------------------- |
| vLLM decode, seqlen 1 | 2.85 / 10.43 / 29.44 | 4.15 / 16.50 / 52.54 | 4.14 / 16.58 / 52.09 |
| vLLM verify, seqlen 4 | 4.13 / 19.56 / 50.71 | 6.77 / 29.27 / 84.02 | n/a |
| SGLang chain, seqlen 4 | 3.69 / 16.40 / 47.02 | 5.17 / 21.27 / 70.09 | 5.57 / 21.88 / 71.37 |
| SGLang tree, seqlen 4 | 4.55 / 19.94 / 57.92 | 7.25 / 25.93 / 83.27 | n/a |

Aggregating every measured shape per baseline:

| base kernel | shapes compared | median |
| ------------- | ------------------------------- | ------ |
| vLLM Triton | 72 (decode + verify) | 1.50x |
| SGLang Triton | 48 (chain + tree verify) | 1.47x |
| aiter Triton | 48 (vLLM decode + SGLang chain) | 1.51x |

### Gated delta rule, MTP verify

Full 144-cell sweep (3 contracts x 4 TP settings x 2 state dtypes x 2 draft lengths x
3 batch sizes), averaged over two independent runs.

#### gfx950

The two runs returned a geomean of 1.451x and 1.431x. All 144 cells are at or above
parity, the lowest at 1.034x.

| contract | base kernel | geomean | median |
| ------------- | ---------------------------------------------- | ------- | ------ |
| vllm_chain | vLLM fused_recurrent_gated_delta_rule | 1.522x | 1.466x |
| sglang_tree | SGLang, tree/EAGLE arms | 1.446x | 1.319x |
| sglang_chain | SGLang fused_sigmoid_gating_delta_rule_update | 1.363x | 1.309x |
| all 144 cells | | 1.442x | 1.360x |

On `vllm_chain` FlyDSL is also 1.168x over aiter's own Triton kernel, over those 48 cells.

Geomean speedup by shape parameter:

| parameter | | | | |
| ----------- | ----------- | ----------- | ----------- | ---------- |
| batch | b1 1.905x | b32 1.341x | b128 1.175x | |
| TP | tp1 1.354x | tp2 1.414x | tp4 1.464x | tp8 1.545x |
| state dtype | fp32 1.551x | bf16 1.341x | | |

The gain is largest where a verify step actually runs -- small batch, high TP -- since
both leave less work per launch.

Latency in us at tp1 (32 v-heads), fp32 state, seqlen 4, batch 1 / 32 / 128:

| contract | FlyDSL | upstream base | aiter Triton |
| ------------- | --------------------- | ---------------------- | ---------------------- |
| vllm_chain | 8.13 / 55.13 / 262.09 | 19.36 / 65.50 / 281.66 | 11.22 / 55.46 / 260.21 |
| sglang_chain | 8.09 / 63.33 / 253.22 | 16.61 / 73.39 / 312.18 | n/a |
| sglang_tree | 7.85 / 80.91 / 325.04 | 17.13 / 98.52 / 359.02 | n/a |

Per-cell speedups agree between the two runs to 0.91% at the median. Two cells move by
more than half, both `vllm_chain` at 16 v-heads with a bf16 state, and the movement is
upstream's: FlyDSL holds to 8% across the two runs there while the vLLM baseline swings
by 70%. Read the aggregates rather than individual cells.

#### gfx942

The two runs returned a geomean of 1.374x and 1.370x.

| contract | base kernel | geomean | median |
| ------------- | ---------------------------------------------- | ------- | ------ |
| vllm_chain | vLLM fused_recurrent_gated_delta_rule | 1.414x | 1.200x |
| sglang_tree | SGLang, tree/EAGLE arms | 1.428x | 1.262x |
| sglang_chain | SGLang fused_sigmoid_gating_delta_rule_update | 1.278x | 1.128x |
| all 144 cells | | 1.371x | 1.211x |

On `vllm_chain` FlyDSL is also 1.233x over aiter's own Triton kernel, over those 48 cells.

Geomean speedup by shape parameter:

| parameter | | | | |
| ----------- | ----------- | ----------- | ----------- | ---------- |
| batch | b1 2.046x | b32 1.142x | b128 1.103x | |
| TP | tp1 1.215x | tp2 1.347x | tp4 1.473x | tp8 1.466x |
| state dtype | fp32 1.463x | bf16 1.285x | | |

The gain is largest where a verify step actually runs -- small batch, high TP -- since
both leave less work per launch.

Latency in us at tp1 (32 v-heads), fp32 state, seqlen 4, batch 1 / 32 / 128:

| contract | FlyDSL | upstream base | aiter Triton |
| ------------- | --------------------- | ---------------------- | ---------------------- |
| vllm_chain | 16.21 / 158.11 / 684.92 | 30.96 / 182.22 / 734.09 | 18.76 / 193.92 / 741.64 |
| sglang_chain | 17.52 / 181.90 / 710.60 | 30.06 / 190.33 / 712.32 | n/a |
| sglang_tree | 22.65 / 194.38 / 811.42 | 34.44 / 220.44 / 862.08 | n/a |

Per-cell speedups agree between the two runs to 0.83% at the median. Read the
aggregates rather than individual cells.

`aiter Triton` is n/a where it cannot express the call: it implements the chain contract
only, and neither upstream implements the other's.

### Tiling: the rule against the table it replaces, on gfx950

All ratios are **table time / rule time, so above 1.00 means the rule is faster**. A cell
is one timed launch configuration -- one shape at one batch size -- which is also the unit
the rule and a table row each decide a tiling for.

#### Decode — 112 cells

Every one of the 32 to 44 tilings a cell admits was timed in one pass against the same
tensors, so both arms' choices are read off a single measurement:

| | rule vs table | distance from the cell's best tiling |
| --------- | ------- | ------ |
| the rule | **1.0019** | **0.73%** |
| the table | - | 0.92% |

Best cell 1.0363x. The rule is the one closer to optimal: it derives the split from the
grid the launch actually has, where a row was fitted per batch on a surface that moves
with `batch * num_v_heads`. Timing the two arms separately instead lands inside a floor
spanning 0.9483 to 1.0055 over the 4 cells where both arms compile an identical kernel,
so a difference this size is only resolvable in the sweep above.

#### MTP — 123 cells

| | rule vs table | vs each contract's upstream |
| --------- | ------- | ------ |
| the rule | **1.0078** | **1.363x** |
| the table | - | 1.353x |

| contract | rule vs table | rule vs upstream |
| ------------- | ------- | ------ |
| `sglang_chain` | 1.0092 | 1.334x |
| `sglang_tree` | 1.0065 | 1.316x |
| `vllm_chain` | 1.0081 | 1.444x |

Best cell 1.5885x. The rule is ahead of the table on every contract and further ahead of
upstream than the table is. Scored against the best tiling per cell, the four gfx950
ladder changes take the rung from 3.94% off optimal to **1.76%**.

Decode's constants carry to gfx950 unchanged: refitting them on this part scores 0.75%
off optimal under five-fold cross-validation, behind the 0.73% they already reach, so the
refit is fitting noise.

### Public-entry MTP latency

The public-entry benchmark measures the complete
`fused_rearrange_sigmoid_gated_delta_rule()` call using a warmed-up per-call
wall-clock median.

On gfx950 with `B=1`, `T=4`, `HV=32`:

* Direct FlyDSL: 36.034 us
* Public entry to FlyDSL: 49.995 us
* Public entry to Triton: 53.861 us
* Vendored vLLM reference: 87.211 us

The public FlyDSL route is 3.87 us, or approximately 7.2%, faster than the same
public entry with FlyDSL disabled.

The absolute values differ from the reviewer's measurements because the benchmark
harness and software environment are not identical. The relative synchronization
gap is consistent: the reviewer observed a 33.6 us regression, while reproducing the
previous synchronized dispatch in this environment showed a 32.71 us regression.
After passing the draft window through host metadata and removing duplicate
validation, the gap becomes a 3.87 us improvement.

The direct and public FlyDSL results are reported separately because their layouts
differ. The direct benchmark uses separate contiguous Q/K/V tensors, while the public
entry receives packed 2-D QKV and constructs strided Q/K/V views. The public-entry
comparison therefore represents the production dispatch benefit.

## Submission Checklist

* Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests


aiter/ops/flydsl/causal_conv1d_update_kernels.py
```diff
@@ -0,0 +1,827 @@
...
+
+def _n_cu(device: torch.device) -> int | None:
+    """Compute-unit count of ``device``, or ``None`` if it could not be queried."""
+    key = device.index if device.index is not None else 0
+    if key not in _CU_COUNT_CACHE:
+        try:
+            count = torch.cuda.get_device_properties(device).multi_processor_count
+        except Exception:  # noqa: BLE001 - no live device, or a meta/CPU tensor
+            count = None
+        _CU_COUNT_CACHE[key] = count
```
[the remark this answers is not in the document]
yiijin: Fixed in 9b4901539 — _n_cu() now goes through get_num_sms(), since under CU_NUM=64 it was reading 256 while the gated delta rule path in this same PR already read 64.

aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py
```diff
@@ -15,6 +18,171 @@
...
+    take a packed batch that unflattens to ``[N, T, ...]``. A ragged batch is a
+    different kernel per row, so it stays on Triton.
+    """
+    if cu_seqlens is None:
+        return -1
+    # One device-to-host copy; the window has to be known on the host to pick
+    # the kernel, so the comparisons are done here rather than one .item() each.
+    lens = (cu_seqlens[1:] - cu_seqlens[:-1]).to("cpu")
+    if lens.numel() == 0:
+        return -1
```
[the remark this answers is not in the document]
yiijin: Measured it and the suggestion is slower — the cost is the number of syncs rather than the bytes, so one bulk copy beats three .item() calls at every batch size (26 µs vs 67 µs at B=128, and still 41 µs vs 67 µs at B=8192 where the copy is 64 KB).

op_tests/triton_tests/utils/gdr_mtp_refs.py
```diff
@@ -0,0 +1,1061 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
yiijin: This file is adapted from vLLM, which is Apache-2.0, so tagging it MIT would misstate the license — kda_ref.py is MIT because its upstream (flash-linear-attention) is.

op_tests/triton_tests/utils/causal_conv1d_update_refs.py
```diff
@@ -0,0 +1,1246 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
yiijin: This is adapted from vLLM and SGLang, both Apache-2.0, so the tag differs from kda_ref.py because the upstream does.

op_tests/test_flydsl_gdr_mtp.py
```diff
@@ -0,0 +1,1782 @@
+# SPDX-License-Identifier: MIT
```
coderfeli, line 1: clean tests using unit test skill

aiter/ops/flydsl/kernels/gdr_decode.py
```diff
@@ -36,6 +41,22 @@ def _store_vec(atom, tile, value, width, numeric):
     fx.copy(atom, frag, tile)
 
 
+def _fast_exp(x):
+    return rocdl.exp2(T.f32, _to_raw(fx.Float32(x) * _LOG2E))
```
coderfeli, line 57: clean to to_raw?

aiter/ops/triton/conv/causal_conv1d.py
```diff
@@ -11,6 +14,48 @@
...
+def _flydsl_conv1d_update_available() -> bool:
+    # Probes the module the fast path imports below: `aiter.ops.flydsl` raises
+    # on a flydsl that is missing or too old, so a bad install is a False here
+    # rather than an exception out of a Triton entry point.
+    try:
+        from aiter.ops.flydsl import causal_conv1d_update_kernels  # noqa: F401
+
+        return True
+    except (ImportError, OSError, RuntimeError):
+        return False
```
coderfeli: why need this?

aiter/ops/triton/conv/causal_conv1d.py
```diff
@@ -11,6 +14,48 @@
...
+def _flydsl_conv1d_update_enabled() -> bool:
+    """Opt-in gate for the FlyDSL causal_conv1d update port.
+
+    Off by default, so Triton keeps serving every call. Not memoized, so the env
+    var can be toggled at runtime; only the availability probe is cached.
+    """
+    return (
+        os.environ.get("AITER_CONV1D_UPDATE_FLYDSL", "") == "1"
+        and _flydsl_conv1d_update_available()
+    )
```
coderfeli: why need this?

aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py
```diff
@@ -15,6 +18,171 @@
...
+        return False
+
+
+def _flydsl_gdr_enabled() -> bool:
+    """Opt-in gate for the FlyDSL gated-delta-rule MTP port.
+
+    Off by default, so Triton keeps serving every call. Not memoized, so the env
+    var can be toggled at runtime; only the availability probe is cached.
+    """
+    return os.environ.get("AITER_GDR_FLYDSL", "") == "1" and _flydsl_gdr_available()
```
coderfeli: clean

aiter/ops/flydsl/causal_conv1d_update_kernels.py
```diff
@@ -0,0 +1,838 @@
...
+    and the two ports cannot end up sizing their launches for different
+    machines.
+    """
+    # Keyed on the type as well, or a device that cannot be queried at all takes
+    # the answer cached for ``cuda:0``: both spell their index ``None``.
+    key = (device.type, device.index if device.index is not None else 0)
+    if key not in _CU_COUNT_CACHE:
+        try:
+            with torch.cuda.device(device):
+                count = get_num_sms()
```
coderfeli, line 142: get from csv?

aiter/ops/flydsl/kernels/gdr_decode.py
```diff
@@ -36,6 +41,22 @@ def _store_vec(atom, tile, value, width, numeric):
...
+
+
+def _fast_log1p(x):
+    return fx.math.log1p(x, fastmath=fx.FastMathFlags.fast)
```
coderfeli: use fast math compile ops for all instead of setting one by one?

aiter/ops/flydsl/kernels/causal_conv1d_update.py
```diff
@@ -0,0 +1,1134 @@
...
+            return buffer_ops.create_buffer_resource_from_addr(ptr)
+
+        # A term scaling with the batch or the cache size goes in the 64-bit base;
+        # the buffer offset is 32 bits. Both factors widen separately, their
+        # product being what overflows.
+        def _rsrc_at(ptr, index, stride):
+            byte_offset = (
+                index.to(fx.Int64) * stride.to(fx.Int64) * fx.Int64(ELEM_BYTES)
+            )
+            return buffer_ops.create_buffer_resource_from_addr(ptr + byte_offset)
```
coderfeli: using make buffer tensor and fx.copy/tilecopy？

op_tests/test_flydsl_gdr_mtp.py
yiijin: Both files now call `pytest.main([__file__], ...)` instead of the hand-written `cases` table, which was a second dispatch duplicating the `parametrize` sites already there.

aiter/ops/flydsl/kernels/gdr_decode.py
yiijin: Dropped `_fast_log1p`, `_fast_rcp` and `_fast_rsqrt` (now `fx.math.log1p`, plain division, `fx.math.rsqrt`), since the compile hint leaves them bare aliases costing 13 MTP instructions for no measurable time; `_fast_exp` stays on the intrinsic because `fx.math.exp` guards each call site against denormals for +22 instructions on decode and +65 on MTP.

aiter/ops/triton/conv/causal_conv1d.py
yiijin: Removed.

aiter/ops/triton/conv/causal_conv1d.py
yiijin: Now a single uncached env read, so it reads the same at every call.

aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py
yiijin: Done.

aiter/ops/flydsl/causal_conv1d_update_kernels.py
yiijin: `_n_cu()` goes through `get_num_sms()` like the rest of the repo, but stays a runtime query because conv1d derives its tiling from an analytic occupancy criterion rather than a tuned CSV.

aiter/ops/flydsl/kernels/gdr_decode.py
yiijin: Done.

aiter/ops/flydsl/kernels/causal_conv1d_update.py
yiijin: Migrated the value paths to `make_buffer_tensor` + `fx.copy`, taking buffer_ops from 28 sites to 10; the 10 left are uniform i32 index loads that no `BufferCopy*` atom can express, and migrating them anyway measured +13% mean latency.

aiter/ops/triton/conv/causal_conv1d.py
```diff
@@ -11,6 +13,32 @@
 _LOGGER = AiterTritonLogger()
 
 
+def _flydsl_conv1d_update_enabled() -> bool:
+    """Opt-in gate for the FlyDSL causal_conv1d update port.
+
+    Off by default, so Triton keeps serving every call. Not memoized, so the env
+    var can be toggled at runtime.
+    """
+    return os.environ.get("AITER_CONV1D_UPDATE_FLYDSL", "") == "1"
```
coderfeli, line 22: why still need this env?

yiijin: The env var is how a deployment picks the FlyDSL or the Triton path per run without a rebuild.

xytpai:
### 1. `aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py:40`

**From agent — [P1] Avoid device-to-host reads during graph capture**

Enabling `AITER_GDR_FLYDSL=1` makes an otherwise graph-capturable MTP call fail here when `cu_seqlens` is on the GPU. This synchronous `.to("cpu")` runs during dispatch, before the code can fall back to Triton.

Reproduced on gfx950 with `cu_seqlens=[0,4,8]`: eager execution succeeds, but `torch.cuda.graph()` raises `Cannot copy between CPU and CUDA tensors during CUDA graph capture`.

Please obtain the draft window from host-side metadata, or decline the FlyDSL route before this copy when capturing. A pinned destination alone would not resolve the subsequent host-side dependency on the copied values.

---

### 2. `aiter/ops/flydsl/kernels/gdr_decode.py:777–779`

**From agent — [P1] Preserve aiter’s valid slot-zero semantics**

The new aiter dispatch route uses `CHAIN`, but aiter’s existing Triton kernel treats slot `0` as valid: it skips only negative read indices and writes checkpoints to any nonnegative index. `MIN_LIVE_SLOT=1` therefore changes the public entry point’s behavior when the flag is enabled.

With `ssm_state_indices=[[0,1]]` and `num_accepted_tokens=[1]`, I reproduced normal output/state updates with the flag off, but an entirely skipped sequence with it on: the output buffer retained its prefilled `-99` values and the state pool was not updated.

Please parameterize the null-slot convention so the aiter route allows slot `0` while the vLLM interface retains its reserved-zero behavior. Regression tests should cover slot `0` as both a rollback source and a checkpoint destination.

---

### 3. `aiter/ops/triton/conv/causal_conv1d.py:280–299`

**From agent — [P2] Check the snapshot dtype before dispatching to FlyDSL**

The support predicate does not receive `intermediate_conv_window`, so a bf16 convolution with an fp32 snapshot buffer is routed to FlyDSL. Triton converts values to the snapshot pointer’s element type, whereas the new kernel uses the input dtype and two-byte addressing for snapshot stores. This silently corrupts the snapshot.

Reproduced with `B=2, D=128, T=2, W=4`: Triton updates all 1536 fp32 snapshot elements correctly, while FlyDSL produces corrupted values and leaves 768 elements untouched.

Please screen the snapshot’s dtype/device before dispatch and reject unsupported combinations in the direct FlyDSL wrapper, or specialize snapshot stores separately.

yiijin:
> ### 1. `aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py:40`
> **From agent — [P1] Avoid device-to-host reads during graph capture**
> 
> Enabling `AITER_GDR_FLYDSL=1` makes an otherwise graph-capturable MTP call fail here when `cu_seqlens` is on the GPU. This synchronous `.to("cpu")` runs during dispatch, before the code can fall back to Triton.
> 
> Reproduced on gfx950 with `cu_seqlens=[0,4,8]`: eager execution succeeds, but `torch.cuda.graph()` raises `Cannot copy between CPU and CUDA tensors during CUDA graph capture`.
> 
> Please obtain the draft window from host-side metadata, or decline the FlyDSL route before this copy when capturing. A pinned destination alone would not resolve the subsequent host-side dependency on the copied values.
> 
> ### 2. `aiter/ops/flydsl/kernels/gdr_decode.py:777–779`
> **From agent — [P1] Preserve aiter’s valid slot-zero semantics**
> 
> The new aiter dispatch route uses `CHAIN`, but aiter’s existing Triton kernel treats slot `0` as valid: it skips only negative read indices and writes checkpoints to any nonnegative index. `MIN_LIVE_SLOT=1` therefore changes the public entry point’s behavior when the flag is enabled.
> 
> With `ssm_state_indices=[[0,1]]` and `num_accepted_tokens=[1]`, I reproduced normal output/state updates with the flag off, but an entirely skipped sequence with it on: the output buffer retained its prefilled `-99` values and the state pool was not updated.
> 
> Please parameterize the null-slot convention so the aiter route allows slot `0` while the vLLM interface retains its reserved-zero behavior. Regression tests should cover slot `0` as both a rollback source and a checkpoint destination.
> 
> ### 3. `aiter/ops/triton/conv/causal_conv1d.py:280–299`
> **From agent — [P2] Check the snapshot dtype before dispatching to FlyDSL**
> 
> The support predicate does not receive `intermediate_conv_window`, so a bf16 convolution with an fp32 snapshot buffer is routed to FlyDSL. Triton converts values to the snapshot pointer’s element type, whereas the new kernel uses the input dtype and two-byte addressing for snapshot stores. This silently corrupts the snapshot.
> 
> Reproduced with `B=2, D=128, T=2, W=4`: Triton updates all 1536 fp32 snapshot elements correctly, while FlyDSL produces corrupted values and leaves 768 elements untouched.
> 
> Please screen the snapshot’s dtype/device before dispatch and reject unsupported combinations in the direct FlyDSL wrapper, or specialize snapshot stores separately.



Thanks for the detailed review and reproductions. I’ve addressed all three issues:

1. Fixed by falling back to Triton during CUDA graph capture before any device-to-host read.
2. Fixed by parameterizing the minimum live slot so aiter accepts slot 0 while vLLM continues reserving it.
3. Fixed by validating the snapshot dtype and device in both the dispatch predicate and direct FlyDSL entrypoint.

coderfeli:
LGTM. look at yutao's feedback.

op_tests/test_flydsl_causal_conv1d_update.py
```diff
@@ -0,0 +1,2093 @@
...
+            causal_conv1d_update_sglang_flydsl,
+        )
+        from aiter.ops.triton.conv.causal_conv1d import (
+            causal_conv1d_fn as causal_conv1d_fn_triton,
+        )
+        from aiter.ops.triton.conv.causal_conv1d import (
+            causal_conv1d_update as causal_conv1d_update_triton,
+        )
+    except ImportError as exc:
+        _SKIP_REASON = f"the FlyDSL causal_conv1d update kernels do not import ({exc})"
```
coderfeli: clean ut. too big.

yiijin: Done.

xytpai:
### [P1] Validate index dtypes before routing conv1d updates to FlyDSL

**Location:** [`aiter/ops/triton/conv/causal_conv1d.py:280–304`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/triton/conv/causal_conv1d.py#L280-L304)

The support predicate does not check `conv_state_indices`, but the FlyDSL kernel always reads its contents through `_load_i32()`. The existing Triton path reads indices according to the tensor’s actual dtype, so enabling `AITER_CONV1D_UPDATE_FLYDSL` silently changes the behavior for valid int64 indices.

I reproduced this on gfx950 with `batch=2`, `dim=64`, `width=4`, and:

```python
conv_state_indices = torch.tensor([1, 2], dtype=torch.int64, device="cuda")
```

Switching the environment variable from `0` to `1` produced a maximum output difference of **8.25** and modified **cache slot 0**, which should remain untouched. Using int32 indices produced identical outputs and states.

Please validate index dtype/device/layout before dispatch and fall back to Triton when unsupported. The public FlyDSL wrappers should also reject unsupported index types explicitly. Apply the same checks to `num_accepted_tokens` and other index buffers read as int32.

---

### [P1] Select the default MTP stream from the input device

**Location:** [`aiter/ops/flydsl/linear_attention_kernels.py:860–862`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/flydsl/linear_attention_kernels.py#L860-L862)  
**Also applies to:** `flydsl_gdr_mtp_sglang`, lines 926–928.

`torch.cuda.current_stream()` selects the stream for the current device, which need not be `query.device`. Entering the target device context later in `_mtp_launch()` does not correct the already-selected stream.

I reproduced this with all operands on `cuda:1` while a non-default stream on `cuda:0` was current. The wrapper passed the `cuda:0` stream to the launch, which failed with `hipErrorInvalidHandle`.

Please select the default stream explicitly:

```python
if stream is None:
    stream = torch.cuda.current_stream(query.device)
```

Also validate the device of explicitly supplied streams and add a multi-GPU regression test where the current device differs from the input device.

---

### [P2] Honor or reject non-contiguous MTP output buffers

**Location:** [`aiter/ops/flydsl/kernels/gdr_decode.py:856–866`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/flydsl/kernels/gdr_decode.py#L856-L866)

`out_view` uses hard-coded contiguous strides, but `_mtp_common_checks()` checks only the output’s dtype and device—not its shape or contiguity.

A correctly shaped output view such as:

```python
backing = torch.full(
    (2, 2, 4, 256), 123.0, device="cuda", dtype=torch.bfloat16
)
out = backing[..., ::2]  # Shape: (2, 2, 4, 128)
```

is accepted, but the kernel writes as though `out` were contiguous. In my gfx950 reproduction, the result differed from the contiguous-output baseline, and **1,024 elements of `backing[..., 1::2]` were modified**, even though those elements are outside the supplied output view.

Please pass the actual output strides to the kernel, or explicitly validate the expected shape and require contiguous output before launching. Creating a temporary contiguous copy alone would not preserve the caller-owned output-buffer contract.

---

### [P2] Forward `accepted` into the rollback test fixture

**Location:** [`op_tests/test_flydsl_gdr_mtp.py:550–556`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/op_tests/test_flydsl_gdr_mtp.py#L550-L556)

`_check_chain_case()` receives `accepted`, but does not forward it to `_make_problem()`. Consequently, the parameterized cases labeled `"mixed"` and `"first"` use the fixture’s default `"full"` acceptance pattern.

Please add:

```python
accepted=accepted,
```

to the fixture call so these cases exercise their intended rollback positions. I reran the four chain cases with this correction, and they passed; this is a coverage gap rather than an observed recurrence failure.

---

### Measure public-entry-point latency, including the dispatch synchronization

**Location:** [`aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py:29–48`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py#L29-L48)

`_uniform_draft_window()` performs a GPU-to-CPU copy and synchronization on every eligible dispatch. This overhead can outweigh the kernel-level improvement for small batches.

In a warmed-up gfx950 wall-clock microbenchmark of the public entry point with `B=1`, `T=4`, and `HV=32`, I measured:

| Configuration | Median time per call |
|---|---:|
| FlyDSL disabled | 35.9 µs |
| FlyDSL enabled | 69.5 µs |

This is one local microbenchmark, not a model-level performance result, but it demonstrates why kernel-only speedups should be reported separately from public-API latency.

Please add a benchmark that includes dispatch overhead and consider accepting the draft window through host-side metadata to avoid synchronization in the hot path. Also document that the current dispatch falls back to Triton during CUDA Graph capture.

---

### Align the “bit-exact” claim with the current validation

The PR description claims bit-exact agreement with the upstream kernels, but the current tests generally validate computed outputs and recurrent states against an error budget. Exact-equality assertions mainly cover copied state, untouched regions, or routed-versus-direct calls to the same implementation.

In a same-dtype gfx950 comparison against the vendored vLLM reference, I observed nonzero differences in both output and updated state, with a maximum output difference of approximately `4.88e-4`.

These differences do not by themselves establish a numerical correctness issue, but they do not support an unconditional bit-exact claim. Please distinguish numerical agreement within tolerance from exact equality, and update the reported test counts to match the current suite.

---

**Review result: `REQUEST_CHANGES`**

Please address the index interpretation, stream-device selection, and output-layout issues and add regression coverage before merging.

yiijin:
> ### [P1] Validate index dtypes before routing conv1d updates to FlyDSL
> **Location:** [`aiter/ops/triton/conv/causal_conv1d.py:280–304`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/triton/conv/causal_conv1d.py#L280-L304)
> 
> The support predicate does not check `conv_state_indices`, but the FlyDSL kernel always reads its contents through `_load_i32()`. The existing Triton path reads indices according to the tensor’s actual dtype, so enabling `AITER_CONV1D_UPDATE_FLYDSL` silently changes the behavior for valid int64 indices.
> 
> I reproduced this on gfx950 with `batch=2`, `dim=64`, `width=4`, and:
> 
> ```python
> conv_state_indices = torch.tensor([1, 2], dtype=torch.int64, device="cuda")
> ```
> 
> Switching the environment variable from `0` to `1` produced a maximum output difference of **8.25** and modified **cache slot 0**, which should remain untouched. Using int32 indices produced identical outputs and states.
> 
> Please validate index dtype/device/layout before dispatch and fall back to Triton when unsupported. The public FlyDSL wrappers should also reject unsupported index types explicitly. Apply the same checks to `num_accepted_tokens` and other index buffers read as int32.
> 
> ### [P1] Select the default MTP stream from the input device
> **Location:** [`aiter/ops/flydsl/linear_attention_kernels.py:860–862`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/flydsl/linear_attention_kernels.py#L860-L862) **Also applies to:** `flydsl_gdr_mtp_sglang`, lines 926–928.
> 
> `torch.cuda.current_stream()` selects the stream for the current device, which need not be `query.device`. Entering the target device context later in `_mtp_launch()` does not correct the already-selected stream.
> 
> I reproduced this with all operands on `cuda:1` while a non-default stream on `cuda:0` was current. The wrapper passed the `cuda:0` stream to the launch, which failed with `hipErrorInvalidHandle`.
> 
> Please select the default stream explicitly:
> 
> ```python
> if stream is None:
>     stream = torch.cuda.current_stream(query.device)
> ```
> 
> Also validate the device of explicitly supplied streams and add a multi-GPU regression test where the current device differs from the input device.
> 
> ### [P2] Honor or reject non-contiguous MTP output buffers
> **Location:** [`aiter/ops/flydsl/kernels/gdr_decode.py:856–866`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/flydsl/kernels/gdr_decode.py#L856-L866)
> 
> `out_view` uses hard-coded contiguous strides, but `_mtp_common_checks()` checks only the output’s dtype and device—not its shape or contiguity.
> 
> A correctly shaped output view such as:
> 
> ```python
> backing = torch.full(
>     (2, 2, 4, 256), 123.0, device="cuda", dtype=torch.bfloat16
> )
> out = backing[..., ::2]  # Shape: (2, 2, 4, 128)
> ```
> 
> is accepted, but the kernel writes as though `out` were contiguous. In my gfx950 reproduction, the result differed from the contiguous-output baseline, and **1,024 elements of `backing[..., 1::2]` were modified**, even though those elements are outside the supplied output view.
> 
> Please pass the actual output strides to the kernel, or explicitly validate the expected shape and require contiguous output before launching. Creating a temporary contiguous copy alone would not preserve the caller-owned output-buffer contract.
> 
> ### [P2] Forward `accepted` into the rollback test fixture
> **Location:** [`op_tests/test_flydsl_gdr_mtp.py:550–556`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/op_tests/test_flydsl_gdr_mtp.py#L550-L556)
> 
> `_check_chain_case()` receives `accepted`, but does not forward it to `_make_problem()`. Consequently, the parameterized cases labeled `"mixed"` and `"first"` use the fixture’s default `"full"` acceptance pattern.
> 
> Please add:
> 
> ```python
> accepted=accepted,
> ```
> 
> to the fixture call so these cases exercise their intended rollback positions. I reran the four chain cases with this correction, and they passed; this is a coverage gap rather than an observed recurrence failure.
> 
> ### Measure public-entry-point latency, including the dispatch synchronization
> **Location:** [`aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py:29–48`](https://github.com/ROCm/aiter/blob/041211b7ec423aee851eb3099a2460aafbd36dc0/aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py#L29-L48)
> 
> `_uniform_draft_window()` performs a GPU-to-CPU copy and synchronization on every eligible dispatch. This overhead can outweigh the kernel-level improvement for small batches.
> 
> In a warmed-up gfx950 wall-clock microbenchmark of the public entry point with `B=1`, `T=4`, and `HV=32`, I measured:
> 
> Configuration	Median time per call
> FlyDSL disabled	35.9 µs
> FlyDSL enabled	69.5 µs
> This is one local microbenchmark, not a model-level performance result, but it demonstrates why kernel-only speedups should be reported separately from public-API latency.
> 
> Please add a benchmark that includes dispatch overhead and consider accepting the draft window through host-side metadata to avoid synchronization in the hot path. Also document that the current dispatch falls back to Triton during CUDA Graph capture.
> 
> ### Align the “bit-exact” claim with the current validation
> The PR description claims bit-exact agreement with the upstream kernels, but the current tests generally validate computed outputs and recurrent states against an error budget. Exact-equality assertions mainly cover copied state, untouched regions, or routed-versus-direct calls to the same implementation.
> 
> In a same-dtype gfx950 comparison against the vendored vLLM reference, I observed nonzero differences in both output and updated state, with a maximum output difference of approximately `4.88e-4`.
> 
> These differences do not by themselves establish a numerical correctness issue, but they do not support an unconditional bit-exact claim. Please distinguish numerical agreement within tolerance from exact equality, and update the reported test counts to match the current suite.
> 
> **Review result: `REQUEST_CHANGES`**
> 
> Please address the index interpretation, stream-device selection, and output-layout issues and add regression coverage before merging.



Thanks for the thorough review and the concrete reproductions — they were very helpful in identifying both the correctness issues and the public-dispatch performance gap.

### 1. Validate index dtypes before routing conv1d updates to FlyDSL

Addressed.

The Conv1d support predicates now validate the dtype, device, rank, and required contiguity of all index tensors read as int32, including:

- `conv_state_indices`
- `num_accepted_tokens` / `num_accept_tokens`
- `query_start_loc`
- `block_idx_last_scheduled_token`
- `initial_state_idx`
- `intermediate_state_indices`
- EAGLE retrieval maps

Unsupported inputs now fall back to Triton at the dispatch seam. Direct FlyDSL wrapper calls reject unsupported index tensors explicitly with `ValueError`.

A regression test uses the reported int64 case:

`conv_state_indices = torch.tensor([1, 2], dtype=torch.int64, device="cuda")`

With FlyDSL dispatch enabled, the call now falls back to Triton. Its output and complete state pool match the disabled path exactly, including leaving slot 0 unchanged. Parameterized wrapper tests also cover all other int32 index operands.

### 2. Select the default MTP stream from the input device

Addressed.

Both MTP wrappers now select the default stream with:

`torch.cuda.current_stream(query.device)`

Explicitly supplied streams are checked against `query.device` and rejected when they belong to another device.

The multi-GPU regression test places all operands on `cuda:1` while `cuda:0` is current. It verifies that the selected stream belongs to `cuda:1`, and that an explicitly supplied `cuda:0` stream is rejected.

### 3. Honor or reject non-contiguous MTP output buffers

Addressed by rejecting unsupported output layouts.

The common MTP validation now requires `out` to have the expected shape and to be contiguous before launch.

The regression test passes a correctly shaped `backing[..., ::2]` view and verifies that:

- the call raises `ValueError`;
- the complete backing tensor remains unchanged;
- no elements outside the supplied view are overwritten.

### 4. Forward `accepted` into the rollback test fixture

Fixed.

`_check_chain_case()` now forwards `accepted=accepted` to `_make_problem()`. The `"full"`, `"mixed"`, and `"first"` cases therefore exercise their intended rollback positions. All four parameterized chain cases pass.

### 5. Measure public-entry-point latency, including dispatch synchronization

Addressed.

The public dispatch now accepts `draft_window` as host-side metadata. It no longer reads `cu_seqlens` back to the CPU to determine the window.

The resulting behavior is:

- valid host `draft_window` in eager mode: dispatch to FlyDSL;
- missing or unsupported metadata: fall back to Triton;
- CUDA Graph capture: fall back to Triton.

The CUDA Graph fallback is now documented.

We also removed duplicated validation between the support predicate and the public FlyDSL wrapper by adding a private validated launch path, and cached the architecture check.

The benchmark now measures the complete `fused_rearrange_sigmoid_gated_delta_rule()` public entry with a warmed-up per-call wall-clock median. Environment-variable changes are performed outside the timed call.

On gfx950 for the requested `B=1`, `T=4`, `HV=32` case, using three runs with 101 measured calls per run:

- Direct FlyDSL: **36.034 µs**
- Public entry → FlyDSL: **49.995 µs**
- Public entry → Triton: **53.861 µs**
- Vendored vLLM reference: **87.211 µs**

The current public FlyDSL route is therefore **3.87 µs / 7.2% faster** than the same public entry with FlyDSL disabled.

The absolute values differ from the reported 35.9/69.5 µs because the benchmark harness and software environment are not identical. However, the synchronization gap is consistent:

- Reported regression: `69.5 - 35.9 = 33.6 µs`
- Previous synchronized dispatch in our environment: `86.24 - 53.53 = 32.71 µs`
- Current metadata dispatch: `49.995 - 53.861 = -3.87 µs`

Thus, the previous synchronization regression is reproduced by the gap and is removed by the metadata-based dispatch.

The direct and public FlyDSL results are reported separately because their input layouts differ. The direct benchmark uses separate contiguous Q/K/V tensors. The production public entry receives packed 2-D QKV; Triton consumes it directly, while FlyDSL currently constructs 4-D strided Q/K/V views. Consequently, the direct result is a kernel baseline, while the public-entry result represents the production dispatch benefit.

### 6. Align the “bit-exact” claim with the current validation

Addressed in the PR description.

The result is now described as numerical agreement with the vendored vLLM and SGLang kernels within the configured tolerances, rather than unconditional bit-exact agreement.

Exact-equality claims are limited to copied state, untouched storage, rollback state, and routed-versus-direct execution of the same implementation.

The reported test counts are also updated to match the current collected suite:

- Conv1d: **41 cases**
- GDR MTP: **19 cases**

### Validation

The focused Conv1d and GDR MTP suites pass all **60 collected cases**. The GDR MTP plus public fused-entry regression run reports **31 passed**. Formatting, lint, and diff checks also pass.

xytpai:
## Review result: **REQUEST_CHANGES**

Reviewed the latest head: **`23ef665f5580`**.

The previous major issues are fixed. I reran the relevant tests on gfx950: **90 passed**. However, two input-validation gaps remain.

### Previous findings — resolved

- **Conv int64 indices:** unsupported indices now fall back to Triton; direct FlyDSL calls reject them.
- **MTP stream selection:** defaults to the input device’s stream. I also verified an actual cross-device launch produces matching output and state.
- **Non-contiguous MTP output:** now rejected before launch.
- **Rollback fixture:** `accepted` is correctly forwarded.
- **Dispatch synchronization:** the D2H inference was removed; dispatch now requires host-side `draft_window`.
- **Bit-exact claim:** the description now distinguishes numerical tolerance from exact equality.

---

### [P2] Validate the snapshot buffer’s device in the public SGLang MTP entry point

**Location:** [`linear_attention_kernels.py:990–1013`](https://github.com/ROCm/aiter/blob/23ef665f55800781286b490c17f8bf6a44d4ab0e/aiter/ops/flydsl/linear_attention_kernels.py#L990-L1013)

`flydsl_gdr_mtp_sglang()` validates the snapshot buffer’s shape, K stride, and element size, but does not check that `intermediate_states_buffer.device == query.device`.

`_mtp_common_checks()` does not receive this buffer, and the public entry point does not call `_flydsl_gdr_mtp_sglang_supported()`, which already checks device placement.

I verified that, with inputs on `cuda:0`, both a CPU snapshot buffer and a `cuda:1` snapshot buffer reach `_mtp_launch()` without rejection. This verification intercepted the launch; I did **not** execute a kernel against foreign memory.

Please reject the buffer before launch:

```python
if intermediate_states_buffer.device != query.device:
    raise ValueError(
        "`intermediate_states_buffer` must be on the same device as `query`."
    )
```

Also enforce its supported dtype set explicitly rather than checking element size alone. Add direct-entry-point tests for CPU and wrong-GPU snapshot buffers.

---

### [P2] Preserve index-shape validation when bypassing the public MTP wrapper

**Location:** [`fused_rearrange_sigmoid_gdr.py:151–167`](https://github.com/ROCm/aiter/blob/23ef665f55800781286b490c17f8bf6a44d4ab0e/aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py#L151-L167)

The dispatch now calls `_launch_flydsl_gdr_mtp()` directly to avoid repeated validation. However, `_flydsl_gdr_mtp_supported()` does not enforce all the index-shape checks performed by the public wrapper.

For a two-sequence batch with:

```python
num_accepted_tokens.shape == (1,)
```

I verified:

- `_flydsl_gdr_mtp_supported()` returns `True`.
- Direct `flydsl_gdr_mtp()` correctly raises `ValueError`.
- The routed entry point forwards the undersized tensor to the private launch.

The kernel reads one accepted-token count per sequence, so this permits a read beyond the supplied tensor’s logical extent. I intercepted the private launch rather than executing this invalid access.

Please make the support predicate enforce the necessary index shapes, including:

```python
num_accepted_tokens.ndim == 1
num_accepted_tokens.shape[0] == query.shape[0]
ssm_state_indices.shape[0] == query.shape[0]
```

The private launch should only be reached after equivalent safety checks—not merely a subset of the public wrapper’s validation.

---

### CI status still needs clarification

The latest head has **50 successful, 20 skipped, and 3 failed checks**:

- [Triton Tests — MI35X / Shard 0](https://github.com/ROCm/aiter/actions/runs/34554268331/job/103124683110)
- [Accuracy — DeepSeek-R1-0528](https://github.com/ROCm/aiter/actions/runs/34554268304/job/103123837125)
- [Accuracy — Kimi-K2.7-Code-MXFP4](https://github.com/ROCm/aiter/actions/runs/34554268304/job/103123837064)

The log API returned HTTP 403, so I cannot attribute these failures to this PR. Please provide the failure logs or a successful rerun before merging.

**Bottom line:** the original correctness regressions are resolved. The remaining findings concern invalid-input handling, not a newly reproduced numerical failure on supported inputs. I recommend closing these two validation gaps and clarifying the failed CI checks before approval.

xytpai (dismissed):

yiijin:
> ## Review result: **REQUEST_CHANGES**
> Reviewed the latest head: **`23ef665f5580`**.
> 
> The previous major issues are fixed. I reran the relevant tests on gfx950: **90 passed**. However, two input-validation gaps remain.
> 
> ### Previous findings — resolved
> * **Conv int64 indices:** unsupported indices now fall back to Triton; direct FlyDSL calls reject them.
> * **MTP stream selection:** defaults to the input device’s stream. I also verified an actual cross-device launch produces matching output and state.
> * **Non-contiguous MTP output:** now rejected before launch.
> * **Rollback fixture:** `accepted` is correctly forwarded.
> * **Dispatch synchronization:** the D2H inference was removed; dispatch now requires host-side `draft_window`.
> * **Bit-exact claim:** the description now distinguishes numerical tolerance from exact equality.
> 
> ### [P2] Validate the snapshot buffer’s device in the public SGLang MTP entry point
> **Location:** [`linear_attention_kernels.py:990–1013`](https://github.com/ROCm/aiter/blob/23ef665f55800781286b490c17f8bf6a44d4ab0e/aiter/ops/flydsl/linear_attention_kernels.py#L990-L1013)
> 
> `flydsl_gdr_mtp_sglang()` validates the snapshot buffer’s shape, K stride, and element size, but does not check that `intermediate_states_buffer.device == query.device`.
> 
> `_mtp_common_checks()` does not receive this buffer, and the public entry point does not call `_flydsl_gdr_mtp_sglang_supported()`, which already checks device placement.
> 
> I verified that, with inputs on `cuda:0`, both a CPU snapshot buffer and a `cuda:1` snapshot buffer reach `_mtp_launch()` without rejection. This verification intercepted the launch; I did **not** execute a kernel against foreign memory.
> 
> Please reject the buffer before launch:
> 
> ```python
> if intermediate_states_buffer.device != query.device:
>     raise ValueError(
>         "`intermediate_states_buffer` must be on the same device as `query`."
>     )
> ```
> 
> Also enforce its supported dtype set explicitly rather than checking element size alone. Add direct-entry-point tests for CPU and wrong-GPU snapshot buffers.
> 
> ### [P2] Preserve index-shape validation when bypassing the public MTP wrapper
> **Location:** [`fused_rearrange_sigmoid_gdr.py:151–167`](https://github.com/ROCm/aiter/blob/23ef665f55800781286b490c17f8bf6a44d4ab0e/aiter/ops/triton/gated_delta_net/fused_rearrange_sigmoid_gdr.py#L151-L167)
> 
> The dispatch now calls `_launch_flydsl_gdr_mtp()` directly to avoid repeated validation. However, `_flydsl_gdr_mtp_supported()` does not enforce all the index-shape checks performed by the public wrapper.
> 
> For a two-sequence batch with:
> 
> ```python
> num_accepted_tokens.shape == (1,)
> ```
> 
> I verified:
> 
> * `_flydsl_gdr_mtp_supported()` returns `True`.
> * Direct `flydsl_gdr_mtp()` correctly raises `ValueError`.
> * The routed entry point forwards the undersized tensor to the private launch.
> 
> The kernel reads one accepted-token count per sequence, so this permits a read beyond the supplied tensor’s logical extent. I intercepted the private launch rather than executing this invalid access.
> 
> Please make the support predicate enforce the necessary index shapes, including:
> 
> ```python
> num_accepted_tokens.ndim == 1
> num_accepted_tokens.shape[0] == query.shape[0]
> ssm_state_indices.shape[0] == query.shape[0]
> ```
> 
> The private launch should only be reached after equivalent safety checks—not merely a subset of the public wrapper’s validation.
> 
> ### CI status still needs clarification
> The latest head has **50 successful, 20 skipped, and 3 failed checks**:
> 
> * [Triton Tests — MI35X / Shard 0](https://github.com/ROCm/aiter/actions/runs/34554268331/job/103124683110)
> * [Accuracy — DeepSeek-R1-0528](https://github.com/ROCm/aiter/actions/runs/34554268304/job/103123837125)
> * [Accuracy — Kimi-K2.7-Code-MXFP4](https://github.com/ROCm/aiter/actions/runs/34554268304/job/103123837064)
> 
> The log API returned HTTP 403, so I cannot attribute these failures to this PR. Please provide the failure logs or a successful rerun before merging.
> 
> **Bottom line:** the original correctness regressions are resolved. The remaining findings concern invalid-input handling, not a newly reproduced numerical failure on supported inputs. I recommend closing these two validation gaps and clarifying the failed CI checks before approval.



Thanks for the re-review. Both validation gaps are closed in the next push.

**[P2] Snapshot buffer device in the public SGLang MTP entry**

`flydsl_gdr_mtp_sglang()` now rejects `intermediate_states_buffer` before the launch when `.device != query.device`, and enforces the supported dtype set explicitly rather than element size alone — an fp16 snapshot against an fp32 state passed the old itemsize check and no longer does. The narrower-than-state itemsize rule stays after it, since it is a separate constraint.

`test_mtp_rejects_invalid_snapshot_buffer` exercises the direct entry point with a CPU buffer, a `cuda:1` buffer (skipped on a single-GPU host), and an unsupported dtype.

**[P2] Index-shape validation behind the dispatch seam**

`_flydsl_gdr_mtp_supported()` now enforces the three shapes you listed: `num_accepted_tokens.ndim == 1`, `num_accepted_tokens.shape[0] == query.shape[0]`, and `ssm_state_indices.shape[0] == query.shape[0]`. Your two-sequence case with `num_accepted_tokens.shape == (1,)` returns False, so the routed entry falls back to Triton instead of reaching `_launch_flydsl_gdr_mtp()`, while the well-formed case still routes. `test_mtp_dispatch_rejects_undersized_index_batches` covers both.

On the general point that the private launch should only be reached after equivalent checks: I walked the rest of `_mtp_common_checks` against `_try_flydsl_mtp`. The `A_log` dtype, operand device agreement, the 16-byte `state` alignment, and `core_attn_out`'s dtype/device/contiguity are each screened there before the predicate runs, and `min_live_slot` is fixed at 0. The index shapes were the only gap.

Reran on gfx950: `op_tests/test_flydsl_gdr_mtp.py` 21 passed, `op_tests/triton_tests/conv/test_causal_conv1d.py` 4921 passed.

coderfeli (approved):

PR #4922, opened 2026-08-22
[Config] perf(kimi-k3): tune K2 c8 BF16 projections on gfx950

JohnQinAMD:
## Purpose

Kimi-K3 K2 verification at concurrency 8 produces `M=24`. The bf16 tuned-GEMM table
has no exact `gfx950` entry at that M for either the routed down projection
(`N=3584, K=7168`) or the fused QKVGFA projection (`N=6288, K=7168`), so both fall
back to torch solution 0 on every layer.

This adds the two measured hipblaslt solutions. `tuned_gemm` already dispatches
`libtype == "hipblaslt"` and honours an explicit `solidx`; these are the first
hipblaslt rows in this particular table, which is why the shapes were missing rather
than mistuned.

Only the exact `gfx950` M24 dispatch changes. Every other batch size, arch and shape
takes the same path as before.

## Test Plan

Measured through the production wrapper under HIP graph replay on MI355X, comparing
the fallback against the selected solution at the exact shapes.

## Test Result

| shape | before | after | |
|---|---|---|---|
| `M24 N3584 K7168` (routed down) | 20.27 us | 18.43 us | **1.100x** |
| `M24 N6288 K7168` (fused QKVGFA) | 24.67 us | 21.21 us | **1.163x** |

Solutions `441426` and `440174` respectively.

Note that hipblaslt solution indices are tied to the ROCm build they were tuned on;
these were selected on ROCm 7.2.


JohnQinAMD:
Closing for now — will re-open once the surrounding work is prepared.

PR #4924, opened 2026-08-23
fix(dist): make raw IPC input pools usable — remove init_dist_env's vestigial signal/buffer block, add explicit raw-pool override

### LINKED ISSUE #4921: [Bug] raw_cached IPC pool is unusable: init_dist_env reads _pool['input'].tensor, which raises for plain-hipMalloc buffers
### Summary

`aiter/ops/communication.py:78` (`init_dist_env`) unconditionally does:

```python
ca_comm.buffer = ca_comm._pool["input"].tensor
```

But when `CustomAllreduce` builds its input pool with `raw_cached=True`, that buffer is a plain `hipMalloc` region with **no backing `torch.Tensor`**, so the `tensor` property raises by design (`custom_all_reduce.py:447`):

```
RuntimeError: Uncached IPCBuffer has no backing tensor; use .data_ptr
```

Net effect: **the `raw_cached` (plain-`hipMalloc`, IPC-exportable) input pool cannot be used at all.** Any configuration that selects it fails during distributed init.

That matters because `raw_cached` is the path that keeps the pool exportable via `hipIpcGetMemHandle`, and it is currently the *only* mitigation for co-resident engines — see Impact.

### Reproduction

8× MI350X (gfx950, SR-IOV VF), ROCm 7.2, `rocm/atom-dev:nightly_202608191459` (bundled AITER `4fa508ef`), GLM-5.2-MXFP4, one TP4 engine:

```
RuntimeError: Uncached IPCBuffer has no backing tensor; use .data_ptr
  File "/app/aiter-test/aiter/ops/communication.py", line 78, in init_dist_env
    ca_comm.buffer = ca_comm._pool["input"].tensor
  File "/app/aiter-test/aiter/dist/device_communicators/custom_all_reduce.py", line 447, in tensor
  File "/app/ATOM/atom/model_engine/model_runner.py", line 941, in _setup_device_and_distributed
```

`raw_cached` is gated at `custom_all_reduce.py:1062`:

```python
raw_cached = _expandable_segments_enabled()
```

so it is reached two ways, and **both fail identically**:

1. `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True` — the supported trigger.
2. A one-line local patch adding an explicit override (`... or _env_flag("AITER_CAR_RAW_CACHED")`).

**Controls run to exclude our own environment:**

| configuration | result |
|---|---|
| patched file mounted, override **off** (`raw_cached=False`) | server **READY** — so the patch itself is harmless |
| override **on**, with our unrelated local AITER patches mounted | dies |
| override **on**, **pristine** container (no other patches) | dies — same traceback |

So this is not caused by anything we mount, and it is not a version skew: the only `custom_all_reduce*` change between `fc2e5d57` (v0.1.20) and `4fa508ef` is `@compile_ops(..., develop=True)` on three `fused_allreduce_mhc_*` stubs.

### Impact

On this VF hardware, running **two TP4 engines on one 8-GPU node** fails at the default envelope:

```
custom_all_reduce.cu:417 fail to call hipIpcGetMemHandle(...) ---> [HIP error](invalid argument)
```

The first engine starts; the second cannot export its handle. Consistent with the `torch.empty` pool returning a caching-allocator pointer that is not IPC-exportable — exactly what `raw_cached` exists to avoid, and what #4174 / #4621 address for expandable segments.

Because `raw_cached` is unusable, the only working mitigation we found is to buy allocator headroom:

| `--gpu-memory-utilization` / `--max-num-seqs` | two engines? |
|---|---|
| 0.78 / 256 | no |
| 0.70 / 128 | no |
| 0.62 / 256 | no |
| **0.62 / 128** | **yes** |

Both dimensions must drop together. That costs real KV cache and half the sequence slots. Measured node throughput at 0.62/128 is 8465 tok/s aggregate; the alternative (guarding the second engine with `NCCL_P2P_DISABLE=1` + `AITER_DISABLE_IPC_COLLECTIVES=1`) drops that engine's own end-to-end throughput by ~50% and yields 6440 aggregate.

### Suggested fix

`init_dist_env` should use the pointer rather than the tensor, since `IPCBuffer` already exposes `data_ptr` for exactly this case — e.g. keep `.tensor` only where a tensor is genuinely required, or have `init_dist_env` branch on `uncached`/`raw_cached`. Happy to send a PR if you confirm the intended contract for `ca_comm.buffer` (tensor vs raw pointer) — it is consumed downstream and I did not want to guess.

Separately, it would help operationally if `raw_cached` were selectable on its own (e.g. `AITER_CAR_RAW_CACHED=1`) rather than only as a side effect of `expandable_segments`: co-resident deployments need the exportable pool but do not necessarily want expandable segments, which independently prevents this engine from starting.

### Environment

- 8× MI350X (gfx950) SR-IOV VF, ROCm 7.2, TP4 per engine
- `rocm/atom-dev:nightly_202608191459`, bundled AITER `4fa508ef2935110ff99adf2743ea93807dbd9c67`
- Model GLM-5.2-MXFP4, `--level 3 --method mtp --num-speculative-tokens 3`
- `PYTORCH_HIP_ALLOC_CONF` unset unless stated

ThomasNing:
Fixes #4921.

## Summary

Two commits:

1. **`init_dist_env` no longer breaks raw IPC input pools** — the vestigial signal/buffer block is removed.
2. **`AITER_CUSTOM_AR_RAW_INPUT_POOL`** — an explicit trigger for the raw (plain-`hipMalloc`) input pool that works, because the existing trigger (expandable segments) turns out to be broken further down regardless.

## Root cause — three stacked failures, not one

#4921 reports the crash at `communication.py:78` (`ca_comm.buffer = ca_comm._pool["input"].tensor`). Measured on 4×/8× MI350X (gfx950, both VF and bare metal), the full chain under `PYTORCH_HIP_ALLOC_CONF=expandable_segments:True` is:

1. **The first failure is earlier than the issue's traceback**: `register_input_buffer(signal)` exports the `torch.zeros` signal tensor's pointer via `hipIpcGetMemHandle`, and under expandable segments that pointer is VMM-backed — the export dies at `custom_all_reduce.cu:417` ("invalid argument"). The issue's line-78 traceback is the signature of the *explicit-override* trigger, where the allocator stays default and registration survives.
2. **Line 78** — the raw pool has no backing `torch.Tensor`, and `.tensor` raises by design.
3. **With 1 and 2 fixed, a third failure appears** (previously unreachable): CUDA-graph capture completes, then `flush_graph_buffers` dies at `custom_all_reduce.cuh:3448` exporting **capture-time output tensors** — every `out = torch.empty_like(inp)` allocated during capture is VMM-backed, `get_output_buffer_RD` records it, and the post-capture flush tries to IPC-export it. #4621's copy-in guard covers inputs only.

So expandable segments cannot arm the raw pool usefully today (failure 3 needs output staging — a separate, kernel-level change). The explicit override gives the raw pool a trigger that works: under the default allocator everything else (meta pool, capture-time outputs, graph flush) stays exportable, and only the input pool moves to `hipMalloc`.

## Why removal instead of `.tensor` → `.data_ptr`

The issue asked which contract `ca_comm.buffer` serves. Answer: none —

- `ca_comm.buffer` / `ca_comm.signal` are only ever **written**, never read (swept aiter and a downstream engine, including `hasattr` gates);
- every internal consumer of the input pool goes through `_pool["input"].data_ptr`;
- C++ `register_input_buffer` only inserts a pointer-translation entry keyed by the registered tensor's own address, consulted only when an allreduce is invoked *with that exact tensor as input* — which never happens for the signal tensor (`open_ipc_handle`'s cache fills on demand, so no pre-warming is lost);
- gfx1250 already skips the entire block in-tree and works.

The old comment said the attributes are "never read anywhere" while the code still ran the registration — removal makes the code match the comment. **Numerics were validated, not just startup** (see below), since a wrong guess here would corrupt allreduce silently.

## Measured validation (GLM-5.2-MXFP4, TP4, `--level 3` graphs, `gpu_memory_utilization 0.90`)

Two environments, stated per row — provenance matters here because bring-up
behaviour in this problem space has proven state-sensitive (see motivation):

- **[A]** bare-metal gfx950, image `nightly_202608041536` + a source overlay of the two files this PR touches plus their csrc (JIT-rebuilt in-container; ≡ aiter `e404860` + this PR).
- **[B]** gfx950 **VF**, stock image `nightly_202608191459` (aiter `4fa508e`) with this PR's files bind-mounted, next to another tenant's live engine.

| config | env | result |
|---|---|---|
| unpatched, expandable segments | A | ❌ dies at `cu:417` (signal export — failure 1) |
| unpatched, expandable segments | B | ❌ same signature |
| patched, expandable segments | A | ❌ dies at `cuh:3448` post-capture (failure 3 — expected, out of scope) |
| **patched + `AITER_CUSTOM_AR_RAW_INPUT_POOL=1`** | A | ✅ **READY at full envelope, 0 IPC errors** |
| patched, default alloc | A | ✅ READY (no-regression) |
| patched + override, second co-resident TP4 engine | A | ✅ both READY at 0.90; **TPOT identical to solo** (0.0196 s) — see IPC-matrix caveat below |
| patched + override | B | ✅ READY, 0 IPC errors; TPOT 0.01561 vs 0.01664 unpatched-default (no regression) |

Numerics: inference through the raw pool produces correct output (spot-checked deterministic arithmetic prompts; allreduce path exercised end-to-end at TP4 with HIP graphs).

Operational notes for the co-resident-engines use case that motivated #4921:

- **With this patch**, on the older engine-stack image used for the dual-engine run, second-engine bring-up was **nondeterministic across IPC-namespace configurations** (measured matrix, one run per cell: each engine solo READY under host or private IPC; host+host → e2 dies at NCCL init; host+private sequential → both READY, benched below; private+private → e2 dies at NCCL init whether brought up sequentially or concurrently). We do **not** claim an IPC-namespace mechanism — and, importantly, **no configuration axis has survived testing**: an independent replication ran the *identical* unpatched two-engine command that had succeeded 3/3 in one window and had it fail hours later on a verified-clean node (no orphaned IPC segments, no stale KFD holders, 0% VRAM). Unpatched second-engine bring-up appears **state- or timing-dependent under identical configuration**; image, IpcMode, envelope, GC settings, capabilities and shm size have all failed to separate outcomes. A success-rate measurement is in progress; treat every single-run cell in the matrix above (including the successes) accordingly. The dual-engine numbers in the table were measured on the one configuration of that matrix that came up (engine 1 `--ipc host`, engine 2 `--ipc private`, verified via `docker inspect`).
- **Without this patch**, the unpatched two-engine workaround is **nondeterministic under identical configuration**: an independent VRAM-instrumented effort (clean boxes, IpcMode captured via `docker inspect`) had *one* successful bring-up of the `0.62/128` two-engine configuration — whose engine pair then stayed durable through six consecutive benchmark passes — and the *identical command* failed at `hipIpcGetMemHandle` hours later on a node verified clean beforehand. Every envelope above `0.62/128` has failed in all unpatched attempts, and no configuration axis tried so far (image, IpcMode, GC settings, capabilities, shm size) separates success from failure. **The failure is concentrated entirely in bring-up**: it fails fast (~50 s), leaves the first engine healthy, and pairs that do come up stay durable — so the operational shape of the problem is a bring-up lottery, not a serving instability. That is the practical motivation for this override — with one scope limit found by the same investigation: under co-residency on at least one stack, `NCCL_DEBUG=INFO` shows **RCCL's own P2P transport failing `hipIpcGetMemHandle` first** (`p2p.cc:256`), with AITER's `cu:417` failure downstream of it. So this override **removes the AITER-side export dependency** (an exportable-by-construction hipMalloc pool instead of allocator-state-dependent pointers); it does **not** claim to make two co-resident engines reliable by itself, because RCCL's own export sits upstream and is outside this PR. (`NCCL_DMABUF_ENABLE=1` was evaluated for that layer and did **not** help: 0/3 with it and 0/3 without, interleaved, `p2p.cc:256` present in all six.) Side by side: `0.62/128` unpatched with an unquantified success rate — vs **`0.90` util / 512 seqs patched** (measured above, single-engine).

## Co-residency bring-up rates (appended after measurement — the promised control)

Measured on the matched image (`nightly_202608191459`), sequential bring-up, second engine on disjoint GPUs, `NCCL_DEBUG` enabled, teardown + clean-node checks between attempts:

| arm | host | attempts | successes | first failure |
|---|---|---|---|---|
| patched + override (pool armed, logline-proven) | bare-metal, `iommu=pt` | 5 | 0 | RCCL `p2p.cc:256` every time |
| **unpatched control, same envelope, same host** | bare-metal, `iommu=pt` | 3 | 0 | RCCL `p2p.cc:256` every time |
| patched + override | VF, no `iommu` flag | 9+ | 1 | RCCL `p2p.cc:256` every time |

Conclusions, stated at their real strength:

- **This patch is orthogonal to co-residency bring-up** — patched 0/5 vs unpatched 0/3 on the same host and envelope, identical failure signature. It neither causes nor cures the bring-up lottery, which is what "no co-residency claim" means concretely.
- The first failure on **both hosts, both stacks, both IpcMode shapes** is RCCL's own P2P transport (`hipIpcGetMemHandle → invalid argument` at `p2p.cc:256`), *before* any AITER pool exists. A separate issue against RCCL is being prepared with full `NCCL_DEBUG=INFO` evidence from both hosts.
- The one successful bring-up (pooled ~1/30 across hosts) was **durable**: six benchmark passes without incident, and across ~25 consecutive second-engine failures the incumbent engine never dropped or needed restarting — the fault is scoped to the joining process's export path.
- What this PR fixes is unaffected by all of the above: the #4921 contract break dies at init deterministically without it, and the override arms the raw pool (logline-proven on two hosts).
- The dual-engine numbers above are not "co-residency is free": at c=32 each engine ran ~21–27% below solo in a short window (aggregate +51%), with **identical TPOT** — decode is unaffected; the loss is prefill/queueing contention. Note the dual aggregate curve **turns over past the knee** (c=128+128 measured *below* c=32+32 in this short-window harness); don't extrapolate aggregate upward with concurrency.

## Test

`op_tests/multigpu_tests/test_init_dist_env.py` (new): brings up `init_dist_env` per rank and checks one allreduce under three modes — `default`, `expandable`, `raw_override` — asserting the override actually selects the raw pool. The existing `test_custom_allreduce.py` performs its own init and never executes `init_dist_env`, which is how the regression shipped.

Not covered by CI: failure 3 (needs graph capture + expandable segments); documented here and reproducible with the table above.

### DIFF
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -1059,7 +1059,15 @@ class CustomAllreduce:
         # PyTorch caching allocator (its memory accounting is what downstream
         # consumers profile against). Gated on the same condition as the
         # capture copy-in path above; _init_ipc only runs for non-VMM.
-        raw_cached = _expandable_segments_enabled()
+        # AITER_CUSTOM_AR_RAW_INPUT_POOL forces the raw pool without
+        # expandable segments. Its use case is co-resident engines on one node
+        # (#4921): a second engine's torch.empty input pool can fail
+        # hipIpcGetMemHandle outright, and the raw pool sidesteps that while
+        # everything else (meta pool, capture-time outputs) stays exportable
+        # under the default allocator.
+        raw_cached = _expandable_segments_enabled() or _env_flag(
+            "AITER_CUSTOM_AR_RAW_INPUT_POOL"
+        )
         self._pool.create("input", max_size, raw_cached=raw_cached)
 
         handles, offsets = self._pool.get_ipc_meta("meta")
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -1069,6 +1069,17 @@ class CustomAllreduce:
             "AITER_CUSTOM_AR_RAW_INPUT_POOL"
         )
         self._pool.create("input", max_size, raw_cached=raw_cached)
+        # One line per rank so every run self-documents which pool it actually
+        # got: a silently-inert trigger is indistinguishable from a working one
+        # by behaviour alone (both serve fine single-engine), and this class of
+        # bug (#4921) was reachable only in specific pool modes.
+        logger.info(
+            "CustomAllreduce input pool: %s (expandable_segments=%s, "
+            "AITER_CUSTOM_AR_RAW_INPUT_POOL=%s)",
+            "raw_cached/hipMalloc" if raw_cached else "torch.empty",
+            _expandable_segments_enabled(),
+            _env_flag("AITER_CUSTOM_AR_RAW_INPUT_POOL"),
+        )
 
         handles, offsets = self._pool.get_ipc_meta("meta")
         self._ptr = self._ops_init_custom_ar(
--- a/aiter/dist/device_communicators/custom_all_reduce.py
+++ b/aiter/dist/device_communicators/custom_all_reduce.py
@@ -1584,7 +1584,7 @@ class CustomAllreduce:
                     residual_inp,
                     w=weight,
                     eps=eps,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                     use_1stage=use_1stage,
                     out_hidden_dim=out_hidden_dim,
                     gemma_norm=gemma_norm,
@@ -1688,7 +1688,7 @@ class CustomAllreduce:
                     residual_inp,
                     w=weight,
                     eps=eps,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                     use_1stage=use_1stage,
                     post_per_token_quant=True,
                     gemma_norm=gemma_norm,
@@ -1886,7 +1886,7 @@ class CustomAllreduce:
                     q_w,
                     k_w,
                     eps,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                 )
             else:
                 return (
@@ -1955,7 +1955,7 @@ class CustomAllreduce:
                     head_dim,
                     rotary_dim,
                     eps,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                 )
             else:
                 return (
@@ -2057,7 +2057,7 @@ class CustomAllreduce:
                     w=weight,
                     eps=eps,
                     group_size=group_size,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                     use_1stage=use_1stage,
                     emit_bf16=emit_bf16,
                     transpose_scale=transpose_scale,
@@ -2116,7 +2116,7 @@ class CustomAllreduce:
                     residual_inp,
                     w=weight,
                     eps=eps,
-                    registered=True,
+                    registered=self.enable_register_for_capturing,
                     use_1stage=use_1stage,
                     emit_bf16=emit_bf16,
                 )
--- a/aiter/ops/communication.py
+++ b/aiter/ops/communication.py
@@ -11,7 +11,7 @@ from ..dist.parallel_state import (
     destroy_distributed_environment,
     destroy_model_parallel,
     ensure_model_parallel_initialized,
-    get_tp_group,
+    get_tp_group,  # noqa: F401 -- re-exported; downstreams import via this module
     init_distributed_environment,
     set_custom_all_reduce,
 )
@@ -57,25 +57,24 @@ def init_dist_env(
         prefill_context_model_parallel_size=prefill_context_model_parallel_size,
     )
 
-    if tensor_model_parallel_size > 1:
-        # hack custom_allreduce
-        tp_grp = get_tp_group()
-        ca_comm = tp_grp.device_communicator.ca_comm
-        # gfx1250 (MI450) uses a VMM-based custom allreduce whose input buffer
-        # is already registered in CustomAllreduce._init_gfx1250(). The extra
-        # register_input_buffer(signal) below routes through get_external_ipc_meta,
-        # whose vmm_exchange rendezvous key is process-local (id(self)+data_ptr),
-        # so it can never align across TP ranks and deadlocks at startup. The
-        # `signal`/`buffer` attributes set here are never read anywhere, so skip
-        # the whole block on gfx1250.
-        if ca_comm is not None and not getattr(ca_comm, "_is_gfx1250", False):
-            # signal
-            signal = torch.zeros(
-                tensor_model_parallel_size * 64, dtype=torch.int64, device=rankID
-            )
-            ca_comm.signal = signal
-            ca_comm.register_input_buffer(signal)
-            ca_comm.buffer = ca_comm._pool["input"].tensor
+    # No per-rank signal/input-buffer registration here. An earlier version
+    # registered a torch.zeros "signal" tensor with CustomAllreduce and mirrored
+    # the input pool as `ca_comm.buffer`. All of it was vestigial:
+    #   * `ca_comm.signal` / `ca_comm.buffer` are never read anywhere;
+    #   * register_input_buffer only inserts a pointer-translation entry keyed
+    #     by the registered tensor's own address, consulted when an allreduce
+    #     is invoked with that exact tensor as input -- which never happens for
+    #     the signal tensor;
+    #   * gfx1250 has skipped the whole block (deadlock in its vmm_exchange
+    #     rendezvous) and works without it.
+    # It was also what made raw IPC input pools unusable (#4921):
+    #   * under PYTORCH_HIP_ALLOC_CONF=expandable_segments:True the torch.zeros
+    #     signal is VMM-backed, so registering it dies in hipIpcGetMemHandle
+    #     (custom_all_reduce.cu:417, "invalid argument");
+    #   * the raw_cached input pool has no backing torch.Tensor, so
+    #     `ca_comm.buffer = ca_comm._pool["input"].tensor` raised by design.
+    # CustomAllreduce.__init__ already sets up its own meta/input pools and the
+    # copy-in path for expandable segments (#4174); nothing further is needed.
     logger.debug(f"RANK: {rankID}/{tensor_model_parallel_size} init_dist_env...")
 
 
--- a/op_tests/multigpu_tests/test_init_dist_env.py
+++ b/op_tests/multigpu_tests/test_init_dist_env.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""init_dist_env must come up whatever allocation mode backs the IPC input pool.
+
+Regression test for #4921. Under PYTORCH_HIP_ALLOC_CONF=expandable_segments:True
+init_dist_env used to fail twice over:
+
+  * it registered a torch.zeros "signal" tensor, whose VMM-backed pointer
+    hipIpcGetMemHandle rejects (custom_all_reduce.cu:417), and
+  * it read `_pool["input"].tensor`, which raises by design for the raw_cached
+    (plain-hipMalloc) input pool that expandable segments selects.
+
+So every raw_cached configuration died at distributed init. Both allocator
+modes are exercised here and share one allreduce-correctness check.
+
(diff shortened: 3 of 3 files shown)


ThomasNing:
Solved with patches

aiter/dist/device_communicators/custom_all_reduce.py
TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223: I think this line shoud change to "registered=self.enable_register_for_capturing,"

TennyWang1223 (changes requested):

aiter/dist/device_communicators/custom_all_reduce.py
ThomasNing: Fixed in dd08de3b5. custom_fused_ar_rms now passes self.enable_register_for_capturing.

ThomasNing: Fixed in dd08de3b5. custom_fused_ar_rms_quant now passes self.enable_register_for_capturing.

ThomasNing: Fixed in dd08de3b5. custom_fused_qknorm_ar now passes self.enable_register_for_capturing.

ThomasNing: Fixed in dd08de3b5. custom_fused_qknorm_ar_rope now passes self.enable_register_for_capturing.

ThomasNing: Fixed in dd08de3b5. custom_fused_ar_rms_per_group_quant now passes self.enable_register_for_capturing.

ThomasNing: Fixed in dd08de3b5. custom_fused_ar_rms_mxfp4_quant now passes self.enable_register_for_capturing.

ThomasNing:
@TennyWang1223 Addressed all six capture-registration comments in dd08de3b5. The fused RMS, quantized RMS, QK-norm, QK-norm+RoPE, per-group quant, and MXFP4 wrappers now pass self.enable_register_for_capturing, matching the existing reduce-scatter/all-gather policy. Ruff 0.16.0, Black, byte-compilation, diff checks, and a focused AST policy check all pass. Please re-review when convenient.

ThomasNing:
The refreshed Checks and Aiter Test workflows are currently marked action_required for this fork head. An upstream maintainer will need to approve those workflow runs before GitHub can execute the CI gates.

TennyWang1223 (approved):
LGTM

ThomasNing:
CI follow-up for head `dd08de3b57a4047fbfc5b9f890dd4f241dbb012f`:

The four required checks are still shown as **Expected** because GitHub blocked the fork workflows before job creation (`conclusion=action_required`, zero jobs), not because the code-style or test jobs failed.

Could an upstream maintainer please approve and run the workflows for this fork head?

- Checks (emits **Check Code Style with Black**, **Check Code Style with Ruff**, and **Check Repository Dependency**): https://github.com/ROCm/aiter/actions/runs/33017532341
- Aiter Test (emits **Standard Test Results**): https://github.com/ROCm/aiter/actions/runs/33017532593

Local reproduction against the exact PR head:

- Black: `1079 files would be left unchanged`; all 3 PR-changed Python files pass.
- Ruff `0.16.0` (the workflow-pinned version): all 3 PR-changed Python files pass.
- The two repository-wide Ruff findings are in unrelated pre-existing files; the workflow uses reviewdog `diff_context`, so they are outside this PR's report scope.

No style-only source change is required for this head; the missing step is workflow approval.

valarLip (approved):

PR #4933, opened 2026-08-23
[Triton/Gluon] [Config] Move GEMM-AFP8WFP8_PRESHUFFLED configs to nested layout

Boss2002n:


Boss2002n:
the 4 deletions are because the configs exist in the new path already

brunomazzottiamd (approved):

PR #4947, opened 2026-08-23
[Triton/Gluon] Satya/unify gluon a8w8 blockscale config resolution

Boss2002n:
https://github.com/ROCm/aiter/pull/4927
satya/migrate-configs-gemm-a8w8-blockscale must land before this


aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -22,6 +22,56 @@ def _dtype_dir(config_name: str) -> str:
     return config_name.lower().replace("-", "_")
 
 
+def resolve_config_dir(
+    op: str,
```
brunomazzottiamd, line 26: Can you please improve the docstring of `resolve_config_dir` function? I think it's a very important function and it should be documented accordingly. For instance, I miss the description and explanation of input arguments. What's the purpose of returning the second tuple element (`name_prefix`)? Can you add an example on how the caller should use it?

Another point of improvement: Do we have invariants to be checked with assertions? I may be wrong, but I think `op` and  `config_name` arguments shouldn't be an empty string. `assert` everything that makes sense w.r.t. input arguments.

brunomazzottiamd: I also have the impression that `resolve_config_dir` should be in `aiter.ops.triton.utils.core` module (together with `load_config_json`) instead of `aiter.ops.triton.utils.gemm_config_utils` module. It's a very fundamental behavior. For instance, https://github.com/ROCm/aiter/pull/5018 rely on `resolve_config_dir` - this PR is related to convolution kernels, not GEMM kernels (at the end it's all GEMM, but I think you get the point of my reasoning...).

aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -22,6 +22,56 @@ def _dtype_dir(config_name: str) -> str:
...
+            (f"{AITER_TRITON_CONFIGS_PATH}/{dev}/gluon/{op}/{dtype_dir}", ""),
+        ]
+        if legacy_dir:
+            candidates.append(
+                (
+                    f"{AITER_TRITON_CONFIGS_PATH}/{legacy_dir}",
+                    arch_prefix,
+                )  # TODO(satya): legacy, remove
+            )
+    else:
```
brunomazzottiamd, line 50: In this case, the `else` block, does it make sense to check if the backend is `triton` or `gluon`? Do we have other backend options?

aiter/ops/triton/utils/gemm_config_utils.py
```diff
@@ -22,6 +22,56 @@ def _dtype_dir(config_name: str) -> str:
...
+                    f"{AITER_TRITON_CONFIGS_PATH}/{legacy_dir}",
+                    arch_prefix,
+                )  # TODO(satya): legacy, remove
+            )
+    for cfg_dir, name_prefix in candidates:
+        # Nested dirs (empty prefix) name their default DEFAULT.json.
+        stem = f"{name_prefix}{config_name}" if name_prefix else "DEFAULT"
+        if os.path.exists(f"{cfg_dir}/{stem}.json"):
+            return cfg_dir, name_prefix
+    return candidates[-1]
```
brunomazzottiamd, line 72: What's the reasoning behind returning the last candidate if no config directory could be successfully resolved? What about raising an exception or returning `None`? What the caller can do with the last candidate?

brunomazzottiamd (commented):
@Boss2002n, I was deliberately very strict with the `resolve_config_dir` function. I think it's a very important function and it should have a solid implementation and documentation. So, I started many related discussions... please don't get me wrong, it's for our own good!

Once this PR is merged, we can proceed faster with the following ones.

aiter/ops/triton/utils/gemm_config_utils.py
Boss2002n: those changes are going to be included in https://github.com/ROCm/aiter/pull/5023

aiter/ops/triton/utils/gemm_config_utils.py
Boss2002n: not as of now but just in case we do in the future

aiter/ops/triton/utils/gemm_config_utils.py
Boss2002n: This was done in case they have defaults within the code blocks but I will raise an error with other changes in https://github.com/ROCm/aiter/pull/5023

aiter/ops/triton/utils/gemm_config_utils.py
brunomazzottiamd: Please address the documentation improvement and assertions in the scope of https://github.com/ROCm/aiter/pull/5023, so we can close this discussion.

aiter/ops/triton/utils/gemm_config_utils.py
brunomazzottiamd: If this is the case then we should check for invalid backend values.

aiter/ops/triton/utils/gemm_config_utils.py
brunomazzottiamd: Please implement this in the scope of https://github.com/ROCm/aiter/pull/5023.

aiter/ops/triton/utils/gemm_config_utils.py
Boss2002n: Yes I will add that check in my utils cleanup PR

Boss2002n:
This PR is just an intermediary PR so that the other Config move PRs do not fail CI 
I will address all these comments in my utils cleanup PR

brunomazzottiamd (approved):

PR #4954, opened 2026-08-24
[AMD][DSV4] feat: MXFP8 activation passthrough in fused_moe

karverma-amd:
DSv4's MoE is w4a8: MXFP4 weights (per_1x32), so the kernels consume fp8 activations carrying group-32 e8m0 microscales. When a caller has already produced exactly that -- an MXFP8 all-to-all dispatch, for instance -- fused_moe should take the activations as-is and only sort the scale, which is what the neighbouring fp4x2 pre-quantized branch already does.

Without this, such a caller lands on a path that re-derives scales, which is the round trip the format was chosen to avoid. Nothing errors; the cost is simply paid again.

12 lines, mirroring the existing pre-quantized branch. No existing path changes: the new branch is reached only when activations arrive as fp8 with a scale already attached.

Motivating consumer is SGLang's MoRI expert-parallel dispatch, which can then quantize on the send side over live tokens rather than leaving the receiver to quantize over a padded all-to-all buffer.

Measured with the matching SGLang change on DeepSeek-V4-Pro, MI355X, TP8 + DP8 attention + EP8/MoRI + EAGLE MTP, 8192 in / 1024 out: +6.4% to +9.8% output throughput and -6.2% to -9.3% TPOT across concurrency 4-64. GSM8K 0.945 over 1319 questions, zero invalid.

Adds op_tests/test_moe_mxfp8_passthrough.py: the branch fails silently in both directions -- absent, it costs throughput with no error; miswired, it reinterprets scale bytes and returns wrong numbers rather than raising. The test pins the group-32 one-byte scale layout and checks the passthrough returns finite, correctly shaped output across the decode token range including the single-token and zero-token cases.

## Submission Checklist

### DIFF
--- a/op_tests/test_moe_mxfp8_passthrough.py
+++ b/op_tests/test_moe_mxfp8_passthrough.py
@@ -5,15 +5,30 @@
 
 The passthrough lets a caller hand over activations that are already fp8 with
 group-32 e8m0 microscales, so fused_moe sorts the scale and skips
-requantization. Correctness means one thing: taking that shortcut must produce
-the same result as letting fused_moe quantize the activations itself.
+requantization. Its entire claim is that taking the shortcut changes nothing,
+and that is testable exactly: pre-quantize with the same HIP MX quantizer the
+internal path uses, and both sides feed identical fp8 bytes and identical scale
+values to the same GEMM. The only thing left varying is which code sorts the
+scale, so the two outputs must be *bit-identical* -- not merely close.
 
-This mirrors the SGLang MoRI a8w4 call pattern: PER_1X32 MXFP4 weights,
-GateMode.INTERLEAVE, and fp8 activations even at decode batch sizes (the
-production path sets AITER_BF16_FP8_MOE_BOUND=0). Without those knobs a
-standalone harness routes fp8 inputs into the mxfp4 quant branch and aborts
-inside the HIP launcher -- that is a test-setup failure, not evidence that the
-passthrough is broken in serving.
+Exactness matters here because the failure mode is silent. The scale is consumed
+bytewise, so a mis-sorted or mis-strided scale returns plausible-looking wrong
+numbers rather than raising, and any tolerance wide enough to absorb quantizer
+noise would also absorb a real defect. There is no quantizer noise to absorb.
+
+Scope, stated so it is not overclaimed: this pins the passthrough against the
+path it replaces. It is not an absolute-correctness test for the a8w4 MoE, which
+op_tests/test_moe_2stage.py and op_tests/flydsl_tests/test_flydsl_moe_a8w4.py
+already cover against a torch reference.
+
+The harness mirrors how SGLang's MoRI dispatch calls this: per_1x32 MXFP4 expert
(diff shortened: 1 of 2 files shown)


op_tests/test_moe_mxfp8_passthrough.py
```diff
@@ -0,0 +1,88 @@
...
+    """per_1x32 is the whole point. A group-128 layout would send the caller
+    back through the fp8->bf16 upscale this branch exists to avoid."""
+    assert MODEL_DIM % MX_BLOCK == 0
+    assert MODEL_DIM // MX_BLOCK == 224
+
+
+def test_e8m0_scale_is_one_byte():
+    """The dispatch buffer is sized from this. fp32 scales would need 4x the
+    room, and the payload would be truncated rather than rejected."""
+    assert torch.float8_e8m0fnu.itemsize == 1
```
valarLip: these tests looks nothing help...

valarLip (dismissed):

op_tests/test_moe_mxfp8_passthrough.py
karverma-amd: @valarLip  - You were right — the old assertions checked PyTorch facts and source text, not aiter behaviour. Replaced with a numerical one.

Pre-quantizing the activations with the same HIP MX quantizer the internal path uses removes the quantizer as a variable: both sides then feed identical fp8 bytes and identical e8m0 scales to the same GEMM, and the only thing still varying is which code sorts the scale. The outputs are bit-identical, so the test asserts `torch.equal` rather than a tolerance — any tolerance wide enough to absorb quantizer noise would also absorb a real defect.

```
$ python3 -m pytest op_tests/test_moe_mxfp8_passthrough.py -v

test_passthrough_is_bit_identical_to_internal_quantization[1]   PASSED [ 33%]
test_passthrough_is_bit_identical_to_internal_quantization[8]   PASSED [ 66%]
test_passthrough_is_bit_identical_to_internal_quantization[112] PASSED [100%]

======================== 3 passed, 2 warnings in 5.26s =========================
```

MI355X (gfx950), ROCm 7.2, torch 2.9.1+rocm7.2.

To show it discriminates rather than merely passes, I substituted the branch in three wrong forms that return bad numbers rather than raising: `no_sort` leaves the scale unsorted, `wrong_cols` sorts with `inter_dim` instead of `model_dim`, and `rolled` sorts a row-rotated scale (shape and dtype stay valid, only the numbers are wrong).

| | correct | no_sort | wrong_cols | rolled |
|---|---|---|---|---|
| tokens=1   | PASSED | FAILED | FAILED | FAILED |
| tokens=8   | PASSED | FAILED | FAILED | FAILED |
| tokens=112 | PASSED | FAILED | FAILED | FAILED |

Scope: this pins the passthrough against the path it replaces, not absolute a8w4 correctness, which `op_tests/test_moe_2stage.py` and `flydsl_tests/test_flydsl_moe_a8w4.py` already cover. Pushed as `da20e2649`.

yifehuan (approved):
LGTM

PR #4955, opened 2026-08-24
[Triton/Gluon] Force fp4gemm preshuffle to triton on gfx1250

zufayu:
## Summary

On gfx1250, `gemm_afp4wfp4_preshuffle` hard-selected the gluon kernel from the arch string:

```python
use_gluon = arch_info.get_arch() == "gfx1250"
```

That single arch check silently governs three separate things, because the gluon and triton
preshuffle kernels are not interchangeable:

| | gluon | triton |
|---|---|---|
| scale layout | `preshuffle_factor=16`, `scale_kwidth=4` | `32`, `8` |
| config params | `NUM_BUFFERS` | `NUM_KSPLIT`, `GROUP_SIZE_M`, `num_stages`, `matrix_instr_nonkdim`, ... |
| config dir | `configs/gfx1250/gluon/...` | `configs/gfx1250/triton/...` |

So "which kernel runs" also decides which config directory is valid and which scale layout
callers must produce. With the arch check, those three could not be moved together.

This PR routes gfx1250 fp4 preshuffle GEMM to the **triton** kernel, and replaces the arch
check with a single predicate that all three sites consult.

## Changes

1. **`aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py`** — arch check becomes a flippable predicate:
   ```python
   _USE_GLUON_PRESHUFFLE = False
   def use_gluon_preshuffle() -> bool:
       return _USE_GLUON_PRESHUFFLE and arch_info.get_arch() == "gfx1250"
   ```
   Default `False` — gfx1250 now runs the triton kernel; flipping one flag hands it back to gluon.

2. **Config resolution follows the backend, not the arch** — `_get_config(..., backend="gluon" if
   use_gluon else "triton")` threaded through to `get_gemm_config`. The `backend` parameter already
   exists in `aiter/ops/triton/utils/gemm_config_utils.py`; this only wires it up. Since the two
   backends take disjoint params, the config must come from the directory of the backend actually
   launched.

3. **Configs** — the existing gluon tuning moves to a new
   `configs/gfx1250/gluon/gemm/gemm_afp4wfp4_preshuffled/DEFAULT.json`, and the triton config is
   expanded from 2 buckets to 8 (`M_LEQ_8/31/32/64/128/256/512/any`) with per-bucket
   `num_stages` / `waves_per_eu` / `matrix_instr_nonkdim` / `NUM_KSPLIT`.

4. **`op_tests/triton_tests/gemm/basic/test_gemm_afp4wfp4.py`** — scale shuffling now keys off
   `use_gluon_preshuffle()` instead of `DEVICE_ARCH == "gfx1250"`. Without this the test would build
   scales in the gluon layout while the triton kernel runs, i.e. feed the wrong layout silently.

## Notes for reviewers

- The behavior change is limited to gfx1250 fp4 preshuffle GEMM; other arches are untouched.
- `use_gluon_preshuffle()` is exported deliberately: any caller that shuffles scales itself must
  pick the layout from this predicate rather than from the arch.
- Authored by @valarLip on branch `gfx1250/test_main_823`; I am taking the branch over and opening
  the PR for review.
- The branch is currently a few commits behind `main` — happy to rebase if preferred.
- gfx1250 accuracy/performance numbers to be attached from a CI run on gfx1250 hardware.

### DIFF
--- a/aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py
+++ b/aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py
@@ -28,10 +28,23 @@ _LOGGER = AiterTritonLogger()
 
 _USE_GEMM_SPLITK_BF16 = False
 
+# gfx1250 runs the triton preshuffle kernel; flip this to hand it back to gluon.
+_USE_GLUON_PRESHUFFLE = False
+
 
 def set_use_gemm_splitk_bf16(value: bool):
     global _USE_GEMM_SPLITK_BF16
     _USE_GEMM_SPLITK_BF16 = value
+
+
+def use_gluon_preshuffle() -> bool:
+    """Whether gemm_afp4wfp4_preshuffle dispatches to the gluon kernel.
+
+    The two kernels consume different scale layouts (gluon: preshuffle_factor
+    16 / scale_kwidth 4, triton: 32 / 8), so callers that shuffle scales
+    themselves must pick the layout according to this predicate.
+    """
+    return _USE_GLUON_PRESHUFFLE and arch_info.get_arch() == "gfx1250"
 
 
 def get_splitk(K: int, BLOCK_SIZE_K: int, NUM_KSPLIT: int):
@@ -449,7 +462,7 @@ def gemm_afp4wfp4_preshuffle(
     """
 
     assert arch_info.is_fp4_avail(), "MXFP4 is not available on your device"
-    use_gluon = arch_info.get_arch() == "gfx1250"
+    use_gluon = use_gluon_preshuffle()
 
     M, K_bytes = x_fp4.shape
     n16, _ = w_preshuf.shape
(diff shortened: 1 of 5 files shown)


zufayu:
Scope note for reviewers: **this is a workaround, not the final fix.**

DSR currently cannot run on the gluon fp4 preshuffle path on gfx1250. Forcing the wrapper to the
triton kernel is what unblocks it. The proper fix is expected from the Triton team, and this PR is
deliberately shaped so that switching back costs one line:

```python
_USE_GLUON_PRESHUFFLE = False   # flip to True once the gluon path is fixed
```

The gluon tuning is not deleted either — it moves intact to
`configs/gfx1250/gluon/gemm/gemm_afp4wfp4_preshuffled/DEFAULT.json`, so the gluon path stays fully
restorable.

Please review with that scope in mind: the goal is to unblock DSR on gfx1250 now, and to revert to
gluon once the official Triton-side solution lands.

aiter/ops/triton/gemm/basic/gemm_afp4wfp4.py
```diff
@@ -28,12 +28,25 @@
 
 _USE_GEMM_SPLITK_BF16 = False
 
+# gfx1250 runs the triton preshuffle kernel; flip this to hand it back to gluon.
+_USE_GLUON_PRESHUFFLE = False
```
valarLip: replace by ENV?

zufayu: Done in 48834df — good call, this needs to be flippable without a rebuild.

```python
# gfx1250 runs the triton preshuffle kernel by default; set
# AITER_FORCE_GFX1250_GLUON_FP4=1 to force the gluon path back on. This is a
# workaround switch kept for validation until the Triton-side fix lands.
_USE_GLUON_PRESHUFFLE = os.environ.get("AITER_FORCE_GFX1250_GLUON_FP4", "0") == "1"
```

Naming and read-at-import follow `AITER_FORCE_GFX1250_EX` in `gemm_a8w8_blockscale.py`.

Default is **0 = triton**, deliberately: the point of the PR is to unblock DSR, which cannot run on
the gluon path, so the workaround has to be active out of the box. `FORCE` is the opt-in override
for validating gluon.

`use_gluon_preshuffle()` is unchanged and keeps the `arch == "gfx1250"` guard, so the env var stays
a no-op on every other arch, and dispatch / config dir / test scale layout still resolve through
the one predicate.

junhaha666 (approved):

PR #4320, opened 2026-07-21
Add opus fp8 mxscale BMM kernels for gfx950

yzhou103:
## Motivation

Adds the opus fp8 **e8m0 mxscale (block-scale) batched GEMM (BMM)** path for
gfx950, exposed as `aiter.batched_gemm_a8w8_mxscale`. The per-kid device
kernels and host launchers are **fully code-generated** (mirroring the opus_gemm
module), and dispatch is driven by a tuned CSV with padded-M / M-split /
heuristic fallbacks. Also includes a round of dead-kernel and naming cleanup.

```python
# mmajor DSV4 layout: x [M, G, K] fp8, wo_a [G, N, K] fp8 (batch-major, zero-copy),
# x_scale [M, G, K/128], w_scale [G, N/128, K/128], out optional [M, G, N].
aiter.batched_gemm_a8w8_mxscale(x, wo_a, x_scale, w_scale, dtype=torch.bfloat16)
```
## Technical Details

Reuses the shared opus GEMM gfx950 backend; every kid is codegen'd into its own <tile>_C{void,bf16,fp32}.device.cu (device symbol) + inlined host launcher, no hand-written adapters.
Table-driven kid -> launcher dispatch (opus_bmm_a8w8_mxscale_tune_dispatch) instead of a large switch.
Kernel families: fused, pipeline, minterleave, mouter / mouter_tunable, wave8n2, wave4m2_selfload, plus OOB-masked sub-tiles (311/313/320/321/324/640/650/653).

Dispatch logic (kernelId=None)
Exact tuned CSV hit → use its kid + splitK.
Padded-M: round M up to the nearest tuned bucket in the same (g,n,k) family; accepted only for OOB-masked sub-tiles (run the real smaller M with no pad/copy).
M-split: for large tile-unaligned M, split into an aligned strong-tile bulk (157/150) + a <256-row OOB-safe tail. Gated on total bulk work g*m >= 8192 (measured break-even), not on g alone.
Heuristic single-kid fallback otherwise.

## Test Plan

op_tests/test_opus_a8w8_bmm.py: correctness (fp8 vs bf16 einsum, err 0) across g∈{2,8}, m=1..16384, both m-major and batch-first layouts, and per-family kids.

## Test Result

together with https://github.com/ROCm/aiter/pull/4428

gfx950 x8, DeepSeek-V4-Pro (`kv_cache_dtype=fp8`, `-tp 8`, expert parallel).
[...]

PR #4961, opened 2026-08-24
[HIP] [OPUS] Unify OPUS GEMM/BMM interfaces and use Torch workspaces

Fyzyukk:
## Summary

This PR refactors the OPUS GEMM/BMM around one strict, exact-kernel-id contract from Python through generated C++ dispatch.

- Add public `opus_gemm()` and `opus_bmm()` entry points,
- Separate kernel selection from exact execution,
- Use per-call Torch workspaces for split-K; C++ only validates and uses them,
- Keep existing non-OPUS backend behavior intact.

## Motivation

The previous path mixed kernel selection, validation, dispatch, and workspace management across Python and C++, resulting in duplicated wrappers, heuristic dispatch, and global workspace state.

This PR defines a strict exact-kid boundary: callers select the kernel and own the output/workspace tensors, while OPUS C++ validates and launches that exact kernel.

### A16W16

```text
GEMM: TunedGemm / gemm_a16w16()
  -> tuned-row validation and fallback policy
  -> final kid + split_k
  -> opus_gemm()                         [2D]

BMM: caller-selected kid + split_k
  -> opus_bmm()                          [batch-first 3D]

opus_gemm() / opus_bmm()
  -> canonical a16w16 registry lookup
  -> _execute_a16w16()
  -> A16W16LaunchPlan
       - validate exact kid, shape, dtype, bias, and split-K
       - reuse caller workspace or allocate a per-call Torch workspace
  -> _launch_a16w16_backend()
       - eager / graph capture: cached C ABI
       - torch.compile / Meta / FakeTensor: torch.ops boundary
  -> opus_gemm_a16w16_launch()
  -> per-architecture exact-kid table
  -> generated launcher
  -> direct kernel or split-K kernel + reducer
```

### A8W8

```text
aiter/ops/gemm_op_a8w8.py
  -> gemm_a8w8()                         [no-scale GEMM]
  -> gemm_a8w8_blockscale()              [plain block-scale GEMM]
  -> gemm_a8w8_blockscale_bpreshuffle()  [B-preshuffle GEMM]

aiter/ops/batched_gemm_op_a8w8.py
  -> batched_gemm_a8w8_mxscale()         [MXFP8 BMM]
             |
             v
fixed, tuned, or caller-resolved final kid
  -> opus_gemm()                         [three GEMM families]
  -> checked raw fast path               [MXFP8 BMM, split_k <= 1]
  -> opus_bmm(layout="mxscale_bmm")      [MXFP8 BMM, split_k > 1]
             |
             v
canonical A8 family registry lookup and contract validation
  -> family adapter
  -> MXFP8 split-K only: build plan and reuse/allocate FP32 Torch workspace
  -> _launch_a8w8_backend()               [pybind]
  -> family C++ entry
  -> exact-kid table
  -> generated launcher
  -> exact kernel; MXFP8 split-K also launches the reducer
```

## Production paths

| Family | High-level or public path | C++ exact entry | Workspace | Current capability |
|---|---|---|---|---|
| A16W16 GEMM | `TunedGemm.mm` / `gemm_a16w16()` -> `opus_gemm()` -> `_execute_a16w16()` | `opus_gemm_a16w16_launch()` | Optional, Torch-owned for workspace kids | gfx942, gfx950, gfx1250 |
| A16W16 BMM | direct `opus_bmm()` -> `_launch_a16w16_bmm()` -> `_execute_a16w16()` | same A16 3D exact launcher | Optional, batch-aware Torch workspace | Exact API only; there is no new high-level BF16 BMM dispatcher |
| A8W8 no-scale GEMM | `gemm_a8w8()` -> `opus_gemm()` -> A8 backend | `opus_gemm_a8w8_launch()` | None | gfx950 kid 2, FP32 output |
| A8W8 plain block-scale GEMM | `gemm_a8w8_blockscale()` -> `opus_gemm()` | `opus_gemm_a8w8_blockscale_launch()` | None | gfx950 kid 1, FP32 output |
| A8W8 block-scale B-preshuffle GEMM | tuned `libtype=opus` row -> `opus_gemm(layout="bpreshuffle")` | `opus_gemm_a8w8_blockscale_bpreshuffle_launch()` | None | gfx942 kid 11000, BF16 output |
| A8W8 MXFP8 BMM | `batched_gemm_a8w8_mxscale()` -> split-1 checked raw path or `opus_bmm(layout="mxscale_bmm")` | `opus_gemm_a8w8_mxscale_bmm_launch()` | FP32 Torch workspace when split-K requires it | gfx950, 45 global ids, BF16/FP32 output |

## Workspace ownership and execution

- Split-K workspace is either supplied by the caller or allocated by Torch for that call. Python validates it, and C++ only uses its pointer during the launch.
- Direct kernels reject workspace; workspace kernels require it.
- A16 uses pybind for initial JIT loading. Later eager and graph-capture calls use the cached C ABI, while `torch.compile`, MetaTensor, and FakeTensor keep the registered `torch.ops` path.

Current A16 two-stage workspace layouts are:

| Architecture | Logical workspace shape | Storage |
|---|---|---|
| gfx950 | `[split_capacity, batch, padded_M, padded_N]` | FP32 |
| gfx942 | `[split_capacity, batch, padded_M, padded_N]` | exact BF16/FP32 instance dtype |
| gfx1250 | `[split_capacity, padded_M, padded_N]` | BF16; current workspace kernels require batch 1 |

The experimental gfx1250 fused split-K source remains available for repair, but it is not registered and cannot be selected by the public API.

## Tuning

| Path | OPUS tuning behavior |
|---|---|
| A16W16 GEMM | Full multi-kid/multi-split tuning through `csrc/gemm_a16w16/gemm_a16w16_tune.py --libtype opus` |
| A8W8 no-scale GEMM | Fixed kid 2; no OPUS offline selection |
| A8W8 plain block-scale GEMM | Fixed kid 1; no OPUS offline selection |
| A8W8 B-preshuffle GEMM | OPUS candidate generation for gfx942 only; current candidate is kid 11000 |
| A8W8 MXFP8 BMM | Dedicated gfx950 tuner covering applicable kids and split-K values 1/2/4/8 |

## Performance

| Comparison on gfx950 | Result |
|---|---:|
| A16W16 interface| -24.051% |
| Executable A8W8 no-scale + FP32 block-scale | -0.062% (flat) |
| Torch vs implicit stream workspace, all 48 workspace kids, eager | -17.801% |
| Torch vs implicit stream workspace, all 48 workspace kids, graph replay | -10.618% |
| Public router vs private family adapter, eager | +1.028% |
| Public router vs private family adapter, graph replay | +0.005% |

All 96 eager and 96 graph cases improved. Matching device show that the gains came from Python/C ABI dispatch, not kernel changes.

## Guide

- `aiter/ops/opus/__init__.py`: strict public GEMM/BMM router and contracts
- `aiter/ops/opus/policy.py`: high-level candidate and tuned-row policy
- `aiter/ops/opus/launch_plan.py`: immutable split-K/workspace planning
- `aiter/ops/opus/gemm_op_a16w16.py`: A16 executor, Torch workspace, and C ABI
- `aiter/ops/opus/gemm_op_a8w8.py`: A8 family adapters and pybind backend
- `csrc/opus_gemm/opus_gemm.cu`: A16/A8 GEMM family entries and exact dispatch
- `csrc/opus_gemm/opus_bmm.cu`: MXFP8 BMM exact dispatch
- `csrc/opus_gemm/gen_instances.py` and `codegen/gen_instances_gfx*.py`:
  generated manifests, subset selection, and typed exact-kid tables
- `csrc/opus_gemm/opus_gemm_common.py`: canonical registry and family metadata


csrc/ck_gemm_a8w8_blockscale/gemm_a8w8_blockscale_tune.py
```diff
@@ -164,8 +162,14 @@ def run_gemm_a8w8_blockscale_opus(
     """
     Run gfx942 Opus a8w8 blockscale bpreshuffle tuned kernel.
     """
-    return opus_gemm_a8w8_blockscale_bpreshuffle_tune(
-        x, weight, x_scale, w_scale, out, kernelId=kernel_id
+    return opus_gemm(
```
demonsan: we should not touch ck part. you can add standalone tune script in opus

Fyzyukk: csrc/ck_gemm_a8w8_blockscale/gemm_a8w8_blockscale_tune.py to the base version, so this PR no longer changes CK-owned code.

demonsan:
## Review Summary — ⚠️ NEEDS WORK

**Scope**: OPUS kernel workspace management & interface refactor. 63 files, ~310 counting CO binaries + main merge. Adds exact kid-based dispatch, removes 3 C++ heuristic dispatch headers, adds `launch_plan.py` + `policy.py`, integrates 219 gfx1250 pre-built CO kernels.

**Test results (gfx950)**: Config shape collision ✅ 16/16 · A16W16 GEMM ✅ 8 passed/6 skipped · A8W8 BMM ✅ guards + sweep

---

### Must-Fix (6)

**🔴 M1 — `__init__.py` contains ~250 lines of function implementations**

`aiter/ops/opus/__init__.py` grew from 82 to 347 lines (4.2×). Three functions — `_validate_a16w16_public_contract` (29 lines), `_resolve_contract` (50 lines), `_opus_dispatch` (155 lines) — embed validation, contract resolution, and family dispatch in the init file. These belong in dedicated modules (e.g. `dispatch.py`, `contract.py`). Only the three 1-line delegates (`opus_gemm`, `opus_bmm`, `gemm_a16w16_opus`) belong here.

**🔴 M2 — OPUS code leaked into `gemm_op_a8w8.py` (~105 lines)**

Two dtype-driven dispatch paths embed OPUS routing in the central dispatcher:
- No-scale `gemm_a8w8` branch: triggers on `x_scale is None`, changes function signature (`x_scale`/`w_scale` → `Optional`), breaking existing API
- FP32 `gemm_a8w8_blockscale` branch: triggers on `dtype == fp32`, couples OPUS to dtype-based dispatch

These should be separate OPUS entry points (e.g. `opus.gemm_a8w8_noscale`). Additionally, `_OPUS_A8W8_NOSCALE_KID` and `_OPUS_A8W8_BLOCKSCALE_KID` are OPUS kernel IDs hardcoded in a non-OPUS file — they belong in `aiter.ops.opus`.

**🔴 M3 — 11 redundant `getattr` calls on always-present fields**

`arch_prefix`, `splitk_workspace_dtype`, `fuse_split_k`, `fuse_m_cluster`, `direct_only` (dataclass fields on `OpusGemmInstance`), `gcnArchName` (on `torch.cuda.get_device_properties`), `__file__` (on Python modules), `torch.compiler.is_compiling`. These fields are always present; `getattr` with defaults silently swallows `AttributeError`.

Files: `_arch.py`, `gemm_op_a16w16.py`, `codegen/common.py`, `gen_instances_gfx1250.py`, `opus_gemm_tune.py` (×2), `opus_bmm_mxscale_tune.py`, `test_opus_a16w16_gemm.py`

**🔴 M4 — Unicode character in code**

`aiter/ops/batched_gemm_op_a8w8.py`: em dash `—` in a logger.warning string. Comments/strings should be ASCII.

**🔴 M5 — `develop=True` on 4 `@compile_ops` bindings**

`aiter/ops/opus/gemm_op_a8w8.py` lines 52, 77, 104, 159. Bypasses JIT cache — every restart recompiles. Remove before merge.

**🔴 M6 — No GPU-side correctness test for gfx1250 CO kernels**

`test_opus_co_integration.py` is CPU-only — validates registry, assets, codegen but never launches a CO kernel on GPU. 219 kernels, zero numerical validation. A `.co` binary with a tile config bug that passes all structural checks would not be caught.

---

### Should-Fix (7)

**⚠️ S1 — Downstream CI skipped** (E4)

`tuned_gemm.py` and `batched_gemm_op_a8w8.py` changed — consumed by ATOM/SGLang/vLLM. Add `ci:all` and require green.

**⚠️ S2 — Breaking API rename needs owner sign-off** (E5)

`bmm_a8w8_mxscale_opus` and `opus_gemm_workspace_init/release/release_all` are removed. `gemm_a16w16_opus` was restored as compat shim ✅. Top committers: @Fyzyukk (14), @demonsan (10). Requesting explicit sign-off.

**⚠️ S3 — Arch string literals in new dispatch code** (C4)

`policy.py` and `launch_plan.py` use raw `"gfx942"`/`"gfx950"`/`"gfx1250"` comparisons. `co_kernels.json` uses `"gfx1250"` as dispatch key. Extract to shared constant or registry.

**⚠️ S4 — Perf-claim commits without benchmark numbers** (P1)

"Optimize A16 eager dispatch", "Optimize A16 C ABI launch path", "Optimize Task2 Python dispatch hot paths". Add numbers with units or rephrase.

**⚠️ S5 — A8W8 GEMM families have zero test coverage**

`_launch_a8w8_gemm`, `_launch_a8w8_blockscale_gemm`, `_launch_a8w8_blockscale_bpreshuffle_gemm` — no test. `opus_gemm` for `a8w8`/`a8w8_blockscale`/`a8w8_blockscale_bpreshuffle` families never called by any test.

**⚠️ S6 — 220 `.co` binary files + 9.6MB JSON committed**

Opaque binaries unreviewable in diff. ~14MB repo weight. Document regeneration procedure; add CI validation that `.co` files match source.

**⚠️ S7 — `_resolve_contract` lazy imports surface errors at call time**

`launch_plan` and `gemm_op_a8w8` imported inside function body. If either fails to import, error at call time not import time. Consider module-level import.

---

### Notes (5)

- **📝 Abstraction layers growing**: `common.py` (2058 lines) → `launch_plan.py` (666 lines) → `policy.py` (861 lines). Three Python layers on C++ dispatch.
- **📝 No motivation (WHY) for the core refactor**: 30+ commits, 63 files, but no explanation of WHY this restructuring is needed.
- **📝 `gemm_a16w16_opus` restored** ✅ — backward-compat shim addresses the deletion concern.
- **📝 CO path ships with documented issues**: `gen_co/KNOWN_ISSUES.md` (385 lines) — good practice, but issues should be resolved before production.
- **📝 All 25 test functions pass unit-test admission audit**: no mirror/tautology tests found.

---

### What's Clean

- API propagation (all aiter-internal callers updated) ✅
- Fake functions (every `@compile_ops` has `gen_fake`) ✅
- No hipblaslt, temp scripts, env vars, sys.path mutations ✅
- Stream sync: `OpusCabiDeviceStreamGuard` properly manages HIP stream state ✅
- No atomics on uninitialized buffers; contiguous checks present ✅
- Comment style improvements throughout ✅

Fyzyukk:
### M5 

  I rechecked the implementation and kept the four `develop=True` arguments.

  In this tree, `compile_ops(..., develop=True)` does not control rebuilding or bypass the JIT cache. Module loading still goes through the same `get_module()` path, while `AITER_REBUILD` controls rebuild behavior.

  The `develop` branch is the pybind call adapter that converts `torch.Tensor` arguments to the Torch-free `aiter_tensor_t` ABI and propagates the current HIP stream. These four C++ entry points accept `aiter_tensor_t`, so deleting the flag alone would break both the ABI conversion and stream propagation.

  If removing this adapter is required, it should be handled as a separate C-ABI migration with non-default-stream and graph-capture coverage, rather than as a four-line flag removal.

### N1

  To clarify, these modules are not three serial dispatch layers.

  `policy.py` is used only by shape-driven callers to select the final kid and split-K. Explicit `opus_gemm` and `opus_bmm` calls bypass it. `dispatch.py` validates and routes the public exact-kid contract, while `launch_plan.py` derives immutable split-K and workspace metadata. `opus_gemm_common.py` is the canonical registry
  shared by runtime and code generation rather than another dispatcher.

  The GEMM and MoE family modules remain flat under `aiter/ops/opus/`, and we will keep these responsibility boundaries stable without adding another dispatch layer.

  ### N3

  Confirmed. `gemm_a16w16_opus` remains available as a compatibility entry with its existing import path and signature. Its shape-driven tuned/heuristic behavior is preserved and remains covered by the compatibility tests.
  
   ### N5

  Acknowledged. The focused tests remain behavior-based: they exercise the public API, the real family adapter, the checked raw-binding boundary, and independent Torch references where the required GPU architecture is available.

  The current focused suite passes on the available environment. The gfx1250-only numerical cases are correctly gated and their real-hardware execution evidence remains tracked separately under M6.

aiter/ops/batched_gemm_op_a8w8.py
```diff
@@ -261,40 +178,53 @@ def _batched_gemm_a8w8_mxscale_impl(
...
+    # tensors carry concrete integer dimensions here.  Avoid four redundant
+    # Python int() conversions on every short BMM launch.
+    m, g, k = x.shape
+    n = wo_a.shape[1]
+    plan_key = (g, m, n, k)
+    try:
+        raw_launch, opus_bmm, kid, split_k = _MXSCALE_BMM_LAUNCH_PLANS[plan_key]
+    except KeyError:
+        raw_launch, opus_bmm = _get_mxscale_bmm_launchers()
+        kid, split_k = _resolve_a8w8_mxscale_bmm_plan(g, m, n, k)
```
yzhou103: Unbounded cache → memory leak. _MXSCALE_BMM_LAUNCH_PLANS in batched_gemm_op_a8w8.py is a plain dict keyed by (g,m,n,k), replacing an lru_cache(maxsize=1024). In decode, m is a ragged token count, so it grows forever. Use lru_cache. It also stores
  the same two function objects in every entry — those belong in a no-arg @functools.cache.

aiter/ops/opus/policy.py
```diff
@@ -0,0 +1,738 @@
...
+
+    selected_opus_rows = (
+        df["libtype"].eq("opus")
+        if "libtype" in df.columns
+        else pd.Series(True, index=df.index, dtype=bool)
+    )
+    invalid_opus_kids = selected_opus_rows & ~df["kernelId"].between(
+        _MXSCALE_BMM_KID_OFFSET, _MXSCALE_BMM_GLOBAL_KID_MAX
+    )
+    if invalid_opus_kids.any():
```
yzhou103: why here invalid keys raise error but in _load_a16w16_opus_tuned it will go heuristic？

aiter/ops/batched_gemm_op_a8w8.py
Fyzyukk: Fixed. The shape-dependent launch plan now uses @functools.lru_cache(maxsize=1024) and caches only (kid, split_k). The launcher function objects remain in the separate no-argument @functools.cache. Added a regression test for the cache bound.

aiter/ops/opus/policy.py
Fyzyukk: This strict failure was introduced while consolidating the MXFP8 BMM selection logic into policy.py in this PR; I restored the behavior from PR #4320: invalid or unregistered OPUS tuned rows are warned about and skipped, after which the caller falls back to the heuristic.

csrc/include/opus/opus.hpp
```diff
@@ -3202,18 +3202,21 @@ using mfma_scale_f32_16x16x128_fp4_fp4  = mfma_f32_16x16x128_fp4_fp4;
 /////////////////////////////////////////////////////////////////////////////////////////////////////////
 // wmma (RDNA4 / wave32) -- gfx1250 uses wmma-256b builtins (16x16x{4,32,64,128}); gfx1200/gfx1201 (Navi 44/48) use wmma-128b _w32_gfx12 builtins (16x16x16). Dispatch macros for the two arg-list shapes differ -- gfx12 set is DISPATCH_WMMA_GFX12_*.
 #if defined(__gfx1250__) || defined(__gfx1201__) || defined(__gfx1200__) || !defined(__HIP_DEVICE_COMPILE__)
-// f16/bf16/f32 builtins: (a_neg, A, b_neg, B, c_mod, C, matrix_a_reuse, matrix_b_reuse)
+// f16/bf16/f32 builtins: (a_neg, A, b_neg, B, c_mod, C,
```
demonsan: Should not touch opus.hpp. We need to keep this file single line and condensed

aiter/ops/deepgemm.py
```diff
@@ -52,12 +52,22 @@ def opus_gemm_a16w16_tune(
 ) -> torch.Tensor:
     warnings.warn(
         "aiter.ops.deepgemm.opus_gemm_a16w16_tune has moved to "
-        "aiter.ops.opus.gemm_op_a16w16.opus_gemm_a16w16_tune; this "
+        "aiter.ops.opus.opus_gemm; this "
```
demonsan: this file mainly for ck gemm. opus gemm can be added to different gemm op.

yzhou103:
This PR deletes three C++ heuristic dispatch headers and reimplements them in Python, but ships no parity coverage. These are pure (M, N, K, has_bias) functions, so a CPU-only sweep + diff against the pre-PR headers is cheap — please add one
 
  Two things make this more than a formality:
 
  1. GFX942_BF16WS_EXACT_N and the 10210→10200 / 10213→10203 / 10216→drop redirects in _resolve_a16w16_candidate are new — they don't exist before this PR. Please explain the rationale and the blast radius separately from the port.
  2. The tuned-row lookup also changed source and key (codegen'd opus_gemm_lookup.h → runtime AITER_CONFIG_GEMM_BF16_FILE), and _resolve_a16w16_candidate now silently drops a tuned row whose plan fails validation, falling back to the heuristic. So 
  end-to-end kid selection can change even if the heuristics are byte-identical. Please diff "shape → final (kid, splitK)" on main vs this branch over the real tuned CSV plus a set of untuned shapes, and justify every difference.

csrc/include/opus/opus.hpp
Fyzyukk: This PR no longer changes csrc/include/opus/opus.hpp, so the existing condensed single-line formatting is preserved.

aiter/ops/deepgemm.py
Fyzyukk: I reverted the changes in aiter/ops/deepgemm.py, so this PR no longer modifies the CK-facing file. The legacy OPUS compatibility implementation now lives in aiter/ops/opus/gemm_op_a16w16.py.

Fyzyukk:
@yzhou103 
1. All three removed heuristic headers are A16W16-only. I added test_python_heuristics_match_pre_pr_cpp_boundary_sweep() in op_tests/test_opus_a16w16_policy_parity.py. It compares the new Python selectors with frozen references transcribed from the three pre-PR headers across 657,072 boundary cases covering gfx942/gfx950/gfx1250, both has_bias values, and BF16/FP32 outputs. All cases match.
2. 
(1) The Python constant and pre-launch redirects are new, but they make existing generated-launcher behavior explicit rather than introduce new kernel behavior. 
(2) On current upstream/main, the generated gfx942 launchers support the BF16-workspace reducer only when N is one of {64,128,256,384,512,1024,2048}. For other values, 10210 already redirects to its FP32-workspace sibling 10200, 10213 redirects to 10203, and 10216 raises AITER_CHECK because it has no FP32-workspace sibling. “FP32 workspace” refers only to the temporary split-K buffer; the operation remains A16W16 and can still produce BF16 output. This resolution now happens before launch because Python/Torch must allocate the workspace using the final kid’s dtype and capacity.
(3) The blast radius is limited to gfx942 A16W16 kids 10210/10213/10216 with non-exact N; other kids and architectures are unaffected. Also, the heuristic never selects 10216—it can only come from a tuned row or an explicit kid. 
(4) I added test_gfx942_non_exact_n_matches_pre_pr_generated_launcher() and test_gfx942_exact_n_keeps_each_bf16_workspace_kid() in op_tests/test_opus_a16w16_policy_parity.py. The shipped configs contain 31 rows using these BF16-workspace kids, and all of them use supported exact N values, so no shipped tuned row is redirected or rejected.
3. 
(1) I refreshed upstream/main to d06a5f38. The relevant lookup code, headers, and tuned configs are unchanged from the PR merge base. The comparison is covered by op_tests/test_opus_a16w16_policy_parity.py.
(2) Upstream/main already has two lookup stages: aiter/ops/opus/common.py first reads the runtime CSV using a key without gfx; only a miss reaches the architecture-specific generated opus_gemm_lookup.h.
(3) This branch uses AITER_CONFIGS.AITER_CONFIG_GEMM_BF16_FILE because the final kid must be resolved before Python/Torch allocates its workspace. This is the canonical merged runtime config already used by tuned_gemm. The generated header is only a build-time > snapshot and does not retain the CSV splitK.
(4) test_shipped_tuned_selection_diff_is_exhaustive() compares all 470 shipped OPUS rows, including all 208 rows with non-zero splitK. Every branch candidate validates, no shipped row is redirected or rejected, and 463 (kid, splitK) selections are unchanged. > The only differences are:
gfx950 (M,N,K)      upstream/main  this branch
(256,1536,7168)         (20399,5)         (1214,4)
(1024,512,7168)         (20400,3)         (1208,2)
(64,8448,7168)           (20341,3)         (1216,2)
(8192,1024,4096)       (21177,0)         (1401,0)
(16384,512,4096)       (21177,0)         (1401,0)
(8192,1024,7168)       (21177,0)         (1401,0)
(16384,512,7168)       (21177,0)         (1401,0)
(5) All seven differences fix the same upstream/main bug: gfx950 and gfx1250 both use cu_num=256, but the old Python key omits gfx. Those gfx950 requests therefore select gfx1250-only kids and fail exact dispatch. Adding gfx selects the valid gfx950 rows, matching the architecture-specific C++ lookup. The seven expected changes are explicitly allowlisted.
(6) test_untuned_shape_final_selection_matches_pre_pr() covers 16 shapes absent from the shipped configs across gfx942, gfx950, and gfx1250. Their final (kid, splitK) selections all match upstream/main.
(7) Invalid tuned rows are also no longer silent. _resolve_a16w16_candidate() returns None without choosing a heuristic; the global tuned_gemm path now warns with the gfx, shape, kid, and splitK, then tries the next padded row or its existing default backend. > test_global_a16_stale_opus_row_keeps_framework_fallback() verifies the warning and fallback, while test_shape_driven_opus_selection_and_rank_route() verifies that a present invalid row is not silently replaced by an OPUS heuristic.

csrc/opus_gemm/include/opus_gemm.h
```diff
@@ -2,45 +2,45 @@
...
+                             aiter_tensor_t& WQ,
+                             aiter_tensor_t& Y,
+                             std::optional<aiter_tensor_t> bias,
+                             std::optional<aiter_tensor_t> workspace,
+                             int kid,
+                             int split_k);
 
-void opus_gemm_a16w16_tune(aiter_tensor_t& XQ,
+// Production A16W16 C ABI. Optional tensors use nullptr; stream is caller-owned.
+AITER_C_ITFS int opus_gemm_a16w16_launch_cabi(aiter_tensor_t* XQ,
```
demonsan: why we need this? I think keeping 1 launch api  is enough

Fyzyukk: Fixed. A16W16 now keeps a single pybind launch path.

op_tests/test_opus_a16w16_policy_parity.py
```diff
@@ -0,0 +1,460 @@
...
+from csrc.opus_gemm.opus_gemm_common import (
+    GFX942_BF16WS_EXACT_N,
+    get_kernel_instance,
+)
+
+# These are literal control-flow translations of the three headers at the
+# pre-PR merge base. Keep this reference independent of policy.py.
+_PRE_PR_REF = "ded4e3e8eee11f56853054c4ed4bdf2790545e5d"
+_PRE_PR_HEADERS = {
+    "gfx950": "4c8f03542e4459b51a17c0bd9fe224533af0c594",
```
demonsan: why we need these sha?

csrc/opus_gemm/opus_gemm_common.py
```diff
@@ -496,218 +433,156 @@ def _a8w8_mxscale_bmm_flatmm_splitk(
...
     for kid, (bm, bn, bk, wg) in _BMM_MXSCALE_SPLITK_PRELOAD_TILES.items()
 })
 
+# ROCm 7.2.4 clang-22 assigns an illegal register class while compiling this
+# exact high-pressure PRELOAD_SF_LDS + D_OUT=void specialization after the
+# workspace kargs moved to a direct pointer.  Its splitK=1 BF16/FP32 kernels
+# keep PRELOAD_SF_LDS; only the split-K workspace specialization uses the
+# semantically equivalent non-preload implementation (the same geometry as
+# local kid 139).
+_bmm_flatmm_local[326].workspace_preload_sf = False
```
yzhou103: workspace_preload_sf = False on kid 326 — could the cost be recorded?
   The comment does a good job explaining why the override exists (clang-22 illegal register class on the PRELOAD_SF_LDS + D_OUT=void specialization). Two small asks:
   1. "semantically equivalent" understates the perf delta. The comment notes the fallback has the same geometry as local kid 139 — which let me measure the cost directly, since after stripping the _sfpreload suffix the two generated launchers differ 
   in exactly one line:
  
   ```
<<   constexpr bool PRELOAD_SF_LDS = true;      // 326
   ---
   >>   constexpr bool PRELOAD_SF_LDS = false;     // 139
```
  
   Both resolve to the same Traits, so kid 139 today is a faithful stand-in for kid 326's post-PR split-K path. On gfx950:
  
   | G   | M   | N    | K    | splitK | Δ throughput |
   |-----|-----|------|------|--------|--------------|
   | 4   | 512 | 1024 | 4096 | 2      | −13.8%       |
   | 4   | 128 | 1024 | 4096 | 2      | −16.4%       |
  
   (min of 5; the node was contended, so treat the magnitude as indicative — the sign is stable.) Worth stating in the comment that this is a real throughput loss on the split-K path, not a free substitution.

File: aiter/ops/opus/launch_plan.py [hunk 9]
```diff
@@ -0,0 +1,668 @@
...
+        raise ValueError(
+            "opus_gemm_a16w16_launch: workspace capacity split_k must be "
+            f"positive, got {split_k}"
+        )
+
+    block_m = int(instance.B_M)
+    block_n = int(instance.B_N)
+    block_k = int(instance.B_K)
+    max_useful_split_k = (K + block_k - 1) // block_k
+    if split_k > max_useful_split_k:
```
yifehuan: Could we distinguish the workspace capacity from the effective split-K count in this check?

  For example, on gfx942 with BF16, (M, N, K) = (1, 64, 128), kid=10201, and split_k=0, _plan_gfx942_split_k() returns (workspace_capacity=16, abi_split_k=1). This check then compares the capacity 16 against ceil(K / B_K) = 2 and raises ValueError, even though the
  effective split-K count has already been reduced to 1. Explicit split_k=1 passes planning.

  Reserving more workspace than the launch needs should be safe here; the C++ validation also accepts workspace.numel() >= required_numel. Would it make sense to size the workspace using the converged split-K count, or allow this capacity overestimate while validating
  the effective split-K count separately? A regression test covering automatic split-K with a short K would help keep these two steps consistent.

Fyzyukk: The K-tile limit was incorrectly applied to workspace capacity. I moved that check into launch planning and size the gfx942 workspace from the converged split-K count. For your example, launch split-K stays 1, and planned capacity now matches it (16 slices → 1).
Verified on gfx942: auto allocation and caller workspaces with 1 or 16 slices all pass the Torch check.

op_tests/test_opus_gfx1250_splitk.py
```diff
@@ -0,0 +1,231 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+"""gfx1250 reducer-grid admission and FP32 workspace integration."""
```
demonsan: test opus_a16w16 has already covered splitk test.  This file is duplicated. Also, hardcoded kid will easily failed. Remove this test will be better

Fyzyukk: Already fixed.

demonsan:
gfx942 BF16-workspace exact-N guard redirects on one entry point and hard-raises on another — Problem: aiter/ops/opus/policy.py:419-425 (_resolve_a16w16_candidate) redirects kid 10210→10200 and 10213→10203 when N not in GFX942_BF16WS_EXACT_N, but aiter/ops/opus/gemm_op_a16w16.py:419-431 passes the tuned row's solidx and an explicit kernelId verbatim into _execute_a16w16, and aiter/ops/opus/launch_plan.py:285-294 then raises ValueError("gfx942 exact kid 10210 requires N in [64, 128, 256, 384, 512, 1024, 2048]; got N=1000") with no redirect anywhere in between. Impact at runtime: on a gfx942 device, gemm_a16w16_opus(A, B, kernelId=10210, splitK=2) with A: [M,K] bf16 and B: [K,1000] bf16 (N=1000, outside the exact-N set; the guard fires for any splitK) now hard-raises through the compatibility entry the author claims is behavior-preserved, where base's generated launcher (csrc/opus_gemm/codegen/gen_instances_gfx942.py:351-357 at base) silently called the fp32-workspace sibling 10200 and returned a correct Y; the tuned-row path has the same asymmetry via a user AITER_CONFIG_GEMM_BF16_FILE override (the shipped CSV carries no such rows and the base tuner already refused to emit them, so the explicit-kernelId path is the concrete trigger). Action: Author must route gemm_a16w16_opus's tuned-row and explicit-kernelId kids through resolve_a16w16_tuned_candidate (or apply the same 10210/10213 redirect inside _build_a16w16_launch_plan) so all entry points agree. [verified]

op_tests/test_opus_a16w16_gemm.py
```diff
@@ -181,15 +163,9 @@ def run_a16w16_case(
...
-    dist: str = "norm",
-    gen=None,
-    const_val: float = 1.0,
-    iters: int = 101,
-    warmup: int = 2,
-    rotate: int = 0,
 ):
-    A = _make_a(batch, M, K, dist, gen, const_val)
-    B = _make_b(batch, N, K, dist, gen, const_val)
+    A = torch.randn(batch, M, K, device="cuda", dtype=torch.bfloat16)
```
demonsan: shouldn't change rand init method due to gfx1250 benchmark need

op_tests/test_opus_co_integration.py
```diff
@@ -151,17 +151,18 @@ def test_co_manifest_uses_supported_architecture_keys():
 
 
 def test_gfx1250_co_registry_and_launch_contract():
-    assert len(gfx1250_4wave_co_kernels_list) == 219
+    assert len(gfx1250_4wave_co_kernels_list) == 221
     assert GFX1250_4WAVE_CO_KIDS == frozenset(gfx1250_4wave_co_kernels_list)
     assert min(GFX1250_4WAVE_CO_KIDS) == 21016
-    assert max(GFX1250_4WAVE_CO_KIDS) == 21315
+    assert max(GFX1250_4WAVE_CO_KIDS) == 21317
```
demonsan, line 157: better use scope not certain value. can fix later

demonsan (approved):
LGTM

valarLip (approved):

yifehuan (approved):
LGTM

PR #4963, opened 2026-08-24
[FlyDSL] gfx942 fp8_mqa_logits: let _auto_variant choose rows_per_block

jin-amd:
## Motivation

On gfx942 `_auto_variant` returns `f"mfma_r2_w{wpb}"`, so `rows_per_block` is pinned at 2 and
seven of the nine registered variants — including the entire `r4` family — can never be
selected. At the shape vLLM actually issues for long-context indexer prefill (`seq_len` 1024,
`seq_len_kv` 131072) the pinned pick is the median of the nine by speed, with the best 1.65x
faster. vLLM chunks prefill to fit `VLLM_SPARSE_INDEXER_MAX_LOGITS_MB` (512 MB), which caps
`seq_len` at 1024 at that context, so the selector's own `seq_len >= 2048` branch can never
fire there. Since vllm-project/vllm#49544 routes gfx942 `fp8_mqa_logits` to this kernel by
default, every MI300X/MI325X user inherits it.

## Technical Details

Sweeping all nine variants over 180 shapes surfaced two separable effects.

**RPB tracks the logits element count, not `seq_len`.** Over
`seq_len ∈ [1, 8192] × seq_len_kv ∈ [1024, 262144]` the boundaries land on the same value of
`seq_len * seq_len_kv` at all six contexts: `r1` below 2^19 (27/27 shapes), `r2` at 2^19 (6/6),
`r4` from 2^21 up (38/38), with 2^20 a transition band split 3/3.

**RPB must also divide `seq_len`.** When it does not, the launcher pads with four `torch.cat`
calls. That is a flat ~44 us of host-side overhead, independent of `seq_len_kv`: at `seq_len` 1,
`seq_len_kv` 131072, `r1` runs 23.1 us and `r2` 67.8 us, of which the four cats measure 44.1 us
(0.7 us unexplained), and pre-padding the input by hand recovers all of it. So the cost is not
proportional to the wasted rows — one padded row of two costs the same as three of four — and the
pinned `r2` walks into it on *every odd `seq_len`*, 37 of the 180 shapes swept, at a median 2.97x
penalty.

The step-down to a divisor has to be gated to small shapes, where the kernel is cheap relative to
44 us. Above the threshold the reuse wins even when padding: at `seq_len` 1025, `seq_len_kv`
131072, the dividing `r1` takes 3880 us against 2601 us for a padded `r2`.

```python
elems = seq_len * seq_len_kv
if elems < 2**19:
    rpb = 1
elif elems < 2**21:
    rpb = 2 if seq_len % 2 == 0 else 1
else:
    rpb = 4
wpb = 2 if (seq_len >= 2048 and seq_len_kv >= 8192) else 4
```

`waves_per_block` is deliberately left unchanged. It is worth a few percent at most here, and
unlike RPB its optimum moves with the head count — at 64 heads the current rule costs 1.61x worst
case where a fixed `w4` costs 1.07x — so it warrants its own sweep rather than a change fitted to
one head count.

## Test Plan

MI325X (gfx942), all nine variants timed at 180 shapes (1620 timings): a power-of-two
`seq_len × seq_len_kv` grid, plus three held-out sets the rule was not fitted on —
non-power-of-two shapes, a fine `seq_len` sweep through the divisibility cliff, and head counts
16 and 64. Selector changes were verified end to end through `flydsl_fp8_mqa_logits(variant=None)`
with no `FLYDSL_FP8_MQA_LOGITS_VARIANT` set, so the selector itself is under test. Every variant's
output was compared against a reference at every shape.

## Test Result

Old pick vs new, no env override:

| seq_len | seq_len_kv | before | us | after | us | speedup |
| ---: | ---: | --- | ---: | --- | ---: | ---: |
| 1024 | 131072 | r2_w4 | 2598.7 | r4_w4 | 1623.5 | 1.60x |
| 1025 | 131072 | r2_w4 | 2609.3 | r4_w4 | 1594.0 | 1.64x |
| 700 | 50000 | r2_w4 | 661.6 | r4_w4 | 440.3 | 1.50x |
| 333 | 12000 | r2_w4 | 118.4 | r4_w4 | 97.1 | 1.22x |
| 3 | 131072 | r2_w4 | 66.4 | r1_w4 | 28.1 | 2.36x |
| 1 | 1024 | r2_w4 | 66.2 | r1_w4 | 22.2 | 2.98x |

No shape measured regresses; the smallest gain is 1.04x. Scored against the fastest of the nine
at each shape, pooled over the held-out data, the geometric mean cost falls from 1.45x to 1.03x
and the worst case from 3.17x to 1.41x.

Logits are bitwise identical across all nine variants at all 180 shapes, so this is purely a
blocking/occupancy change with no numerical trade-off.

End to end on 8x MI325X, TP8, GLM-5.2-FP8, 131072 in / 1024 out, concurrency 8: median TPOT
improves 7.22% and output throughput 6.60%. This kernel is 16.8% of GPU time at that point, the
largest single item in the profile.

## Submission Checklist


jin-amd:
@vpietila-amd @valarLip — would appreciate a look when you have a moment.

Short version: on gfx942 `_auto_variant` returns `f"mfma_r2_w{wpb}"`, so `rows_per_block` is
pinned at 2 and seven of the nine registered variants are unreachable. Choosing it from the
logits element count instead is 1.60x at the long-context prefill shape vLLM actually issues,
and up to 2.99x at small or odd `seq_len`. Outputs stay bitwise identical, and
`test_flydsl_fp8_mqa_logits.py` shows no new failures (368-case default matrix plus a 160-case
matrix that reaches `r4`, run patched and unpatched).

aiter/ops/flydsl/kernels/mqa_logits/fp8_mqa_logits.py
```diff
@@ -387,12 +387,47 @@ def _mk_builder(rpb, wpb):
...
+# (MI325X) over seq_len 1..8192 x seq_len_kv 1024..262144, the boundaries land
+# on the same element count at every context:
+#
+#   elements < 2**19   RPB=1 fastest at all 27 shapes measured
+#   elements = 2**19   RPB=2 fastest at all 6
+#   elements >= 2**21  RPB=4 fastest at all 38
+#
+# (2**20 is the transition band, split 3/3 between RPB=2 and RPB=4.)
+_RPB2_MIN_ELEMS = 2**19
+_RPB4_MIN_ELEMS = 2**21
```
samremes: Could move this inside the `_auto_variant` as they are very local to the heuristic.

samremes (commented):
Please trim the comments down, they don't need to tell the whole story how it ended up with these. State the logic very briefly.

jin-amd:
> Please trim the comments down, they don't need to tell the whole story how it ended up with these. State the logic very briefly.

Thank you @samremes  Both addressed in d863e4d. The comments are trimmed down to a two-line docstring that just states the rule (RPB from seq_len * seq_len_kv thresholds, stepped down to a divisor of seq_len when padding overhead would dominate; WPB unchanged), and the two element-count constants are now locals inside _auto_variant. The rationale and measurements stay in the commit message. No logic change.

PR #4970, opened 2026-08-24
[FlyDSL] QRInt4: INT4 two-shot all-reduce for gfx942/gfx950

samremes:
## Summary
- Add `QRInt4`: FlyDSL INT4 two-shot all-reduce (reduce-scatter, then all-gather) for gfx942/gfx950, TP 2/4/8.
- Rank tile is 1152 B: 1024 B INT4 + 128 B group-16 E4M3. Fanout is 18×64 B NT stores (8+8+2). Super-tile is 1 or 8.
- Host: HIP IPC inbox, ST pick, bf16 payload. Tests gate SQNR ≥ 18 dB vs fp32 NCCL on every rank.

This is a Python launcher (`from aiter.ops.flydsl import QuickAllReduceInt4`). It is not wired into the HIP QuickReduce dispatch tables.

## Why this is faster than HIP QR INT4
HIP QR INT4 (`CodecQ4`) is one tile per handshake. This kernel packs ST=8 tiles under one last-sector handshake, and uses lockstep 64 B NT fanout.

Group-16 E4M3 also beats HIP CodecQ4 numerics: SQNR ~19.2 dB vs fp32 all-reduce (HIP INT4 is ~18.3 dB / rel MAE ~0.123).

### gfx942 (MI325X, 304 CU)

| tokens × hidden | ST | FlyDSL µs | HIP QR INT4 µs | vs HIP |
| 512 × 5120 | 1 | 23.62 | 30.75 | −23% |
| 512 × 4096 | 1 | 22.00 | 28.56 | −23% |
| 32768 × 5120 | 8 | 834 | 1121 | −26% |

### gfx950 (MI355X, 256 CU)

| tokens × hidden | ST | FlyDSL µs | HIP QR INT4 µs | vs HIP |
| 512 × 5120 | 1 | 20.4 | 28.8 | −29% |
| 512 × 4096 | 1 | 18.1 | 26.5 | −32% |
| 32768 × 5120 | 8 | 558 | 1061 | −47% |

## Test plan
- [ ] `python3 op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py` -- validity and perf tests


aiter/ops/flydsl/kernels/qr_int4.py
```diff
@@ -0,0 +1,264 @@
...
+    def close(self):
+        for eng in self._by_st.values():
+            eng.close()
+        self._by_st = {}
+
+    def allreduce(self, inp, out, stream=None):
+        live_bytes = self._check_payload(inp, out)
+        num_tiles = max(1, (live_bytes + TILE_BYTES - 1) // TILE_BYTES)
+        st = self._pick_st(num_tiles)
+        self._launch_eng(self._by_st[st], inp, out, stream)
```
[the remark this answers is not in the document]
samremes: That's responsibility of the caller.

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
```diff
@@ -0,0 +1,172 @@
...
+        hip = cls._load_hip()
+        buf = ctypes.c_void_p()
+        err = hip.hipExtMallocWithFlags(
+            ctypes.byref(buf),
+            ctypes.c_size_t(size),
+            ctypes.c_uint(cls._HIP_DEVICE_MALLOC_UNCACHED),
+        )
+        cls._hip_check(err, what="hipExtMallocWithFlags")
+        err = hip.hipMemset(buf, 0, ctypes.c_size_t(size))
+        cls._hip_check(err, what="hipMemset")
```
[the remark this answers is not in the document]
samremes: it raises anyway if _hip_check fails -> moot point.

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
```diff
@@ -0,0 +1,172 @@
...
+
+import ctypes
+
+
+class UncachedIpcHeap:
+    """Allocate uncached device memory and share it across ranks via HIP IPC."""
+
+    _HIP_IPC_HANDLE_BYTES = 64
+    _HIP_IPC_MEM_LAZY_ENABLE_PEER_ACCESS = 0x1
+    _HIP_DEVICE_MALLOC_UNCACHED = 0x3
```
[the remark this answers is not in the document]
samremes: HIP docs:
```
#define hipDeviceMallocUncached   0x3
```

jin-amd:
I took this through a full end-to-end serving A/B on GLM-5.3, and it holds up well. Summary below:

**Setup:** 8x MI325X (gfx942), TP8, bf16. Op-level on a flydsl 0.3.0 image; end-to-end on vLLM serving GLM-5.3-FP8 (hidden 6144, 78 layers, TP8, 263k max model len).


## End-to-end on GLM-5.3: needs a minimum-size gate

The PR ships the launcher without wiring it into any dispatch table, so we added a vLLM communicator that mirrors `QuickAllReduce` and sits ahead of it in the `CudaCommunicator` chain, scoped to TP groups only (EP reduces stay on PYNCCL). Happy to upstream that separately.

The important finding: **claiming every payload is a net loss, and a minimum-size gate is required.** With no gate, every one of 11 sweep points loses a near-constant 1.25 ms per output token:

| point | baseline ITL p50 | ungated | gated at 2 MiB |
| --- | --- | --- | --- |
| 128k/c16 | 19.12 | 19.63 | 19.16 |
| 128k/c8 | 16.46 | 17.47 | 16.51 |
| 128k/c2 | 13.18 | 14.41 | 13.23 |
| 256k/c8 | 16.88 | 18.09 | 17.16 |

The cause is the fixed cost. GLM-5.3 does one TP all-reduce per layer across 78 layers, and decode payloads here are 24 KiB (c2) to 192 KiB (c16) — far below HIP QuickReduce's own 2 MiB INT4 floor, so stock vLLM routes them to the unquantized custom all-reduce at ~3 us. QRInt4 has a ~18.4 us floor regardless of payload, so claiming decode costs 78 x ~16 us = ~1.25 ms/token, which matches the measured 1.23 ms mean offset.

Setting the floor to 2 MiB (matching HIP QuickReduce) restores decode to neutral and keeps the prefill win. Geomean over 11 points, oriented so >1.00 favours QRInt4:

| metric | ungated | gated at 2 MiB |
| --- | --- | --- |
| output throughput | 1.026x | **1.020x** (positive at 11/11 points) |
| TTFT p50 | 1.067x | 1.021x |
| TPOT p50 | 1.000x | 1.024x |
| ITL p50 | **0.934x** | 0.994x |
| ITL p99 | 1.019x | 1.049x |

Worth being precise about where the end-to-end gain comes from: the kernel is only claiming prefill-sized batches, ITL p50 (a pure decode step) is flat, and the win shows up in TTFT and in ITL p99 — decode steps that were stalled behind a prefill chunk. TPOT improves as a consequence, not because decode got faster.

**Suggestion:** ship a recommended minimum payload size with the kernel, ideally as a per-(world_size, dtype) table like HIP QuickReduce's `_QR_MIN_SIZE`, rather than leaving it to the integrator to discover. As written, a naive integration is a regression on any decode-heavy workload.

## Accuracy

gsm8k 5-shot, 1319 questions, chat template, seed 4321:

| arm | strict-match | flexible-extract |
| --- | --- | --- |
| baseline (HIP QR INT4) | 0.9757 | 0.9742 |
| QRInt4 ungated | 0.9757 | 0.9750 |
| QRInt4 gated at 2 MiB | 0.9765 | 0.9757 |

No movement beyond a single question, which is what you would expect given QRInt4's codec is the more accurate of the two.

samremes:
@copilot resolve the merge conflicts in this pull request

aiter/ops/flydsl/quick_allreduce_int4.py
```diff
@@ -0,0 +1,358 @@
+# SPDX-License-Identifier: MIT
```
coderfeli, line 1: this is not kernel. put outside kernels folder.

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
```diff
@@ -0,0 +1,172 @@
+# SPDX-License-Identifier: Apache-2.0
```
coderfeli, line 1: this is not kernel. put outside kernels folder.

aiter/ops/flydsl/kernels/qr_int4_kernel.py
```diff
@@ -0,0 +1,759 @@
...
+def _f16x2(packed):
+    return fx.Vector.from_elements([packed], fx.Int32).bitcast(fx.Float16)
+
+
+def _i32(vec):
+    return vec.bitcast(fx.Int32)[0]
+
+
+def _minnumf(a, b):
+    return fx.Vector(fx.arith.minnumf(a, b), a.shape, a.dtype)
```
coderfeli: fx.min?

aiter/ops/flydsl/kernels/qr_int4_kernel.py
```diff
@@ -0,0 +1,759 @@
...
+    wave: one lane at a time, copy that lane's pointer to a scalar
+    register, mask to that lane, issue the load, repeat. ``readfirstlane``
+    copies lane 0's value into a scalar register once so the whole wave
+    issues a single buffer op.
+
+    Do this at the inbox descriptor, not on the peer list: fanout stores
+    use a different peer per lane, so those addresses must stay in vector
+    registers. ``T.i64`` is the result type ``readfirstlane`` requires.
+    """
+    return fx.Int64(rocdl.readfirstlane(T.i64, as_ir_value(addr)))
```
coderfeli: why need this wrapper? rocdl.readfirstlane() is ok?

aiter/ops/flydsl/kernels/qr_int4_kernel.py
```diff
@@ -0,0 +1,759 @@
...
+    """
+    ptr_ty = ir.Type.parse("!llvm.ptr<1>")
+    ptr = llvm.IntToPtrOp(ptr_ty, as_ir_value(addr_i64)).result
+    llvm.InlineAsmOp(
+        None,
+        [ptr, as_ir_value(data)],
+        "global_store_dwordx4 $0, $1, off nt",
+        "v,v",
+        has_side_effects=True,
+    )
```
coderfeli: not make sense to have inline asm here.

aiter/ops/flydsl/quick_allreduce_int4.py
```diff
@@ -0,0 +1,358 @@
...
+# Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
```
coderfeli, line 2: quick_allreduce instead of qr?  difficult to understand the kernel name

aiter/ops/flydsl/kernels/qr_int4_kernel.py
```diff
@@ -0,0 +1,759 @@
...
+import flydsl.compiler as flyc
+import flydsl.expr as fx
+from flydsl._mlir import ir
+from flydsl._mlir.dialects import llvm
+from flydsl.expr import gpu, range_constexpr, rocdl
+from flydsl.expr.typing import Int32, Int64, Stream, T, as_ir_value
+
+from . import buffer_ops
+
+WORLD = 8  # Default value for world size
```
coderfeli: too many global variables.

coderfeli:
+ @TennyWang1223  and @xytpai

aiter/ops/flydsl/quick_allreduce_int4.py
samremes: moved and renamed

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
```diff
@@ -0,0 +1,571 @@
+# SPDX-License-Identifier: MIT
```
TennyWang1223, line 1: move to dir: op_tests/multigpu_tests/

TennyWang1223 (dismissed):

samremes:
@TennyWang1223 @coderfeli Could you have a look again?

TennyWang1223 (dismissed):
LGTM

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
```diff
@@ -0,0 +1,571 @@
...
+import statistics
+import sys
+from multiprocessing import Pool, freeze_support, set_start_method
+
+_REPO_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
+if _REPO_ROOT not in sys.path:
+    sys.path.insert(0, _REPO_ROOT)
+
+import pandas as pd
+import pytest
```
valarLip: remove pytest?

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
```diff
@@ -0,0 +1,178 @@
...
+    _HIP_DEVICE_MALLOC_UNCACHED = 0x3
+    _HIP_MEMCPY_HOST_TO_DEVICE = 1
+    _hip = None
+    _hipIpcMemHandle_t = None
+
+    @classmethod
+    def _load_hip(cls):
+        if cls._hip is not None:
+            return cls._hip
+        for name in (
```
coderfeli: load hip so here?

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
```diff
@@ -0,0 +1,571 @@
...
+    # compile() launches every ST binary on this shape and all ranks must pass
+    # the same one, so keep the JIT buffer small at the widest hidden size.
+    compile_tokens = min(512, max(tokens))
+    compile_hidden = max(hiddens)
+    compile_inp = torch.empty(
+        (compile_tokens, compile_hidden), device=device, dtype=torch.bfloat16
+    )
+    compile_out = torch.empty_like(compile_inp)
+    dist.barrier()
+    fly.compile(compile_inp, compile_out)
```
valarLip, line 194: what's this compile for?

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
```diff
@@ -0,0 +1,571 @@
...
+            dist.barrier()
+
+            out = torch.empty_like(inp)
+            fly.allreduce(inp, out)
+            got = out.to(torch.float32)
+            dist.barrier()
+
+            nbytes = int(inp.numel()) * int(inp.element_size())
+            n_tiles = max(1, (nbytes + TILE_BYTES - 1) // TILE_BYTES)
+            st_used = fly._pick_st(n_tiles)
```
valarLip: _pick_st？ maybe more formal way?

valarLip (changes requested):

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
```diff
@@ -0,0 +1,178 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2025 FlyDSL Project Contributors
+# Modifications Copyright (C) 2026 Advanced Micro Devices, Inc.
```
coderfeli, line 3: not for this kernel only?

zufayu:
**Review: ⚠️ needs work** — the INT4 codec, LDS races, cross-launch color handshake (i32 wraparound safe) and ST=8 barriers all check out; no correctness defect found. Two things to address before merge:

1. **CI never exercises the codec edge cases.** The real validity cases — the 4 E4M3 fills (pos/neg underflow, overflow-512, zeros), partial tile, TP2/TP4, hidden=4096 — are pytest-only, but default CI scans `op_tests/*.py` at depth 1 (excludes `multigpu_tests/`) and even under the `multigpu` label `aiter_test.sh` runs `python3 <file>`, i.e. `main()`'s TP8/hidden=5120 normal-fill sweep. So the reserved-zero/saturation paths this PR adds are never tested in CI. Please wire the pytest cases into a CI job, and fix the test-plan path (`op_tests/test_flydsl_quick_allreduce_int4.py` doesn't exist — it's under `multigpu_tests/`).

2. **Rebase needed** — `git apply` fails at `aiter/ops/flydsl/__init__.py:76`; main has since added entries to the same `__all__` block.

Minor: no in-tree consumer yet (which PR activates it?); ~195 MiB/rank persistent IPC inbox at TP8 is unmentioned; `_RESIDENT_WGS_PER_CU` lacks a stated measurement basis.

Validation/Perf: not run (no idle multi-GPU box here).

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
samremes: Just for printing and verifying the extra detail here, I could probably drop it completely though. `_pick_st` is not supposed to be a public API.

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
samremes: Warmup, so it won't compile it later.

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
samremes: sure, I can remove the pytest cases and only include the main.

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
samremes: Some of these IPC helpers were adapted from another kernel in FlyDSL repo originally. I think these could be refactored later on to common reusable pieces, but this kernel is probably the only user in aiter now.

samremes:
>CI never exercises the codec edge cases

Yes, I know the CI does not run pytest. But should I put all of those tests through main manually if I remove pytest completely from the tests? So to simplify I would maybe rather just remove the extra validity tests and let main run the normal tests only.

aiter/ops/flydsl/quick_allreduce_int4_ipc.py
samremes: Yes — this helper needs the HIP runtime for IPC (`hipIpcGetMemHandle` / uncached malloc). I dropped the duplicated `libamdhip64.so*` loop and now load through `aiter.dist.device_communicators.vmm_allocator.load_hip_runtime`, then still bind the IPC ctypes signatures here. No new C++ HIP wrapper for this PR.

op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py
```diff
@@ -0,0 +1,453 @@
...
+        "grid_cap": row["grid_cap"],
+        "flydsl us": us,
+        "flydsl TFLOPS": (flops / us / 1e6) if us else 0.0,
+        "flydsl TB/s": (nbytes / us / 1e6) if us else 0.0,
+        "flydsl err": max(r[0]["err"] for r in ranks),
+        "flydsl sqnr_db": min(r[0]["sqnr_db"] for r in ranks),
+    }
+
+
+test_quick_allreduce_int4.__test__ = False
```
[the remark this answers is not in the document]
samremes: pytest was removed so avoid collecting it as a pytest case.

zufayu:
Removing them "to simplify" loses the part that mattered most: the four codec fills (pos/neg underflow at 2^-8, overflow-512, true-zero-scale) are the reserved-zero-remap and E4M3-saturation edges — for a lossy INT4 codec those are exactly where silent corruption hides, and normal-fill sweeps never touch them.

You don't need pytest or a CI change to keep the coverage, though. `main()` already asserts correctness (it calls `_assert_validity`, SQNR + per-tile floor) — it just sweeps TP8/hidden=5120/normal-fill. So fold the four fills (and ideally one TP2 or TP4 shape + hidden=4096) into `main()`'s default sweep as assert-only cases. Then the existing `multigpu`-label run, which already invokes `python3 <file>` → `main()`, exercises the edges too — zero new CI infra.

Dropping the fills entirely is the one option I'd push back on; folding them into `main()` gets you the simplicity without shipping the codec edges untested.

zufayu:
need to clean Copilot comment

aiter/ops/flydsl/kernels/quick_allreduce_int4.py
```diff
@@ -0,0 +1,730 @@
...
+    # Each rank owns this many 16-byte atoms of a 32 KiB tile
+    # (8 GPUs → 1, 4 → 2, 2 → 4). LDS still holds all ATOMS atoms.
+    rank_atoms = ATOMS // world_size
+    # Last-sector pad is ST * rank_atoms * RANK_TILE_I32 after the ST tiles.
+    rank_payload_i32 = rank_atoms * RANK_TILE_I32
+    release_i32_off = super_tile * rank_payload_i32
+    wire_tile_i32 = release_i32_off + 16
+    wire_tile_bytes = wire_tile_i32 * 4
+
+    flags_i32 = PHASES * grid * world_size
```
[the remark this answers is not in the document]
samremes: Fixed in 4b7a9f97c, as a hard precondition rather than padding.

`make_quick_allreduce_int4_kernel` now rejects any grid whose flag prefix is not a whole 64 B sector. The required multiple is `16 // (PHASES * world_size)`: 4 at TP2, 2 at TP4, 1 at TP8. So `TP2/grid=1`, `TP2/grid=2` and `TP4/grid=1` raise instead of silently building a wire area where every rank-tile and release sector straddles two hardware sectors; `TP8/grid=1` is already 64 B aligned and stays legal.

I went with the guard instead of rounding the prefix to 16 i32s because no supported configuration is affected: `clamp_grid_cap` over TP2/4/8 x ST1/8 on 80/228/256/304 CU with the default cap of 1216 always lands on a multiple of 4. Padding would instead shift every peer's data offset, so it would need a fresh 8-GPU wire-protocol re-validation, and the only thing it would buy is tiny hand-set grid caps that are not useful for an all-reduce.

Verified with `python3 op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py` on 8x gfx950: codec-fill and TP2/4/8 validity asserts pass, SQNR 19.2 dB, latencies unchanged (30.1 / 187.6 / 566.2 us).

aiter/ops/flydsl/quick_allreduce_int4.py
```diff
@@ -0,0 +1,363 @@
...
+            raise ValueError(
+                f"world_size={world_size} does not match group size {group_world}"
+            )
+        if group_rank != int(rank):
+            raise ValueError(f"rank={rank} does not match group rank {group_rank}")
+        _validate_ipc_process_group(group, rank=int(rank))
+        arch = get_gfx_runtime()
+        if arch not in _SUPPORTED_ARCHS:
+            raise RuntimeError(
+                f"QuickAllReduceInt4 supports {', '.join(_SUPPORTED_ARCHS)}, got {arch}"
```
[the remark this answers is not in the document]
samremes: Leaving this as is: `get_gfx_runtime()` is the aiter-wide architecture check, and this launcher should not be the one module that deviates from it.

It is what 23 modules use for arch gating today, including `aiter/ops/mha.py`, `aiter/mla.py`, `aiter/fused_moe.py`, `aiter/ops/gemm_op_a8w8.py` and `aiter/jit/core.py`. The `aiter/ops/opus/_arch.py` path cited here is the outlier, and even it falls back to `get_gfx_runtime()` when `gcnArchName` is not usable.

The CU count is already read from the target device, not from the cached agent: `torch.cuda.get_device_properties(self._device_index).multi_processor_count`. Only the arch string is process-wide.

If per-device arch resolution is the right standard for a heterogeneous host, it belongs in the shared helper so every op gets it, not in a bespoke probe inside one FlyDSL launcher. Changing the shared helper is out of scope for this PR.

samremes:
> need to clean Copilot comment

@zufayu 
1. Asserting against the minor nit mentioned by Copilot for tiny grids. 
2. The `get_gfx_runtime` is aiter-wide helper, so I will use it also here rather than deviate.

samremes:
Copilot:
>The launcher does not reject streams belonging to a different GPU, risking invalid cross-device launches.

It is caller's responsibility to provide a valid stream, especially when it is an optional parameter and default is already correct, we could do endless validation for anything here. Other ops in aiter don't do this either so I won't deviate from the common pattern.

valarLip (approved):

PR #4972, opened 2026-08-24
[TRITON] Make RNG deterministic in unit tests

brunomazzottiamd:
## Motivation

Triton unit test `op_tests/triton_tests/attention/test_mla.py` was failing in a non-deterministic way. This PR fixes this issue, as well as other non-deterministic unit tests.

## Technical Details

The fix it to deterministically set the RNG seed in the unit tests, on a case by case basis.

## Test Plan

Run the entire Triton test suite on `gfx942` and `gfx950`, through CI AITER runners.

## Test Result

Triton test suite passed on both `gfx942` and `gfx950`. ✅

## Submission Checklist

### DIFF
--- a/op_tests/triton_tests/attention/test_mha_v3.py
+++ b/op_tests/triton_tests/attention/test_mha_v3.py
@@ -896,6 +896,7 @@ def test_mha_fp8(
         pytest.skip(f"FP8 not supported on {_arch}")
 
     torch.cuda.empty_cache()
+    torch.manual_seed(20)
     q = torch.randn((BATCH, SEQLEN_Q, NUM_Q_HEADS, HEAD_SZ), device="cuda", dtype=dtype)
     k = torch.randn((BATCH, SEQLEN_K, NUM_K_HEADS, HEAD_SZ), device="cuda", dtype=dtype)
     v = torch.randn((BATCH, SEQLEN_K, NUM_K_HEADS, HEAD_SZ), device="cuda", dtype=dtype)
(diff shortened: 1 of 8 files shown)


op_tests/triton_tests/attention/test_mha_v3.py
```diff
@@ -896,6 +896,7 @@ def test_mha_fp8(
         pytest.skip(f"FP8 not supported on {_arch}")
 
     torch.cuda.empty_cache()
+    torch.manual_seed(20)
```
vgokhale, line 899: Why 20?

brunomazzottiamd: I don't know why. All other FP8 tests in this file are already using 20 as the RNG seed. I just stuck to the established pattern. I can change it if you wish!

vgokhale: Nah that's fine. I was just curious as it was a specific magic number. Thanks. We can keep as is for consistency.

vgokhale (approved):

Boss2002n (changes requested):
why not seed all relevant libraries at the file's global level

At the top of the test files do the following instead of adding it to every function:

```
random.seed(0)
torch.manual_seed(0)
```

brunomazzottiamd:
> why not seed all relevant libraries at the file's global level
> 
> At the top of the test files do the following instead of adding it to every function:
> 
> ```
> random.seed(0)
> torch.manual_seed(0)
> ```

This approach is flawed because it makes the test file deterministic instead of making each test case deterministic. You can run just a single test case, or a subset of test cases, with `pytest` - this should work deterministically in CI and in any machine. Some bug reproducers can be triggered by just a single test case, with no need to running the entire file.

Furthermore, setting the RNG at global level leave us at the mercy of `pytest` test collection order. If the collection order changes, we can get a 1 / thousands mismatched element error at random.

Boss2002n (approved):

PR #4974, opened 2026-08-24
[HIP] Fix (custom_all_reduce): correct output RankData slot prediction during graph capture

afriedri:
## Problem

Capture-time and flush-time disagree about where an output's `RankData` lives.

`get_output_buffer_RD()` predicts, while the stream is capturing:

```cpp
ptrs = d_rank_data_base_ + graph_unreg_input_buffers_.size()
                         + graph_unreg_output_buffers_.size();
```

`register_graph_buffers()` later writes output *i* to:

```cpp
output_buffers_[self_ptr] = d_rank_data_base_ + num_input_buffers + i;
```

The input count is **still growing** during capture, so the prediction can't be derived from it.
With N all-reduces captured into one graph, the k-th call bakes slot `2k+1` into the kernel while
the flush writes the real peer pointers to slot `N+k`. They coincide only for the last call.

Concretely, with 2 ARs in one graph:

| call | predicted | actual |
|---|---|---|
| `in[0]` | base+0 | base+0 ✅ |
| `out[0]` | base+1 | base+**2** ❌ |
| `in[1]` | base+1 | base+1 ✅ |
| `out[1]` | base+3 | base+3 ✅ |

For `k < N/2` the mis-indexed slot is an **input** `RankData`, so
`cross_device_reduce_2stage_write_mode` writes reduced output through eight peer pointers that
belong to a different tensor — running off the end of the peer allocation.

## Symptom

```
Memory access fault by GPU node-9 (Agent handle: 0x...) on address 0x... Reason: Unknown.
Kernel Name: _ZN5aiter37cross_device_reduce_2stage_write_modeIDhLi8ELb1EEE...
```

Under vLLM this surfaces as `VllmWorker-N died unexpectedly` → `EngineDeadError`, which is easy to
misread as a throughput regression: median TPOT is unchanged and only P99 moves, because the run
is being *truncated*, not slowed.

## Why it needs all of TP=8 + gfx942 + cudagraphs

- `_output_dp` is dereferenced by exactly one kernel, `cross_device_reduce_2stage_write_mode`
  (the other four take it as an unused parameter). It is reached only when
  `world_size_ == 8 && bytes > 4 MiB && arch.find("gfx942")`.
- `get_output_buffer_RD()` is only called when `is_broadcast_reg_outptr`
  (`reg_inp_ptr == 0`, i.e. `registered_input=true`).
- Slot prediction only happens under `hipStreamCaptureStatusActive`.

vLLM makes it near-certain on large dense models: it opens a **single** `capture()` around all
~51 cudagraph sizes, so for Llama-3.1-405B (126 layers × 2 AR) roughly 12,852 captured calls
accumulate into one flush batch and nearly all of them mis-index.

## Fix

Reserve output slots from the **top of the arena downward** (`d_rank_data_end_ - 1 - i`) and use
the identical mapping in `register_graph_buffers()`, so the two agree by construction. Inputs keep
growing upward from `d_rank_data_base_`; the regions converge and the existing capacity check
bounds both.

Also adds `check_rank_data_capacity()` to the capture-time predictions, which were previously
unguarded — an exhausted arena now throws a clear error instead of silently aliasing live
`RankData`.

Slot consumption is unchanged; only placement differs. The eager (non-graph) registration path is
untouched.

## Testing

8× MI300X (gfx942), aiter v0.1.19, vLLM 0.27.1, torch 2.11.0+gitd0c8b1f, triton 3.6.0.

**Llama-3.1-405B-Instruct-FP8-KV, TP=8** — `--max-concurrency 256 --num-prompts 512`, in/out 1024,
`--ignore-eos` (so generated tokens must be exactly 524288):

| Arm | Gen. tokens | EngineDeadError | Out tok/s | P99 TPOT |
|---|---|---|---|---|
| stock v0.1.19 | 518438 ❌ | 137 | 1788 | 169.3 |
| **this patch** | **524288** ✅ | **0** | **1867 / 1871** | **135** |
| `VLLM_ROCM_USE_AITER_CUSTOM_AR=0` | 524288 ✅ | 0 | 1793 | 140.6 |
| vLLM 0.24.0 (pre-aiter-AR) | 524288 ✅ | 0 | 1884 | 133.9 |

5/5 clean runs. Also faster than disabling aiter AR, and within ~1% of the pre-regression
0.24.0 baseline.

**Accuracy** — gsm8k via lm_eval, `--limit 250`:

| Arm | flexible-extract | strict-match | API failures |
|---|---|---|---|
| this patch | 0.972 ±0.011 | 0.968 ±0.011 | 0 |
| vLLM 0.24.0 | 0.960 ±0.012 | 0.944 ±0.015 | 0 |

**Non-regression:**

| Model | TP | Patched | Stock |
|---|---|---|---|
| gpt-oss-120b | 8 | 65536 tok ✅, 3072 tok/s | 65536 tok ✅, 2639 tok/s |

---
Prepared with help of Claude; reviewed by Andy


TennyWang1223 (approved):
LGTM

yzhou103 (approved):

PR #4975, opened 2026-08-24
[Build] [Feature] Add MK1 persistent decoder provider

ssharma4-amd:
## Summary

Adds an optional binary-only `aiter.MK1` provider for persistent (megakernel)
GPT-OSS-120B decoding on AMD Instinct MI355X (`gfx950`). ATOM imports the
provider directly from `aiter.MK1`.

## Design

- AITER owns compiled-checkpoint validation and loading, native provider
  selection, tensor lifetime, cache binding, and quantum execution.
- ATOM retains request scheduling, ordinary model loading, KV-cache allocation,
  and the local persistent-checkpoint path.
- Checkpoints are local-only; the provider performs no Hugging Face networking.
- The public package contains the Python integration, a thin Python extension,
  and provider binaries. It contains no private kernel source and has no
  `libanvil.so` dependency.
- The package-root `aiter.persistent_decoder` compatibility module is
  intentionally absent so there is one public import path.

## Native compatibility

- `module_persistent_decoder.so` reports its compiled `native_abi_version` to
  the Python loader. AITER requires version 1.
- The host adapter uses `dlvsym` to resolve the provider factory at the exact
  ELF symbol version `REDLINE_PERSISTENT_DECODER_TYPED_RUNTIME_1`.
- No separate JSON native manifest is shipped. Compatibility is checked at the
  two native boundaries that are actually invoked.

## Packaging

- Includes `aiter.MK1` and its three native artifacts in the AITER wheel.
- Adds `safetensors` because AITER owns loading the compiled persistent
  checkpoint shards.
- Uses explicit BF16 and FP16 provider filenames; the initial ATOM integration
  selects BF16 through ATOM's existing cache configuration.

## Supported integration

- GPT-OSS-120B revision `b5c939de`
- AMD Instinct MI355X / `gfx950`
- ATOM BF16 KV cache
- TP1 / PP1 / PCP1 / DCP1
- Batch-one greedy persistent decode

## Validation

Validated on AMD Instinct MI355X:

- Native rebuild: Fleet `job-e51aa84a`
  - extension reports ABI version 1;
  - wrapper imports `dlvsym`;
  - BF16 provider exports the required versioned factory symbol;
  - BF16 provider construction succeeds;
  - gfx950 code object is present;
  - no `libanvil.so` dependency.
- Isolated public AITER wheel: Fleet `job-687ce457`
  - wheel contains all three required native artifacts;
  - wheel contains no `native_manifest.json` or `libanvil.so`;
  - installed extension reports ABI version 1;
  - an incompatible ABI is rejected;
  - BF16 provider is constructed and mapped successfully;
  - all seven focused AITER MK1 tests pass.
- Existing end-to-end validation remains green: ATOM chat produced coherent
  output, the serving benchmark completed 60/60 requests, and three matched
  1024-input/128-output concurrency-one persistent runs averaged 1.962 ms TPOT.

The native provider implementation is distributed only as compiled binaries;
no private kernel source is included in this PR.


aiter/MK1/persistent_decoder.py
```diff
@@ -0,0 +1,304 @@
...
+    validate_checkpoint,
+)
+from .config import MK1Config, is_supported
+from .errors import BackendLaunchError, CheckpointError, PrelaunchError
+
+ATOM_PHYSICAL_PAGE_SIZE = 16
+
+
+@dataclass(frozen=True, slots=True)
+class AtomCacheBinding:
```
valarLip: things like these Atomxxxxx should be in aiter

ssharma4-amd: Updated

PR #4977, opened 2026-08-25
[HIP] [Feat] Single-kernel Lamport fused all-reduce + RMSNorm (opt-in)

EricKing626:
## Motivation

`custom_fused_ar_rms` runs as two kernels — two-shot all-reduce, then RMSNorm — separated by an
`end_sync` grid barrier. On ROCm there is no Programmatic Dependent Launch, so the two strictly
serialise.

On GLM-5.2 FP4 decode (isl 8192 / osl 1024 / conc 64, TP4) this pair is the largest single
contributor to our remaining MI355X latency gap. Removing the serialisation means collapsing the
two kernels into one, which means removing the barrier between them.

## Approach

Lamport wait-free signalling: **the data is the flag.**

- The buffer is pre-armed with a sentinel (`-0.0`).
- Producers canonicalise `-0.0 -> +0.0`, so a written value can never be the sentinel.
- A pack counts as arrived once no word in it equals the sentinel.

`end_sync` is no longer needed, so reduce and norm become one kernel — one launch, no gap.

Two supporting details worth reviewing:

1. **`start_sync_acqrel`** — RELEASE/ACQUIRE variant of the existing RELAXED `start_sync`, so the
   re-arming sentinel stores of iteration *N* cannot be reordered past a peer's producer stores of
   iteration *N+1*.
2. **Occupancy-clamped grid** — the kernel is persistent-style; a block spins on data produced by
   other blocks of the same grid. The grid is clamped to the co-resident block count via
   `hipOccupancyMaxActiveBlocksPerMultiprocessor`. Without the clamp, unscheduled producer blocks
   can never run behind spinning consumers and the kernel deadlocks.

## Changes

Confined to `csrc/include/custom_all_reduce.cuh`. Adds a Lamport sentinel/arm/spin helper set,
`start_sync_acqrel`, and the fused kernel plus its dispatch. Nothing existing is modified.

## Enabling

**Off by default.** Opt in with `AITER_AR_RMSNORM_LAMPORT=1`.

The fast path also requires: `n % pack_size == 0`, `packs_per_row >= 64`, `1 <= n_loop <= 4`,
`(m * packs_per_row) % world_size == 0`, and `world_size` in {2, 4, 8}. Any failure falls through
to the existing 1-stage / 2-stage paths unchanged.

## Testing status

**Not yet compiled or run on hardware.** Opened early for design review of the Lamport protocol
and the occupancy clamp. Correctness and performance numbers will be posted before merge.

Planned: TP=4/8, `n=7168`, `m` in {1, 64, 256}, elementwise vs `AITER_AR_RMSNORM_LAMPORT=0`;
several thousand back-to-back iterations for the re-arming race; end-to-end GLM-5.2 decode A/B.

## Known gaps

- Graph capture records the lazy `lamport_ensure_armed` call; needs a warm-up before capture.
- `tnum` fixed at 512, not swept.
- Only `local_device_load_rmsnorm` semantics ported; `_naive`, `_512n` and quantised variants keep
  the two-kernel path.
- No unit test yet.

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -1745,7 +1745,6 @@ __global__ void __launch_bounds__(tnum, 1)
 
     const int warp_id = threadIdx.x / tnum_gpu;
     const int lane_id = threadIdx.x % tnum_gpu;
-    const int tid     = blockIdx.x * tnum_gpu + lane_id;
 
     const P* ptrs[ngpus];
     P* peer_tmps[ngpus];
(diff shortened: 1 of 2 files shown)


stefanskiasan:
Measured this PR end-to-end on 4x MI355X (gfx950, ROCm 7.2.3, aiter 0.1.19 wheel + this module JIT-built with `pybind11==3.0.1`, see #4996): full GLM-5.3 (`glm_moe_dsa`, MXFP4 experts, fp8 KV, TP4, MTP k=3, FULL cudagraphs) in vLLM, all-reduce + residual + RMSNorm fused at both norms of all 78 layers. Three variants on the same GPUs, same image, no prefix cache, needle/perplexity checks identical for all three:

| variant | decode bs=1 (tok/s) | throughput 64 users (tok/s) | mean TPOT |
|---|---|---|---|
| baseline (RCCL all-reduce, separate RMSNorm) | 121 / 122 | 617 / 634 | 45.6 / 44.6 ms |
| 2-kernel fused AR+RMSNorm (existing path) | 123 / 122 | 614 / 626 | 45.9 / 44.9 ms |
| this PR, `AITER_AR_RMSNORM_LAMPORT=1` | 108 / 113 | 560 / 572 | 50.5 / 48.4 ms |

So for hidden 6144 / bf16 / TP4 decode-sized messages the Lamport path is ~10 % slower than the 2-kernel path here, and the 2-kernel fusion itself is neutral vs. RCCL. Numerically both are fine. Happy to re-run if there is a tuning knob (message-size threshold, wavefront count) I should set for gfx950.

PR #4798, opened 2026-08-17
config(glm5): retune MXFP4 MoE token=8 to a4w4

jiaryang:
## Summary

- Follow-up to #4266: `glm5_fp4_tuned_fmoe.csv` already moved several `inter_dim=512` tokens to `flydsl_mxmoe` a4w4, but **token=8** was still on afp4/mfma. Non-MTP decode **c8** (`nextPow2(8)=8`) hit that slow path.
- Retuned that single shape with `gemm_moe_tune.py --mxfp4-flydsl` on MI355X (gfx950).
- Best: `flydsl_mxmoe_g1_a4w4_16x256x256_f16in_nt` + `flydsl_mxmoe_g2_a4w4_16x256x256_atomic` (`block_m=16`, coupled `us=68.581`).

## Selected kernels (token=8, `6144×512`, E=257, topk=9)

Confirmed via `fused_moe` log on all three serving shapes below.

| stage | before (afp4) | after (a4w4, this PR) |
| --- | --- | --- |
| g1 | `flydsl_moe1_afp4_wfp4_bf16_t32x32x256_w3_xcd4_kw4_fp4` | `flydsl_mxmoe_g1_a4w4_16x256x256_f16in_nt` |
| g2 | `flydsl_moe2_afp4_wfp4_bf16_t32x128x256_reduce_bnt2` | `flydsl_mxmoe_g2_a4w4_16x256x256_atomic` |

## Serving A/B (MI355X TP4, GLM-5.2-MXFP4, non-MTP, c8)

Recipe: `max-running-requests=8`, DSA triton, kv fp8, `SGLANG_ROCM_FUSED_DECODE_MLA=0`, `ROCM_QUICK_REDUCE_QUANTIZATION=INT4`, `SGLANG_OPT_USE_TOPK_V2=0`.

- **TTT** = Total token throughput (tok/s) / 4
- **TTT vs B200** = MI / B200 (higher is better)
- **TPOT vs B200** = B200 / MI (higher is better)

| shape | before TTT | after TTT | Δ TTT | TTT vs B200 (b→a) | before TPOT | after TPOT | Δ TPOT | TPOT vs B200 (b→a) |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1024/1024 | 284.6 | **323.0** | **+13.5%** | 78% → **88%** (365) | 13.72 | **11.98** | **−12.7%** | 77% → **88%** (10.6) |
[...]

PR #4179, opened 2026-07-10
[fmoe][gfx950]Flydsl mxmoe v2

charlieguo1106:
## Test Plan

AITER_MOE_EXPERT_BALANCE=true python3 op_tests/test_moe_2stage.py   --no-flydsl-csv  -d bf16   -dim 6144,512 -q 4   -a silu   -s f   -e 257   -k 9   -p t --kernel -t 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 16384 32768

AITER_FLYDSL_STAGE2_FP8=0 AITER_MOE_EXPERT_BALANCE=true python3 op_tests/test_moe_2stage.py   --no-flydsl-csv  -d bf16   -dim 6144,512 -q 4   -a silu   -s f   -e 257   -k 9   -p t --kernel -t 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 16384 32768

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4983, opened 2026-08-25
[Config] config(glm5): restore MXFP4 MoE token=16/32/64/128/8192 to a4w4

Jacob0226:
## Motivation

#4179 moved GLM-5.2 MXFP4 MoE token=16/32/64/128/8192 from `flydsl_mxmoe_g1_a4w4_*` to flydsl bf16 v2. a4w4 is reachable only from `Mxfp4FlydslTuner`, bf16 only from `FmoeTuner` — two modes of one script — yet both write one key: the dedupe drops `block_m` and the kernel name, and `update_tunedf` overwrites in place with no guard unless `--compare --update_improved`. A sweep replaced five entries it never measured. #4798 fixed token=8 the same way; its c8-only A/B missed these.

| token | overwritten row, `Mxfp4FlydslTuner` `us` | replacement — **what `main` ships today** — `FmoeTuner` `us` |
| ----- | ---------------------------------------- | ----------------------------------------------------------- |
| 16    | 90.96                                    | 120.57                                                       |
| 32    | 138.56                                   | 205.89                                                       |
| 64    | 188.04                                   | 207.54                                                       |
| 128   | 206.49                                   | 211.92                                                       |
| 8192  | 876.67                                   | 935.66                                                       |

Direction only, never a speedup: the left times one launch including `moe_sorting`, the right two stages without it — the left measures more work and is still smaller.

## Test Plan

MI355X (gfx950), ROCm 7.2, `rocm/sgl-dev:v0.5.18-rocm720-mi35x-20260824`, aiter `c16d44b93`.

**Step 1 — restore the five rows verbatim, no re-tune.** They come straight out of `db24351ee^`:

    git show db24351ee^:aiter/configs/model_configs/glm5_fp4_tuned_fmoe.csv

**Step 2 — `test_moe_2stage.py` benchmarks each CSV through `fused_moe`, the function the model calls at every MoE layer.** Not the tuner's `us`: the modes time different regions, and `_tag`, the only free-form column, is empty on every row, so nothing records which wrote what. Same inputs, differing only in the kernels the CSV names. Nine repeats, `main` against this branch:

    AITER_CONFIG_FMOE=<config>.csv python3 op_tests/test_moe_2stage.py --no-legacy --no-situv2 -q 4

## Test Result

**`fused_moe`, us/call** from `test_moe_2stage.py`, median of 9 with p25–p75. GLM-5.2 MoE `model_dim=6144, inter_dim=512, expert=257, topk=9`. The last column counts repeats this PR won; the box is shared, so that is the robust read. `logits_diff` ~5.9e-4 both sides.

| token | main   | p25–p75       | this PR    | p25–p75       | Δ          | won  |
| ----- | ------ | ------------- | ---------- | ------------- | ---------- | ---- |
| 16    | 134.4  | 129.5–136.8   | **98.2**   | 95.4–101.5    | **−26.9%** | 9/9  |
| 32    | 188.2  | 188.1–190.7   | **153.1**  | 151.1–154.1   | **−18.6%** | 8/9  |
| 64    | 228.5  | 221.4–230.7   | **216.0**  | 215.5–217.3   | **−5.5%**  | 8/9  |
| 128   | 241.3  | 239.7–241.8   | **226.3**  | 225.0–228.4   | **−6.2%**  | 9/9  |
| 8192  | 1070.6 | 1065.8–1075.1 | **1028.2** | 1023.7–1031.9 | **−4.0%**  | 8/9  |

**Boundary** — same A/B past the restored buckets:

| token | main (bf16) | a4w4 `bm=16` | Δ      |
| ----- | ----------- | ------------ | ------ |
| 256   | 246.8       | 246.0        | −0.3%  |
| 512   | 250.9       | 353.2        | +40.8% |
| 1024  | 279.0       | 536.0        | +92.1% |

256 ties; 512 up belongs to bf16. 8192 restores its own `128x256x256` entry.

**Serving,** SGLang `bench_serving`, GLM-5.2-MXFP4 TP4, median TPOT ms; conc*N* is the decode batch size, so it maps onto a row above. It restored token=16/32 to the rows shipped here.

| shape      | main  | restored | Δ         |
| ---------- | ----- | -------- | --------- |
| i1k conc16 | 16.68 | 15.65    | **−6.2%** |
| i1k conc32 | 22.48 | 20.51    | **−8.8%** |
| i8k conc16 | 20.49 | 19.19    | **−6.3%** |
| i8k conc32 | 30.07 | 27.91    | **−7.2%** |

GSM8K 0.927, unchanged.


Jacob0226:
@yzhou103 
Closing this PR — @charlieguo1106, the author of #4179, has already fixed the regression in #5006, which is now merged into main. On top of that, #5045 retunes the GLM5.2 MXFP4 MoE table across a much broader set of shapes and supersedes the rows this PR touches.

PR #4984, opened 2026-08-25
[FlyDSL] [CI] feat(mega_moe/gfx1250): quantize before dispatch, on an fp8 or fp4 wire

jhchouuu:
## What

MegaMoE's EP dispatch can carry an **already-quantized** payload — fp8 or fp4 — instead of bf16.

```
MegaMoEGfx1250(dispatch_wire=...)        bf16 | fp8 | fp4     default bf16
$MEGA_DISPATCH_WIRE                      same, when the kwarg is not given
--dispatch_wire                          the harness flag
```

The default is bf16, which is what every existing caller gets today.

## Why

Dispatch ships bf16 and every receiver quantizes the copy it got, so a token routed to `topk` peers is quantized `topk` times **on the same values**. Quantizing on the sender does it once per local token and shrinks the wire. At hidden 7168 a token crosses as:

| wire | payload | scale row | total |
|---|---|---|---|
| bf16 | 14336 B | — | 14336 |
| fp8 | 7168 | 256 | 7424 |
| fp4 | 3584 | 256 | 3840 |

MX quant is destination-independent, so moving it changes no bytes — the numbers below are bit-identical, not "close enough".

## Results

**Module** (`test_mega_moe_gfx1250.py`, EP4, hidden 7168, topk 6, medians of 3 interleaved, ranges disjoint), per-layer wall clock at 16384 tokens/rank:

| quant | bf16 wire | quantized wire | |
|---|---|---|---|
| a8w4 | 4654 us | 4378 us (fp8) | **−5.9%** |
| a4w4 | 3512 us | 3286 us (fp4) | **−6.4%** |

512 tokens/rank is parity. Per-kernel attribution puts all of it in dispatch — 449 → 197 us/layer for fp4 — with both GEMMs and combine unchanged.

**End to end**, DSV4-Pro on 4x gfx1250, DP4+EP4, ISL/OSL 1024/1024, concurrency 2048, 8192 requests:

| | bf16 wire | fp4 wire | |
|---|---|---|---|
| total token throughput | 57851.6 tok/s | 60226.5 tok/s | **+4.1%** |
| mean TTFT | 11967.9 ms | 11248.0 ms | **−6.0%** |
| mean TPOT | 58.93 ms | 56.79 ms | −3.6% |
| benchmark duration | 290.0 s | 278.6 s | −3.9% |

## How it works

**Sender.** `per_1x32_mx_quant_hip` once per local token, into the wire dtype the GEMM wants. The wire must MATCH that dtype (a8w4 → fp8, a4w4 → fp4): a mismatch is a row-width error, not a slow path, so it is checked rather than inferred.

**Receiver.** The route-gather kernel keeps its gather and its scale preshuffle and drops only the quant — the preshuffle cannot move to the sender, because its destination is the grouped row *this* rank assigns, which no sender knows. A branch inside `_emit_quant_block_loop` loads what the quant pass would have computed; the store pass, which owns the preshuffled-scale addressing, stays shared.

**Combine is untouched** and still carries bf16. `combine_token_nbytes` is split from the old `token_nbytes` on purpose: deriving the combine slot stride from a now-wire-dependent token size would silently halve it while the rows staged for combine are still bf16.

**Naming.** Everything that means the dispatch wire says so — `is_quant_dispatch_wire`, `dispatch_scale_nbytes`, `dispatch_scale_dst_nbytes`, `dispatch_wire_elem_count`, `_DispatchWire`, and the arena region `disp_out_scales`. `_DispatchWire` is deliberately not generalised: three of its four fields are structurally dispatch-only (mori carries no scales on a combine, a combine quant would happen inside the gemm2 epilogue on an LDS tile rather than host-side on a whole tensor, and `recv_dtype` exists only to build a torch view combine has no equivalent of), so sharing the table would hand a future combine wire three dead fields.

## Also fixed

* **a4w4 could not reach the grouped GEMM.** `is_grouped_a4w4` required `q_dtype_w == fp4x2` while the a8w4 arm also accepted `w1.dtype == uint8`. mxfp4 weights arrive in both spellings (ATOM's loader keeps them uint8), so a4w4 with uint8 weights failed the eligibility test and fell through to the 2-stage mxfp4 kernels — a silent detour to a different kernel family, which on this shape has no tuned config and aborts.

* **`a1_scale` named two different tensors** in `grouped_gemm_gfx1250_a8w4`: the parameter is the caller's per-token e8m0 rows, and the quant pass rebound the same name to the preshuffled grouped scale. Both are uint8 with a plausible shape, so a wrong read was silent.

* **The gfx1250 MegaMoE CI job was not running.** It passes `--combine scatter_fused`, a value #5052 removed, so it exited on argparse.

## Dependency

The quantizing wire needs a mori whose EP dispatch forwards a per-token scale row — **ROCm/mori#593** (merged). The import is inside the property that needs it, so a bf16 wire works with any mori; asking for fp8/fp4 without one raises a message naming the PR.

## Testing

4x gfx1250, `test_mega_moe_gfx1250.py` against an fp32 reference:

```
a8w4 + bf16 wire   logits_diff 0.002174
a8w4 + fp8  wire   logits_diff 0.002174     <- identical
a4w4 + bf16 wire   logits_diff 0.042041
a4w4 + fp4  wire   logits_diff 0.042041     <- identical
```

The harness gains `--dispatch_wire {auto,bf16,fp8,fp4}` and drives `AITER_FORCE_A8W4` off the quant key, so `-q a4w4_mxfp4` cannot silently measure a8w4.

## Note for anyone with a launch script

`MEGA_WIRE` was the name during development and is now `MEGA_DISPATCH_WIRE`, because combine will want a wire of its own. The old name **raises** rather than falling back: an env var that is silently ignored would send a run that asked for fp4 down the bf16 path and report nothing. Setting both names to the same value is allowed, so a fleet can be rolled over one script at a time.


junhaha666 (dismissed):

op_tests/multigpu_tests/test_mega_moe_gfx1250.py
```diff
@@ -1017,6 +1056,15 @@ def _parse_args():
         default="auto",
         help="dispatch transport (communication) dtype",
     )
+    p.add_argument(
+        "--mega_wire",
```
yanboshao: This parameter appears to have overlapping functionality with parameter dispatch_commu_dtype. Would it be feasible to retain only one?

aiter/ops/flydsl/kernels/mega_moe_gfx1250/mega_moe.py
```diff
@@ -118,9 +143,35 @@ class MegaMoEStage2Config:
     dispatch_block_num: int | None = None
     dispatch_warp_num_per_block: int | None = None
     schedule: tuple | None = None
-    dispatch_backend: str = "flydsl"
+    dispatch_backend: str = "mori"
```
yanboshao: This class is the config for stage2. We can add a new stage1 config to hold dispatch‑related parameters. Subsequent stage1 fusion can be modified based on your implementation.

aiter/ops/flydsl/kernels/mega_moe_gfx1250/mega_moe.py
```diff
@@ -37,9 +37,58 @@
...
 
 
+def read_dispatch_wire_env() -> str:
+    """$MEGA_DISPATCH_WIRE, and a loud death for the name it replaced.
+
+    Not a fallback: an env var that is silently ignored sends a run that asked
+    for fp4 down the bf16 path and reports nothing, which is the one failure
+    mode a wire benchmark cannot survive.
+    """
+    stale, current = os.environ.get("MEGA_WIRE"), os.environ.get("MEGA_DISPATCH_WIRE")
```
yadaish, line 53: what about using one more readable name? These two looks quite similar.

yadaish:
LGTM

aiter/ops/flydsl/kernels/mega_moe_gfx1250/mega_moe.py
jhchouuu: MEGA_WIRE is the old name, not a second knob, the live name is MEGA_DISPATCH_WIRE, it appears only in this guard. A launch script that still exports it fails loudly instead of silently serving bf16, and the reference goes away once those scripts are updated. 

And these will also be refactor after dispatch + gemm1 were implemented. 
cc @yanboshao

junhaha666 (approved):

PR #4990, opened 2026-08-25
[HIP] [ROCm] Widen the AR+RMSNorm reduce-scatter producer

EricKing626:
**Branch:** `rs-wide-producer`
**Status:** opt-in, default off. Measured, correct, small.

## Summary

`reduce_scatter_cross_device_store` — stage 1 of the fused all-reduce + RMSNorm
pair — maps threads to peers rather than to data. This PR adds a variant that
maps them to data instead, behind `AITER_AR_RS_WIDE=1`.

The variant is bit-for-bit identical to the kernel it replaces and about **2.7%
faster** at the decode shapes measured. That is below what an end-to-end serving
run can resolve, so it is proposed as a default-off option rather than as a fix
for anything. The measurements are reported in full below, including the ones
that contradict the original hypothesis.

## What this changes

Adds `reduce_scatter_cross_device_store_wide`:

- One pack per thread. All `ngpus` remote loads are issued back to back before
  any of them is consumed.
- Reduction in registers. No LDS staging, no `__syncthreads()`.
- The broadcast store is likewise issued `ngpus`-wide by the same thread.

Addressing, the tmp-buffer layout, and the `end_sync<ngpus, false>`
(RELEASE/ACQUIRE) ordering are unchanged, so stage 2 cannot distinguish the two
variants.

All call sites go through a single `launch_reduce_scatter_cross_device_store`
wrapper so the two variants cannot drift apart. Each call site passes in the
element count it already used to size the stock kernel's grid, rather than
letting the wrapper recompute it: the call sites do not agree on that count
(`dispatchFusedAllReduceRMSNorm` sizes from `m * hidden_dim`,
`allreduce_mhc_post_split_launcher` from `m * input_hidden_dim`), and those
differ whenever the input is padded. With the flag off, the launch is
bit-identical to what it was before this PR.

`kMaxBlocks` remains a hard bound rather than a tuning choice: `start_sync` /
`end_sync` index `Signal::start` / `end` / `_flag` by `blockIdx.x`, and those
arrays are sized `[kMaxBlocks]`.

## The generated code is what was intended

From the gfx950 ISA (`hipcc --cuda-device-only -S`), bf16, `ngpus = 4`:

| | stock | wide |
|---|---|---|
| `group_segment_fixed_size` | 8192 B | **0** |
| VGPR | 32 | 48 (`ngpus = 8`: 78) |
| `s_barrier` | 5 | 3 |
| `ds_write` / `ds_read` | 2 / 5 | **0 / 0** |
| remote loads per trip | 1 | **4, back to back** |

The four loads issue with a single `s_waitcnt` after all of them, so there
really are `ngpus` requests in flight per thread. At `ngpus = 8` the 78 VGPR do
not cost occupancy at 512 threads per block, so the `in[ngpus]` array is not a
problem.

## Measured effect

Producer kernel, from profiler traces of a real decode workload (mean over
ranks, 155 calls per step):

| | flag off | flag on | delta |
|---|---|---|---|
| producer, per call | 13.94 us (median 13.48) | 13.57 us (median 13.08) | **−2.7%** |
| consumer, per call | 4.41 us | 4.49 us | +1.8% |
| collective total, per rank | — | — | −2.1% |

End to end, on a serving benchmark:

| | effect |
|---|---|
| median inter-token latency | **−0.10%** |
| median TPOT | −0.12% |
| output throughput | +0.10% |

That is nothing measurable, and it is the expected result. The producer is
~13.9 us of an 18.1 us collective call, the collective is ~12% of a decode step,
so 2.7% on the producer predicts

```
0.027 x 13.9us x 155 calls / 24.1ms = 0.24%
```

against a single-A/B resolution floor of roughly 0.3%. The optimization is real
at the kernel level and simply too small to resolve end to end.

## The original hypothesis was wrong

This PR was first written on the premise that the kernel is limited by
outstanding cross-device requests: each thread issues one remote load, then
blocks on `__syncthreads()` while only `warp_id == 0` does the add. That
description of the stock kernel is accurate, and the compiler did exactly what
was asked — but the kernel is not limited by it.

`AITER_AR_RS_WIDE_BLOCKS` sweep, TP = 4, producer us/call (median):

| shape | stock | 8 | 16 | 32 | 64 | 80 |
|---|---|---|---|---|---|---|
| (1, 7168) | 6.62 | 7.03 | 6.56 | 8.01 | 7.06 | 8.00 |
| (64, 7168) | 14.25 | 15.13 | 13.83 | 14.03 | 14.27 | 14.58 |
| (256, 7168) | 41.18 | 39.24 | 39.11 | 39.91 | 41.31 | 41.12 |

The curve is flat. Time does not fall as block count rises, which is the
signature of a fabric-bound kernel, not a latency-bound one. Adding outstanding
requests per thread cannot help a kernel that already cannot issue faster than
the link drains, which is consistent with the 2.7% observed.

Decomposing the 18.1 us collective call at the decode shape: ~6.0 us is data
movement, at a per-rank rate close to the fabric ceiling; ~6.8 us is the two
barriers; ~4.3 us is the consumer RMSNorm. The kernel body this PR rewrites is
the part that was already near its limit.

## Enabling

| env | default | effect |
|---|---|---|
| `AITER_AR_RS_WIDE=1` | off | use the wide variant |
| `AITER_AR_RS_WIDE_BLOCKS=N` | 0 (auto) | force the grid; wide variant only |

With `AITER_AR_RS_WIDE` unset the original kernel runs with the original grid.

`AITER_AR_RS_WIDE_BLOCKS` is a diagnostic, not a tuning parameter: sweeping it
separates latency-bound (time falls as block count rises) from fabric-bound
(flat), as above.

Both are read once per process into a function-local static, and the branch is
resolved at graph capture time, so the environment must be set before the
process starts.

## Correctness

Bit-for-bit identical, not merely inside tolerance. Both arms were given the
same seed and their output tensors compared exactly:

| world size | shapes (n = 7168) | tensors compared | result |
|---|---|---|---|
| TP = 2 | m = 1, 8, 64, 256 | 16 | bitwise identical |
| TP = 4 | m = 1, 8, 32, 64, 128, 256 | 48 | bitwise identical |
| TP = 8 | m = 1, 8, 64, 256 | 64 | bitwise identical |

This is the expected result: the wide kernel accumulates the peers in the same
order at the same fp32 precision and writes the same bytes to the same
tmp-buffer offsets, so there is no reordering for the arithmetic to notice.

`op_tests/multigpu_tests/test_fused_ar_rms.py --test fused` (bf16, m = 1..256,
eager, `AITER_AR_1STAGE=0` so every shape takes the two-kernel path) passes with
the flag both off and on, with identical reported error at every shape.
`test_fused_ar_rmsnorm` is now seeded like the other cases in that file, so two
runs differing only in an env var see identical inputs.

## Known, pre-existing, unrelated

The two-kernel path hangs under graph capture in the op-test harness at m >= 64,
where the graph path selects it. This reproduces identically with
`AITER_AR_RS_WIDE=0` and is not introduced by this PR. It does not affect
serving, where the same path runs under a captured graph without trouble.

## Notes on measurement

Recorded because both produced badly wrong numbers before being fixed.

- **Do not run two arms side by side for a kernel measurement.** Both saturate
  the same cross-device fabric; running them concurrently inflated the producer
  from ~14 us to ~29 us on *both* sides. Correctness runs in parallel are fine,
  and so is an end-to-end serving A/B, where both arms are equally loaded.
- **Report the median, not the mean.** The producer opens with a cross-rank
  barrier, so rank skew lands inside the kernel's own duration. Without a
  per-iteration realignment, m = 1 appeared to cost as much as m = 128. Adding a
  barrier per iteration and reporting the median brought repeat runs to within
  0.24 us of each other, against a mean that swung by a factor of 2.4.
- **A/B against a same-configuration A/A run.** The two GPU groups used for the
  two arms differ by 0.16–0.34% on their own, which is larger than the effect
  being measured. Running A/A, A/B and the swapped A/B gives the arm bias and
  the effect separately.

### DIFF
--- a/csrc/include/custom_all_reduce.cuh
+++ b/csrc/include/custom_all_reduce.cuh
@@ -24,6 +24,7 @@
 #include <iostream>
 #include <limits>
 #include <map>
+#include <cstdlib>
 #include <string>
 #include <type_traits>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


stefanskiasan:
Independent measurement on MI355X (gfx950), in case it is useful: **the wide kernel is exactly as fast as the stock one for our shape**, so on this workload the producer looks neither latency- nor bandwidth-bound.

**Setup.** vLLM with GLM-5.3 (`GlmMoeDsaForCausalLM`, MXFP4 experts + FP8 attention), TP4, hidden 6144, bf16, fp8 KV cache, MTP k=3, 64 concurrent requests on a real-text dataset. aiter 0.1.21. The patch applied cleanly (5 of 6 hunks; the sixth is the `#include <cstdlib>`, which our tree already had). Since `custom_all_reduce` is JIT-built, this was two Docker layers.

**Both kernels present and dispatched**, confirmed from the torch profile and from `AITER_AR_RS_WIDE` in `/proc/<worker>/environ`:

| | stock | `AITER_AR_RS_WIDE=1` |
|---|---|---|
| kernel in trace | `reduce_scatter_cross_device_store` | `…_cross_device_store_wide` |
| total | 269.4 ms / 3200 calls | 269.0 ms / 3200 calls |
| **per call** | **84.2 µs** | **84.1 µs** |
| end-to-end throughput (2 rounds) | 2396 / 2695 tok/s | 2282 / 2654 tok/s |
| GSM8K (150 items) | 147/150 | 147/150 |

Accuracy is unaffected, as expected from the byte-identical traffic.

**Why I expected a difference in either direction.** For our decode shape (m=64, hidden 6144, pack_size 8, ngpus 4) `part` is 12288 packs, so the grids differ a lot:

* stock: `baseline_grid_elems/NGPUS/THREAD_NUM` = 192, clamped to `kMaxBlocks` = **80 blocks**, each retiring `THREAD_NUM/ngpus` = 128 packs per trip;
* wide: `12288/512` = **24 blocks**, each retiring 512.

On a 256-CU part that is 31 % versus 9 % of the CUs, and the wide variant additionally drops the LDS staging and both `__syncthreads()`. Neither effect moved the number by more than 0.1 µs.

A back-of-the-envelope check says the kernel is far from any bandwidth limit too: ~1.5 MB read plus written in 84 µs is about 18 GB/s.

**What the time actually is, at least here:** waiting for the last rank. Per-rank kernel times from the same profile — every compute kernel matches across ranks, only the collectives diverge:

| kernel | rank 0 | rank 1 | rank 2 | rank 3 |
|---|---|---|---|---|
| `reduce_scatter_cross_device_store` | 269.4 | 316.9 | 313.3 | **139.2** |
| MoE GEMM (`mfma_moe1_silu_mul_afp4_wfp4`) | 198.6 | 195.8 | 196.9 | 196.6 |
| MLA attention | 93.1 | 93.7 | 93.8 | 93.7 |

The lagging rank is whichever one the host scheduler happens to delay — pinning each worker process to its own cores moved the laggard from rank 3 to rank 1 and cut the spread from 177.7 ms to 113.5 ms.

So this is not an argument against the patch — it removes real work and real synchronisation, and on a setup where the ranks arrive together it should show. It just means the producer was not the limit for us, and `AITER_AR_RS_WIDE_BLOCKS` reports "flat" in the sense your comment describes.

One note on measuring this on ROCm: with `librocprofiler-sdk` **1.1.0** the torch profile is unusable — `hipGraphLaunch` inflates and the HIP graph fast path is disabled, which showed up as 77 % idle time between kernels in our traces. All numbers above are with **1.3.2** (vllm-project/vllm#53182), where idle time drops to 2–6 %.

stefanskiasan:
Small correction to the last paragraph of my comment above, since it might otherwise read as a recommendation: I wrote that pinning each worker to its own cores moved the laggard and cut the spread from 177.7 ms to 113.5 ms. That profile observation holds, but pinning turned out **not** to be a usable lever for us — on our second instance (identical image and config, different NUMA node) the same pinning is never faster and sporadically collapses by 43 %. We reverted it.

That does not change any of the `AITER_AR_RS_WIDE` numbers, which were all taken **without** pinning: 84.2 µs stock versus 84.1 µs wide, 3200 calls each. The point that the producer was not our limit stands — I just no longer have a fix for what the limit is.

PR #4991, opened 2026-08-25
[HIP] opt inverse rope group quant gfx1250

yzhou103:
## Motivation

Follows #4806. The payload now goes through the gfx1250 tensor engine, the
tier orders apply to row-major, and the op carries one stated policy for
non-finite input. Also fixes an out-of-bounds LDS write, a SIGABRT that
#4806's layout gate introduced, and a dispatch band that was removed by
mistake

## Technical Details

1. Tier sort extended to row-major narrow slices. The orders were restricted
to n32k4, but row-major admits the same sort and it is worth 13-18% once
the slice is narrow. The gain is the narrow quantize path -- a pass whose
groups all miss the rotary tail needs no widening, no cos/sin and no f32 --
which only a uniform segment per wave can select statically. Gated on
TDS == 16: below that a wave spans too few groups to amortize the sort.
2. Payload staged through the tensor engine. Every access order this kernel
uses is a regular 2D tile, tiering included, so the window is
dim0 = run*GROUP_SIZE by dim1 = slots/run and the tier only moves the
origin. Most of the effect is the staged wait: s_wait_tensorcnt<0>
serialises every tile behind the first consumer.
3. WAR stalls removed. s_wait_xcnt was 45% of all stall cycles in the s=16384
ATT trace -- three exec-mask writes landing on a store still in flight:
store_scale's saveexec around a one-byte store whose lanes already agree on
address and value, the cos/sin load's lane-term guard, and a wave-uniform
tdm_wave that divergence analysis could not see through.
4. Four-group heads tier at the exact cut. At GS=128 rope lives only in group
3, so cutting at GROUPS_PER_HEAD/2 filed the pure-nope group 2 under rope
and it paid f32 widening for a rotation it never performs. Held to four
groups per head; the same correction at eight costs 11.7%.

Each scale layout is built only where it can run. mfma_tile feeds CDNA's
V_MFMA_SCALE and n32k4 feeds RDNA's WMMA scaleB, so a third of the module
was kernels the target could never launch.
  gfx1250   372 -> 264 instantiations, 5.31 -> 3.88 MiB device code
  gfx950    372 -> 264 instantiations, 5.31 -> 3.27 MiB device code
  build     26.6 -> 22.3 s
The gate lives entirely in the translation unit; both macros keep an
#ifndef escape hatch for a target that genuinely needs both.

## Test Plan

384 checks pass on gfx1250, HIP graph capture/replay included. gfx942 is
added to SUPPORTED_GFX: its exclusion cited permlane16_swap, which the amax
reduction stopped using, and the op builds and runs there with mfma_tile
included.
Two gaps the sweep could not reach are now covered. Rows whose Ks leaves a
part-wave block, at 9 and 3 heads -- every default shape is a whole number
of waves. And the invalid-group policy, since none of the DATA_DISTS can
produce a non-finite input, which is why that behaviour had moved twice
under optimisation with nothing watching it.

## Test Result

gfx1250, h,g = 128,16, against main. Four interleaved rounds -- both builds
measured inside each round, since absolute timings drift between batches and
only a ratio taken within a round holds up. Median of the per-round ratios:
                          main      branch
  row   GS=128  s=2048   36.18 us   24.89 us   -31.2%
  row   GS=128  s=16384 295.59 us  211.08 us   -28.6%
  row   GS=32   s=2048   35.93 us   27.04 us   -24.7%
  row   GS=128  s=512     7.99 us    6.27 us   -21.3%
  row   GS=32   s=16384 293.00 us  237.78 us   -18.8%
  n32k4 GS=32   s=2048   33.12 us   29.86 us    -9.8%
  n32k4 GS=32   s=16384 261.91 us  238.36 us    -8.9%
The two remaining s=512 points, row/GS=32 and n32k4/GS=32, land around -5%
but spread several points round to round, so they are quoted that way rather
than to a figure.
A wider sweep -- 2 head configs x 3 layout/group-size combinations x 6
values of S -- is uniformly in the same direction with no regression, but it
was collected one build after the other and so carries that batch drift;
treat it as direction rather than magnitude.

## Submission Checklist


yzhou103:
image

zufayu:

**Three gfx1250-focused changes to the DeepSeek-V4 attention-output inverse-RoPE + FP8 group-quant kernel: a corrected nope/rope tier cut for four-group heads, per-family build gates that compile the wrong-family scale layout out of the device image, and TDM staging of the payload into dynamic LDS.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0) and no report supplied; the only GPU on this box is gfx942, which this module does not build for (the test's SUPPORTED_GFX is gfx950/gfx1250/gfx1201 — the DPP reduction's permlane steps are gfx950+), so no target could run here either way.
Perf (advisory): NOT RUN — the box's only GPU is gfx942 and the module requires gfx950/gfx1250/gfx1201; the TDM staging and tier changes are gfx1250-only. The PR's own table (gfx1250, 5-run medians, six configs, base vs head) is the only perf evidence and was collected in batches.

🔴 [verified] The TDM staging overflows the dynamic LDS allocation on partial-wave blocks. csrc/kernels/inverse_rope_group_quant.cu:1428 sizes the LDS at block_size*TDS*KPT elements, but the tile the DMA writes is wave-sized (kSlotsPerWave*GROUP_SIZE elements per wave per pass, :629/:675) — the two agree only when block_size is a multiple of wave_size. The k_slots halving loop (:1662) produces partial-wave blocks whenever scale_n % kSlotsPerWave != 0: at H=96/G=64/GS=32/row (D=768, scale_n=24) a 16-thread block's 512-element tile lands in a 256-element allocation, and at H=33/G=16/GS=32/row a 1-thread block writes a 2048-byte tile into 64 bytes of LDS; simulating the host dispatch over all legal (H,G,group_size,layout,S) configs, 5838 of 11632 overflow. The DMA's write is bounded only by the tensor shape (the row), not the block, so bytes past the allocation land in a co-resident workgroup's LDS on the same CU — silently corrupting that block's staged payload and its quantized output. Every shape the PR and its test run (D=4096 throughout) divides evenly, so the test cannot catch it. **Author must** gate the staging on k_slots % kSlotsPerWave == 0 (in tdm_used and the kernel) or size the LDS by ceil(block_size/wave_size)*kSlotsPerWave*GROUP_SIZE*KPT elements.

⚠️ [verified] The exec-mask removals and the packed rotation are in the shared kernel body, not behind the gfx1250 gate: store_scale now stores from every lane of a group (csrc/kernels/inverse_rope_group_quant.cu:825), the cos/sin load is unguarded off a clamped row (:984), and the rotation is f32x2 (:1022) — all of these compile into gfx950's wave64 row/mfma_tile kernels too, where the xcnt-drain argument was never measured and nope lanes now issue cos/sin loads (row 0) they previously skipped entirely, since gfx950 has no native-quant path to early-continue them. The automatic gfx950 op-test CI covers correctness there, but the PR quotes no gfx950 timing. **Author must** run the op test's timing on gfx950 (row + mfma_tile) base vs head and confirm the mask removals do not regress wave64.

⚠️ [verified] inverse_rope_group_quant is the op ATOM's DeepSeek-V4 attention-output path calls (the test's unfused baseline mirrors atom's _V4RoPE.inverse), and this diff changes its default kernel on every arch it builds for — but the PR carries no ci:* label, and every downstream roster quoted in its CI guide (ci:atom: DeepSeek-R1-0528/GPT-OSS-120B; ci:sglang: DSv3-MXFP4/Qwen 3.5; ci:vllm: GPT-OSS/DSv3/Kimi) names no DSv4 model, so it is unclear any downstream job exercises the changed path at all. **Reviewer should ask** which downstream job covers the DSv4 attention-output path on gfx950/gfx1250 and require it green before merge.

📝 [verified] The build gates change the public op's contract: n32k4 on CDNA and mfma_tile on RDNA now raise ValueError (aiter/ops/inverse_rope_group_quant.py:43) where they previously returned a scale tensor, and 108 kernel instantiations per wrong family are gone from the device image. The rejection is tested (check_layout_rejected) and the layouts have no consumer on the wrong family, but a C++-level caller still hits an abort rather than a raise. **Reviewer should ask** whether any downstream caller passes the wrong-family layout today.

📝 [verified] The PR's measurement note is truncated mid-sentence ("with interleaved A/B on an idle GPU moved from -2.2 is a lower bound") and the table itself was collected in batches (all-before, then all-after), which the note admits makes the quoted deltas lower bounds. **Author must** complete the note and re-quote the six configs from an interleaved A/B run on an idle gfx1250.

yhl-amd (dismissed):
fully test on mi450, LGTM

valarLip (dismissed):

PR #4892, opened 2026-08-20
[Triton/Gluon] Moe A8W4 tuning

lburzawa:
## Motivation

Tune moe for deepseek shapes.

## Technical Details

Add persistent kernel option to configs for each shape and do further tuning on them.

## Test Plan

Correctness and performance.

## Test Result

UTs pass and performance improves on targeted shapes.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4836, opened 2026-08-18
[Triton/Gluon] [GFX1250] add triton support for gfx1250 MoE a8w4 and a4w4

nsusanto:
- Support triton backend for both a8w4 and a4w4 MoE kernels in gfx1250
    - by default use gluon kernel for 1250 and triton kernel for 950
    - 1250 can use both triton or gluon kernel by passing in `--backend` arg but 950 can only use triton backend
- Added repr to both a8w4 and a4w4 MoE

## Submission Checklist

- [ x ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4562, opened 2026-08-04
[GFX1250] [Gluon] Add multicast support for MoE a8w4

nsusanto:
## Motivation

Implemented multicast by launchinf CTAs as a cluster CGA a single TDM load into the LDS of every CTA in that cluster. 

Multicast is off by default (num_ctas = 1).

## Technical Details
- Layouts moved from the kernels to a host-side helper as _make_cga_layout_ for multicast can't be called in a triton jit function and it requires a layout already made.
- get_gluon_a8w4_ctas_per_cga decides the cluster shape:
    - decode (m < 1024): shard along N only  [1, num_ctas]
    - prefill (m >= 1024): shard both axes -> [2, 2], 8 -> [2, 4], 16 -> [4, 4]
    - returns [1, 1] when num_ctas == 1 or the arch is not gfx1250
- block_m accounting and the new tile_m_scale routing argument
    - Routing now contains tile_m_scale to scale block_m based on num_ctas[0]

## Test Result

Passes all UT but was not giving perf improvements for decode and prefill. Needs support to get expected perf improvements in the future.

## Submission Checklist

- [ x ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4470, opened 2026-07-30
[GLUON] Add more fine grained tuning based on M

lburzawa:
## Motivation

Add more fine grained tuning based on M/num_gates. Idea and initial PR by @amd-hhashemi 

## Technical Details

Separate M into buckets and tune block_n/block_k/num_buffers using json configs.

## Test Plan

Correctness and performance.

## Test Result

UTs pass and performance improves on tuned configs.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4993, opened 2026-08-25
[Triton/Gluon] Fix PR#4562 and PR#4470 for GPT-OSS-120b on gfx1250

sogalin:
## Problem
On gfx1250, the Gluon MoE a8w4 path returns garbage for GPT-OSS-120B
(MXFP4 W4A8). GSM8K via SGLang, 200 questions: accuracy 0.010, invalid 0.990.
## Root cause
Two independent regressions:
1. **#4562** hoisted the per-kernel WMMA/shared layouts into
   `get_moe_a8w4_layouts()`. Its `num_warps == 4` branch spreads all warps
   along N (`[[0, 1], [0, 2]]`), but this helper only feeds
   `_moe_gemm_a8w4_prefill`, which previously hardcoded `[[0, 1], [1, 0]]`.
   Dropping the M basis corrupts the output. The 8-warp prefill branch kept
   its M basis, so this looks like an oversight.
2. **#4470** replaced the shape-derived Gluon config with a tuned lookup that
   falls back to `bm{block_m}_any`. Those generic entries force
   `BLOCK_K=256`, while the previous logic used `BLOCK_K=512` for
   `block_m <= 32`. GPT-OSS shapes (N=5760/2880, K=2880) have no tuned entry,
   so both GEMMs land on the fallback.
## Fix
1. Restore `[[0, 1], [1, 0]]` for `num_warps == 4`.
2. For untuned shapes, fall back to the shape-derived config instead of
   `bm*_any`. Shapes with a tuned entry are unaffected.
## Validation
gfx1250, GPT-OSS-120B MXFP4 W4A8, SGLang GSM8K, 200 questions @ parallel 1319:
| Build | Accuracy | Invalid |
| --- | --- | --- |
| before | 0.010 | 0.990 |
| fix 1 only | 0.725 | 0.060 |
| fix 1 + 2 | **0.865** | **0.005** |
| revert #4470 + #4562 + #4892 + #4836 | 0.860 | 0.005 |

This matches a full revert of both regressions and their dependents, while
keeping the a4w4 Triton backend that #4836 added.

### DIFF
--- a/aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
+++ b/aiter/ops/triton/_gluon_kernels/gfx1250/moe/moe_op_gemm_a8w4.py
@@ -1709,7 +1709,9 @@ def get_moe_a8w4_layouts(
         warp_bases = [[0, 1]]
         reg_bases = []
     elif num_warps == 4:
-        warp_bases = [[0, 1], [0, 2]]
+        # This helper only feeds the prefill kernel, which needs a warp basis
+        # along M; distributing all four warps along N produces garbage output.
+        warp_bases = [[0, 1], [1, 0]]
         reg_bases = []
     else:
         warp_bases = (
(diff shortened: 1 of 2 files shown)


lburzawa:
We found this to be an issue with upstream triton compiler. Here is the ticket https://github.com/AMD-Triton/triton-tickets/issues/1983.

PR #4994, opened 2026-08-25
[AMD][DSV4] Fix(fmoe): fuse stage-1 fp8 quant on the heuristic FlyDSL fallback

karverma-amd:
## Summary

`get_2stage_cfgs`' tuned path selects the fused `_fp8` stage-1 variant per shape, folding stage 2's mxfp8 quant and scale-sort into stage 1's CShuffle epilogue. Stage 1 then writes one byte per element instead of two, and the standalone quant+sort kernel disappears entirely. Most rows in `dsv4_fp8fp4_tuned_fmoe.csv` carry that suffix.

The **heuristic fallback never considered it**, so any shape without a tuned row paid for the separate conversion even though the fused variant exists for every tile. This makes the fallback consistent with the tuned path.

The pairing is safe: stage 2 already consumes fp8 (`kn2='flydsl_moe2_afp8_...'`, where `afp8` is the A-input dtype), so promoting stage 1 to emit fp8 directly removes the intermediate conversion rather than changing what stage 2 expects. The change is gated on `get_flydsl_kernel_params()` resolving the fused name, so it is a no-op wherever that variant is absent, and scoped to `_a_type == "fp8"` so fp4/bf16 paths are untouched.

## Why this matters

DeepSeek-V4 at EP8 takes the fallback. From a server log on MI355X:

```
[fused_moe] no tuned FlyDSL config for ('gfx950', 256, 32768, 7168, 3072, 49, 6, ...),
using heuristic FlyDSL fallback (kn1='flydsl_moe1_afp8_wfp4_bf16_t64x128x256_w4_bnt0_gui_fp8',
kn2='flydsl_moe2_afp8_wfp4_bf16_t64x128x256_atomic', fuse_quant=fp8)
```

Before this change that shape ran a standalone per-group quant kernel on every MoE layer, measured at **116 us per layer, 8.8% of decode GPU busy time**.

## Measured

DeepSeek-V4-Pro, MI355X, TP8/EP8/DP8, mxfp4 weights + mxfp8 activations, ISL 8192 / OSL 1024. Both arms are the **same build**, toggled only by `AITER_S1_FUSE_FP8Q`, and each arm's `fuse_quant=` selection was verified in its server log on all 8 ranks.

| concurrency | unfused tok/s | fused tok/s | gain | TPOT unfused -> fused |
|---|---|---|---|---|
| 4  | 42.92  | 56.95  | **+32.7%** | 91.26 -> 68.50 ms (-24.9%) |
| 8  | 84.14  | 111.39 | **+32.4%** | 92.90 -> 69.86 ms (-24.8%) |
| 16 | 160.34 | 210.35 | **+31.2%** | 96.68 -> 73.25 ms (-24.2%) |
| 32 | 299.99 | 387.00 | **+29.0%** | 101.62 -> 78.14 ms (-23.1%) |
| 64 | 552.36 | 702.52 | **+27.2%** | 107.17 -> 83.27 ms (-22.3%) |

GSM8K (1319 questions): **0.947 fused vs 0.944 unfused**, 0 invalid -- no numerical regression.

## Why fused unconditionally

The tuned table disagrees by `inter_dim` (at 384 all 11 rows chose unfused; at 2048 and 3072 nearly all chose fused), which suggests a floor. A direct sweep of `inter_dim` 384..3072 found **no crossover**: end-to-end `fused_moe` was faster fused in 14 of 16 cells (-0.4% to -4.2%), and the two exceptions (+1.0%, +0.1%) were within run-to-run spread.

The table's preference for unfused at 384 reflects its joint stage1+stage2 choice at its own shapes, not a standalone loss for the fused epilogue: timing stage 1 in isolation charges the fused arm for its epilogue while never charging the unfused arm for the quant kernel that fusion deletes.

The concurrency sweep above corroborates this from the other direction. The gain rises monotonically as concurrency falls, which is what the mechanism predicts if the eliminated kernel is a fixed per-layer cost -- so the low-token regime where fusion has least room to pay off is in fact where it pays off most.

`AITER_S1_FUSE_FP8Q=0` disables it; `AITER_S1_FUSE_FP8Q_MIN_INTER_DIM` reimposes a floor, with `=1` forcing fusion below that floor.

## Test plan

`op_tests/test_fused_moe_s1_fuse_fp8q.py`, 4 passed on gfx950:

- the fallback promotes stage 1 to the fused fp8 variant
- `AITER_S1_FUSE_FP8Q=0` restores the unfused stage 1
- a floor above `inter_dim` opts the shape out
- `=1` overrides the floor

The tests force the fallback by emptying the module-level `cfg_2stages` cache rather than picking a shape that happens to be untuned, so they cannot quietly stop exercising the fallback if that shape is later tuned. They also clear `get_2stage_cfgs`' `lru_cache` per test, since the env knobs are not part of the cache key -- without that, every case after the first asserts against the first case's metadata.

Note: the two commits can be squashed on merge; the second fixes defects in the test added by the first.


yadaish:
why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.

karverma-amd:
> why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.

Because --enforce-shared-experts-fusion makes the local expert count 49 (48 routed at EP8 + the fused shared expert, topk 6+1), and there is no E=49 row in dsv4_fp8fp4_tuned_fmoe.csv. The table's four expert values are 48, 256, 384, 385 — so it covers shared-expert fusion only at EP1 (385/7) and EP8 only without fusion (48/6). EP8 with fusion is the configuration the DSv4 recipe actually ships (+21.6% on its own), and it's the one that was never tuned. Separately, the E=48 rows stop at token=16384 while this shape is token=32768, so the bucket misses too. dsv4_fp8fp4_untuned_fmoe.csv has no expert=49 rows at all, which is why the sweep never produced any.

Adding EP8 rows is a good idea and I'm happy to do it — about 15–17 rows for (49, 7, 7168, 3072) over token 1…32768. I'd still keep this patch, for three reasons. The token bucket is a function of concurrency, draft width and phase, so the table will always have holes and the fallback shouldn't be silently worse than the tuned path when the fused variant exists for every tile. The fused choice also isn't shape-dependent here: sweeping inter_dim 384…3072 found no crossover (14/16 cells faster fused, the two exceptions inside run-to-run spread). And the two changes address different things — the table picks a joint stage1+stage2 pair per shape, while this makes the fallback stop paying for a quant kernel the fused epilogue removes.

Happy to add the EP8 tuned rows in this PR or a follow-up, whichever you prefer.

aiter/fused_moe.py
```diff
@@ -75,6 +75,11 @@ def is_flydsl_available():
 _ACT_TYPE_DISABLED_KEY = "__ignore__"
 _SWIGLU_MXFP4_BF16_BOUND = int(os.environ.get("GPTOSS_SWIGLU_MXFP4_BF16_BOUND", "256"))
 _MOE_A8W4_BYPASS_QUANT = os.environ.get("AITER_MOE_A8W4_BYPASS_QUANT", "0") == "1"
+# Smallest inter_dim at which the heuristic fallback may fuse stage 2's mxfp8
+# quant into stage 1's epilogue. 0 means always; see get_2stage_cfgs for why.
+_S1_FUSE_FP8Q_MIN_INTER_DIM = int(
```
yadaish: What about using env to enable/disable a8 requant on/off,  _S1_FUSE_FP8Q_MIN_INTER_DIM env is only one factor to take effect on requant.

aiter/fused_moe.py
```diff
@@ -2781,9 +2786,50 @@ def get_block_m() -> int:
         if get_flydsl_kernel_params(kn2) is None:
             kn2 = _base_kn2
 
+        # Fold stage 2's mxfp8 quant and scale-sort into stage 1's epilogue.
```
yadaish: I think the comment is useless if we already have clear env name

aiter/fused_moe.py
yadaish: And you should give the env one clear name to make other developers understand the meaning. If so, redundant comment should be removed.

yadaish:
> > why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.
> 
> Because --enforce-shared-experts-fusion makes the local expert count 49 (48 routed at EP8 + the fused shared expert, topk 6+1), and there is no E=49 row in dsv4_fp8fp4_tuned_fmoe.csv. The table's four expert values are 48, 256, 384, 385 — so it covers shared-expert fusion only at EP1 (385/7) and EP8 only without fusion (48/6). EP8 with fusion is the configuration the DSv4 recipe actually ships (+21.6% on its own), and it's the one that was never tuned. Separately, the E=48 rows stop at token=16384 while this shape is token=32768, so the bucket misses too. dsv4_fp8fp4_untuned_fmoe.csv has no expert=49 rows at all, which is why the sweep never produced any.
> 
> Adding EP8 rows is a good idea and I'm happy to do it — about 15–17 rows for (49, 7, 7168, 3072) over token 1…32768. I'd still keep this patch, for three reasons. The token bucket is a function of concurrency, draft width and phase, so the table will always have holes and the fallback shouldn't be silently worse than the tuned path when the fused variant exists for every tile. The fused choice also isn't shape-dependent here: sweeping inter_dim 384…3072 found no crossover (14/16 cells faster fused, the two exceptions inside run-to-run spread). And the two changes address different things — the table picks a joint stage1+stage2 pair per shape, while this makes the fallback stop paying for a quant kernel the fused epilogue removes.
> 
> Happy to add the EP8 tuned rows in this PR or a follow-up, whichever you prefer.

I think adding EP8 csv rows are better option and what you 

> > why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.
> 
> Because --enforce-shared-experts-fusion makes the local expert count 49 (48 routed at EP8 + the fused shared expert, topk 6+1), and there is no E=49 row in dsv4_fp8fp4_tuned_fmoe.csv. The table's four expert values are 48, 256, 384, 385 — so it covers shared-expert fusion only at EP1 (385/7) and EP8 only without fusion (48/6). EP8 with fusion is the configuration the DSv4 recipe actually ships (+21.6% on its own), and it's the one that was never tuned. Separately, the E=48 rows stop at token=16384 while this shape is token=32768, so the bucket misses too. dsv4_fp8fp4_untuned_fmoe.csv has no expert=49 rows at all, which is why the sweep never produced any.
> 
> Adding EP8 rows is a good idea and I'm happy to do it — about 15–17 rows for (49, 7, 7168, 3072) over token 1…32768. I'd still keep this patch, for three reasons. The token bucket is a function of concurrency, draft width and phase, so the table will always have holes and the fallback shouldn't be silently worse than the tuned path when the fused variant exists for every tile. The fused choice also isn't shape-dependent here: sweeping inter_dim 384…3072 found no crossover (14/16 cells faster fused, the two exceptions inside run-to-run spread). And the two changes address different things — the table picks a joint stage1+stage2 pair per shape, while this makes the fallback stop paying for a quant kernel the fused epilogue removes.
> 
> Happy to add the EP8 tuned rows in this PR or a follow-up, whichever you prefer.

I think  EP8 rows shou be added also.

karverma-amd:
@yadaish - Good points — done. Collapsed the two knobs into a single on/off switch AITER_MOE_FUSE_STAGE1_FP8_QUANT (default on; =0 keeps the separate quant kernel) and dropped the _MIN_INTER_DIM floor entirely — the inter_dim sweep found no crossover, so the floor was dead weight rather than a real tuning axis. With the self-explanatory name the long
 rationale block is redundant, so I trimmed it to two lines. Net −49 lines. Tests updated to the single switch (the two floor tests are gone); both pass on gfx950. Pushed

karverma-amd:
> > > why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.
> > 
> > 
> > Because --enforce-shared-experts-fusion makes the local expert count 49 (48 routed at EP8 + the fused shared expert, topk 6+1), and there is no E=49 row in dsv4_fp8fp4_tuned_fmoe.csv. The table's four expert values are 48, 256, 384, 385 — so it covers shared-expert fusion only at EP1 (385/7) and EP8 only without fusion (48/6). EP8 with fusion is the configuration the DSv4 recipe actually ships (+21.6% on its own), and it's the one that was never tuned. Separately, the E=48 rows stop at token=16384 while this shape is token=32768, so the bucket misses too. dsv4_fp8fp4_untuned_fmoe.csv has no expert=49 rows at all, which is why the sweep never produced any.
> > Adding EP8 rows is a good idea and I'm happy to do it — about 15–17 rows for (49, 7, 7168, 3072) over token 1…32768. I'd still keep this patch, for three reasons. The token bucket is a function of concurrency, draft width and phase, so the table will always have holes and the fallback shouldn't be silently worse than the tuned path when the fused variant exists for every tile. The fused choice also isn't shape-dependent here: sweeping inter_dim 384…3072 found no crossover (14/16 cells faster fused, the two exceptions inside run-to-run spread). And the two changes address different things — the table picks a joint stage1+stage2 pair per shape, while this makes the fallback stop paying for a quant kernel the fused epilogue removes.
> > Happy to add the EP8 tuned rows in this PR or a follow-up, whichever you prefer.
> 
> I think adding EP8 csv rows are better option and what you
> 
> > > why the dsv4 EP8 falls into fallback path? What about improving the tuned csv to support EP8.
> > 
> > 
> > Because --enforce-shared-experts-fusion makes the local expert count 49 (48 routed at EP8 + the fused shared expert, topk 6+1), and there is no E=49 row in dsv4_fp8fp4_tuned_fmoe.csv. The table's four expert values are 48, 256, 384, 385 — so it covers shared-expert fusion only at EP1 (385/7) and EP8 only without fusion (48/6). EP8 with fusion is the configuration the DSv4 recipe actually ships (+21.6% on its own), and it's the one that was never tuned. Separately, the E=48 rows stop at token=16384 while this shape is token=32768, so the bucket misses too. dsv4_fp8fp4_untuned_fmoe.csv has no expert=49 rows at all, which is why the sweep never produced any.
> > Adding EP8 rows is a good idea and I'm happy to do it — about 15–17 rows for (49, 7, 7168, 3072) over token 1…32768. I'd still keep this patch, for three reasons. The token bucket is a function of concurrency, draft width and phase, so the table will always have holes and the fallback shouldn't be silently worse than the tuned path when the fused variant exists for every tile. The fused choice also isn't shape-dependent here: sweeping inter_dim 384…3072 found no crossover (14/16 cells faster fused, the two exceptions inside run-to-run spread). And the two changes address different things — the table picks a joint stage1+stage2 pair per shape, while this makes the fallback stop paying for a quant kernel the fused epilogue removes.
> > Happy to add the EP8 tuned rows in this PR or a follow-up, whichever you prefer.
> 
> I think EP8 rows shou be added also.

Agreed the tuned rows are the better long-term fix — DSV4 at EP8 (expert 49) has no matching entry today, so it takes this fallback. Since the tuning table is orthogonal to this fallback-correctness fix, I'd like to keep this PR scoped and add the EP8 rows in a dedicated follow-up PR shortly: (expert=49, topk=7, model_dim=7168, inter_dim=3072) swept over token 1…32768 (~15–17 rows). That keeps the tuning data reviewable on its own and lets this fix merge independently.

yadaish:
LGTM

zufayu (approved):

zufayu (approved):
Reviewed the 15-line change. The gating is what makes this safe: it fires only when `_a_type == "fp8"`, the switch is on, and `get_flydsl_kernel_params(f"{kn1}_fp8")` actually resolves — so it is a strict no-op wherever the fused variant is absent, and the tuned path is untouched. `AITER_MOE_FUSE_STAGE1_FP8_QUANT=0` is a clean kill switch, and the added `fuse_quant=` in the fallback warning makes the selection auditable from a server log after the fact.

CI is green (35 success, 0 failures) and no review threads are open.

One doc nit, not a blocker: the description says both arms were "toggled only by `AITER_S1_FUSE_FP8Q`", but the variable in the code is `AITER_MOE_FUSE_STAGE1_FP8_QUANT`. Renamed at some point and the description did not follow — anyone reproducing the measurement from the PR text will set a variable that does nothing. Worth fixing the description so the numbers stay reproducible.

PR #4996, opened 2026-08-25
[CI] [JIT] [Build] [Bugfix] Build pybind modules against torch's bundled pybind11; drop pip pybind11 dep

### LINKED ISSUE #4770: [Bug] JIT-built module_fmha_fwd_bf16_opus rejects tensors from prebuilt module_aiter_core: pybind11 internals mismatch (v11 wheel vs v12 runtime JIT)
# [Bug] JIT-built `module_fmha_fwd_bf16_opus` rejects tensors from prebuilt `module_aiter_core`: pybind11 internals mismatch (v11 wheel vs v12 runtime JIT)

## Summary

On images that ship prebuilt aiter binaries (e.g. `vllm/vllm-openai-rocm:nightly`, amd-aiter 0.1.19) together with pip `pybind11 3.1.0`, any module that aiter JIT-compiles at runtime is built against pybind11 internals **v12**, while the wheel's prebuilt modules were built against internals **v11**. pybind11 does not share registered types across modules with different internals IDs, so the JIT module registers its own `aiter_tensor_t` and rejects the `aiter_tensor_t` objects produced by `module_aiter_core`:

```
TypeError: fmha_fwd_bf16_opus_fwd(): incompatible function arguments. The following argument types are supported:
    1. (q: aiter_tensor_t, k: aiter_tensor_t, v: aiter_tensor_t, out: aiter_tensor_t, ...)
Invoked with: <aiter.jit.module_aiter_core.aiter_tensor_t object at ...>, ...
```

The call is argument-correct; only the pybind type identity differs.

Currently the one affected module in practice is `module_fmha_fwd_bf16_opus` (not in the wheel's gfx950 prebuild set, so it JIT-builds on first use). This has blocked serving Kimi-K3/K2 on `vllm/vllm-openai-rocm:nightly` since the 2026-08-11 nightly: MLA prefill backend `ROCM_AITER_FA` calls `fmha_fwd_bf16_opus_varlen_fwd` and dies during the warmup prefill. Workaround in vLLM: `--attention-config '{"mla_prefill_backend": "FLASH_ATTN"}'`.

## Evidence

Internals strings of the two binaries inside the nightly image:

```
$ strings aiter/jit/module_aiter_core.so | grep pybind11_internals
__pybind11_internals_v11_system_libstdcpp_gxx_abi_1xxx_use_cxx11_abi_1__     # prebuilt (wheel)

$ strings aiter/jit/module_fmha_fwd_bf16_opus.so | grep pybind11_internals
__pybind11_internals_v12_system_libstdcpp_gxx_abi_1xxx_use_cxx11_abi_1__     # JIT-built at runtime
```

Header versions available in the image:

- torch-bundled pybind11 (`torch/include/pybind11/detail/internals.h`): `PYBIND11_INTERNALS_VERSION 11`
- pip `pybind11 3.1.0` (`pybind11/include/.../internals.h`): `PYBIND11_INTERNALS_VERSION 12` ("the minimum for pybind11 v3.1.0")

## Root cause

`aiter/jit/utils/cpp_extension.py` (`_write_ninja_file_and_build_library` path, ~line 1662):

```python
    # FIXME: build python module excluded with torch, use `pybind11`
    # But we can't use this now because all aiter op based on torch
    # which means pybind11 related build flags must from torch now
    common_cflags = []
    if is_python_module:
        import pybind11
        extra_include_paths.append(pybind11.get_include())
```

The pip pybind11 include is appended to the *user* include paths, which take precedence over the torch system includes — so whenever the environment's pip pybind11 generation differs from the one the wheel binaries were built with, every runtime-JIT'd module lands in an incompatible internals world. The existing FIXME comment already notes that pybind11 flags must come from torch.

This became load-bearing recently because (a) the detorch refactor made `aiter_tensor_t` — a pybind-registered class — the cross-module currency (previously `torch::Tensor` flowed through torch's casters), and (b) pybind11 3.1.0 bumped internals v11 → v12.

**What actually pulled the trigger — no aiter or vLLM code change at all.** `pybind11` is installed as
amd-aiter's own *unpinned* runtime dependency (`pip show pybind11` → `Required-by: amd-aiter`).
pybind11 3.1.0 was released on PyPI on **2026-08-06**; every image dependency resolution after that
date installs a pybind11 that is ABI-incompatible with amd-aiter's own prebuilt binaries. First
broken nightly observed 2026-08-11. Before 08-06, the unpinned dep resolved to 3.0.x (internals
v11, matching the wheel), so the include-precedence flaw was benign.

## Repro

On a gfx950 host:

```bash
docker run --rm --device /dev/kfd --device /dev/dri vllm/vllm-openai-rocm:nightly \
  vllm serve moonshotai/Kimi-K3 --trust-remote-code --tensor-parallel-size 8
# (with VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_MLA=1)
```

fails at the warmup prefill with the TypeError above, after JIT-building `module_fmha_fwd_bf16_opus` (~16 s). Any direct call into `aiter.ops.mha.fmha_fwd_bf16_opus_varlen_fwd` in that image reproduces it without vLLM.

## Suggested fixes

1. **Minimal:** in `jit/utils/cpp_extension.py`, stop appending `pybind11.get_include()` for torch-based python modules — torch's include root (already in the system includes) provides the pybind11 that matches libtorch_python and the wheel binaries.
2. **Packaging:** add `module_fmha_fwd_bf16_opus` to the gfx950 wheel prebuild set so nothing JIT-compiles at runtime (also removes the 16 s cold-boot build).
3. **Robust:** make `aiter_tensor_t` a header-only structural type caster (duck-typed / DLPack extraction) instead of a registered class, so type identity never has to cross module boundaries.

(1) or (2) unblocks immediately; (3) removes the failure mode for good.

## Why this wasn't caught earlier

Four independent conditions had to line up, and no CI exercises the combination:

- aiter CI builds all modules from source in one pass, so everything always shares one pybind11 —
  the bug requires prebuilt wheel binaries meeting a runtime JIT in a *different* environment,
  which only happens in downstream images.
- The include-path precedence flaw is old but was benign: pip pybind11 3.0.x has internals v11,
  same as torch's bundled copy. pybind11 3.1.0 (internals v12) arming the bug only reached the
  nightly image ~2026-08-11 via a dependency bump; nothing in aiter changed that day.
- Only `module_fmha_fwd_bf16_opus` falls through to runtime JIT (everything else is prebuilt for
  gfx950), and its only caller is the MLA prefill path of MLA models at TP8 on gfx950 — not
  covered by image smoke tests.
- The failure mode itself is new: until the detorch refactor, tensors crossed module boundaries
  as `torch::Tensor` via libtorch_python's casters (one shared registration inside torch),
  which is immune to internals mismatch.

## Preventing recurrence

- At JIT time, assert that the current pybind11 headers' internals ID matches the loaded
  prebuilt modules' (readable from the binary), and fail with an explicit message instead of a
  distant `TypeError`.
- Add a CI job that installs the released wheel in a fresh environment, force-JITs one module,
  and calls it with tensors created by a prebuilt module — exactly this scenario.
- Pin `pybind11` (e.g. `pybind11==3.0.*`) in the wheel's dependency metadata so environments
  can't drift from the build toolchain silently.

Rohan138:
## Problem

aiter's JIT builder (`aiter/jit/utils/cpp_extension.py`) appends the **pip-installed** pybind11's include path via `pybind11.get_include()`, emitted as a `-I` **ahead of** torch's bundled headers (`-isystem <torch>/include`). So every compiled pybind module resolves `#include <pybind11/...>` to whatever pybind11 pip has.

pybind11 **3.1.0** (2026-08-06) bumped `PYBIND11_INTERNALS_VERSION` **11 → 12**. Modules built at different pip epochs then disagree on the internals version, and pybind11 does not share registered types across internals versions. The prebuilt `module_aiter_core` (which registers `aiter_tensor_t`) and a JIT module built later end up incompatible:

```
TypeError: fmha_fwd_bf16_opus_fwd(): incompatible function arguments.
  Invoked with: <aiter.jit.module_aiter_core.aiter_tensor_t object at ...>
```

Fixes #4770.

## Root cause detail

`module_aiter_core` is a **torch-free** (`torch_exclude:True`) pybind module — it registers `aiter_tensor_t` and is loaded via `ctypes`. It has no torch headers on its include path, so it previously took its pybind headers **only from pip**. Torch-backed consumer modules (e.g. `fmha`) also took pip's. Because they were compiled at different times, a drifting pip pybind11 desynchronized producer and consumers.

## Fix

torch already bundles pybind11 at `torch/include` (3.0.1 / internals **v11** in torch 2.10), which satisfied aiter's former `pybind11>=3.0.1`. This PR builds **every** python module — torch-backed **and** torch-free — against torch's bundled pybind headers:

- Drop the pip `get_include()` append entirely.
- For `torch_exclude` builds (which otherwise omit `torch/include`), add `torch/include` back **only for header resolution** — pybind is header-only, so the torch-free `.so` gains **no runtime libtorch dependency**.

Now producer (`module_aiter_core`) and all consumers compile against the *same* pybind (torch's), so `aiter_tensor_t` stays ABI-compatible regardless of any pip pybind11. (Vendoring pybind instead would re-break this: torch-backed modules must use torch's pybind to interop with torch types, so everything has to converge on torch's copy.)

With the ABI sourced from torch, **aiter no longer needs a pip pybind11 dependency at all**. Removed from:
- `requirements.txt`, `pyproject.toml` (`build-system.requires`), `setup.py` (`install_requires`)
- the `pip install --upgrade pybind11` steps in CI workflows (`aiter-test`, `atom-test`, `kimi-downstream`, `kimi-perf-downstream`, `vllm_benchmark`) and `.github/requirements/triton-test.txt`

## Test

`rocm/pytorch:latest` (torch 2.10.0+rocm, bundled pybind **v11**), **with no pip pybind11 installed**, building trivial pybind modules through aiter's `load()`:

| Module | `is_python_module` | `torch_exclude` | `.so` internals | links libtorch? |
|---|---|---|---|---|
| torch-backed | True | False | `__pybind11_internals_v11__` ✅ | yes (expected) |
| torch-free (like `module_aiter_core`) | True | True | `__pybind11_internals_v11__` ✅ | **no** ✅ |

Both compile with **zero pip pybind11 present**. For contrast, with `pip install pybind11==3.1.0`, the pre-change code produced `__pybind11_internals_v12__` (the mismatch); this change yields `v11` regardless.


Rohan138:
The fix in this PR (using the pytorch-bundled pybind11) does tie AITER to the corresponding PyTorch version it was built with, but I assume that's already the case for now. Might need to be revisited if we want to make AITER PyTorch-stable.

Also I don't know if the deps should be pinned here in the first place: https://github.com/ROCm/aiter/blob/main/requirements.txt

Or at the very least, the build and install deps should be kept aligned: https://github.com/ROCm/aiter/blob/main/setup.py#L468

If the deps need to be pinned for CI/test purposes, it might be better to have a separate `tests/requirements.txt` file the way we do in vLLM: https://github.com/vllm-project/vllm/blob/main/requirements/test/rocm.txt; the pins right now, if copied over to `setup.py` will e.g. break vLLM CI builds if vLLM bumps/pins a different pandas versions

stefanskiasan:
Data point from the field (aiter **0.1.19** prebuilt wheel, gfx950 / 8x MI355X, ROCm 7.2.3, torch 2.12.0, vLLM ROCm image that ships pip `pybind11 3.1.0`): JIT-building a single module (`module_custom_all_reduce` from #4977) next to the prebuilt wheel modules produced exactly the mismatch described here, at call time:

```
TypeError: all_reduce(): incompatible function arguments
```

(the JIT module's `aiter_tensor_t` carries pybind internals v12, the wheel's `module_aiter_core` v11). Pinning `pybind11==3.0.1` before the JIT build fixed it, and the rebuilt module serves correctly (GLM-5.3 TP4, needle/perplexity checks unchanged). Until this is merged, anyone JIT-building against a wheel needs that pin — worth a line in the README, since the failure only shows up when the module is first called.

zufayu:
Thanks for the investigation and proposed fix. Since #4847 addresses the immediate OPUS FMHA failure and #5312 provides an ABI strategy aligned with AITER’s de-Torch direction, we do not plan to merge #4996. Could you please close it as superseded?

We will validate the ROCm 10 backport with Kimi-K2.6 TP8 and track the remaining ABI work separately.

zufayu:
## ROCm/aiter PR #4996 — [CI] [JIT] [Build] [Bugfix] Build pybind modules against torch's bundled pybind11; drop pip pybind11 dep

**Repoints aiter's JIT pybind-module builds from the pip-installed pybind11 headers to torch's bundled copy (torch/include, re-added as -isystem for torch_exclude builds like module_aiter_core) so prebuilt-wheel and runtime-JIT modules share one pybind11 internals version, and drops the pip pybind11 dependency from requirements/pyproject/setup and the CI workflows.**

Validation (deterministic): NOT RUN — the PR changes runtime code (aiter/jit/utils/cpp_extension.py, executed at JIT-build time in the serving path) but ships no test target; the changed path runs at run time, so the missing target is reported as a finding below
Perf (advisory): NOT RUN — no perf target: the PR changes runtime code but ships no test target (build-flag change only, no steady-state kernel surface)

⚠️ [verified] The patch no longer applies to main and the maintainer has stated the PR will not be merged: main restructured the exact is_python_module block in aiter/jit/utils/cpp_extension.py that this PR edits (ABI build flags now split by torch_exclude, torch_exclude builds hardcode `-D_GLIBCXX_USE_CXX11_ABI=1` where this PR reads `torch._C` unconditionally, and `-DAITER_EXTENSION_NAME` was added), so a rebase is semantic, not textual; the maintainer comment of 2026-09-10 asks to close as superseded by #4847 + #5312. Any CI result on this branch describes a tree that has moved. **Author must** rebase onto main's restructured block (resolving the `_GLIBCXX_USE_CXX11_ABI` divergence) or close the PR as superseded.
⚠️ [inferred] The fix moves the ABI coupling from pip pybind11 to torch's bundled copy: a JIT module now bakes whatever internals version the runtime torch ships (verified mechanism: internals-version-keyed type registries; torch 2.9.1 bundles pybind11 3.0.1 = v11), so a prebuilt wheel reused with a torch whose bundled pybind11 internals differ (e.g. a torch shipping pybind11 3.1.0 / internals v12) reopens the same `TypeError: incompatible function arguments` mismatch between module_aiter_core and JIT modules — the author concedes the torch-version tie in PR comments, and the field report shows wheels do run against other torch versions (aiter 0.1.19 wheel on torch 2.12.0). **Reviewer should ask** whether #5312's ABI strategy (the de-Torch-aligned approach the maintainer prefers) closes this window before any part of this approach is revived.
📝 [verified] The dependency removal is incomplete: the PR drops the `pip install pybind11` lines from the six workflows it touches (.github/workflows/aiter-test.yaml, atom-test.yaml, kimi-downstream.yaml, kimi-perf-downstream.yaml, vllm_benchmark.yaml) but a seventh, .github/workflows/vllm-disagg-ci-smoke-workflow.yaml:119, still runs `pip install --upgrade "pybind11>=3.0.1"` — dead weight after the change and inconsistent with the PR's "no longer needs a pip pybind11 dependency at all" claim. **Author must** drop that line too if the approach is revived.
📝 [verified] csrc/cpp_itfs/pa_gluon_aot/Makefile:3 still runs `python3 -m pybind11 --includes` into PYBIND11_INC; with the pip package gone from requirements.txt this shells out to a missing module (stderr noise, empty PYBIND11_INC), and the dev build survives only because `-I$(TORCH_PATH)/include` is already on the hipcc line — which is accidentally the torch-bundled resolution this PR wants. **Author must** drop the PYBIND11_INC call from the Makefile.
📝 [verified] No regression test asserts the invariant the PR exists to fix: nothing in op_tests/ checks that a JIT-built python module and the prebuilt module_aiter_core bake the same `__pybind11_internals_v*` symbol, so the #4770 class of mismatch can silently return (a drifting torch or builder change would pass every existing test until a user hits the call-time TypeError). **Author must** add a test asserting the internals symbol of a JIT-built python module matches module_aiter_core's.

Rohan138:
#5312 LGTM, thanks for the pointer! Closing

PR #4999, opened 2026-08-25
[Triton/Gluon] Fix AMDGCN codegen abort in fp8_mqa_logits past ~32K context

kzjeef:
## Summary

On gfx950, `fp8_mqa_logits` aborts inside the AMDGCN backend at JIT-compile time
once the context passes roughly 32K tokens. It is a `Fatal Python error: Aborted`,
so the process dies rather than raising — an inference server hosting a DSA/NSA
model (GLM-5.2, DeepSeek-V3.2-style indexers) goes down with no recoverable error.

```
python3: /__w/triton/triton/llvm-project/llvm/include/llvm/ADT/Sequence.h:275:
llvm::iota_range<unsigned int>::iota_range(T, T, bool) [T = unsigned int]:
Assertion `Begin <= End && "Begin must be less or equal to End."' failed.

Current thread (most recent call first):
  File "triton/backends/amd/compiler.py", line 544 in make_amdgcn
  File "triton/compiler/compiler.py", line 327 in compile
  File "aiter/ops/triton/attention/fp8_mqa_logits.py", line 228 in fp8_mqa_logits
```

## Root cause

Two independent constexprs in the Gluon launch path combine into a shape the
AMDGCN backend cannot codegen:

- `block_m = 2 if (num_heads <= 32 and seq_len > 4096) else 1` — 2 for any
  prefill chunk over 4096 tokens on a ≤32-head indexer.
- `use_buffer_store = logits.numel() * logits.element_size() < 2 GiB` — False
  once the fp32 `[seq_len, seq_len_kv]` logits buffer clears the 32-bit byte
  offset of a buffer resource descriptor.

`BLOCK_M == 2` **and** `USE_BUFFER_STORE == False` together abort the compile.
Either one alone is fine.

The pair had never been compiled before because nothing reaches it until the
logits buffer passes 2 GiB, i.e. `seq_len * seq_len_kv >= 2^29`. With a
16384-token prefill chunk that is a KV length of 32768 — every context past
~32K tokens, and nothing below it.

Measured on MI355X / gfx950, ROCm 7.2.4, torch 2.11.0+rocm7.2, Triton
`3.7.0+amd.rocm7.2.0.git89002410`, `num_heads=32`, `head_dim=128`:

| seq_len | seq_len_kv | logits | BLOCK_M | USE_BUFFER_STORE | result |
|--------:|-----------:|-------:|:-------:|:----------------:|:-------|
| 16384 | 16384 | 1.00 GiB | 2 | True (auto) | ok |
| 16384 | 32768 | 2.00 GiB | 2 | False (auto) | **abort** |
| 16384 | 2048 | 0.12 GiB | 2 | False (forced) | **abort** |
| 4096 | 131072 | 2.00 GiB | 1 | False (auto) | ok |
| 8192 | 65536 | 2.00 GiB | 2 | False (auto) | **abort** |
| 16384 | 32768 | 2.00 GiB | 2 | True (forced) | ok |

Rows 3 and 4 isolate it: the abort tracks the constexpr pair, not the tensor size,
and neither flag triggers it alone.

## Fix

Hoist the buffer-limit computation above the arch dispatch and gate `block_m` on
it, so the non-buffer store path always walks one query row per program:

```python
block_m = 2 if (num_heads <= 32 and seq_len > 4096 and use_buffer_store) else 1
```

`BLOCK_M=1` is the same kernel without the row pairing, and it is already what
every shorter context uses. This is a workaround at the aiter level — the
underlying codegen bug in the `BLOCK_M=2` + `gl.store` combination is still there
and worth fixing separately in the compiler.

**Scoped to one Triton build.** Everything here was measured on
`3.7.0+amd.rocm7.2.0.git89002410`; no other version was tried, so I cannot say
whether an earlier or later backend codegens the pair. The fix does not depend on
the answer — it only removes a configuration that no shape reached before ~32K
context — but the abort itself may not reproduce elsewhere.

## Correctness

Same inputs through the buffer-store and non-buffer-store paths:

```
buffer-store vs non-buffer-store: max|diff| = 0.000e+00     (bit-identical)
vs torch reference:  buffer = 1.479e-04   non-buffer = 1.479e-04
```

## Performance

No shape that worked before changes behaviour: `use_buffer_store` is still True
there, so `block_m` is still 2 and the emitted kernel is unchanged. The fix only
takes effect where `BLOCK_M=2` could not be compiled at all.

Kernel time, median of 20, MI355X:

| shape | BLOCK_M=2 | BLOCK_M=1 | selected after fix |
|:------|----------:|----------:|:------------------:|
| 8192 × 8192 | 0.456 ms | 0.458 ms | 2 (unchanged) |
| 16384 × 16384 | 1.491 ms | 1.617 ms | 2 (unchanged) |
| 16384 × 32768 | **abort** | 3.27 ms | 1 (fix applies) |

For reference, giving up the row pairing costs 8.4% at 16384 × 16384 — but the
fix never applies at a shape where the pairing was available.

End to end, SGLang v0.5.17 + GLM-5.2-MXFP4, TP4 on MI355X, 45K-token prompt,
prefix cache disabled, median of 4:

| configuration | TTFT | LLVM aborts |
|:--------------|-----:|------------:|
| stock, `--chunked-prefill-size 16384` (default) | **crash**, HTTP 500 | 4 |
| stock, `--chunked-prefill-size 4096` (workaround) | 4.31 s | 0 |
| **this fix**, `--chunked-prefill-size 16384` | **3.91 s** | 0 |

Shrinking the prefill chunk to 4096 also avoids the abort (it forces
`BLOCK_M=1`), but costs 9.3% TTFT because every GEMM and MoE stage loses 4× of
its M dimension. The fix keeps the default chunk size.

## Test

`(8192, 65536)` added to the existing `test_fp8_mqa_logits` parametrization in
`op_tests/triton_tests/attention/test_fp8_mqa_logits.py` — exactly 2 GiB of
logits, the smallest shape that reaches the combination. The list previously
topped out at 1024 queries and far under 2 GiB, which is why this shipped.

Two things the shape needs that the rest of the list does not, both guarded in
the body rather than in a separate test:

- It runs for one configuration (`num_heads=32, head_dim=128`) instead of all
  sixteen. `BLOCK_M=2` needs `num_heads <= 32`, so half the cross product cannot
  reach the bug at all, and each case costs 2 GiB plus a reference.
- The reference is checked on 32-row windows at each end of the query range
  rather than the whole tensor — `ref_fp8_mqa_logits` materializes
  `num_heads × s_q × s_k` floats, which is 64 GiB here. Rows are independent, so
  the two windows cover both a paired block and the trailing one. Smaller shapes
  still compare in full; the window list is just `[slice(0, s_q)]` for them.

It also skips below 8 GiB free VRAM.

- Before the fix: aborts the pytest process (fatal, not a test failure).
- After: `145 passed, 15 skipped in 16.83s` — one new case, the 144 pre-existing
  ones unchanged, and the 15 skips are the guarded-out configurations.

### DIFF
--- a/aiter/ops/triton/attention/fp8_mqa_logits.py
+++ b/aiter/ops/triton/attention/fp8_mqa_logits.py
@@ -194,6 +194,12 @@ def fp8_mqa_logits(
             matrix_instr_nonkdim=matrix_instr_nonkdim,
         )
     else:
+        # Buffer ops use a 32-bit byte offset (2 GiB resource descriptor cap).
+        # Fall back to plain global load/store when a tensor exceeds that.
+        BUFFER_LIMIT_BYTES = 2 * 1024 * 1024 * 1024
+        use_buffer_load = KV.numel() * KV.element_size() < BUFFER_LIMIT_BYTES
+        use_buffer_store = logits.numel() * logits.element_size() < BUFFER_LIMIT_BYTES
+
         num_buffers = 2
         USE_FOLDED_REDUCTION = FOLDED_REDUCTED_SUPPORT and num_heads > 16
         if arch == "gfx950":
@@ -203,7 +209,12 @@ def fp8_mqa_logits(
             num_chains = 4 if USE_FOLDED_REDUCTION else 0
             num_warps = 2 if num_heads <= 32 else 1
             block_kv = 64 if num_heads <= 32 else 32
-            block_m = 2 if (num_heads <= 32 and seq_len > 4096) else 1
+            # BLOCK_M=2 with the non-buffer store path aborts the AMDGCN backend
+            # at JIT time (Sequence.h:275 "Begin must be less or equal to End").
+            # Only reachable once logits pass 2 GiB, i.e. context past ~32K.
+            block_m = (
+                2 if (num_heads <= 32 and seq_len > 4096 and use_buffer_store) else 1
+            )
             mfma_nonk_dim = 32 if (head_size <= 64 or num_heads == 32) else 16
             other = {
                 "USE_PADDED_SHARED_LAYOUT": ASYNC_COPY_SUPPORTS_DISTRIBUTED,
@@ -220,11 +231,6 @@ def fp8_mqa_logits(
             block_m = 1
             other = {"LOOP_VARIANT": loop_variant}
 
-        # Buffer ops use a 32-bit byte offset (2 GiB resource descriptor cap).
-        # Fall back to plain global load/store when a tensor exceeds that.
-        BUFFER_LIMIT_BYTES = 2 * 1024 * 1024 * 1024
-        use_buffer_load = KV.numel() * KV.element_size() < BUFFER_LIMIT_BYTES
-        use_buffer_store = logits.numel() * logits.element_size() < BUFFER_LIMIT_BYTES
         _gluon_fp8_mqa_logits_kernel[((seq_len + block_m - 1) // block_m,)](
             Q_ptr=Q,
             KV_ptr=KV,
--- a/op_tests/triton_tests/attention/test_fp8_mqa_logits.py
+++ b/op_tests/triton_tests/attention/test_fp8_mqa_logits.py
@@ -95,6 +95,10 @@ def generate_cp_test_data(seq_len, seq_len_kv):
         (128, 1024),
         (1024, 1024),
         (1024, 1560),
+        # Exactly 2 GiB of logits, which is where the gfx950 kernel pairs
+        # BLOCK_M=2 with a non-buffer store -- the combination that used to abort
+        # the AMDGCN backend at JIT time. Nothing above it comes close.
+        (8192, 65536),
     ],
 )
 @pytest.mark.parametrize("num_heads", [32, 64])
@@ -110,6 +114,16 @@ def test_fp8_mqa_logits(
     disable_cp: bool,
     clean_logits: bool,
 ) -> None:
+    # The 2 GiB shape is here for one configuration only: BLOCK_M=2 needs
+    # num_heads <= 32, and the shape costs 2 GiB of logits plus a reference,
+    # which is not worth paying sixteen times over.
+    big = s_q * s_k * 4 >= 2 * 1024**3
+    if big:
+        if (num_heads, head_dim, disable_cp, clean_logits) != (32, 128, True, True):
+            pytest.skip("large shape is checked once")
+        if torch.cuda.mem_get_info()[0] < 8 * 1024**3:
+            pytest.skip("needs ~8 GiB free")
+
     torch.manual_seed(0)
     q = torch.randn(s_q, num_heads, head_dim, device="cuda", dtype=torch.bfloat16)
     kv = torch.randn(s_k, head_dim, device="cuda", dtype=torch.bfloat16)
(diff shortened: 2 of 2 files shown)


aiter/ops/triton/attention/fp8_mqa_logits.py
```diff
@@ -203,7 +209,17 @@ def fp8_mqa_logits(
             num_chains = 4 if USE_FOLDED_REDUCTION else 0
             num_warps = 2 if num_heads <= 32 else 1
             block_kv = 64 if num_heads <= 32 else 32
-            block_m = 2 if (num_heads <= 32 and seq_len > 4096) else 1
+            # BLOCK_M=2 paired with the non-buffer store path makes the AMDGCN
```
cagrikymk: Could you trim this comment a bit? I think this indicates a compiler bug but I will take a closer look later. It doesnt have to block this PR.

Besides that, LGTM!

cagrikymk (dismissed):

op_tests/triton_tests/attention/test_fp8_mqa_logits.py
```diff
@@ -149,3 +149,47 @@ def test_fp8_mqa_logits(
...
         return  # nothing left to compare
     assert diff < 1e-3, f"{diff=}"
+
+
+# Both shapes land on exactly 2 GiB of logits, which is where the gfx950 kernel
+# pairs BLOCK_M=2 with a non-buffer store and used to abort the AMDGCN backend at
+# JIT time. Nothing above reaches it: every shape there is <= 1024 queries or far
+# under 2 GiB.
+@pytest.mark.parametrize("s_q, s_k", [(8192, 65536), (16384, 32768)])
+@torch.inference_mode()
```
cagrikymk: Could you extend the existing test cases with a short note about the issue?

We can just add (8192, 65536) there (or something smaller but large enough to reproduce the bug)

cagrikymk: We dont have to have a dedicated tests I think, extending the existing ones should do the job.

kzjeef: Done in 42df52b8b — dedicated test removed, `(8192, 65536)` added to the existing parametrization with a note on why that shape.

Two things it needed that the rest of the list does not, both guarded in the body:

- It runs for one configuration (`num_heads=32, head_dim=128, disable_cp, clean_logits`) rather than all sixteen. `BLOCK_M=2` requires `num_heads <= 32`, so half the cross product cannot reach the bug at all, and every case costs 2 GiB of logits.
- The reference is checked on 32-row windows at each end of the query range instead of the whole tensor. `ref_fp8_mqa_logits` materializes `num_heads x s_q x s_k` floats — 64 GiB at this shape, and the `relu`/`mul` intermediates multiply that. Rows are independent, so the two windows cover both a paired `BLOCK_M=2` block and the trailing odd row. Every smaller shape still compares in full; the window list is just `[slice(0, s_q)]` for them.

The shape cannot be made smaller — the trigger *is* the 2 GiB logits buffer (`use_buffer_store` flips at `s_q * s_k >= 2^29`), and `BLOCK_M=2` needs `s_q > 4096`, so `(8192, 65536)` is the minimum. It also skips below 8 GiB free VRAM.

Net -21 lines. Verified on MI355X: `145 passed, 15 skipped in 16.83s` with the fix, and the one new case still aborts the pytest process without it.

aiter/ops/triton/attention/fp8_mqa_logits.py
kzjeef: Trimmed in 5df5465d7 — down to three lines naming the assertion and the reachability condition.

Agreed it is a compiler bug; this is only a workaround at the aiter level and the `BLOCK_M=2` + `gl.store` codegen path is still broken underneath. Happy to hand over the repro if useful — `s_q=16384, s_k=2048` with `USE_BUFFER_STORE` forced False reproduces it at 0.12 GiB, so it does not need a large allocation to debug. One caveat worth recording: everything was measured on Triton `3.7.0+amd.rocm7.2.0.git89002410` only, so I cannot say whether another backend version codegens the pair.

cagrikymk (approved):

cagrikymk (approved):

azaidy (approved):
LGTM!

PR #5002, opened 2026-08-26
[ROCm] Add fused Q/K norm, RoPE, gate, and FP8 quantization

nholmber:
## Purpose

Add a fused gfx950 operator that combines Q/K RMSNorm, RoPE, output-gate extraction, and grouped FP8 quantization while preserving BF16 model-visible outputs. Forward precomputed Q/K/V descales through varlen FMHA so framework integrations can invoke FP8 attention without requantizing Q/K/V.

## Changes

- Add the generic `fused_qk_norm_rope_gate_fp8_quant` public wrapper and Triton kernels.
- Support BF16 source activations and E4M3-FN FP8 Q/K/V outputs on gfx950.
- Return stable per-sequence/per-KV-head descale buffers.
- Validate input/output layouts, shapes, dtypes, and devices.
- Bound suffix reductions and quantization to the requested token and sequence ranges.
- Forward complete Q/K/V descale sets through varlen FMHA to CK; reject partial descales and CK-disabled use.
- Add correctness, suffix-boundary, padded-allocation, validation, and FMHA ABI tests.

## Dependency

The end-to-end framework integration is in:

```text
vllm-project/vllm#53823
```

## Validation

Validated after rebasing onto current `origin/main` on MI355X/gfx950:

```text
focused GPU suite: 18 passed
ruff check: passed
python bytecode compilation: passed
git diff --check: passed
```

Earlier end-to-end validation with the paired vLLM branch:

```text
long-context and packed coherence: passed
GSM8K: 97.57% flexible-extract
GPQA-Diamond: 92.42%
FP8 ASM FMHA dispatch: confirmed
```

## AI assistance disclosure

This change was developed with OpenAI Codex assistance. The human submitter must review every changed line and confirm they understand and can defend the implementation before marking the PR ready.


aiter/ops/triton/_triton_kernels/rope/qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,613 @@
...
+    BLOCK_D: tl.constexpr,
+    SEARCH_STEPS: tl.constexpr,
+):
+    token = tl.program_id(0)
+    head = tl.program_id(1)
+    valid_token = token < num_actual_tokens
+
+    low = 0
+    high = num_sequences
+    for _ in tl.static_range(0, SEARCH_STEPS):
```
damien-lejeune, line 436: The 3 binary search implementations, respectively in `quantize_qk_grouped_kernel`, `quantize_qk_grouped_offset_kernel` and `quantize_v_grouped_offset_kernel` could be extracted to a `@triton.jit` device helper function. The last two are seemingly the same, but for some reason here the implementation is slightly different. But here, wouldn't setting `high = num_sequences - 1` remove the need to have the `mask` and `other` params in the `tl.load`, along with the `tl.minimum`? If so, all would fit nicely to extract this feature in a common function to ease maintenance.

aiter/ops/triton/rope/qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,399 @@
...
+            gqa_ratio=gqa_ratio,
+            head_dim=head_dim,
+            FP8_MAX_VALUE=FP8_MAX,
+            BLOCK_D=triton.next_power_of_2(head_dim),
+            SEARCH_STEPS=8,
+            num_warps=4,
+        )
+        search_steps = max(0, (num_quant_sequences - 1).bit_length())
+    else:
+        search_steps = max(0, (num_quant_sequences - 1).bit_length())
```
damien-lejeune, line 377: We could have the `search_steps = max(0, (num_quant_sequences - 1).bit_length())` outside of the if/else statements to have only one instance.

aiter/ops/triton/rope/qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,399 @@
...
+            query_fp8.stride(1),
+            key_fp8.stride(0),
+            key_fp8.stride(1),
+            num_q_heads=num_query_heads,
+            num_kv_heads=num_kv_heads,
+            gqa_ratio=gqa_ratio,
+            head_dim=head_dim,
+            FP8_MAX_VALUE=FP8_MAX,
+            BLOCK_D=triton.next_power_of_2(head_dim),
+            SEARCH_STEPS=8,
```
damien-lejeune, line 372: Is there a reason why `SEARCH_STEPS=8` here, instead of being parameterized like the other variants? I assume it's ok, runtimewise, given `num_sequences <= 256`. But if `ceil(log_2(num_sequences)) < 8` it will still run a binary search with 8 iterations (and emit extra, wasteful, loads).

op_tests/triton_tests/rope/test_qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,366 @@
...
+
+import aiter
+from aiter import flash_attn_varlen_func
+from aiter.ops.triton.rope.qwen3_next_fp8_qkv import (
+    FP8_MAX,
+    qwen3_next_fp8_qkv_prep,
+)
+
+NUM_QUERY_HEADS = 8
+NUM_KV_HEADS = 1
```
damien-lejeune, line 15: Would it be possible to test it with `NUM_KV_HEADS > 1` too? Otherwise with `NUM_KV_HEADS = 1` only we are not testing part of the kernel implementation (e.g. indexing are using `num_kv_heads = 0`; setting it up to something else than 0 might makes increase coverage of indexing issue, among other things)

op_tests/triton_tests/rope/test_qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+def _make_inputs(lengths: list[int]):
+    device = torch.device("cuda")
+    total_tokens = sum(lengths)
+    torch.manual_seed(1234 + total_tokens)
+    q_gate = torch.randn(
+        total_tokens,
+        NUM_QUERY_HEADS * 2 * HEAD_DIM,
+        dtype=torch.bfloat16,
```
damien-lejeune, line 118: The kernel seems to be support `float16` too. It would be nice to test this type too?

aiter/ops/triton/rope/qwen3_next_fp8_qkv.py
```diff
@@ -0,0 +1,399 @@
...
+        rotary_dim=rotary_dim,
+        half_rotary=half_rotary,
+        eps=eps,
+        INPUT_DTYPE=(tl.bfloat16 if q_gate.dtype == torch.bfloat16 else tl.float16),
+        HEAD_BLOCK=head_block,
+        ROT_HALF_BLOCK=rotary_half_block,
+        HAS_PASS=rotary_dim < head_dim,
+        TOKENS_PER_PROGRAM=QK_TOKENS_PER_PROGRAM,
+        ADD_GEMMA_OFFSET=True,
+        REUSE_NORMALIZED_ROTARY=True,
```
damien-lejeune, line 293: Am I understanding right that these two are *always* true? If so, then we have a some dead code (e.g. i.e. branches when `REUSE_NORMALIZED_ROTARY=False`). We could either remove this dead code or allow those to be set by the users.

damien-lejeune (commented):
Overall nice structure!

PR #4761, opened 2026-08-14
[Triton/Gluon] [PERF] Optimize Triton unified attention prefill and decode

vorapolsiloai:
## Summary
- skip per-element causal and bounds masking for fully visible prefill tiles in the Triton 2D unified-attention kernel
- restrict the fast path to gfx950, head size 128, FP8 Q/FP8 KV, query length at least 256, non-windowed, non-shuffled prefill
- preserve current-main dispatch and launch configurations, including the gfx950 tuning merged in #4918; this PR no longer overrides decode routing or launch configs
- select the eligible family through the gfx950 JSON configuration introduced by #5088, with exact safety checks kept compile-time in the kernel and the refactored wrapper unchanged

## Integration context

The PR branch is rebased onto AITER `main` at
[`5e4a8119`](https://github.com/ROCm/aiter/commit/5e4a8119cc2a6b578667102b75b4f321852dfd8c),
including [#5088](https://github.com/ROCm/aiter/pull/5088) and
[#4918](https://github.com/ROCm/aiter/pull/4918). The kernel patch is unchanged
from the benchmarked candidate. A 480-case configuration and effective-gate
comparison confirmed that the JSON migration preserves all existing launch
parameters and enables the fast loop only for the benchmarked shape family.

The end-to-end comparison used two derivatives of the same pinned Image C
parent (`sha256:653c62aad83496c96748cd28151810f13757509120f548a389d555527957fbb4`).
The baseline derivative installed the two unified-attention source files from
`35c652ed`; the candidate installed those files from this PR at `c4638c76`.
All other image contents, model files, runtime flags, and workload inputs were
held fixed.

## Performance validation

Tested on MI355X/gfx950 with Qwen3-VL-235B-A22B-Instruct-MXFP4. Servers were
NUMA-pinned to disjoint GPUs, clients were launched synchronously, and results
used same-GPU or same-pair crossover placement. Full-node GPU utilization and
temperatures were sampled throughout the runs.

- TP1, 60k input / 600 output, concurrency 32: **+9.10%** median output throughput; **8.34%** lower duration; **9.25%** lower mean TTFT; **7.91%** lower mean TPOT across four same-GPU comparisons
- TP2, 60k input / 600 output, concurrency 32: **+5.64%** median output throughput; **5.33%** lower duration; **5.98%** lower mean TTFT; **5.17%** lower mean TPOT across two same-pair comparisons
[...]

PR #4614, opened 2026-08-06
[Triton/Gluon] [GFX950] Add Unified Attention Gluon Kernel

cagrikymk:
## Motivation
This PR adds gfx950 gluon kernel for unified attention. It uses single kernel that supports both 2d and 3d grid.

Performance is tuned for head sizes [64, 128, 256], dtype fp8 x fp8 and bf16 x bf16.

## Remaining Steps

- [x] Test e2e performance, at least for GPT-OSS-120b
- [x] Extend the gluon routing to make it more bullet-proof

## Technical Details
Performance numbers are collected using Triton 3.7.1.

### Geomean speedup by head_size against Triton

This is the geomean speedup of various run configs with varrying concurrency, kv/seq. length, number of KV/Q heads,

| head_size | decode bf16 | decode fp8 | prefill bf16 | prefill fp8 | mixed bf16 | mixed fp8 |
|--:|--:|--:|--:|--:|--:|--:|
| 256 | **1.58x** | **1.02x** | **2.17x** | **2.22x** | **1.90x** | **1.81x** |
| 128 | **1.10x** | **1.08x** | **1.19x** | **1.38x** | **1.23x** | **1.41x** |
| 64 | **1.11x** | **1.24x** | **1.51x** | **1.55x** | **1.54x** | **1.62x** |

Detailed performance report: 
[gluon_gfx950_perf_report.md](https://github.com/user-attachments/files/30803810/gluon_gfx950_perf_report.md)

## Test Plan

Existing tests are routed to the gluon kernel if conditions are met

## Test Result

Tests pass

PR #5004, opened 2026-08-26
[Triton/Gluon] [Perf] Optimize live-window unified-attention decode

andyluo7:
## Summary

- Partition the 3-D unified-attention decode kernel over the live sliding window instead of the full logical sequence.
- Apply the same live-segment bound in the Triton and gfx1250 Gluon reducers.
- Select 3-D from decode state, active KV length, and available launch parallelism, independent of GPU architecture and head size.
- Cover a large, off-tile sliding window in the existing unified-attention regression suite; no standalone dispatcher-coupled test file is added.
- On gfx1250, use the existing Gluon 3-D producer for the affected unshuffled FP8 live-window path. This backend choice is separate from the architecture-agnostic 2-D/3-D workload selection.

## Motivation

For sliding-window decode, the previous 3-D kernel derived segment boundaries from the full sequence length. It could therefore visit expired KV tiles before masking them and leave leading reduction segments empty when the logical context was much longer than the active window.

This change rebases 3-D segment ranges onto the first tile intersecting the live window and makes each reducer consume only active live-window segments. It changes neither the attention mask nor the stored KV layout.

## Dispatch and platform scope

The 2-D/3-D selector now chooses 3-D only when all of the following are true:

- the batch is decode-only;
- the active KV length exceeds the existing 2-D short-context boundary;
- the 2-D launch does not already exceed the target launch parallelism.

Short windows, prefills, and already-parallel launches retain the previous 2-D route. There is no model name, product name, GPU-architecture gate, or head-size gate in this workload decision.

The existing architecture-specific kernel implementations and low-level launch tuning remain architecture-specific. In particular, gfx1250 already uses its Gluon 3-D producer for shuffled KV. The wrapper now also uses that available producer for unshuffled FP8 live-window 3-D decode, after gfx1250 testing showed non-finite partials in the generic Triton FP8 producer. No ASIC-revision or model-specific condition is introduced.

## Correctness and regression coverage

Current head: `1138500476bcdcc6f1f695e1f36f98b383149ca4`  
Tested base: `d58537b705269f3179c3ba62fa34c108360a0aa2`

Upstream `main` subsequently advanced to `ce788f58feb47c66c23bfde5d39fab8e5db76606` through #5123, which changes only DCP/top-k files. The current PR head merge-simulates cleanly with that commit and has no file overlap with it.

Static checks on all five changed Python files:

```text
black --check                         PASS
ruff 0.16.0 check                     PASS
python -m py_compile                  PASS
git diff --check                      PASS
```

The gfx1250 runs used one physical `gfx1250` A0 GPU on runner
`heliosr-1b114-a06-4-mi45x`, image
`rocm/ufb-private@sha256:7472a35db180b9e0dd3850be862fea30fb7f5ff7663c7b3740dea01b847bae3a`,
and Triton `3.8.0+gite411a7fd.gfx1250a0v5`. The compiler wheel SHA-256 is
`d6f0db98f683b0f289b150ce048aab380203101cb3e233477f60ed65d21f1668`
([build run](https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33551006583)).
The validation branch is `58baf857564c6ddd24d24d923f8098a4dda71ffb`; all five PR-file blob IDs
match this public head exactly.

Focused gfx1250 A0 producer/reducer diagnostic:

- Run: https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33569822567
- `2 passed`; preflight and postflight GPU-health checks passed.
- `segm_output`: 524,288/524,288 finite values.
- `segm_max`: 8,192/8,192 finite values.
- `segm_expsum`: 8,192/8,192 finite values.
- Torch, Triton, synchronized Gluon, and normal-wrapper reductions all had a 0.0 mismatch ratio.
- Maximum finite difference was `0.00439453125` for BF16 output and `0.015625` for FP8 output.
- The test fixes scale tensors at 1.0 so an all-zero result cannot pass only because of loose FP8 tolerances.

Complete gfx1250 3-D selection:

- Run: https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33569985137
- `100 passed, 20 skipped`; the skips are the unsupported unshuffled-NVFP4 cases.
- All 20 supported large off-tile sliding-window configurations passed; four unsupported NVFP4 configurations skipped.
- Preflight/postflight health checks passed and the kernel-fault log was empty.

Complete gfx1250 unified-attention file, compared with the same 2,168-case matrix on the pre-fix candidate:

- Current run: https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33572592033 -- `1172 passed, 660 skipped, 336 failed`.
- Pre-fix candidate: https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33564202060 -- `1170 passed, 660 skipped, 338 failed`.
- The selected-node lists are byte-identical (`sha256:634677cf32cfef001e2329a3ea7a6fbe93f1f9a2fc564466e4a7d256cf15f2d9`).
- The current failure set is a strict subset of the pre-fix set: zero new failures and exactly the two unshuffled-FP8 large-window failures removed.
- All 2,168 selected cases reached terminal outcomes. Preflight/postflight health checks passed, the kernel-fault log was empty, and the downloaded artifact hashes match the host-preserved evidence.
- The workflow is red because 336 known failures from the pre-fix candidate remain, not because this change introduced a new failure.

Upstream exact-head CI:

- [Triton Test](https://github.com/ROCm/aiter/actions/runs/33573357062): all eight MI35X shards, all eight MI300X shards, and both aggregate result jobs passed.
- On both MI35X and MI300X, shard 6 executed all 2,168 cases in `test_unified_attention.py`: `1104 passed, 1064 skipped`. All 16 supported large-window configurations passed; the eight skips are unsupported NVFP4 combinations.
- [OPUS Test](https://github.com/ROCm/aiter/actions/runs/33570073257): MI35X and MI300X passed.
- [Checks](https://github.com/ROCm/aiter/actions/runs/33570073056): Black, Ruff, and repository dependency checks passed; the title/label check also passed.
- [Aiter Test](https://github.com/ROCm/aiter/actions/runs/33570073242): both wheel builds, all eight MI35X and eight MI300X standard-test shards, result aggregation, tuned-op bench, and the final AITER gate passed.

The repository's `ci:gfx1250-ffm-triton` workflow accepts only PR heads hosted in `ROCm/aiter`, so it skips this fork-authored PR by policy. The physical gfx1250 exact-head runs above cover that gap.

## Performance

For the listed shapes, the current selector still chooses the same gfx950 3-D path and the measured gfx950 kernel implementation is unchanged; later edits broadened eligibility and fixed the gfx1250 producer route. MI355X runs used 10 warmups and 50 timed iterations; values below are median speedups of live-window 3-D over the existing 2-D route.

| Batch | Logical context / live window | Speedup |
| ---: | ---: | ---: |
| 1 | 2K / 2K | 1.34x |
| 32 | 2K / 2K | 1.38x |
| 1 | 8,203 / 8,203 | 3.93x |
| 32 | 8,203 / 8,203 | 3.98x |
| 1 | 262K / 32K | 14.30x |
| 32 | 262K / 32K | 9.10x |

These are MI355X kernel measurements, not MI300X or gfx1250 performance claims.

## Related work and merge behavior

- #4761 overlaps the two unified-attention source files but targets a separate non-windowed 2-D prefill/decode optimization. Its current head `0a20fb8f9544b7156f27b4fc61a20f41bb4837d4` merge-simulates cleanly.
- #4614 overlaps the wrapper and test file. Its current head `e20141dcc2d87c29c798d0775fb183a4db8bdec6` merge-simulates cleanly, but its eligible gfx950 Gluon dispatch runs before this Triton selector, so landing order and performance ownership should be coordinated rather than treating the effects as cumulative.
- #4676 overlaps only the wrapper and its current head `e59fa7b41e83803c407db274a606023d380462f2` merge-simulates cleanly. Its FlyDSL path requires a full window, so sliding-window calls fall through to this path.
- #3959 changes only `aiter/ops/triton/attention/pa_decode_shuffle.py`; it has no file overlap with this PR.
- #5088 is a broad wrapper/config refactor and does not implement this live-window partition/reduction fix. Its current head `67abf386c25d4641bc0f0286caded832aa8d648a` conflicts mechanically in `aiter/ops/triton/attention/unified_attention.py`; if it lands first, this PR must be rebased into the new backend/config structure.
- #5157's separate 2-D prefill work merge-simulates cleanly. #5158 is also semantically separate 2-D prefill work but currently conflicts in the shared wrapper/kernel files, so its landing order must likewise be coordinated.

## Review feedback addressed

- Removed architecture and head-size conditions from the 2-D/3-D workload decision.
- Folded the large-window coverage into `test_unified_attention.py`.
- Removed the standalone dispatcher test and test file.
- Kept backend-specific routing separate from the workload selector.

## AI assistance

OpenAI Codex assisted with rebase verification, patch cleanup, duplicate-work analysis, test construction, hardware diagnostics, and PR drafting.

### DIFF
--- a/aiter/ops/triton/_gluon_kernels/gfx1250/attention/unified_attention_3d.py
+++ b/aiter/ops/triton/_gluon_kernels/gfx1250/attention/unified_attention_3d.py
@@ -722,7 +722,7 @@ class AttentionProgram:
         query_offset_1_pv,
         query_mask_0_pv,
         query_mask_1_pv,
-        segm_idx,
+        segment_tile_start,
         tiles_per_segment,
         stride_k_cache_0,
         stride_k_cache_1,
@@ -903,8 +903,8 @@ class AttentionProgram:
 
         # Calculate tile range
         num_tiles = (max_seq_prefix_len + cfg.TILE_SIZE - 1) // cfg.TILE_SIZE
-        tile_start = segm_idx * tiles_per_segment
-        tile_end = min((segm_idx + 1) * tiles_per_segment, num_tiles)
+        tile_start = segment_tile_start
+        tile_end = min(segment_tile_start + tiles_per_segment, num_tiles)
         if cfg.SLIDING_WINDOW > 0:
             qpos_lo = q_block_local_idx * cfg.BLOCK_Q
             qpos_hi = gl.minimum(
@@ -913,8 +913,10 @@ class AttentionProgram:
             )
             first_allowed_key = context_len + qpos_lo - cfg.SLIDING_WINDOW + 1
             last_allowed_key = context_len + qpos_hi
-            tile_start = gl.maximum(0, first_allowed_key // cfg.BLOCK_SIZE)
-            tile_end = gl.minimum((last_allowed_key // cfg.BLOCK_SIZE) + 1, num_tiles)
+            first_allowed_tile = gl.maximum(0, first_allowed_key // cfg.TILE_SIZE)
+            last_allowed_tile = (last_allowed_key // cfg.TILE_SIZE) + 1
+            tile_start = gl.maximum(tile_start, first_allowed_tile)
+            tile_end = gl.minimum(tile_end, gl.minimum(last_allowed_tile, num_tiles))
 
         query_pos_qk = gl.convert_layout(
             query_pos_qk, gl.SliceLayout(1, cfg.QK_WMMA_UNPACKED_LAYOUT)
@@ -1835,17 +1837,30 @@ def cdiv_fn(x, y):
 def get_seq_metadata(
     seq_lens_ptr,
     seq_idx,
+    segm_idx,
     TILE_SIZE: gl.constexpr,
     NUM_SEGMENTS_PER_SEQ: gl.constexpr,
+    SLIDING_WINDOW: gl.constexpr,
+    ALL_DECODE: gl.constexpr,
 ):
     # sequence len for this particular sequence
     seq_len = gl.load(seq_lens_ptr + seq_idx)
 
-    # number of segments for this particular sequence
+    # Decode only needs tiles intersecting the live window. Rebase the segment
+    # grid onto its first tile so different segments do not repeat that work.
+    total_seq_tiles = cdiv_fn(seq_len, TILE_SIZE)
     num_segments = NUM_SEGMENTS_PER_SEQ
-    tiles_per_segment = cdiv_fn(seq_len, num_segments * TILE_SIZE)
+    if ALL_DECODE and SLIDING_WINDOW > 0:
+        first_allowed_key = gl.maximum(0, seq_len - SLIDING_WINDOW)
+        first_tile = first_allowed_key // TILE_SIZE
+        live_tiles = total_seq_tiles - first_tile
+        tiles_per_segment = cdiv_fn(live_tiles, num_segments)
+        segment_tile_start = first_tile + segm_idx * tiles_per_segment
+    else:
+        tiles_per_segment = cdiv_fn(seq_len, num_segments * TILE_SIZE)
+        segment_tile_start = segm_idx * tiles_per_segment
 
-    return seq_len, tiles_per_segment
+    return seq_len, tiles_per_segment, segment_tile_start, total_seq_tiles
 
 
 @gluon.jit
@@ -1910,6 +1925,7 @@ unified_attention_gluon_kernel_3d_repr = make_kernel_repr(
         "BLOCK_SIZE",
         "TILE_SIZE",
         "HEAD_SIZE",
+        "SLIDING_WINDOW",
         "NUM_BLOCKS_GATHER_PER_TILE",
         "NUM_SEGMENTS_PER_SEQ",
         "num_warps",
@@ -2053,14 +2069,17 @@ def _unified_attention_gluon_kernel_3d(
         if q_block_local_idx * cfg.BLOCK_Q >= cur_batch_query_len:
             return
 
-    seq_len, tiles_per_segment = get_seq_metadata(
+    seq_len, tiles_per_segment, segment_tile_start, total_seq_tiles = get_seq_metadata(
         seq_lens_ptr,
         seq_idx,
+        segm_idx,
         cfg.TILE_SIZE,
         cfg.NUM_SEGMENTS_PER_SEQ,
+        cfg.SLIDING_WINDOW,
+        cfg.ALL_DECODE,
     )
 
-    if segm_idx * tiles_per_segment * cfg.TILE_SIZE >= seq_len:
+    if segment_tile_start >= total_seq_tiles:
         return
 
     qk_factor: gl.float32 = cfg.QK_SCALE
@@ -2235,7 +2254,7 @@ def _unified_attention_gluon_kernel_3d(
         query_offset_1_pv,
         query_mask_0_pv,
         query_mask_1_pv,
-        segm_idx,  # for 2D, segm_idx = 0
+        segment_tile_start,
         tiles_per_segment,  # for 2D, tiles_per_segment = num_tiles = (max_seq_prefix_len + cfg.BLOCK_SIZE - 1) // cfg.BLOCK_SIZE
         stride_k_cache_0,
         stride_k_cache_1,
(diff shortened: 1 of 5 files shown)


ChuanLi1101 (dismissed):
LGTM. gfx950 live-window unified-attention decode path looks ready; Aiter Test Gate green on MI35X and MI300X. Pair merge with vLLM #53827 for the EAGLE3 k=4 spec-decode stack.

cagrikymk:
@andyluo7 I assume this targets models with very large sliding windows (for reference, initial target was GPT-OSS with SWA 128).

If that is the case, can we merge the conditions for the 2d/3d selection in an arch/head size agnostic way?

For such large windows, I think head size/GPU arch shouldnt matter and I expect this to benefit regardless as we underfill the GPU. Also, the code will be simpler as well and I dont think we need specific tests for this. If you wanna make sure the large SWA path is hit, you can add to the existing tests.

op_tests/triton_tests/attention/test_unified_attention_live_window.py
```diff
@@ -0,0 +1,183 @@
...
+        (128, 32768, True, 1024, 1024, 1, True),
+        (128, 32768, True, 2048, 1024, 1, False),
+        (128, 1024, True, 8192, 1024, 1, True),
+        (128, 2048, True, 8192, 1024, 1, False),
+        (128, 32768, True, 8192, 1024, 1025, True),
+        (64, 32768, True, 8192, 1024, 1, True),
+        (128, 32768, False, 8192, 1024, 1, True),
+    ],
+)
+def test_sliding_decode_dispatch(
```
cagrikymk: No need for new test file + new tests, adding a large SWA case should be enough to have good coverage.

Also, a test for the use_2d_kernel is not needed IMO. The dispatcher is already a conditional so this would just lead 2 different places to couple whenever the 2d selection part is modified.

andyluo7:
@cagrikymk Thanks -- addressed in `1138500476bcdcc6f1f695e1f36f98b383149ca4`.

- The live-window 2-D/3-D decision is now based only on decode state, active KV length, and launch underfill. The large-window branch has no GPU-architecture or head-size condition.
- Short windows, prefills, and already-parallel launches retain their existing 2-D behavior.
- Backend availability remains separate from workload selection: on gfx1250, the existing Gluon 3-D producer is used for shuffled KV and for the affected unshuffled-FP8 live-window route.
- The standalone dispatcher test/file was removed. Coverage now adds one large off-tile SWA case (`8203` logical tokens / `2053` live tokens) to the existing `test_unified_attention.py` matrix.

Validation on the exact PR files:

- gfx1250 producer/reducer diagnostic: [2 passed](https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33569822567), with finite partials and zero mismatches across Torch, Triton, synchronized Gluon, and wrapper reductions.
- Complete gfx1250 3-D selection: [100 passed, 20 expected NVFP4 skips](https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33569985137); all 20 supported large-window configurations passed.
- Complete 2,168-case gfx1250 file: [1172 passed, 660 skipped, 336 known failures](https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33572592033). Compared with the byte-identical test selection on the [pre-fix candidate](https://github.com/andyluo7/aiter-mi45x-ci/actions/runs/33564202060), there are zero new failures and exactly the two targeted unshuffled-FP8 large-window failures are fixed. Preflight/postflight GPU health passed and the kernel-fault log is empty.
- Upstream [Triton Test](https://github.com/ROCm/aiter/actions/runs/33573357062): all eight MI35X shards, all eight MI300X shards, and both aggregate result jobs passed. On both platforms, the shard containing `test_unified_attention.py` ran all 2,168 cases and all 16 supported large-window configurations passed.
- Upstream [Aiter Test](https://github.com/ROCm/aiter/actions/runs/33570073242): both wheel builds, all 16 MI35X/MI300X standard-test shards, and the final gate passed.

I also checked active related work. The current heads of #4761, #4614, #4676, and #5157 merge cleanly. #5088 is a structural unified-attention refactor rather than a duplicate of this fix, but the current heads conflict mechanically in the shared wrapper; if it lands first, this PR will need a rebase into its new backend/config structure.

op_tests/triton_tests/attention/test_unified_attention_live_window.py
andyluo7: Addressed in `1138500476bcdcc6f1f695e1f36f98b383149ca4`: I removed the standalone test file and the direct `use_2d_kernel` tests, and folded a large off-tile SWA case (`8203` logical tokens / `2053` live tokens) into the existing `test_unified_attention.py` 3-D matrix.

On the exact PR files, all 20 supported large-window gfx1250 configurations passed; the four skips are unsupported unshuffled-NVFP4 combinations. The complete 2,168-case gfx1250 file had zero new failures versus the identical pre-fix selection and fixed the two targeted unshuffled-FP8 large-window failures.

PR #5005, opened 2026-08-26
[Triton/Gluon] [ASM] [HIP] Block-sparse MHAv4 with load-balancing

nsakkine:
Block-sparse MHA v4 with load-balanced tile dispatch

## Summary

Adds a block-sparse forward path to MHA v4: a boolean tile mask becomes a ragged LUT, and a
sparse code object per quantization recipe walks only the selected KV tiles.

gfx950 gets a sparse sibling for every quantized dense recipe at 256×128 tiles — `fp8`, `i8fp8`,
`mxfp8`, `f8f6`, `mxfp6`, `f6f4`, `mxfp4`, `f4f4`; BF16 has an upstream dense gfx950 row but no
sparse row yet, so it stays dense-only. gfx942 gets the two recipes that have dense rows, native
FP8/FP8 (E4M3FNUZ) and signed INT8/FP8, at 256×**64** tiles. GQA works on both: LUT rows stay one
per query head, and K/V addressing uses the GQA ratio.

LUT lengths differ per (batch, head, q-tile), so the flat grid is dispatched from a work table
ordered longest-LUT-first, built on device: one fused kernel ranks and packs up to 8192 entries
with a wave per entry, and larger tables fall back to a sort.

Entry points are `mha_v4(..., block_mask=...)` and `mha_v4_mxfp8(..., block_mask=...)`, or the
ragged LUT triple straight to `mha_v4_packed`. `block_mask` is rejected for
`AttentionFormat.BF16` until a sparse BF16 row exists.

Because the sparse tile is now architecture-dependent, mask geometry and the key-length padding
rule are no longer hardcoded to 128: the new `mha_v4_kv_tile()` export reports the active GPU's
KV tile, the C++ padding check reads `cfg.ts_kv` from the selected manifest row instead of a
literal, and mask constructors including `bench_sage.py` go through the helper.

## Method

Wan 2.x 720p self-attention, 75648 tokens, hd=128, non-causal. `hq=5` is the Ulysses degree-8
shard (1480 work-table entries, fused build); `hq=40` is unsharded (11840 entries, sort
fallback). Each (batch, head, q-tile) row selects exactly `round(d * n_kv)` KV tiles at random
positions, the locality-pessimistic choice. Quantization and LUT construction sit outside the
timed region; the work-table build sits inside it, where it lives at runtime.

`n_kv` is the architecture KV-tile count for this sequence: **591** on gfx950 (256×128) and
**1182** on gfx942 (256×64). Entry counts match on both, since the query tile is 256 either way.
Every figure below is a ratio against that recipe’s own dense kernel, so none depend on the
clocks of the machine they were taken on; ideal speedup is `n_kv / selected_tiles`. gfx950 was
measured on one MI355X across all eight quantized sparse recipes, gfx942 on one MI300X for its
two (`fp8`, `i8fp8`).

## Zero sparsity

An all-true mask makes the sparse kernel visit every tile, so it should reproduce the dense
kernel exactly at the same cost. Output is bitwise identical (`torch.equal`) for all eight gfx950
recipes and for both gfx942 recipes at hq=5 and hq=40. The residual is LUT indirection plus the
work-table build.

### gfx950 (MI355X)

| recipe | dense cost, relative to `fp8` | sparse overhead, hq=5 | sparse overhead, hq=40 |
|---|---|---|---|
| `i8fp8` | 1.27 | +0.6% | +1.1% |
| `mxfp8` | 1.04 | +2.6% | +2.3% |
| `fp8`   | 1.00 | +2.3% | +2.5% |
| `f8f6`  | 1.00 | +2.8% | +2.7% |
| `mxfp6` | 0.88 | +3.7% | +3.6% |
| `f6f4`  | 0.82 | +2.7% | +2.8% |
| `mxfp4` | 0.82 | +2.8% | +2.7% |
| `f4f4`  | 0.75 | +3.6% | +3.2% |

The relative dense costs are taken at hq=5 and agree with hq=40 to within 0.02.

### gfx942 (MI300X)

| recipe | dense cost, relative to `fp8` | sparse overhead, hq=5 | sparse overhead, hq=40 |
|---|---|---|---|
| `i8fp8` | 1.05 | +3.1% | +2.8% |
| `fp8`   | 1.00 | +0.8% | +1.9% |

Dense `i8fp8` sits much closer to dense `fp8` on gfx942 than on gfx950 (1.05× vs 1.27×). The
all-true tax stays in the same few-percent band, though `i8fp8` pays a little more than `fp8`
here — the reverse of gfx950.

## Scaling with density

Timings are back-to-back repeats, so the sparse KV working set stays resident in the 256 MB LLC.

### gfx950 (MI355X)

All eight recipes track the ideal within three points of each other, so the ranges below span the
full set; per-recipe numbers are folded out underneath.

| density | KV tiles/row | ideal | hq=5 measured | of ideal | hq=40 measured | of ideal |
|---|---|---|---|---|---|---|
| 0.9 | 532 | 1.11x | 1.07-1.10x | 96-99% | 1.07-1.10x | 97-99% |
| 0.8 | 473 | 1.25x | 1.20-1.24x | 96-99% | 1.20-1.24x | 96-99% |
| 0.7 | 414 | 1.43x | 1.37-1.42x | 96-99% | 1.37-1.41x | 96-99% |
| 0.6 | 355 | 1.66x | 1.60-1.65x | 96-99% | 1.60-1.65x | 96-99% |
| 0.5 | 296 | 2.00x | 1.91-1.97x | 96-99% | 1.91-1.97x | 96-99% |
| 0.4 | 236 | 2.50x | 2.39-2.46x | 95-98% | 2.39-2.47x | 95-99% |
| 0.3 | 177 | 3.34x | 3.16-3.27x | 95-98% | 3.16-3.28x | 95-98% |
| 0.2 | 118 | 5.01x | 4.69-4.86x | 94-97% | 4.72-4.91x | 94-98% |
| 0.1 |  59 | 10.02x | 8.99-9.48x | 90-95% | 9.12-9.64x | 91-96% |

<details>
<summary>Per-recipe speedup vs its own dense kernel</summary>

hq=5, Ulysses degree 8:

| density | ideal | `fp8` | `i8fp8` | `mxfp8` | `f8f6` | `mxfp6` | `f6f4` | `mxfp4` | `f4f4` |
|---|---|---|---|---|---|---|---|---|---|
| 0.9 | 1.11 | 1.07 | 1.10 | 1.08 | 1.08 | 1.07 | 1.08 | 1.08 | 1.07 |
| 0.8 | 1.25 | 1.21 | 1.24 | 1.22 | 1.21 | 1.20 | 1.21 | 1.21 | 1.20 |
| 0.7 | 1.43 | 1.38 | 1.42 | 1.39 | 1.38 | 1.37 | 1.38 | 1.38 | 1.37 |
| 0.6 | 1.66 | 1.60 | 1.65 | 1.62 | 1.61 | 1.60 | 1.61 | 1.61 | 1.60 |
| 0.5 | 2.00 | 1.92 | 1.97 | 1.94 | 1.93 | 1.91 | 1.92 | 1.93 | 1.92 |
| 0.4 | 2.50 | 2.40 | 2.46 | 2.42 | 2.41 | 2.39 | 2.40 | 2.41 | 2.39 |
| 0.3 | 3.34 | 3.18 | 3.27 | 3.21 | 3.20 | 3.16 | 3.18 | 3.19 | 3.17 |
| 0.2 | 5.01 | 4.71 | 4.86 | 4.77 | 4.75 | 4.69 | 4.72 | 4.73 | 4.70 |
| 0.1 | 10.02 | 9.13 | 9.48 | 9.21 | 9.15 | 8.99 | 9.09 | 9.03 | 9.04 |

hq=40, unsharded:

| density | ideal | `fp8` | `i8fp8` | `mxfp8` | `f8f6` | `mxfp6` | `f6f4` | `mxfp4` | `f4f4` |
|---|---|---|---|---|---|---|---|---|---|
| 0.9 | 1.11 | 1.08 | 1.10 | 1.08 | 1.08 | 1.07 | 1.08 | 1.08 | 1.08 |
| 0.8 | 1.25 | 1.21 | 1.24 | 1.22 | 1.21 | 1.20 | 1.21 | 1.21 | 1.21 |
| 0.7 | 1.43 | 1.39 | 1.41 | 1.39 | 1.38 | 1.37 | 1.38 | 1.38 | 1.38 |
| 0.6 | 1.66 | 1.61 | 1.65 | 1.62 | 1.61 | 1.60 | 1.61 | 1.61 | 1.60 |
| 0.5 | 2.00 | 1.93 | 1.97 | 1.93 | 1.93 | 1.91 | 1.93 | 1.93 | 1.91 |
| 0.4 | 2.50 | 2.41 | 2.47 | 2.42 | 2.41 | 2.39 | 2.41 | 2.41 | 2.39 |
| 0.3 | 3.34 | 3.20 | 3.28 | 3.21 | 3.19 | 3.17 | 3.19 | 3.19 | 3.16 |
| 0.2 | 5.01 | 4.77 | 4.91 | 4.84 | 4.76 | 4.72 | 4.76 | 4.76 | 4.72 |
| 0.1 | 10.02 | 9.33 | 9.64 | 9.35 | 9.31 | 9.14 | 9.25 | 9.12 | 9.20 |

</details>

Below 0.1 the curve keeps climbing but the gap to ideal widens, because a shrinking tile count
per row stops filling the machine:

| density | KV tiles/row | ideal | `fp8` hq=5 | `f4f4` hq=5 | `fp8` hq=40 | `f4f4` hq=40 |
|---|---|---|---|---|---|---|
| 0.05   | 30 | 19.7x | 17.4x (88%) | 16.5x (84%) | 18.2x (92%) | 17.1x (87%) |
| 0.025  | 15 | 39.4x | 29.6x (75%) | 28.5x (72%) | 32.1x (81%) | 29.7x (75%) |
| 0.0125 |  7 | 84.4x | 49.9x (59%) | 46.0x (55%) | 54.1x (64%) | 47.4x (56%) |

### gfx942 (MI300X)

Same protocol, 1182 KV tiles/row. The two recipes stay within a few points of each other, so the
ranges span `fp8` and `i8fp8`.

| density | KV tiles/row | ideal | hq=5 measured | of ideal | hq=40 measured | of ideal |
|---|---|---|---|---|---|---|
| 0.9 | 1064 | 1.11x | 1.08-1.10x | 97-99% | 1.07x | 96-97% |
| 0.8 |  946 | 1.25x | 1.21-1.24x | 97-99% | 1.19-1.20x | 95-96% |
| 0.7 |  827 | 1.43x | 1.38-1.41x | 96-99% | 1.35-1.37x | 94-96% |
| 0.6 |  709 | 1.67x | 1.60-1.64x | 96-98% | 1.56-1.58x | 94-95% |
| 0.5 |  591 | 2.00x | 1.91-1.96x | 96-98% | 1.86-1.88x | 93-94% |
| 0.4 |  473 | 2.50x | 2.38-2.44x | 95-98% | 2.31-2.32x | 92-93% |
| 0.3 |  355 | 3.33x | 3.15-3.22x | 94-97% | 3.04-3.05x | 91-92% |
| 0.2 |  236 | 5.01x | 4.67-4.78x | 93-95% | 4.51-4.56x | 90-91% |
| 0.1 |  118 | 10.02x | 9.01-9.19x | 90-92% | 8.72-8.73x | 87% |

<details>
<summary>Per-recipe speedup vs its own dense kernel (gfx942)</summary>

hq=5, Ulysses degree 8:

| density | ideal | `fp8` | `i8fp8` |
|---|---|---|---|
| 0.9 | 1.11 | 1.10 | 1.08 |
| 0.8 | 1.25 | 1.24 | 1.21 |
| 0.7 | 1.43 | 1.41 | 1.38 |
| 0.6 | 1.67 | 1.64 | 1.60 |
| 0.5 | 2.00 | 1.96 | 1.91 |
| 0.4 | 2.50 | 2.44 | 2.38 |
| 0.3 | 3.33 | 3.22 | 3.15 |
| 0.2 | 5.01 | 4.78 | 4.67 |
| 0.1 | 10.02 | 9.19 | 9.01 |

hq=40, unsharded:

| density | ideal | `fp8` | `i8fp8` |
|---|---|---|---|
| 0.9 | 1.11 | 1.07 | 1.07 |
| 0.8 | 1.25 | 1.20 | 1.19 |
| 0.7 | 1.43 | 1.37 | 1.35 |
| 0.6 | 1.67 | 1.58 | 1.56 |
| 0.5 | 2.00 | 1.88 | 1.86 |
| 0.4 | 2.50 | 2.32 | 2.31 |
| 0.3 | 3.33 | 3.05 | 3.04 |
| 0.2 | 5.01 | 4.51 | 4.56 |
| 0.1 | 10.02 | 8.73 | 8.72 |

</details>

Below 0.1 the same underfill shows up. gfx942 has twice as many (half-width) tiles per row, so
the knee is in the same density range rather than later:

| density | KV tiles/row | ideal | `fp8` hq=5 | `i8fp8` hq=5 | `fp8` hq=40 | `i8fp8` hq=40 |
|---|---|---|---|---|---|---|
| 0.05   | 59 | 20.0x | 17.3x (86%) | 17.0x (85%) | 16.3x (81%) | 16.6x (83%) |
| 0.025  | 30 | 39.4x | 30.9x (78%) | 30.7x (78%) | 29.6x (75%) | 30.2x (77%) |
| 0.0125 | 15 | 78.8x | 53.0x (67%) | 53.0x (67%) | 49.9x (63%) | 51.4x (65%) |

**gfx942 takeaway.** The 256×64 siblings scale like gfx950’s 256×128 ones on the recipes that
exist: Ulysses-8 (`hq=5`) holds 96-99% of ideal from 0.9 down to 0.5 retained tiles and is still
90-92% at 0.1, with `fp8` the closer of the two. Unsharded `hq=40` sits a few points further from
ideal than `hq=5` (sort fallback, more but shorter waves), where gfx950’s extra MX recipes and
128-wide tiles filled the larger MI355X more evenly. None of that changes the dispatch story:
keep the longest-LUT-first work table, and treat gfx942 `fp8` / `i8fp8` as production-ready
sparse rows for Wan Ulysses-8.

## Why the work table

The kernel takes an arbitrary tile selection — one LUT row per (batch, head, q-tile), each with
its own length — and nothing in the interface constrains those lengths to be equal. In raster
order a flat grid finishes with a tail: whichever expensive tiles go last run with most of the
machine idle, so the cost depends both on how uneven the mask is and on where the heavy rows
happened to land. Ordering longest-first removes both dependencies, so the kernel need not assume
anything about how the mask was produced.

The ablation below holds the mask fixed and changes only dispatch order, at hq=5 and 0.1 mean
density. Row lengths are drawn with a fixed mean and increasing dispersion, reported as the ratio
of the longest row to the mean. "Best/worst layout" is raster order over those same lengths
permuted so the heavy rows sit first or last. Both columns are normalized to LPT at the same
spread.

| LUT spread (max/mean) | raster, best layout | raster, worst layout |
|---|---|---|
| 1.0 (uniform) | 0.99x | 1.00x |
| 2.1 | 0.99x | 1.06x |
| 4.2 | 0.99x | 1.21x |
| 7.7 | 0.98x | 1.39x |
| 10.1 | 1.00x | 1.24x |

LPT holds within 1% across all three layouts, while raster costs up to 39% more on identical work
depending only on where the heavy rows sit. When raster order already happens to be longest-first
the two agree, which is what identifies the ordering rather than any per-call overhead as the
cause. Uniform lengths order to the identity permutation, so the balanced case pays nothing.

The gain narrows as the table grows: at hq=40 (11840 entries) LPT is again flat while raster
costs at most 6% more, since 46 tiles per CU amortize a straggler far better than 5.8 do.
Ordering matters most in the sequence-parallel sharded case, which is also the case the fused
builder covers.

## Work-table build cost

Ordering the grid costs one device-side pass over the LUT lengths, and composite keys keep it
stable so uniform lengths leave raster order untouched. The builder is shared across
architectures; relative cost on gfx942 matches the gfx950 split:

| entries | path | build cost, relative to 1480 entries fused |
|---|---|---|
| 1480 (hq=5)   | fused rank kernel | 1.0x |
| 5920 (hq=20)  | fused rank kernel | 2.2x |
| 11840 (hq=40) | sort fallback | 4.4x |

At 8192 entries the 8-byte keys fill 64 KB of LDS, which is where the fused kernel stops. As a
share of the whole call the build is 0.8% at hq=5 and 0.1 density on gfx950, rising to 4.3% at
0.0125 density; on the sort fallback at hq=40 it is 3.4% at 0.0125 density, and the same
0.0125 / hq=5 point on gfx942 is ~3%.

End-to-end from the host-side optimizations (Wan 720p, hq=5, 1480 entries, gfx950): the
work-table build is ~4.4× faster (wave-parallel rank count), the full sparse call ~1.12× faster
at 1.6% density, and the packed sparse call ~2.6× at 2% density and ~1.2× at full density against
the pre-optimization baseline.

## Testing

`op_tests/test_mha_v4.py` on gfx950: Covers all-true sparse masks
bitwise-equal to dense across quantized recipes and under GQA; work-table LPT order against a
stable reference over seven LUT-length patterns and table sizes straddling the fused limit;
uniform lengths left in raster order; hd128 rotation against an explicit orthonormal Hadamard
matmul; mask-shape and partial-LUT rejection; BF16 dense dispatch restore; `torch.compile` schema
and parity checks.

The sparse tests now run on gfx942 as well: arch gates changed from "gfx950 only" to "gfx942 or
gfx950", the FP8 parameters resolve through `native_fp8_format()` so gfx942 selects the E4M3FNUZ
manifest row rather than E4M3, the MX-only parameters stay gfx950-skipped, and mask shapes derive
their KV extent from `mha_v4_kv_tile()`.

`op_tests/test_mha_v4.py` on gfx942 (MI300X): Some skips with gfx950-only MX recipes and their
packed-layout / compile-parity cases. Every sparse test that applies to gfx942 passed, and all-true sparse
vs dense is bitwise identical for `fp8` and `i8fp8` at Wan 720p hq=5 and hq=40.

## Documentation

- [x] `aiter/ops/mha_v4.md` — sparse API, GQA, LUT/work-table contract, per-arch sparse geometry

## Dependencies

- [x] No new third-party dependencies. Ten sparse `.co` binaries added: eight under
  `hsa/gfx950/fmha_v4_fwd/` and two under `hsa/gfx942/fmha_v4_fwd/MI300/`.


csrc/include/torch/mha_v4_fwd.h
```diff
@@ -23,5 +23,33 @@ void fmha_v4_fwd(const at::Tensor& q,
                  int64_t v_scale_mode,
                  double softmax_scale);
 
+// Sorted block-sparse sibling. Same packed operands as fmha_v4_fwd, plus a ragged LUT.
+// Builds the work table internally (identity raster if lut_count is uniform, else LPT).
+void fmha_v4_fwd_sparse(const at::Tensor& q,
```
valarLip, line 28: can you please switch the new style by remove libtorch dependcy?

valarLip (changes requested):

csrc/include/torch/mha_v4_fwd.h
nsakkine: Thanks for the pointer — agreed, new code shouldn't be landing in the ATen style when the rest of the ASM launchers have moved on. But would you be open to it as a follow-up? IIUC the reason being that this work can't be scoped to fmha_v4_fwd_sparse alone: that entry shares validate_packed_mha_v4 and populate_dense_kernarg with the dense fmha_v4_fwd, which predates this PR, so the whole translation unit has to convert together.

valarLip (approved):

PR #5007, opened 2026-08-26
[FlyDSL] One-stage split-K for the a8w8 preshuffle GEMM

XiaobingSuper:
## What

Split-K for the FlyDSL a8w8 preshuffle GEMM, plus retuned Kimi-K3 configs for
M <= 32.

K is split across `gridDim.z`. Every split writes its fp32 partial into a
`[k_split, M, N]` workspace, then bumps a per-tile semaphore; the last split to
arrive reduces the k_split planes, converts to bf16/fp16 and resets the
semaphore, all in the same kernel. Accumulation stays fp32 with a single
rounding, so precision matches `k_split == 1`.

The partials are handed between CTAs that may sit on different XCDs, each with
its own L2, so they have to reach a common point. An agent-scope fence does
that with a whole-L2 `buffer_wbl2` per CTA, which also evicts the A/B tiles
every other in-flight CTA is still reading. Marking just these accesses
`sc0|sc1` writes them through and leaves the rest of L2 alone.

`k_split == 1` is untouched — same kernel, same cached stores — and its output
is bitwise identical to main on every shape checked.

## Performance vs main

Production dispatch (`gemm_a8w8_bpreshuffle`), CUDA graph capture, single idle
gfx950. Weights are rotated over ~1 GB of copies so every call reads them cold
from HBM, the way `run_perftest` does — decode sweeps every layer, so a hot
weight tensor is not the case to optimise for. main and branch alternated twice
on the same GPU, best of each; repeatability across the two rounds of the same
config was median 0.37%, p90 0.94%, max 2.69%.

| M | N | K | main (us) | branch (us) | |
|---|---|---|---|---|---|
| 32 | 7168 | 4224 | 26.275 | 11.404 | **+56.6%** |
| 1 | 576 | 7168 | 10.173 | 4.699 | **+53.8%** |
| 8 | 576 | 7168 | 10.574 | 4.931 | **+53.4%** |
| 2 | 576 | 7168 | 10.218 | 4.766 | +53.4% |
| 4 | 576 | 7168 | 10.330 | 5.251 | +49.2% |
| 16 | 576 | 7168 | 9.715 | 5.218 | +46.3% |
| 16 | 1536 | 7168 | 11.649 | 6.297 | +45.9% |
| 8 | 1536 | 7168 | 10.969 | 6.117 | +44.2% |
| 4 | 1536 | 7168 | 10.640 | 6.055 | +43.1% |
| 2 | 1536 | 7168 | 10.520 | 5.990 | +43.1% |
| 1 | 1536 | 7168 | 10.481 | 6.010 | +42.7% |
| 16 | 2176 | 7168 | 11.884 | 6.854 | +42.3% |
| 32 | 576 | 7168 | 10.014 | 5.779 | +42.3% |
| 8 | 2176 | 7168 | 11.096 | 6.672 | +39.9% |
| 4 | 2176 | 7168 | 10.835 | 6.587 | +39.2% |
| 32 | 1536 | 7168 | 11.504 | 7.031 | +38.9% |
| 1 | 2176 | 7168 | 10.702 | 6.542 | +38.9% |
| 2 | 2176 | 7168 | 10.797 | 6.648 | +38.4% |
| 32 | 2176 | 7168 | 11.677 | 7.757 | +33.6% |
| 16 | 6400 | 7168 | 16.336 | 11.684 | +28.5% |
| 32 | 7168 | 1536 | 5.818 | 4.906 | +15.7% |
| 32 | 8448 | 7168 | 21.379 | 18.558 | +13.2% |
| 4 | 2304 | 1536 | 4.213 | 3.993 | +5.2% |
| 2 | 2304 | 1536 | 4.154 | 3.968 | +4.5% |
| 1 | 2304 | 1536 | 4.135 | 3.951 | +4.4% |
| 8 | 2304 | 1536 | 4.224 | 4.049 | +4.1% |
| 32 | 2304 | 1536 | 4.291 | 4.148 | +3.3% |
| 32 | 6400 | 7168 | 15.652 | 16.002 | -2.2% |

```
56 shapes total:  503.8 -> 394.9 us    +21.6%
geomean speedup:  1.244x
27 faster / 1 slower / 28 unchanged (within 2%)
```

The wins land on long K with small N, where the tile grid does not fill the GPU
and K is long enough to split. The unchanged shapes are mostly K <= 1536. The
single regression is at the edge of the run-to-run spread.

## Configs

29 rows, all M <= 32: 21 stay on flydsl with a better config, 8 move from ck to
flydsl. 20 use split-K, mostly `k_split=7`. Every changed row was re-measured
old config against new on an idle GPU, and rows that were not actually faster
there were kept at main's value.

## Testing

- Precision: M in {1, 8, 64} x k_split in {1, 2, 7, 14}, error flat against
  `k_split == 1`
- 50 CUDA graph replays, including ragged M=17, so the semaphore returns to
  zero and the workspace is safe to reuse
- `k_split == 1` output bitwise identical to main
- Kernel-name parsing for the dispatch and AOT parsers, with and without the
  scheduler token, xcd/lds tokens, and `_ksN`


aiter/ops/flydsl/kernels/preshuffle_gemm.py
```diff
@@ -118,6 +119,34 @@ def _get_preload(tile_m, tile_n, tile_k):
     )
 
 
+# Row bound of the A/C layout views; gemm_kernels mirrors it to guard on the host.
+PRESHUFFLE_M_MAX = 65536
```
coderfeli, line 123: 64K max?

aiter/ops/flydsl/kernels/preshuffle_gemm.py
```diff
@@ -738,12 +806,90 @@ def apply_activation(val_s):
                 T.vec(acc_size, out_elem_cls.ir_type), out_elems
             )
             frag_C_out.store(out_vec)
-            fx.copy(buf_copy_out, frag_C_retile, pC_g)
+
+        fx.copy(buf_copy_out, frag_C_retile, pC_g)
+        if const_expr(split_k > 1):
```
coderfeli, line 782: make it common utils then we can use for other dtypes?

aiter/ops/flydsl/kernels/preshuffle_gemm.py
```diff
@@ -738,12 +806,90 @@ def apply_activation(val_s):
...
+                        col = fx.Int32(bid_y) * tile_n + local_n
+                        if row < i32_m:
+                            out_linear = row * fx.Int32(N) + col
+                            # All loads first, so the adds pipeline behind one
+                            # memory latency.
+                            parts = []
+                            for s in range_constexpr(split_k):
+                                parts.append(
+                                    Vec(
+                                        buffer_ops.buffer_load(
```
coderfeli: copy atom instead of buffer ops?

coderfeli:
Some comments inline. Others look good.

aiter/ops/flydsl/kernels/preshuffle_gemm.py
```diff
@@ -738,12 +806,90 @@ def apply_activation(val_s):
...
+                is_last = (arrival == fx.Int32(split_k - 1)).select(
+                    fx.Int32(1), fx.Int32(0)
+                )
+                fx.ptr_store(Vec.from_elements([is_last], fx.Int32), split_flag_ptr)
+            gpu.barrier()
+
+            is_last = Vec(fx.make_view(split_flag_ptr, fx.make_layout(1, 1)).load())[0]
+            if is_last != fx.Int32(0):
+                workspace_rsrc = buffer_ops.create_buffer_resource(arg_c, max_size=True)
+                out_rsrc = buffer_ops.create_buffer_resource(
```
xytpai: Please remove `buffer_ops`

aiter/ops/flydsl/kernels/preshuffle_gemm.py
XiaobingSuper: `M_max = 65536` is pre-existing on main (`preshuffle_gemm.py:812` there) -- it is the row bound of the A/C layout views, not something this PR introduces. What the PR adds is the name `PRESHUFFLE_M_MAX` and a matching host-side guard in `gemm_kernels.py` so an M above it fails loudly instead of silently wrapping. Happy to raise the bound in a follow-up if any model needs M > 64K in one launch.

aiter/ops/flydsl/kernels/preshuffle_gemm.py
XiaobingSuper: Done -- moved to `aiter/ops/flydsl/kernels/splitk_epilogue.py` as `splitk_reduce_epilogue(...)`. `out_elem_cls` is the only dtype knob, so any 16/32-bit output dtype works; the semaphore protocol and the sc0|sc1 release are part of the util.

aiter/ops/flydsl/kernels/preshuffle_gemm.py
XiaobingSuper: Done. The reduce now builds a buffer-tensor descriptor and partitions it with copy atoms (`make_layout_tv` -> `make_tiled_copy` -> `fx.copy(..., soffset=)`). The TV layout gives each thread 4 contiguous columns so the loads stay dwordx4 and the stores dwordx2, and the descriptor bounds cover the ragged-M tail. On the 20 tuned Kimi-K3 split-K shapes it is 0-3% faster than the buffer_ops version it replaces.

aiter/ops/flydsl/kernels/preshuffle_gemm.py
XiaobingSuper: Removed from the split-K path -- the reduce is copy-atom based now, see the reply above. The remaining `buffer_ops` uses in this file are the pre-existing scale_a/scale_b/bias epilogue loads on main, which this PR does not touch; I would rather convert those in a separate change so this one stays reviewable against the split-K feature.

xytpai (approved):

coderfeli (approved):

PR #5009, opened 2026-08-26
[HIP] [Feature] Radix-select top-k for wide ungrouped MoE routers

fanxingran:
grouped_topk_kernel runs `topk` serial rounds of scan / wave arg-max /
write -INF back to LDS, so latency scales as topk * ceil(num_experts /
(WARP_SIZE * vec_size)) -- 8 rounds for the shape it was tuned for (256
experts, 8 groups, top-8) but 64 for a wide ungrouped router such as
Kimi-K3's 896 experts at top-16.

Add grouped_topk_radix_kernel for that regime: scores stay in registers and
the exact topk-th largest is found by radix select over the order-preserving
uint32 image of the score, with DPP for the wave prefix sum and ballot+mbcnt
for slot assignment. Dispatch is gated to the measured win (single expert
group, sigmoid, topk >= 16 plus a narrow top-8 window), leaving 472
unclaimed shapes verified bit-identical to stock. row_bcast DPP was removed
in GFX10, so the wave scan selects it under #if defined(__GFX9__) with a
__shfl_up fallback.

gfx950, 896 experts at top-16: 1.33x - 2.07x per call across 1..163840
tokens; +1.54% output tok/s in a vLLM TP8 serving A/B over twelve alternating
rounds; gsm8k 1319 questions 0.9644 against 0.9636 stock.

## Motivation

`grouped_topk_kernel` selects the top-k experts by running `topk` serial rounds of
scan / wave arg-max / write `-INF` back to LDS. Latency therefore scales as
`topk * ceil(num_experts / (WARP_SIZE * vec_size))`: 8 rounds for the shape the
kernel was tuned for (256 experts, 8 groups, top-8), but 64 rounds for a wide
*ungrouped* router such as Kimi-K3's 896 experts at top-16.

The router sits on the decode critical path, so those extra rounds show up
directly in inter-token latency. This PR adds a kernel for that regime and leaves
everything else on the existing path.

## Technical Details

New `grouped_topk_radix_kernel`. Instead of repeatedly masking the winner in LDS,
scores stay in registers and the exact `topk`-th largest value is found by radix
select over the order-preserving `uint32` bit image of the score. Per pass, a
bucket histogram narrows the candidate range; once the threshold is known, one
ballot plus `mbcnt` assigns output slots so weights stay paired with their ids.
The wave prefix sum uses DPP. LDS use drops to a single bucket histogram per pass,
versus `2 * num_experts` floats for the scan kernel.

`row_bcast` DPP was removed in GFX10, so the wave scan selects it under
`#if defined(__GFX9__)` and falls back to `__shfl_up` elsewhere. The fallback only
needs to compile and stay correct, since the dispatch gate requires wave64.

Dispatch is gated to the measured win, so unclaimed shapes keep bit-identical
behaviour:

```cpp
if(num_expert_group != 1 || is_softmax)          return false;   // grouped reduction wants LDS
if(get_warp_size_func() != 64)                   return false;
if(num_experts > RADIX_SLOTS * 64 * vec_width)   return false;   // register capacity
if(topk >= 16)                                   return true;
return topk >= 8 && num_experts <= 512 && num_tokens <= 1024;
```

The token-count bound on the top-8 branch is deliberate: that shape wins at low
token counts but loses roughly 7% once the GPU saturates, so the gate stops
claiming it there.

| file | change |
| --- | --- |
| `csrc/kernels/topk_softmax_kernels_group.cu` | +440 |
| `op_tests/test_moeTopkSoftmax.py` | +211 |

## Test Plan

All on gfx950 (MI355X).

1. New contract tests in `op_tests/test_moeTopkSoftmax.py`, runnable standalone
   via `python op_tests/test_moeTopkSoftmax.py --radix-tests-only`:
   - envelope: 9 expert counts (896 / 1024 / 1028 / 510 / 514 / 898 / 255 / 257 /
     897, covering the vec4, vec2 and vec1 widths) x topk 8 / 16 / 32 x fp32 /
     bf16 x 1 / 3 / 64 / 1024 tokens, against a torch reference;
   - ties: 2, 3 and 5 distinct logit values, which forces the non-exact radix
     path, for (896, 16), (510, 8) and (255, 16).
2. aiter's existing benchmark suite in the same file, base versus new, across the
   Kimi-K3, Kimi-K2.5 and DeepSeek-R1 router shapes.
3. Bit-identity check for shapes the gate does not claim: same ids, same weights,
   no tolerance.
4. vLLM TP8 serving A/B on Kimi-K3-MXFP4, alternating rounds to cancel drift.
5. gsm8k, 1319 questions.

## Test Result

**Correctness.** 2592 envelope cases and 72 ties cases pass, 0 failures. Verified
over fp32 / bf16 / fp16, token counts 1 to 163840, strided logits, renorm on and
off, routed scaling factors, heavy ties, both the biased and non-biased entry
points, and all three vector widths.

**472 shapes outside the gate are bit-identical to the stock build** — every
expert-group count, both scoring functions, both renorm settings.

**Per call, Kimi-K3 (896 experts, 1 group, top-16):** 1.33x - 2.07x across
1..163840 tokens, with no measured regression at any point.

| tokens | stock | radix | speedup |
| --- | --- | --- | --- |
| 1 | 13.19 us | 7.71 us | 1.71x |
| 1024 | 13.21 us | 9.93 us | 1.33x |
| 16384 | 65.98 us | 39.57 us | 1.67x |
| 163840 | 552.38 us | 334.55 us | 1.65x |

**Neighbouring shapes.** Kimi-K2.5 (384 experts, 1 group, top-8): 1.28x at 1
token, 1.13x at 1024, unchanged from 2048 up where the gate stops claiming it.
DeepSeek-R1 (256 experts, 8 groups): 1.00x - 1.04x, i.e. untouched.

**End to end.** vLLM TP8 serving A/B on Kimi-K3-MXFP4, 12 alternating rounds,
n=6 per arm:

| metric | stock | radix | delta |
| --- | --- | --- | --- |
| output tok/s | 578.50 [576.81, 580.06] | 587.43 [584.52, 589.72] | **+1.54%** |
| TPOT | 27.07 ms | 26.66 ms | -1.55% |
| ITL p50 | 26.43 ms | 26.00 ms | -1.62% |
| e2e mean | 28313 ms | 27882 ms | -1.52% |

The two arms do not overlap on any of the four metrics across all 12 rounds;
stock's best round and radix's worst round are still 4.5 tok/s apart. TTFT is
unchanged (580 - 633 ms both arms), which is what a decode-path router change
should look like.

**Accuracy.** gsm8k, 1319 questions: 0.9644 with this change against 0.9636 on
stock.

**Pre-existing flake, unrelated to this change.** `op_tests` reports an
intermittent tie-ordering mismatch in the 256-expert 8-group family at a
different token count on each run, because that test does not seed its input.
The radix gate cannot reach that family (it requires `num_expert_group == 1`),
and the family is among the 472 shapes verified bit-identical above. Over five
runs of the same command: stock 11 failures, this branch 1.

## Submission Checklist

### DIFF
--- a/csrc/kernels/topk_softmax_kernels_group.cu
+++ b/csrc/kernels/topk_softmax_kernels_group.cu
@@ -280,6 +280,98 @@ __inline__ __device__ void warpReduceMax(float& val_o, int& idx)
     //         }
 }
 
+// ---------------------------------------------------------------------------
+// Helpers for grouped_topk_radix_kernel.
+
+struct radix_topk_traits
+{
+    // Bits per radix pass, buckets per pass, and passes to resolve a full float
+    // key. Four 8-bit digits cover the key; two passes suffice on real routing
+    // data.
+    static constexpr int Bits   = 8;
+    static constexpr int Bins   = 1 << Bits;
+    static constexpr int Passes = 32 / Bits;
+    // Vector registers of scores each lane holds. This fixes the register
+    // footprint at Slots * vec_size values per lane for every instantiation,
+    // and caps the expert count the kernel can serve at
+    // Slots * WARP_SIZE * vec_size (1024 for the vec4 path, which is what wide
+    // routers like Kimi-K3 use).
+    static constexpr int Slots = 4;
+};
+
+// Order-preserving map float -> uint32, so integer compares reproduce float
+// compares. Positive floats keep their order once the sign bit is set;
+// negative floats need a full bit flip.
+__device__ __forceinline__ unsigned int radix_mono_key(float f)
+{
+    unsigned int b = __builtin_bit_cast(unsigned int, f);
+    return (b & 0x80000000u) ? ~b : (b | 0x80000000u);
+}
+
+// Set bits strictly below this lane. mbcnt is the native wave prefix popcount.
+__device__ __forceinline__ int radix_ballot_rank(unsigned long long m)
+{
+    return static_cast<int>(__builtin_amdgcn_mbcnt_hi(
+        static_cast<unsigned int>(m >> 32),
+        __builtin_amdgcn_mbcnt_lo(static_cast<unsigned int>(m), 0u)));
+}
+
+template <int CTRL, int ROW_MASK, int BANK_MASK, typename T>
+__device__ __forceinline__ T radix_dpp_fetch(T x)
+{
+    // old = 0, so lanes disabled by the masks and lanes whose source lane is
+    // out of range contribute nothing to the accumulation.
+    return __builtin_bit_cast(T,
+                              __builtin_amdgcn_update_dpp(0,
+                                                          __builtin_bit_cast(int, x),
+                                                          CTRL,
+                                                          ROW_MASK,
+                                                          BANK_MASK,
+                                                          false));
+}
+
+// Inclusive add scan across the wave, kept in the VALU. Hillis-Steele via
+// row_shr inside each row of 16 lanes, then row_bcast to carry across rows.
+// This is the hot alternative to a __shfl_up chain, which lowers to
+// ds_bpermute and pays LDS latency on every step.
+template <typename T>
+__device__ __forceinline__ T radix_wave_scan_add(T x)
+{
+#if defined(__GFX9__)
+    x += radix_dpp_fetch<0x111, 0xf, 0xf>(x); // row_shr:1
+    x += radix_dpp_fetch<0x112, 0xf, 0xf>(x); // row_shr:2
+    x += radix_dpp_fetch<0x114, 0xf, 0xf>(x); // row_shr:4
+    x += radix_dpp_fetch<0x118, 0xf, 0xf>(x); // row_shr:8
+    x += radix_dpp_fetch<0x142, 0xa, 0xf>(x); // row_bcast:15 -> rows 1,3
+    if constexpr(WARP_SIZE > 32)
+        x += radix_dpp_fetch<0x143, 0xc, 0xf>(x); // row_bcast:31 -> rows 2,3
+    return x;
+#else
+    // GFX10+ dropped the row_bcast DPP modes this scan carries rows with
+    // ("broadcasts are not supported on GFX10+"), the same reason
+    // wave_reduce() in hip_reduce.h guards them. The radix path only dispatches
+    // on wave64 (see radix_topk_applicable below), so non-GFX9 targets just
+    // need something that compiles and stays correct if it ever runs; speed
+    // there is irrelevant.
+    const int lane = static_cast<int>(__lane_id());
+    for(int off = 1; off < WARP_SIZE; off <<= 1)
+    {
+        T up = __shfl_up(x, off, WARP_SIZE);
+        if(lane >= off)
+            x += up;
+    }
+    return x;
+#endif
+}
+
+template <typename T>
+__device__ __forceinline__ T radix_wave_total_add(T x)
+{
+    return __builtin_bit_cast(
+        T,
+        __builtin_amdgcn_readlane(__builtin_bit_cast(int, radix_wave_scan_add(x)), WARP_SIZE - 1));
+}
+
 __device__ void blockReduceMax(float& val, int& idx)
 {
     __shared__ float shared_vals[32];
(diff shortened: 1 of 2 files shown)


fanxingran:
@carlushuang @junhaha666 could you take a look? This adds a radix-select path inside `grouped_topk` for the wide ungrouped regime Kimi-K3 lands in (896 experts, `num_expert_group=1`, sigmoid, top-16).

The gate keeps everything else untouched: 472 shapes outside it are verified bit-identical to stock, DeepSeek-R1 unaffected. Thanks a lot!

fanxingran:
@carlushuang thanks for the review! The four namespace-scope `static constexpr` are gone, folded into `struct radix_topk_traits` in the same shape. Thanks again!

valarLip:
can you fix [Checks / Check Code Style with Black (pull_request)](https://github.com/ROCm/aiter/actions/runs/33032945731/job/98469571320?pr=5009)

fanxingran:
> can you fix [Checks / Check Code Style with Black (pull_request)](https://github.com/ROCm/aiter/actions/runs/33032945731/job/98469571320?pr=5009)

Fixed — three hand-wrapped asserts in the new test now go through black 26.5.1, formatting only and AST-identical, squashed into the existing commit. Thanks for review!

fanxingran:
@valarLip Black is fixed — the whole tree outside 3rdparty/ is clean locally. The workflows on the new commit are all sitting in action_required, could you approve the run? The other red checks in the last round were all downstream of Black (check_signal.sh gates on the Checks conclusion), and the Kimi-K2.5 perf (vllm) perf floor also fired on main's own nightly on Aug 26 and Aug 27, so it's unrelated to this change. Thanks!

valarLip (changes requested):
## Review

Nice piece of work — the radix-select algorithm itself is sound. I simulated the exact lane/slot/ballot layout on CPU over 20,000 randomized cases (all vector widths, expert counts 64–1024, topk 1..num_experts, 1–8 distinct score levels including all-equal) and it wrote exactly `topk` distinct in-range slots every time; the DPP wave64 scan matches LLVM's canonical `buildScan` and was confirmed exact on real gfx950. Recording that up front so it doesn't get re-litigated.

The findings below are about the edges around that core: the dispatch gate, inputs the gate lets through, and what the new tests can actually detect.

### Blocking

**1. `csrc/kernels/topk_softmax_kernels_group.cu:869` — `topk > num_experts` produces an empty ballot, and topk slots are left unwritten.**
`radix_topk_applicable` never checks `topk <= num_experts`. Take `num_experts=8, num_expert_group=1, sigmoid, topk=16` on wave64: the gate returns true on `topk >= 16`. Pass-0 total population is 8 < need=16, so `__ballot(found) == 0` and `__ffsll(0) - 1 == -1`. I verified on gfx950 that `__builtin_amdgcn_readlane(v, -1)` returns lane 63; lane 63 has `hit == 0`, so `packed == 0` → `need = 0, cnt = 0` → `exact` is true with `thresh == 0`. The emit loop then writes the 8 live experts into slots 0..7 and slots 8..15 are never written at all. The caller gets whatever was in the buffer (the op_tests pre-fill is -1) and `fused_moe`/`moe_sorting` gathers on a garbage or -1 expert id. The stock `grouped_topk_kernel` writes all `topk` slots for the same input. Fix: `if(topk > num_experts) return false;` in the gate, and/or guard the empty ballot.

**2. `:305` — `radix_mono_key` ranks NaN above `+inf`, so a NaN score is always selected first.**
Positive NaN maps to `0xFFC00000`, strictly above `+inf`. That's the opposite of both stock kernels, which either never select NaN (`tmp[i] > max_val` is false) or explicitly sanitize it (`gating[i] = ::isnan(gating[i]) ? -INFINITY : gating[i];`). With one NaN logit or one NaN `correction_bias` entry in a Kimi-K3 row (896 experts, 1 group, top-16): sigmoid propagates the NaN, the NaN expert takes slot 0, its weight is NaN, and under `need_renorm` the `routed_scaling_factor / radix_wave_total_add(psum)` divisor is NaN — so *all* topk weights for that token become NaN and poison the hidden state. Stock degrades gracefully: the NaN expert is simply never picked. And because the kernel is chosen by a heuristic that includes `num_tokens`, whether a NaN corrupts the row now depends on batch size. Please apply the same `isnan -> -INFINITY` sanitization before `radix_mono_key`.

**3. `:1646` — `num_tokens` in the dispatch gate makes the output permutation a function of batch size.**
`return topk >= 8 && num_experts <= 512 && num_tokens <= 1024;` selects between two kernels with *different output orders* (stock: descending score; radix: ballot order). So a 384-expert / top-8 ungrouped router (Kimi-K2.5, named in the PR body) returns ballot-ordered ids at M=1024 and descending-score ids at M=1025, from the same weights, in the same process. Consequences:
- Any consumer treating `topk_ids[:,0]` as the argmax expert, or overwriting `topk_ids[:,-1]` for shared-expert fusion — sglang's `topk.py` does exactly this — gets a different expert on decode vs prefill.
- `fused_moe` accumulates in slot order, so the MoE output is bitwise different between decode and prefill for the same token, breaking prefix-cache and speculative-decode bit-exactness.

aiter's own op_tests can't catch this: every comparison sorts by id first. Either make the gate a pure function of model shape, or normalize the emission order between the two kernels.

### Should fix before merge

**4. `:731` — the stated output-order contract doesn't match the code.**
The kernel header comment (and the PR body) say "Selected experts are emitted in ascending expert order", but the emit loop is `for s { for j { ballot over lanes } }` with `id = (s*WARP_SIZE + lane)*vec_size + j`, so for `vec_size > 1` the order is j-major. Simulating the exact emit order for 896 experts / vec4 / top-8 gives `[53, 10, 67, 83, 383, 762, 782, 879]` — grouped by `id mod 4`, neither ascending id nor descending score (true descending is `[83, 782, 762, 10, 879, 383, 53, 67]`). Anyone who trusts the stated contract and skips a sort before `moe_sorting` gets a wrong result. The ordering is also documented only in the `.cu` comment — `aiter/ops/topk.py:103/176`, `csrc/include/moe_op.h`, and `rocm_ops.hpp` carry no ordering statement and weren't updated.

**5. `:893` — both emit loops read uninitialized registers.**
`key[SLOTS][vec_size]` and `weight[...]` are only written under `if(!live[s]) continue;`, but line 893 computes `const unsigned int k = key[s][j] & hi_mask;` unconditionally, *before* the `live[s] &&` short-circuit. That's UB / LLVM poison. Clang currently lowers `&&` to a poison-safe `select`, but InstCombine folds select → `and i1` when it can prove the operand isn't poison-generating, and `and i1 poison, false` is poison. A future LLVM, or an unroll/rotation that speculates the `&`, makes `hit` poison for a dead slot — which then feeds `__ballot(hit)`, a convergent op whose result is wave-uniform, corrupting `base`, every lane's `slot[][]`, and `psum` for the whole wave. Worst for `num_experts <= 256` with vec1, where 3 of 4 SLOTS are dead. One-line fix: set `key[s][j] = 0u; weight[s][j] = 0.0f;` in the `!live` branch instead of `continue`.

**6. `:801` — the unbiased instantiation formally reads an uninitialized `bias`.**
`vec_i bias;` is declared unconditionally, assigned only under `if constexpr(isBiased)`, then consumed through a *runtime* ternary: `radix_mono_key(isBiased ? sig + static_cast<float>(bias[j]) : sig)`. Since `isBiased` is a template bool, -O3 folds it and this is harmless today — but the false instantiation is still formed with a live read of `bias[j]`, and the unbiased launcher passes `nullptr` for `correction_bias`. Any codegen that keeps the read alive (-O0, sanitizers, or an opus vector type with a non-trivial `operator[]`) turns it into a null deref. Suggest `if constexpr(isBiased) key[s][j] = radix_mono_key(sig + (float)bias[j]); else key[s][j] = radix_mono_key(sig);` with the declaration moved inside the constexpr branch.

**7. `:1641` — the gate has no architecture check, but every perf number justifying it is gfx950-only.**
The only hardware condition is `get_warp_size_func() != 64`, while the comment above states "gfx950, ungrouped, tokens 1..4096". gfx90a (MI200) and gfx942 (MI300X) are also wave64 and build this TU, so they silently switch to the new kernel on an extrapolated measurement. Concretely, a 128-expert / top-8 / ungrouped router on MI300X satisfies `topk >= 8 && num_experts <= 512 && num_tokens <= 1024` and takes the radix path. The kernel's cost is dominated by 4×256 LDS atomicAdds plus a per-token DPP scan, both of which have different throughput on gfx90a — so it can be a net regression there while the comment asserts "wins everywhere, 1.35x-4.3x". The adjacent `use_opt_sort` gate at least pins an exact shape. Please add a gfx950 (or explicitly enumerated arch) condition, or re-measure.

**8. `:1616` — `radix_vec_width()` is a hand copy of `LAUNCH_KERNEL()`'s residue switch, with no static_assert.**
The capacity gate `num_experts > Slots * 64 * radix_vec_width(...)` depends on the two never drifting, and additionally hardcodes 64 instead of the wave size it just queried. It's correct today (verified: `>` is the right comparison — at exactly 1024 experts with vec4, `num_vec = 256` and max `v = 3*64+63 = 255 < 256`). But add a vec8 case to `LAUNCH_KERNEL`, or change the residue rule, and the gate over-estimates capacity: it would admit e.g. 768 experts while `LAUNCH_KERNEL` instantiates a narrower vector, real capacity is `SLOTS*WARP_SIZE*vec`, and `live[s] = v < num_vec` never covers the tail. Those experts are never loaded, never histogrammed, never eligible — a top-k over a truncated prefix of the row, with correct-looking normalized weights, no crash and no assert. Please derive the width from one shared helper used by both sites.

**9. `:1494` — `if constexpr(NUM_GRP == 1)` with no `else` is a silent no-launch path.**
Both new launcher macros wrap the whole launch in `if constexpr(NUM_GRP == 1)`, so under LAUNCHER4's `if(use_radix){...}` a `NUM_GRP != 1` instantiation expands to `if(use_radix) { }` — no kernel, no error. The comment justifying the guard ("only keeps the other NUM_GRP instantiations from being emitted") isn't accurate: `grouped_topk_radix_kernel`'s template list is `<DTYPE_I, f32vec, need_renorm, isBiased>` and doesn't depend on `NUM_GRP` at all. Unreachable today because LAUNCHER2 switches on the same `num_expert_group` the gate tests — but the coupling lives in three places that don't reference each other, and `grouped_topk` carries `// TODO: expand usage in the future` right above `use_opt_sort`. Relax the gate to allow `num_expert_group == 2` and the build stays green, no kernel launches, and `biased_grouped_topk` returns with `topk_ids`/`topk_weights` holding whatever the caller's `torch.empty_strided` allocation contained. Please add `else { AITER_CHECK(false, ...); }`. While there: the two 23-line macros differ in only three tokens (dispatch name, isBiased, bias pointer vs nullptr) and could collapse into one parameterized by `isBiased` + a `BIAS_ARG`, matching the existing `LAUNCHER_grouped_topk_kernel` shape.

**10. `:861` — the 12-bit `hit` packing has no static_assert tying it to the expert cap.**
`(bucket << 24) | ((need - before) << 12) | mine[i]` gives `need - before` and `mine[i]` 12 bits each. Safe today since `Slots = 4` caps `num_experts` at 4*64*4 = 1024 < 4096 — but the traits comment explicitly frames `Slots` as the knob for wider routers. Raising it to 16 for a 4096-expert router lets a single bucket population overflow `cnt` into the `need` field and `need` into the `bucket` field, giving a silently wrong thresh/need: wrong experts, no crash. Add `static_assert(radix_topk_traits::Slots * 64 * 4 < (1 << 12));` next to the traits, or widen the pack.

### Reuse / efficiency

**11. `:319` — the new DPP/scan/key helpers duplicate primitives already available in this exact TU.**
- `radix_dpp_fetch` == `aiter::upd_dpp_` (`csrc/include/warp_sort.h:30-49`, already included)
- `radix_wave_scan_add` == `warp_cumsum` (`warp_sort.h:1763` — already called from *this file* at lines 114, 119, 986, 996)
- `radix_wave_total_add` == `wave_reduce(x, Sum{})` (`hip_reduce.h:257` — already called from this file at line 740)
- `radix_mono_key` is a verbatim copy of `pack_score_key`'s body in `csrc/kernels/sparse_attn/msa/pa_sparse_block_select_kernels.cuh:707-711`

That's four independent copies of the same DPP plumbing and two of the float→orderable-uint32 trick. When a future target changes the `#if defined(__GFX9__)` row_bcast assumptions, whoever updates `hip_reduce.h`/`warp_sort.h` has no reason to look inside `topk_softmax_kernels_group.cu`, so this private copy keeps the stale lowering and produces a wrong prefix sum — wrong topk ids, not a compile error. Same hazard for a NaN fix (finding #2) applied to one `radix_mono_key` copy and not the other. Worth noting the existing `warp_cumsum` also carries a `ds_bpermute` cross-row carry that works on GFX10+, so reusing it would remove the new `#if`/`__shfl_up` fallback entirely.

**12. `:766` — histogram sizing, the unrolled break-loop, and dead-slot ballots.**
The histogram is sized and zeroed for all 4 PASSES up front (4096 B LDS, 16 stores per lane) even though the loop is documented to break after 2, and `#pragma unroll` on a pass loop carrying a runtime `break` replicates the whole body 4×. Lane *l* reads only the `BINS_PER_LANE` buckets it exclusively owns, so a single 256-entry array reused across passes (zero-after-read plus one extra `__syncthreads()`, near-free for a one-wave block) drops LDS from 4096 B to 1024 B. At 4 KB per single-wave block, concurrent blocks per CU caps at 40 on MI355X's 160 KB, and this kernel launches one block per token — block occupancy is exactly what hides HBM latency at large M.

Separately, both emit loops issue 16 unconditional `__ballot` + `__popcll` + mbcnt pairs even when whole slots are dead for every lane. `active_slots = min(SLOTS, ceil(num_vec / WARP_SIZE))` recovers them: at 384 experts (vec4, num_vec=96) half the ballots are wasted; at 256 experts with topk>=16 (num_vec=64) 12 of 16 are — roughly 60 instructions of serial, fully dependent cross-lane work (`base += __popcll(m)` feeds the next iteration) in a kernel whose entire claimed win at one token is ~2x. Kimi-K3's 896/vec4 fills all four slots, so benchmarking only the headline shape hides this.

### Tests

**13. `op_tests/test_moeTopkSoftmax.py:836` — the new tests run at import time, with no `__main__` guard, and end in `SystemExit`.**
Both functions are invoked unconditionally at module top level, after `args = parser.parse_args()` and before every existing benchmark table — against this repo's op-test convention that all argparse and the sweep live inside `main()`. The envelope grid is 9 experts × 3 topk × 2 dtypes × 4 token counts × 2 renorm × 3 scales × 2 biased = 2592 op invocations (~5.1 s measured on MI355X), plus 72 ties cases, plus a `torch.topk` reference per case and up to 64 per-row `torch.unique` calls (~165k extra device round trips). CI runs `timeout 60m python3 op_tests/test_moeTopkSoftmax.py`, and `split_tests.sh:108` still budgets `FILE_TIMES[op_tests/test_moeTopkSoftmax.py]=39` — not updated in this PR — so the greedy shard balancer now mis-sizes that shard. And because the block ends in `raise SystemExit(...)`, one bf16 tolerance blip in 2664 cases aborts before *any* pre-existing benchmark table is produced. `--radix-tests-only` only skips the tables; it never skips the tests.

**14. `:575` — the new tests can't detect the failure modes this PR introduces.**
- `except Exception` swallows every failure into a counter, so a HIP launch failure or an `AITER_CHECK` abort is indistinguishable from a contract violation, and one bad case cascades into 2591 misleading "FAIL" lines.
- The duplicate-id check is capped at `min(ids.shape[0], 64)` rows, so a slot collision from the `base`/`need` bookkeeping on row 700 of an M=1024 batch passes — 94% of rows are never checked, and M=1024 is exactly the boundary the `num_tokens <= 1024` gate turns on.
- Nothing asserts the radix kernel was actually dispatched — the docstring itself says the checks pass on stock aiter. If `radix_topk_applicable` ever regresses to always-false (a gate typo, or the arch condition from finding #7 added later), all 2664 assertions still pass and the PR's entire speedup silently disappears with green CI. Nine of the swept expert counts (897, 898, 1028, the wide topk=32 combos) already fall outside the gate and silently exercise stock, with nothing distinguishing them.
- The kernel's only strided write path (`w_out = topk_weights + token_idx * stride_tk`) is untested with a non-unit stride, even though this file's own `test_grouped_topk` deliberately uses `torch.empty_strided((token, topk), (topk + 10, 1))`.

**15. `:828` — conventions.**
`--radix-tests-only` is the bespoke behavior-toggle flag the op-test convention rules out (every flag should be a list the sweep iterates); the new tests also skip `@benchmark()`, `run_perftest`, `checkAllclose` (already imported at line 17), and the summary table. The hand-rolled tolerances are the practical problem: `slack = 2e-3 if logits.dtype != torch.float32 else 2e-6` and `wtol = 3e-3` replace the shared comparator, and 2e-3 is loose for sigmoid outputs over 896 experts, where adjacent order statistics near the top-k boundary are routinely closer than that — genuinely wrong selections pass. A shared reference already exists (`biased_grouped_topk_torch`, `aiter/ops/topk.py:217`) and isn't used.

PR #5010, opened 2026-08-26
[Triton/Gluon] Support caller-defined padding cache slot in fused MLA writer

tanth47:
## Summary

Add an optional `pad_slot_id` to `fused_qk_rope_cat_and_cache_mla` so callers can reserve a non-negative cache slot for padding without changing the default AITER behavior.

The argument defaults to `-1`, so slot 0 remains writable for existing callers. When a caller opts in, both the decode/mixed and extra-prefill KV-write branches skip the selected slot. The behavior is implemented consistently in the Triton and gfx1250 Gluon kernels.

## Motivation

Some runtimes reserve a physical KV-cache slot while using padding rows in `slot_mapping`. The generic SGLang MLA writers already protect this invariant in [sgl-project/sglang#36003](https://github.com/sgl-project/sglang/pull/36003), but the fused AITER writer bypasses those writers and can still overwrite the reserved slot.

This change keeps AITER generic: no slot is reserved by default, and callers explicitly select one when needed. Downstream context is tracked in [sgl-project/sglang#36207](https://github.com/sgl-project/sglang/issues/36207).

## Tests

Added focused coverage that verifies:

- default behavior still writes slot 0;
- `pad_slot_id=0` preserves slot 0 in the decode/mixed branch;
- `pad_slot_id=0` preserves slot 0 in the extra-prefill branch;
- ordinary positive slots are still written.

The new value is a compile-time kernel argument, so the default predicate can be constant-folded to the existing behavior.

Local checks: Python compilation, Black, Ruff, and `git diff --check`. GPU execution is left to AITER CI.

### DIFF
--- a/aiter/ops/triton/_gluon_kernels/gfx1250/fusions/fused_kv_cache.py
+++ b/aiter/ops/triton/_gluon_kernels/gfx1250/fusions/fused_kv_cache.py
@@ -419,6 +419,7 @@ def _fused_qk_rope_cat_and_cache_mla_kernel(
     OUTPUT_Q_NOPE_ZEROS_AND_Q_PE: gl.constexpr = False,
     HAVE_K_SCALE: gl.constexpr = False,
     UPCAST_OPERAND: gl.constexpr = False,
+    PAD_SLOT_ID: gl.constexpr = -1,
 ):
     # 1-warp (wave32) blocked layouts matching the Triton-generated ttgir.
     L_NOPE: gl.constexpr = gl.BlockedLayout(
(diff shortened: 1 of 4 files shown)


tanth47:
@k50112113 Could you take a look and approve the fork workflows? This adds an opt-in padding-slot guard to the fused MLA writer while preserving the existing default. It also updates the gfx1250 Gluon mirror; please add `ci:gfx1250-ffm-triton` if that suite is appropriate (contributors cannot apply the label).

PR #4646, opened 2026-08-09
FlyDSL: port a16wi4 to new pipeline, clean old moe_gemm_2stage

coderfeli:
FlyDSL: port a16wi4 to new pipeline, clean old moe_gemm_2stage

PR #5017, opened 2026-08-26
[FlyDSL] gfx942 a16wi4: pack f32->bf16 with lshr-16 instead of scalar

msaffari-amd:
## Summary
On gfx942, FlyDSL a16wi4 (bf16 × int4) became slower after #4646.
gfx942 has no `v_cvt_pk_bf16_f32`. The new pipeline fell back to scalar `f32.to(bf16)` / truncf for the int4 → bf16 upconvert. That path is much more VALU than packing two f32 values into a bf16 pair with `lshr 16`.
This PR restores that pack for **a16wi4 on gfx942 only**. Nibble order is unchanged. gfx950 still uses the packed convert. MXFP4 is not touched.
## Why it is faster
truncf is ~5 VALU ops per element. The pack is:
* take the high 16 bits of the first f32 (`lshr 16`)
* keep the high 16 bits of the second f32
* OR them into one i32 = two bf16s
For scaled int4 this is exact (no extra rounding vs the old truncf fallback).
## What changed
* `_int4_nibble_to_bf16x8(..., use_k16=True)` — gfx942 pack, not truncf

## Perf (convert only, same tile)
MI300X gfx942. a16wi4, SiLU, `h=3584`, `inter=512`, `E=896`, `topk=16`, `ksplit=1`. Same gemm1 tile for both: `16×64×128`, `k_wave=2`. CUDA events, 10 warmup + 51 iters. `origin/main` (truncf) vs this branch (lshr).
| tokens | truncf gemm1 | lshr gemm1 | speedup |
|--------|--------------|------------|---------|
| 1      | 60.5 µs      | 39.0 µs    | **1.55×** |
| 16     | 283.5 µs     | 169.1 µs   | **1.68×** |
| 64     | 667.6 µs     | 445.2 µs   | **1.50×** |

### DIFF
--- a/aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm1.py
+++ b/aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm1.py
@@ -496,6 +496,7 @@ def compile_gemm1_a16w4_port(
         arg_out: fx.Int64,
     ):
         lds_raw_ptr = fx.SharedAllocator().allocate(SharedStorage).peek().raw.ptr
+        # gfx942 W upconvert is lshr-16 bf16 pack (utils), not scalar truncf.
         tx_i32 = fx.Int32(gpu.thread_id("x"))
         bx_i32 = fx.Int32(gpu.block_id("x"))
         lane = tx_i32 % fx.Int32(64)
--- a/aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm1.py
+++ b/aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm1.py
@@ -496,7 +496,6 @@ def compile_gemm1_a16w4_port(
         arg_out: fx.Int64,
     ):
         lds_raw_ptr = fx.SharedAllocator().allocate(SharedStorage).peek().raw.ptr
-        # gfx942 W upconvert is lshr-16 bf16 pack (utils), not scalar truncf.
         tx_i32 = fx.Int32(gpu.thread_id("x"))
         bx_i32 = fx.Int32(gpu.block_id("x"))
         lane = tx_i32 % fx.Int32(64)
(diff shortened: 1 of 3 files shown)


aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm1.py
```diff
@@ -496,6 +496,7 @@ def gemm1_kernel(
         arg_out: fx.Int64,
     ):
         lds_raw_ptr = fx.SharedAllocator().allocate(SharedStorage).peek().raw.ptr
+        # gfx942 W upconvert is lshr-16 bf16 pack (utils), not scalar truncf.
```
samremes: comment out of place

aiter/ops/flydsl/kernels/moe_2stage_a16wmix/gemm2.py
```diff
@@ -381,6 +381,7 @@ def gemm2_kernel(
         arg_out: fx.Int64,
     ):
         lds_raw_ptr = fx.SharedAllocator().allocate(SharedStorage).peek().raw.ptr
+        # gfx942 W upconvert is lshr-16 bf16 pack (utils), not scalar truncf.
```
samremes: comment out of place?

aiter/ops/flydsl/kernels/moe_2stage_a16wmix/utils.py
```diff
@@ -442,6 +449,10 @@ def make_b_loader(
 
     Returns a :class:`_BLoader`. The closures emit no IR until called, so building them
     here rather than inline in the kernel body does not perturb instruction order.
+
+    Cache-key note: FlyDSL hashes this factory's source (not nested helpers).
+    gfx942 a16wi4 W upconvert is lshr-16 bf16 pack in ``_int4_nibble_to_bf16x8``,
+    not scalar truncf.
```
samremes: does this need to say what it is not? explain the new behavior only.

msaffari-amd: Done.

coderfeli (approved):

PR #5024, opened 2026-08-26
[HIP] [CK] Add MHA forward tuning scripts and kernel-info dumping utit…

huishi-hs:
…lies

Ship an end-to-end MHA tuning + inspection tooling suite under `csrc/cpp_itfs/mha/tools/`:

1. Runtime CKtile MHA kernel-info dump. Enabled by setting env `AITER_DUMP_MHA_FWD_INFO=<stride>` (int). `AITER_DUMP_MHA_FWD_INFO=1` dumps every FMHA forward shape info; AITER_DUMP_MHA_FWD_INFO_FILE specify output log file. larger stride subsamples one out of every `stride` calls. Used to collect real-workload MHA shape distribution as the input of the subsequent tuning pipeline.

2. `countMhaShape.py` -- process collected dump info: 2.1 stats hot shape, 2.2 produce different MHA groups, 2.3 generate an untuned csv for each specific group, e.g. `mha_untune_0_group_bf16_hq256_hv256_mask2.csv`.

3. `mha_tune.py` -- for each MHA group: 3.1 enumerate possible tile and config, 3.2 build binary for each config, 3.3 benchmark and produce the final tuned csv.

4. `mha_merge_tuned_and_gen_runtime_json.py`: 4.1 merge multiple per-group tuned configs into a single JSON that can override CK-tile's default MHA config at runtime (consumed by the `CK_TILE_FMHA_FWD_CUSTOM_TUNE_CONFIG_FILE` hook on the CK side).

Co-Authored-By: Claude Opus 4.7

## Motivation

Tune CK tile MHA fwd in AITER. more background in https://github.com/ROCm/rocm-libraries/pull/11298

## Technical Details

Decribed in csrc/cpp_itfs/mha/tools/README.md 

## Submission Checklist


csrc/include/mha_fwd_dump.h
```diff
@@ -0,0 +1,182 @@
...
+            // still empty and users need the pointer regardless.
+            std::fprintf(stderr,
+                         "[MHA_FWD] AITER_DUMP_MHA_FWD_INFO enabled, "
+                         "writing to %s\n",
+                         path.c_str());
+            // Flush residual buffer on normal shutdown. SIGKILL bypasses
+            // this, but per-write fflush() below still keeps the log
+            // durable on disk up to the last completed line.
+            std::atexit([] {
+                if(std::FILE* g = mha_dump_sink())
```
asleepzzz, line 116: mha_dump_sink
->atexit ->
mha_dump_sink

please verify more cases

asleepzzz (commented):
LGTM, only some comments

csrc/include/mha_fwd_dump.h
huishi-hs: @asleepzzz Thanks for your comments!

The recursion is syntactic only. std::call_once guarantees the initializer runs exactly once, so the atexit lambda's call to mha_dump_sink() is just a cached load of fp, not a re-entry — no deadlock, no double registration. And atexit fires before function-local static destructors, so fp is still valid when we fclose it. If dumping is disabled or the fopen fails, the handler is never registered in the first place.

Alternative option  is hoisting fp to a file-scope static and drop the self-call.
If removing self-call is prefered, the cleanest fix is to move the FILE* out of mha_dump_sink() into an anonymous-namespace variable, so the atexit lambda closes it directly without re-entering the accessor. 

namespace {
    // File-scope sink state. Declared in an anonymous namespace so it
    // has internal linkage per translation unit that includes this
    // header (this header is only included from mha_fwd.cu and
    // mha_common.h, both of which end up in the same TU set).
    std::FILE*     g_mha_dump_fp = nullptr;
    std::once_flag g_mha_dump_once;
} // namespace

            std::atexit([] {
                if(g_mha_dump_fp)
                {
                    std::fclose(g_mha_dump_fp);
                    g_mha_dump_fp = nullptr;   // defensive, in case
                                               // any late writer races
                                               // with shutdown.
                }
            });

Which is prefered by AITER community rules/styles?

huishi-hs:
> LGTM, only some comments

@asleepzzz following loop is removed from code with file scope static File pointer
mha_dump_sink
->atexit ->
mha_dump_sink

ck pr is merged https://github.com/ROCm/rocm-libraries/pull/11298
This PR is rebased with latest aiter and verified again with latest CK (USE_NEW_UNIFIED_FRAMEWORK=0).
CK MHA fwd tuning from aiter can work when update CK.

Could you please help review updated PR? Thanks!

asleepzzz:
can we disable batch mode tuning and splitkv / appendkv / pagedkv?
and please add unit test of causal / GQA/ seqlen_q != seqlen_k

csrc/include/mha_common.h
```diff
@@ -142,4 +154,110 @@ inline void print_fmha_fwd_args(ARG args)
...
+// =============================================================================
+// Runtime dump of mha forward args (group-mode extension).
+// The torch-free primitives (env-var sampling, mutex, common field builder,
+// batch-mode dumper) are defined in mha_fwd_dump.h and reused here.
+// =============================================================================
+
+// Helper: copy an int32 device tensor of length `n` to a host vector.
+inline std::vector<int32_t> mha_dump_copy_int32_to_host(const at::Tensor& t)
+{
+    auto host = t.to(torch::kCPU, /*non_blocking=*/false, /*copy=*/true).contiguous();
```
rocking5566, line 166: This CPU copy fails inside `torch.cuda.graph()` when dumping is enabled, so collecting shapes can break graph-captured varlen attention. I reproduced the same copy operation on ROCm: it succeeds eagerly but raises during capture (isolated test, not a full PR build).

Could we skip/defer dumping during capture and add a regression test?

csrc/cpp_itfs/mha/tools/mha_gen_runtime_json.py
```diff
@@ -0,0 +1,631 @@
...
+    loaded: List[TunedCsv] = [TunedCsv.load(p) for p in csv_paths]
+
+    # Group by (dtype, compiled_hq, compiled_hv). Each group can contain
+    # multiple csv sources (e.g. same shape tuned twice with different mask
+    # types); item 4 of the spec says we ignore mask/bias/lse/dropout so
+    # we merge them into one tile pool.
+    grouped_rows: Dict[Tuple[str, int, int], List[Dict[str, Any]]] = {}
+    grouped_sources: Dict[Tuple[str, int, int], List[TunedCsv]] = {}
+    for lc in loaded:
+        key = (lc.dtype, lc.compiled_hdim_q, lc.compiled_hdim_v)
```
rocking5566, line 378: Dropping `mode` and `mask_type` here merges groups that may need different tiles. With two CSVs for no-mask and causal attention at the same dimensions and length, only the first winner survives; reversing `--in` order changes the output (reproduced).

Could we preserve these distinctions in dispatch, or reject conflicting inputs?

csrc/cpp_itfs/mha/tools/mha_tune.py
```diff
@@ -0,0 +1,2233 @@
...
+    return [
+        f"-prec={dtype}",
+        f"-mode={mode_int}",  # 0:batch, 1:group
+        "-b=1",  # single sequence
+        f"-h={nhead_q}",
+        f"-h_k={nhead_k}",
+        f"-d={hdim_q_bench}",
+        f"-d_v={hdim_v_bench}",
+        f"-s={max_seqlen}",
+        f"-s_k={max_seqlen}",
```
rocking5566, line 1047: Setting both `-s` and `-s_k` to `max_seqlen` always benchmarks equal Q/K lengths. A workload with Q=512 and K=4096 is therefore measured as 512×512, changing both the work and causal mask geometry.

Could we carry separate Q/K lengths through the CSV and benchmark arguments, or explicitly limit this tool to equal-length workloads?

csrc/cpp_itfs/mha/tools/mha_tune.py
```diff
@@ -0,0 +1,2233 @@
...
+        return 1
+
+    # ---- Auto step 2: build ----------------------------------------------
+    # Check whether every plan already has its target binary on disk. If any
+    # are missing, run the full build stage and keep only the plans that
+    # built successfully; otherwise skip to bench directly.
+    missing = [
+        p
+        for p in plans
+        if not os.path.isfile(_binary_path(str(p.build_dir), args.build_target))
```
rocking5566, line 1704: Checking only whether the binary exists misses changes to its build settings. Reusing `--work-dir` with `--lse 1` after `--lse 0`, for example, rewrites the JSON but skips rebuilding. The old binary may lack the requested kernel and produce `no_perf`/`run_failed`.

Could we validate the build settings before reuse, or include them in the cache key?

csrc/cpp_itfs/mha/tools/mha_gen_runtime_json.py
```diff
@@ -0,0 +1,631 @@
...
+
+    mids: List[int] = []
+    for i in range(n - 1):
+        a = rows[i]["max_seqlen"]
+        b = rows[i + 1]["max_seqlen"]
+        # (a + b) // 2 is well-defined and strictly greater than a whenever
+        # b > a + 1; for adjacent (b == a + 1) it equals a, giving an empty
+        # interval on the smaller side, which is harmless because ranges
+        # are half-open and the boundary flips to the next row anyway.
+        mids.append((a + b) // 2)
```
rocking5566, line 266: For samples 512 and 513, the midpoint is 512, producing `<512` and `>=512`. Length 512 therefore selects the tile tuned for 513 instead of its own winner (reproduced).

Could we use a ceiling midpoint with these half-open intervals and add an adjacent-length regression test?

rocking5566 (changes requested):
Please address the graph-capture and tuning issues in the five inline comments. Validation was limited to static review and targeted remote reproductions; no full PR build or end-to-end MHA test was run.

huishi-hs:
> can we disable batch mode tuning and splitkv / appendkv / pagedkv? and please add unit test of causal / GQA/ seqlen_q != seqlen_k

@asleepzzz  
Thank you for your review! I'd like catch up with you before update this PR.

Since only group/varlen mode is end-to-end validated in this PR, do you suggest tune/codegen path only support group/varlen-only for now? And skip disable batch mode tuning and splitkv / appendkv / pagedkv now?

About " add unit test of causal / GQA/ seqlen_q != seqlen_k"
This PR is a tuning tool and not a kernel change; Just to make sure I understand the intent: do you want me to add unit tests that feed synthetic dump traces covering causal masks, GQA (nhead_q != nhead_k), and seqlen_q != seqlen_k through the full tool chain — from mha_count_shape.py grouping, to mha_tune.py — and assert that the generated CK build command and the tile_example_fmha_fwd invocation are correct for each case?
Or are you looking for something else (e.g. actually running the built binary on GPU and checking numerical correctness)?

PR #5027, opened 2026-08-26
[HIP] [ROCm][Perf] Add head_dim 512 + weightless V-norm to fused_qk_norm_rope_cache_pts_quant_shuffle

mpashkovskii:
## Purpose

Add **head_dim 512** and a **weightless V-norm** to `fused_qk_norm_rope_cache_pts_quant_shuffle` so vLLM can use the fused QK-norm + RoPE + KV-cache op for **Gemma4**, whose full-attention layers use `head_dim=512` and whose every attention layer applies a weightless V-norm (`RMSNorm(head_dim, has_weight=False)`).

## Changes (`csrc/kernels/rope/rope_common.h`, `fused_qk_norm_rope_cache_quant.cu`, headers, pybind, python op)

- **head_dim 512**: add `case 512` to the `fused_rope_rms_set_kv` head-size switch and relax the guard. At `WARP_SIZE=32` this is `VEC_SIZE=16` (`vec_t<bf16,16>`); it compiles and runs (no register-pressure failure).
- **Weightless V-norm**: add `warp_rms_norm_no_weight_` (RMS-normalize a head with no learned gamma) and apply it to V in `fused_mrope_rms_kv_kernel` before the KV-cache write, gated on a new runtime `bool v_norm` (default false; the mrope-3D launcher is unchanged). Thread `v_norm` through the pts entrypoint, the header, the pybind `.def`, and the python wrapper.

Learned (`has_weight=true`) V-norm is out of scope here (Gemma4 is weightless); tracked as a follow-up.

## Test

`op_tests/test_fused_qk_norm_rope_cache_quant.py::test_fused_qk_norm_rope_cache_pts_v_norm` — exercises the pts op with `v_norm` on/off at head_dim 256 and 512; asserts the cached V is weightless RMS-normalized when `v_norm=True` and raw otherwise. **Passes** locally.

Consumed by vLLM PR vllm-project/vllm#53918 (enables `fuse_qk_norm_rope_kvcache` on Gemma4; gsm8k parity 0.7293 vs 0.726 baseline).

## Perf

End-to-end on `RedHatAI/gemma-4-31B-it-FP8-block` (ROCM_AITER_UNIFIED_ATTN, fp8 KV, input 1024 / output 512, 2 seeds; fused vs the same `splitting_ops:[]` config with the fusion off):

| conc | fused tok/s | no-fuse tok/s | fusion |
|---|---|---|---|
| 1  | 51.4  | 48.4  | **+6.2%** |
| 8  | 349.9 | 320.8 | **+9.0%** |
| 32 | 877.0 | 833.2 | **+5.3%** |

A microbenchmark of this op vs the separate ops it replaces (rms_norm x3 + reshape_and_cache, RoPE excluded) shows the fused op is ~2x cheaper per call at head_dim 256 and 512.

AI assistance (Claude) was used; a human is reviewing every line and running the tests before this leaves draft.


op_tests/test_fused_qk_norm_rope_cache_quant.py
```diff
@@ -3454,3 +3454,55 @@ def run(layout: str):
...
+    kw = torch.randn(head_size, dtype=dtype, device=dev)
+    cos_sin = torch.randn(8, head_size, dtype=dtype, device=dev)
+    positions = torch.zeros(nt, dtype=torch.int64, device=dev)
+    q_out = torch.empty(nt, hq * head_size, dtype=dtype, device=dev)
+    # NHD contiguous cache: [num_blocks, block_size, num_heads, head_size]
+    k_cache = torch.zeros(1, block_size, hk, head_size, dtype=dtype, device=dev)
+    v_cache = torch.zeros(1, block_size, hv, head_size, dtype=dtype, device=dev)
+    slot_mapping = torch.zeros(nt, dtype=torch.int64, device=dev)
+    k_scale = torch.ones(1, dtype=torch.float32, device=dev)
+    v_scale = torch.ones(1, dtype=torch.float32, device=dev)
```
yzhou103: Test coverage: v_norm=True with v_scale != 1.0 is not covered. The current test uses v_scale=ones(1), so it doesn't verify that norm-then-quantize ordering is correct when the scale actually changes the magnitude. Worth adding a case with a non-trivial v_scale (e.g. 0.5) to catch any norm/scale misordering.

op_tests/test_fused_qk_norm_rope_cache_quant.py
```diff
@@ -3454,3 +3454,55 @@ def run(layout: str):
         "pts_quant_shuffle_block_layout parity summary (markdown):\n%s",
         df.to_markdown(index=False),
     )
+
+
+def _run_pts_v_norm_case(head_size: int, v_norm: bool):
```
yzhou103: Test format doesn't follow the file's convention: missing @benchmark() decorator, not wired into the __main__ block, and no markdown summary table. The existing tests in this file all use @benchmark() + __main__ loop + df.to_markdown(). Please align with that pattern so the test runs consistently with python op_tests/tquant.py .  there is a skill in .claude about aiter-op test

csrc/kernels/rope/rope_common.h
```diff
@@ -8336,9 +8365,10 @@ void fused_rope_rms_set_kv(const T* qkv,
...
                            int64_t v_token_stride   = 0,
-                           int64_t v_head_stride    = 0)
+                           int64_t v_head_stride    = 0,
+                           bool v_norm              = false)
 {
-    AITER_CHECK(head_size == 64 || head_size == 128 || head_size == 256);
+    AITER_CHECK(head_size == 64 || head_size == 128 || head_size == 256 || head_size == 512);
     // po2_div/po2_mod in the paged and shuffle-layout address math assume these are
     // powers of two. Every vLLM block_size and every x = 16 / sizeof(elem) satisfies
     // it, but check rather than corrupt the cache silently if that ever changes.
```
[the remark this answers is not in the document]
yzhou103: please check the copilot review

op_tests/test_fused_qk_norm_rope_cache_quant.py
mpashkovskii: aligned

op_tests/test_fused_qk_norm_rope_cache_quant.py
mpashkovskii: added the case

csrc/kernels/rope/rope_common.h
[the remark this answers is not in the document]
mpashkovskii: added the check

yzhou103 (approved):
LGTM

PR #5035, opened 2026-08-26
[Triton/Attention] Clamp unified_attention 3D config to the LDS budget on gfx1201

### LINKED ISSUE #4329: [Issue]: `select_3d_config` picks a 3D unified-attention config that exceeds LDS on non-gfx1250 Wave32 (Required 65792 > 65536)

lobo235:
Fixes #4329.

## Summary

`select_3d_config` chooses the 3D split-KV tiling without any check that the resulting shared-memory (LDS) request fits the target device. On gfx1201 (Radeon AI PRO R9700), the Triton stack honors the requested pipeline depth and double-buffers **both** the K and V tiles, so the shipped config (TILE_SIZE=64, head_size=128, bf16 KV) requests 65792-73728 B, over the 64 KB limit, and Triton aborts at kernel launch:

```
triton.runtime.errors.OutOfResources: out of resource: shared memory
    Required: 65792, Hardware limit: 65536
```

Downstream this is a hard `vllm serve` startup failure for head_size=128 models on the AITER unified-attention backend (vllm-project/vllm#48723).

## Fix

A budget check in `select_3d_config`, scoped to gfx1201:

- The shared request is modeled as `2 (K, V) * TILE_SIZE * head_size_padded * kv_elem_bytes * num_stages + block_m * head_size_padded * q_elem_bytes`. The formula reproduces all three reported footprints (65792 / 69632 / 73728 B, measured via `triton.compile(...).metadata.shared`) exactly.
- When the budget is exceeded: drop the pipeline depth 2 -> 1 first (halves the K/V term and keeps the tiling the segment/gather sizing depends on), then shrink TILE_SIZE 64 -> 32 -> 16 where the tiling is not coupled to the KV block layout (shuffled cache / gather mode). If still infeasible, raise a clear `ValueError` instead of letting Triton reject the config at launch.
- The device budget comes from the existing `get_shared_memory_per_block` helper (live query with a per-gfx fallback).

## Why scoped to gfx1201

- **gfx1250** has 320 KB of LDS; the config never overflows there.
- **gfx1151** (Strix Halo): the issue's measurements show the same request resolves to 33024 B on that stack, i.e. it fits today. Clamping there would change a working config. Notably, the stage-1 result of this clamp is exactly that 33024 B footprint. If a future stack change pushes gfx1151 over the same cliff, widening the scope is a one-line change.
- **CDNA** is untouched — no behavior change for any shipping CDNA config.

## Notes

- Makes the vllm-side stopgap (vllm-project/vllm#49264) unnecessary for gfx1201.
- Related unmerged candidates: #4385 (aiter, different shape), vllm-project/vllm#49264 (downstream stopgap).
- No kernel changes; pure config selection. The only cost is stages 2 -> 1 for configs that previously failed to launch at all.

## Test

New `op_tests/triton_tests/attention/test_unified_attention_lds_budget.py`:

- footprint model reproduces the three reported byte counts (runs on any arch)
- gfx1201: overflow configs are clamped inside the budget; a fitting fp8-KV config is left untouched (no over-clamping)
- gfx1151: config unchanged (scope guard)

Verified on gfx1201 (R9700): 3 passed, 1 skipped (gfx1151 guard); with the clamp disabled the regression test fails with `assert 65792 <= 65536`.

### DIFF
--- a/aiter/ops/triton/attention/unified_attention.py
+++ b/aiter/ops/triton/attention/unified_attention.py
@@ -34,6 +34,7 @@ except:  # noqa: E722
     _reduce_segments_gluon = None
 
 from aiter.ops.triton._triton_kernels.flash_attn_triton_amd.utils import get_arch
+from aiter.ops.flydsl.utils import get_shared_memory_per_block
 from aiter.ops.triton.utils._triton import arch_info
 from aiter.ops.triton.utils.types import e4m3_dtype
 
@@ -156,6 +157,29 @@ def select_2d_config(
     }
 
 
+def _unified_3d_lds_footprint(
+    tile_size: int,
+    head_size_padded: int,
+    kv_elem_bytes: int,
+    q_elem_bytes: int,
+    block_m: int,
+    num_stages: int,
+) -> int:
+    """Shared memory (LDS) requested by kernel_unified_attention_3d on the gfx1201 stack.
+
+    The gfx1201 Triton stack honors the requested pipeline depth and double-buffers
+    both the K and V tiles: ``2 (K, V) * tile * head * kv_elem_bytes * num_stages``,
+    plus the Q tile (``block_m * head * q_elem_bytes``). At TILE_SIZE=64 / head_size=128 /
+    bf16 KV that is 65792-73728 B, over the 64 KB limit (issue #4329, downstream
+    vllm-project/vllm#48723). The formula reproduces all three reported footprints
+    exactly (measured via ``triton.compile(...).metadata.shared``).
+    """
+    return (
+        2 * tile_size * head_size_padded * kv_elem_bytes * num_stages
+        + block_m * head_size_padded * q_elem_bytes
+    )
+
+
 def select_3d_config(
     head_size,
     block_size,
@@ -167,6 +191,7 @@ def select_3d_config(
     shuffled_kv_cache: bool = False,
     NUM_BLOCKS_GATHER_PER_TILE: int = 1,
     SLIDING_WINDOW: int | None = None,
+    block_m: int = 16,
 ):
     arch = get_arch()
     reduce_num_warps = 2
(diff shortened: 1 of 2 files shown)


lobo235:
## Measured impact on gfx1201 (Radeon AI PRO R9700)

Beyond fixing the startup crash, this clamp unlocks the AITER unified attention backend at long
context on gfx1201. Measured on the same node (Qwen3.8-27B FP8, TP=2, vLLM v0.28.0rc2), engine
ITL (inter-token latency, lower is better):

| config | decode-short (1k ctx) | 68k context |
|---|---|---|
| Triton attention + MTP k=3 (previous default) | 65.1 ms | 124.8 ms |
| **AITER unified attention + this clamp, MTP off** | **46.0 ms** | **49.8 ms** |

That is **1.42x at 1k and 2.5x at 68k context** from the attention backend swap alone — the
backend was previously unusable on gfx1201 because `select_3d_config` requested 65792 B of shared
memory (> 64 KB limit), crashing at CUDA-graph warmup:

```
triton.runtime.errors.OutOfResources: out of resource: shared memory,
    Required: 65792, Hardware limit: 65536
```

With the clamp, the config fits (stages 2 -> 1, 33024 B) and the engine boots and serves.
Note: with the AITER backend, MTP speculative decoding becomes a net loss (draft overhead > 
acceptance gain), so the recommended gfx1201 config is AITER unified attention without MTP.

Verification: engine-level benchmark above + unit tests in this PR (3 passed on gfx1201, 1 gfx1151
skip; red-green: clamp disabled reproduces `assert 65792 <= 65536`).

PR #5045, opened 2026-08-27
[FlyDSL] Retune GLM5.2 mxfp4 MoE and fix a scale-view cache leak

XiaobingSuper:
## Summary

Retunes all 64 GLM5.2 mxfp4 MoE shapes (`model_dim=6144`, `inter_dim` 256/512/1024/2048, `expert=257`, `topk=9`, `token` 1..32768) for gfx950, and fixes two bugs that the tuning runs exposed.

27 of the 64 shapes now select the coupled `flydsl_mxmoe_g1/g2_a4w4_*` port; the previous config reached it on only 5 rows.

## Performance

Measured through the production `fused_moe` path, not by comparing `us` columns across two tuned CSVs. Each shape was built and timed identically under both configs (`AITER_CONFIG_FMOE` pointed at the old or the new file), with the same shape pinned to the same GPU in both phases, `num_warmup=5, num_iters=30`.

| | |
|---|---|
| median | **+6.7%** |
| mean | **+8.6%** |
| faster by >1% | 44 / 64 |
| slower by >1% | 6 / 64 (worst -3.5%) |
| flat (±1%) | 14 / 64 |

By `inter_dim` (median): 256 → +7.4%, 512 → +0.5%, 1024 → +9.8%, 2048 → +3.4%

Six shapes (`2048/256`, `4/1024`, and `2`/`16`/`64`/`128` at `2048`) keep their existing main entries rather than the retuned ones, at review request; their gains are counted as 0 above.

By token bucket (median): ≤64 → **+11.4%** (max +33.4%), 65–1024 → +3.3%, >1024 → +5.9%

Largest gains: `16/256` +33.4%, `8/256` +32.2%, `2/1024` +31.5%, `1/1024` +30.7%, `1/2048` +30.1%, `32768/1024` +24.2%.

Regressions, all small: `256/256` -1.1%, `256/512` -1.4%, `8192/256` -1.6%, `1/512` -2.3%, `32768/256` -3.3%, `16384/256` -3.5%.

## Bug 1 — `_mxfp4_scale_u8` leaks device memory (`aiter/fused_moe.py`)

The helper was wrapped in `@functools.lru_cache(maxsize=2048)`. Its body is a bare `.view(torch.uint8)`, so the memo buys nothing — but tensors hash by identity, so every per-call intermediate scale is a guaranteed miss and is then **pinned in the cache**. Four of the call sites pass freshly allocated per-call tensors (`inter_states`, `a2_scale`, `w2`, `w2_scale`).

A tuning sweep leaked ~0.75 GiB of *live* (not merely reserved) memory per timed iteration and eventually exhausted a 288 GiB card. Confirmed by walking `gc.get_referrers` back to `aiter.fused_moe._mxfp4_scale_u8 CacheInfo(hits=46, misses=74, maxsize=2048, currsize=74)`. Because the memory is live, allocator settings such as `expandable_segments:True` cannot help.

Fix: drop the decorator. The retry sweep that previously OOM'd then completed 7/7 shapes with zero OOMs.

The other `lru_cache(maxsize=2048)` sites in the file (`get_inter_dim`, `get_block_size_M`, `use_nt`, `get_ksplit`, `get_2stage_cfgs`) all key on ints/shapes and are left alone.

## Bug 2 — per-row `.item()` sync in the tuner cosine check (`aiter/ops/flydsl/mxfp4_v2_tune_utils.py`)

`v2_stage1_dequant_cosine_err` looped over sorted rows one at a time, costing one device sync per row — roughly 295k syncs per timed candidate at `token=32768, topk=9`. It is now batched in chunks of 32768 rows, which bounds the int64 gather inside `mxfp4_to_f32` while preserving the equal-weight-per-row average.

## Bug 3 — `_f4out` rows get no AOT cache entry (`aiter/aot/flydsl/mxfp4_moe.py`)

A `flydsl_mxmoe_g2_a4w4_*_f4out` kernel only takes the mxfp4-out path when both gates are open: `AITER_MXFP4_INTERMEDIATE` (off by default) and the runtime shape check in `fused_moe` (`D_HIDDEN == 7168`). Otherwise `_f4out` is stripped from the name and the plain kernel launches.

The AOT generator skipped those rows entirely, so it never pre-compiled the kernel that actually launches. GLM5 is `D_HIDDEN=6144`, so every `_f4out` row there falls back — and the config only worked because the `32768/512` row happened to name the plain kernel and seed that cache entry for `16384/512` as well. Retuning `32768/512` to `_f4out` removed the last seed, and CI hit `FLYDSL_RUNTIME_RUN_ONLY=1 but no usable AOT cache for launch_gemm2`.

Fix: always emit the non-f4out fallback job, plus the f4out one when `AITER_MXFP4_INTERMEDIATE` is set. Reproduced and verified on gfx950 with a two-row CSV and a fresh `FLYDSL_RUNTIME_CACHE_DIR`: before, `Stage2 jobs: 0` and the exact CI traceback; after, `Stage2 jobs: 1` and both shapes pass under `FLYDSL_RUNTIME_RUN_ONLY=1`.

## Test plan

- Full 64-shape retune completed on gfx950, one shape per GPU (GPUs 4–7), cosine checks passing.
- The A/B above re-runs every shape through `fused_moe` under both configs.
- Smoke-tested the tuner with the in-repo fix and no runtime patching: 0 OOM, 0 tracebacks, cosine passed.
- `black --check` clean on both changed Python files.


XiaobingSuper:
Rebased onto current main to clear the conflict.

The conflict was with #5006 (fix glm5 regression), which hand-patched 4 rows (`inter_dim=512`, token 16/32/64/128) to `block_m=16` + `flydsl_mxmoe_g1_a4w4_16x256x256_f16in_nt`. This retune converges on the same choice for all 4 independently, and additionally re-measures stage 2 for them (`token=16` → `..._atomic_sbm16`, `token=64` → `..._t16x128x128_atomic_nt_sbm16`, `token=128` → `..._atomic_persist_nt_sbm16`), so the `us` columns on those rows are now measured rather than carried over.

Also restored the CSV to main's column order and kept the trailing `_tag` column, so the diff is values only and the schema is unchanged.

coderfeli (dismissed):

coderfeli (approved):

PR #4815, opened 2026-08-18
[Config] Add Qwen3.6 35B-A3B FMoE configs for gfx1201

keneoneth:
## Motivation

Add Qwen3.6 35B-A3B FMoE tuning config files for gfx1201/RDNA4.

This is part of the Qwen3.6 gfx1201 AITER MoE enablement stack:

- Issue: <https://github.com/ROCm/aiter/issues/4492>
- Dependency: <https://github.com/ROCm/aiter/pull/4385>
- Dependency PR1: gfx1201 BF16 G1U1 small-token-count direct HIP path, <https://github.com/ROCm/aiter/pull/4740>
- Follow-up PR3: gfx1201 BF16 G1U1 selected large-M Triton MoE path, <https://github.com/ROCm/aiter/pull/5059>

## Technical Details

This PR adds model-specific FMoE tuned and untuned CSVs for Qwen3.6 35B-A3B on gfx1201:

- `qwen3_6_35b_a3b_bf16_tuned_fmoe.csv`
- `qwen3_6_35b_a3b_bf16_untuned_fmoe.csv`
- `qwen3_6_35b_a3b_fp8_tuned_fmoe.csv`
- `qwen3_6_35b_a3b_fp8_untuned_fmoe.csv`

The configs cover Qwen3.6 MoE shapes with:

- `model_dim = 2048`
- `num_experts = 256`
- `topk = 8`
- `activation = Silu`
- `use_g1u1 = 1`
- BF16 unquantized FMoE rows
- FP8 block-quantized FMoE rows

This PR only adds tuning/config data. It does not add new kernels.

The tuned rows are recorded for the measured gfx1201 `cu_num=32` target. The previously duplicated `cu_num=64` rows were removed so the CSV does not imply separate measurements.

The FP8 untuned file lists 24 requested shapes. The tuned file currently contains 22 valid rows; `M=1024, inter_dim=128` and `M=2048, inter_dim=256` did not produce valid candidates in local retuning and are left for follow-up kernel/tuner work.

The BF16 `M in {1,2,4,8,16}` decode tier is handled by the direct small-M kernel in PR1. These config rows are for the remaining FMoE/CK-backed Qwen3.6 shapes where the generic path is still appropriate.

## Test Plan

Validate the CSV files and build-time model-config/prebuild enumeration behavior:

```bash
python3 -m unittest op_tests.tuning_tests.test_csv_validation -v
python3 -m unittest op_tests.tuning_tests.test_config_shape_collision -v
```

## Test Result

[...]

PR #4740, opened 2026-08-13
[HIP] [JIT] Fix/gfx1201 bf16 g1u1 small m moe

keneoneth:
## Motivation

Enable Qwen3.6 35B-A3B BF16 AITER MoE serving on gfx1201/RDNA4 systems such as Navi48/R9700S.

Currently, setting `VLLM_ROCM_USE_AITER_MOE=1` for Qwen3.6 35B-A3B on gfx1201 can route small decode MoE shapes into AITER paths that do not have a valid gfx1201 kernel, making AITER MoE unusable for this model family. The observed serving failure is:

```text
RuntimeError: wrong! device_gemm with the specified compilation parameters does not support this GEMM problem
```

This PR adds the missing BF16 G1U1 small-token-count tier with a guarded direct HIP path for:

```text
M in {1, 2, 4, 8, 16}
inter_dim in {128, 256}
model_dim = 2048
num_experts = 256
topk = 8
```

This is part of the Qwen3.6 35B-A3B gfx1201 AITER MoE enablement stack:

- Issue: <https://github.com/ROCm/aiter/issues/4492>
- Dependency: <https://github.com/ROCm/aiter/pull/4385>
- Follow-up PR2: gfx1201 Qwen3.6 35B-A3B FMoE config data, <https://github.com/ROCm/aiter/pull/4815>
- Follow-up PR3: gfx1201 BF16 G1U1 selected large-M Triton MoE path, <https://github.com/ROCm/aiter/pull/5059>

## Technical Details

This PR adds `fmoe_g1u1_bf16_small_m`, a direct two-kernel path for the narrow unquantized BF16 G1U1 case.

Source files:

- Python dispatch: `aiter/fused_moe.py`
- Python op binding: `aiter/ops/moe_op.py`
- HIP entrypoint and kernels: `csrc/py_itfs_cu/asm_fmoe.cu`
- Focused test: `op_tests/test_moe_small_m_direct.py`

Implementation summary:

- Stage 1 computes `Silu(gate) * up` into a BF16 activation workspace.
- Stage 2 applies the down projection and top-k weighted accumulation.
- The path bypasses general MoE sorting and CK grouped GEMM for this small-token-count case only.
- The path accepts canonical W1/W2 and the standard AITER `shuffle_weight(..., layout=(16,16))` W1/W2 runtime layout.
- A narrow early Python fast path avoids the generic custom-op wrapper overhead for this exact small-M case.

The guard remains limited to:

- gfx1201 only
- BF16 activations and weights
[...]

Issue #4492, opened 2026-07-31
[Issue]: AITER fused MoE is not usable on gfx1201/R9700S: no gfx1201 dispatch entry and `fmoe_g1u1` finds no suitable kernel


keneoneth:
### Problem Description

AITER fused MoE does not appear usable on gfx1201 machines.

This was found while trying to enable AITER MoE for Qwen3.6 MoE models through vLLM and enabling AITER MoE fails before the vLLM server starts.

The following AITER MoE environment variables were set:

```bash
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
```

The issue appears to have two layers:

1. vLLM reports AITER fused MoE as unsupported on gfx1201 because `rocm_aiter_ops.is_fused_moe_enabled()` returns false.
2. Direct AITER fused MoE testing on gfx1201 also fails. After resolving the CK header/build dependency by testing with AITER main and the CK submodule initialized, `fmoe_g1u1` still cannot find a suitable kernel for representative MoE shapes.

The goal of this issue is to clarify whether gfx1201 AITER fused MoE is expected to be supported for now.

### Operating System

Ubuntu 24.04 container on Linux host

### CPU

AMD EPYC 9554 64-Core Processor

### GPU

AMD Radeon AI Pro R9700S / gfx1201

### ROCm Version

ROCm 7.14.0rc3

### Code Sources

vLLM commit: c8db00b16cc188b46b7b9517a5836a0da4aa8c3e
vLLM ref: main
vLLM version: 0.26.0.dev0+main

AITER ref: refs/pull/4385/head
AITER commit: 81b2b29f7c8064899df6fe5b15a246261a7220a7
AITER PR #4385 LDS guard present: True
AITER install mode: AITER_TRITON_ONLY=1 AITER_USE_SYSTEM_TRITON=1

Local AITER main test:
local AITER HEAD: 6042519a2af604e556b6de01f9dfd9966b695476
local AITER base main: 152cf1b99
CK submodule: f33252cebe5a52362ec1ee12c124dde7800dda3a

### 1. vLLM failure with AITER MoE enabled

Set:

```bash
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
```

Start a Qwen3.6 MoE model through vLLM. Example BF16 model:

```bash
vllm serve /mnt/nas_share/models/Qwen/Qwen3.6-35B-A3B \
  --dtype auto \
[...]

PR #5059, opened 2026-08-27
[CI] [JIT] Add gfx1201 BF16 G1U1 selected large-M Triton MoE path

keneoneth:
## Motivation

Enable the remaining Qwen3.6 35B-A3B BF16 AITER MoE large-M serving shapes on gfx1201/RDNA4 systems such as Navi48/R9700S when `VLLM_ROCM_USE_AITER_MOE=1`.

After the small-token-count direct HIP path and the Qwen3.6 35B-A3B FMoE config data are in place, there is still a band of larger BF16 G1U1 rows where the standard gfx1201 path is valid but not consistently faster than the vLLM Triton MoE baseline. This PR adds a guarded gfx1201 Triton route only for exact large-M rows that were measured to be stable and faster in kernel-level validation.

This is part of the Qwen3.6 35B-A3B gfx1201 AITER MoE enablement stack:

- Issue: https://github.com/ROCm/aiter/issues/4492
- Dependency: https://github.com/ROCm/aiter/pull/4385
- Dependency PR1: gfx1201 BF16 G1U1 small-token-count direct HIP path, https://github.com/ROCm/aiter/pull/4740
- Dependency PR2: gfx1201 Qwen3.6 35B-A3B FMoE config data, https://github.com/ROCm/aiter/pull/4815

## Technical Details

This PR adds a guarded gfx1201 BF16 G1U1 Triton route for selected Qwen3.6 35B-A3B large-M rows.

Source files:

- Triton dispatch, exact config lookup, and execution helper: `aiter/fused_moe.py`
- Triton FMoE config registration: `aiter/jit/core.py`
- Base Triton FMoE config files:
  - `aiter/configs/tuned_triton_fmoe.csv`
  - `aiter/configs/untuned_triton_fmoe.csv`
- Qwen3.6 35B-A3B tuned rows: `aiter/configs/model_configs/qwen3_6_35b_a3b_bf16_tuned_triton_fmoe.csv`
- Dispatch/config tests: `op_tests/test_triton_fmoe_dispatch.py` (updated)

Implementation summary:

- The PR adds a separate `AITER_CONFIG_TRITON_FMOE` config family so Triton launch configs remain data-driven rather than hardcoded in `fused_moe.py`.
- `_load_triton_fmoe_configs()` loads the merged Triton FMoE config table.
- `_get_gfx1201_triton_fmoe_config(...)` performs a strict exact-key lookup using the current gfx, CU count, token count, model shape, dtype, activation, and quantization parameters.
- `_triton_bf16_g1u1_moe(...)` uses existing AITER Triton MoE kernels:
  - `moe_align_block_size_triton`
  - `aiter.ops.triton.moe.moe_op_silu_fused.fused_moe_silu`
  - `aiter.ops.triton.moe.moe_op.fused_moe`
  - `moe_sum`
- The Triton lookup and dispatch happen before the normal AITER `get_2stage_cfgs()` and `moe_sorting()` path.
- If there is no exact matching Triton config row, execution falls through to the existing AITER path unchanged, including the existing `get_2stage_cfgs(get_padded_M(M), ...)` lookup.
- There is no padded-M fallback, nearest-CU fallback, or range interpolation for the Triton route.
- The exact-match config structure allows additional validated rows to be added without changing the existing AITER fallback behavior.

The guard remains limited to:

- gfx1201 only
- BF16 activations and weights
- unquantized MoE
- Silu / G1U1
- `model_dim = 2048`
- `inter_dim = 128`
- `num_experts = 256`
- `topk = 8`
- `doweight_stage1 = False`
- no expert parallel mask
- no bias
- no hidden/intermediate padding
- no quant scales
- no local-token override
- `GateMode.SEPARATED`
- exact tuned CSV row only

Selected rows:

| inter_dim | M values |
|---:|---|
| 128 | `1152`, `1543`, `1728`, `1799`, `1803`, `1815`, `1837`, `1944`, `1945`, `2048` |

### Dispatch and Config Validation

The dispatch tests verify that:

- an exact selected gfx1201 row returns the expected Triton launch config;
- a neighboring unselected row does not inherit another config;
- `inter_dim=256` is not enabled;
- non-gfx1201 requests do not use this route;
- quantized, expert-parallel, biased, padded, scaled, local-token, and non-separated-gate requests fall through;
- the Triton FMoE config family resolves correctly.

### Kernel-Level Validation

Kernel-level validation compares this AITER path against the vLLM Triton MoE path using the same BF16 hidden states, weights, routing inputs, and top-k weights.

AITER path under test:

- The benchmark helper imports `aiter.fused_moe.fused_moe` as `aiter_fused_moe`.
- For an exact selected row, `aiter_fused_moe(...)` returns through `_triton_bf16_g1u1_moe(...)` before the normal AITER two-stage config/sorting path.
- The selected Triton path uses `moe_align_block_size_triton`, `fused_moe_silu`, `fused_moe`, and `moe_sum`.

Exact AITER call under test:

```python
from aiter.fused_moe import fused_moe as aiter_fused_moe

def run_aiter():
    return aiter_fused_moe(
        hidden,
        w1_aiter,
        w2_aiter,
        weights,
        ids,
        activation=aiter.ActivationType.Silu,
        quant_type=aiter.QuantType.No,
        dtype=dtypes.bf16,
    )
```

vLLM baseline path under test:

- The benchmark helper imports `vllm.model_executor.layers.fused_moe.fused_moe.fused_experts`.
- This exercises the vLLM unquantized Triton MoE path with AITER disabled.

Exact vLLM call under test:

```python
from vllm.model_executor.layers.fused_moe.activation import MoEActivation
from vllm.model_executor.layers.fused_moe.fused_moe import fused_experts

def run_vllm():
    return fused_experts(
        hidden,
        w1,
        w2,
        weights,
        ids,
        activation=MoEActivation.SILU,
        global_num_experts=expert,
    )
```

Exact latency timing used for both paths (`fn()` represents the functions described above):

```python
def bench(fn, warm: int, iters: int) -> tuple[float, float]:
    for _ in range(warm):
        fn()
    torch.cuda.synchronize()

    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)
    times = []
    for _ in range(iters):
        start.record()
        fn()
        end.record()
        torch.cuda.synchronize()
        times.append(start.elapsed_time(end) * 1000.0)
    return statistics.mean(times), statistics.median(times)
```

### Kernel-Level Validation Result

Kernel-level validation result from the Qwen3.6 35B-A3B BF16 selected large-M harness, comparing:

- vLLM Triton baseline;
- original AITER path before this change (base branch with the Qwen3.6 BF16 FMoE tuned config supplied from the PR2 artifact);
- updated AITER path after this change.

| inter_dim | M | vLLM mean us | AITER original mean us | AITER updated mean us | Original delta vs vLLM | Updated delta vs vLLM |
|---:|---:|---:|---:|---:|---:|---:|
| 128 | 1152 | 1015.051 | 1978.551 | 987.241 | -97.15% | 2.74% |
| 128 | 1543 | 1088.187 | 2471.005 | 1074.476 | -128.76% | 1.26% |
| 128 | 1728 | 1101.682 | 2716.054 | 1111.353 | -144.49% | -0.88% |
| 128 | 1799 | 1116.192 | 2819.401 | 1112.582 | -151.02% | 0.32% |
| 128 | 1803 | 1114.117 | 2820.434 | 1110.769 | -151.43% | 0.30% |
| 128 | 1815 | 1107.736 | 2827.023 | 1100.336 | -154.12% | 0.67% |
| 128 | 1837 | 1111.264 | 2840.152 | 1107.088 | -154.58% | 0.38% |
| 128 | 1944 | 1138.056 | 2984.190 | 1131.887 | -162.02% | 0.54% |
| 128 | 1945 | 1142.209 | 2998.672 | 1134.125 | -161.63% | 0.71% |
| 128 | 2048 | 1160.809 | 3136.352 | 1163.372 | -168.22% | -0.22% |

Notes:

- The original AITER measurements above were collected on the base branch with `AITER_CONFIG_FMOE` pointed at the Qwen3.6 35B-A3B BF16 tuned FMoE config artifact from the PR2 validation context, so the comparison reflects the intended dependency-backed default path rather than an unsupported no-config state before PR2's tuning updates.
- The updated AITER measurements were collected on the current PR4 head.

## Submission Checklist


zufayu:
## aiter PR #5059 — [JIT] Add gfx1201 BF16 G1U1 selected large-M Triton MoE path

**Routes ten measured Qwen3.6 35B-A3B token counts on gfx1201 to the existing Triton MoE kernels, reaching parity with the vLLM baseline that the current AITER path loses to by 2–2.7x.**

Validation (deterministic): NOT RUN — no exact-head `validation_report.json`. The index-width scan over this diff returns 0 candidates, correctly: the only products here are `M * topk` and `E * block_m`, both bounded well under 2^31.

Worth stating plainly, because it changes how the rest should be read: this PR adds no kernel. Every kernel it calls — `fused_moe_silu`, `triton_moe`, `moe_align_block_size_triton`, `moe_sum` — already exists. The 260 lines in `fused_moe.py` are a gate, a config loader and an orchestration helper. And the measured gap it closes is large: the existing AITER path runs 1979 µs against vLLM's 1015 µs at M=1152, and 3136 µs against 1161 µs at M=2048. That is a real regression to fix, not a marginal tuning win.

⚠️ **The 128-line test never executes in any CI job.** `op_tests/tuning_tests/test_triton_fmoe_dispatch.py` uses `unittest.TestCase`, which is the right framework for where it lives — but `tuning-tests.yaml` invokes a hard-coded module list across three jobs (`test_csv_validation`, `test_tuner_infra`, `test_mp_tuner_logic`, `test_a6w6_tuning` at line 96; `test_tune_pipeline` at 179; `test_run_config` at 262), and the new module is in none of them. It also cannot arrive via the standard shards, which collect `op_tests` at depth 1 only (`aiter_test.sh:30`).

More importantly, that workflow's triggers are `schedule` and `workflow_dispatch` — there is no `pull_request`. So even after the module is added to the list, this test gives no PR-time protection; it would report the next morning at the earliest. That matters here because the gate below is the whole PR. **Author must** add the module to `tuning-tests.yaml`, and it is worth asking separately whether a dispatch-gate test belongs somewhere that runs on PRs at all. [verified]

⚠️ **The config lookup matches the token count exactly, so `M = 1543` hits and `M = 1544` falls through.** `_get_gfx1201_triton_fmoe_config` builds a 14-field key including `str(M)` and resolves it with `.get(key)`, returning `None` on any miss. The shipped rows are `1152, 1543, 1728, 1799, 1803, 1815, 1837, 1944, 1945, 2048` — values that read like observed batch sizes rather than a designed sweep. The very next statement in `_fused_moe_impl` is `get_2stage_cfgs(get_padded_M(M), ...)`, i.e. the path this one sits in front of buckets M rather than matching it.

Under real serving M moves continuously with batch composition, so the fast path will fire far less often than a ten-row table suggests, and the 2–2.7x gap remains for everything in between. `cu_num` is in the key as well, so a Navi48 reporting a different CU count misses every row. **Reviewer should ask** whether bucketing or nearest-M is intended, and — given the measured gap grows monotonically with M from 97% to 168% — whether the underlying HIP path should be fixed rather than bypassed at ten points. [verified]

📝 **Two silences stack.** `@unittest.skipUnless(fused_moe_mod is not None, ...)` turns an import failure into a green skip, so even once the module is in CI a broken import reports success. And `_load_triton_fmoe_configs` is wrapped in `functools.lru_cache(maxsize=1)`, so `AITER_CONFIG_TRITON_FMOE` is read once per process and a later change to it has no effect. **Author must** make the import failure an error rather than a skip. [verified]

📝 **`.contiguous()` is called unconditionally on the hot path.** `_triton_bf16_g1u1_moe(hidden_states.contiguous(), w1.contiguous(), w2.contiguous(), ...)` runs on every forward. It is free when the tensors are already contiguous, but if the expert weights ever arrive non-contiguous this copies all 256 experts per call, which would dwarf the gain the PR is measuring. **Reviewer should ask** whether the gate should assert contiguity instead of silently repairing it. [verified]

📝 **`stage1` is allocated with `torch.empty` and its padding rows are never written.** `sorted_token_ids` is filled with the sentinel `topk_ids.numel()`, so `fused_moe_silu` masks those rows out, leaving them uninitialized in `stage1`. Whether the second GEMM or `moe_sum` ever reads them depends on how `triton_moe` walks `sorted_token_ids`, which I did not trace — so this is a question, not a finding. **Reviewer should ask** the author to confirm the padded rows are unreachable downstream. [inferred]

keneoneth:
> ## aiter PR #5059 — [JIT] Add gfx1201 BF16 G1U1 selected large-M Triton MoE path
> **Routes ten measured Qwen3.6 35B-A3B token counts on gfx1201 to the existing Triton MoE kernels, reaching parity with the vLLM baseline that the current AITER path loses to by 2–2.7x.**
> 
> Review (advisory): ⚠️ NEEDS WORK Validation (deterministic): NOT RUN — no exact-head `validation_report.json`. The index-width scan over this diff returns 0 candidates, correctly: the only products here are `M * topk` and `E * block_m`, both bounded well under 2^31.
> 
> Worth stating plainly, because it changes how the rest should be read: this PR adds no kernel. Every kernel it calls — `fused_moe_silu`, `triton_moe`, `moe_align_block_size_triton`, `moe_sum` — already exists. The 260 lines in `fused_moe.py` are a gate, a config loader and an orchestration helper. And the measured gap it closes is large: the existing AITER path runs 1979 µs against vLLM's 1015 µs at M=1152, and 3136 µs against 1161 µs at M=2048. That is a real regression to fix, not a marginal tuning win.
> 
> ⚠️ **The 128-line test never executes in any CI job.** `op_tests/tuning_tests/test_triton_fmoe_dispatch.py` uses `unittest.TestCase`, which is the right framework for where it lives — but `tuning-tests.yaml` invokes a hard-coded module list across three jobs (`test_csv_validation`, `test_tuner_infra`, `test_mp_tuner_logic`, `test_a6w6_tuning` at line 96; `test_tune_pipeline` at 179; `test_run_config` at 262), and the new module is in none of them. It also cannot arrive via the standard shards, which collect `op_tests` at depth 1 only (`aiter_test.sh:30`).
> 
> More importantly, that workflow's triggers are `schedule` and `workflow_dispatch` — there is no `pull_request`. So even after the module is added to the list, this test gives no PR-time protection; it would report the next morning at the earliest. That matters here because the gate below is the whole PR. **Author must** add the module to `tuning-tests.yaml`, and it is worth asking separately whether a dispatch-gate test belongs somewhere that runs on PRs at all. [verified]
> 
> ⚠️ **The config lookup matches the token count exactly, so `M = 1543` hits and `M = 1544` falls through.** `_get_gfx1201_triton_fmoe_config` builds a 14-field key including `str(M)` and resolves it with `.get(key)`, returning `None` on any miss. The shipped rows are `1152, 1543, 1728, 1799, 1803, 1815, 1837, 1944, 1945, 2048` — values that read like observed batch sizes rather than a designed sweep. The very next statement in `_fused_moe_impl` is `get_2stage_cfgs(get_padded_M(M), ...)`, i.e. the path this one sits in front of buckets M rather than matching it.
> 
> Under real serving M moves continuously with batch composition, so the fast path will fire far less often than a ten-row table suggests, and the 2–2.7x gap remains for everything in between. `cu_num` is in the key as well, so a Navi48 reporting a different CU count misses every row. **Reviewer should ask** whether bucketing or nearest-M is intended, and — given the measured gap grows monotonically with M from 97% to 168% — whether the underlying HIP path should be fixed rather than bypassed at ten points. [verified]
> 
> 📝 **Two silences stack.** `@unittest.skipUnless(fused_moe_mod is not None, ...)` turns an import failure into a green skip, so even once the module is in CI a broken import reports success. And `_load_triton_fmoe_configs` is wrapped in `functools.lru_cache(maxsize=1)`, so `AITER_CONFIG_TRITON_FMOE` is read once per process and a later change to it has no effect. **Author must** make the import failure an error rather than a skip. [verified]
> 
> 📝 **`.contiguous()` is called unconditionally on the hot path.** `_triton_bf16_g1u1_moe(hidden_states.contiguous(), w1.contiguous(), w2.contiguous(), ...)` runs on every forward. It is free when the tensors are already contiguous, but if the expert weights ever arrive non-contiguous this copies all 256 experts per call, which would dwarf the gain the PR is measuring. **Reviewer should ask** whether the gate should assert contiguity instead of silently repairing it. [verified]
> 
> 📝 **`stage1` is allocated with `torch.empty` and its padding rows are never written.** `sorted_token_ids` is filled with the sentinel `topk_ids.numel()`, so `fused_moe_silu` masks those rows out, leaving them uninitialized in `stage1`. Whether the second GEMM or `moe_sum` ever reads them depends on how `triton_moe` walks `sorted_token_ids`, which I did not trace — so this is a question, not a finding. **Reviewer should ask** the author to confirm the padded rows are unreachable downstream. [inferred]

Thanks for the review, this is helpful. Here's the updated fix point-by-point:

1. **Dispatch test coverage in CI**

Agreed. This was a real gap.

- We added `op_tests.tuning_tests.test_triton_fmoe_dispatch` to `.github/workflows/tuning-tests.yaml`.
- We also added an explicit PR-time unittest step for this module in `.github/workflows/aiter-test.yaml`, because the standard `op_tests` sharding only collects depth-1 files and would not naturally pick up `op_tests/tuning_tests/test_triton_fmoe_dispatch.py`.

So this test is now both:
- included in the tuning test workflow, and
- exercised in PR-time validation.

2. **Import failure was previously a green skip**

Agreed. This was also a real issue.

- We removed the class-level `@unittest.skipUnless(...)`.
- Import/setup failure in `test_triton_fmoe_dispatch.py` now hard-fails instead of silently skipping.

3. **Exact `M` lookup**

This behavior is intentional for this first step.

The current design here is an exact-match allow-list of validated gfx1201 rows, not a generic bucket/range policy. That is why the lookup key includes the exact token count and returns `None` on a miss.

We agree this is not a general solution for all nearby `M` values. For this PR, though, we want conservative behavior:
- exact validated rows take the Triton route
- everything else falls through to the existing AITER path unchanged

We do plan to revisit bucket/range policy in follow-up work, but only after we have broader measurements and test coverage to justify it.

4. **Unconditional `.contiguous()` on the hot path**

Agreed for `w1` / `w2`.

We tightened this:
- `w1` and `w2` contiguity is now part of Triton-route eligibility
- if either is non-contiguous, the Triton selector declines the route and we fall back to the existing path
- we no longer silently copy `w1` / `w2` inside the Triton fast path

We kept `hidden_states.contiguous()` as a bounded per-request activation repair, but removed the silent weight-copy behavior that would be problematic for a performance path.

5. **`stage1 = torch.empty(...)` / padded-row concern**

We reviewed this path.

In the Triton helper, `stage1` is allocated as:

```python
stage1 = torch.empty((M * topk, inter_dim), ...)
```

Here `M` is the token count for this MoE call, and `topk` is the number of routed experts per token. So `stage1` holds exactly one row per real token-expert route, i.e. exactly `M * topk` payload rows.

The padding only exists in the routing metadata:

```python
sorted_token_ids
sorted_expert_ids
```

`sorted_token_ids` is padded up to the block-aligned routing length and the padded entries are filled with the sentinel `topk_ids.numel()` (which is also `M * topk`).

So the distinction is:

- `stage1`: exact payload buffer for the real `M * topk` routed pairs
- `sorted_token_ids` / `sorted_expert_ids`: padded metadata used for block alignment

In other words, there are no extra padded payload rows allocated in `stage1` itself. The padded portion is only in the metadata arrays, not in the `stage1` storage.

.github/workflows/aiter-test.yaml
```diff
@@ -937,6 +937,16 @@ jobs:
           aiter_test \
           bash -c "SHARD_TOTAL=${{ matrix.shard_total }} SHARD_IDX=${{ matrix.shard_idx }} ./.github/scripts/aiter_test.sh"
 
+      - name: PR dispatch gate test
```
gyohuangxin: Please don't add this Triton FMoE-specific unittest as a hardcoded step in the general aiter-test workflow?

gyohuangxin (changes requested):

.github/workflows/aiter-test.yaml
keneoneth: Thanks for the review. I removed the Triton FMoE-specific unittest step from the general aiter-test workflow. And the dispatch test has been moved into the standard top-level Aiter test suite as op_tests/test_triton_fmoe_dispatch.py.

PR #5060, opened 2026-08-27
[Config] [Tune] Add GLM-5.3 BF16 GEMM configs for gfx950

andyluo7:
## Motivation

GLM-5.3-Flash TP8/EP1 decode on gfx950 exercises small BF16 GEMMs across more than the original batch-1 and batch-32 paths. This update extends measured coverage to the decode graph batch-size grid used by the serving configuration:

`M={1,2,4,8,16,32,64,128}`

## Technical details

- Add `glm53_bf16_tuned_gemm.csv` under `aiter/configs/model_configs/` so it is discovered by the existing BF16 config merge.
- Keep 29 model-specific rows, distributed as M1: 3, M2: 4, M4: 4, M8: 2, M16: 7, M32: 3, M64: 4, and M128: 2.
- Scope every row to `gfx950`, `cu_num=256`, BF16 input/output, no bias, and no preshuffle.
- Retain only candidates with at least 3% measured improvement that also change the selected dispatch.
- Remove four original rows and five newly tuned candidates that select the same fallback backend/solution and therefore do not change runtime dispatch.
- Reuse the existing global `q3vl_bf16_tuned_gemm.csv` rows for the overlapping `N=128, K=4096` family instead of duplicating runtime keys. For M=128, the existing global row is faster than the new candidate (6.4639 us versus 6.7474 us).

## Validation

Source:

- AITER base: `0200ada1aa4cc7fe47aa14abd4677b35105b1553`
- PR head: `ae578dd357a46bd5adde4710410cf99274bd7aeb`
- Hardware: 8x AMD Instinct MI355X (`gfx950`, 256 CUs per GPU)

The added-M tuning sweep covered 48 candidates: six added M values times eight GLM shape families. It used the `asm`, `flydsl`, `triton`, `torch`, and `skinny` backends, with 5 warmups and 20 timed iterations. All 48 candidates tuned successfully; 33 met the 3% production pre/post threshold and all post-tune correctness checks passed within the configured tolerance.

After removing dispatch no-ops and globally covered keys, 23 new rows remain. Their measured improvements range from 4.01% to 32.07%. Together with the six meaningful original M1/M32 rows, the final table contains 29 rows.

Exact-head checks on MI355X:

- 30 CSV validation and runtime collision tests passed.
- The production `csrc/gemm_a16w16/gemm_tuner.py --run_config` path executed all 29 fallback cases successfully with zero retries.
- The same production path executed all 29 tuned rows successfully with zero retries.
- `git diff --check` passed.

Full-grid SGLang integration on the same 8x MI355X node (Slurm job `1882`):

- TP8/EP1 started with full decode graph mode and explicitly captured `bs=[1,2,4,8,16,32,64,128]` on all eight TP ranks in 35.60-35.86 seconds.
- The server-lifetime route audit recorded 648 tuned BF16 selections over 61 unique padded `(M,N,K)` keys. All eight requested decode M buckets were exercised.
- 28 of the 29 final model-specific BF16 rows were selected. The remaining `(16,32,4096)` BF16 row was operator-validated but was not invoked in this serving configuration because the live `N=32,K=4096` projection is FP32.
- A concurrency-128 random 1K-input/1K-output run completed 128/128 requests, generated 131,072 tokens in 19.81 seconds, and reached 6,615.64 output tok/s.
- All 128 requests had exactly 1,024 input and 1,024 output tokens and empty error strings.
- The server log had no traceback, HIP error, OOM, or engine fault; Slurm stderr was empty; all validation exit files were zero.

For isolation, current-main AITER source was used for the static/collision and production GEMM checks. The end-to-end serving run used the previously validated packaged AITER runtime with the exact current PR config directory mounted over it, so this serving result validates the configuration change without claiming unrelated current-main FP8 MoE JIT compatibility in that older SGLang image.

The original M1/M32 table was also validated end to end with SGLang GLM-5.3-Flash TP8/EP1, FP8 KV cache, AITER MoE, TileLang DSA, full decode graphs, disabled radix cache, and 1,024 input/output tokens:

| Case | Requests | Total throughput | Median interactivity |
|---|---:|---:|---:|
| concurrency 1 | 10/10 | 287.85 tok/s | 145.61 tok/s/user |
| concurrency 32 | 320/320 | 5,731.36 tok/s | 93.72 tok/s/user |

Every request produced exactly 1,024 output tokens with an empty error string, and all client/supervisor exits were zero.

## Submission checklist

- [x] Rebased onto current AITER `main`.
- [x] Covered the serving decode grid through M=128.
- [x] Removed default-equivalent rows.
- [x] Checked for runtime-key collisions and reused existing global rows.
- [x] Ran production operator validation on MI355X.
- [x] Limited the PR to model-specific gfx950 tuning data.


andyluo7:
Downstream integration and reproducibility references: SGLang engine support is in sgl-project/sglang#36607, and the exact 8x MI355X TP8/EP1 recipe plus 1K/1K c1/c32 measurements are in sgl-project/sglang#36732.

valarLip:
how about other M..

stefanskiasan:
**Measured on 8× MI350X (gfx950): +10.9 % generated throughput from this CSV alone, single-variable.**

Mounted `glm53_bf16_tuned_gemm.csv` over `aiter/configs/model_configs/` in an otherwise unchanged container and re-ran the same benchmark. AITER picked the file up (8 references in the startup log).

| | generated |
|---|---:|
| without the CSV | 2,100 tok/s |
| **with the CSV** | **2,328 tok/s** |

Setup: `zai-org/GLM-5.3-Flash`, vLLM `ZJY0516/vllm@glm-release` (the GLM-5.3 branch plus its ROCm implementation), TP8/EP1, BF16 KV, Triton MoE, decode CUDA graphs on, `--max-model-len 8192 --max-num-seqs 64 --gpu-memory-utilization 0.85`, ROCm 7.2.3, Triton 3.7.1. Workload: 8 fixed short English prompts, `temperature=0.0`, `max_tokens=1024`, N=64 concurrent, generated = `completion_tokens / wall_clock`. Same container, same image, only the mount added.

Two notes that may be useful:

- Raising `--max-num-seqs` compounds with this: with the CSV in place, 64 → 256 → 512 gives 2,328 → 2,766 → 3,034 tok/s. So the M={1,32} rows help the low-batch decode path exactly as described, and the tuned shapes do not become irrelevant once the batch grows.
- We also measured **fp8 KV against BF16 KV** on this model, since your MI355X recipe (sgl-project/sglang#36732) selects FP8 E4M3. On our box fp8 was slightly *slower* (2,717 vs 2,766 tok/s at N=256), gave essentially no KV-capacity gain (31.5M vs 32.3M tokens — only 11 of 45 layers are sparse MLA, the 34 KDA layers carry constant-size state that fp8 does not shrink), and cost tool-call accuracy above ~50k context (100k: 1/4 correct with fp8 vs 3/4 with BF16, `tool_choice` unset). At 1k/1k that difference would not show, which may be why the recipe benchmark did not surface it. Worth a look before fp8 KV is recommended for long-context serving.

Happy to re-run any of this or test further tuned shapes — the sweep takes about 25 minutes per configuration here.

andyluo7:
Good point. I expanded the branch to cover the full decode grid we use for this model: `M={1,2,4,8,16,32,64,128}`.

I tuned 48 additional candidates on 8x MI355X (six added M values times eight GLM shape families). All tuned successfully; 33 met the >=3% production pre/post threshold. I then removed rows that do not actually change dispatch and avoided duplicating keys already supplied by the global Q3VL BF16 table.

The final model-specific table has 29 meaningful rows:

- M1: 3
- M2: 4
- M4: 4
- M8: 2
- M16: 7
- M32: 3
- M64: 4
- M128: 2

Accounting:

- Removed four original M1/M32 rows and five new rows that select the same fallback backend/solution.
- Reused five accepted `N=128, K=4096` rows from `q3vl_bf16_tuned_gemm.csv` instead of adding duplicate runtime keys.
- At M=128, the existing global row is faster than the new candidate: 6.4639 us versus 6.7474 us.

Validation on the rebased head (`ae578dd35`): 30 CSV/collision tests pass, and both fallback and tuned production `--run_config` executions pass all 29 rows with zero retries. The 23 newly retained rows measured 4.01%-32.07% improvement.

andyluo7:
Follow-up full-grid validation is complete on 8x MI355X (job `1882`).

- Full TP8/EP1 decode graphs captured `bs=[1,2,4,8,16,32,64,128]` on all eight ranks in 35.60-35.86 seconds.
- The server-lifetime audit recorded 648 tuned BF16 selections over 61 unique padded `(M,N,K)` keys, with every requested decode M bucket exercised.
- 28/29 model-specific BF16 rows were selected. The remaining `(16,32,4096)` row passed operator validation but was not invoked because this live SGLang path emits the `N=32,K=4096` projection as FP32.
- The c128 1K-input/1K-output run completed 128/128 requests, generated 131,072 output tokens in 19.81 seconds, and reached 6,615.64 output tok/s.
- All request error strings were empty; the server had no traceback, HIP error, OOM, or engine fault; Slurm stderr was empty.

Current-main AITER source was used for the 30 static/collision tests and both 29-row production `--run_config` passes. The serving run intentionally overlaid the exact current config onto the previously validated packaged AITER runtime, isolating this CSV change from unrelated FP8 MoE JIT compatibility in the older SGLang image.

valarLip (approved):

PR #5066, opened 2026-08-28
[HIP] [DCP] Enable fused indexer QK preparation

ZhiweiYan-96:
## Summary
Enable the fused indexer QK preparation kernel for DCP. On non-owner DCP ranks, `slot_mapping` is negative, but those ranks must still produce valid `q_out` and `weights_out` for distributed index scoring. The previous early return skipped the entire kernel and left these outputs uninitialized.

## Changes
- Add an optional `compute_all_q_rope` argument, following the approach used by #4342.
- Compute Q RoPE, FP8 quantization, and `weights_out` on every DCP rank.
- Keep K normalization, quantization, and cache writes restricted to valid owner slots.
- Clamp stale padding positions before indexing the RoPE cache.
- Preserve the existing non-DCP behavior by default.
- Add coverage for mixed valid/invalid slots, stale positions, and the default non-DCP path.

### DIFF
--- a/csrc/kernels/cache_kernels.cu
+++ b/csrc/kernels/cache_kernels.cu
@@ -1439,7 +1439,9 @@ __global__ void indexer_qk_rope_quant_and_cache_kernel(
     const float weights_scale,
     const bool use_ue8m0,
     const bool preshuffle,
-    const bool is_neox)
+    const bool is_neox,
+    const int max_position,
+    const bool compute_all_q_rope)
 {
     static_assert(HEAD_DIM == 128, "Indexer fused qk cache currently supports head_dim=128");
     static_assert(ROPE_DIM == 64, "Indexer fused qk cache currently supports rope_dim=64");
@@ -1451,10 +1453,10 @@ __global__ void indexer_qk_rope_quant_and_cache_kernel(
         return;
 
     const int64_t slot_idx = slot_mapping[token_idx];
-    if(slot_idx < 0)
+    if(!compute_all_q_rope && slot_idx < 0)
         return;
-
-    const int64_t pos = positions[token_idx];
+    int64_t pos = positions[token_idx];
+    pos = pos < 0 ? 0 : (pos >= max_position ? max_position - 1 : pos);
     const scalar_t* cos_ptr = cos_cache + pos * cos_stride0;
     const scalar_t* sin_ptr = sin_cache + pos * sin_stride0;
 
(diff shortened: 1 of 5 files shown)


csrc/kernels/cache_kernels.cu
```diff
@@ -1451,10 +1453,10 @@ __global__ void indexer_qk_rope_quant_and_cache_kernel(
         return;
 
     const int64_t slot_idx = slot_mapping[token_idx];
-    if(slot_idx < 0)
+    if(!compute_all_q_rope && slot_idx < 0)
         return;
-
-    const int64_t pos = positions[token_idx];
+    int64_t pos = positions[token_idx];
+    pos = pos < 0 ? 0 : (pos >= max_position ? max_position - 1 : pos);
```
yitingw1: The clamp sits before the head_idx != 0 || slot_idx < 0 return, so it also applies to valid owner rows on the pre-existing non-DCP path.

csrc/kernels/cache_kernels.cu
```diff
@@ -4102,6 +4106,7 @@ void indexer_qk_rope_quant_and_cache(
     AITER_CHECK(cos_cache.size(0) == sin_cache.size(0) &&
                     cos_cache.size(1) == sin_cache.size(1),
                 "cos_cache and sin_cache shapes must match");
+    AITER_CHECK(max_position > 0, "cos_cache and sin_cache must not be empty");
```
yitingw1, line 4112: Maybe add `AITER_CHECK(!compute_all_q_rope || q.size(0) == num_tokens,
            "compute_all_q_rope requires q to have exactly num_tokens rows");`, too?

csrc/kernels/cache_kernels.cu
ZhiweiYan-96: The position clamp is now restricted to slot_idx < 0, so the existing valid-owner and non-DCP paths remain unchanged.

csrc/kernels/cache_kernels.cu
ZhiweiYan-96: added, thanks for reminding.

yitingw1 (dismissed):
LGTM

XiaobingSuper (dismissed):

op_tests/test_indexer_qk_rope_quant_and_cache.py
```diff
@@ -0,0 +1,158 @@
+# SPDX-License-Identifier: MIT
+
+import torch
+
+from aiter import dtypes, indexer_qk_rope_quant_and_cache
+
+
```
yzhou103, line 123: it seems there's no test for indexer_qk_rope_quant_and_cache before. it's better add test with/without comupute_all_q_rope  and there is a skill in .claude about aiter-op test. you can refer to that.

valarLip (dismissed):

op_tests/test_indexer_qk_rope_quant_and_cache.py
ZhiweiYan-96: image
case updated, thanks for reminding.

valarLip (approved):

PR #5081, opened 2026-08-28
[HIP] Add fused SiTUv2 activation + per-token FP8 quant kernel

XiaobingSuper:
## What

`situv2_and_mul_quant(out, input, scale, d, beta, linear_beta)` computes the
SiTUv2 gated activation and the per-token FP8 quantization of its result in a
single pass:

```
y = beta * tanh(g / beta) * sigmoid(g) * (linear_beta * tanh(u / linear_beta))
out, scale = per_token_fp8(y)
```

with `input` laid out as `[m, 2 * d]` (gate half followed by up half). Callers
that previously ran the activation, materialized a bf16 intermediate, and
quantized it in a second kernel now read the input once and write only the fp8
output plus the scale.

Two forms are dispatched. When the row fits one chunk per thread the
activations stay in registers and a single DPP wave reduction produces the
scale; a single-wave block skips LDS and both barriers entirely. Wider rows
stage through FP32 LDS and sweep the row in a loop. The block/vector-width
ladder is tuned per `d` on gfx950 so the block stays close to `d / VecSize` —
a loose fit leaves whole waves idle. Supports `d` up to 16384, divisible by 8.

## Performance

Kernel-level, measured from a TP8 decode trace over 1536 steps. The fused
kernel replaces the standalone activation plus the MLP share of the per-token
quant kernel:

| per decode step | before | after |
|---|---:|---:|
| standalone SiTUv2 activation | 154.4 us | — |
| MLP share of the per-token fp8 quant | 146.8 us | — |
| `situv2_and_mul_quant` | — | 140.3 us |
| **replaced set** | **301.1 us** | **140.3 us** |

**2.15x on the replaced set.** Call counts corroborate the substitution: the
activation kernel drops to zero and the quant kernel loses exactly the 31.1
MLP calls per step that the fused kernel absorbs. That set is ~1% of total GPU
time, so end-to-end throughput moves within run-to-run variance and is not
claimed here.

Standalone throughput, m = 32768 tokens, bf16 in / fp8 out, MI355X:

| d | us | GB/s |
|---:|---:|-----:|
| 512  |  16.84 | 4990 |
| 768  |  27.01 | 4663 |
| 1024 |  31.36 | 5354 |
| 2048 |  58.91 | 5698 |
| 4096 | 113.03 | 5938 |
| 4224 | 133.83 | 5172 |
| 8192 | 235.28 | 5705 |

A bare `copy_` at the same byte volume measures 6021 / 6768 / 5845 / 5938 GB/s
for d = 512 / 1024 / 2048 / 4096, so from d = 2048 up the kernel runs at the
pure-copy bandwidth ceiling.

## Correctness

`op_tests/test_situv2_and_mul_quant.py` has one case per host dispatch branch
(compile-time `d`, the two `d <= 512` / `d <= 1024` single-pass routes, and the
runtime-`d` multi-pass route), each at two token counts on the shapes that
actually run, plus an all-zero row that exercises the `row_max == 0` guard, an
empty batch, and a `torch.compile(fullgraph=True)` call. Output is within 1 ULP
of the fp32 reference and the scale relative error is ~2e-6.

The fp8 bound comes from `opus::finfo<fp8_t>::max()`, so the OCP e4m3fn range
on gfx950 and the fnuz range on gfx942 are both handled.

### DIFF
--- a/op_tests/test_situv2_and_mul_quant.py
+++ b/op_tests/test_situv2_and_mul_quant.py
@@ -1,0 +1,168 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+import argparse
+import itertools
+
+import pandas as pd
+import torch
+
+import aiter
+from aiter import dtypes
+from aiter.jit.utils.chip_info import get_gfx
+from aiter.ops.activation import situv2_and_mul_quant
+from aiter.test_common import (
+    benchmark,
+    checkAllclose,
+    run_perftest,
+)
+
+torch.set_default_device("cuda")
+
+SUPPORTED_GFX = ["gfx942", "gfx950", "gfx1250"]
+
+# Kimi-K3's values. Neither drives a dispatch branch, so one pair covers them.
+BETA = 4.0
+LINEAR_BETA = 25.0
+
+
+def run_torch(x, beta=BETA, linear_beta=LINEAR_BETA):
+    """Reference activation in fp32; not timed, not in the table."""
+    d = x.shape[-1] // 2
+    gate = x[..., :d].float()
+    up = x[..., d:].float()
+    return (
+        beta
+        * torch.tanh(gate / beta)
+        * torch.sigmoid(gate)
+        * (linear_beta * torch.tanh(up / linear_beta))
+    )
+
+
+@benchmark()
+def test_situv2_and_mul_quant(m, d, dtype):
+    x = torch.randn((m, 2 * d), dtype=dtype)
+    ref = run_torch(x)
+    ref_scale = ref.abs().amax(dim=-1, keepdim=True) / torch.finfo(dtypes.fp8).max
+
+    # The model preallocates both outputs, so the test does too.
+    out = torch.empty((m, d), dtype=dtypes.fp8)
+    scale = torch.empty((m, 1), dtype=dtypes.fp32)
+
+    _, us = run_perftest(situv2_and_mul_quant, out, x, scale, d, BETA, LINEAR_BETA)
+
+    err = checkAllclose(
+        ref,
+        out.to(dtypes.fp32) * scale,
+        rtol=8e-2,
+        atol=3e-1,
+        msg="situv2_and_mul_quant out",
+    )
+    # The scale is an amax over activations the kernel and torch compute to
+    # slightly different last bits, so it is tight but not bit-exact.
+    checkAllclose(
+        ref_scale,
+        scale,
+        rtol=2e-5,
+        atol=3e-8,
+        msg="situv2_and_mul_quant scale",
+    )
+
+    # Reads 2*d bf16 per token, writes d fp8 plus one fp32 scale. Arithmetic is
+    # ~11 ops per output element (three transcendentals counted as one each,
+    # plus the quant multiply), which only serves to place the op on the
+    # roofline -- it is memory bound at every shape here.
+    nbytes = m * (2 * d * x.element_size() + d * out.element_size() + 4)
+    flops = 11 * m * d
+    return {
+        "gfx": get_gfx(),
+        "us": us,
+        "TFLOPS": flops / us / 1e6,
+        "TB/s": nbytes / us / 1e6,
+        "err": err,
+    }
+
+
+def run_edge_shapes():
+    """Two shapes the sweep cannot carry: an all-zero row, whose zero scale the
+    randn inputs never produce, and an empty batch, which would only add a NaN
+    row to the table. Correctness only, so both stay out of it."""
+    d = 768
+    out = torch.empty((4, d), dtype=dtypes.fp8)
+    scale = torch.empty((4, 1), dtype=dtypes.fp32)
+    situv2_and_mul_quant(
+        out, torch.zeros((4, 2 * d), dtype=dtypes.bf16), scale, d, BETA, LINEAR_BETA
+    )
+    assert torch.count_nonzero(out.float()) == 0, "zero row: quantized output"
+    assert torch.count_nonzero(scale) == 0, "zero row: scale"
+
+    situv2_and_mul_quant(
+        torch.empty((0, d), dtype=dtypes.fp8),
+        torch.empty((0, 2 * d), dtype=dtypes.bf16),
+        torch.empty((0, 1), dtype=dtypes.fp32),
+        d,
+        BETA,
+        LINEAR_BETA,
+    )
+    aiter.logger.info("situv2_and_mul_quant edge shapes passed")
+
+
+def main():
+    if get_gfx() not in SUPPORTED_GFX:
+        aiter.logger.warning(
+            "situv2_and_mul_quant unsupported on %s; skipping", get_gfx()
+        )
+        return
+
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
+        description="config input of test",
+    )
+    parser.add_argument(
+        "-d",
+        "--dtype",
+        type=dtypes.str2Dtype,
+        nargs="*",
+        default="bf16,",
+        help="""Data type of the activation input.
+    e.g.: -d bf16""",
+    )
+    parser.add_argument(
+        "-m",
+        "--m",
+        type=int,
+        nargs="*",
+        default=[1, 32, 129, 1024, 8192, 32768],
+        help="""Number of tokens.
+    e.g.: -m 32""",
+    )
+    parser.add_argument(
+        "-n",
+        "--dim",
+        dest="d",
+        type=int,
+        nargs="*",
+        # 384 takes the single-wave path, 768 the compile-time specialization
+        # Kimi-K3's shared experts run, 1024 the VecSize 16 path, and 4224 the
+        # runtime-d dense MLP.
+        default=[384, 768, 1024, 4224],
+        help="""Hidden size per gate/up half.
+    e.g.: -n 768""",
+    )
+    args = parser.parse_args()
+
+    run_edge_shapes()
+
+    for dtype in args.dtype:
+        df = [
+            test_situv2_and_mul_quant(m, d, dtype)
+            for d, m in itertools.product(args.d, args.m)
+        ]
+        df = pd.DataFrame(df)
+        aiter.logger.info(
+            "situv2_and_mul_quant summary (markdown):\n%s", df.to_markdown(index=False)
+        )
+
+
+if __name__ == "__main__":
+    main()
--- a/aiter/ops/activation.py
+++ b/aiter/ops/activation.py
@@ -50,6 +50,25 @@ def silu_and_mul_quant(
 
 
 @compile_ops("module_activation", develop=True)
+def situv2_and_mul_quant(
+    out: Tensor,
+    input: Tensor,
+    scale: Tensor,
+    group_size: int,
+    beta: float,
+    linear_beta: float,
+    shuffle_scale: bool = False,
+) -> None:
+    """Apply SiTUv2 and per-token FP8 quantization.
+
+    All tensors must be contiguous ROCm tensors. ``input`` is BF16 with shape
+    ``[..., 2 * d]`` where ``d`` is divisible by 8, ``out`` is FP8 with
+    ``input.numel() / 2`` elements, and ``scale`` is FP32 with one element per
+    token. Only ``group_size == d`` and ``shuffle_scale=False`` are supported.
+    """
+
+
+@compile_ops("module_activation", develop=True)
 def gelu_and_mul(out: Tensor, input: Tensor) -> None: ...
 
 
--- a/csrc/include/activation.h
+++ b/csrc/include/activation.h
@@ -23,6 +23,13 @@ void scaled_silu_and_mul(const aiter_tensor_t& out, const aiter_tensor_t& input,
 void silu_and_mul_quant(const aiter_tensor_t& out, const aiter_tensor_t& input,
                         const aiter_tensor_t& scale, int group_size,
                         float limit = 0.0f, bool shuffle_scale = false);
+void situv2_and_mul_quant(const aiter_tensor_t& out,
+                          const aiter_tensor_t& input,
+                          const aiter_tensor_t& scale,
+                          int group_size,
+                          float beta,
+                          float linear_beta,
+                          bool shuffle_scale = false);
 void gelu_and_mul(const aiter_tensor_t& out, const aiter_tensor_t& input);
 void gelu_tanh_and_mul(const aiter_tensor_t& out, const aiter_tensor_t& input);
 void gelu_fast(const aiter_tensor_t& out, const aiter_tensor_t& input);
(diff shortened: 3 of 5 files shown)


XiaobingSuper:
The ci failed cases are not related to this PR change.

zufayu:
## aiter PR #5081 — [HIP] Add fused SiTUv2 activation + per-token FP8 quant kernel

**Fuses the SiTUv2 gated activation with per-token FP8 quantization so the bf16 intermediate never reaches HBM.**

Validation (deterministic): NOT RUN — no exact-head `validation_report.json`. The index-width scan over this diff returns 0 candidates, which is correct here: `token_idx` is already `int64_t` and the address arithmetic carries no unwidened multiply.

The activation math checks out. `two_log2e = 2.8853900817779268` is `2·log2(e)`, `gate_tanh_mul = -2·log2(e)/beta` is the correct exponent substitution for `tanh(g/beta)`, the `exp2 → 1+x → rcp` chain reconstructs `sigmoid(g)` exactly, and `beta > 0` is enforced host-side, so the `beta = 0` division is unreachable.

🔴 **`fp8_max` is hard-coded to 448.0 and is wrong on gfx942.** `activation_kernels.cu:265` declares `constexpr float fp8_max = 448.0f;` and the quantization path carries no architecture branch — the only `#if` in the file guards the gfx1250 asm helpers, both of which have portable `#else` fallbacks. But `aiter/utility/dtypes.py:10` maps gfx942 to `torch.float8_e4m3fnuz`, whose maximum is **240**, not 448.

The consequence is not an edge case. `inv_scale = 448 / row_max` maps the row peak to 448; fnuz saturates at 240, so every element above `row_max × 240/448 = 53.6%` of the peak collapses to the same value, and dequantization returns `0.536 × row_max` for all of them — up to 46% error, on every row, on every call. `GPU_ARCH_LIST` is `gfx942;gfx950` and MI300X runs eight standard shards, so this is not an uncovered architecture.

The correct form is already in this file. `:523` computes `constexpr float inverted_DTYPE_MAX = 1.f / static_cast<float>(opus::finfo<DTYPE_O>::max());`, deriving the bound from the output type at compile time; `quant_kernels.cu` and `dsv4_rotate_quant.cu` use the same helper. **Author must** derive the bound from `opus::finfo<DTYPE_O>` instead of hard-coding it. [verified]

⚠️ **The test that would have caught that never executes.** `op_tests/test_situv2_and_mul_quant.py` computes its reference as `ref.abs().amax(dim=-1) / torch.finfo(dtypes.fp8).max` — architecture-aware, so on gfx942 `assert_close(scale, ref_scale, rtol=2e-5)` fails by a factor of 448/240 ≈ 1.87. It does not fail, because the file has no `__main__` guard: CI runs `op_tests` at depth 1 as `timeout 60m python3 "$file"` (`.github/scripts/aiter_test.sh:85`), which imports the module, defines the four test functions, and exits 0. pytest collection applies only to `op_tests/triton_tests/`. The test is correct and inert. **Author must** add a `__main__` block that calls the four cases, so the file actually validates something under the runner CI uses. [verified]

⚠️ **The quantization half was rewritten rather than reused, and that is where the bound went wrong.** The kernel at `:481` already templates the activation as `float (*ACT_FN)(const DTYPE_I&)` — a pattern used in four kernels in this file — and the surrounding machinery (row max, reduction, type-derived bound, scale write, fp4 branch, `group_size`, `shuffle_scale`) is shared across all of them. This PR reimplements that machinery in 592 new lines and diverges on exactly the one constant that the shared version derives correctly.

To be fair to the design: SiTUv2 genuinely does not fit the existing `ACT_FN` signature. It is binary — the `up` side carries its own `linear_beta·tanh(u/linear_beta)` rather than a plain multiply — and it needs two runtime scalars, so the signature would have to be extended. The activation half is a real change. The quantization half is not. **Reviewer should ask** whether the quantization path can be shared with the existing template, with the activation signature extended rather than the whole kernel duplicated. [verified]

📝 **The packed path, and with it the gfx1250 asm, is reachable for exactly one shape.** The guard is `if constexpr(StaticD == 768 && (VecSize == 8 || VecSize == 4))`, and the wave64 route for `d == 768` is `LAUNCH_SITUV2_SP(64, 16, 768)` — `VecSize = 16` fails that condition and falls to the scalar path. So `situv2_bf16_mul_lo/hi` only run at `d == 768` on wave32. In the PR's own table `d = 768` is the slowest row at 4663 GB/s, against 5354 at 1024 and 5698 at 2048. `v_fma_mix_f32_bf16`, `op_sel_hi` and the `neg(0)` operand form have no precedent anywhere in `csrc/`. **Reviewer should ask** whether that path earns its complexity. [verified]

📝 **The headline speedup has no supporting data, and the asm has never been compiled.** "Against activation-then-quant as two kernels this is 1.5-2.8x" is the central justification, but the table reports only the fused kernel's absolute throughput — the two-kernel baseline is not measured anywhere in the description. Separately, the asm compiles only under `#if defined(__gfx1250__)`, and gfx1250 testing is opt-in behind `ci:gfx1250-ffm-triton`, so that block has likely never been through a compiler in CI. Validating it does not need the hardware: a build-only `hipcc --offload-arch=gfx1250 -c` job would settle whether `neg(0)` assembles at all. [verified]

XiaobingSuper:
Thanks — the two blocking findings are both real and are fixed in `4c415bc`.

**`fp8_max` hard-coded to 448.** Confirmed, and the reason it is not caught by the
existing dtype check is worth recording: `situv2_and_mul_quant` does validate
`out.dtype() == AITER_DTYPE_fp8`, but `aiter_enum.h` has a single `AITER_DTYPE_fp8`
and `dtypes.py:48` folds both `float8_e4m3fn` and `float8_e4m3fnuz` onto it, so the
check passes on gfx942 and the kernel then scales to 448 against a format that
saturates at 240. Now derived from the output type, matching `:523` and the other
quant kernels:

```cpp
// 448 on OCP e4m3fn, 240 on the gfx942 fnuz format -- derive it from the
// output type rather than hard coding the OCP bound.
constexpr float fp8_max = opus::finfo<opus::fp8_t>::max();
```

The `"output must be OCP FP8 E4M3FN"` message was also inaccurate for the same
reason and is now `"output must be FP8 E4M3"`. (Minor: the line reference in the
review is `:265`, which is `act_and_mul_bias_kernel`; the constant is at `:838`.)

**The test never executes.** Also confirmed — `aiter_test.sh:30` collects
`op_tests` at depth 1 and runs `python3 "$file"`, so without a `__main__` guard the
module imports and exits 0. Added the same guard the other pytest-style files in
`op_tests` use. Verified by running `python3 op_tests/test_situv2_and_mul_quant.py`
against a clean rebuild: 9 passed. The test was already architecture-aware, so on
gfx942 it would have failed by 448/240 before this fix.

**gfx1250 asm has never been compiled.** It has now:
`hipcc --offload-arch=gfx1250 --cuda-device-only -S` builds clean and emits 249
`v_fma_mix_f32_bf16`, including the `op_sel` and negated-operand forms —
`v_fma_mix_f32_bf16 v9, v1, s17, -v6 op_sel:[1,0,0] op_sel_hi:[1,0,0]`. So the
block assembles; agreed that a build-only job would be the right way to keep it
that way.

**Headline speedup unsupported.** Fair, and the description is rewritten around
kernel-level trace numbers instead: the fused kernel replaces a 154.4 us activation
plus the 146.8 us MLP share of the quant kernel with one 140.3 us kernel per decode
step, 2.15x on the replaced set, with call counts confirming the substitution. That
set is ~1% of total GPU time, so end-to-end is within run-to-run variance and is no
longer claimed.

**Sharing the quantization path with the existing template.** Considered and not
taken. Beyond the signature problem you note, the quantization half here is not a
copy: it is single-wave and register-resident, with the activations held in VGPRs
across a DPP-only reduction and no LDS or barrier at all on the shapes that matter.
That is where the performance comes from — measured 6% over the two-wave form at
m = 32768 — and folding it back into the shared template would give it up. The one
thing that genuinely should have been shared was the bound, and that is now shared.

**Packed path reachable for one shape.** Correct: `d == 768` on wave64 dispatches
`VecSize = 16`, which fails the guard, so `situv2_bf16_mul_lo/hi` only run on
wave32. On gfx950 packed and scalar measure within 0.6% of each other across
m = 16..32768 in paired build-and-benchmark rounds — the compiler already folds the
bf16 convert into the multiply there. The path is kept for gfx1250, where
`v_fma_mix_f32_bf16` does select-half, convert and multiply in one VALU op.

XiaobingSuper:
Copilot's three inline comments are addressed in `91fafc9`.

**Strict aliasing on the bf16→packed-uint32 pun** — real, and the fix is free.
`reinterpret_cast<const packed_bf16x2&>(gate)` is now
`__builtin_bit_cast(packed_bf16x2, gate)`. Codegen is unchanged: a build-only
`hipcc --offload-arch=gfx1250 --cuda-device-only -S` emits the same 249
`v_fma_mix_f32_bf16` before and after, and the gfx950 op tests still pass.

**Binding docstring says "per-group quantization"** — correct, the implementation
rejects `group_size != d`. Now "Fused SiTUv2 and per-token FP8 quantization.
Requires group_size == d.", which matches the Python docstring and the host check.

**`"output must be FP8 E4M3"` still names a format** — fair; both variants are E4M3
so it was not wrong, but it implies the check distinguishes them when
`AITER_DTYPE_fp8` cannot. Now "output must be the aiter FP8 dtype", which is what
the check actually tests.

Rebased onto the main merge; `python3 op_tests/test_situv2_and_mul_quant.py` is
9 passed on the merged base.

op_tests/test_situv2_and_mul_quant.py
```diff
@@ -0,0 +1,113 @@
...
+    def compiled(out: torch.Tensor, inp: torch.Tensor, out_scale: torch.Tensor):
+        situv2_and_mul_quant(out, inp, out_scale, d, 4.0, 25.0)
+        return out, out_scale
+
+    compiled(quantized, x, scale)
+    ref = _situv2_ref(x, 4.0, 25.0)
+    torch.testing.assert_close(quantized.float() * scale, ref, rtol=8e-2, atol=3e-1)
+
+
+if __name__ == "__main__":
```
valarLip, line 167: https://github.com/ROCm/aiter/tree/main/.claude/skills/aiter-op-test

valarLip (changes requested):

op_tests/test_situv2_and_mul_quant.py
XiaobingSuper: Trimmed further in 8dd386a -- the first pass was over-built. There is only one kernel under test, so it is now the single-kernel-vs-reference shape from the skill rather than a candidate loop: the unfused torch+`get_hip_quant` comparison and the `--betas` sweep axis are both gone (beta drives no dispatch branch, so sweeping it only added two constant columns; the fused-vs-unfused numbers live in the PR description as trace data).

What stays is what the standard asks for: `@benchmark` on `test_situv2_and_mul_quant(m, d, dtype)`, `run_perftest` + `checkAllclose` on both the dequantized output and the scale, `us` / `TFLOPS` / `TB/s` / `err` + `gfx` columns, an `itertools.product` sweep over the four `d` values that select different dispatch paths, the markdown table, and the `get_gfx()` allow-list in `main()` under a `__main__` guard.

Full default sweep on gfx950, 24 cases, all `err = 0`:

```
|     m |    d | dtype          | gfx    |        us |      TFLOPS |       TB/s |   err |
|------:|-----:|:---------------|:-------|----------:|------------:|-----------:|------:|
|     1 |  384 | torch.bfloat16 | gfx950 |   1.6188  |  0.00260934 | 0.00118854 |     0 |
|    32 |  384 | torch.bfloat16 | gfx950 |   2.41682 |  0.0559281  | 0.0254748  |     0 |
|   129 |  384 | torch.bfloat16 | gfx950 |   2.48751 |  0.219053   | 0.0997771  |     0 |
|  1024 |  384 | torch.bfloat16 | gfx950 |   2.77062 |  1.56116    | 0.711096   |     0 |
|  8192 |  384 | torch.bfloat16 | gfx950 |   6.58166 |  5.25749    | 2.39475    |     0 |
| 32768 |  384 | torch.bfloat16 | gfx950 |  16.9879  |  8.14769    | 3.71121    |     0 |
|     1 |  768 | torch.bfloat16 | gfx950 |   1.93341 |  0.00436948 | 0.0019882  |     0 |
|    32 |  768 | torch.bfloat16 | gfx950 |   2.88306 |  0.093767   | 0.0426658  |     0 |
|   129 |  768 | torch.bfloat16 | gfx950 |   2.91862 |  0.373393   | 0.169901   |     0 |
|  1024 |  768 | torch.bfloat16 | gfx950 |   3.46183 |  2.4989     | 1.13704    |     0 |
|  8192 |  768 | torch.bfloat16 | gfx950 |  10.3055  |  6.71542    | 3.05564    |     0 |
| 32768 |  768 | torch.bfloat16 | gfx950 |  29.8069  |  9.28724    | 4.22587    |     0 |
|     1 | 1024 | torch.bfloat16 | gfx950 |   1.9431  |  0.00579692 | 0.00263702 |     0 |
|    32 | 1024 | torch.bfloat16 | gfx950 |   2.81346 |  0.128116   | 0.0582798  |     0 |
|   129 | 1024 | torch.bfloat16 | gfx950 |   2.91226 |  0.498944   | 0.22697    |     0 |
|  1024 | 1024 | torch.bfloat16 | gfx950 |   3.69063 |  3.1253     | 1.4217     |     0 |
|  8192 | 1024 | torch.bfloat16 | gfx950 |  11.9443  |  7.72544    | 3.51431    |     0 |
| 32768 | 1024 | torch.bfloat16 | gfx950 |  32.356   | 11.4074     | 5.18925    |     0 |
|     1 | 4224 | torch.bfloat16 | gfx950 |   2.51044 |  0.0185083  | 0.00841447 |     0 |
|    32 | 4224 | torch.bfloat16 | gfx950 |   3.45583 |  0.430243   | 0.195602   |     0 |
|   129 | 4224 | torch.bfloat16 | gfx950 |   3.7277  |  1.60792    | 0.731012   |     0 |
|  1024 | 4224 | torch.bfloat16 | gfx950 |   7.74686 |  6.14173    | 2.79223    |     0 |
|  8192 | 4224 | torch.bfloat16 | gfx950 |  39.6095  |  9.60965    | 4.36885    |     0 |
| 32768 | 4224 | torch.bfloat16 | gfx950 | 144.951   | 10.5037     | 4.77533    |     0 |
```

op_tests/test_situv2_and_mul_quant.py
```diff
@@ -0,0 +1,206 @@
...
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
+        description="config input of test",
+    )
+    parser.add_argument(
+        "-d",
+        "--dtype",
+        type=dtypes.str2Dtype,
+        nargs="*",
+        default="bf16,",
```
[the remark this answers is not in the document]
XiaobingSuper: @copilot This one does not hold. Argparse *does* apply `type=` to a **string** default -- `parse_args` ends with a pass that converts any default still holding its original string through `_get_value`, so `str2Dtype("bf16,")` runs and the trailing comma makes it a tuple:

```
>>> p.add_argument("-d", "--dtype", type=dtypes.str2Dtype, nargs="*", default="bf16,")
>>> p.parse_args([]).dtype
(torch.bfloat16,)
```

The loop therefore iterates dtypes, not characters. This is also the pattern `op_tests/test_quant.py` uses, and the full sweep run emits one table with `dtype = torch.bfloat16`, not five character-keyed ones. No change made.

The `--betas` comment is against an older revision -- that option was removed in the meantime, so it no longer applies either.

valarLip (approved):

PR #5101, opened 2026-08-29
[Triton/Gluon] Resolve conv configs in the launch layer

Boss2002n:


saeid-rostami (approved):
Thanks for cleaning this up. LGTM!

brunomazzottiamd (changes requested):
I'm not sure if it's a good idea to move kernel config resolution to the PyTorch wrapper layer. https://github.com/ROCm/aiter/pull/939 did a reorg, splitting kernel code + kernel config resolution from PyTorch wrappers. The idea at the time was to provide a PyTorch-free path to JAX, so JAX wrappers would be able to reuse kernel code + kernel config resolution through https://github.com/jax-ml/jax-triton without bringing PyTorch as a bloated dependency.

The code looks good, but I'm requesting changes just to prevent merging. We should discuss this approach with the team before proceeding.

The following PRs are also blocked by this discussion:

- https://github.com/ROCm/aiter/pull/5102
- https://github.com/ROCm/aiter/pull/5103
- https://github.com/ROCm/aiter/pull/5104
- https://github.com/ROCm/aiter/pull/5105
- ~https://github.com/ROCm/aiter/pull/5106~ → this is **wrong**!

FYI: @rahulbatra85 + @azaidy + @vgokhale

PR #5107, opened 2026-08-29
[HIP] [OPUS] [JIT] [gfx1250] Add auto-detect stepping and B0-only guard

dbyoung18:
Current GFX1250 ASM Kernel are targeting B0 only. Add auto-detect and guard to ensure not dispatch on A0 to avoid unexpected behaviours.

image


valarLip (changes requested):
## Review

The goal is right — a B0-only code object on an A0 device should fail loudly rather than misbehave. Everything below is verified against `3a124e1` in a worktree; commands and measurements are included.

The three structural problems: the guard is not the single choke point it claims, its C++ failure mode is `std::abort()` on every pybind path, and its Python half fails **open** while reading a different GPU than its C++ half.

### Blocking

**1. `csrc/include/aiter_hip_common.h:456` — not a single choke point; `AiterAsmKernelFast` bypasses it entirely.**

The guard lives in `AiterAsmKernel`'s constructor, under a comment reading "Single choke point". But `AiterAsmKernelFast` has a public constructor at :251 (`AiterAsmKernelFast(const char* kernel_name, const void* hsaco) { init(...); }`) with no check, and three places construct it directly:

- `csrc/opus_gemm/include/gfx1250/opus_co_launch_gfx1250.cuh:68` and `:119` (`co_kernel()`) — the **gfx1250-only** OPUS `.co` loader
- `csrc/cpp_itfs/moe/asm_moe.cpp.jinja:71`
- `csrc/cpp_itfs/mla/asm_mla_decode_fwd.cpp.jinja:93`

So on gfx1250 A0 the OPUS GEMM path loads and launches its code objects with zero checks — exactly the unguarded case the PR exists to prevent. The check belongs in `AiterAsmKernelFast::init()`, which both classes funnel through.

**2. `:458` — `AITER_CHECK` does not throw by default; it calls `std::abort()`.**

`aiter_detail::check_fail` (same file, :46-64) throws only when the thread-local `g_aiter_can_throw` is true. It is declared `inline thread_local bool g_aiter_can_throw = false` at :32, and set true in exactly five places: `csrc/include/aiter_ctypes_error.h:87` (the ctypes path), three sites in `csrc/opus_gemm/opus_gemm.cu`, and one in `opus_bmm_launchers_a8w8_mxscale_gfx950.cuh`. Pybind modules — `csrc/py_itfs_cu/pa_sparse_prefill_opus_kernels.cu`, the raw `AITER_C_ITFS` entries in `csrc/py_itfs_cu/asm_mla.cu` — run with it false.

Net effect on gfx1250 A0: a hard SIGABRT with no Python traceback and no chance to fall back. That is strictly worse than the "unexpected behaviour" being guarded against, and it contradicts the Python half of the same PR, which raises `RuntimeError`.

**3. `aiter/jit/utils/asm_guard.py:19` — the safety gate fails OPEN, and `@lru_cache` pins the wrong answer for the process lifetime.**

```python
    except Exception:  # noqa: BLE001
        return True
```

`get_asic_revision()` is written to raise `RuntimeError` when `rocminfo` is missing from PATH, exits non-zero, or its output lacks an `ASIC Revision` line; `get_gfx_runtime()` raises on an unknown arch. All of that is swallowed and converted to "asm supported". In a slim ROCm container without `rocminfo`, on a real gfx1250 A0 board, every `require_gfx1250_asm()` becomes a no-op and B0-only asm is dispatched on A0 — the precise case this PR was written to block.

Because `@functools.lru_cache(maxsize=1)` sits above the `try`, a single transient failure (fork under memory pressure, PATH not yet set in an entrypoint) is memoized forever and never retried. A hardware-compat gate should fail closed, or at minimum log the swallowed exception.

**4. `aiter/jit/utils/chip_info.py:99` — the Python gate reads a different GPU than the C++ gate.**

Python takes the stepping of the **first** `rocminfo` GPU agent; C++ reads the **current HIP device**. `rocminfo` does not honour `HIP_VISIBLE_DEVICES` — measured on this node:

```
all agents                   -> 8
HIP_VISIBLE_DEVICES=3        -> 8      (ignored)
ROCR_VISIBLE_DEVICES=3       -> 1      (honoured)
```

vLLM/SGLang/ATOM workers set `HIP_VISIBLE_DEVICES`. On a mixed-stepping gfx1250 bring-up node — the realistic setting for an A0/B0 guard — with A0 in slot 0 and B0 in slot 3, `HIP_VISIBLE_DEVICES=3` makes Python raise and block a perfectly good B0 GPU; with the reverse layout Python permits and the C++ `AITER_CHECK` aborts the process. `lru_cache(maxsize=1)` with no device argument also means the answer never follows a later `torch.cuda.set_device()`. Reading the revision through HIP for the current device would make both halves agree by construction.

**5. `aiter/jit/utils/asm_guard.py:9` — neither helper is wrapped in `@torch_compile_guard()`, so `fullgraph=True` over any guarded op now fails.**

Reproduced with the real functions and a warm `lru_cache`:

```
>>> is_gfx1250_asm_supported()          # warm the cache
>>> torch.compile(lambda x: (require_gfx1250_asm("probe"), x+1)[1], fullgraph=True)(t)
Unsupported: Attempted to call function marked as skipped
  Explanation: Dynamo does not know how to trace the builtin `posix.stat.`
```

It dies on `posix.stat` inside `executable_path("rocminfo")` before reaching `subprocess.run`, and the `except Exception` does not rescue a tracer-level `Unsupported`. `require_gfx1250_asm` is the unconditional **first statement** of `aiter/mla.py:574/1188/1243/1679`, `aiter/ops/gemm_op_a8w8.py:1316`, `gemm_op_a4w4.py:397/431`, `gemm_op_a8w4.py:50`, `attention.py:475` — all plain Python wrappers Dynamo traces.

The fix is one decorator, and the same file already uses it at `chip_info.py:40` and `:161`. `aiter/ops/mha_v4.py:214-219` documents this exact failure mode: *"Dynamo traces the body of a cached function regardless of whether the cache is warm ... fails outright under `fullgraph=True`. `get_gfx()` keeps its own rocminfo probe opaque the same way."*

### Should fix before merge

**6. `csrc/include/aiter_hip_common.h:192` — `is_gfx1250_asm_supported()` is documented "Non-raising" but can `std::abort()`.**

Both callees use `HIP_CALL`: `get_gpu_arch()` (:472/480/481) and `get_asic_revision()` (:509). `HIP_CALL` routes to `aiter_detail::aiter_check_fatal` → `std::abort()` (:78-90). `csrc/cpp_itfs/mha_bwd.cu:385` is the one caller the comment promises is soft — it is supposed to `return -1` so `mha_bwd` falls back to CK. A transient `hipGetDevice`/`hipGetDeviceProperties` failure now aborts the process from inside a predicate documented as non-raising.

CK's equivalent, `ck_tile::get_device_revision()` (`3rdparty/composable_kernel/include/ck_tile/host/device_prop.hpp:55-70`), returns -1 on HIP error precisely to stay non-fatal. The comment is also wrong in the other direction: with `hipGetDeviceCount() == 0`, `get_gpu_arch()` returns `"No GPU Found"` and the predicate returns **true**.

**7. `aiter/ops/pa_sparse_prefill_opus.py:103` — sweep gaps: several gfx1250 asm entry points are ungated and land on the C++ abort.**

- `aiter/ops/pa_sparse_prefill_opus.py:103` / `:244` — gfx1250-only, loads `hsa/gfx1250/mla_v4_opus/*.co`, no Python guard
- `aiter/ops/mha.py:584` / `:656` / `:758` — `fmha_fwd_with_sink_asm`, `fmha_fwd_with_sink_varlen_asm`, `fmha_fwd_mxfp8_asm`; only the internal `can_impl_*` predicates were gated, not the public wrappers
- `aiter/ops/attention.py:788` / `:887` / `:1071` — the raw `@compile_ops` bindings `mla_decode_stage1_asm_fwd`, `mla_prefill_asm_fwd`, `mla_prefill_ps_asm_fwd`. These are public `aiter.*` symbols external consumers call directly, bypassing the `mla.py` wrappers that did get the guard.

**8. `aiter/mla.py:574` — inconsistent failure policy, and no capability predicate for frameworks to select on.**

`aiter/ops/mha.py:1937/1973/2966/3647` fold the check into `can_impl_*`/`can_try_*` so the router degrades to CK/Triton, and `mha_bwd.cu:385` likewise returns -1. But `mla.py:574/1188/1243/1679`, `attention.py:475` and the four gemm ops **raise**.

Worse, `require_gfx1250_asm` is the first statement of `mla_decode_fwd`, above branches (persistent vs non-persistent split-KV) that may never touch a B0-only `.co` — the guard sits above the dispatch it is meant to gate. And `is_gfx1250_asm_supported` lives in `aiter.jit.utils.asm_guard` and is **not** re-exported from `aiter`, so SGLang/vLLM have nothing to select on at init: the server crashes at the first decode instead of choosing a Triton/CK backend at startup. Exporting the predicate would be a small, high-value addition.

**9. `aiter/ops/mha.py:1937` — the two halves of each new gate read the arch from different sources.**

`mha.py:12` imports `get_gfx` (the build-time one: `get_gfx_custom_op()` → `GFX_MAP`), while `is_gfx1250_asm_supported()` internally uses `get_gfx_runtime()` (always rocminfo). `chip_info.py:75-79` documents the split explicitly.

With `GPU_ARCHS=gfx1250` exported for a cross-compile/CI run on a gfx950 box, `get_gfx() == "gfx1250"` is true while `get_gfx_runtime()` returns gfx950 — so `is_gfx1250_asm_supported()` takes its non-gfx1250 early return and yields `True`. The gate passes and a gfx1250 asm kernel is selected on gfx950. The new conjunct reads as if it hardened this path but adds no protection there.

Note also that `batched_gemm_op_a8w8.py`, `batched_gemm_op_bf16.py`, `gemm_op_a4w4.py`, `gemm_op_a6w6.py` and `gemm_op_a8w8.py` all alias `get_gfx_runtime as get_gfx`, so the same identifier means different things in different files feeding the same guard. And the 4th site (`mha.py:3647`) was reformulated as a split `if` while the other three stayed as `and` conjunctions — same predicate, three spellings.

**10. `csrc/py_itfs_cu/pa_sparse_prefill_opus_kernels.cu:380` — the gate also blocks code objects the repo itself documents as compiler output, and the B0-only premise has no in-tree evidence.**

This file says, at :377-379:

> *aiter's prebuilt-code-object loader. Its name comes from the hand-written assembly kernels it was introduced for; **these code objects are compiler output**, but the load-and-launch-by-symbol path is the same.*

…then does `using PrebuiltKernel = AiterAsmKernel;` and instantiates at :484/490/627/633 — so the new constructor guard applies to them. Meanwhile the PR's own comment asserts "every shipped gfx1250 asm kernel is B0-only", and a repo-wide grep for `A0`/`B0`/`asicRevision` across docs, CSVs and manifests returns nothing to support it. The code objects don't encode it either — `readelf -h` on `hsa/gfx1250/mla_v4_opus/*.co` shows `Flags: 0x449, sramecc any`, a plain target-id with no stepping feature.

(I could not reproduce a `.comment`-section producer note on any of the ten `hsa/gfx1250/*` families — `llvm-objdump -s -j .comment` returns nothing for all of them — so I'm not claiming which families are compiler-generated beyond what the source comment states. The point stands on the source comment alone.)

Net effect on gfx1250 A0: two kernels whose A0 status is undocumented go from working to killing the process, since `aiter/ops/pa_sparse_prefill_opus.py` has no Python guard and lands on the abort path (finding #2). If the B0-only claim is based on validation you have, it would help to record which kernels were tested and how.

### Minor

**11. `csrc/include/aiter_hip_common.h:204` — the allowlist escape hatch is unreachable from Python, and the match is unanchored.**

`kA0AllowList` containing only the sentinel is intentional and correct today (no A0-capable kernels), and the `// keep last; add entries above` comment is right — no complaint there. Two real issues:

- `asm_guard.py`'s `require_gfx1250_asm` has **no allowlist counterpart**, so for all ten Python-guarded ops the `RuntimeError` fires before C++ is ever entered. An engineer follows the documented procedure, adds a substring, rebuilds — and the op still raises.
- Matching is unanchored `name.find(*p)`, so adding `"mla"` or `"fmha_fwd"` to allow one kernel would whitelist every B0-only variant sharing that substring. Worth an exact-match or prefix rule before the list is ever non-empty.

Also `const std::string_view name{kernel_name}` is constructed unconditionally for a loop that currently cannot run, and is UB if `kernel_name` is ever null.

**12. `aiter/jit/utils/chip_info.py:105` — the regex is tied to rocminfo's current print format, and the docstring is wrong in two ways.**

`re.search(r"ASIC Revision:\s*(\d+)", line)` against today's `ASIC Revision:  0(0x0)` works (verified on this node). If a ROCm release prints `0x1` only, `\d+` matches the leading `0` and reports A0 — disabling all asm on a genuine B0 part.

The docstring says rocminfo was chosen *"so it stays stable across ROCm releases"*, but scraping human-readable text is a weaker contract than the structured `hipDeviceProp_t.asicRevision` field this PR's own C++ half uses, and which CK already wraps as `ck_tile::get_device_revision` — consumed with the same `== 0 => A0` semantics in ~10 CK tests, which does confirm `>= 1` is the right B0 discriminator. The second claim, *"always reflects the actual running GPU"*, is false per finding #4: it returns the first `Name: gfx` agent's revision.

**13. `aiter/jit/utils/chip_info.py:91` — a second (in practice third) `rocminfo` spawn for data already captured.**

`_detect_native()` (:22) already runs `rocminfo`, and its captured stdout already contains the `ASIC Revision` lines. `rocminfo` measures 0.23-0.44 s per spawn on this node.

`import aiter` already spawns it twice, because `aiter/jit/core.py:24` does `sys.path.insert(0, .../utils/)` and imports the flat `chip_info`, while ~30 runtime files import `aiter.jit.utils.chip_info` — two distinct module objects with independent `lru_cache` slots (verified: `a is not b` → True). `asm_guard` binds only the package copy, so the first guard hit adds a third fork. One `@lru_cache _rocminfo_text()` shared by both parsers would fix this and also guarantee the arch and the revision come from the same agent, rather than two independent scans of two separate invocations.

**14. `aiter/jit/utils/asm_guard.py:23` — `[SKIP]` is misleading and the name doesn't match the behavior.**

Nothing is skipped — control never returns — so anyone grepping for `[SKIP]` concludes a fallback ran. Every A0 failure is reported twice, the warning is unconditioned (unlike the cached predicate) so a serving loop that catches and retries floods the log at request rate, and raw `\033[93m` bytes land in redirected CI logs.

Separately, `require_gfx1250_asm` does not require gfx1250 — it requires B0 stepping and is a no-op on every other arch. `aiter/ops/mla_sparse_prefill.py:99-102` already shows the confusion: a real arch check and the stepping check sit stacked as two consecutive, differently-worded raises. Something like `require_gfx1250_b0_stepping` would say what it does. (`# noqa: BLE001` at :19 is also inert — `pyproject.toml` states the rule set stays exactly ruff's default, which doesn't include BLE001.)

**15. No tests, despite an exact in-repo precedent.**

`get_asic_revision` (text parsing) and `is_gfx1250_asm_supported` / `require_gfx1250_asm` (comparison + threshold) are pure Python and testable without a gfx1250 A0 device, yet the PR ships zero tests — so the regex, the agent-selection loop, the `>= 1` threshold and the fail-open branch are all unverified, and the A0 path can only be exercised on hardware almost nobody has.

`op_tests/test_jit_arch_guard.py` is the precedent: *"Unit tests for the JIT arch-coverage guard offload-target parser"*, a plain `test_*()` file that monkeypatches `core.get_gfx_runtime` and needs no GPU. It's registered at `.github/scripts/split_tests.sh:171`, so a new file would need the same registration. `CONTRIBUTE.md` ("Before Submitting": *"Includes relevant tests"*) asks for this.

zufayu:

**Adds gfx1250 silicon-stepping auto-detection (ASIC revision via HIP) and blocks every shipped hand-written gfx1250 asm code object from dispatching on A0 stepping — 10 Python wrapper guards, 4 mha capability predicates, and a C++ loader gate — while leaving OPUS compiler-output .co paths deliberately ungated.**

Validation (deterministic): NOT RUN — REVIEW_AUTO_VALIDATE=0 (auto-validation disabled); triage required a run (15 runtime paths changed) and a target existed (op_tests/test_asm_guard.py); the 9 unit tests pass when run by hand on this gfx942 box.
Perf (advisory): MEASURED — require_gfx1250_asm no-op path on gfx942 (torch 2.9.1/ROCm 7.0): head 8.5–15.2 μs/call (3000 calls × best-of-5 × 3 fresh processes) vs base 0 μs (the call does not exist on base — head-only, so this is the added overhead, not a base-vs-head regression measurement); the underlying cached decision costs 0.13 μs.

🔴 [verified] aiter/jit/utils/chip_info.py `_asic_revision_via_hip` resolves `hipDeviceGetAttribute` via `ctypes.CDLL(None)` with the comment "resolve HIP symbols already loaded by torch" — verified false on this ROCm 7.0 + torch 2.9.1 environment: the symbol is absent from the process global scope before CUDA init, after CUDA init, and after `import aiter` (torch dlopens its HIP runtime RTLD_LOCAL), and torch 2.9.1 device properties expose no `asicRevision`, so this is the only detection path. Reproduced end-to-end with this PR's own head code: `get_asic_revision()` raises AttributeError → the guard logs "could not read ASIC revision" and fails closed → `require_gfx1250_asm` raises "asm is only supported on gfx1250 B0+ (current device is gfx1250 A0)" — on B0-class silicon (this GPU reads asicRevision=1 via the soname dlopen the PR's comment rejects, which works here). On any gfx1250 node whose torch loads HIP this way, every guarded op hard-fails on good B0 parts and frameworks selecting on `is_gfx1250_asm_supported` fall back off asm everywhere; the PR's tests mock exactly this layer, so its own CI cannot catch it. **Author must** confirm the symbol resolution works in the actual gfx1250 deployment image, or fall back through soname/full-path dlopen (`ctypes.CDLL('libamdhip64.so')` works here) and add one unmocked test of the real detection chain.

🔴 [verified] csrc/include/aiter_hip_common.h `require_gfx1250_asm_or_throw` claims it throws "so pybind/ctypes callers get a Python RuntimeError, not a SIGABRT" — a C++ exception crossing a raw ctypes call is uncatchable and terminates the process: reproduced on this box with a hipcc-built extern "C" thrower called via ctypes (exit 134, "terminate called after throwing an instance of 'std::runtime_error'"). The two gates this PR adds in csrc/cpp_itfs/moe/asm_moe.cpp.jinja (FMoeKernel) and csrc/cpp_itfs/mla/asm_mla_decode_fwd.cpp.jinja sit in plain extern "C" functions invoked ffi_type="ctypes" (module_moe_fmoe_asm, module_mla_asm) with no `aiter_safe_call`/AITER_CTYPES bridge — the bridge mechanism exists on main and is used by 17 other asm .cu files, but not these templates; the AiterAsmKernel constructor gate is likewise reached unbridged via asm_mla.cu's raw AITER_C_ITFS entries. Concrete trigger: gfx1250 A0 + a direct call to any public raw binding (aiter.mla_prefill_asm_fwd, aiter.mla_decode_stage1_asm_fwd, aiter.fmoe_g1u1 — exported symbols with no Python guard) kills the serving process instead of raising. The 10 Python-wrapped paths raise cleanly first, so the backstop is exactly what aborts. **Author must** route the throw through the existing aiter_safe_call/AITER_CTYPES bridge in both jinja templates (as the other asm entry points on main already do) or add Python-side guards to the raw bindings.

⚠️ [verified] aiter/mla.py:552 puts `require_gfx1250_asm` as the first statement of `mla_decode_fwd` — the production MLA decode path on gfx942/gfx950 (DSv3/Kimi, per-layer per-step) — and 9 other ops, and the call costs a measured 8.5–15.2 μs on archs where the gate can never fire (arch != gfx1250 short-circuits inside): the cost is the `@torch_compile_guard` custom-op machinery (CPU tensor alloc + torch dispatcher) that `require_gfx1250_asm` routes through on every call, ~60–100× the 0.13 μs cached decision. At 61 layers that is ~0.5–0.9 ms of pure guard overhead per DSv3 decode step on MI300X/MI355X, archs this PR has no intent to change. **Author must** short-circuit on a once-cached arch check (e.g. a module-level `get_gfx_runtime() == "gfx1250"` flag) or call the cached predicate directly instead of through the custom op, so non-gfx1250 archs pay ~0.

⚠️ [verified] The PR is stale against the merge target: the patch no longer applies (fails at aiter/mla.py:1215), and csrc/py_itfs_cu/pa_sparse_prefill_opus_kernels.cu was renamed on main to mla_v4_prefill_opus_kernels.cu in #5228 — so the PrebuiltKernel/SkipGfx1250Gate OPUS opt-out this diff adds lands on a file that no longer exists at that path, and any CI result on the current head describes a moved tree. **Author must** rebase and re-apply the opt-out to the renamed file before merge.

📝 [verified] The new C++ `get_asic_revision` in csrc/include/aiter_hip_common.h has zero callers in the tree and uses HIP_CALL (std::abort on query failure) — dead code that reintroduces the abort-on-query hazard if ever wired into a soft path; and the Python `require_gfx1250_asm` has no counterpart to the C++ exact-name A0 allowlist (`is_gfx1250_asm_a0_ok`), so a kernel added to that allowlist would still be blocked at all 10 Python-guarded ops. **Reviewer should ask** whether to drop the dead helper now and mirror the allowlist before either is ever needed.

valarLip (commented):
Reviewed the full diff. The direction is right — shipping a B0-only code object on A0 silicon should fail loudly rather than fault — but I think a few things need to be resolved before merge. Grouped by severity.

## Blocking

### 1. The premise of the whole gate is unverified

`aiter/jit/utils/chip_info.py:150`

Everything here rests on `asicRevision == silicon stepping` (0 = A0, 1 = B0). I can't find anything in the PR or the tests that confirms this against a real gfx1250 B0 board. On the 8x MI355X box I checked, every gfx950 GPU agent reports `ASIC Revision: 0(0x0)` — so the one board family I can actually read reports 0 on production silicon.

If shipped gfx1250 B0 also reports 0, then `get_asic_revision()` returns 0, `is_gfx1250_asm_supported()` is False, and MLA decode/prefill, `pa_decode_bf16_asm`, every f4/mxfp8 GEMM and every sink FMHA raise or silently fall back to CK/Triton **on healthy hardware**, with no env var to override it.

`op_tests/test_asm_guard.py` only asserts the gate is self-consistent with whatever this box reports, so it can never catch this. Could you add a reading from a real B0 part before merging?

### 2. Raw `throw` bypasses the repo's `g_aiter_can_throw` convention, and the sweep is incomplete

`csrc/include/aiter_hip_common.h:229`

`require_gfx1250_asm_or_throw` uses a bare `throw std::runtime_error`. The existing convention (`check_fail:59`) only throws on bridged paths and otherwise prints and aborts. The PR converts the three `asm_mla.cu` entry points to `AITER_CTYPES_DEFINE_ENTRYPOINT_VOID`, but the other raw `AITER_C_ITFS` entry points that construct an `AiterAsmKernel` were missed:

- `csrc/py_itfs_cu/asm_pa.cu:167` (`pa_fwd`) and `:320` (`pa_ps_fwd`)
- `csrc/py_itfs_cu/asm_topksoftmax.cu:62`
- `csrc/py_itfs_cu/asm_topk_per_row_decode.cu:15`
- `csrc/py_itfs_cu/asm_topk_per_row_prefill.cu:17`
- `csrc/py_itfs_cu/asm_flatmm_a8w8_blockscale.cu:30`

On gfx1250 A0, any of these unwinds an exception out of an `extern "C"` frame into libffi/ctypes — `std::terminate`/SIGABRT, taking down the whole TP rank. That is exactly the outcome the header comment ("THROWS (never `std::abort()`) ... not a SIGABRT") promises to avoid. Using `AITER_CHECK(...)` instead would route through the existing mechanism.

### 3. Throwing from the constructor body calls `__hipUnregisterFatBinary(nullptr)`

`csrc/include/aiter_hip_common.h:293`

`init()` throws inside the `AiterAsmKernel` constructor body, at which point the `AiterAsmKernelFast` base subobject is fully constructed (protected `= default` ctor). Unwinding runs `~AiterAsmKernelFast()`, which unconditionally calls `__hipUnregisterFatBinary(module)` — and `module` is still `nullptr`, because the throw happens before `init_ungated`.

This path is new in this PR: the pre-existing `AITER_CHECK` throw sites all sit after `module` is assigned. So every `static AiterAsmKernel impl(name, path)` construction that throws by design (`asm_mla.cu`, `asm_gemm_a4w4.cu`, `asm_fmha_fwd_with_sink.cu`, `asm_pa_decode_bf16.cu`, ...) hands a null pointer to HIP's fat-binary deregistration instead of producing the intended clean Python `RuntimeError`.

```cpp
if (module) __hipUnregisterFatBinary(module);
```

### 4. Resource leak on the newly-reachable unwind path

`csrc/py_itfs_cu/asm_mla.cu:776`

`AITER_CTYPES_DEFINE_ENTRYPOINT_VOID` wraps the whole function body in `aiter_safe_call`, which sets `g_aiter_can_throw = true` — so every `AITER_CHECK`/`HIP_CALL` in the body, plus the new gate, now unwinds instead of aborting.

The persistent branch holds two unowned raw resources across that point: `new uint64_t[10]` (`:776`) and `hipMalloc(&dev_PS_META_DATA, ...)` (`:782`). I couldn't find a matching `delete[]`/`hipFree` anywhere in the file. The `HipDeviceGuard` above them is RAII; these two are not.

This is a pre-existing leak, but the PR changes its character: it used to be "we're aborting anyway", and now it's a recoverable path. A serving loop that catches and continues, or an autotuner sweeping shapes, leaks 80 bytes host + 80 bytes device per attempt.

Same shape at `asm_topk_per_row_prefill.cu:33` (hipMalloc'd workspace on the throw path). Minor related point: `load_hsaco_file()` (full `.co` read + `validate_hsaco_lds`) runs *before* the gate, so A0 reads the object off disk just to discard it.

## Should fix before merge

### 5. The guard is at the wrong altitude in `mla_decode_fwd`

`aiter/mla.py:552`

`require_gfx1250_asm` is at the very top of `mla_decode_fwd`, but that function is a multi-backend dispatcher. `use_flydsl_ps1` at `:755` (whose condition is literally `get_gfx() == "gfx1250"`) routes to `flydsl_mla_pagesize1_fp8_fp8` at `:913`; there are also `mla_decode_fwd_opus_stage1` and `hk_mla_decode_fwd` branches. Only the final `else` at `:969` reaches asm.

So gfx1250 A0 + fp8 + `page_size == 1` raises at `:552` and never reaches the FlyDSL selection — and FlyDSL loads no shipped asm at all. It also contradicts the C++ side of this same PR, which deliberately exempts OPUS code objects via `SkipGfx1250Gate` on the grounds that arch coverage is the OPUS layer's job; the Python side kills the OPUS branch anyway.

`gemm_op_a4w4.py:195` in this same PR gets this right — the guard is inside the gfx1250 branch. `mla_decode_fwd_v4_nm:1700` has the same issue as `:552`.

### 6. The gate is cached per kernel-name but reads per-device state

`csrc/include/aiter_hip_common.h:250`

The gate runs once per kernel name per process, inside `AiterAsmKernelFast::init()`. What it reads — `hipGetDevice()` + `prop.asicRevision` — is a property of the *current* device. Every call site caches the kernel object process-wide (`SynchronizedCache<std::string, AiterAsmKernel> impl_ptr_map` across 28 TUs, or function-local statics), and `launch_kernel` never rechecks.

On a mixed-stepping node (dev0 = B0, dev1 = A0): call on dev0 first, gate reads rev 1 and passes, kernel is cached; then `hipSetDevice(1)` and call the same op — cache hit, `init()` never runs again, and the B0-only code object launches on A0 silicon with no throw and no warning. Reverse order permanently disables every B0 device in the process.

### 7. Python and C++ gates disagree, with opposite failure directions

`aiter/jit/utils/chip_info.py:161`

Two same-named gates, no shared source of truth:

| | Python | C++ |
|---|---|---|
| scope | `min()` over all same-arch agents node-wide | current HIP device |
| source | `rocminfo` text (docstring notes it ignores `HIP_VISIBLE_DEVICES`) | `hipGetDeviceProperties` |
| on failure | fail-closed | fail-open (returns true) |

Concretely: one A0 experimental card in a chassis of seven B0, worked around with `HIP_VISIBLE_DEVICES=1..7`. C++ says supported; Python still sees the A0 in `rocminfo`, `min() == 0`, and every `require_gfx1250_asm` throws and all four `can_impl_*` in `mha.py` return False — asm disabled on seven good cards, with no override. In the other direction, on a mixed-arch node where the first agent isn't gfx1250, `_detect_native()` takes the first gfx string, `_ARCH_IS_GFX1250` is False, and the Python gate silently disappears.

Suggestion: `_get_pci_chip_id()` (`chip_info.py:504`) already does the `hipDeviceGetAttribute` ctypes dance. Reusing that pattern would make the Python side per-HIP-device, matching C++, and drop the rocminfo text dependency entirely.

### 8. Refactor smuggles in a behavior change affecting every arch

`aiter/jit/utils/chip_info.py:22`

The new shared `_rocminfo_output()` uses `check=True`. The call site it replaced in `get_cu_num_custom_op` deliberately used `check=False` — `rocminfo` commonly prints a complete agent list and still exits non-zero (KFD/topology warnings, a single failed agent query, `/dev/kfd` permissions). That tolerance isn't reconstructed anywhere.

On such a node, `get_cu_num()` used to return the CU count and now raises `RuntimeError`, taking `get_build_targets()`, every cu_num-keyed GEMM/MoE tuned-config dispatch, and JIT builds down with it — on machines that worked before, on every arch, unrelated to gfx1250.

Also: `lru_cache` doesn't cache exceptions, so on a box without `rocminfo` every call re-forks (0.22–0.34 s measured), and `get_cu_num_custom_op` itself isn't cached (only the outer `get_cu_num` is).

### 9. The rocminfo parser is brittle, and its failure mode is the one the comment says to avoid

`aiter/jit/utils/chip_info.py:137`

- `rev = int(value.split("(")[0])` is unguarded — `N/A`, empty, or non-numeric raises `ValueError` and takes down the whole parser
- the key regex hardcodes exactly two leading spaces
- the append condition assumes `Device Type:` always prints before `ASIC Revision:`; a reordering silently drops the agent rather than erroring
- `\s+` on a present-but-empty field yields a phantom `("", rev)` agent

Any rocminfo version that changes indentation, field order, or prints `N/A` makes `_rocminfo_gpu_agents()` return `[]` or raise → `get_asic_revision()` raises → `_gfx1250_stepping_ok()` swallows it and fails closed → every gfx1250 asm path disabled on a healthy B0 machine, with a single `logger.warning` as the only symptom.

That is precisely what the comment at `:167` says must not happen ("A bogus parse must not read as A0 and disable asm on good silicon"). Meanwhile C++ reports supported, so the two gates disagree and whichever runs first wins.

### 10. `@torch_compile_guard()` on a zero-arg bool is a ~190x regression and defeats its own stated purpose

`aiter/jit/utils/asm_guard.py:41`

`torch_guard.py` takes the `not sig.parameters` branch and registers `aiter::is_gfx1250_asm_supported(Tensor dummy) -> (Tensor, bool)`. Every call allocates two `torch.empty(1, cpu)` and goes through the dispatcher.

Measured: **2.66 us/call decorated vs 0.014 us undecorated**.

The comment at `:22-23` states the design goal is "a constant lets `torch.compile(fullgraph=True)` trace through" — the decorator does the opposite, turning a constant into an opaque custom op. Under `fullgraph=True`, `ret = ... and is_gfx1250_asm_supported()` becomes a data-dependent branch on an opaque op's output: graph break or hard error.

On gfx1250 the four per-call predicates in `mha.py` (`:1946`, `:1982`, `:3004`, `:3737`) pay 2.7–5.3 us of pure CPU per attention — roughly 0.3 ms/step at 60 layers, invisible in a kernel profile. Note the comparable `get_gfx()`/`get_cu_num()` in `chip_info.py` wrap the custom op in `functools.lru_cache`; this one doesn't. Dropping the decorator fixes it.

### 11. Mixing two different arch sources in one boolean

`aiter/ops/mha.py:1946` (same at `:1982`, `:3004`, `:3737`)

`get_gfx() == "gfx1250" and is_gfx1250_asm_supported()` combines `get_gfx()` (`chip_info.py:44` — honors `GPU_ARCHS`, takes the last entry when multiple) with `is_gfx1250_asm_supported()` (internally `get_gfx_runtime()`, which explicitly ignores `GPU_ARCHS`).

With the common container setting `GPU_ARCHS=gfx1250` on a gfx950 machine: the first half passes because `get_gfx()` says gfx1250, and the second half passes because the runtime is gfx950 (not A0 gfx1250) so it returns True. `can_impl_fmha_fwd_with_sink_asm` then claims the gfx1250 asm backend is available on a gfx950 device — the new guard is inert exactly where it's needed most. Conversely `GPU_ARCHS=gfx942;gfx950` on real B0 gfx1250 never selects the path.

Using `get_gfx_runtime()` on both halves removes the discrepancy.

### 12. Fix-then-sweep misses

`aiter/ops/gemm_op_a4w4.py:489`

`gemm_a4w4o8` is a gfx1250-only asm entry point (`assert get_gfx() in _GFX1250` in the body, then an unconditional `_f4gemm_asm_dispatch` — the same call its two guarded siblings make). `gemm_mxfp4_asm` (`:398`) and `gemm_nvfp4_asm` (`:432`) both got guards; this one didn't. On A0 it skips the Python layer and throws from the C++ load site with a lower-level message instead of the PR's unified one — and its op name can never appear in `_A0_ALLOWLIST`, so the future escape hatch is structurally unavailable to it.

Also unguarded: `aiter/ops/pa_sparse_prefill_opus.py` `pa_sparse_prefill_gfx1250_opus_fwd` / `_fp8_` variants — no Python guard, and the C++ side exempts them via `SkipGfx1250Gate`, so neither layer checks. If those `.co` files contain B0-only encodings, A0 gets a memory fault or silent wrong values rather than a clean error.

### 13. The capability probe doesn't exist in three supported configurations

`aiter/__init__.py:149`

The docstring says "Frameworks can call this at startup to select a backend", but both names are defined inside the `else` arm of the `if AITER_TRITON_ONLY: ... elif AITER_AOT_IMPORT: ... else:` chain starting at `:73`. So under `AITER_TRITON_ONLY=1`, `AITER_AOT_IMPORT=1`, or `sys.platform == 'win32'` (`:64` forces Triton-only unconditionally), they're absent.

vLLM/SGLang writing `if aiter.is_gfx1250_asm_supported():` get an `AttributeError` instead of `False` — a capability probe turning into a startup crash, forcing callers into `hasattr`. Hoisting the two imports out of the chain (or defining them unconditionally at module level) fixes it.

### 14. Both allowlists are inert, and their keys don't agree

`aiter/jit/utils/asm_guard.py:11`

Python `_A0_ALLOWLIST` is keyed by op name; C++ `kA0AllowList` (`:217`) by exact kernel symbol name. They're declared "keep in sync" by comment, but they aren't in the same namespace (one op fans out to several kernels), both are empty, and nothing validates either. `kA0AllowList` contains only a nullptr sentinel, so its loop provably never executes and `|| is_gfx1250_asm_a0_ok(...)` is identically false.

The Python call sites don't even agree with themselves: `gemm_*` passes a Python op name, while `mla.py` passes a *different function's* C entry name (`mla_decode_fwd` → `"mla_decode_stage1_asm_fwd"`, `mla_decode_fwd_v4_nm` → `"mla_decode_v4_asm"`).

So the first real exemption is near-certain to be written wrong: C++ only → the Python guard throws first and the entry is dead code; Python only → it passes and then C++ throws from deeper, with a message *less* comprehensible than the guard's; and the intuitive key (the Python function name, or the kernel name from the `.co`) matches neither call site. `op_tests/test_asm_guard.py::test_a0_allowlist_is_in_sync` only asserts `isinstance(..., frozenset)`, so the name oversells it. A0 users also see symbol names that don't appear anywhere in their own code.

### 15. Unconditional device query on the gfx942/gfx950 backward hot path

`csrc/cpp_itfs/mha_bwd.cu:403`

`(!is_gfx1250_asm_supported())` is appended as the last term of an `||` elimination chain. `||` only short-circuits on a *true* term, so any genuinely asm-eligible path (all earlier terms false) always evaluates it. Inside: an uncached `hipGetDevice` + `hipGetDeviceProperties` plus two `std::string` heap allocations — while `:399` in the same function just called `get_gpu_arch()`, which does the same thing. (There are already five gcnArchName-normalizing copies under `csrc/include/`: `get_gpu_arch`, `get_device_gfx`, `opus_get_arch_info`, `isGPUArch`; the last two are cached.)

Measured 0.108 us on ROCm 7.1.1 — tolerable — but `hipGetDeviceProperties` is tens of us on some ROCm 6.x, which would make this a measurable training-hot-path regression from a gate that has nothing to do with those two arches:

```cpp
(arch_id == "gfx1250" && !is_gfx1250_asm_supported())
```

Separately, the Python `can_impl_fmha_v3_bwd*` still only accept gfx942/gfx950 and there's no gfx1250 bwd entry, so this term is currently unreachable from Python. And under an `ONLY_FAV3` build, `-1` is no longer a fallback signal — it surfaces at `asm_mha_bwd.cu:326` as a misleading "invalid argument for fmha_v3_bwd".

## Architecture

The recurring cost here is that "this code object needs arch X stepping >= N" is expressed as ~20 hardcoded `"gfx1250"` + `>= 1` sites across two languages. That fact belongs on the code object as metadata — `hsa/<arch>/*.csv` already carries `knl_name`/`co_name` columns, and `validate_hsaco_lds` in this same class is exactly that shape. One declaration per `.co`, one check at load, and the allowlist problem above dissolves along with it.

There's also no env escape hatch (the repo already has the `AITER_ENABLE_EXPERIMENTAL` convention). On an A0 board, CI turns into a wall of failures instead of skips.

Smaller, related: the `init`/`init_ungated` split + `SkipGfx1250Gate` tag + `using` re-export + derived `struct PrebuiltKernel` are five moving parts expressing one boolean; a defaulted parameter would do.

## Nits

- `asm_mla.cu`: the new `#include "aiter_ctypes_error.h"` isn't `.clang-format` clean (`SortIncludes: true`). `pa_sparse_block_select_entry.cu` documents that this header must come first, separated by a blank line, or a mechanical reformat becomes a compile error. `aiter_hip_common.h:200-201` and `:233-235` aren't clang-format clean either, and `.githooks/pre-commit` covers `.h`.
- `docs/isa_kernel_optimization.md:79-80` still describes `AiterAsmKernel` as loading unconditionally — no mention of the gate, allowlist, or opt-out.
- Both new files use `Copyright (C) 2024-2026`; the repo convention for files created in 2026 is a single year (114 occurrences of `Copyright (C) 2026`).
- `gemm_op_a4w4.py`: the guard replaced `# F4GEMM is kept on a separate dispatch (preload kargs layout).`, which was the only record of why that branch exists.
- `aiter/mla.py`: an unrelated blank line was removed in `_mla_prefill_reduce_kernel`.

## What I verified

- `ruff` and `black` pass on all changed Python files; the new `aiter/__init__.py` imports are inside an `# isort: off` region, so isort doesn't apply.
- All three `AITER_CTYPES_DEFINE_ENTRYPOINT_VOID` parameter lists and forwarding lists check out item by item — 27/18/11 parameters, order and count all match.
- The new rocminfo parser runs correctly on an 8x MI355X box: CPU agents correctly skipped, `Marketing Name` not mismatched. The brittleness in #9 is about other rocminfo versions, not this one.

dbyoung18:
> Reviewed the full diff. The direction is right — shipping a B0-only code object on A0 silicon should fail loudly rather than fault — but I think a few things need to be resolved before merge. Grouped by severity.
> 
> ## Blocking
> ### 1. The premise of the whole gate is unverified
> `aiter/jit/utils/chip_info.py:150`
> 
> Everything here rests on `asicRevision == silicon stepping` (0 = A0, 1 = B0). I can't find anything in the PR or the tests that confirms this against a real gfx1250 B0 board. On the 8x MI355X box I checked, every gfx950 GPU agent reports `ASIC Revision: 0(0x0)` — so the one board family I can actually read reports 0 on production silicon.
> 
> If shipped gfx1250 B0 also reports 0, then `get_asic_revision()` returns 0, `is_gfx1250_asm_supported()` is False, and MLA decode/prefill, `pa_decode_bf16_asm`, every f4/mxfp8 GEMM and every sink FMHA raise or silently fall back to CK/Triton **on healthy hardware**, with no env var to override it.
> 
> `op_tests/test_asm_guard.py` only asserts the gate is self-consistent with whatever this box reports, so it can never catch this. Could you add a reading from a real B0 part before merging?
> 
> ### 2. Raw `throw` bypasses the repo's `g_aiter_can_throw` convention, and the sweep is incomplete
> `csrc/include/aiter_hip_common.h:229`
> 
> `require_gfx1250_asm_or_throw` uses a bare `throw std::runtime_error`. The existing convention (`check_fail:59`) only throws on bridged paths and otherwise prints and aborts. The PR converts the three `asm_mla.cu` entry points to `AITER_CTYPES_DEFINE_ENTRYPOINT_VOID`, but the other raw `AITER_C_ITFS` entry points that construct an `AiterAsmKernel` were missed:
> 
> * `csrc/py_itfs_cu/asm_pa.cu:167` (`pa_fwd`) and `:320` (`pa_ps_fwd`)
> * `csrc/py_itfs_cu/asm_topksoftmax.cu:62`
> * `csrc/py_itfs_cu/asm_topk_per_row_decode.cu:15`
> * `csrc/py_itfs_cu/asm_topk_per_row_prefill.cu:17`
> * `csrc/py_itfs_cu/asm_flatmm_a8w8_blockscale.cu:30`
> 
> On gfx1250 A0, any of these unwinds an exception out of an `extern "C"` frame into libffi/ctypes — `std::terminate`/SIGABRT, taking down the whole TP rank. That is exactly the outcome the header comment ("THROWS (never `std::abort()`) ... not a SIGABRT") promises to avoid. Using `AITER_CHECK(...)` instead would route through the existing mechanism.
> 
> ### 3. Throwing from the constructor body calls `__hipUnregisterFatBinary(nullptr)`
> `csrc/include/aiter_hip_common.h:293`
> 
> `init()` throws inside the `AiterAsmKernel` constructor body, at which point the `AiterAsmKernelFast` base subobject is fully constructed (protected `= default` ctor). Unwinding runs `~AiterAsmKernelFast()`, which unconditionally calls `__hipUnregisterFatBinary(module)` — and `module` is still `nullptr`, because the throw happens before `init_ungated`.
> 
> This path is new in this PR: the pre-existing `AITER_CHECK` throw sites all sit after `module` is assigned. So every `static AiterAsmKernel impl(name, path)` construction that throws by design (`asm_mla.cu`, `asm_gemm_a4w4.cu`, `asm_fmha_fwd_with_sink.cu`, `asm_pa_decode_bf16.cu`, ...) hands a null pointer to HIP's fat-binary deregistration instead of producing the intended clean Python `RuntimeError`.
> 
> ```c++
> if (module) __hipUnregisterFatBinary(module);
> ```
> 
> ### 4. Resource leak on the newly-reachable unwind path
> `csrc/py_itfs_cu/asm_mla.cu:776`
> 
> `AITER_CTYPES_DEFINE_ENTRYPOINT_VOID` wraps the whole function body in `aiter_safe_call`, which sets `g_aiter_can_throw = true` — so every `AITER_CHECK`/`HIP_CALL` in the body, plus the new gate, now unwinds instead of aborting.
> 
> The persistent branch holds two unowned raw resources across that point: `new uint64_t[10]` (`:776`) and `hipMalloc(&dev_PS_META_DATA, ...)` (`:782`). I couldn't find a matching `delete[]`/`hipFree` anywhere in the file. The `HipDeviceGuard` above them is RAII; these two are not.
> 
> This is a pre-existing leak, but the PR changes its character: it used to be "we're aborting anyway", and now it's a recoverable path. A serving loop that catches and continues, or an autotuner sweeping shapes, leaks 80 bytes host + 80 bytes device per attempt.
> 
> Same shape at `asm_topk_per_row_prefill.cu:33` (hipMalloc'd workspace on the throw path). Minor related point: `load_hsaco_file()` (full `.co` read + `validate_hsaco_lds`) runs _before_ the gate, so A0 reads the object off disk just to discard it.
> 
> ## Should fix before merge
> ### 5. The guard is at the wrong altitude in `mla_decode_fwd`
> `aiter/mla.py:552`
> 
> `require_gfx1250_asm` is at the very top of `mla_decode_fwd`, but that function is a multi-backend dispatcher. `use_flydsl_ps1` at `:755` (whose condition is literally `get_gfx() == "gfx1250"`) routes to `flydsl_mla_pagesize1_fp8_fp8` at `:913`; there are also `mla_decode_fwd_opus_stage1` and `hk_mla_decode_fwd` branches. Only the final `else` at `:969` reaches asm.
> 
> So gfx1250 A0 + fp8 + `page_size == 1` raises at `:552` and never reaches the FlyDSL selection — and FlyDSL loads no shipped asm at all. It also contradicts the C++ side of this same PR, which deliberately exempts OPUS code objects via `SkipGfx1250Gate` on the grounds that arch coverage is the OPUS layer's job; the Python side kills the OPUS branch anyway.
> 
> `gemm_op_a4w4.py:195` in this same PR gets this right — the guard is inside the gfx1250 branch. `mla_decode_fwd_v4_nm:1700` has the same issue as `:552`.
> 
> ### 6. The gate is cached per kernel-name but reads per-device state
> `csrc/include/aiter_hip_common.h:250`
> 
> The gate runs once per kernel name per process, inside `AiterAsmKernelFast::init()`. What it reads — `hipGetDevice()` + `prop.asicRevision` — is a property of the _current_ device. Every call site caches the kernel object process-wide (`SynchronizedCache<std::string, AiterAsmKernel> impl_ptr_map` across 28 TUs, or function-local statics), and `launch_kernel` never rechecks.
> 
> On a mixed-stepping node (dev0 = B0, dev1 = A0): call on dev0 first, gate reads rev 1 and passes, kernel is cached; then `hipSetDevice(1)` and call the same op — cache hit, `init()` never runs again, and the B0-only code object launches on A0 silicon with no throw and no warning. Reverse order permanently disables every B0 device in the process.
> 
> ### 7. Python and C++ gates disagree, with opposite failure directions
> `aiter/jit/utils/chip_info.py:161`
> 
> Two same-named gates, no shared source of truth:
> 
> Python	C++
> scope	`min()` over all same-arch agents node-wide	current HIP device
> source	`rocminfo` text (docstring notes it ignores `HIP_VISIBLE_DEVICES`)	`hipGetDeviceProperties`
> on failure	fail-closed	fail-open (returns true)
> Concretely: one A0 experimental card in a chassis of seven B0, worked around with `HIP_VISIBLE_DEVICES=1..7`. C++ says supported; Python still sees the A0 in `rocminfo`, `min() == 0`, and every `require_gfx1250_asm` throws and all four `can_impl_*` in `mha.py` return False — asm disabled on seven good cards, with no override. In the other direction, on a mixed-arch node where the first agent isn't gfx1250, `_detect_native()` takes the first gfx string, `_ARCH_IS_GFX1250` is False, and the Python gate silently disappears.
> 
> Suggestion: `_get_pci_chip_id()` (`chip_info.py:504`) already does the `hipDeviceGetAttribute` ctypes dance. Reusing that pattern would make the Python side per-HIP-device, matching C++, and drop the rocminfo text dependency entirely.
> 
> ### 8. Refactor smuggles in a behavior change affecting every arch
> `aiter/jit/utils/chip_info.py:22`
> 
> The new shared `_rocminfo_output()` uses `check=True`. The call site it replaced in `get_cu_num_custom_op` deliberately used `check=False` — `rocminfo` commonly prints a complete agent list and still exits non-zero (KFD/topology warnings, a single failed agent query, `/dev/kfd` permissions). That tolerance isn't reconstructed anywhere.
> 
> On such a node, `get_cu_num()` used to return the CU count and now raises `RuntimeError`, taking `get_build_targets()`, every cu_num-keyed GEMM/MoE tuned-config dispatch, and JIT builds down with it — on machines that worked before, on every arch, unrelated to gfx1250.
> 
> Also: `lru_cache` doesn't cache exceptions, so on a box without `rocminfo` every call re-forks (0.22–0.34 s measured), and `get_cu_num_custom_op` itself isn't cached (only the outer `get_cu_num` is).
> 
> ### 9. The rocminfo parser is brittle, and its failure mode is the one the comment says to avoid
> `aiter/jit/utils/chip_info.py:137`
> 
> * `rev = int(value.split("(")[0])` is unguarded — `N/A`, empty, or non-numeric raises `ValueError` and takes down the whole parser
> * the key regex hardcodes exactly two leading spaces
> * the append condition assumes `Device Type:` always prints before `ASIC Revision:`; a reordering silently drops the agent rather than erroring
> * `\s+` on a present-but-empty field yields a phantom `("", rev)` agent
> 
> Any rocminfo version that changes indentation, field order, or prints `N/A` makes `_rocminfo_gpu_agents()` return `[]` or raise → `get_asic_revision()` raises → `_gfx1250_stepping_ok()` swallows it and fails closed → every gfx1250 asm path disabled on a healthy B0 machine, with a single `logger.warning` as the only symptom.
> 
> That is precisely what the comment at `:167` says must not happen ("A bogus parse must not read as A0 and disable asm on good silicon"). Meanwhile C++ reports supported, so the two gates disagree and whichever runs first wins.
> 
> ### 10. `@torch_compile_guard()` on a zero-arg bool is a ~190x regression and defeats its own stated purpose
> `aiter/jit/utils/asm_guard.py:41`
> 
> `torch_guard.py` takes the `not sig.parameters` branch and registers `aiter::is_gfx1250_asm_supported(Tensor dummy) -> (Tensor, bool)`. Every call allocates two `torch.empty(1, cpu)` and goes through the dispatcher.
> 
> Measured: **2.66 us/call decorated vs 0.014 us undecorated**.
> 
> The comment at `:22-23` states the design goal is "a constant lets `torch.compile(fullgraph=True)` trace through" — the decorator does the opposite, turning a constant into an opaque custom op. Under `fullgraph=True`, `ret = ... and is_gfx1250_asm_supported()` becomes a data-dependent branch on an opaque op's output: graph break or hard error.
> 
> On gfx1250 the four per-call predicates in `mha.py` (`:1946`, `:1982`, `:3004`, `:3737`) pay 2.7–5.3 us of pure CPU per attention — roughly 0.3 ms/step at 60 layers, invisible in a kernel profile. Note the comparable `get_gfx()`/`get_cu_num()` in `chip_info.py` wrap the custom op in `functools.lru_cache`; this one doesn't. Dropping the decorator fixes it.
> 
> ### 11. Mixing two different arch sources in one boolean
> `aiter/ops/mha.py:1946` (same at `:1982`, `:3004`, `:3737`)
> 
> `get_gfx() == "gfx1250" and is_gfx1250_asm_supported()` combines `get_gfx()` (`chip_info.py:44` — honors `GPU_ARCHS`, takes the last entry when multiple) with `is_gfx1250_asm_supported()` (internally `get_gfx_runtime()`, which explicitly ignores `GPU_ARCHS`).
> 
> With the common container setting `GPU_ARCHS=gfx1250` on a gfx950 machine: the first half passes because `get_gfx()` says gfx1250, and the second half passes because the runtime is gfx950 (not A0 gfx1250) so it returns True. `can_impl_fmha_fwd_with_sink_asm` then claims the gfx1250 asm backend is available on a gfx950 device — the new guard is inert exactly where it's needed most. Conversely `GPU_ARCHS=gfx942;gfx950` on real B0 gfx1250 never selects the path.
> 
> Using `get_gfx_runtime()` on both halves removes the discrepancy.
> 
> ### 12. Fix-then-sweep misses
> `aiter/ops/gemm_op_a4w4.py:489`
> 
> `gemm_a4w4o8` is a gfx1250-only asm entry point (`assert get_gfx() in _GFX1250` in the body, then an unconditional `_f4gemm_asm_dispatch` — the same call its two guarded siblings make). `gemm_mxfp4_asm` (`:398`) and `gemm_nvfp4_asm` (`:432`) both got guards; this one didn't. On A0 it skips the Python layer and throws from the C++ load site with a lower-level message instead of the PR's unified one — and its op name can never appear in `_A0_ALLOWLIST`, so the future escape hatch is structurally unavailable to it.
> 
> Also unguarded: `aiter/ops/pa_sparse_prefill_opus.py` `pa_sparse_prefill_gfx1250_opus_fwd` / `_fp8_` variants — no Python guard, and the C++ side exempts them via `SkipGfx1250Gate`, so neither layer checks. If those `.co` files contain B0-only encodings, A0 gets a memory fault or silent wrong values rather than a clean error.
> 
> ### 13. The capability probe doesn't exist in three supported configurations
> `aiter/__init__.py:149`
> 
> The docstring says "Frameworks can call this at startup to select a backend", but both names are defined inside the `else` arm of the `if AITER_TRITON_ONLY: ... elif AITER_AOT_IMPORT: ... else:` chain starting at `:73`. So under `AITER_TRITON_ONLY=1`, `AITER_AOT_IMPORT=1`, or `sys.platform == 'win32'` (`:64` forces Triton-only unconditionally), they're absent.
> 
> vLLM/SGLang writing `if aiter.is_gfx1250_asm_supported():` get an `AttributeError` instead of `False` — a capability probe turning into a startup crash, forcing callers into `hasattr`. Hoisting the two imports out of the chain (or defining them unconditionally at module level) fixes it.
> 
> ### 14. Both allowlists are inert, and their keys don't agree
> `aiter/jit/utils/asm_guard.py:11`
> 
> Python `_A0_ALLOWLIST` is keyed by op name; C++ `kA0AllowList` (`:217`) by exact kernel symbol name. They're declared "keep in sync" by comment, but they aren't in the same namespace (one op fans out to several kernels), both are empty, and nothing validates either. `kA0AllowList` contains only a nullptr sentinel, so its loop provably never executes and `|| is_gfx1250_asm_a0_ok(...)` is identically false.
> 
> The Python call sites don't even agree with themselves: `gemm_*` passes a Python op name, while `mla.py` passes a _different function's_ C entry name (`mla_decode_fwd` → `"mla_decode_stage1_asm_fwd"`, `mla_decode_fwd_v4_nm` → `"mla_decode_v4_asm"`).
> 
> So the first real exemption is near-certain to be written wrong: C++ only → the Python guard throws first and the entry is dead code; Python only → it passes and then C++ throws from deeper, with a message _less_ comprehensible than the guard's; and the intuitive key (the Python function name, or the kernel name from the `.co`) matches neither call site. `op_tests/test_asm_guard.py::test_a0_allowlist_is_in_sync` only asserts `isinstance(..., frozenset)`, so the name oversells it. A0 users also see symbol names that don't appear anywhere in their own code.
> 
> ### 15. Unconditional device query on the gfx942/gfx950 backward hot path
> `csrc/cpp_itfs/mha_bwd.cu:403`
> 
> `(!is_gfx1250_asm_supported())` is appended as the last term of an `||` elimination chain. `||` only short-circuits on a _true_ term, so any genuinely asm-eligible path (all earlier terms false) always evaluates it. Inside: an uncached `hipGetDevice` + `hipGetDeviceProperties` plus two `std::string` heap allocations — while `:399` in the same function just called `get_gpu_arch()`, which does the same thing. (There are already five gcnArchName-normalizing copies under `csrc/include/`: `get_gpu_arch`, `get_device_gfx`, `opus_get_arch_info`, `isGPUArch`; the last two are cached.)
> 
> Measured 0.108 us on ROCm 7.1.1 — tolerable — but `hipGetDeviceProperties` is tens of us on some ROCm 6.x, which would make this a measurable training-hot-path regression from a gate that has nothing to do with those two arches:
> 
> ```c++
> (arch_id == "gfx1250" && !is_gfx1250_asm_supported())
> ```
> 
> Separately, the Python `can_impl_fmha_v3_bwd*` still only accept gfx942/gfx950 and there's no gfx1250 bwd entry, so this term is currently unreachable from Python. And under an `ONLY_FAV3` build, `-1` is no longer a fallback signal — it surfaces at `asm_mha_bwd.cu:326` as a misleading "invalid argument for fmha_v3_bwd".
> 
> ## Architecture
> The recurring cost here is that "this code object needs arch X stepping >= N" is expressed as ~20 hardcoded `"gfx1250"` + `>= 1` sites across two languages. That fact belongs on the code object as metadata — `hsa/<arch>/*.csv` already carries `knl_name`/`co_name` columns, and `validate_hsaco_lds` in this same class is exactly that shape. One declaration per `.co`, one check at load, and the allowlist problem above dissolves along with it.
> 
> There's also no env escape hatch (the repo already has the `AITER_ENABLE_EXPERIMENTAL` convention). On an A0 board, CI turns into a wall of failures instead of skips.
> 
> Smaller, related: the `init`/`init_ungated` split + `SkipGfx1250Gate` tag + `using` re-export + derived `struct PrebuiltKernel` are five moving parts expressing one boolean; a defaulted parameter would do.
> 
> ## Nits
> * `asm_mla.cu`: the new `#include "aiter_ctypes_error.h"` isn't `.clang-format` clean (`SortIncludes: true`). `pa_sparse_block_select_entry.cu` documents that this header must come first, separated by a blank line, or a mechanical reformat becomes a compile error. `aiter_hip_common.h:200-201` and `:233-235` aren't clang-format clean either, and `.githooks/pre-commit` covers `.h`.
> * `docs/isa_kernel_optimization.md:79-80` still describes `AiterAsmKernel` as loading unconditionally — no mention of the gate, allowlist, or opt-out.
> * Both new files use `Copyright (C) 2024-2026`; the repo convention for files created in 2026 is a single year (114 occurrences of `Copyright (C) 2026`).
> * `gemm_op_a4w4.py`: the guard replaced `# F4GEMM is kept on a separate dispatch (preload kargs layout).`, which was the only record of why that branch exists.
> * `aiter/mla.py`: an unrelated blank line was removed in `_mla_prefill_reduce_kernel`.
> 
> ## What I verified
> * `ruff` and `black` pass on all changed Python files; the new `aiter/__init__.py` imports are inside an `# isort: off` region, so isort doesn't apply.
> * All three `AITER_CTYPES_DEFINE_ENTRYPOINT_VOID` parameter lists and forwarding lists check out item by item — 27/18/11 parameters, order and count all match.
> * The new rocminfo parser runs correctly on an 8x MI355X box: CPU agents correctly skipped, `Marketing Name` not mismatched. The brittleness in #9 is about other rocminfo versions, not this one.

Thanks — went through all of it. Fixed, and the gate is now verified on real silicon in both directions: A0 board reports ASIC Revision 0 and is correctly blocked, B0 reports 1 and runs the shipped asm. So asicRevision == stepping is no longer an assumption.

Fixed
- aiter/__init__.py: exports moved out of the else: arm — defined in all modes (Triton-only/Windows get a no-op stub), so frameworks get True/no-op instead of AttributeError.
- gemm_a4w4o8: guard added.
- ~AiterAsmKernelFast(): now null-checks module — you're right that a throw from the derived ctor body still runs the base dtor.
- chip_info: _rocminfo_output() back to check=False; that behaviour change was unintentional.
- @torch_compile_guard() dropped: the probe is folded into an import-time constant, so the function returns a plain bool that dynamo constant-folds — 2.449 µs → 0.015 µs, no opaque graph node.
- mla.py: guard moved off the dispatcher down to the two actual mla_decode_stage1_asm_fwd call sites, so FlyDSL/OPUS/hk paths are no longer blocked. The other three guards sit in functions with no non-asm backend branch.
- C++ gate: now scans all visible devices, takes the lowest stepping, cached once — same node-wide semantics as the Python side, so the per-kernel-name caching is no longer device-dependent. Both sides now fail the same way (gfx1250 with unreadable revision → reject; unreadable arch → allow) and the comments cross-reference each other.
- Allowlist test renamed to test_a0_allowlists_are_both_empty and now actually asserts both sides are empty (parses kA0AllowList from the header), so adding to one side alone fails.

Not changed, with reasons
- Unbridged extern "C" entries: I mapped every gfx1250-touching .cu to its entry macro — all 9 are AITER_CTYPES_DEFINE_ENTRYPOINT_VOID. The files listed (asm_pa.cu, asm_topksoftmax.cu, asm_flatmm_a8w8_blockscale.cu, asm_topk_per_row_*.cu) are gfx9-only and have no hsa/gfx1250/ code object, so the gate returns early there. Also deliberately not AITER_CHECK: g_aiter_can_throw is false on the pybind path, which would downgrade a catchable exception to an abort. Added a std::cerr line before the throw as a backstop.
- asm_mla.cu new uint64_t[10] / hipMalloc: real, but pre-existing and leaks on the success path too — and dev_PS_META_DATA outlives the call, so it needs a stream sync or a cached buffer. Separate PR.
- stepping metadata on the .co: agree with the direction — it touches the csv schema, codegen.py and the load path, so I'd rather do it as a follow-up issue than grow this PR.

Verified on gfx1250: test_asm_guard 6/6, test_f4gemm bf16+fp8, test_mxfp8fp4gemm; black + ruff check . clean.

valarLip (approved):

PR #5111, opened 2026-08-29
[dist] reuse comm groups when possible for memory saving

gbyu-amd:
## Motivation
The comm group reuse was originally done in https://github.com/ROCm/ATOM/pull/2050. Now implement it in aiter side to make it safer.

The purpose is to reuse comm buffers for different comm groups when they share the same ranks and collectives are issued in sequence. This can reduce considerable non-torch memory allocation, therefore leaving more capacity to KV cache.

Before:
```bash
[atom 21:34:21] Memory budget: total_gpu=287.98GB, free=63.91GB, utilization=0.86, budget=247.67GB, peak_torch=192.92GB, non_torch=30.21GB, cudagraph_est=0.40GB, safety=5.76GB, available_for_kv=18.37GB, block_bytes=1769472, num_kvcache_blocks=6832
```

By reusing the comm groups with this PR, non-torch memory now only takes 13GB out of the budget, almost doubles the kv cache:
After:
```bash
[atom 04:36:43] Memory budget: total_gpu=287.98GB, free=80.78GB, utilization=0.86, budget=247.67GB, peak_torch=192.91GB, non_torch=13.40GB, cudagraph_est=0.40GB, safety=5.76GB, available_for_kv=35.20GB, block_bytes=2138112, num_kvcache_blocks=13727
```

## Submission Checklist


aiter/dist/parallel_state.py
```diff
@@ -1843,21 +1909,60 @@ def initialize_model_parallel(
...
+    # Dedup helper: a group whose rank set matches an already-built one reuses it
+    # (reuse_from). The first builder is the source. Only multi-rank groups are
+    # tracked (single-rank groups hold no communicator). No-op when the flag is off.
+    _local_rank = get_world_group().local_rank
+    _built_by_ranks: dict[tuple[int, ...], GroupCoordinator] = {}
+
+    def _build_group(group_name, group_ranks, use_message_queue_broadcaster=False):
+        group_ranks = [x.tolist() for x in group_ranks]
+        my_ranks = next((r for r in group_ranks if rank in r), None)
+        key = tuple(sorted(my_ranks)) if my_ranks is not None else None
```
TennyWang1223: Reuse keys on the sorted rank set but copies the source's rank order

  _build_group keys dedup on tuple(sorted(my_ranks)), while the reuse branch copies the source's order:

  self.ranks = reuse_from.ranks
  self.rank_in_group = reuse_from.rank_in_group

  So two groups with the same rank set but different order collapse to one key, and the later group inherits the source's ranks/rank_in_group. Fine for all-reduce, but wrong for order-dependent ops (all_gather layout, reduce_scatter, send/recv neighbors).

  Today all groups come from single-axis slices/transposes of all_ranks, so every rank list is ascending and sorted(x) == x — it doesn't trigger. But it's an implicit invariant a future group could break silently.

  Fix (parallel_state.py:1939): key on the ordered tuple so reuse only happens when the order is truly identical:

  key = tuple(my_ranks) if my_ranks is not None else None

TennyWang1223 (changes requested):

aiter/dist/parallel_state.py
gbyu-amd: thanks for catching this. Updated the key in commit https://github.com/ROCm/aiter/pull/5111/changes/e62ccb37c7be8a6630fd587326ca86832e7b7dc9

valarLip (changes requested):
## Review

The memory saving is worth having and the dedup logic itself is correct — I ran the three topologies the new test covers against the PR branch and the reuse graph is exactly right in each:

```
all_alias     tp[0,1,2,3]  <- dcp, ep
dp_attention  dp[0,1,2,3]  <- ep            (tp is a singleton; source is dp, not tp)
partial       tp[0,1] <- dcp;  dp[0,2];  ep[0,1,2,3] independent
```

The findings below are about what happens around that core: fields the reuse path inherits that it shouldn't, a guard that can't fire, and a test suite that doesn't execute the code it's guarding. Everything is verified against `fc1d2cf` unless marked otherwise.

### Blocking

**1. `aiter/dist/parallel_state.py:580` — the reuse path ignores the group's own `use_message_queue_broadcaster=False`.**

`self.mq_broadcaster = reuse_from.mq_broadcaster` is unconditional. TP is the first `_build_group` call and the only one passing `use_message_queue_broadcaster=True` (line 1971), and line 640 builds a real `MessageQueue` when `world_size > 1`. DCP and EP pass `False` — and get TP's queue anyway.

The consequence is that `broadcast_object` (line ~1212) switches transport for those groups: instead of `torch.distributed.broadcast_object_list` it takes the mq path, which hard-asserts `src == 0`. So a group that explicitly opted out of the broadcaster now has one, and `get_ep_group().broadcast_tensor_dict(td, src=N)` for `N != 0` — a signature that accepts any src (`communication_op.py:381`) — becomes an `AssertionError`. I checked: no in-repo or ATOM caller passes `src != 0` today, so the immediate blast radius is the silent transport swap plus TP and EP object broadcasts now multiplexing one shm ring with a single write cursor. Still, this is a behavior change nobody asked for.

Fix: inherit only when `use_message_queue_broadcaster` is True, otherwise `self.mq_broadcaster = None`.

**2. `op_tests/multigpu_tests/test_reuse_identical_rank_groups.py:127` — the GPU-free tests never execute the PR's code, and the file doesn't run in CI on this PR.**

`_record_groups` captures `original = ps.init_model_parallel_group` but `recorder` returns a `_FakeGroup` without ever calling it. So `GroupCoordinator.__init__(reuse_from=...)`, `CudaCommunicator(reuse_from=...)`, the ownership flags and the guarded `destroy()` are never touched by the 7 GPU-free tests — only the ~8 lines of key/dict bookkeeping in `_build_group`.

The fake has already diverged from the real class in a load-bearing way: the real `GroupCoordinator` has **no `self.reuse_from` attribute at all** (`grep -n "self.reuse_from"` → zero hits); only `_FakeGroup` defines it. `_assert_reuse_invariants` therefore can only ever run against fakes, and would `AttributeError` the moment someone pointed it at a real group.

On the CI side: `.github/scripts/aiter_test.sh:23-25` collects `op_tests/multigpu_tests/**` only when `MULTIGPU == TRUE`, which `aiter-test.yaml` gates on the `multigpu` PR label. This PR currently carries no labels, so none of this file runs. Dropping `_owns_process_groups = False` (→ double `destroy_process_group` at teardown) would ship entirely green. Worth adding the label before merge, at minimum.

**3. `aiter/dist/device_communicators/communicator_cuda.py:48` — `reuse_from` copies four communicator handles with zero validation, on a now-public kwarg.**

Lines 53-56 assign `pynccl_comm` / `ca_comm` / `qr_comm` / `symm_mem_comm` straight from the source and `return` — no assert that `reuse_from.world_size == self.world_size`, `.ranks == .ranks`, or `.device == .device`. Correctness rests entirely on the `key = tuple(my_ranks)` dedup in `_build_group`, and the PR's own comment (1939-1944) stresses that rank *order* must match byte-for-byte.

`reuse_from` is now a public kwarg on `init_model_parallel_group` (parallel_state.py:1590). Any external caller that gets the key logic wrong produces a communicator whose CustomAllreduce/ncclComm spans a different rank set or is sized for a different world_size: `all_gather` concatenates in the wrong order, `reduce_scatter`/`send`/`recv` address the wrong neighbours, CA/QR kernels index out of bounds. All silent wrong answers, no exception. Three cheap asserts turn that into a startup failure.

`GroupCoordinator`'s reuse branch has the same shape — it ignores its `group_ranks`, `torch_distributed_backend` and `local_rank` arguments entirely, setting `self.local_rank = local_rank` (line 535) while `self.device = reuse_from.device`, so the two can disagree.

### Should fix before merge

**4. `parallel_state.py:544` — a reusing EP group hands mori a gloo PG that TP is still using.**

The EP branch passes `cpu_group=self.cpu_group` (= `reuse_from.cpu_group`) into `CudaCommunicator`, and `MoriAll2AllManager._init_mori_shmem` (`device_communicators/all2all.py:33-35`) does `_register_process_group("mori", cpu_group)` followed by `shmem_torch_process_group_init("mori")`, which issues its own bootstrap collectives. Before this PR, EP always owned a private gloo PG for that.

`all2all_manager` is built lazily on the first MoE forward, so mori's bootstrap barrier can land on the same gloo PG while the driver rank is mid-`send_object`/`broadcast_object_list` for TP (an asymmetric, source-rank-only send). Interleaved gloo traffic from two independent producers on one PG hangs or cross-delivers. The fixed registration name also means a second EP-like group silently overwrites the first registration. The DeepEP backends make the same exclusivity assumption. I haven't reproduced a hang — flagging the assumption, since it's the one thing the reuse path takes away that the source group can't give back.

**5. `parallel_state.py:1922` — the unanimity `all_reduce` runs before `assert _TP is None`.**

The `torch.distributed.all_reduce(_flag, group=get_world_group().cpu_group)` at 1923 executes ~45 lines before the first precondition assert at 1968. Today a second `initialize_model_parallel()` call — or one reached by only a subset of ranks (a re-init/recovery path, or ranks that didn't raise earlier) — fails fast and locally. After this PR those ranks block forever in gloo instead of raising: exactly the silent-hang mode this check was added to prevent, relocated one level up.

It also injects a mandatory world gloo barrier into every startup, for 100% of users, to police a feature that is off by default — including the `custom_group_config` path where `need_std_comm` is False and reuse is a guaranteed no-op. Move it after the precondition asserts and gate it.

**6. `parallel_state.py:1925` — the guard is a bare `assert`, so `-O` deletes it.**

`python -O` / `PYTHONOPTIMIZE` strips asserts, removing the exact protection whose stated purpose (line 1920) is to "turn a silent hang into an error" — while the `all_reduce` cost is still paid. Production serving images commonly set PYTHONOPTIMIZE. Export `AITER_REUSE_IDENTICAL_COMM_GROUPS=1` on some ranks only (a per-node systemd env file, or `srun --export` dropping it on one node) and the ranks with the flag skip `new_group()` for DCP/EP while the others call it; `new_group` requires participation from every process in the default PG, so the job deadlocks in init with no output. Use `if _n_set not in (0, world_size): raise RuntimeError(...)`.

**7. `parallel_state.py:554` — `if "ep" in self.unique_name:` is a second copy of a rule that already exists.**

`base_device_communicator.py:123` has `self.is_ep_communicator = "ep" in unique_name`, and :124 derives `use_all2all` from it. Those are the only two sites in the tree carrying this rule, and they can drift silently.

The false-negative direction is the damaging one: an EP-role group that takes the `else` branch gets `self.device_communicator = src_dc`, an object whose `unique_name` is `"tp:0"` → `is_ep_communicator` False → `use_all2all` False → `all2all_manager` stays None → `prepare_communication_buffer_for_model` returns early at base:261 (MoE weights never prepared, wrong outputs, no error) and `dispatch()` trips `assert self.all2all_manager is not None`. False positive: any group name containing 'ep' as a substring ('deepep', 'prefill_ep', 'step'). Thread an explicit `is_ep` from `_build_group(group_name == "ep")`, or read `src_dc.is_ep_communicator`.

**8. `parallel_state.py:1547` — the ownership flags default to the destructive branch.**

`_owns_process_groups` / `_owns_device_communicator` are assigned only at lines 549, 571 and 576, all inside the `reuse_from is not None` block; the normal constructor never assigns either. They're then read as `getattr(self, ..., True)` — a defensive default whose fallback is "I own it, destroy it". So the invariant exists nowhere a reader or a linter can see it, any consumer reading `group._owns_process_groups` directly gets an AttributeError on every non-reused group, and a future edit that populates `device_communicator` on a path where the flag isn't set makes the object claim ownership of a borrowed ProcessGroup — `destroy_process_group()` called twice raises, aborting teardown mid-way. (`_owns_device_communicator` is already unset when the reuse path has `use_device_communicator=False` or `world_size == 1`; harmless today only because `device_communicator` is None there.) Declare both in the annotated attribute block near line 500 and set them explicitly on both paths.

**9. `parallel_state.py:1543` — a borrower's `destroy()` leaves `device_communicator` live.**

`del self.device_group` / `del self.cpu_group` run unconditionally while `self.device_communicator.destroy()` is gated on `owns_dc`, and nothing sets `self.device_communicator = None`. After `_DCP.destroy()` (owns_pg=False, owns_dc=False), `_DCP.device_communicator` still points at TP's `CudaCommunicator`, whose `.cpu_group`/`.device_group` keep the destroyed PGs reachable. A compiled-graph node dispatching through `_groups["dcp:0"]` then reaches `device_communicator.all_reduce` on a destroyed ProcessGroup instead of getting the intended `ValueError("Group ... is destroyed.")`. The non-reuse path never had this window. One line: `self.device_communicator = None` at the end of `destroy()`.

**10. `parallel_state.py:1917` — hand-rolled env parsing accepts only the literal `"1"`.**

`os.environ.get("AITER_REUSE_IDENTICAL_COMM_GROUPS", "0") == "1"` re-implements what `_env_flag()` (`device_communicators/custom_all_reduce.py:37`) already does, accepting `1/true/yes/on` case- and whitespace-insensitively. That helper backs every other AITER_* boolean in `aiter/dist` (`AITER_CUSTOM_AR_DISABLE_GFX1250`, `AITER_CUSTOM_AR_FORCE_IPC`, `AITER_CUSTOM_AR_FORCE_VMM`, `AITER_CUSTOM_AR_RAW_INPUT_POOL`). A user who exports `=true` — which works for all of those — gets the feature silently off, with no warning and no log line reporting the resolved value. In a mixed launcher, `=1` on some nodes and `=true` on others hard-fails the whole job via the unanimity check with a message saying the flag "disagrees across ranks" while both operators believe they enabled it. Import `_env_flag` (or hoist it into `aiter/dist/utils.py`, which would then have two consumers) and log the resolved boolean once from rank 0.

### Structure and tests

**11. `parallel_state.py:537` — the 55-line `reuse_from` block is a second constructor body, and it has already drifted.**

It re-assigns the same 8 fields by hand and ends in a bare `return`, so the `logger.debug("Initialized GroupCoordinator ...")` at 616-622 is never emitted for reused groups — the groups hardest to reason about at runtime are precisely the ones invisible in the startup log. The missing `self.reuse_from` attribute (finding #2) is the same drift from the other direction. Every field added to `__init__` from now on must be added in two places, and one added only to the normal path yields an AttributeError — or a `getattr(..., default)` that quietly picks wrong behavior — that appears solely on DCP/EP and solely with the env flag on. Suggest restructuring so only *acquisition* differs:

```python
if reuse_from is not None:
    ranks, cpu_group, device_group = reuse_from.ranks, reuse_from.cpu_group, reuse_from.device_group
else:
    <existing new_group loop>
# ... single shared tail
```

**12. `parallel_state.py:581` — the `MessageQueue` branch on the reuse path is unreachable.**

`if use_message_queue_broadcaster and self.world_size > 1 and self.mq_broadcaster is None:` can never be true: TP is the only caller passing True and TP is the first group built, hence never a reuser. Ten lines that will be maintained, reviewed and copied forward by the next person adding a broadcaster-using group, untested in either direction — and if it ever did fire it would build a MessageQueue on a *borrowed* cpu_group with no ownership tracking. Delete it (and per finding #1, replace line 580 with `self.mq_broadcaster = None`).

**13. `test_reuse_identical_rank_groups.py:246` — the deadlock-guard test accepts any `AssertionError`, and nothing has a timeout.**

The worker body is `except AssertionError: raised = True` with no message check, and `grep -n timeout` on the file returns nothing. `initialize_model_parallel` is full of unrelated asserts (`is_initialized`, `_TP is None` :1968, `_DCP is None` :1975, custom-group validation) — remove the unanimity check, have any of those fire instead, and the test still reports PASSED while the guard is gone. Conversely, if the check is merely weakened, ranks 1-3 take the reuse path while rank 0 calls `new_group`, all four processes block forever, `mp.spawn` joins forever, and the CI job hangs until the global timeout instead of failing. Use `pytest.raises(AssertionError, match="disagrees across ranks")` plus a spawn timeout. (The `destroy_model_parallel()` at line 237 also covers nothing — the assert fires strictly before any global is assigned.)

**14. `test_reuse_identical_rank_groups.py:302` — nothing asserts the memory saving this PR exists for.**

`_assert_shares` is three `is` comparisons on `pynccl_comm`/`ca_comm`/`qr_comm` plus `device_group`/`cpu_group` identity — which any implementation that allocates a second set and *then* aliases would also satisfy. Nothing counts `CustomAllreduce` instances, ProcessGroups created, or a `torch.cuda.memory_allocated()` delta against the flag-off baseline, so a regression that reintroduces the duplicate allocation is undetectable — even though the PR body's whole justification is non-torch memory dropping 30.21 GB → 13.40 GB.

Relatedly, the `_gpu_init` block at 280-286 that wires `ca.signal` / `register_input_buffer` is dead weight: the only collective `_gpu_worker` exercises is `ep.device_communicator.pynccl_comm.all_reduce`, which bypasses `ca` entirely — so the riskiest shared object (CustomAllreduce, with per-object registered buffers and a signal pad) is never driven. And `assert "ep" in g.unique_name` at line 163 is vacuous in the GPU-free path, where `_FakeGroup` sets `unique_name = group_name` verbatim and bypasses `_get_unique_name`.

**15. `parallel_state.py:1844` — docs and the operator-facing message weren't updated.**

`initialize_model_parallel`'s docstring (1844-1871) documents only the parallel sizes, backend and `custom_group_config`. A caller reading it cannot learn that groups may now share process groups and communicators, nor that the function is now **collective** (it all_reduces on the world cpu_group before any group is built). `destroy()`'s contract changed from always-destroy to conditionally-destroy with no docstring at all.

And the assert message says `reuse_identical_rank_groups disagrees across ranks` — the name of an internal local. An operator debugging an 8-rank init failure searches their config and environment for that string and finds nothing; it needs to say `AITER_REUSE_IDENTICAL_COMM_GROUPS`.

gbyu-amd:
Thanks — this was a genuinely useful review; several of these were real defects, not style. All 15 are addressed in `a074cc4a`. Point by point, with the code as it stands now.

### Blocking

**1. Reuse path ignored `use_message_queue_broadcaster=False`.** Fixed, and merged with #12. `parallel_state.py:657` now unconditionally starts from `self.mq_broadcaster = None` on *both* paths; the queue is only built when the group itself asked for it:

```python
self.mq_broadcaster = None
if use_message_queue_broadcaster and self.world_size > 1:
    if reuse_from is not None:
        raise NotImplementedError(...)   # see #12
    self.mq_broadcaster = MessageQueue.create_from_process_group(...)
```

So DCP/EP no longer inherit TP's shm ring, `broadcast_object` keeps `torch.distributed.broadcast_object_list` for them, and the `src == 0` assert is no longer reachable via a borrowed queue. `destroy()` also clears it (`:1593`).

**2. GPU-free tests didn't execute the PR's code / no CI.** The fake group is gone. `op_tests/multigpu_tests/test_reuse_identical_rank_groups.py` now patches **only the CUDA leaf** — `_patch_cuda_communicator()` (`:136`) swaps `communicator_cuda.CudaCommunicator` for `_StubCommunicator` (`:89`), which works because `GroupCoordinator.__init__` imports it at call time. Everything above the leaf is the real thing: `_build_group`, `GroupCoordinator.__init__(reuse_from=...)`, the validation asserts, the ownership flags and the guarded `destroy()` all run on every GPU-free test.

`_assert_reuse_invariants` (`:177`) was rewritten to assert on real attributes only — ordered `ranks`, `rank_in_group`, `device_group is src.device_group`, `not _owns_device_group`, EP's private `cpu_group` + `_owns_cpu_group`, `mq_broadcaster is None`, `is_ep_communicator`/`use_all2all`, shared `pynccl_comm`. `self.reuse_from` is now a declared attribute on the real class (`:522`, set at `:547`), so the invariants are valid against real groups too.

I proved the suite is non-vacuous by mutation: flipping `is_ep=group_name == "ep"` to `is_ep=False`, and forcing `_owns_device_group = True` on the borrow path, are each caught by the GPU-free tests. (An earlier no-op mutation was, correctly, *not* caught — that's why the two above were chosen.)

On CI: agreed, the file doesn't run without the `multigpu` label. I've left the label off so far because I don't want to add it silently mid-review — please add it (or say the word and I will) before merge.

**3. `communicator_cuda.py` copied handles with zero validation.** Three asserts added before any assignment (`communicator_cuda.py:56-69`): `world_size` equality, ordered `ranks` equality, `device` equality — so an external caller with a wrong dedup key now fails at startup instead of silently addressing the wrong peers.

The same shape is validated one level up in `GroupCoordinator` (`parallel_state.py:562-569`): ordered `reuse_from.ranks == my_ranks`, and `reuse_from.local_rank == local_rank` — closing the "sets `self.local_rank` while taking `reuse_from.device`" disagreement you pointed at.

### Should fix before merge

**4. EP handed mori a gloo PG TP still owns.** Fixed by giving a reusing EP group its **own private gloo `cpu_group`** (`parallel_state.py:578-588`):

```python
if is_ep and use_device_communicator and self.world_size > 1:
    # mori registers this group under a fixed name and bootstraps on it,
    # assuming exclusive use. A gloo PG holds none of the NCCL/CA/QR
    # buffers reuse saves, so EP keeps a private one.
    ... new_group(ranks, backend="gloo") ...
    self._owns_cpu_group = True
```

This costs nothing the PR was trying to save — the memory is in NCCL/CA/QR, which EP still shares — and restores the exclusivity mori and the DeepEP backends assume. The GPU test asserts `ep.cpu_group is not source.cpu_group` while the rank list still matches.

**5. Unanimity `all_reduce` ran before the preconditions.** Moved. The six `assert _TP/_DCP/_PCP/_PP/_DP/_EP is None` (+ `_CUSTOM`) now run first (`:1970-1977`), then the collective (`:1982`). It's also gated: `if need_std_comm and world_size > 1:` — so the `custom_group_config` path and single-rank runs pay nothing, and the barrier is no longer charged to feature-off users on a path where reuse is a guaranteed no-op. (The `global` statement was hoisted to `:1967` so the precondition reads are legal.)

**6. Bare `assert` stripped by `-O`.** Now a `raise` (`:1988-1996`):

```python
if _n_set not in (0, world_size):
    # Not an assert: `python -O` would strip this guard.
    raise RuntimeError(...)
```

**7. Duplicated `"ep" in unique_name` rule.** Removed. `is_ep` is threaded explicitly from the one place that knows: `_build_group(..., is_ep=group_name == "ep")` (`:2036`) → `init_model_parallel_group(is_ep=...)` (`:1624`) → `GroupCoordinator.__init__(is_ep=...)` (`:538`). No name matching in the reuse path. Belt-and-braces, the non-EP share branch asserts it isn't inheriting an EP communicator (`:636`):

```python
assert not src_dc.is_ep_communicator, "...use_all2all / all2all_manager state is per-EP-group..."
```

**8. Ownership flags defaulted to the destructive branch.** All three are now declared in the annotated attribute block (`:525-527`) and assigned unconditionally on **both** paths — borrow path `:576/:588/:591`, normal path `:616-617`, communicator `:631/:655`. No `getattr(..., True)` remains; `destroy()` reads the attributes directly. Also renamed: the single `_owns_process_groups` became `_owns_cpu_group` / `_owns_device_group`, because after #4 an EP borrower owns one and not the other.

**9. Borrower's `destroy()` left `device_communicator` live.** Fixed (`:1587-1591`):

```python
if self.device_communicator is not None:
    if self._owns_device_communicator:
        self.device_communicator.destroy()
    # Drop it either way: a kept reference would route collectives into
    # process groups that are destroyed by now.
    self.device_communicator = None
```

`_decision_worker` asserts `grp.device_communicator is None` for every borrower after teardown.

**10. Hand-rolled env parsing accepted only `"1"`.** Hoisted, as you suggested. `aiter/dist/utils.py:176`:

```python
def env_flag(name: str) -> bool:
    """Parse a boolean AITER_* env var: ``1/true/yes/on``, case-insensitive."""
    return os.environ.get(name, "").strip().lower() in ("1", "true", "yes", "on")
```

`custom_all_reduce.py` now imports it and its five `_env_flag(` call sites were renamed to `env_flag(` (`_env_flag` no longer exists anywhere in the tree). `parallel_state.py:1962` uses `env_flag("AITER_REUSE_IDENTICAL_COMM_GROUPS")`, and rank 0 logs the resolved value once (`:1997`): `AITER_REUSE_IDENTICAL_COMM_GROUPS resolved to True`.

(Note `_parse_expandable_segments` was deliberately left alone — it parses a value out of a `PYTORCH_*_ALLOC_CONF` string, not an env var by name, so it isn't foldable into this helper. There are ~90 other ad-hoc env parses under `aiter/ops` and `aiter/aot`; sweeping them would drag an `aiter/dist` fix into unrelated trees, and a truly global helper probably belongs in `aiter/utility/` to avoid an ops→dist dependency. Happy to do that as a follow-up.)

**11. 55-line second constructor body.** Restructured exactly as you sketched — only *acquisition* branches now (`:549-617`), everything after is a single shared tail, so the `logger.debug("Initialized GroupCoordinator ...")` fires for reused groups too, and it now reports `reuse_from=<unique_name or None>` (`:627`). Fields added to `__init__` from here on land in one place.

**12. Unreachable `MessageQueue` branch on the reuse path.** Deleted; replaced by an explicit `NotImplementedError` (`:663`) documenting *why* the shm ring can't be shared (single write cursor, `src == 0` assert), rather than ten lines of untested fallback that would build a queue on a borrowed `cpu_group`.

**13. Deadlock-guard test accepted any `AssertionError`, no timeout.** Both fixed. `_unanimity_worker:357`:

```python
with pytest.raises(RuntimeError, match=_ENV):
    initialize_model_parallel(**_dims(...))
```

— matches on the operator-facing env var name (see #15), and it's a `RuntimeError` now (#6), so unrelated `assert`s can't satisfy it. `_spawn(..., timeout=180)` (`:364`) uses `mp.spawn(join=False)` plus a `ctx.join(timeout=1)` deadline loop that terminates the processes and raises `TimeoutError` — so the weakened-check hang mode fails the test in 3 minutes instead of wedging the CI job.

**14. Nothing asserted the memory saving.** New test `test_reuse_saves_process_groups_and_handles` (`:392`), parametrized over all three topologies. `_saving_worker` (`:299`) A/Bs the same topology in one process, flag off then on, and counts three things:

```python
counts[reuse] = (len(new_group_calls), len(comms), len({id(c.pynccl_comm) for c in comms}))
...
assert on_pg < off_pg          # process groups actually not created
assert on_handles < off_handles # distinct allreduce handles actually not allocated
assert on_comm <= off_comm
```

Counting *distinct handle ids* is what rules out the "allocate a second set, then alias" implementation `_assert_shares` would have accepted. `torch.distributed.new_group` is wrapped by a counting shim (`_count_new_group`, `:149`).

The dead `ca.signal` / `register_input_buffer` wiring in `_gpu_init` is deleted, with a comment noting `ca_comm` is checked by identity only and that driving CA would need `init_dist_env`'s full buffer handshake. `assert "ep" in g.unique_name` is no longer vacuous — the stub keeps the real `unique_name` and derives `is_ep_communicator` the same way `base_device_communicator.py:123` does.

End-to-end, on an 8-GPU node running Kimi-K3 (tp8/dcp8, EP aliasing TP), with reuse on: `non_torch` = **12.77–13.61 GB** per rank against the ~30 GB no-reuse baseline in the PR body, and `AITER_REUSE_IDENTICAL_COMM_GROUPS resolved to True` logged exactly once from rank 0.

**15. Docs and operator-facing message.** `initialize_model_parallel`'s docstring gained an `Environment:` section describing the flag, what borrowing means, the EP exception, and owner-only release (`:1909-1918`), plus an explicit `Collective:` note that all ranks must call it together. `destroy()` now has a docstring stating the conditional-teardown contract (`:1572`). The error message names `AITER_REUSE_IDENTICAL_COMM_GROUPS`, reports `N/world_size ranks have it enabled`, and mentions that `1/true/yes/on` are all accepted so mixed spellings are still unanimous.

---

13/13 tests pass, `black --check` and `ruff` clean. The only thing outstanding on my side is the `multigpu` label so the new test file actually runs in CI.

valarLip (approved):

PR #5112, opened 2026-08-29
[FlyDSL] 1250 clean moe aux kernel codes and ir, add ut

coderfeli:
clean moe aux kernel codes and ir, add ut


coderfeli:
E2E validated ok by  @XingerZhu

PR #5123, opened 2026-08-30
[HIP] [FlyDSL] dcp topk merge

ganyi1996ppo:
Three related changes for the Decode Context Parallel path of the DeepSeek-V3.2-style
sparse indexer. The first two are what GLM-5.2 DCP decode needs; the third is an
independent crash fix found while bringing that path up, and is useful on its own.

---
## 1. `[FlyDSL] Add DCP decode TopK merge kernel`

Under DCP each rank holds `1/W` of the KV and takes a local top-k, so the ranks must
agree on a global top-k before attention. The existing sequence reconstructs that
global set on **every** rank and then drops the `~(W-1)/W` of it the rank does not
own — W times the work, most of it discarded.

This op emits only the rank's **owned** physical KV slots, already localized and
compacted, plus the `kv_indptr` attention needs. Two kernels: a per-row radix
select for the global threshold, then a per-row pack that also computes the
cross-row prefix sum.

Two properties make the fusion possible:

- **Ownership is positional.** Candidate column `c` came from rank `c // k_loc`, so
  global ids never have to travel and the caller can exchange scores alone.
- **The slot formula collapses.** For local index `j` the physical slot is
  `block_table[j // page] * page + (j % page)` — independent of `W` and of the
  interleave size `S`. Verified by exhaustive enumeration over 50400 combinations
  of `(W, S, page, j)`, 0 mismatches. So the localize folds into the emit loop
  instead of needing a separate pass.

Allocates no device scratch (the caller owns `staging`), so it is safe inside a
captured CUDAGraph.

### Why the prefix sum lives inside `pack`

It started as its own `grid=(1,)` kernel walking the rows on a single thread. That
is O(rows) serial work sandwiched between two parallel kernels, and it showed —
**37% of the op's runtime at rows=128** (15.7 of 42.3 µs) to prefix-sum 128
integers.

Each `pack` block now recomputes the prefix it needs, in parallel. The redundancy
is real (rows blocks each scanning rows counts) but a block scan over a few hundred
values is a handful of barriers: `pack` went 3.3 → 3.5 µs while a whole kernel
launch disappeared. The loop is chunked, so `rows > 256` stays correct — verified at
rows ∈ {1, 128, 256, **257**, 512}.

That is also why `select` and `pack` cannot merge further: `pack` needs
`kv_indptr[row]`, which depends on every *other* row's count. Closing that gap needs
either a cross-block barrier (unsafe — block scheduling order is not guaranteed, so
a block spinning on a not-yet-scheduled block deadlocks) or `grid=(1,)`, which would
serialise `select` by a factor of rows.

### Performance

Against the equivalent six-op sequence, at the production shape `W=8`,
`k_loc = topk = 2048`:

| bs | six-op | fused | speedup |
|---:|-------:|------:|--------:|
| 1 | 166.5 µs | 33.9 µs | 4.91× |
| 4 | 173.6 µs | 33.8 µs | 5.14× |
| 16 | 165.2 µs | 34.3 µs | 4.82× |
| 32 | 167.3 µs | 33.6 µs | 4.98× |
| 64 | 174.6 µs | 34.3 µs | 5.09× |
| 128 | 167.4 µs | 34.0 µs | 4.92× |

The curve is **flat** — 33.6-34.3 µs regardless of bs — where the six-op path grows
with rows. Plus half the all-gather payload (`[rows, k_loc]` instead of
`[2, rows, k_loc]`) and 2 kernels instead of ~16.

This is a **single-op** measurement. It has not been attributed against a full decode
step, so the wall-clock share depends on how often the indexer runs per forward.

### Correctness

`op_tests/flydsl_tests/test_flydsl_dcp_topk_merge.py` — **45 passed**. Covers a
differential test against the six-op reference, edge cases (`topk > n_cand`, empty
rows, all-padding rows), a shape sweep, and tie-heavy inputs where the equal-to-
threshold admission rule is what decides the partition.

Two GPU races were found and fixed during development, each with a reproducible
before/after:

- **Bucket-hit test.** Every thread reads `remaining_k` from LDS but only the thread
  owning the k-th element writes it. Without a barrier between the two, a fast writer
  updates it while a slow thread is still reading, the slow thread then satisfies the
  hit condition against the *new* value and writes a second bucket — a corrupted
  threshold.
- **Block scan.** A fast wave entering the *next* `block_exclusive_prefix_i32` call
  would overwrite `scan[warp]` while a slow wave was still reading the current call's
  totals. Fixed with a trailing barrier.

The second one I initially dismissed as a non-race based on 40 clean iterations of a
single shape. It reproduced later under a different shape. A race is structural, not
a property of one observed schedule — recording that here because the first ruling
was wrong.

---

## 2. `[HIP] Add optional score output to top_k_per_row_decode`

`top_k_per_row_decode` now takes an optional `values` tensor and writes each selected
index's logit alongside it, mirroring `top_k_per_row_prefill`. The one-block kernel
already supported this (`WRITE_TOPK_VALUES` is a template parameter of
`dispatch_topk_oneblock`); the decode entry point hardcoded it to false with
`/*out=*/nullptr`. Wiring it through the header, the pybind arg list and the Python
wrapper is the whole change — **the default path compiles to the same instantiation
as before**.

This lets a DCP decode indexer drop the separate gather whose only job was to read
these same scores back out of the logits.

### Also: a padding-value fix

Where the index is padded with `-1`, the score was padded with `0.0`. A `0.0` pad does
**not** sort below real scores — logits are routinely negative, so the padding
outranks them, and a consumer that ranks these scores (a DCP merge across ranks) lets
padding steal real candidates' slots. Pad with `-inf` instead.

Measured impact before the fix: at `W=8` the disjoint owned-partition collapsed from
256 entries to 30.

The changed line is shared with prefill, but **every in-tree prefill caller passes
`values=None`** (checked all 5), so `WRITE_TOPK_VALUES` is false there and the store
never executes.

---

## Validation

- `op_tests/flydsl_tests/test_flydsl_dcp_topk_merge.py`: **45 passed**
- HIP `values`: padding `== -inf`, score `==` gathered logit, `values=None`
  bit-identical to before, non-stable path — all pass
- Gluon fix: 3 previously-aborting shapes compile; reverting the patch reproduces the
  abort on `main`, so the fix is not a no-op
- End-to-end: GLM-5.2, TP=8, DCP=8, 65536-token inputs, **8/8 requests succeeded**,
  32 prefill batches including the `done=[16384]` chunk that previously aborted. Zero
  `iota_range` / `proc died` / `exitcode=-6` in the log.

## Not covered

- **The DCP merge has not been validated on real multi-GPU DCP.** The all-gather
  itself is untested; the consumer keeps it behind a default-off flag.
- Tested on one gfx950 machine with Triton 3.7.0.
- The `waves_per_eu` timings are 10-iteration wall-clock, not a rigorous benchmark.

### DIFF
--- a/aiter/ops/flydsl/dcp_topk_merge.py
+++ b/aiter/ops/flydsl/dcp_topk_merge.py
@@ -1,0 +1,162 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""FlyDSL DCP decode TopK merge interface."""
+
+import functools
+
+import torch
+
+from .kernels.dcp_topk_merge import build_dcp_topk_merge_module
+from .kernels.kernels_common import get_warp_size
+from .kernels.tensor_shim import _run_compiled
+
+
+@functools.cache
+def _get_launcher(n_cand, k_loc, topk_tokens, page_size, world_size):
+    return build_dcp_topk_merge_module(
+        n_cand, k_loc, topk_tokens, page_size, world_size
+    )
+
+
+@functools.lru_cache(maxsize=128)
+def _validate(
+    sc_shape,
+    sc_dtype,
+    li_shape,
+    li_dtype,
+    bt_shape,
+    bt_dtype,
+    idx_numel,
+    idx_dtype,
+    indptr_shape,
+    indptr_dtype,
+    counts_shape,
+    counts_dtype,
+    stage_shape,
+    stage_dtype,
+    dcp_rank,
+    world_size,
+    topk_tokens,
+    page_size,
+    device_warp,
+):
+    if len(sc_shape) != 2 or sc_dtype != torch.float32:
+        raise ValueError("gathered_scores must be a 2D float32 tensor")
+    rows, n_cand = sc_shape
+    if world_size <= 0 or n_cand % world_size:
+        raise ValueError("gathered_scores.shape[1] must be divisible by world_size")
+    k_loc = n_cand // world_size
+    if not 0 <= dcp_rank < world_size:
+        raise ValueError("dcp_rank out of range")
+    if li_shape != (rows, k_loc) or li_dtype != torch.int32:
+        raise ValueError("local_idx must be int32 [rows, n_cand // world_size]")
+    # The kernel slices this 2D (fx.slice(block_table, (row, None))), so a 1D
+    # tensor fails deep inside the JIT rather than here.
+    if len(bt_shape) != 2:
+        raise ValueError(f"block_table must be 2D [rows, max_blocks], got {bt_shape}")
+    # rows == num_req: DCP+DSA is qlen=1 decode only.
+    if bt_shape[0] != rows:
+        raise ValueError("block_table.shape[0] must equal rows (qlen=1 decode only)")
+    # The kernel reads every one of these as i32. torch builds page tables as
+    # int64 by default, and an int64 block_table passes a shape-only check and
+    # then yields silently wrong physical slots -- no crash, wrong KV.
+    if bt_dtype != torch.int32:
+        raise ValueError("block_table must be int32")
+    if idx_dtype != torch.int32:
+        raise ValueError("out_kv_indices must be int32")
+    if indptr_dtype != torch.int32:
+        raise ValueError("out_kv_indptr must be int32")
+    if counts_dtype != torch.int32:
+        raise ValueError("owned_counts must be int32")
+    if stage_shape != (rows, k_loc):
+        raise ValueError("staging must be int32 [rows, n_cand // world_size]")
+    if stage_dtype != torch.int32:
+        raise ValueError("staging must be int32 [rows, n_cand // world_size]")
+    if indptr_shape[0] < rows + 1:
+        raise ValueError("out_kv_indptr must hold rows + 1 entries")
+    if counts_shape[0] < rows:
+        raise ValueError("owned_counts must hold rows entries")
+    # Per row a rank owns at most its own plane (k_loc) and at most the whole
+    # global winner set (topk_tokens) -- whichever binds first. Requiring k_loc
+    # unconditionally would reject a valid buffer whenever topk_tokens < k_loc.
+    if idx_numel < rows * min(k_loc, topk_tokens):
+        raise ValueError("out_kv_indices too small for the worst case")
+    if page_size <= 0:
+        raise ValueError("page_size must be positive")
+    if device_warp != 64:
+        raise ValueError("the FlyDSL DCP merge kernel requires a wave64 GPU")
+
+
+def flydsl_dcp_topk_merge(
+    gathered_scores: torch.Tensor,
+    local_idx: torch.Tensor,
+    block_table: torch.Tensor,
+    out_kv_indices: torch.Tensor,
+    out_kv_indptr: torch.Tensor,
+    owned_counts: torch.Tensor,
+    staging: torch.Tensor,
+    dcp_rank: int,
+    world_size: int,
+    topk_tokens: int,
+    page_size: int,
+) -> None:
+    """Select this DCP rank's share of the global top-k and emit packed KV slots.
+
+    `gathered_scores[:, c]` MUST come from rank `c // k_loc` at that rank's local
+    top-k position `c % k_loc` -- the natural all_gather(dim=0) order. Ownership
+    is positional, so a permuted layout silently produces wrong KV.
+
+    PRECONDITION: unused candidate slots in `gathered_scores` must be -inf, not
+    0.0. Padding is detected via `local_idx < 0` for THIS rank's plane only; the
+    other W-1 planes carry no liveness signal, so their padding has to lose the
+    global threshold comparison on score alone. Logits are routinely negative,
+    so a 0.0 pad outranks real candidates and this rank silently under-emits.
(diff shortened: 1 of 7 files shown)


aiter/ops/topk.py
```diff
@@ -482,6 +489,44 @@ def top_k_per_row_decode(
         k,
         workspace,
         stable,
+        values,
```
lirui927, line 506: should add value checks as well — dtype, shape, and contiguous.

op_tests/test_flydsl_dcp_topk_merge.py
```diff
@@ -0,0 +1,579 @@
...
+    parser.add_argument("-p", "--page", type=int, nargs="*", default=[64])
+    parser.add_argument(
+        "-m", "--mode", type=str, nargs="*", default=CASE_MODES, choices=CASE_MODES
+    )
+    args = parser.parse_args()
+
+    df = []
+    for rows, world, k_loc, topk, page, mode in itertools.product(
+        args.rows, args.world, args.k_loc, args.topk, args.page, args.mode
+    ):
```
lirui927, line 460: The test organization doesn’t follow the existing CI conventions.

aiter/ops/flydsl/dcp_topk_merge.py
```diff
@@ -0,0 +1,156 @@
...
+
+@functools.cache
+def _get_launcher(n_cand, k_loc, topk_tokens, page_size, world_size):
+    return build_dcp_topk_merge_module(
+        n_cand, k_loc, topk_tokens, page_size, world_size
+    )
+
+
+@functools.lru_cache(maxsize=8)
+def _warp_size(device_index: int) -> int:
```
coderfeli: there is get_warp_size in kernel common.py

zufayu:
Nit (non-blocking): the commented-out `mb/ob dispatch` block at the tail of `top_k_per_row_decode` is dead code that predates this PR. Since the function is already being rewritten here, worth dropping it in a follow-up — git history keeps the reference, and leaving it invites rot (uncompiled, unmaintained, but scanned by every future reader to confirm it's unreachable).

```diff
     } else {
         aiter::ob::dispatch_topk_oneblock<float, int, 1024, false, aiter::ob::Phase::Decode>(
             ws_ptr, buf_size, logits_ptr, static_cast<int*>(nullptr),
             batch, stride0,
             /*rowStarts=*/nullptr, /*rowEnds=*/seq_lens_ptr,
             kTopK, /*out=*/nullptr, indices_ptr,
             select_min, stream, /*sorted=*/true, static_cast<int>(next_n));
     }
-
-    // --- Original mb/ob dispatch (commented out for reference) ---
-    // Both branches take the caller-provided `workspace` (no device allocation
-    // here); the mb buffer is zeroed + self_reset by the kernel (prezeroed=true).
-    // if (aiter::should_use_mulblocks(batch, stride0)) {
-    //     aiter::mb::standalone_stable_radix_topk<float, int, false, true, aiter::mb::Phase::Decode>(
-    //         ws_ptr, buf_size, logits_ptr, batch, stride0,
-    //         /*rowStarts=*/nullptr, /*rowEnds=*/seq_lens_ptr,
-    //         kTopK, /*out=*/nullptr, indices_ptr,
-    //         is_largest, stream, static_cast<int>(next_n), /*prezeroed=*/true);
-    // } else {
-    //     aiter::ob::dispatch_topk_oneblock<float, int, 1024, false, aiter::ob::Phase::Decode>(
-    //         ws_ptr, buf_size, logits_ptr, static_cast<int*>(nullptr),
-    //         batch, stride0,
-    //         /*rowStarts=*/nullptr, /*rowEnds=*/seq_lens_ptr,
-    //         kTopK, /*out=*/nullptr, indices_ptr,
-    //         select_min, stream, /*sorted=*/true, static_cast<int>(next_n));
-    // }
 }
```

Unrelated to the `values` correctness — just pre-existing hygiene that this PR's churn makes cheap to clean up.

zufayu (dismissed):

coderfeli (dismissed):

valarLip (dismissed):

op_tests/test_flydsl_dcp_topk_merge.py
```diff
@@ -0,0 +1,619 @@
...
+
+
+# @benchmark does not hide the fn from pytest: collected by its test_ name, it
+# would be called with zero fixtures and raise TypeError. The parameterized
+# wrapper below is this file's pytest entry point into the same code.
+test_dcp_topk_merge.__test__ = False
+
+
+@pytest.mark.parametrize("mode", CASE_MODES)
+@pytest.mark.parametrize("rows,world,page", [(1, 8, 64), (17, 4, 16), (256, 2, 128)])
```
valarLip: we don't use pytest.. need clean

op_tests/test_flydsl_dcp_topk_merge.py
```diff
@@ -0,0 +1,619 @@
...
+def test_modes(rows, world, page, mode):
+    """Drive the benchmark fn's assertions over edge shapes under pytest."""
+    k_loc = 64
+    # "overtake" is the topk >= n_cand degenerate case; the rest stay selective.
+    topk = world * k_loc if mode == "overtake" else 128
+    test_dcp_topk_merge(rows, world, k_loc, topk, page, mode)
+
+
+def test_reference_model_selects_global_topk():
```
valarLip: test for reference?

valarLip (approved):

PR #5132, opened 2026-08-31
[Triton/Gluon] [Fix] fix post-merge issues in dev-lumen cherry-pick (#4978)

WuLei-AMD:
Four correctness/quality fixes found during review:

- moe_wgrad: widen offs_token to int64 before stride multiply to prevent int32 overflow at DSv4 scale (batch=512, seq=16384, top_k=7, hidden=7168 puts the product ~4.2e11, well past 2^31)
- triton/__init__: export cross_entropy_forward_chunked which was declared in cross_entropy.__all__ but omitted from the package re-export
- fused_norm_quant_gemm: remove dead _get_fused_cpp probe; module_fused_norm_quant_gemm was never registered in JIT so the C++ branch always fell through to the Python path anyway
- triton/__init__: gate mxfp8_attention lazy-load behind is_cdna4() check so non-gfx950 callers get NotImplementedError instead of a cryptic AssertionError

op_tests fixes:
- test_sonicmoe: add pytest test_sonicmoe_correctness so CI can discover it; replace string PASS/FAIL with assert
- test_moe_aux_loss: bwd reference was missing the grad_aux_loss factor; add parametrize over [0.5, 1.0, 2.0] to catch the regression
- test_dsv4_indexer / test_sparse_mla_dsv4_train: replace importlib+sys.modules hack with standard imports
- test_topk_softmax: merged into test_moe_topk_gating (removes duplicate coverage); fix weight comparison to sort by expert ID before assert_close

## Submission Checklist


zufayu:
## [aiter] PR #5132 — [Triton/Gluon] [Fix] fix post-merge issues in dev-lumen cherry-pick (#4978)

**Fixes four issues left by the #4978 cherry-pick — int64 widening in `moe_wgrad`, a missing `__init__` re-export, a dead C++ probe, and a gfx950 gate on `mxfp8_attention` — plus five op_test cleanups.**

Validation (deterministic): NOT RUN — no exact-head `validation_report.json`. Environment fact worth recording: `moe_wgrad` has no test target anywhere in the repository, so a validation run would have nothing to select for the headline fix.

🔴 **`test_sonicmoe.py` exits 0 when correctness fails, and this PR rewrote the lines that decide it.** `main()` accumulates `all_pass`, prints `"Some tests FAILED."`, and then returns — there is no `sys.exit(1)` and no raise. Under CI, which runs `timeout 60m python3 "$file"` (`.github/scripts/aiter_test.sh:85`), any SonicMoE regression producing `max_err >= 0.05` on any of the seven activations prints a failure line and still reports success. The hole predates this PR (#4978 introduced it), but this PR edits exactly that branch (`status != "PASS"` → `max_err >= 0.05`) without closing it. **Author must** make `main()` exit non-zero when `all_pass` is false. [verified]

⚠️ **The new pytest cases do not execute in aiter CI, so the stated goal is not met.** The description says the new `test_sonicmoe_correctness` is added "so CI can discover it", but CI runs `op_tests/*.py` as scripts; pytest is applied only to `op_tests/triton_tests/` (`triton-test.yaml:102`). `python3` executes module-level code and `__main__`, and neither `main()` invokes the new functions — this applies both to `test_sonicmoe_correctness` and to the `test_topk_softmax` cases folded into `test_moe_topk_gating.py`, which sit after `if __name__ == "__main__": main()`. The root cause is repository-wide rather than this PR's fault (across `op_tests/*.py`, 23 files define pytest cases with no call site at all, totalling 171 cases that never run), and I am filing that separately. For this PR the local fix is small. **Author must** either call the new functions from each `main()`, or make the script entry point run pytest on itself, e.g. `sys.exit(pytest.main([__file__, "-q"]))` behind a flag that preserves the existing benchmark entry point. [verified]

⚠️ **`moe_wgrad` has no test and no in-repo caller, so the headline int64 fix cannot be verified or regression-guarded.** `grep -rln moe_wgrad --include=*.py` matches only `aiter/ops/triton/moe/moe_wgrad.py` and its package `__init__`; the op is exported through `__all__` but nothing exercises it. The widening itself is complete — `off_expert` and `offs_token_id` were already `int64`, and all three pointer expressions now widen, with no unadapted sibling path — but the DSv4 scale that motivates it is not reproducible by anything in the tree. **Author must** add a small correctness test under `op_tests/triton_tests/moe/`, where CI already runs pytest. [verified]

⚠️ **The rewritten bf16 case no longer uses bf16 while keeping the tolerance that only bf16 justified.** The deleted `test_topk_softmax_bf16_input` built `torch.randn(..., dtype=torch.bfloat16)` and cast to fp32, which is why it used `atol/rtol=2e-3`. The replacement `test_topk_softmax_bf16_caller_casts` builds `torch.randn(32, 8, dtype=torch.float32)` directly — no bfloat16 anywhere — while keeping 2e-3, against 1e-4 in the parametrized case next to it. It is now a strictly weaker duplicate of a case that already exists, and it cannot test the caller-cast contract its name describes. **Author must** restore a bf16 input tensor, or rename the case and tighten the tolerance to 1e-4. [verified]

📝 **Three smaller items.** The description cites the overflow product as `~4.2e11`, which is `58.7M x 7168`, i.e. `offs_token` before the divide; the expression actually changed is `offs_token // top_k`, giving `8.4M x 7168 ~= 6.0e10` — still far past `2^31`, so the fix is right, but the number traces to a different expression. The deleted `test_topk_softmax_n_zero` was a pure `pytest.skip` recording that the ASM kernel cannot launch on 0 tokens; that limitation is no longer written down anywhere. And `_CDNA4_ONLY` guards only the `_BACKWARD_COMPAT_MAP` branch of `__getattr__`, while the migration target that map points at — `attention.mxfp8_attention` — is directly importable and bypasses the gate. **Reviewer should ask** whether the 0-token limitation should survive as an `xfail`, and whether the arch gate belongs at module level in `mxfp8_attention.py`. [verified]

zufayu (dismissed):
LGTM. All four findings from my review are addressed at `37ef02e96`, verified in the diff rather
than from the commit message: `sys.exit(1)` on failure in `test_sonicmoe.main()`, bf16 input
restored in `test_topk_softmax_bf16_caller_casts`, the pytest cases now invoked from `main()` so
they execute under CI's script mode, and `moe_wgrad` covered by a new test in
`op_tests/triton_tests/moe/` with an independent reference.

The 8 `Triton Tests (MI35X)` shards have since gone green, which is the first execution of that new
test — worth noting because it is the only coverage the headline int64 fix has.

One thing to keep in mind rather than fix here: that test guards functional regressions, not width
regressions. Its shapes are small, so removing the `.to(tl.int64)` would still pass. The overflow
scale quoted in the description is not reachable in CI.

vgokhale:
Please submit separate PRs for different kernels and different backends

PR #5136, opened 2026-08-31
[Triton/Gluon] Clamp the split-K heuristic so it cannot divide by zero

Boss2002n:


brunomazzottiamd (approved):
LGTM, we can merge it! However, I have a request for the scope of another PR...

`get_splitk` function is duplicated across all six GEMM and fused-GEMM files!

The implementation is almost identical, with two minor differences:

- `gemm_afp4wfp4.py` recomputes `NUM_KSPLIT` before returning.
- `batched_gemm_afp4wfp4.py` contains leftover debug comments.

Notably, the copies in `_triton_kernels/gemm/basic/gemm_a16wfp4.py` and `_triton_kernels/gemm/batched/batched_gemm_a16wfp4.py` appear unused. Their wrappers already import `get_splitk` from another module.

Code duplication makes fixes to the split-K heuristic error-prone and has already allowed the implementations to drift slightly. Please extract the shared split-K calculation into a common utility module, update all callers to import it, and remove the local copies.

Issue #5128, opened 2026-08-31
[skills] Make kernel-review tooling measurable: corpus, growth mechanism, promotion bar


zufayu:
Follow-up to #4870, filed before it merges so the sequence after it is on the record rather than
promised.

#4870 lands a review/validation layer at the **advisory tier**, which is the right tier for its
evidence — and also where the evidence stops. Its own description states the historical campaign is
local and makes no durable recall or consistency claim. The PR also demonstrates why that matters:
the original D9 rule missed 0 of 3 seeded index-width defects across three revisions, and that was
only visible because someone measured it. Prose edits to review rules cannot be scored by reading
them.

This issue covers the three steps that make the tooling improvable instead of static. They are one
programme with one dependency chain, so they are one issue.

---

### Step 1 — Commit the corpus as a replayable asset

The durable part of #4870 is not the skill text, it is the labelling methodology behind it: 14 PRs,
17 defects, each pinned to its merge commit, labels taken from the `[Fix]`/revert PRs. Today that
lives in a PR description and evaporates on merge.

- `corpus/defects.jsonl` — one record per labelled defect: introducing PR, merge SHA, file/line,
  family, severity, fixing PR, one-line description.
- `corpus/replay.sh` — checks out each case at its merge commit, time-boxes `gh`, runs the skill
  under test, scores against the labels, repeats n times.
- `corpus/metrics.json` — committed: corpus version, skill SHA, per-family recall@5, false-clearance
  rate, findings/PR, n, variance.
- **`train`/`holdout` split from the first commit.** Retrofitting it once numbers are quoted is
  expensive, and without it the ratchet overfits — #4870's production-scale table already shows the
  risk, since the scale facts came from the same fix PRs that define the labels.

Seed data: four reports from an independent gfx942 run of head `706eafafc` against ROCm/aiter#5089
are available as the first holdout records — as-shipped, instrumented, a seeded positive control
that correctly returned `BLOCK`, and a post-fix run. Details in #4870.

**Done when** one command reproduces the reported numbers, so any rule edit can be scored before and
after.

### Step 2 — Make the corpus grow for free

A corpus that only grows when someone remembers to grow it is a one-off engineering project, not an
asset. The mechanism already exists in this repo's workflow: every `[Fix]` and revert PR is by
construction a labelled defect with a known introducing commit, a known fixing commit, and a
maintainer who has just finished understanding it.

Add one step to the contributing guidance for `[Fix]`/revert PRs: append the record to
[...]

PR #4885, opened 2026-08-20
[Triton/Gluon] [FlyDSL] feat(flydsl): Add HSTU Backward kernel

SamiAario-AMD:
## Motivation

The purpose of this work is to implement the backward counterpart to the FlyDSL HSTU forward kernel.

Jira ID : SILOTIGER-616

## Technical Details

Depends on the FlyDSL HSTU forward kernel in [#4441](https://github.com/ROCm/aiter/pull/4441). This PR adds the backward pass for the same attention block defined in [HSTU (arxiv 2402.17152v2)](https://arxiv.org/pdf/2402.17152v2).

The backward does **not** stash activations from the forward — it recomputes `S = alpha * Q K^T` and the SiLU derivative from `(Q, K)`, then produces `(dQ, dK, dV)` via two lock-free FlyDSL kernels:

1. **Fused KV-owned kernel** (`hstu_attention_bwd.py`) — owns a KV tile, streams query tiles, and computes **dV** and **dK** from a single `S` recompute.
2. **Q-owned kernel** (`hstu_attention_bwd_dq.py`) — mirrors the forward layout; each program owns a query tile and streams KV tiles to compute **dQ**.

Shared indexing/layout helpers live in `hstu_attention_common.py`. The public API is `flydsl_hstu_attention_bwd` in `hstu_attention_kernels.py`, wired into `FlydslHstuAttention` for end-to-end autograd.

**Feature parity with the forward kernel:**
- Jagged (variable-length) batching via `seq_offsets`
- Causal masking plus the same optional variants: `num_targets`, sliding window (`max_attn_len`), and contextual prefix (`contextual_seq_len`)
- `attn_dim != hidden_dim` (asymmetric head dims)
- `{f16, bf16}` inputs, fp32 accumulation, fast-math SiLU (same conventions as forward: `alpha`-in-score, `1/N` on `dS`)

**Tuning & dispatch:**
- Per-kernel tuned configs in `hstu_attention_bwd_tuned.csv`

## Test Plan

- **Correctness (pytest)** — `op_tests/flydsl_tests/test_flydsl_hstu_attention_bwd.py`

- **Perf / op-test gate** — `op_tests/test_flydsl_hstu_attention_bwd.py`

- **Benchmark harness** — extended `op_tests/op_benchmarks/triton/bench_hstu_attn.py` with `--provider flydsl --mode bwd` (and optional `--sort_by_length`).

## Test Result

[...]

PR #5142, opened 2026-08-31
[skills] Add kernel PR validation and structural D9 scanning (supersedes #4870)

zhiding512:
Supersedes #4870. Same skill, plus the fixes found in the review rounds since.

**Why a new PR rather than more commits on #4870:** #4870's head lives on `jhinpan:skill/validate-kernel-pr`, a fork, so fixes found during review cannot be pushed there by the people finding them. This branch is `zhiding512/validate-kernel-pr-amdsmi-fix` in ROCm/aiter, so review rounds land directly. Nothing about the design is being relitigated here.

**Authorship:** 12 of the 23 commits are @jhinpan's, preserved unmodified with their original authorship and messages — `342ffd0e` through `a7f52abc`, identical to #4870's head. The 11 on top are review-round fixes. @jhinpan, this is yours to close #4870 in favour of, or to reject if you would rather carry these on your own branch instead.

**The review history is on #4870 and is not reproduced here:** @zufayu's two acceptance-bar reviews (including the held-out run against #5089 on MI308X/gfx942), @zhiding512's end-to-end report, and @jhinpan's replies. Read that thread for the reasoning behind the current design; #5128 tracks the measurement work agreed there.

---

### Round 1 — portability and evidence gaps found by held-out validation

**`79c62806` — treat GPU activity as optional, and hand the worktree back clean.** `amdsmi_get_gpu_activity` was a hard dependency at three call sites, so a host where that single query fails degraded the whole run to `NO_GPU` even with idle GPUs available. Hit on MI308X / ROCm 7.0, where the call returns `AMDSMI_STATUS_UNEXPECTED_DATA (43)` while enumeration, BDF, ASIC and VRAM queries all work. Activity is now optional and `gpu_claim.idleness_basis` names the evidence the claim rests on: `activity+vram`, or `vram-only`. Unknown stays distinct from zero — the previous `gfx if isinstance(gfx, int) else 0` reported an unavailable metric as a measured idle GPU. Separately, `merge_sim` applied the patch but only the baseline path set `BASE_ACTIVE`, so any run that skipped correctness exited with the candidate patch still applied, and the next run in that worktree blamed the caller for it.

**`d02515f6` — let script targets carry grid and receipt evidence.** A target driven by a `__main__` guard could reach neither `correctness_s1_grid` nor `execution_receipt`, so `INCONCLUSIVE` was the hard ceiling for every aiter `op_tests/*.py` — the repository's standard test shape. Neither skip described a property of the target: the route profiler is `sys.setprofile` plus a receipt write and pytest was only where the hook happened to be installed, while the grid was injected solely through an environment variable. `run_script_with_probe.py` runs the file under `runpy` with `run_name="__main__"`, calling the probe's own hooks so `producer` stays truthful. `--shape-arg` names the target's own CLI flag, held to the same standard of proof as the env channel: the flag must appear in an `add_argument` call, and the invalid-grid probe must still make the target fail. This is the commit that relaxes `report_schema.json`'s `PASS` constraint `test_selection.runner` from `const "pytest"` to `enum ["pytest", "script"]` — the constraint encoded exactly the assumption being removed, and the schema is what caught the violation when the first script `PASS` was produced.

**`f8904577` — ask the target whether it needs a GPU, and name what that answer rests on.** When no GPU was claimed, the target is run once with no visible device; `gpu_requirement` becomes `not-required` only if it passes **and executes at least one test**, because a suite guarded by `skipif(not torch.cuda.is_available())` also exits 0 while proving nothing. `PASS` still requires `gpu_claim: pass`, so this reaches no clearance that was previously unreachable.

### Round 2 — review-pr's side of the contract

**`73e6f575`, `2a6f4857`, `493f769f`, `dd6ff6a7`** state the promotion bar in review-pr's header, let it accept a script target's `PASS`, have it triage whether a PR changes runtime surface at all and run the validator itself when it can, and call the validator directly rather than through a `bin/` wrapper. A PR with no runtime surface is now reported `N/A` rather than unvalidated.

**`27b7033d` — measure a kernel PR's cost, not just its correctness.** A kernel PR could pass every stage while running slower, because nothing timed it. The `perf` stage benches the target with the patch reversed and with it applied, inside the base phase's lock on the same GPU. Three choices, each forced by a measurement rather than by taste: minimum over per-column medians (a mean lets an untouched `torch avg` reference column bury the regression next to it); best-of-3 per cell (five warm runs of unchanged `test_layernorm2d.py` gave `ck avg` 13.10/20.98/20.70/13.28/13.17 — 1.60x and bimodal, so single-shot at a 0.95 threshold would call unchanged code a regression about half the time; best-of-3 brings the same data to 1.014x); and a worktree snapshot around the timing run, because a bench that drops `tuned_op_bench.csv` in the repo root otherwise fails the baseline cleanliness check and skips the entire head correctness phase.

### Round 3 — this round: the skill's own honesty rules, applied to itself

The three commits below came out of reviewing this branch for stale and redundant content. Each is a case of the skill making a claim its code does not support — the failure mode it exists to catch.

**`4e053e43` — describe the validator this branch actually ships.** `perf` landed in `27b7033d` and its own skill still said it did not exist, instructing a reader to disbelieve a stage that now measures and gates, while `review-pr` said the opposite in three places. `perf` is now documented where the implementation lives, including `--perf-args`, `--no-perf` and the four `PERF_*` env knobs, which were public already and only undocumented. Also: "review-pr never builds and never runs" has been false since `493f769f`; the `bin/validate-kernel-pr` comment outlived the wrapper; and the mutant manifest's `tolerance_table` disagreed with the file it describes (`test_softmax.py` at the pinned base uses `f16=1e-2`, `bf16=2e-2`).

**`b03d15a2` — stop the report schema promising evidence nothing produces.** `report_schema.json` is the contract `review-pr` validates against, so a declared field reads as a shipped capability. `claims` declared a stage no producer writes — removed. `source_sha` was hardcoded null while the honesty rules listed source SHA as recorded evidence — now the resolved module's checkout commit, `-dirty` when the tree has uncommitted changes, which the head phase always does; null for an installed package, because attributing a commit to a prebuilt binary is the provenance claim this skill refuses to make. `reference_tolerances` is recorded and never compared, and cannot be as the data stands (dtype-keyed table, positional scraped list, nothing to join on), so the field now says it is context rather than implying it is a gate. Separately, the PASS gate's two byte-identical correctness blocks collapse into one `$ref` — 74 duplicated lines in the clause that decides what earns a clearance.

**`b3532864` — name the channel the shape grid actually travelled on.** `--shape-arg` was added in `d02515f6`, but the code around it still assumed the grid could only be an environment variable. `GRID_HOOK_OK` was assigned twice, so supplying both flags proved the env var while `run_pytest` put the shapes on argv — the report attested a channel the run did not use. A `--shape-arg` naming a flag the target does not accept fell through to the branch reporting "no shape grid was configured", which is an environment fact standing in for a capability gap. And `run_pytest` took `"$SHAPE_ENV=$GRID"` and re-split on the first `=`, so a CLI-only run worked only because an unset `SHAPE_ENV` left a leading `=` the split then removed.

---

## Verification

**Round 3 (this round), on the checked-in regression suite:**

| check | result |
|---|---|
| `tests/test_validator.py` | 40 passed |
| new test against the **pre-fix** script | fails on `'no configured shape override' unexpectedly found`, i.e. on the reported reason, not merely on the field added to carry it |
| PASS gate after the `$ref` collapse | a clean `PASS` report validates; `executed=0`, `failures=1`, `errors=1`, `exit=1`, `status=skip`, missing `stats`, and `stats` missing `skipped` are each still rejected on **both** correctness stages |
| schema backward compatibility | a report with no `grid_channel` and a null `source_sha` still validates |
| `black` / `ruff check` | clean on the changed files |

**Round 1, on #4538 (gfx950, MI355 OAM), base `89178864`:**

| check | result |
|---|---|
| script target with `--shape-arg -s` | `PASS` 9/9, exit 0, review-pr consumable, `idleness_basis: activity+vram` |
| receipt contents | route `…fp8_mqa_logits:_auto_variant` over the four injected production shapes |
| `--shape-arg` naming a flag the target does not accept | `correctness_s1_grid: skip`, receipt still `pass` |
| **seeded defect** (fused `-inf` fill dropping `[e, seq_len_kv)`) | `BLOCK`, exit 1, with `correctness_s1_grid` failing **independently** of the PR's own suite |

The seeded-defect row is the one that matters: the injected grid caught the defect on its own, so the stage has real discriminating power rather than being decoration. Note that round 3 found the "flag the target does not accept" row above was reporting the right verdict for the wrong stated reason — the run skipped, but told the reader no grid had been configured. That is fixed in `b3532864`, and the new test pins it.

## Known gaps, not addressed here

- The picker's `amdsmi` import fallback covers `/usr/lib/python3/dist-packages` and `/usr/lib/python3.X/dist-packages`, but not `/usr/local/lib/python3.*/dist-packages`, where ROCm 7.2.4 ships the bindings — so from inside a virtualenv the picker can exit 2 unless `PYTHONPATH` or `PYTHON_BIN` is set.
- The executor sets `AITER_TRITON_ONLY=1`, which guarantees `native_artifacts: []` for aiter and excludes the CK/HIP op path from every run. Fine for FlyDSL/Triton PRs; a scope limit that belongs in "What this skill does not do" and is not yet written there.
- Choosing the right `--target` from a diff is still the unsolved part. An irrelevant target can still produce `PASS`; the report names the target so a reviewer can reject that evidence, but the executor cannot decide relevance itself.
- Remaining redundancy that is noise rather than wrong, deliberately left for a follow-up so it does not bury the fixes above: `tests/test_validator.py` copy-pastes a 12-line git-commit block seven times and builds the same `SCALE=1.25` perf patch three times while bypassing the helper that exists for it; `validate_pr.sh` has two `elif`/`else` branches with identical bodies and extracts one JSON field three times in two styles; `--tests` is an alias with no in-tree caller.

Two earlier known gaps are now closed: `pick-idle-gpu.py` has its shebang, and `validate_evidence.py` no longer hardcodes `source_sha: None`.


zufayu:
Reviewed the delta over `706eafafc` with the `review-pr` on `main`, so the PR isn't grading itself.

For the description: `--shape-arg` takes the S1 grid from **0 of 125** aiter op_tests reachable (none read a shape env var) to **11** (`--shape`, `--shapes`, `--input_shapes`). The difference between a stage that never runs and one that does.

🔴 `scan_index_width.py` (unchanged from #4870) misses what it was written for. On the diff that introduced `moe_wgrad` in #4978 — the overflow #5132 is fixing — it reports one candidate, `off_expert * stride_dwe`, already `int64` on line 102, and misses both real lines. The operand regex cannot span `[:, None]`, and a hit still requires `INDEXY`, a hard-coded name list — the name dependence moved from the trigger to the operand rather than going away. `SKILL.md:654` then has the reviewer assert the list is complete. **Author must** fix the operand match and drop the name-list dependency.

⚠️ 340 new lines, no new tests. `tests/` is byte-identical to #4870; `read_activity`, `idleness_basis`, `shape_arg`, `PATCH_APPLIED`, `run_script_with_probe` occur 0 times in the 806-line suite.

⚠️ `validate_pr.sh:899` sets `GRID_HOOK_OK` from `--shape-arg`; `:918` overwrites it unconditionally. Supplying both flags drops the working CLI channel and reports the wrong reason.

zufayu:
**A separate note on why misses like these happen — not a merge blocker, and nothing here should hold up this PR.**

Step 3 already maps PR type to the exact rules that apply, but it is prose the model self-applies. Step 1 already runs bash — let it derive the type from the diff and emit only the matching rules. Paths give Triton / FlyDSL / tuning-config / tests-only, a grep of added lines gives new `@compile_ops` and FP8, and `scan_index_width.py` already decides the index-arithmetic case. A typical PR matches three or four types, so the model reads ~10 rules instead of 44, at 1x cost.

Two that pair with it for free: any aggregate ships with a sample of the underlying rows, and each rule returns `n/a` / `cleared` / `fired` plus a resolvable `file:line`. Both are aimed at errors I made while writing this review — a claim taken from an earlier round's description instead of the current file, and a recall figure whose denominator counted assertion strings.

Unmeasured, n=1 from one session. Worth scoring before acting on.

jhinpan:
Opened **#5174** against this branch — the three review items above, plus eleven defects found by applying the skill to six real open FlyDSL kernel PRs. It targets `zhiding512/validate-kernel-pr-amdsmi-fix` rather than `main`, so merging it lands the work inside this PR instead of competing with it. No push access to ROCm/aiter, hence the fork.

**@zufayu's items.** The scanner is now an AST pass with both name lists deleted: on aiter#4978 it reports 4 of 4 real `moe_wgrad` overflow lines where the old one reported 0 and emitted an already-`int64` line as its only candidate, and it reports 0 on #5132, the fix. The operand regex could not span `[:, None]`, which is the standard Triton pointer-arithmetic idiom, so the miss was structural rather than incidental. `GRID_HOOK_OK` is split into independent probes. 27 tests added, suite at 55, all green, every test checked against the pre-fix sources.

**What running it changed about the diagnosis.** Four consecutive customer PRs reached INCONCLUSIVE for one reason: `correctness_s1_grid` alone forces it, and the stage was inert on this repository's dominant test shape — 0 of 7 files in `op_tests/flydsl_tests/` expose either existing channel. That is the same measurement as `--shape-arg`'s 0/125 → 11/125, from the other side. `--shape-argnames` adds pytest parametrization as a third channel under the same burden of proof.

**Four stages were reporting `pass` on empty evidence** — `executed_shapes: []`, `tolerances_head: []`, `idleness_basis: "unknown"` beside a concrete activity number, `native_artifacts: []`. Each was a status decided by "this code path completed" rather than "evidence was collected". The tolerance one mattered most: a file declaring `DEFAULT_REL_TOL` read as having no tolerances, which made the m2 mutant undetectable. None of these are visible to a static review — they do not exist until the tool runs.

**Four of the eleven were mine**, introduced by the new channel and closed in the same series; all four published a fact about the run as a defect of the code under test, including a BLOCK against an author whose own 138 tests passed in the same report. They are written up in #5174 rather than folded away, because the discipline that caught them is the transferable part: each fix was graded by re-running the case that had produced the decisive verdict, not by finding an invocation that passes.

**Held out:** #4885 and #5025 were never used while fixing. #4885 BLOCKs on a real merge-blocking defect (`flydsl.expr.buffer_ops` eager import breaks `import aiter.ops.flydsl` on head) and still does on the final tree; #5025 PASSes, with the grid running three widths its own suite never tests. #5025 also found the scanner claiming kernel scope while visiting every Python function.

Known limits, `grid_channel_reason` coverage, and the two measurements taken mid-edit that should not be quoted are all listed in #5174. Happy to split it if you would rather take the scanner and the honesty fixes separately from the new channel.

zufayu:
Re-reviewed at `a8d46dc`. CI 56/56 clean, #4870 closed, my 08-31 🔴 is resolved — the scanner is an AST pass now, both name lists gone, `[:, None]` handled.

**Don't merge ahead of #5174.** `validate_evidence.py:94` is `cross_namespace = grid_channel == "pytest"`, gating the relaxation at `:104`. Nothing on this branch can set it: `report_schema.json:357` allows only `["cli-flag", "env-var", "cli-flag-unusable", "none"]`, and `validate_pr.sh:953-975` assigns nothing else. So the misreport it fixes — `missing required shapes` on a run whose every grid case passed — still fires. #5174 is what makes the line live (adds `shape_grid_plugin.py`, widens the enum to `["", "cli", "env", "pytest"]`, sets `GRID_CHANNEL="pytest"`).

**Please freeze this branch until #5174 lands on it.** #5174 has gone dirty twice: its base is `dd6ff6a71`, part of it is already cherry-picked here, so every push re-conflicts the rest.

**Four Copilot comments are false positives**, resolvable as-is:
- `source_sha` "hardcoded null" (`validate_evidence.py:241`, `SKILL.md:341`) — `def source_sha` is at `:166` and `runtime_identity()` calls it.
- `review-pr/SKILL.md:578` "still requires pytest" — `:576` is already `in {"pytest", "script"}`.
- `validation_probe.py:66` "leaks `pending_frames`" — profile functions get `'return'` on exception unwind; `'exception'` is trace-only.

zufayu:
Re-reviewed at `ee68f943` now that #5174 has landed. **The objection I raised earlier is resolved:** `report_schema.json` now allows `grid_channel` of `["", "cli", "env", "pytest"]`, so `cross_namespace` in `validate_evidence.py` is reachable and the shape-containment relaxation actually fires. CI is clean — 14 success, 18 skipped, no failures.

The remaining blocker is mechanical: 15 unresolved Copilot threads and no approving review. Here is where each thread stands, so they can be cleared without re-deriving the analysis.

**Safe to resolve — verified false positives (4):**
- `validate_evidence.py:241` and `SKILL.md:505`, "`source_sha` hardcoded to null" — `def source_sha` exists and `runtime_identity()` calls it as `"source_sha": source_sha(module_path)`.
- `review-pr/SKILL.md`, "acceptance snippet still requires `runner == "pytest"`" — it already reads `in {"pytest", "script"}`.
- `validation_probe.py`, "ignoring the exception event leaks `pending_frames`" — CPython delivers `'return'` to a profile function when a Python frame unwinds on an exception; `'exception'` is trace-only. The `pop` runs.

**Safe to resolve — fixed or harmless (2):**
- `validate_evidence.py:94` `cross_namespace` — fixed by #5174, as above.
- `report_schema.json` `runner_risk` defined twice (405 and 465) — real, but the two definitions are byte-identical in `type` and `description`, so the overwrite changes nothing. Worth deleting one; not worth a round of CI.

**Real, not fixed, and not carried by #5174 (2).** Neither affects whether the validator reaches a correct PASS, so I do not think they should hold this PR — but they should not be lost either:
- `run_script_with_probe.py:57` catches `SystemExit` only. An uncaught `AssertionError` — the standard shape of aiter's `op_tests/*.py` — leaves `status` at `0`, and `finally` writes the receipt before the exception propagates, so a red run lands a receipt claiming `pytest_exitstatus: 0`. Evidence quality on failure, not correctness on success.
- `validate_pr.sh:875` derives `kernel_files_changed` from `git diff --name-only HEAD`, and the script has no `git add` or `--intent-to-add` anywhere, so files `git apply` introduces as untracked are invisible to the test-policy analyzer. A PR adding a new kernel can read as touching no kernel code.

**Not yet verified (7)** — multi-route `--expected-route` handling, the per-phase shared `execution-receipt.json` path, `python3` used directly while `TARGET_PYTHON` may fall back to `python`, `sys.setprofile` installed when `routes` is empty, `parse_diff._pending` across file boundaries, and the `unannotated_runtime_parameters` key emitted by `scan_index_width.py:546` with no counterpart in the schema. I have not read enough of each to call them either way.

Suggest merging on the strength of the above and opening one follow-up PR for the two confirmed defects plus whatever survives review of the seven. Two rebase races have already cost this branch a day; a third to land two evidence-quality fixes is a bad trade.

zufayu (approved):

zufayu:
It can be merged once the required CI passes.

PR #4813, opened 2026-08-18
[HIP] [JIT] Fused MiniMaxM3 QKNorm+RoPE+CacheInsert

weitliao:
Introduces minimax_m3_qknorm_rope_cache_shuffle_insert, a single HIP kernel that replaces the three-kernel sequence (fused QK-norm/RoPE -> reshape_and_cache(asm_layout=True) -> index-cache scatter) used by the MiniMax-M3 sparse attention layer, writing K/V directly into the page-16 SHUFFLE layout consumed by pa_decode_gluon.

Includes correctness tests against a torch reference and, where vLLM is available, end-to-end equivalence tests against the existing unfused path.

## Motivation

The current 3-kernel path adds launch overhead and extra memory traffic (K/V written to `qkv`, then read again for cache insert). Fusing into one kernel reduces both on a hot prefill/decode path.

## Technical Details

- **Input:** packed `qkv` row per token: `[q | k | v | index_q | index_k]`, `head_dim=128`
- **Processing:** Gemma RMSNorm + partial NeoX RoPE on Q/K/index_q/index_k; V passes through unchanged
- **Outputs:** `q_out`, `index_q_out`, and paged `k_cache` / `v_cache` / `index_cache` in page-16 SHUFFLE layout
- **Also supports:** FP8 KV/index cache, padded slots (`slot_mapping=-1`), `skip_index_branch=True`
## Test Plan

pytest op_tests/test_minimax_m3_fused_qknorm_rope_cache_shuffle.py -v
pytest op_tests/test_minimax_m3_fused_shuffle_vs_unfused.py -v  # requires vLLM

## Test Result

- Platform: linux, Python 3.12.13, pytest 9.1.1
- Command: `pytest op_tests/test_minimax_m3_fused_qknorm_rope_cache_shuffle.py -v`
- Result: **14 passed** in 20.78s
| Test | Cases | Result |
|---|---|---|
| `test_fused_insert` (tokens × fp8_index × quantized) | 1/7/64 × {False,True} × {False,True} (12) | PASSED |
| `test_skip_index_branch_leaves_index_outputs_untouched` | 1 | PASSED |
| `test_multi_kv_head_and_full_rotary` | 1 | PASSED |
## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5143, opened 2026-08-31
[HIP] Extend fused QK norm for MiniMax-M3

weitliao:
## Motivation

Extend the existing `fused_qknorm_idxrqknorm` operator from [#4362](https://github.com/ROCm/aiter/pull/4362) with the MiniMax-M3 behavior explored in [#4813](https://github.com/ROCm/aiter/pull/4813), without introducing a duplicate public operator.
The goal is to fuse Q/K normalization, RoPE, KV-cache insertion, and optional index-cache processing for vLLM’s MiniMax-M3 ROCm path, and to emit the unit-scale e4m3 `index_q` that [#4787](https://github.com/ROCm/aiter/pull/4787)'s score kernels require.

**This PR is rebased on [#4787](https://github.com/ROCm/aiter/pull/4787)** (`e53569d5`). Until #4787 merges, the GitHub diff vs `main` includes those MSA score/top-k files. This PR does not modify them. The commits that belong here are the three after `e53569d5`.
Depends on [#4787](https://github.com/ROCm/aiter/pull/4787).

## Technical Details

- Added compile-time index-branch skipping through `skip_index_branch`.
- Added `fp8_e4m3_static` for fixed device-scalar K/V input scales.
- Preserved existing per-token FP8, unit-scale FP8, dense, sparse, and page-128 behavior.
- Added support for packed LBHNC page-16 SHUFFLE caches where K and V have unequal physical page spans.
- Decoupled index-cache validation from packed K-cache physical capacity.
- `index_q_out` may be qkv dtype **or** fp8 e4m3. The fp8 path writes **unit-scale** e4m3 (`scale=1.0`), matching #4787's `q_idx` contract (no scale tensor). Main KV static/per-token scales are unchanged.
- IQ dtype is independent of the main KV cache, so fp8 index + bf16 KV still works. Mixed bf16 `index_q` + fp8 index-K is preserved via a separate fp8 IQ pointer.
- When `index_cache_dtype` is omitted, infer it from `index_cache` / `index_q_out`, not from `kv_cache_dtype`.
- Capability markers for downstream probes: `FUSED_QKNORM_IDXRQKNORM_SUPPORTS_PACKED_SHUFFLE` and `FUSED_QKNORM_IDXRQKNORM_SUPPORTS_FP8_INDEX_Q`.
- Kept `fused_qknorm_idxrqknorm` as the only public API.

## Test Plan
- `op_tests/test_fused_qknorm_idxrqknorm.py --minimax_tp4` (bf16 and fp8 index, full and skip-index, packed SHUFFLE).
- `op_tests/test_msa_block_select.py -H 1 -b 4 -c 4096 -q 1` to confirm #4787 still scores after the rebase.

## Test Result

Both passed on gfx950. Fused `index_q` / index-K in e4m3 is accepted by `pa_sparse_block_score_decode`.

## Submission Checklist


ChuanLi1101 (approved):
LGTM with follow-ups, not merge blockers.

Kernel logic looks correct: skip-index vs packed qkv layout, fp8_e4m3_static, packed LBHNC SHUFFLE views, and unit-scale e4m3 index_q all hang together, and standard AITER CI is green.

Please follow up:
- Run the MiniMax cases in default CI (`--minimax_tp4` is currently opt-in and also skips DEFAULT_CASES).
- Infer `index_cache_dtype=None` from `index_cache` only, not from `index_q_out`.
- Packed SHUFFLE still trusts the caller on K/V page spans; a mismatched separate SHUFFLE pair can now OOB.

zufayu (approved):
Reviewed the diff and the CI state. Findings from my earlier pass are resolved or accounted for:

- The `ci:vllm` run covers this PR's actual target — `vLLM Benchmark (amd/MiniMax-M3-MXFP4)` passed twice, alongside DeepSeek-V4-Pro and Kimi-K2.5 on both sglang and vllm backends. That closes the downstream-coverage gap I raised.
- I withdrew two findings after reading the head implementation rather than the diff hunk. The dropped `assert q_out is not None` / `assert index_q_out is not None` are an intentional relaxation, not an omission: both are `std::optional` on the C++ side, every call site is `has_value() ? data_ptr() : nullptr`, and the kernel guards with `if(is_q && q_out != nullptr)`. Separately, `fp8_e5m2` cannot reach the changed `use_per_token_kv_scale` whitelist — the `AITER_CHECK` at the top of the fp8 insert path admits only `fp8` / `fp8_e4m3` / `fp8_e4m3_unit` / `fp8_e4m3_static`.
- The `index_cache` capacity check is not removed, only replaced in the packed page-16 branch; the page-128 branch still carries the original `numel() >= num_blocks * block_size * kHeadDim`.

`Flash Attention - Triton / MI35X` is red and unrelated to this PR. The same six `2048-2048-32` cases (mha/mqa/gqa x True/False) abort with `Fatal Python error: Aborted` / exit 134 on the `opus_kernel_workspace_management` branch, which shares nothing with this change. That workflow tracks Dao-AILab upstream `main`, so the regression rides in from there. It is also not a required check — five of the last six commits touching `aiter/__init__.py` landed on main with this job red or cancelled, including #4787, this PR's own dependency.

One item worth a follow-up answer, not a blocker: `scale_page_span = min(kv_cache_k->size(0), kv_cache_v->size(0))` bounds the per-token scale allocation, but `scale_idx` is derived from `mapped_slot / block_size` — a logical page id from `slot_mapping`, not a physical span. The two coincide only under an invariant that is not written down anywhere, and the test allocates K at 48 pages against V at 40. Worth confirming the logical page domain is bounded by the shorter span.

PR #5146, opened 2026-08-31
[FlyDSL] Buffer ops refactoring

amd-nprotaso:
Replace old rocdl api ussage to new fx.


aiter/ops/flydsl/kernels/buffer_ops.py
```diff
@@ -168,6 +141,53 @@ def _ptr8_to_v4i32(ptr8_val) -> ir.Value:
...
+def _as_num_records(num_records_bytes) -> fx.Int64 | None:
+    """Normalize a descriptor byte count to ``fx.Int64`` for ``make_buffer_ptr``."""
+    if num_records_bytes is None:
+        return None
+    if isinstance(num_records_bytes, int):
+        return fx.Int64(max(0, min(int(num_records_bytes), _MAX_NUM_RECORDS)))
+    if isinstance(num_records_bytes, ir.Value):
+        if isinstance(num_records_bytes.type, ir.IndexType):
+            num_records_bytes = _unwrap_value(
+                std_arith.IndexCastOp(T.i64(), num_records_bytes).result
```
jli-melchior: could we use fx.Int32/fx.Int64  here to replace arith.index_cast,  we want to avoid the use of arith dialect, also other side we can use fx Type as much as possible

amd-nprotaso:
Kernels will be refactor to use layout API, instead of use low level buffer operations

PR #5116, opened 2026-08-30
[Triton/Gluon] [HIP] [CK] Remove obsolete availability helpers

coderfeli:
## Summary
- Remove `is_flydsl_available`, obsolete FlyDSL import fallbacks, and the duplicated `aiter.ops.flydsl.utils` layer now that FlyDSL is required; preserve lazy public imports for unsupported architectures.
- Centralize architecture LDS limits in `aiter.jit.utils.chip_info`; tuners and capacity checks no longer depend on retired `flydsl.utils.smem_allocator` metadata/check APIs.
- Move the GDR tuned CSV and MXFP4 v2 tuner helper into their canonical directories; delete unreferenced legacy config data.
- Move and standardize the GDR decode test under top-level `op_tests/` so standard CI discovers it.
- Drop the redundant standalone A4W4 test; `op_tests/test_moe_2stage.py -q 4` already covers the production path and isolated stage1/stage2 timing.
- Delete the superseded split-K HGEMM test rather than relocating it.
- Derive every LDS budget from the resolved arch instead of a hardcoded `gfx950`/`gfx1250` string, and drop the local `get_gfx()` wrapper now that `chip_info.get_gfx()` already honours `GPU_ARCHS`. Tuner modules resolve via `get_gfx()`; kernel builders resolve via `get_rocm_arch()` so AOT cross-compilation under `FLYDSL_GPU_ARCH` still sees the target arch's capacity.

## Compatibility
- Intentionally removes the obsolete public `is_flydsl_available` probe and unused `buffer_ops.extract_base_index` export.
- Semantic `flydsl_config is None` checks remain because they detect stale tuned CSV kernel names, not import failures.
- Guards remain only for genuinely optional external components such as Mori, VLLM, and optional HIP helpers.
- Legacy `SmemAllocator/SmemPtr` kernel migrations are deferred to a separate structural follow-up.

## Test plan
- [x] `python3 -m compileall -q aiter op_tests csrc`
- [x] Ruff, Black, IDE lints, and `git diff --check`
- [x] Related pytest suites pass on gfx950, including GDR, MLA dispatch, causal-conv, tuner guards, GEMM padding, and FlyDSL prefill smoke
- [x] Tuner kernel lists are unchanged after the LDS-limit refactor: gfx950 8wave 36 / preshuffle 2312, gfx1250 wmma 1420 / mxfp8 wmma 1156 (identical before and after)
- [x] `test_moe_2stage.py -q 4` A4W4 production path passes and reports stage1/stage2 timing
[...]

PR #5148, opened 2026-08-31
[HIP] [CK] [FlyDSL] FlyDSL split-K preshuffle decode GEMM for small-M shapes (FP8 + MX, gfx950/gfx942)

johannes-graner:
## Summary

Adds a FlyDSL split-K preshuffle GEMM tuned for decode (small-M) shapes, covering FP8 per-tensor and FP8 blockscale, plus a direct-output (split_k=1) epilogue for the whole scale-mode family. Adds MX datatype support (mxfp8, mxfp4) as correctness-validated, tunable dispatch candidates. Enables the blockscale and epilogue split-K kernel families on gfx942 in addition to gfx950. Rides along: a rework of the existing FP8 decode-GEMM dispatch/layout/AOT plumbing that this work shares infrastructure with.

## What's included

- `gemm_decode`-family split-K preshuffle GEMM kernel (`aiter/ops/flydsl/kernels/preshuffle_gemm_splitk.py` + reduce/op helpers), wired into dispatch, the tuner, and AOT prebuild for both FP8 per-tensor and FP8 blockscale.
- Direct-output single-pass (`split_k=1`) epilogue, validated per scale mode against the FP32 oracle.
- mxfp8 (128-block E8M0, hardware-scaled MFMA) and mxfp4 (4-bit packed, 32-block E8M0) kernel support, wired as a third scale mode symmetric to existing per-tensor/blockscale dispatch.
- gfx942 enablement for the blockscale and epilogue split-K families via a software-dequant path (4x standard MFMA per 128-K scale block in place of the gfx950-only scaled MFMA); mx128 stays gfx950-only (needs a hardware scale operand gfx942 lacks).
- New `op_tests/flydsl_tests` coverage for the decode GEMM, split-K preshuffle GEMM, and small-M HGEMM kernels.
- Shared DRY infrastructure: a common split-K bpreshuffle dispatch helper, a shared tuner base class, and a shared kernel-fit check reused across the a8w8/a4w4 families.

## Performance

Measured on gfx950, M in {1, 16}, N in {2048, 16384}, K=7168. "Best existing" is the best-tuned pre-existing CK/asm/FlyDSL incumbent for that shape, not a cold-dispatch heuristic fallback. Fuller benchmark data across more shapes is tracked separately.

**FP8 per-tensor** (this PR's split-K decode GEMM vs best-tuned existing kernel):

| M | N | GB/s (this PR) | % of HBM bandwidth wall | speedup vs best existing |
|---|---|---|---|---|
| 1 | 2048 | 1974 | 55% | 1.54x |
| 1 | 16384 | 5760 | 98% | 1.05x |
| 16 | 2048 | 1728 | 48% | 1.50x |
| 16 | 16384 | 5691 | 97% | 1.03x |

**FP8 blockscale** (best existing -> this PR):

| M | N | best existing GB/s | this PR GB/s | speedup |
|---|---|---|---|---|
| 1 | 2048 | 632 | 2106 | 3.33x |
| 1 | 16384 | 4394 | 4620 | 1.05x |
| 16 | 2048 | 636 | 1917 | 3.01x |
| 16 | 16384 | 4247 | 4531 | 1.07x |

Gains are largest at small N, the latency-bound decode regime; at large N both tables converge toward parity as the kernel approaches the HBM-bandwidth wall.

## Correctness

Measured on gfx950, current `main` + FlyDSL 0.3.2. The tracked FlyDSL pytest suite passes (17/17, 0 skipped) against a torch FP32 reference. mxfp8 is exact across the tile/split sweep. mxfp4 is bit-exact on the identity case and reaches ~3e-3 max relative error on real data, with a negative control that correctly diverges.

## Notes

- MX (mxfp8, mxfp4) is correctness-complete and wired as tunable dispatch candidates; performance tuning is deferred and no tuned configuration rows are shipped in this PR.
- gfx942 support is newly enabled in this PR, validated with a single measured winning shape out of the decode shape set tested — an enablement, not a performance headline.


andyluo7:
Hi @johannes-graner — while validating this PR against current `main` (`0598b04f`, which contains #5116), I found a concrete rebase integration issue.

Several files introduced or changed by #5148 still import `aiter.ops.flydsl.utils`, but #5116 removed `aiter/ops/flydsl/utils.py`. In an exact integration tree, `import aiter` succeeds, but this fails:

```bash
python3 -c "from aiter.ops.flydsl.utils import is_flydsl_available"
# ModuleNotFoundError: No module named 'aiter.ops.flydsl.utils'
```

This also affects downstream capability checks and some of the new small-M/split-K paths. The PR head still contains `utils.py`; current main does not. For local gfx950 Hy4 profiling I restored the PR-head helper as an isolated, hashed integration repair, but I have not pushed anything to your branch.

When rebasing, could you either retain the helper or update all #5148-added imports to the post-#5116 `chip_info` / direct FlyDSL availability pattern? I can share the exact integration diff and runtime probe results if useful.

johannes-graner:
Hi @andyluo7,
Thanks for the heads-up, I'll refactor to use the new patterns.

andyluo7:
Hi @johannes-graner — one more downstream integration point from Hy4-preview FP8 TP8/C1 validation on gfx950.

vLLM's correctness-preserving path splits the routed and shared experts because they use different activation/clamp semantics. That changes the small-M MXFP8 FMoE lookups from the existing combined MiniMax-M3 rows (`expert=129, topk=5, inter_dim=768/384, Swiglu`) to these two generic call shapes:

- routed: `model_dim=6144, inter_dim=256, expert=256, topk=8, Silu`
- shared: `model_dim=6144, inter_dim=256, expert=1, topk=1, Silu`
- token tiers: `1, 2, 4`
- `gfx950, cu_num=256, bf16 output, MXFP8 per_1x32`

I tuned those six rows against the #5148 kernel family. The resulting CSV is SHA-256 `1f05eadfb3654b7d4adb059f069fe1d0de7d4ce24d9e8cb3e1b6f23699b92534`. Preliminary end-to-end vLLM TP8/C1 graph-mode runs using it are about 19.29-19.39 ms mean ITL, versus 25.562 ms on matched gfx942 (~1.32x); the clean three-repeat rerun is queued behind another user's workload.

Current #5148 head `f8500b854` has the existing combined rows in `aiter/configs/model_configs/minimax_m3_mxfp8_tuned_fmoe.csv`, but not these split-call rows, so downstream performance currently requires an external `AITER_CONFIG_FMOE` file.

Would you be comfortable including the six tuned rows in that existing per-model table after I send the clean repeat/profile evidence, or do you prefer another AITER-owned configuration location? I will not hard-code these shapes in vLLM. I can provide the exact rows and per-kernel measurements once the uncontaminated rerun completes.

johannes-graner:
Hi @andyluo7,

Of course I can include the tuned rows. Either send them to me, or add them to the tuned csv and push directly to this branch, either is fine by me.

PR #4789, opened 2026-08-17
[Config] update gemm stage2 v2 kernel

charlieguo1106:
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4556, opened 2026-08-04
[PERF] Add opt-in atomic stage2 override

vorapolsiloai:
## Summary

Add an opt-in `AITER_FLYDSL_FORCE_ATOMIC=1` override that replaces an eligible tuned FlyDSL stage-2 `_reduce` kernel with its registered `_atomic` sibling while constructing MoE metadata.

Safety and precedence are explicit:

- unchanged when the override is unset;
- unchanged when no registered atomic sibling exists;
- ignored for non-FlyDSL, already-atomic, or otherwise ineligible kernels;
- `AITER_FLYDSL_FORCE_REDUCE=1` takes precedence when both variables are set;
- output allocation and MoE sorting use the final metadata, so accumulation semantics match the selected epilogue.

## Motivation and measured result

Using [#4554](https://github.com/ROCm/aiter/pull/4554)'s Qwen3-VL tuning rows on 8× MI355X, 1024 input/output tokens, concurrency 4:

- tuned `_reduce`: **375.19 output tok/s**, 889.47 total tok/s;
- forced `_atomic`: **391.46 output tok/s**, 928.27 total tok/s;
- reference range: 389.48–391.05 output tok/s, 924.04–928.77 total tok/s.

The opt-in override recovers approximately **4.3% output throughput** for this point while retaining safe fallback behavior.

## Qwen3-VL integration stack

- [vLLM #50212](https://github.com/vllm-project/vllm/pull/50212): fused attention prologue; requires AITER main containing the strided-cache support merged via #4531.
- [AITER #4531](https://github.com/ROCm/aiter/pull/4531): **merged** into AITER main as [`b0cc5392`](https://github.com/ROCm/aiter/commit/b0cc5392dd838a4bec46a5553e7fbeed60218ab5); provides full strided MRoPE KV-cache support.
- [AITER #4554](https://github.com/ROCm/aiter/pull/4554): tuning rows used by the measurement above.
- **This PR — [AITER #4556](https://github.com/ROCm/aiter/pull/4556):** optional atomic stage-2 override.
- [AITER #4761](https://github.com/ROCm/aiter/pull/4761): independent Triton unified-attention prefill/decode optimization used by the later Image C stack.

## Rebase provenance

- Frozen AITER `main`: `fc2e5d57fb5b8ad8e7e23f7103071dde798ea618`
- Current PR head: `5390e8912bcfd58f6132a30428dcb5a2bf7744de`

## Validation

[...]

PR #5155, opened 2026-08-31
[Config] Retune Qwen3-VL MXFP4 MoE without stage-2 reduce kernels

vorapolsiloai:
## Summary

- Replace the initial manual substitution of 32 `kernelName2` values with a generated 60-shape MoE retune that refreshes both kernel stages and all generated timing metadata.
- Consolidate the result into the existing canonical `qwen3_vl_fp4_tuned_fmoe.csv`, retain its four non-overlapping TP1 tiers, expand the matching untuned manifest, and remove the duplicate `q3vl_fp4_tuned_fmoe.csv`.
- Merge current `main` and resolve the overlap with [#4789](https://github.com/ROCm/aiter/pull/4789) per row, keeping whichever side's tuner measured the lower time for each shape.
- Keep kernel selection at the model-config level, as requested in the review of [#4556](https://github.com/ROCm/aiter/pull/4556); this PR does not add or require a runtime atomic override.
- Exclude FlyDSL `_reduce` kernels from the tuner candidate space while leaving atomic and CKTile candidates eligible.

## Motivation and duplicate-config resolution

[#4556](https://github.com/ROCm/aiter/pull/4556) proposed forcing atomic stage-2 kernels at runtime. Maintainer feedback recommended expressing the model-specific choice in the tuned configuration instead. The first version of this PR did that by replacing 32 eligible stage-2 kernel names while retaining the old tuning metadata.

This revision addresses the follow-up feedback to update the complete tuner-generated config: all 60 `q3vl` shape keys were retuned under the non-reduce constraint rather than changing `kernelName2` after tuning.

During serving validation, I also found that two model-config files described the same Qwen3-VL MXFP4 workload. They overlapped on 12 TP1 runtime keys. The production config merger keeps the first matching key, so the longer-named `qwen3_vl_fp4_tuned_fmoe.csv` shadowed the retuned `q3vl` rows at TP1. That made the original TP1 serving result invalid as evidence for those 12 replacements: it had measured the old effective kernels.

The committed layout now has one source of truth:

- `qwen3_vl_fp4_tuned_fmoe.csv`: 64 unique runtime keys (all 60 retuned `q3vl` keys plus four non-overlapping existing TP1 tiers).
- `qwen3_vl_fp4_untuned_fmoe.csv`: the same 64 unique runtime keys.
- `q3vl_fp4_tuned_fmoe.csv`: removed.
- Production-merge validation resolves exactly 64 Qwen3-VL keys with no cross-file collision or order-dependent selection.

## Tuner provenance

The retune ran the AITER FMoE tuner on MI355X (`gfx950`) with four visible GPUs and the complete set of 60 `q3vl` input shape keys:

```bash
TUNE_MOE_KERNEL_REGEX='^(?!.*_reduce)' \
TUNE_MOE_EXPERT_BALANCE=False \
AITER_FLYDSL_FORCE=1 \
AITER_FLYDSL_FORCE_ATOMIC=0 \
AITER_FLYDSL_FORCE_REDUCE=0 \
python3 gemm_moe_tune.py \
  -i q3vl_fp4_full.csv \
  -o q3vl_fp4_tuned_fmoe.csv \
  -o2 retuned_all_candidates.csv \
  --all --mp 4 --batch 5 --sort True \
  --timeout 1800 --warmup 5 --iters 101
```

Generated-output validation:

- 60/60 input shape keys are present and unique.
- 540 candidates were profiled; no `_reduce` candidate passed the candidate filter.
- Generated 60-row stage-2 winners: 57 atomic and 3 CKTile; 0 reduce.
- Generated output SHA-256: `46a2ec113dfd647990ef43ea64e64cb687e07bec5fddd945cd5ae238dcd5ddb7`.

Relative to the previous PR config (`baea2c37...`):

- `kernelName1` changed for 57/60 rows and `kernelName2` changed for 57/60 rows.
- The selected kernel pair changed for 60/60 rows, and all 60 generated timing records were refreshed.
- Stage-2 composition changed from 56 legacy atomic + 4 CKTile to 42 layout atomic + 15 legacy atomic + 3 CKTile.
- 39 rows moved from legacy atomic to layout atomic. The result is therefore a paired retune of both stages, variants, launch metadata, and timings—not a post-tuning kernel-name replacement.

After consolidation and the `main` merge below, the committed 64-row canonical config has 61 atomic (47 layout + 14 legacy) + 3 CKTile stage-2 selections and no reduce selection. Exact committed checksums:

- Tuned canonical: `cc107c0aaf8d2526045044dedc11e62c6e53dd58cdc965659776a4957f44263e`
- Untuned manifest: `4976668431c73ebacbe87c7e19730fcb33876cfaab19656ae0cb952352f5928c`

## Merge resolution against #4789

[#4789](https://github.com/ROCm/aiter/pull/4789) landed on `main` while this PR was open and retuned all 16 TP1 `E=128` rows of `qwen3_vl_fp4_tuned_fmoe.csv` -- the same file this PR rewrites -- so the two conflicted on 12 rows.

Rather than taking either side wholesale, the merge resolves the file **per shape**: for each contested key, keep whichever side's tuner recorded the lower total time. Each winning row is copied **whole** (kernel names, `block_m`, `ksplit`, `us1`/`us2`/`us`, error columns), so no row mixes fields from two different tuning runs. Both sides' headers are identical and neither selects a `_reduce` stage-2 kernel, so the atomic-only property of this PR is preserved.

Outcome over the 16 shared shapes:

| shape (token, E=128) | #4789 | this branch | kept |
| --- | --- | --- | --- |
| 1 | 22.43 us | 23.01 us | #4789 |
| 2 | 31.09 us | 35.50 us | #4789 |
| 4 | 54.49 us | 60.82 us | #4789 |
| 8 | 108.68 us | 93.35 us | this branch |
| 16 | 207.68 us | 137.41 us | this branch |
| 32 | 209.31 us | 180.43 us | this branch |
| 64 | 208.74 us | 206.01 us | this branch |
| 128 | 211.50 us | 206.79 us | this branch |
| 256 | 218.25 us | 211.57 us | this branch |
| 512 | 224.26 us | 295.64 us | #4789 |
| 1024 | 246.49 us | 235.12 us | this branch |
| 2048 | 370.47 us | 335.22 us | this branch |
| 4096 | 637.36 us | 565.67 us | this branch |
| 8192 | 1110.31 us | 1505.91 us | #4789 |
| 16384 | 2031.31 us | 2112.24 us | #4789 |
| 32768 | 4005.44 us | 4065.76 us | #4789 |

The merged file therefore takes 7 rows from #4789 and 57 from this branch (9 contested winners plus the 48 TP2/4/8 shapes, which exist only here and which #4789 does not touch). By tuned time over the 16 shared shapes this is a 6.76% geomean improvement over #4789 alone and 5.81% over this branch alone.

This also closes the reviewer's non-blocking follow-up about the four `E=128` rows at tokens 512 / 8192 / 16384 / 32768. They were deliberately carried over unchanged here (deleting them is unsafe: `_PADDED_M_TIERS = [32768, 131072]` means the tier fallback in `_lookup_cfg` only applies for `token > 32768`, so 512 / 8192 / 16384 would miss the config entirely). #4789 retuned all four, and the merge adopts its faster rows -- token 512 goes 295.64 us -> 224.26 us, in line with the 212.74 us the reviewer measured locally.

**Scope of the serving numbers below:** they were measured on the pre-merge canonical config (`cab43a49...`). The merge only substitutes rows where the other tuner measured strictly faster on the same shape, so it is not expected to regress those results, but the serving sweep was not re-run after the merge.

## Serving performance

The benchmark compares the original PR #5155 runtime configuration against the consolidated canonical configuration. Both arms use the same PR binary and server settings; the candidate image is the baseline image with only the canonical tuned/untuned CSVs replaced and the duplicate short-name tuned CSV removed. `AITER_FLYDSL_FORCE_ATOMIC=0` and `AITER_FLYDSL_FORCE_REDUCE=0` in both arms, so runtime forcing did not select the kernels.

**Baseline identity clarification:** because of the duplicate-file merge behavior, the original effective TP1 rows came from `qwen3_vl_fp4_tuned_fmoe.csv` (`5af89c8c...`), while the original TP2/TP4/TP8 rows came from `q3vl_fp4_tuned_fmoe.csv` (`baea2c37...`). The candidate for every topology is the consolidated canonical file (`cab43a49...`). The canonical file contains the exact generated retune rows used by the earlier TP2/TP4/TP8 measurements.

Hardware/model: one 8× MI355X node, `amd/Qwen3-VL-235B-A22B-Instruct-MXFP4`, expert parallel enabled. Each arm ran the same 12 serving workloads at TP1, TP2, TP4, and TP8. Deltas below are canonical versus the original effective PR configuration; positive throughput and negative latency favor the canonical config. Intervals are descriptive 95% bootstrap intervals over the 12 workload rows; each topology has one independent paired sweep, so these intervals do not estimate run-to-run variation.

### Performance test script

The following is the complete per-server invocation and 12-case client loop. It was run separately for each `TP=1,2,4,8` and checksum-pinned arm. Run it inside the corresponding benchmark image, set `HIP_VISIBLE_DEVICES` to `TP` GPUs, and set `ARM` to `baseline` or `canonical`:

```bash
#!/usr/bin/env bash
set -euo pipefail

MODEL=amd/Qwen3-VL-235B-A22B-Instruct-MXFP4
TP=${TP:?set TP to 1, 2, 4, or 8}
ARM=${ARM:?set ARM to baseline or canonical}
PORT=${PORT:-9004}
: "${HIP_VISIBLE_DEVICES:?set HIP_VISIBLE_DEVICES to TP GPUs}"
export HIP_VISIBLE_DEVICES

CONFIG_DIR=/opt/aiter/aiter/configs/model_configs
CANONICAL_CONFIG=${CONFIG_DIR}/qwen3_vl_fp4_tuned_fmoe.csv
SHORT_CONFIG=${CONFIG_DIR}/q3vl_fp4_tuned_fmoe.csv
sha() { sha256sum "$1" | awk '{print $1}'; }

case "${ARM}" in
  baseline)
    [[ "$(sha "${CANONICAL_CONFIG}")" == 5af89c8ccfbe11a01e057afb4e30bba302ed2e87967ecd28efbe958c0da2f967 ]]
    [[ "$(sha "${SHORT_CONFIG}")" == baea2c37f381c531aa1710d99dffc72415f88bd8de5da5a1c7c4c0544217373f ]]
    ;;
  canonical)
    [[ "$(sha "${CANONICAL_CONFIG}")" == cab43a493451c6ca4057cf17c8f72371726e51d5ae258ffb23c397150441399d ]]
    [[ ! -e "${SHORT_CONFIG}" ]]
    ;;
  *) echo "ARM must be baseline or canonical" >&2; exit 2 ;;
esac

GPU_MEMORY_UTILIZATION=0.96
if [[ "${TP}" == "8" ]]; then
  GPU_MEMORY_UTILIZATION=0.94
fi

export SAFETENSORS_FAST_GPU=1
export VLLM_WORKER_MULTIPROC_METHOD=spawn
export HIP_FORCE_DEV_KERNARG=1
export AITER_ONLINE_TUNE=0
export AITER_USE_CK_MOE_SORTING=1
export VLLM_ROCM_USE_AITER=1
export VLLM_ROCM_USE_AITER_MOE=1
export VLLM_ROCM_USE_AITER_LINEAR=1
export VLLM_ROCM_USE_AITER_RMSNORM=1
export VLLM_ROCM_USE_AITER_FP4_ASM_GEMM=1
export VLLM_ROCM_QUICK_REDUCE_QUANTIZATION=INT4
export VLLM_ROCM_USE_AITER_MHA=0
export VLLM_Q3VL_FUSE_ATTN_PROLOGUE=1
export AITER_FLYDSL_FORCE=1
export AITER_FLYDSL_FORCE_ATOMIC=0
export AITER_FLYDSL_FORCE_REDUCE=0

# These were the only topology-specific environment overrides.
if [[ "${TP}" == "2" ]]; then
  export VLLM_ROCM_USE_AITER_CUSTOM_AR=0
fi
if [[ "${TP}" != "8" ]]; then
  export AITER_LOG_LEVEL=INFO
fi

vllm serve "${MODEL}" \
  --host 0.0.0.0 \
  --port "${PORT}" \
  --tensor-parallel-size "${TP}" \
  --gpu-memory-utilization "${GPU_MEMORY_UTILIZATION}" \
  --max-model-len 139264 \
  --max-num-batched-tokens 32768 \
  --max-num-seqs 2048 \
  --async-scheduling \
  --enable-expert-parallel \
  --no-enable-prefix-caching \
  --enable-chunked-prefill \
  --limit-mm-per-prompt '{"image":1}' \
  --compilation-config '{"mode":3,"cudagraph_mode":"FULL_AND_PIECEWISE","custom_ops":["+rms_norm","+quant_fp8","+rotary_embedding"],"splitting_ops":[],"pass_config":{"fuse_qk_norm_rope_kvcache":true,"rope_kvcache_fusion_max_token_num":32768}}' \
  --mm-encoder-tp-mode data \
  --mm-processor-cache-type shm \
  --trust-remote-code \
  --disable-log-stats \
  --kv-cache-dtype fp8_e4m3 \
  --quantization quark \
  >"server_${ARM}_tp${TP}.log" 2>&1 &
server_pid=$!
trap 'kill "${server_pid}" 2>/dev/null || true' EXIT

outdir="results/${ARM}/tp${TP}"
mkdir -p "${outdir}"

# ISL OSL MC seed warmup_requests num_prompts
while read -r isl osl mc seed warmup prompts; do
  python3 -m sglang.bench_serving \
    --backend sglang-oai-chat \
    --dataset-name image \
    --host localhost \
    --port "${PORT}" \
    --image-count 1 \
    --image-resolution 512x512 \
    --model "${MODEL}" \
    --ready-check-timeout-sec 6000 \
    --request-rate inf \
    --random-range-ratio 0.8 \
    --random-input-len "${isl}" \
    --random-output-len "${osl}" \
    --max-concurrency "${mc}" \
    --seed "${seed}" \
    --warmup-requests "${warmup}" \
    --num-prompts "${prompts}" \
    --output-file "${outdir}/isl${isl}_osl${osl}_mc${mc}.json"
done <<'CASES'
8192   1024 4   1  4   32
8192   1024 8   2  8   64
8192   1024 16  3  16  96
8192   1024 32  4  32  128
8192   1024 64  5  64  256
8192   1024 128 6  128 384
8192   1024 256 7  256 768
60000  600  4   8  4   32
60000  600  16  9  16  96
60000  600  64  10 64  256
128000 1024 4   11 4   32
128000 1024 16  12 16  96
CASES

kill "${server_pid}"
wait "${server_pid}" || true
trap - EXIT
```

### Performance comparison

- **TP1 (revalidated after consolidation):** total-token throughput **+1.84%** (95% row-bootstrap CI **+0.82% to +2.94%**); mean E2E latency **-1.76%** (CI **-2.83% to -0.75%**). Runtime kernel pairs differed on the expected 12/16 keys; the other four are the retained non-overlapping canonical tiers. The baseline emitted one primary duplicate-config warning per server (4 total), while the canonical arm emitted none.
- **TP2:** total-token throughput **+2.77%** (95% CI **+1.47% to +4.41%**); mean E2E latency **-2.71%**. Kernel sets differed on 14/16 common keys.
- **TP4:** total-token throughput **+2.74%** (95% CI **+2.06% to +3.63%**); mean E2E latency **-2.67%**. Kernel sets differed on 15/15 common keys.
- **TP8:** total-token throughput **+2.89%** (95% CI **+1.53% to +3.97%**); mean E2E latency **-2.81%**. Kernel sets differed on 15/16 common keys.

The final comparison set contains **96 successful serving cases**: 4 topologies × 2 configs × 12 workloads. Strict pre-shutdown log checks found no failed requests, fatal engine errors, OOMs, or runtime-forcing events. TP8 used `--gpu-memory-utilization 0.94` for both arms; an earlier 0.96 attempt was discarded after ROCm HSA resource exhaustion. This was a serving-only performance run; no GPQA or other accuracy benchmark was included.

The corrected TP1 result confirms the shadowing diagnosis: once the consolidated file actually dispatched the 12 replacement kernel pairs, TP1 improved as well. Total-token throughput is positive across all four tested topologies.

## Test plan

- [x] Run the atomic-constrained tuner across all 60 `q3vl` shapes.
- [x] Validate key preservation, row uniqueness, candidate filtering, and exact checksums.
- [x] Consolidate tuned and untuned manifests to 64 unique canonical keys and remove the duplicate tuned file.
- [x] Run the config-shape collision guard and verify production merging has no Qwen3-VL duplicate keys.
- [x] Revalidate TP1 against the original effective runtime config and confirm all 12 intended kernel-pair changes dispatch.
- [x] Compare the original effective config and canonical retune over 12 serving workloads at TP1/TP2/TP4/TP8.
- [x] Confirm no failed requests, fatal engine/memory errors, or runtime atomic/reduce forcing.
- [x] Merge current `main` and resolve the #4789 overlap per shape on measured tuner time.
- [x] Re-run `op_tests.tuning_tests.test_config_shape_collision` (17 passed, including the real runtime merge and both self-check controls) and `op_tests.tuning_tests.test_csv_validation` (15 passed) on the merged tree.


yzhou103:
LGTM. there is a Non-blocking follow-up (not a request to change this PR):

  The 4 E=128 rows at token 512 / 8192 / 16384 / 32768 are carried over unchanged from
  main, so this PR is a strict improvement and I don't think they should hold it up.
  Two notes for a follow-up though:

  - Deleting them is not a safe shortcut: `_PADDED_M_TIERS = [32768, 131072]` means the
    tier fallback in `_lookup_cfg` only kicks in for token > 32768, so 512/8192/16384
    would miss entirely and fall back to the untuned path. Keeping them is the right
    default here.
  - Retuning them does look worthwhile later. On a local MI355X (gfx950, cu_num=256),
    retuning token=512 / E=128 gave 212.74us vs the carried-over 295.64us (-28.0%), and
    the tuner picked exactly the same kernel pair as this PR's token=256 row
    (block_m=32, `..._t32x64x256_w3_kw2_fp4` + `..._moe2_layout_..._t32x128x256_atomic_nt_sbm32`),
    with err2 dropping 2.5% -> 1.1%. It also restores monotonicity across the E=128
    sweep (256=211.57, 512=212.74, 1024=235.12) — as-is there's a dip at 512.

vorapolsiloai:
@yzhou103  Thank you for the review! I just with updated the co-author to added my colleague contribution and remove AI agent from the commit. Also I merged to main, This trigger new CI test. Please feel free to merge them once the CI passed.

vorapolsiloai:
@yzhou103 a few of the CI test failed would it be possible to rerun those tests for me

yzhou103 (dismissed):

vorapolsiloai:
@yzhou103 I had to resolve merge conflict with #4789 which was merged today. 16 rows were conflicted. so I compared and pick the faster microbench time. Could you reapprove this PR again?

vorapolsiloai:
@yzhou103 I think the CI is broken, I restarted the 

> [Atom Test / Accuracy (Kimi-K2.7-Code-MXFP4, linux-aiter-do-mi350x-4)](https://github.com/ROCm/aiter/actions/runs/34829751291/job/104115875529?pr=5155)

test 3 times and it keep failing. I don't think this is related to my qwen3 vl tuned csv, the test is something to do with kimi k2.7 code mxfp4 accuracy test.

yzhou103 (approved):
LGTM

PR #5159, opened 2026-09-01
[Config] gptoss bf16 tuned gemm: drop the losing large-M QKV rows on gfx950 (tuned rows lose 16-27% to the hipBLASLt fallback)

alexnails:
## What

The gfx950 rows in `gptoss_bf16_tuned_gemm.csv` for the gpt-oss QKV projection shape (N=5120, K=2880, bias) at M=4096/8192/16384 pick kernels that lose to the plain hipBLASLt fallback on MI355X:

| M (lookup) | tuned pick | hipBLASLt (`torch.addmm`) | Δ |
|---|---|---|---|
| 61366 → 8192 row | flydsl, 1.744 ms | 1.271 ms | −27% |
| 4065 → 4096 row | flydsl | | −16.5% |
| 16384 row | "triton auto" | | −22% |

The o-projection shape — which has **no** tuned row — already runs at 93% of the measured 1424 TF/s GEMM ceiling through the same fallback, which is the tell. Delete the three rows so large-M lookups fall through to hipBLASLt; the small-M decode rows (M ≤ 64) keep flydsl, which wins there (+11.5% at M=64).

Verified on MI355X: the shapes take the hipBLASLt path with outputs identical to `torch.addmm` (max_abs 0.0), worth ~16 ms/request on a 61k-token gpt-oss prefill (36 launches). Suggests a broader guard for the tuning flow: reject a candidate row at table-generation time if it does not beat the untuned fallback for that shape.


yifehuan (approved):
LGTM

PR #5164, opened 2026-09-01
[HIP] [OPUS] [FlyDSL] Simplify AITER compiler worker fan-out to smooth-out regressions

Qubitium:
This PR centralizes AITER's CPU compilation fanout and makes the worker policy safe for long-lived inference processes and memory-limited build containers. It removes independently calculated worker pools and nested compiler fanout that could multiply into an OOM, while preserving AOT and runtime compile paths.

## What changed

- Added one shared worker policy based on process-available CPUs and the lower of host-available and finite cgroup memory, using the observed 1.5 GiB worker RSS estimate.
- Recompute automatic CPU and memory limits on every policy call; automatic results are never written into the environment.
- Keep `AITER_MAX_JOBS` as AITER's canonical ceiling and clamp it by live CPU and memory limits. Config-driven pools are also capped to submitted work.
- Force AITER, Ninja, CMake, Make, OpenMP, BLAS, and NumExpr descendants of AITER process-pool workers to one job, preventing nested fanout.
- Add cgroup v1/v2 awareness, fail-closed handling when finite cgroup usage cannot be read, and one-time `memory.stat` diagnostics showing page-cache components when conservative accounting limits the budget.
- Fix AOT worker argument/exception propagation and standalone entrypoint source-tree import behavior.
- Remove the dead runtime Ninja worker helper.

## Final `MAX_JOBS` compatibility policy

For an AITER-owned compile, ceiling precedence is:

1. Explicit `AITER_MAX_JOBS`.
2. A valid positive legacy `MAX_JOBS` when `AITER_MAX_JOBS` is unset.
3. Automatic CPU/memory/cgroup sizing.

Standalone AITER entrypoints may adopt a positive legacy value into `AITER_MAX_JOBS` with a migration warning. Runtime JIT reads the legacy value non-mutating at the AITER-owned Ninja compile boundary. Neither path modifies `MAX_JOBS`; malformed, zero, or negative legacy values fall back to automatic sizing. Plain `import aiter` and the generic `get_worker_count()` policy do not consult `MAX_JOBS`.

Both explicit and legacy ceilings remain upper bounds only: live CPU, host-memory, and finite cgroup-memory limits always clamp them. Nested worker controls remain forced to one inside AITER process-pool children.

## Validation

- Focused worker, entrypoint, blob-generator, and runtime-JIT regression suite: 41 passed, 26 subtests passed.
- Modified Python files pass byte-compilation and `git diff --check`.


valarLip (changes requested):
## Review

The problem this PR targets is real and worth solving. The findings below are all reproduced against `6c79b82` in a worktree; commands are included so you can re-run them.

Two are build-breaking.

### Blocking

**1. `aiter/jit/utils/cpp_extension.py:29` — the new top-level import breaks every codegen subprocess, and the failure is swallowed.**

The blob-gen scripts put `AITER_CORE_DIR` on `sys.path`, and in pip-install mode that resolves to `aiter/jit/utils`, *not* the repo root (`gemm_moe_ck2stages_common.py:9-16`). So:

```
$ python3 csrc/ck_gemm_moe_2stages_codegen/gen_instances.py --working_path /tmp/x
  File ".../aiter/jit/utils/chip_info.py", line 15, in <module>
    from cpp_extension import executable_path
  File ".../aiter/jit/utils/cpp_extension.py", line 29, in <module>
    from aiter_worker_limits import get_worker_count, get_worker_count_per_parent
ModuleNotFoundError: No module named 'aiter_worker_limits'
```

The same command on `origin/main` gets past all imports. These scripts are launched from `aiter/jit/core.py:997` as `os.system(f"{PY} {blob_gen_cmd.format(blob_dir)}")` — the return code is discarded — so on a clean from-source `PREBUILD_KERNELS=1 pip install -e .`, before `aiter_worker_limits` reaches site-packages via the new `py_modules` entry, the affected modules generate zero kernel instances and build silently empty. `optCompilerConfig.json` has 126 entries carrying a `blob_gen_cmd`; every one whose generator routes through `chip_info` is exposed. You hit this exact problem for `build_co.py` and added `sys.path.insert(0, _REPO_ROOT)` there — the `cpp_extension` path needs the same sweep (or the module needs to live somewhere already importable, see the note at the end).

**2. `aiter/aot/pa.py:101` — consuming the `executor.map` iterator turns all five AOT drivers into hard failures.**

`process_config` ends at `csrc/cpp_itfs/utils.py:320 return run_lib(func_name, folder)` → `:277 return getattr(lib, func_name)`, i.e. a ctypes `_FuncPtr`, which cannot be pickled back to the parent:

```
>>> list(ProcessPoolExecutor(1).map(lambda _: ctypes.CDLL(None).getpid, [0]))
ValueError: ctypes objects containing pointers cannot be pickled
```

Base never consumed the iterator, so the error was discarded and the prebuild "succeeded". After this PR `python3 -m aiter.aot.pa` aborts — and that is the exact command the PR adds at `README.md:151`. Same `list(executor.map(...))` at `pa_ragged.py:101`, `pa_v1.py:108`, `asm_mla_decode_fwd.py:50`, and `sampling.py:98-100`. The new tests mock `ProcessPoolExecutor`, so they can't catch it. Simplest fix: have `process_config` return `None` (or the `func_name` string).

### The policy cancels its own guarantees

**3. `aiter_worker_limits.py:209` — the auto-computed value is written into `AITER_MAX_JOBS`, which lines 197-206 then treat as an explicit expert override that bypasses both caps.**

Live:

```python
first  = get_worker_count()          # -> 204, and os.environ["AITER_MAX_JOBS"] == "204"
w._cpu_worker_cap = lambda: 2        # crush the CPU budget
second = get_worker_count()          # -> 204   (expected 2)
```

Base did the opposite: `aiter/jit/core.py:621-643` recomputed `cpu_count` and psutil availability on *every* call and clamped down. Concretely, `setup.py:412` calls `get_worker_count()` on an idle host → 204 is exported; the FlyDSL AOT phase then consumes tens of GB; every later JIT/ninja build in that process, and in every spawned child, reads 204 as "explicit" and skips the memory check entirely. Same in a long-lived vLLM/SGLang process, and a child under a tighter cpuset/cgroup skips its own caps. This structurally cancels the memory cap the PR exists to add. Suggest keeping the computed value in a module-level cache rather than in the environment, and exporting a *different* variable for children if propagation is the goal.

**4. `:108` — `_cgroup_memory_directories()` returns on the first matching mountinfo entry, so a cgroup v1 memory limit is silently dropped on hybrid hosts.**

The loop `return directories` as soon as one mount's ancestry walk reaches `mount_point`. On a hybrid host where the cgroup2 unified mount is listed before the v1 `memory` mount, the v2 entry wins, `_cgroup_memory_remaining_bytes()` returns `None`, and `_available_memory_bytes()` falls back to host psutil availability. In `docker run -m 32g` on a large host that yields a memory budget in the thousands, so the CPU cap wins and ~100 compile workers run inside a 32 GB cgroup → container OOM-kill. This is precisely the container case the PR description says it fixes. Collect across *all* matching mounts and take the tightest bound.

**5. `:145` — the cgroup budget counts reclaimable page cache as used.**

Remaining is `limit - memory.current` (v2) / `limit - memory.usage_in_bytes` (v1). Both counters include page cache and reclaimable slab, and `memory.stat` is never opened (grep confirms zero references). On this host `memory.current` is 1.354 TB, of which inactive_file + active_file + slab_reclaimable = 92.8%; only ~91 GB is anon. In a CI container with a finite `memory.max` that has just cloned the repo and hipified sources, `limit - current` is a few GB → `max(1, avail // 1_500_000_000)` gives 1-2 workers, and `PREBUILD_THREAD_NUM`, ninja `-j`, `get_max_workers` and every AOT pool all serialize silently, getting worse the longer the container runs. The correct bound is `limit - (usage - reclaimable_file_and_slab)`.

**6. `:40` — the psutil failure path floors the budget at exactly one worker, and the `except` tuple is too narrow.**

`_host_available_memory_bytes()` returns `EST_WORKER_RSS_BYTES` on failure, and `1_500_000_000 // 1_500_000_000 == 1`, so `get_worker_count()` returns 1 and exports `AITER_MAX_JOBS=1`: a 128-core build silently goes from minutes to hours with no warning. Base failed open in both places — `setup.py:41-43 except ImportError: max_num_jobs_memory = max_num_jobs_cores` (comment: "psutil may not be available during metadata extraction") and `flydsl/common.py:233 except Exception: return default_workers`. Also, `psutil.Error` derives from `Exception`, not `OSError`:

```
psutil.AccessDenied.__mro__ == (AccessDenied, Error, Exception, BaseException, object)
```

so `AccessDenied`/`ZombieProcess`, and the `NotImplementedError` psutil raises on an unsupported platform, are all outside `(ImportError, OSError, ValueError, KeyError)` and propagate out of `get_worker_count()` into `import aiter` / `pip install` as a hard traceback. Fail open to the CPU budget and catch `Exception`.

**7. `:197` — an explicit `AITER_MAX_JOBS` bypasses every safety cap, and this PR sets it to 64 in five workflows.**

Base clamped *down*: `if max_processes > max_jobs: os.environ["MAX_JOBS"] = str(max_jobs)`, and `_get_num_workers` clamped against `max(1, cpu_count*0.8)`. After the rename, all five of `aiter-test.yaml:24`, `atom-test.yaml:260`, `vllm-disagg-ci-smoke-workflow.yaml:45`, `kimi-perf-downstream.yaml:96`, `kimi-downstream.yaml:112` set `AITER_MAX_JOBS=64` unconditionally — what used to be a ceiling hint is now a cap bypass. On a smaller runner, or in a 32 GB memory cgroup, 64 × 1.5 GB estimated RSS = 96 GB → an OOM-killed build where the old code would have clamped. For a PR titled around bounding fanout, this removes the one bound that existed. Consider clamping an explicit value to the auto budget with a warning, rather than documenting it as an unsafe override.

**8. `:202` — a malformed or empty `AITER_MAX_JOBS` now raises on the runtime JIT path.**

```
$ AITER_MAX_JOBS= python3 -c "import aiter_worker_limits as w; w.get_worker_count()"
ValueError: AITER_MAX_JOBS must be an integer, got ''
$ AITER_MAX_JOBS=auto ...
ValueError: AITER_MAX_JOBS must be an integer, got 'auto'
```

This is reached from `aiter/jit/core.py:989 check_and_set_ninja_worker()` inside `build_module`, so a stray value — a `-e AITER_MAX_JOBS="${VAR}"` passthrough where `VAR` is unset, or a shell `AITER_MAX_JOBS= pip install -e .` — turns the first JIT compile inside a running inference server into an uncaught traceback rather than a slightly-wrong worker count. Base explicitly tolerated it (`core.py:639-640 except ValueError:`, and the old `_get_num_workers` gated on `.isdigit()`). The same hard-fail was newly added for `PREBUILD_THREAD_NUM`. Note `int()` still accepts `' 8 '`, `'+8'` and `'8_0'` (== 80), so the validation is loose in one direction and fatal in the other.

### Fanout regressions

**9. `setup.py:370` — `PREBUILD_THREAD_NUM` loses its hard cap of 5, inverting the build shape.**

Base:
```python
prebuid_thread_num = 5
max_jobs = os.environ.get("MAX_JOBS")
if max_jobs is not None and max_jobs.isdigit() and int(max_jobs) > 0:
    prebuid_thread_num = min(prebuid_thread_num, int(max_jobs))
else:
    prebuid_thread_num = min(prebuid_thread_num, getMaxJobs())
```
Head: `prebuild_thread_num = get_worker_count_for(len(all_opts_args_build))` — one slot per prebuilt module, which is on the order of a hundred. The shape goes from "5 modules × wide ninja" to "N modules × `ninja -j1`", because the nested budget is `get_worker_count_per_parent(N)` and floor-divides to 1. Under CI's `AITER_MAX_JOBS=64` that's 64 × `-j1` vs base 5 × `-j12`; on a 256-core host it's ~128 concurrent jobs against a 204 budget, leaving a third of the budget unused. Work-stealing exists only at module granularity and the nested `-j` is frozen for the whole build, so the tail is serial — the largest CK module compiles all its TUs at `-j1` while everything else sits finished.

Worse, `os.environ["PREBUILD_THREAD_NUM"]` is never popped in setup.py's own process (verified), and `run_pretune_modules` (`aiter/utility/pretune.py:362`) is a plain sequential loop, so each pretune rebuild still divides by that stale count and gets `-j1` on an otherwise idle machine.

**10. `aiter/utility/mp_tuner.py:27` — a compiler-fanout policy is installed as the GPU tuning pool's initializer, pinning JIT builds to `-j1`.**

`_init_task_start_times` is the `initializer=` for both `mp.Pool` sites (`:473-475`, `:651-653`), and it calls `configure_worker_subprocesses()`, which does `os.environ[_WORKER_ENV] = "1"`. But `setup.py`'s `get_exclude_ops()` excludes every `_tune` module from the prebuild, so the `.so` doesn't exist and the first worker to call e.g. `aiter.gemm_a8w8_tune` (`@compile_ops("module_gemm_a8w8_tune")`) triggers `build_module` in-process; `base_tuner._set_config_env_for_run_config` additionally sets `AITER_REBUILD`, forcing rebuilds mid-tune. Those builds now run `-j1` instead of ~0.8×ncpu — on a 256-core / 8-GPU box that's 8 total compile jobs instead of ~200. The right primitive already exists in that module: `get_worker_count_per_parent(mp_num)`.

Two notes on the same function: the `OMP/MKL/OPENBLAS=1` half is largely inert under spawn (torch and libgomp are imported before the initializer runs), which makes the dict misleading rather than harmful; and `NINJAFLAGS` is not an environment variable ninja reads — the string is absent from the ninja binary, which only knows `NINJA_STATUS` — so that entry is dead and the `-j1` effect comes solely from `AITER_MAX_JOBS`. Also `csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py` builds its pool without any initializer, so the two tuners now behave differently.

**11. `csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot_prebuild.py:808` — a GPU-resident process pool is now sized by a CPU/host-RAM compiler policy.**

Base always passed `args.num_processes` (argparse default 32), so the effective value was `min(cpu_count, 32, total)` = 32 and the `if num_processes is None` branch was dead. Head is `get_worker_count_for(total)` → up to ~204 on a 256-core host. These children genuinely use the GPU: the module does `torch.set_default_device("cuda")` at line 30 and `run_pa_gluon_test` sets `device = "cuda:0"` at :167-168 and allocates KV-cache/block-table tensors there. The pool is spawn-based, so that's ~204 separate HIP contexts on `cuda:0` (hundreds of MB each) plus per-test KV caches. GPU memory appears nowhere in `aiter_worker_limits.py`, and with `--num_processes` deleted the only remaining knob is the global `AITER_MAX_JOBS`, which also throttles unrelated compilation.

**12. `aiter/aot/flydsl/common.py:210` — FlyDSL AOT loses its 64-worker ceiling and its per-worker memory assumption drops from 2.0 GB to 1.5 GB.**

Base had `_DEFAULT_MAX_WORKERS = 64` and `_DEFAULT_MEM_PER_WORKER_GB = 2.0`, applied as `min(_affinity_aware_cpu_count(), 64)` then `_memory_worker_cap(...)`. Head is `get_worker_count_for(num_jobs)` with `EST_WORKER_RSS_BYTES = 1_500_000_000` and no absolute ceiling: on a 256-core host with 300 GB available that's 64 → ~200 workers. The base README row said "Each worker uses ~1.5-2.5 GB RSS"; at the 2.5 GB end 200 workers is 500 GB against 300 GB available. The `AITER_FLYDSL_AOT_WORKERS` / `AITER_FLYDSL_AOT_MEM_PER_WORKER_GB` escape hatches were deleted along with the ceiling. Killed workers cost a retry each at a 1200 s timeout, and `_run_file_pool` drops permanently-failed kernels as `results[idx] = None`, so a memory-pressured run can silently ship a build missing kernels.

**13. `csrc/opus_gemm/gen_co/build_co.py:481` — the deliberately conservative hipcc fanout is replaced by the global budget, and `--jobs/-j` is deleted.**

The removed argparse help stated the rationale directly: each hipcc invocation takes ~1 core for ~3 s, a full sweep is thousands of entries, so the default (`cpu_count()//8`) was deliberately well under nproc. On a 256-core host that goes 32 → up to ~204 concurrent hipcc processes in a `ThreadPoolExecutor` whose workers each hold a subprocess. hipcc's peak RSS for opus template instantiation isn't the 1.5 GB AOT-Python-worker figure the budget is derived from, so the memory model under-counts exactly the workload most likely to OOM. `build_co.py -j 8` is now an argparse "unrecognized arguments" error, and `configure_worker_subprocesses` isn't called here (threads, same process), so nothing bounds nested fanout either.

### Compatibility and CI

**14. Generic `MAX_JOBS` is now silently ignored, with no deprecation path.**

The in-repo rename is complete and the surviving `MAX_JOBS` uses (e.g. `.github/scripts/build_aiter_triton.sh:66`) correctly build Triton — but external consumers can't be updated in lockstep. Verified in this workspace: `/app/ATOM/docker/atom_release.dockerfile:78` builds aiter with `MAX_JOBS=$MAX_JOBS PREBUILD_KERNELS=$PREBUILD_KERNELS ...`, and `/app/ATOM/docker/Dockerfile` has the same `ARG MAX_JOBS` / `ENV MAX_JOBS` shape. After this PR those throttles silently evaporate on a shared or memory-limited builder and aiter takes 80% of all CPUs instead — the exact direction the PR intends to prevent, with no error to signal it. A one-line back-compat read ("if `AITER_MAX_JOBS` is unset and `MAX_JOBS` is set, use it as an upper bound and warn once") would preserve every existing deployment at no cost.

**15. The black pre-check goes red, and the new `tests/` tree never runs in CI.**

`.github/workflows/pre-checks.yaml:35` runs `psf/black@stable`; `python3 -m black --check` on the PR-touched files reports exactly:

```
would reformat aiter_worker_limits.py
would reformat setup.py
would reformat op_tests/opus/device/setup.py
would reformat tests/test_asm_mla_decode_aot.py
would reformat tests/test_worker_awareness.py
5 files would be reformatted, 2 files would be left unchanged.
```

Separately, `tests/` has no `__init__.py`, no `conftest.py`, and no pytest `pythonpath`, so the entry points behave inconsistently:

| command | result |
|---|---|
| `python3 -m pytest tests/test_worker_awareness.py` | 23 passed |
| `pytest tests/test_worker_awareness.py` | collection ERROR |
| `python3 tests/test_worker_awareness.py` | `ModuleNotFoundError: No module named 'aiter_worker_limits'` |

The last one is the `if __name__ == "__main__"` entry point the file itself advertises. And no workflow runs this directory: `aiter-test.yaml:22` pins `AITER_TEST: "op_tests"`, and the only `tests/` paths in the workflows are `flash_attention_integration.yaml`'s `tests/test_flash_attn_triton_amd.py`, which does not exist in this repo (it belongs to a different checkout). So none of the new regression coverage that would catch findings #1-#8 actually executes. `CONTRIBUTE.md:114/117/152` also still say tests live in `op_tests/`.

### One structural suggestion

`aiter_worker_limits.py` sits at the repo **root** and ships as an unnamespaced top-level `py_modules` entry, but the repo already has `aiter/jit/utils/` for exactly the "importable before the package exists" role — that's where `cpp_extension.py` and `chip_info.py` live, and it's already on `sys.path` in every context that broke in finding #1. Moving it there removes the two `sys.path.insert(0, _REPO_ROOT)` bandaids and the need for `tests/test_worker_entrypoints.py`.

Worth knowing while you decide: `setup.py`'s module-scope `from aiter_worker_limits import ...` works only because `pyproject.toml` declares `[build-system] requires` with no `build-backend` key. PEP 517 then defaults to `setuptools.build_meta:__legacy__`, the one backend that injects the source directory into `sys.path`. Adding the conventional explicit `build-backend = "setuptools.build_meta"` line — something an unrelated packaging cleanup would plausibly do — would break `pip install .` on this import.

Finally, this PR bundles a real MLA ABI fix and the `executor.map` changes with a large policy refactor and CI renames across 28 commits. If the policy needs to be reverted after landing, the unrelated fixes go with it — worth splitting.

valarLip:
Following up on my review with a scoping suggestion, since I think the goal here is reachable with far less surface area than 30 files.

### What `main` already does

Lowering the fanout is already supported and already clamps downward:

```python
# aiter/jit/core.py:621  check_and_set_ninja_worker()
if max_processes > max_jobs:            # user asked for too much -> clamp
    os.environ["MAX_JOBS"] = str(max_jobs)
```

So `MAX_JOBS=32 pip install -e .` works today, and downstream images already rely on it (e.g. `ATOM/docker/atom_release.dockerfile:78`).

### What's actually broken in `main`

The memory term is effectively dead:

```python
max_num_jobs_cores  = os.cpu_count() * 0.8                 # 256-core host -> 204
free_memory_gb      = psutil.virtual_memory().available    # 2889 GB on my box
max_num_jobs_memory = int(free_memory_gb / 0.5)            # -> 5779
max_jobs            = min(204, 5779)                       # -> 204; memory never binds
```

Two concrete defects:

1. **0.5 GB per job is far too low.** This PR itself uses `EST_WORKER_RSS_BYTES = 1_500_000_000`, and `aiter/aot/flydsl/common.py` used 2.0 GB with a README range of 1.5-2.5 GB. At 0.5 GB the memory term can only bind below ~102 GB available, so on any large builder the CPU term always wins.
2. **`psutil.virtual_memory().available` reports the host.** Inside `docker run -m 32g` the cgroup limit is invisible, so the computed value is wrong in exactly the environment that OOMs. This is the part an operator cannot fix by picking a number — they'd have to know the right number per container.

The same computation is also duplicated verbatim in `setup.py:31 getMaxJobs()` and `core.py:621 check_and_set_ninja_worker()`.

### Suggested smaller scope

Both defects are fixable in roughly 15 lines, in one shared helper replacing those two copies:

- raise the per-job estimate from 0.5 GB to ~1.5 GB
- derive available memory as `min(psutil.available, cgroup_limit - anon_usage)`, where `anon_usage` comes from `memory.stat` (i.e. `memory.current` minus `inactive_file + active_file + slab_reclaimable`); fall back to psutil when no finite cgroup limit is readable
- keep the existing clamp-down semantics for `MAX_JOBS`

That gets container awareness and a realistic memory model, leaves `MAX_JOBS` behaving as it does today, and requires no changes in downstream images.

### Why I'd suggest that over the current shape

From the review, the current implementation doesn't yet deliver the guarantee it's built for:

- an explicit `AITER_MAX_JOBS` now **bypasses** both caps rather than clamping, and this PR's own CI sets it to 64 in five workflows
- the auto-computed value is written back into `AITER_MAX_JOBS`, after which it is read as an explicit override, so the memory check is skipped for the rest of the process and in every child
- the cgroup path counts reclaimable page cache as used, which collapses the budget toward 1 worker in any container that has done I/O
- the outer-pool x inner-ninja product got worse, not better (`128 x -j1` where `main` was `5 x -j12`)

None of those are hard to fix individually, but the smaller version avoids most of them by construction. Happy to be wrong if there's a requirement I'm missing — is there a case where the per-site policies (FlyDSL's 64-worker ceiling, `build_co.py`'s `cpu_count()//8`, `pa_gluon`'s 32 GPU-resident processes) genuinely need to be unified under one number rather than just corrected in place?

Separately and independently of the scope question: the two build-breaking items in the review (the `cpp_extension` import that breaks blob-gen subprocesses, and the `list(executor.map(...))` change that makes all five AOT drivers raise on an unpicklable ctypes pointer) are worth fixing regardless of which direction this goes.

Qubitium:
@valarLip  Thank you for the review! You brought up several bugs in the implementation. I will address them one by one as I read through them. 

> ### The policy cancels its own guarantees
> **3. `aiter_worker_limits.py:209` — the auto-computed value is written into `AITER_MAX_JOBS`, which lines 197-206 then treat as an explicit expert override that bypasses both caps.**
> 
> Live:
> 
> ```python
> first  = get_worker_count()          # -> 204, and os.environ["AITER_MAX_JOBS"] == "204"
> w._cpu_worker_cap = lambda: 2        # crush the CPU budget
> second = get_worker_count()          # -> 204   (expected 2)
> ```
> 
> Base did the opposite: `aiter/jit/core.py:621-643` recomputed `cpu_count` and psutil availability on _every_ call and clamped down. Concretely, `setup.py:412` calls `get_worker_count()` on an idle host → 204 is exported; the FlyDSL AOT phase then consumes tens of GB; every later JIT/ninja build in that process, and in every spawned child, reads 204 as "explicit" and skips the memory check entirely. Same in a long-lived vLLM/SGLang process, and a child under a tighter cpuset/cgroup skips its own caps. This structurally cancels the memory cap the PR exists to add. Suggest keeping the computed value in a module-level cache rather than in the environment, and exporting a _different_ variable for children if propagation is the goal.

This is a state problem that is out of scope of this PR. We don't have (or need):

1. Who is spawning (parent) who
2. By how many children is the fan-out
3. The width/depth is unknown. 
4. There is no process global tracker of fan-ed out workers alive/existed counter. 

So this pr simplifies policy to:

* Establish one bounded top-level aiter fanout budget.
* Propagate that budget instead of independently recalculating it throughout the process tree.
* For aiter controlled process-pool children, explicitly force nested compiler fanout to one.

The primary goal here is also not maximizing compiler throughput. It is preventing relatively short-lived aot jit compilation from creating unbounded fanout and ooming which is a much worser outcome and highly likely in base than slower but mostlly one-time compile before inference begines. 

PR documents that overriding AITER_MAX_JOBS externally is unsafe.

Qubitium:
> 5. :145 — the cgroup budget counts reclaimable page cache as used.
> 
> Remaining is limit - memory.current (v2) / limit - memory.usage_in_bytes (v1). Both counters include page cache and > > reclaimable slab, and memory.stat is never opened (grep confirms zero references). On this host memory.current is > 1.354 TB, of which inactive_file + active_file + slab_reclaimable = 92.8%; only ~91 GB is anon. In a CI container with a > finite memory.max that has just cloned the repo and hipified sources, limit - current is a few GB → max(1, avail // > 1_500_000_000) gives 1-2 workers, and PREBUILD_THREAD_NUM, ninja -j, get_max_workers and every AOT pool all > serialize silently, getting worse the longer the container runs. The correct bound is limit - (usage -  > reclaimable_file_and_slab).

Page cache/buffer are often not released by Linux kernel on time. We should be conservative here. On many systems, page/buffer can be over 40% of the os memory but more often than not, they are released when it is too late (swap triggered stalls).

Qubitium:
>7. :197 — an explicit AITER_MAX_JOBS bypasses every safety cap, and this PR sets it to 64 in five workflows.
>
>Base clamped down: if max_processes > max_jobs: os.environ["MAX_JOBS"] = str(max_jobs), and _get_num_workers >clamped against max(1, cpu_count*0.8). After the rename, all five of aiter-test.yaml:24, atom-test.yaml:260, vllm->disagg-ci-smoke-workflow.yaml:45, kimi-perf-downstream.yaml:96, kimi-downstream.yaml:112 set >AITER_MAX_JOBS=64 unconditionally — what used to be a ceiling hint is now a cap bypass. On a smaller runner, or in a >32 GB memory cgroup, 64 × 1.5 GB estimated RSS = 96 GB → an OOM-killed build where the old code would have >clamped. For a PR titled around bounding fanout, this removes the one bound that existed. Consider clamping an >explicit value to the auto budget with a warning, rather than documenting it as an unsafe override.

I set the CI to 64 because that's was `base` value. Ci can override this value because I believe your CI is >= 128 cores so this is safe. Only override `AIDER_MAX_JOBS` when the env is known and pre-deiminated.  PR has added doc warnigs about unsafe usage of AITER_MAX_JOBS

I will remove the CI static set 64 value that mirrors base.

Qubitium:
> #14 — “Generic MAX_JOBS is now silently ignored, with no deprecation path.”

This was by design. Compat code bloat should be avoided since this PR documents that this is a feature, not compat bug. AITER should not read or override parent (sglang/vllm) MAX_JOBS full-stop since since they may use it for their own aot/jit compilers outside of AITER scope and how MAX_JOBS is genearted is unknown. Aiter should avoid outside noise and do it's own thing.

valarLip:
> MAX_JOBS

it's not only aiter but also sgl/vllm, all are using “MAX_JOBS”, instead of SGLANG_MAX_JOBS  or VLLM_MAX_JOBS, why aiter need AITER_MAX_JOBS ?

Qubitium:
> #10 — “aiter/utility/mp_tuner.py:27 — a compiler-fanout policy is installed as the GPU tuning pool's initializer, pinning JIT builds to -j1.”

Also by design so trying to turn this potential fan-out:

8 GPU tuner processes
× 25 compiler workers each
= 200 compiler workers

into this

8 GPU tuner processes
× 1 compiler workers each
= 8 compiler workers

valarLip:
> > #10 — “aiter/utility/mp_tuner.py:27 — a compiler-fanout policy is installed as the GPU tuning pool's initializer, pinning JIT builds to -j1.”
> 
> Also by design so trying to turn this potential fan-out:
> 
> 8 GPU tuner processes × 25 compiler workers each = 200 compiler workers
> 
> into this
> 
> 8 GPU tuner processes × 1 compiler workers each = 8 compiler workers

this is aiter's design to fanout to use more cpu core for speed up compile

Qubitium:
> > MAX_JOBS
> 
> it's not only aiter but also sgl/vllm, all are using “MAX_JOBS”, instead of SGLANG_MAX_JOBS or VLLM_MAX_JOBS, why aiter need AITER_MAX_JOBS ?

Agree that not using something global/universal like MAX_JOBS seems strange but aiter is a aot/jit compiler that sits inside a user-space program like sgalng/vllm that is thread-light, not thread-heavy like a compiler. Aiter is especially sensitive due to it's fan-out/threads so it needs dedicated control.

Qubitium:
> > > MAX_JOBS
> > 
> > 
> > it's not only aiter but also sgl/vllm, all are using “MAX_JOBS”, instead of SGLANG_MAX_JOBS or VLLM_MAX_JOBS, why aiter need AITER_MAX_JOBS ?
> 
> Agree that not using something global/universal like MAX_JOBS seems strange but aiter is a aot/jit compiler that sits inside a user-space program like sgalng/vllm that is thread-light, not thread-heavy like a compiler. Aiter is especially sensitive due to it's fan-out/threads so it needs dedicated control. 


> > > #10 — “aiter/utility/mp_tuner.py:27 — a compiler-fanout policy is installed as the GPU tuning pool's initializer, pinning JIT builds to -j1.”
> > 
> > 
> > Also by design so trying to turn this potential fan-out:
> > 8 GPU tuner processes × 25 compiler workers each = 200 compiler workers
> > into this
> > 8 GPU tuner processes × 1 compiler workers each = 8 compiler workers
> 
> this is aiter's design to fanout to use more cpu core for speed up compile

I fully understand this. But at what cost. 1TB model crashed because Aiter wanted one time faster 2s compile?

Qubitium:
> > > > MAX_JOBS
> > > 
> > > 
> > > it's not only aiter but also sgl/vllm, all are using “MAX_JOBS”, instead of SGLANG_MAX_JOBS or VLLM_MAX_JOBS, why aiter need AITER_MAX_JOBS ?
> > 
> > 
> > Agree that not using something global/universal like MAX_JOBS seems strange but aiter is a aot/jit compiler that sits inside a user-space program like sgalng/vllm that is thread-light, not thread-heavy like a compiler. Aiter is especially sensitive due to it's fan-out/threads so it needs dedicated control.
> 
> > > > #10 — “aiter/utility/mp_tuner.py:27 — a compiler-fanout policy is installed as the GPU tuning pool's initializer, pinning JIT builds to -j1.”
> > > 
> > > 
> > > Also by design so trying to turn this potential fan-out:
> > > 8 GPU tuner processes × 25 compiler workers each = 200 compiler workers
> > > into this
> > > 8 GPU tuner processes × 1 compiler workers each = 8 compiler workers
> > 
> > 
> > this is aiter's design to fanout to use more cpu core for speed up compile
> 
> I fully understand this. But at what cost. 1TB model crashed because Aiter wanted one time faster 2s compile?

@valarLip Again, I understand the goal and would like to actually address the compilation performance goal end goal of Aiter but it not in the scope of this PR. First, we fix the out of bound thread growth in hostile envs. Then we design a progress global worker take/put token system (or something like it) so we can track show many threads/processes have spawned and in compile due to fan-out. 

Btw, I have another PR wip that will address some of the AIIT compilation speed issues.

Qubitium:
> 8. :202 — a malformed or empty AITER_MAX_JOBS now raises on the runtime JIT path.

Skip

Qubitium:
>#13 — “csrc/opus_gemm/gen_co/build_co.py:481 — the deliberately conservative hipcc fanout is replaced by the global budget, and --jobs/-j is deleted.” 

I think we need some telemtry on this. I am not sure if I hit OPUS so PR code may actually cause oom since it may give opus more threads and a looser 1.5GB guard than base's tigher 2GB assumption. I think this part is where some telemtry can happen to see what happens in the real world.

Qubitium:
> > #13 — “csrc/opus_gemm/gen_co/build_co.py:481 — the deliberately conservative hipcc fanout is replaced by the global budget, and --jobs/-j is deleted.”
> 
> I think we need some telemtry on this. I am not sure if I hit OPUS so PR code may actually cause oom since it may give opus more threads and a looser 1.5GB guard than base's tigher 2GB assumption. I think this part is where some telemtry can happen to see what happens in the real world.

First telemetry is back. This PR is faster than base by 5x for test OPUS run on a 24 core VM with`gfx1250 offline OPUS path`

Second telemtry back:

image


Result, this PR does not enhance or regress OPUS compilation speed. 

> main’s old default launched 3 hipcc/clang chains, peaked at ~0.64 GiB aggregate sampled RSS, and took 86.6 s; the PR launched 19 chains, peaked at ~3.63 GiB, and took 14.8 s.
  No OOM occurred. I’m doing one final PR run at 10 ms sampling to tighten the peak estimate rather than relying on the earlier 100 ms sample.

Qubitium:
> #6 — “:40 — the psutil failure path floors the budget at exactly one worker, and the except tuple is too narrow.”

Pushed more generic Exception catch. 

If a syscall failed, it is more likely than not, it is catastraphic  so we should stick to `1` thread for safety.

Qubitium:
> ### Suggested smaller scope
> Both defects are fixable in roughly 15 lines, in one shared helper replacing those two copies:
> 
> * raise the per-job estimate from 0.5 GB to ~1.5 GB
> * derive available memory as `min(psutil.available, cgroup_limit - anon_usage)`, where `anon_usage` comes from `memory.stat` (i.e. `memory.current` minus `inactive_file + active_file + slab_reclaimable`); fall back to psutil when no finite cgroup limit is readable
> * keep the existing clamp-down semantics for `MAX_JOBS`

>raise the per-job estimate from 0.5 GB to ~1.5 GB 

This was already done by PR

> * derive available memory as `min(psutil.available, cgroup_limit - anon_usage)`, where `anon_usage` comes from `memory.stat` (i.e. `memory.current` minus `inactive_file + active_file + slab_reclaimable`); fall back to psutil when no finite cgroup limit is readable

As stated in later comments, you cannot rely on Linux kernel to free those free able memory, in time. It's too late. You must count them as unrecoverable (as in unrecoveratable in time to cause swap stall).  Once the oom triggers, kernel is in catch 22. It must honor alloc request or start killing user-space process which is just as bad as killing vllm/sglang in this production env (think ssh, shell, tmux).

> keep the existing clamp-down semantics for MAX_JOBS

Also addressed. This is by design. This PR is designed to fix by ignoring incorrectly ctxed MAX_JOBS clamp, not use it.  Parent (sglang/vllm) uses it in it's own ctx so Aiter should not.

Qubitium:
> Fanout regressions
> 9. setup.py:370 — PREBUILD_THREAD_NUM loses its hard cap of 5, inverting the build shape.

No regression found in telemetry. Actually I just removed unnecessary PREBUILD_THREAD_NUM micro-optimization. Tests show PR is just as good as base if not better.  https://github.com/ROCm/aiter/pull/5164/commits/f1f3099132cdc133244f18a11f6ec4d26f8c4dbd

image

Qubitium:
> 2. aiter/aot/pa.py:101 — consuming the executor.map iterator turns all five AOT drivers into hard failures.

> process_config ends at csrc/cpp_itfs/utils.py:320 return run_lib(func_name, folder) → :277 return getattr(lib, > func_name), i.e. a ctypes _FuncPtr, which cannot be pickled back to the parent:

Fixed. This bug was already in but hidden in `base` but revealed by PR by consuming the result of the workers.  https://github.com/ROCm/aiter/pull/5164/commits/b61af1cd96b8029cb62b3e6fd08475aa19b1d078

Qubitium:
> 1. aiter/jit/utils/cpp_extension.py:29 — the new top-level import breaks every codegen subprocess, and the failure is swallowed.

> The blob-gen scripts put AITER_CORE_DIR on sys.path, and in pip-install mode that resolves to aiter/jit/utils, not the repo root (gemm_moe_ck2stages_common.py:9-16). So:

Fixed in https://github.com/ROCm/aiter/pull/5164/commits/87d9919e8e7121ad517d7a2a5145b733873965d6

Qubitium:
>11. csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot_prebuild.py:808 — a GPU-resident process pool is now sized by a CPU/host-RAM compiler policy.

Just removed and replaced `num_processes ` with AITER_MAX_JOBS.  This code is only used by ci and cli so the env var is perfect override. No need to for another var to represent the same meaning. Even though AITER_MAX_JOBS is cpu/ram derived, it can be overriden in this case to be gpu correct. 

https://github.com/ROCm/aiter/pull/5164/commits/95d3dca8e10abcb011364b081247fb90af69bd7b

Qubitium:
Fixes stale and sticky AITER_MAX_JOBS should be recomputed 
https://github.com/ROCm/aiter/pull/5164/commits/66ea3104a177da80dfd33d64e10beac27366f3f0

Qubitium:
Added backward compatibility for MAX_JOBS but only at AITER entrypoints. When AITER_MAX_JOBS is unset, a valid MAX_JOBS is adopted as a deprecated ceiling and remains clamped by live CPU/memory limits. This change should Aiter localized control but still adapt to MAX_JOBS for backward compat. 
  
  https://github.com/ROCm/aiter/pull/5164/commits/7ba708457a302979ad825f94ee5898273fffe64f

Qubitium:
More throughts on why or why not AITER_MAX_JOBS. 

> Added backward compatibility for MAX_JOBS but only at AITER entrypoints. When AITER_MAX_JOBS is unset, a valid MAX_JOBS is adopted as a deprecated ceiling and remains clamped by live CPU/memory limits. This change should Aiter localized control but still adapt to MAX_JOBS for backward compat.
> 
> [7ba7084](https://github.com/ROCm/aiter/commit/7ba708457a302979ad825f94ee5898273fffe64f)

@valarLip  After the compat change. I thiink my view on Aiter owned AITER_MAX_JOBS come down to this:

> > MAX_JOBS
> 
> it's not only aiter but also sgl/vllm, all are using “MAX_JOBS”, instead of SGLANG_MAX_JOBS or VLLM_MAX_JOBS, why aiter need AITER_MAX_JOBS ?

I think Aiter only benefits from owning its own resource control policy instead of ceding it to a global MAX_JOBS that has no package-local context. Aiter entrypoints still respect existing MAX_JOBS ceiling users through the added compat path.

Best of both worlds.

valarLip:
Thanks for working through these one by one, and for bringing telemetry — measured numbers beat my static analysis, and I'll take the OPUS and prebuild results at face value.

I re-checked the current head (`bc37d67`) rather than reading the replies, and I want to correct something of mine first, then separate what `main` already covered from what it genuinely doesn't. I think that distinction is the thing still making this PR contentious, and it's mostly a framing problem at this point rather than a code problem.

### First, a correction on my side

**Review finding #1 (the `cpp_extension` import breaking blob-gen subprocesses) is genuinely fixed** by 87d9919, and my original repro was unfair: I ran `gen_instances.py` directly from a shell, which is not how aiter invokes it. With the `PYTHONPATH=AITER_ROOT_DIR` injection in `_run_blob_generator` the import resolves in both source-tree and pip-install layouts. Your characterisation is also right — the underlying fragility predates the PR and was hidden by `os.system` discarding the status; the PR only surfaced it. Switching to `subprocess.run(check=True)` is the right call.

I also confirmed the following are fixed and behave as intended now:

```
1st call -> 204 | env AITER_MAX_JOBS = None      # no longer self-promotes
2nd call, CPU budget crushed  -> recomputed      # no longer sticky
AITER_MAX_JOBS='' / 'auto'    -> 204             # falls back instead of raising
```

`get_worker_count()` returning `min(automatic, configured)` restores clamp-down semantics, and `adopt_legacy_max_jobs()` at 12 entrypoints is a good compromise on the `MAX_JOBS` question. Findings #2, #3, #6, #8, and the CI `=64` part of #7 all check out as resolved, and black is clean.

### What `main` already covered

Two things, and I think this is where the disagreement keeps recurring:

**`MAX_JOBS` clamping.** `main` clamps in two independent places — `core.py:631-643` (`if max_processes > max_jobs: os.environ["MAX_JOBS"] = str(max_jobs)`) and `cpp_extension.py:1534`. So an operator could already bound the fanout; the knob existed and was safe.

**Nested fanout for the prebuild path.** `main` sets `PREBUILD_THREAD_NUM = 5` at `setup.py:394`, and `cpp_extension.py:1549` reads it as `max_jobs = int(max_jobs) / int(prebuild_thread_num)`. That is exactly an outer-pool ÷ inner-ninja divisor, producing `5 modules × -j40` rather than an unbounded product. So "base has unbounded fanout" is not accurate for the prebuild path specifically.

Your telemetry showing no regression after removing it is convincing, and finer work-stealing granularity plausibly is better. My ask is only that the PR describe that removal as **a deliberate simplification validated by measurement**, not as fixing a missing guard. As written, every reviewer will stop at the same place I did.

### What `main` genuinely does not cover

These two are real, and neither is reachable by picking a different number — which was the premise of my earlier scoping suggestion, so consider that suggestion withdrawn on this point:

**1. cgroup blindness.** `psutil.virtual_memory().available` reports the host. Inside `docker run -m 32g` the limit is invisible, so the computed budget is wrong in precisely the environment that OOMs. An operator can only compensate by knowing the correct number per container — which is not a knob problem, it's a missing input. This is the strongest justification for the PR and I'd lead the description with it.

**2. No nesting divisor on the tuner path.** `main`'s `_init_task_start_times` sets only `_TASK_START_TIMES`; it establishes no budget, so each GPU tuner child inherits `MAX_JOBS=204` independently. `main`'s divisor exists only for prebuild.

One calibration on your `8 × 25 = 200` example: children building the *same* module serialise on `mp_lock`'s `FileBaton`, so the product only materialises when they build *different* modules concurrently. The gap is real; the worst case is an upper bound rather than the typical case.

### And what `main` simply does foolishly

`max_num_jobs_memory = int(free_memory_gb / 0.5)` — at 0.5 GB per job the memory term can only bind below ~102 GB available, so on any large builder it is dead code and the CPU term always wins. Raising it (you use 1.5 GB) is correct and overdue. Same for the computation being duplicated verbatim in `setup.py:31` and `core.py:621`.

### Two things I think are still open

**a. The runtime JIT path no longer honours `MAX_JOBS`.** `adopt_legacy_max_jobs()` is called at 12 entrypoints but not in `aiter/jit/core.py`, and `check_and_set_ninja_worker()` is now a one-line body — `get_worker_count()` — whose return value is discarded and which sets nothing. So under `MAX_JOBS=32 vllm serve ...`, aiter's in-process JIT compilation ignores it entirely, where `main` clamped.

That may well be deliberate under your "aiter shouldn't read the parent's `MAX_JOBS`" principle. But it's exactly the scenario you cited for why aiter needs its own control — aiter embedded in an sglang/vllm process — and it's the one place the compat path doesn't reach. Worth stating explicitly either way, since a downstream operator will observe it as a silent behavior change. (Also, `check_and_set_ninja_worker` no longer checks or sets anything; the name should follow.)

**b. Counting page cache as used has an unquantified cost.** Your reasoning is legitimate — the kernel does not guarantee timely reclaim, and a swap stall is worse than a slow build. But the magnitude matters, and cgroup page cache sits near the limit by construction, since it is only reclaimed under pressure:

```
32 GiB container, after clone + hipify:
  limit - memory.current  =  2.0 GiB  ->  1 worker
  limit - anon            = 29.0 GiB  -> 20 workers
```

So the conservative choice is not "slightly fewer workers", it's serialising the build in any long-lived container, silently, and getting worse the longer it runs. A middle position would be `limit - (anon + dirty + unreclaimable_slab)`, or discounting page cache by a factor rather than counting all of it. If you'd rather keep the current behavior, a one-line log when the cgroup path produces a budget below, say, 4 would at least make it diagnosable instead of silent.

### Net

The cgroup awareness and the tuner-path divisor are worth having and `main` cannot express them. The `MAX_JOBS` and prebuild-divisor changes are reverts-to-parity plus a measured simplification, and describing them that way would remove most of the friction here. Beyond (a) and (b) I don't have blocking concerns with the current head.

For the remaining review items I haven't mentioned — the `pa_gluon` GPU-resident pool, the FlyDSL ceiling, and `build_co`'s hipcc fanout — they're materially less dangerous now that `AITER_MAX_JOBS` clamps instead of overrides, since the budget itself is bounded. The one I'd still keep an eye on is `pa_gluon`: it's a CPU/host-RAM number governing GPU contexts, and GPU memory appears nowhere in the policy.

Qubitium:
Me

> The primary goal here is also not maximizing compiler throughput. It is preventing relatively short-lived aot jit compilation from creating `unbounded` fanout and ooming which is a much worser outcome and highly likely in base than slower but mostlly one-time compile before inference begines.

Review 

> Nested fanout for the prebuild path. main sets PREBUILD_THREAD_NUM = 5 at setup.py:394, and cpp_extension.py:1549 reads it as max_jobs = int(max_jobs) / int(prebuild_thread_num). That is exactly an outer-pool ÷ inner-ninja divisor, producing 5 modules × -j40 rather than an unbounded product. So "base has unbounded fanout" is not accurate for the prebuild path specifically.


@valarLip It was very wrong for me to use the term and imply `unbounded` in my earlier comment. All the fan-out code in `main` was already numerically bounded.  

I should've said some fan-out paths in main are not bounded against the right resource constraints, causing catastrophic edge cases to surface more often than they should.

> Your telemetry showing no regression after removing it is convincing, and finer work-stealing granularity plausibly is better. My ask is only that the PR describe that removal as a deliberate simplification validated by measurement, not as fixing a missing guard. As written, every reviewer will stop at the same place I did.

I think you phrased the current PR execution logic better than I actually did: `deliberate simplification`

This PR is, almost naive,  simplication of execution model of `main`'s fan-out and replacing some of the `N x Small M` for the most part with (Sometimes Larger, Sometimes Smaller N vs main) `N x 1`.  

PR's goal was never to make `main` faster or slower, it was actually only meant to simplify the fan-out and weed out the edge-cases. But during the review, and with some telemetry data is appears the simplification may actually be a net-win with parity to `main`'s compilation performance, though in very limited scenarios.

Qubitium:
> b. Counting page cache as used has an unquantified cost. Your reasoning is legitimate — the kernel does not guarantee timely reclaim, and a swap stall is worse than a slow build. But the magnitude matters, and cgroup page cache sits near the limit by construction, since it is only reclaimed under pressure:

> 32 GiB container, after clone + hipify:
>  limit - memory.current  =  2.0 GiB  ->  1 worker
>  limit - anon            = 29.0 GiB  -> 20 workers
> So the conservative choice is not "slightly fewer workers", it's serialising the build in any long-lived container, silently, > and getting worse the longer it runs. A middle position would be limit - (anon + dirty + unreclaimable_slab), or discounting page cache by a factor rather than counting all of it. If you'd rather keep the current behavior, a one-line log when the cgroup path produces a budget below, say, 4 would at least make it diagnosable instead of silent.

I will add the diagnostic log. For the current PR I think we should keep the current ultra-conservative memory formula. From past experience, I will never trust Linux kernel memory manager to release `claimable` memory in time to save the day.

Qubitium:
@valarLip Resolved the two remaining review issues in the Aiter worker-policy PR:
https://github.com/ROCm/aiter/pull/5164/commits/2a47871d20fdefd2c327c9cb8912fa39b94402ec

I agree with you that embedded/runtime jit should preserve the same worker-policy compatibility once execution reaches an Aiter-owned compile boundary.

I had wrongly assumed vllm/sglang would reach Aiter compilation only through explicit Aiter owned entrypoints. In practice, both frameworks directly import/call Aiter backend ops, and those calls can trigger Aiter's lazy runtime jit without passing through a standalone Aiter cli/aot entrypoint.

* Updated the PR description to more accurately reflect the intent of the PR and the execution-model changes after review/fixes.

* Runtime JIT now honors a valid positive legacy `MAX_JOBS` at the Aiter owned compile boundary when `AITER_MAX_JOBS` is unset, without mutating either environment variable. `AITER_MAX_JOBS` remains preferred, and every ceiling is still clamped by live CPU, host-memory, and cgroup limits. Standalone AITER entrypoints retain the compatibility path that adopts legacy `MAX_JOBS` into `AITER_MAX_JOBS`.

* Removed the dead runtime Ninja worker helper while keeping generic `MAX_JOBS` isolated from plain imports and the generic worker-policy API.

* Kept conservative page-cache accounting, added one-time `memory.stat` diagnostics with correct host-vs-cgroup attribution, and fail closed when finite cgroup usage cannot be read.

Qubitium:
Just noticed I accidentally squashed all previous commits into a force push yesterday.

Qubitium:
@valarLip Can you re-review and/or send this to other reviewers as well? I no longer have AMD MI350X compute to test this if needs arrive, I can rent out the gpu slice again for short burts/pr testing.

valarLip (changes requested):
## Re-review of `b0563e7`

Third pass. Verified in a worktree; the three worker-count collapses below are reproduced, not reasoned. `python3 -m pytest -q tests` passes (49 passed, 26 subtests).

First, credit where it's due since my last comment: you took the middle ground I suggested on page-cache accounting — keeping the conservative semantics but adding a diagnostic that names the reclaimable breakdown. That message is genuinely good:

```
AITER worker budget limited to 1 by cgroup memory: limit=32.00 GiB, current=31.70 GiB,
remaining=0.30 GiB, worker_estimate=1.40 GiB; cgroup current usage conservatively includes
page cache (anon=3.00 GiB, inactive_file=16.00 GiB, active_file=12.50 GiB, slab_reclaimable=1.00 GiB)
```

The problem is when it fires. See finding 4.

### The policy still collapses to 1 worker on three ordinary paths

**1. `aiter_worker_limits.py:183` — a container that has done I/O.**

Reproduced with a synthetic cgroup: 32 GiB limit, `memory.current` 31.7 GiB of which `anon` is 3 GiB and 29 GiB is reclaimable page cache + slab:

```
cgroup remaining bytes -> 322122548
get_automatic_worker_budgets() -> (204, 1)
get_worker_count() -> 1
```

`docker run --memory=32g` on a 128-core runner reaches this state partway through the AITER prebuild from `.o`/`.so` writes alone — page cache the kernel would drop on demand. Because the budget is recomputed on every call, the build starts at full parallelism and silently degrades to serial as it runs.

I understand the rationale you gave (the kernel doesn't guarantee timely reclaim, and a swap stall is worse than a slow build). But `_read_cgroup_memory_stat` already reads `inactive_file` and `slab_reclaimable` — it just never subtracts them, only prints them. Even a partial discount (`limit - (usage - inactive_file)`, or half of it) would keep the conservative posture while removing the pathological case, since clean inactive file pages are the part the kernel reclaims without any stall at all.

**2. `aiter_worker_limits.py:46` — psutil unavailable.**

```
_host_available_memory_bytes() -> 1500000000   # == EST_WORKER_RSS_BYTES
get_automatic_worker_budgets() -> (204, 1)
get_worker_count()             -> 1
```

`1_500_000_000 // 1_500_000_000 == 1`. You said a failed syscall is likely catastrophic, so 1 thread is the safe answer — that's reasonable for a *syscall* failure. But the bare `except Exception` also catches plain `ImportError`, and psutil missing is neither catastrophic nor rare: the code this replaced said so explicitly, `except ImportError: max_num_jobs_memory = max_num_jobs_cores` with the comment *"psutil may not be available during metadata extraction"*, and FlyDSL's `_memory_worker_cap` had `except Exception: return default_workers`. Both degraded to the CPU budget.

Splitting the two would preserve your intent: `ImportError` → fall back to the CPU budget; anything else → 1, and log which happened. As written, nothing is logged at all here — `_maybe_log_cgroup_memory_diagnostic` is gated on the cgroup observation, which is `None` on this path.

**3. `aiter_worker_limits.py:180` — a readable limit with an unreadable usage file.**

```
remaining -> 0        # fail-closed
worker_count -> 1
```

The diagnostic does fire and says so, which is good. But the asymmetry is worth reconsidering: an unreadable *limit* two blocks above `continue`s and skips the directory entirely, while an unreadable *usage* is treated as "the cgroup is completely full". An unknown usage against a known-64-GiB limit isn't evidence of memory pressure. This triggers on a hardened runtime that masks parts of `/sys/fs/cgroup`, on EIO during teardown, and on a v1 hierarchy whose limit parses but whose usage file is absent (see finding 6).

**4. `aiter_worker_limits.py:225` — the diagnostic only fires at budget ≤ 3, so the interesting collapses are silent.**

```python
if _cgroup_memory_diagnostic_emitted or memory_budget > 3:
    return
```

Verified: a 64 GiB cgroup at 52 GiB used gives `budgets -> (204, 8)`, `worker_count -> 8` — a 25× reduction — and logs **nothing**. The user sees a 25×-slower build with zero explanation. The threshold should be relative to `cpu_budget` (say, log whenever `memory_budget < cpu_budget // 4`) rather than a hardcoded 3.

Also `_cgroup_memory_diagnostic_emitted` is never reset, so in a long-lived `setup.py` or inference process the first (possibly mild) event consumes the one-shot and a later, worse collapse is never reported.

### Fanout

**5. `setup.py:376` — the new outer/ninja split still yields `ninja_workers == 1`.**

```python
total_workers = get_worker_count()
outer_workers = min(total_workers, max(1, len(all_opts_args_build)))
ninja_workers = max(1, total_workers // outer_workers)
```

`optCompilerConfig.json` has 128 entries, 118 after dropping `*tune` (verified), plus MoE/MHA variants — always more than `total_workers`. So `outer_workers == total_workers` and the division is always 1:

```
modules=118: total=204 outer=118 ninja_workers=1
modules=128: total=204 outer=128 ninja_workers=1
modules=150: total=204 outer=150 ninja_workers=1
```

That replaces `main`'s `5 modules × ninja -j20` with ~118 concurrent module drivers each at `-j1`. Two costs: the tail is now strictly serial (the largest module — `mha_varlen` / `moe_ck2stages`, thousands of TUs — compiles at `-j1` and is the last thing left once the short modules drain), and 118 simultaneous `build_module` threads each run `rename_cpp_to_cu`/hipify under the GIL and spawn a blob-generator subprocess, versus 5 before. The 1.5 GB/worker model was calibrated on one hipcc, not on "module driver + hipify + CK codegen process".

Your telemetry showed no regression, and I believe the measurement — but it was taken on a 24-core VM where `total_workers` is ~19 and the arithmetic gives a different shape than it does at 204. Worth re-running on a large host, or capping `outer_workers` (e.g. `min(total_workers // 4, len(...))`) so the inner `-j` stays > 1.

**6. `aiter/utility/mp_tuner.py:27` — the tuning pool initializer still pins `AITER_MAX_JOBS=1`.**

You've defended this as intentional (8 × 25 → 8 × 1), and the goal is right. My concern is only the divisor: `configure_worker_subprocesses()` hardcodes 1, where the correct value mirroring `setup.py` would be `total // pool_size`. On a 256-core / 8-GPU box that's 8 total compile jobs instead of ~25 each — an order of magnitude on tuner cold start, in a pool that is at most 8 processes wide and therefore was never the unbounded case.

Two smaller notes on the same function: the `OMP_NUM_THREADS`/`MKL_NUM_THREADS=1` half is inert, because the spawned child imports torch at `mp_tuner.py:9` before the initializer runs; and `csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py` builds its pool without any initializer, so the two tuners now behave differently.

**7. `aiter_worker_limits.py:379` — `NINJAFLAGS` is not a variable ninja reads, and two tests now cement the non-guarantee.**

Verified against the installed binary:

```
ninja 1.13.0.git.kitware.jobserver-pipe-1
NINJAFLAGS absent from binary
present: NINJA_STATUS
present: MAKEFLAGS
```

Upstream ninja reads only `NINJA_STATUS` and `MAKEFLAGS` (the latter solely to detect a GNU Make 4.4+ jobserver via `--jobserver-auth`, which `-j1` does not carry). `NINJAFLAGS` was proposed in ninja PR #1543 and rejected; it's a Qt5/AUR packaging convention consumed by package scripts, not by ninja.

So a nested non-AITER ninja inside a pool worker is unconstrained, while `README.md`, `docs/installation.rst` and the PR body all state "Ninja ... forced to one job", and `tests/test_worker_awareness.py:519` and `:534` assert on the value. What actually works here is `AITER_MAX_JOBS=1` (aiter passes an explicit `-j`) and `CMAKE_BUILD_PARALLEL_LEVEL=1`. Either drop the entry and fix the docs, or keep it and say it's for third-party packaging scripts.

**8. `aiter/jit/utils/cpp_extension.py:1545` — the prebuild opts out of the live recomputation.**

```python
if ninja_workers is not None:
    return max(1, int(ninja_workers))
```

`setup.py:376` computes `ninja_workers` once, before the pool starts, from the most favourable memory reading of the entire build, and passes that fixed integer to all ~150 module builds. The PR's stated policy is "recompute automatic CPU and memory limits on every policy call"; the one path where memory actually gets consumed is the one that snapshots. If memory tightens mid-build the prebuild keeps the stale value; if it loosens, a pessimistic `-j1` is never revised.

### Coverage gaps

**9. `aiter_worker_limits.py:167` — only the cgroup v2 sentinel is handled.**

```python
if raw_limit == "max":
    continue
```

cgroup v1 encodes unlimited as the integer `9223372036854771712` in `memory.limit_in_bytes`, so an unbounded v1 memory cgroup is recorded as a finite limit — which makes finding 3's fail-closed path fire on an *unlimited* cgroup whose usage file is missing, and prints `limit=8388608.00 GiB` in the diagnostic.

In the other direction, `memory.high` is never read. Kubernetes memory QoS commonly sets `memory.high` (the reclaim throttle) while leaving `memory.max == "max"`, so the code `continue`s, reports no bound, and launches ~0.8×nproc compilers into an 8 GB `memory.high` cgroup — thrashing in reclaim, which is precisely the swap-stall outcome finding 1 is conservative to avoid.

**10. `aiter_worker_limits.py:21` — cgroup CPU quota is not read, while every CI ceiling was removed.**

`_process_cpu_count()` uses `os.process_cpu_count()`/`sched_getaffinity`, neither of which reflects `cpu.max` or `cpu.cfs_quota_us`. A k8s pod with `cpu: 8` on a 192-core node and no affinity mask reports 192 → `cpu_budget = 153`. That's 200 lines of care for `memory.max` and nothing for the symmetric CPU knob.

It matters more now because the explicit ceilings are gone: `AITER_PREBUILD_MAX_JOBS: "64"` from `aiter-test.yaml` and `MAX_JOBS=64` from `atom-test.yaml`, `kimi-downstream.yaml`, `kimi-perf-downstream.yaml` and `vllm-disagg-ci-smoke-workflow.yaml` (the one surviving `MAX_JOBS=32`, in `ffm-bringup.yaml:192`, installs something else). So the memory cap — which findings 1-4 show is either 1 or unbounded — is the only remaining bound on CI concurrency.

**11. Still open from my last comment: the runtime JIT path doesn't honour `MAX_JOBS`.**

`adopt_legacy_max_jobs()` is called at 12 entrypoints but still not in `aiter/jit/core.py`, and `check_and_set_ninja_worker()` remains a one-line body whose return value is discarded. So under `MAX_JOBS=32 vllm serve ...`, aiter's in-process JIT ignores it, where `main` clamped. You may well intend this under "aiter shouldn't read the parent's `MAX_JOBS`" — but it's the one place the compat path doesn't reach, and it's exactly the embedded scenario you cited. A line in the docs either way would close it. (`check_and_set_ninja_worker` also no longer checks or sets anything; the name should follow.)

### Minor

**12. `aiter_worker_limits.py:193` — unsynchronized module globals mutated by the ~118-thread executor this PR introduces.**

`_last_cgroup_memory_observation` and `_cgroup_memory_diagnostic_emitted` are plain globals reached from every thread via `get_compile_worker_count()`. Thread A computes `remaining` from cgroup X, thread B overwrites the observation with cgroup Y, and A's diagnostic then prints Y's limit/usage/directory next to A's budget. And `if _cgroup_memory_diagnostic_emitted or ...` / `... = True` is check-then-set, so N threads can all pass the guard and emit N duplicate warnings, defeating the one-time contract in the docstring. Returning the observation from `_cgroup_memory_remaining_bytes()` instead of stashing it fixes both.

**13. `csrc/opus_gemm/gen_co/build_co.py:481` — `--jobs/-j` is still deleted.**

`jobs = get_worker_count_for(len(instances))` replaces `default=max(1, (os.cpu_count() or 2) // 8)`, whose help text read *"the default is deliberately well under nproc"*. On a 192-core box that's 153 concurrent hipcc instead of 24, guarded only by the 1.5 GB/worker estimate, which is below opus gfx1250 template hipcc peaks. You said you wanted telemetry here — the 24-core VM result won't exercise it. Separately, argparse now hard-errors `unrecognized arguments: -j8` for any external script using the documented flag; same silent CLI break in `op_tests/opus/device/setup.py`.

**14. Three dead artifacts.**

- `setup.py:423` — `def run(self): get_worker_count(); super().run()`. `get_worker_count` is explicitly pure (`tests/test_worker_awareness.py:294` asserts it never writes the environment), so this call has no effect; the code it replaced set `os.environ["MAX_JOBS"]`, which was what made the value visible to `super().run()`. It reads as if it configures the build.
- `aiter_worker_limits.py:274` — `_available_memory_bytes()` has no caller anywhere in the tree (verified), and its body duplicates the expression inlined at `:288-292`. A fix to one copy misses the other — the exact shape of findings 1 and 3.
- `aiter/jit/utils/cpp_extension.py:1565` — `if num_workers is not None:` is now always true, since `_get_num_workers` returns `int`.

**15. `.github/scripts/split_tests.sh:53` and `aiter_test.sh` — the CI wiring for `tests/` is inert.**

The PR's own change to `split_tests.sh` (vs merge-base) is:

```diff
-    mapfile -t ALL_FILES < <(find "$TEST_DIR" -maxdepth 1 -name 'test_*.py' -type f | LC_ALL=C sort)
+    mapfile -t ALL_FILES < <(
+        find "$TEST_DIR" -maxdepth 1 -name 'test_*.py' -type f | LC_ALL=C sort
+    )
```

— identical command in a subshell, zero behavioural difference, while the header comment was updated to "shards Aiter and Triton tests".

And `aiter_test.sh`'s new `find tests -maxdepth 1 -name 'test_*.py'` sits inside `if [[ -z "${AITER_TEST:-}" ]]` (`:27-33`), but `aiter-test.yaml:714` unconditionally sets `AITER_TEST` per shard from `aiter_shard_${shard_idx}.list`. So that branch is never taken in CI. The new suite runs only via the separate "Focused regression tests" step, gated on `matrix.label == 'MI35X' && matrix.shard_idx == 0` — 1 of 16 matrix legs, with no MI300X coverage of the worker-policy suite at all.

---

Net: findings 1-4 are the ones I'd fix before merge — they are the policy failing its own guarantee, and 4 is a two-line change that would have surfaced 1 and 3 in your own testing. 5 and 6 are shape questions where your telemetry and my arithmetic disagree because they were taken at different core counts; a run on a 128+ core host would settle both.

Qubitium:
> 2. aiter_worker_limits.py:46 — psutil unavailable.

> _host_available_memory_bytes() -> 1500000000   # == EST_WORKER_RSS_BYTES 
> get_automatic_worker_budgets() -> (204, 1)
> get_worker_count()             -> 1
> 1_500_000_000 // 1_500_000_000 == 1. You said a failed syscall is likely catastrophic, so 1 thread is the safe answer — > that's reasonable for a syscall failure. But the bare except Exception also catches plain ImportError, and psutil missing >is neither catastrophic nor rare: the code this replaced said so explicitly, except ImportError: max_num_jobs_memory = > max_num_jobs_cores with the comment "psutil may not be available during metadata extraction", and FlyDSL's ?>_memory_worker_cap had except Exception: return default_workers. Both degraded to the CPU budget.
>
> Splitting the two would preserve your intent: ImportError → fall back to the CPU budget; anything else → 1, and log  which happened. As written, nothing is logged at all here — _maybe_log_cgroup_memory_diagnostic is gated on the  ?cgroup observation, which is None on this path.

@valarLip  Can we skip this? `psutil` is part of the `aiter` package requirement. If user fails to install `aiter` properly, like just expects git clone and run it, it should fail at first `aiter` entry point but outside the scope of this pr?

This should really go inside a separate pr that has an aiter entry guard that checks all the dependent packages and fails early and/or warns users they are missing package we expected them to install and to install the package, not just git clone it to run.

Qubitium:
> 1. aiter_worker_limits.py:183 — a container that has done I/O.
>
> Reproduced with a synthetic cgroup: 32 GiB limit, memory.current 31.7 GiB of which anon is 3 GiB and 29 GiB is reclaimable page cache + slab:

> cgroup remaining bytes -> 322122548
>get_automatic_worker_budgets() -> (204, 1)
>get_worker_count() -> 1
->docker run --memory=32g on a 128-core runner reaches this state partway through the AITER prebuild from .o/.so writes alone — page cache the kernel would drop on demand. Because the budget is recomputed on every call, the build starts at full parallelism and silently degrades to serial as it runs.
->
->I understand the rationale you gave (the kernel doesn't guarantee timely reclaim, and a swap stall is worse than a slow build). But _read_cgroup_memory_stat already reads inactive_file and slab_reclaimable — it just never subtracts them, only prints them. Even a partial discount (limit - (usage - inactive_file), or half of it) would keep the conservative posture while removing the pathological case, since clean inactive file pages are the part the kernel reclaims without any stall at all.


@valarLip  A one worker limit is reasonable here. With only about 0.3 GiB of headroom before accounting for reclaimable cache, even one worker’s estimated 1.5 GB footprint may require reclaim. Increasing concurrency would make us more dependent on timely reclaim, which this policy deliberately avoids assuming. 128 core system limited to 32gb is asking for trouble here as we have documented that aot/jit can use ~1.5GB per fan-out process. Main had ~2GB but I reduced it to 1.5GB but that can be tweaked based on max observations. Pretty sure the main's 2GB was rounded up from ~1.5xx as I did not find any git comments on why 2GB was used. 

> since clean inactive file pages are the part the kernel reclaims without any stall at all.

Clean file pages are generally cheaper to reclaim, but being reclaimable is different from having enough memory immediately available to satisfy an allocation request. Reclaim itself requires processing, and the allocating process can stall while that work happens. Linux provides no guarantee of immediate reclaim, and my past experience reinforces that concern. One process stall is fine, but fan-out stall creates total havoc.

This policy is deliberately conservative, with an acknowledged compilation-throughput tradeoff. The priority here is to prevent aiter's compilation from creating memory pressure that destabilizes the inference process or other workloads.

We have to keep in mind aiter aot/jit is most likely not the primary process or even the most important one on this system.

Qubitium:
>3. aiter_worker_limits.py:180 — a readable limit with an unreadable usage file.
>
>remaining -> 0        # fail-closed
worker_count -> 1
the diagnostic does fire and says so, which is good. But the asymmetry is worth reconsidering: an unreadable limit two blocks above continues and skips the directory entirely, while an unreadable usage is treated as "the cgroup is completely full". An unknown usage against a known-64-GiB limit isn't evidence of memory pressure. This triggers on a hardened runtime that masks parts of /sys/fs/cgroup, on EIO during teardown, and on a v1 hierarchy whose limit parses but whose usage file is absent (see finding 6).


@valarLip  An unknown usage is not evidence of memory pressure, but a known limit with unknown headroom is not a safe basis for increasing compilation concurrency either, so the one-worker fallback and diagnostic are resonable.

Qubitium:
> 4. aiter_worker_limits.py:225 — the diagnostic only fires at budget ≤ 3, so the interesting collapses are silent.
>
>if _cgroup_memory_diagnostic_emitted or memory_budget > 3:
    return
Verified: a 64 GiB cgroup at 52 GiB used gives budgets -> (204, 8), worker_count -> 8 — a 25× reduction — and logs nothing. The user sees a 25×-slower build with zero explanation. The threshold should be relative to cpu_budget (say, log whenever memory_budget < cpu_budget // 4) rather than a hardcoded 3.
>
>Also _cgroup_memory_diagnostic_emitted is never reset, so in a long-lived setup.py or inference process the first (possibly mild) event consumes the one-shot and a later, worse collapse is never reported.

@valarLip  Addressed in [d67aa2b2](https://github.com/ROCm/aiter/pull/5164/commits/d67aa2b2cd1a772b866df4b976868b828cf53bc5) with ceiling-aware diagnostics and rate-limited reporting when throttling worsens, while preserving the worker policy. However,  a 25× reduction in concurrency does not necessarily imply a 25× slower build.

Qubitium:
>Fanout
>5. setup.py:376 — the new outer/ninja split still yields ninja_workers == 1.
>
>total_workers = get_worker_count()
outer_workers = min(total_workers, max(1, len(all_opts_args_build)))
ninja_workers = max(1, total_workers // outer_workers)
optCompilerConfig.json has 128 entries, 118 after dropping *tune (verified), plus MoE/MHA variants — always more than total_workers. So outer_workers == total_workers and the division is always 1:
>
>modules=118: total=204 outer=118 ninja_workers=1
modules=128: total=204 outer=128 ninja_workers=1
modules=150: total=204 outer=150 ninja_workers=1
That replaces main's 5 modules × ninja -j20 with ~118 concurrent module drivers each at -j1. Two costs: the tail is now strictly serial (the largest module — mha_varlen / moe_ck2stages, thousands of TUs — compiles at -j1 and is the last thing left once the short modules drain), and 118 simultaneous build_module threads each run rename_cpp_to_cu/hipify under the GIL and spawn a blob-generator subprocess, versus 5 before. The 1.5 GB/worker model was calibrated on one hipcc, not on "module driver + hipify + CK codegen process".
>
>Your telemetry showed no regression, and I believe the measurement — but it was taken on a 24-core VM where total_workers is ~19 and the arithmetic gives a different shape than it does at 204. Worth re-running on a large host, or capping outer_workers (e.g. min(total_workers // 4, len(...))) so the inner -j stays > 1.

@valarLip  I don’t have access to a >24-core system with an AMD Instinct GPU. The proposed total_workers // 4 split preserves more inner parallelism, but without comparative measurements, we cannot conclude that it improves build time or memory behavior.

This PR deliberately limits nested concurrency to simplify resource control. I’m open to addressing demonstrated regressions, but I don’t think replacing this policy with another unvalidated heuristic resolves the concern. More adaptive scheduling could follow in a separate PR with visibility into active compilation work and resource usage. If large host validation is required here, can you help run the comparison on available infrastructure?

Qubitium:
@valarLip I think we’re revisiting the same policy tradeoff without an agreed way to solve it. This PR aims to simplify the scattered concurrency controls and reduce the risk of excessive compilation fan-out.  Maximizing compilation throughput is not its goal, but regressions introduced by the change still deserve evaluation, based on real benchmarks.

Can we agree on representative tuner commands, kernels/model configurations, and hardware/container limits for comparing both policies (main vs pr)? We can measure then agree on acceptable tradeoffs. I can test on the hardware available to me, but I would need your help with the large-core-count systems.

I think the acceptance criterion should be improved resource control with an agreed, measured performance tradeoff, not a requirement to improve compilation speed as well. Further throughput optimization can follow in a separate PR, while any material regressions introduced here should be evaluated before merging.

Qubitium:
@valarLip Please trigger CI test so I can at least check if there are regressions. We are at the stage where we need real data, regression tests on real kernels on real hw, no more theories.

valarLip:
Re-reviewed `9a8167b8a1dc340fc708dba6acec8d5883ac88ac` and checked the earlier review discussion. Most earlier correctness issues are now addressed, including ctypes result serialization, blob-generator import/error propagation, live budget recomputation, ceiling enforcement, CPU quota awareness, and runtime `MAX_JOBS` compatibility. The shared compilation policy is a useful improvement.

Two P2 issues remain:

1. **Keep GPU execution concurrency separate from CPU compilation budgets.** [PA-Gluon pool selection](https://github.com/ROCm/aiter/blob/9a8167b8a1dc340fc708dba6acec8d5883ac88ac/csrc/cpp_itfs/pa_gluon_aot/pa_decode_gluon_aot_prebuild.py#L819-L829) now uses the CPU/host-RAM budget, but its tasks allocate device tensors on hardcoded `cuda:0` and execute kernels. On a 256-CPU host, the live policy returns 204; running the actual standard accuracy configuration-generation path with only the executor mocked selected **144 processes** for the first batch, versus the previous default of 32. Host CPU/RAM headroom does not bound GPU memory or execution contention. Please retain a separate GPU-test concurrency bound, or split CPU precompilation from bounded GPU execution. A manual `AITER_MAX_JOBS` override does not correct the automatic default. No GPU OOM was induced or observed in this review.

2. **Run pytest modules through pytest rather than reporting import-only execution as success.** [aiter_test.sh](https://github.com/ROCm/aiter/blob/9a8167b8a1dc340fc708dba6acec8d5883ac88ac/.github/scripts/aiter_test.sh#L27-L34) now discovers `tests/test_*.py`, but still executes ordinary files with `python3 "$file"`. I ran:

   ```bash
   AITER_TEST=tests/test_worker_cpu_quota.py GPU_ARCHS=gfx950 \
   ```

   It printed “Test passed” and “All tests passed” in about 0.082 seconds, while executing **none of the module's 24 pytest cases**. The separate CI pytest step does cover them on its selected leg; the defect is this runner's false success result. Please dispatch `tests/` through pytest and preserve the specialized script/torchrun entrypoints for operator tests.

**Performance still needs representative build measurements.** The [fixed outer/Ninja allocation](https://github.com/ROCm/aiter/blob/9a8167b8a1dc340fc708dba6acec8d5883ac88ac/setup.py#L374-L400) gives every module `-j1` when module count meets or exceeds the budget, and cannot redistribute idle capacity to the remaining large module. A controlled real-Ninja test with synthetic commands, a 16-worker budget, and uneven module sizes took 0.336 s with the previous 5×3 allocation versus 1.004 s with the PR's 16×1 allocation. This demonstrates the scheduling mechanism; **it does not establish a 3× regression in actual HIP compilation**. A comparison of the same cold prebuild/module set on a large-core machine, recording wall time, peak RSS, and active compiler counts, would quantify the tradeoff. I am treating conservative page-cache accounting as an explicit policy choice, not another newly demonstrated correctness bug.

**Validation:** 126 tests and 31 subtests passed in `ljin_dev`. No full HIP/CK/FlyDSL cold build or serving benchmark was run.

Please also **resolve the current merge conflicts with `main`**—GitHub currently reports `mergeable: CONFLICTING` and `mergeStateStatus: DIRTY`—and rerun the relevant checks on the updated head.

PR #5165, opened 2026-09-01
[Triton/Gluon] Add vocab-parallel cross-entropy kernel

WuLei-AMD:
## Summary

  Adds a Triton-based vocab-parallel cross-entropy loss for tensor-parallel
  training, adapted from TransformerEngine's cross-entropy kernels.

  ### New files

  `aiter/ops/triton/_triton_kernels/cross_entropy.py` — three JIT kernels:
  - `_ce_local_softmax_stats_kernel` — per-TP-rank online softmax pass;
    computes row-wise `(max, sum-of-exp, target-logit)` without writing softmax
    output, so peak memory scales with `n_rows × 3` not `n_rows × V`
  - `_ce_fused_loss_grad_kernel` — merges gathered per-rank stats into global
    softmax, computes CE loss, and writes the gradient in-place into the logit
    tensor; supports `label_smoothing`, `reduce_loss`, and `ignore_index`
  - `_ce_grad_scale_kernel` — backward element-wise multiply; skipped as a fast
    path when `grad_output == 1.0`

  `aiter/ops/triton/cross_entropy.py` — Python orchestration layer:
  - `cross_entropy_forward` — local stats → `all_gather` → fused loss+grad
  - `cross_entropy_forward_chunked` — same, chunked over the row dimension to
    cap peak activation memory at `chunk_rows × V` instead of `B×SQ × V`
  - `cross_entropy_backward` — scales the stored gradient by `grad_output`

  ### Test

  `op_tests/triton_tests/test_cross_entropy.py` — 49 pytest cases covering:
  - Loss correctness vs `F.cross_entropy` across dtype × vocab size × batch
  - `label_smoothing` (0 / 0.1 / 0.2) × `reduce_loss` (True / False)
  - `ignore_index` rows produce zero loss and zero gradient
  - Non-power-of-2 vocab sizes (127, 511, 32001)
  - Backward gradient scaling and identity fast-path
  - End-to-end autograd gradient matches torch autograd
  - Chunked forward matches non-chunked for various `chunk_rows`

  ## Test plan

  - [ ] `python3 -m pytest op_tests/triton_tests/test_cross_entropy.py` — 49/49 passed on MI308X (gfx942)
  - [ ] `black --check` and `ruff check` pass on all changed files

### DIFF
--- a/op_tests/triton_tests/test_cross_entropy.py
+++ b/op_tests/triton_tests/test_cross_entropy.py
@@ -253,7 +253,7 @@ def test_cross_entropy_forward_rejects_out_of_range_target(bad):
     logits = torch.randn(B, SQ, V, device="cuda")
     target = torch.randint(0, V, (B, SQ), device="cuda")
     target[0, 0] = bad
-    with pytest.raises(AssertionError):
+    with pytest.raises(ValueError):
         cross_entropy_forward(logits, target, 0.0, False, None, -100)
 
 
--- a/op_tests/triton_tests/test_cross_entropy.py
+++ b/op_tests/triton_tests/test_cross_entropy.py
@@ -6,8 +6,9 @@ Tests the single-GPU (dist_group=None) path of cross_entropy_forward,
 cross_entropy_forward_chunked, and cross_entropy_backward against
 torch.nn.functional.cross_entropy as the reference.
 
-Multi-GPU / TP-parallel paths require a real process group and are not
-covered here; those are exercised by the distributed integration tests.
+The multi-GPU / TP-parallel path (world_size > 1: gathered-stat layout and
+TP label smoothing) requires a real process group and is currently NOT
+covered by any test — it needs a multi-process TP harness (follow-up).
 """
 
 import pytest
@@ -38,8 +39,8 @@ def _ref_forward(logits_2d, target_1d, label_smoothing, reduce_loss, ignore_idx)
 # ── forward correctness ──────────────────────────────────────────────
 
 
-@pytest.mark.parametrize("V", [128, 512, 1000, 4096, 32001])
-@pytest.mark.parametrize("B_SQ", [1, 4, 64])
+@pytest.mark.parametrize("V", [128, 32001])  # aligned + large non-aligned
+@pytest.mark.parametrize("B_SQ", [1, 64])
 @pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16])
 def test_cross_entropy_forward_basic(V, B_SQ, dtype):
     """Loss values match F.cross_entropy for standard CE (no smoothing)."""
@@ -63,8 +64,10 @@ def test_cross_entropy_forward_basic(V, B_SQ, dtype):
     torch.testing.assert_close(loss_triton.float(), ref.float(), atol=1e-3, rtol=1e-3)
 
 
+# ls=0.0 (standard CE) is covered by test_cross_entropy_forward_basic; the
+# smoothing term is linear in ls, so one non-zero value is enough.
 @pytest.mark.parametrize("reduce_loss", [True, False])
-@pytest.mark.parametrize("label_smoothing", [0.0, 0.1, 0.2])
+@pytest.mark.parametrize("label_smoothing", [0.1])
 def test_cross_entropy_forward_label_smoothing(reduce_loss, label_smoothing):
     torch.manual_seed(0)
     B, SQ, V = 2, 8, 256
@@ -128,7 +131,51 @@ def test_cross_entropy_forward_ignore_index():
             assert grad_2d[row].abs().max().item() == 0.0, f"row {row} grad not zero"
 
 
-@pytest.mark.parametrize("V", [127, 511, 32001])  # non-power-of-2 vocab sizes
+def test_cross_entropy_forward_reduce_with_ignore_index():
+    """reduce_loss=True must divide by the non-ignored row count, not n_rows."""
+    torch.manual_seed(13)
+    B, SQ, V = 4, 8, 256
+    ignore_idx = -100
+    logits = torch.randn(B, SQ, V, device="cuda")
+    target = torch.randint(0, V, (B, SQ), device="cuda")
+    # ignore a third of the rows
+    flat = target.reshape(-1)
+    flat[::3] = ignore_idx
+
+    loss_triton, grad = cross_entropy_forward(
+        logits.clone(),
+        target,
+        label_smoothing=0.0,
+        reduce_loss=True,
+        dist_group=None,
+        ignore_idx=ignore_idx,
+    )
+
+    # F.cross_entropy(reduction="mean") already averages over non-ignored rows.
+    ref_loss = _ref_forward(
+        logits.reshape(B * SQ, V), target.reshape(B * SQ), 0.0, True, ignore_idx
+    )
+    torch.testing.assert_close(
+        loss_triton.float(), ref_loss.float(), atol=1e-3, rtol=1e-3
+    )
+
+    # gradient must also be normalized by the non-ignored count
+    logits_ref = logits.clone().requires_grad_(True)
+    F.cross_entropy(
+        logits_ref.reshape(B * SQ, V),
+        target.reshape(B * SQ),
+        reduction="mean",
+        ignore_index=ignore_idx,
+    ).backward()
+    torch.testing.assert_close(
+        grad.reshape(B * SQ, V).float(),
+        logits_ref.grad.reshape(B * SQ, V).float(),
+        atol=1e-3,
+        rtol=1e-3,
+    )
+
+
+@pytest.mark.parametrize("V", [127, 511])  # non-power-of-2 vocab sizes
 def test_cross_entropy_forward_nonaligned_vocab(V):
     """Unaligned vocab sizes must not cause silent OOB reads."""
     torch.manual_seed(3)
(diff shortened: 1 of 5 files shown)


op_tests/triton_tests/test_cross_entropy.py
```diff
@@ -0,0 +1,388 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+"""Correctness tests for the vocab-parallel cross-entropy Triton kernels.
+
+Tests the single-GPU (dist_group=None) path of cross_entropy_forward,
+cross_entropy_forward_chunked, and cross_entropy_backward against
+torch.nn.functional.cross_entropy as the reference.
+
+The multi-GPU / TP-parallel path (world_size > 1: gathered-stat layout and
```
vgokhale: This sounds like a TODO, not something that should be in a docstring. Please remove this, add an assert to guard against this until the "follow-up" is added.

vgokhale:
1) All kernels need a repr decorator. Please see examples of current kernels.

2) Every kernel should have a benchmark at op_tests/op_benchmarks/triton. Please add one per kernel.

vgokhale:
1. torch.equal() in the backward hot path forces a device sync — [cross_entropy.py:286](vscode-webview://1g38rf2ae4pgbul18p1dp4921ret77arv9vu5h87qc2eh3d1dlnf/aiter/ops/triton/cross_entropy.py#L286)


if not is_cg_capturable and torch.equal(
    grad_output, torch.tensor(1.0, device=grad_output.device)
):
torch.equal returns a Python bool, so it must synchronize the device. This runs on every backward call, stalling the pipeline to avoid one elementwise kernel — the guard costs more than what it skips. It also allocates a fresh device tensor per call. Separately, torch.equal compares dtype as well as value, so with a bf16 grad_output (normal in mixed-precision training) it returns False and the fast path never fires anyway. Worth confirming against your torch version, but the sync is unconditional either way.

2. n_valid == 0 produces Inf gradients — [cross_entropy.py:85](vscode-webview://1g38rf2ae4pgbul18p1dp4921ret77arv9vu5h87qc2eh3d1dlnf/aiter/ops/triton/cross_entropy.py#L85)

If every row in a batch is ignore_idx (real in heavily padded batches), n_valid = 0. The kernel then computes blk / n_valid → ±Inf into the gradient, and the wrapper returns loss_1d.sum() / n_valid → NaN. F.cross_entropy also gives NaN loss there, so the loss matches, but PyTorch gives a zero gradient where this gives Inf — that difference poisons the whole step. Untested; the ignore-index tests only ignore a subset of rows.

3. Out-of-range targets fail silently. A label y >= world_size * n_cols (or any negative value that isn't ignore_idx) leaves X_y = -inf on every rank, so loss = -(-inf - m - log d) = +inf and propagates with no diagnostic. A range check in the wrapper is cheap.

4. label_smoothing under TP is known-wrong but accepted. The kernel docstring says smoothing "is only correct for world_size == 1" because scaled_x_sum only covers the local shard — yet cross_entropy_forward happily takes label_smoothing > 0 with a non-None dist_group and returns wrong numbers. This should raise, not be a comment.

5. The cross-rank merge leans on an implicit -inf sentinel — [_triton_kernels/cross_entropy.py:138](vscode-webview://1g38rf2ae4pgbul18p1dp4921ret77arv9vu5h87qc2eh3d1dlnf/aiter/ops/triton/_triton_kernels/cross_entropy.py#L138)


ori_Xy = tl.maximum(ori_Xy, Xy_new)
This recovers the target logit by taking the max across ranks, correct only because non-owning ranks wrote -inf. A genuinely -inf logit (masked vocab entry) is indistinguishable from "not my shard". Works as written; deserves a comment since it isn't obvious.

vgokhale:
So couple things first:

First, several review threads were resolved without a corresponding change. The benchmark and TP-smoothing ones are genuinely handled, but **the `NUM_WARPS` thread (`cross_entropy.py:32`) was resolved with the code untouched** — see item 1 below. If a comment is being declined rather than actioned, a one-line reply saying so is more useful than resolving it silently.

Second, on the TP-smoothing thread specifically: the ask was to gather the global logit sum so smoothing works under tensor parallelism; the resolution instead forbids the combination. That's a legitimate call, but it's a scope reduction, so worth stating explicitly in the PR description — a Lumen caller wanting label smoothing on a TP run now has no path.

### Blocking

**1. `NUM_WARPS = 16` is still fixed regardless of `BLOCK_SIZE` (`cross_entropy.py:32`, used at :119, :148, :264, :291, :346).**
16 warps × 64 lanes is 1024 threads, the maximum workgroup. For a small vocab shard that means most lanes idle and a 1024-wide reduction tree over very few values. `softmax.py` — the closest peer, same shape of op — uses `num_warps=8` and also sets `waves_per_eu=2` and `num_stages=2`, none of which are set here. The benchmark now exists, so this is measurable: could you post numbers for a couple of `num_warps` values across the vocab range you already sweep?

**2. `MAX_FUSED_SIZE = 65536 // 2` hardcodes a 2-byte element, but fp32 is accepted, tested, and is a benchmark mode.**
`softmax.py` derives this as `65536 // x.element_size()`. At `V=32001` fp32 this picks `BLOCK_SIZE=32768` where the element-size-derived value would be `16384` — at 1024 threads that is 32 live fp32 values per lane in the reduction loop, on top of the exp intermediates. One-line fix, and it interacts with item 1.

**3. The forward now takes two host syncs per call, via bare `assert`s.**
`assert valid_target.all()` (:94, :223) forces `__bool__` on a CUDA tensor. The comment acknowledges the trade, and I agree the check earns its keep — but two things make the current form awkward:
- A bare `assert` is stripped under `python -O`, so optimized runs silently lose *both* the check and the sync. Behaviour and performance then differ between `-O` and not, which is a bad property for a training hot path.
- A training loop wants this on for the first few steps and off thereafter.

A `validate_targets: bool = True` parameter would cover both and keeps the negative tests working as-is. (`torch._assert_async` avoids the sync entirely, but it surfaces the failure at a later sync point, which would make `pytest.raises(AssertionError)` unreliable — so I'd go with the flag.)

**4. `clamp_min(1)` silently diverges from the stated reference.** An all-ignored batch now returns loss `0.0`; `F.cross_entropy(reduction="mean")` returns `nan`. I think 0.0 is the better behaviour and the test locks it in — but the tests compare against `F.cross_entropy` throughout, so the one case where the op deliberately disagrees belongs in the public docstring, not only in a code comment.

**5. `tl.debug_barrier()` (`_triton_kernels/cross_entropy.py:178`) and the target-column round trip.** The store loop writes the gradient to global memory, then the kernel barriers and re-loads `X_ptr + y - vocab_start` to add `-(1-ls)/n_valid`. Testing `offs == y - vocab_start` inside the store loop would correct it in-register and remove the barrier, the load and the store. It also fixes a precision loss: the re-load reads the value *after* rounding to bf16, so the correction becomes `round_bf16(p_y/n_valid) - 1/n_valid` rounded again — worst exactly where `p_y → 1` and the true gradient → 0.

**6. `_ce_grad_scale_kernel` is a bandwidth-bound elementwise multiply launched one program per row** with a V-wide block and 16 warps. A flat 1-D grid over `B·SQ·V` would be simpler and faster; better still, fold the scale into the fused kernel when `grad_output` is known at forward time. As written, backward re-reads and re-writes the whole logits tensor for `X *= g`.

**7. `assert reduce(mul, list(target.size())) == n_rows` (:65)** is `target.numel() == n_rows` with two extra imports — and it is a bare `assert` guarding an input contract that the same file guards elsewhere with `raise ValueError`.

**8. No `AiterTritonLogger`.** 47 of 102 public wrapper modules log on entry, including `softmax.py`. Minor consistency point.

**9. `BLOCK_SIZE = min(MAX_FUSED_SIZE, triton.next_power_of_2(V))` is repeated in all three public functions** (:66, :209, :338).

### Still open: the TP path has no coverage

All 28 collected cases pass `dist_group=None`. The `world_size > 1` stat-merge loop and the `all_gather_into_tensor` layout — the reason a *vocab-parallel* CE exists — remain untested, and the benchmark is single-GPU too. The new assert removes the known-wrong smoothing combination, which helps, but the gathered-stat merge itself is still unexercised.

WuLei-AMD:
> **[vgokhale](/vgokhale) **

 Thanks for the detailed review. All items addressed:
  
   1. NUM_WARPS: Changed to num_warps = min(16, max(1, BLOCK_SIZE // 64)) with waves_per_eu=2, num_stages=2. Benchmarked on MI308X:
  
   | Shape        | nw=4     | nw=8     | nw=16    |
   |--------------|----------|----------|----------|
   | 4096×32000   | 1.32 ms  | 1.64 ms  | 1.44 ms  |
   | 4096×128256  | 5.41 ms  | 6.08 ms  | 5.65 ms  |
   | 4096×151936  | 6.63 ms  | 7.48 ms  | 6.95 ms  |
   | 16384×32000  | 5.16 ms  | 6.47 ms  | 5.67 ms  |
   | 16384×128256 | 22.33 ms | 25.70 ms | 21.93 ms |
  
   1. num_warps=4 wins for all practical vocab sizes (BLOCK_SIZE=32768 → 256 cols/warp). nw=16 leads by ~2% only at the largest shape.
   2. MAX_FUSED_SIZE: Now 65536 // _input.element_size() per call, matching softmax.py.
   3. bare assert → validate_targets param: Added validate_targets: bool = True to both forward functions; replaced assert with raise ValueError so python -O doesn't silently drop the check.
   4. clamp_min(1) docstring: reduce_loss arg now documents the 0.0 vs nan divergence from F.cross_entropy.
   5. tl.debug_barrier(): Removed. Target-column correction is now applied in-register via tl.where(offs == local_y, ...) inside the store loop — eliminates the barrier, an extra global load/store, and the bf16 round-trip.
   6. assert reduce(mul,...): Replaced with if target.numel() != n_rows: raise ValueError(...).
   7. AiterTritonLogger: Added _LOGGER with entry-point info() on all three public functions.
   8. BLOCK_SIZE repetition: Extracted to _block_size_and_warps(element_size, n_cols) helper.
  
   TP coverage: Added op_tests/multigpu_tests/triton_test/test_cross_entropy_tp.py using mp.Pool + RCCL all_gather_into_tensor (world_size=2). Three cases: loss parity, grad shard correctness (max_diff=0.00e+00), and ignore_index under TP — all pass.
  
   Item 6 (_ce_grad_scale_kernel fold) deferred: folding into the fused kernel requires knowing grad_output at forward time, which is framework-dependent; will follow up separately.

Dewei-Wang-sh (approved):

PR #5166, opened 2026-09-01
[JIT] Discover matching system ROCm headers for Python SDK compilers

Qubitium:
## Problem

When `hipcc` is provided by the Python ROCm SDK, AITER can resolve `ROCM_HOME` to the Python environment (for example, `/opt/python314`). If `rocm-sdk-devel` is not installed but matching ROCm development packages are present under `/opt/rocm`, `include_paths(cuda=True)` adds the Python include directory but not the system development headers.

This breaks AITER JIT modules that directly include rocPRIM. The observed `module_rmsnorm_quant` build failed with:

```text
fatal error: 'rocprim/rocprim.hpp' file not found
```

The header was present at `/opt/rocm/include/rocprim/rocprim.hpp`; it was simply absent from the generated compiler command.

## Fix

- Detect the alternatives-managed `/opt/rocm/include` development tree when selected Python SDK paths do not provide rocPRIM.
- Read `hip/hip_version.h` and require its HIP major/minor version to match the active compiler before adding system headers.
- Refuse missing or mismatched system headers so AITER cannot silently combine incompatible ROCm releases.
- Add focused regression tests covering matching, mismatched, missing-rocPRIM, and complete `include_paths()` behavior.

## Validation

- `python -m pytest -q tests/test_cpp_extension_rocm_headers.py` — 4 passed.
- Built and imported a real rocPRIM PyTorch extension with `CPATH`, `CPLUS_INCLUDE_PATH`, `ROCM_HOME`, and `ROCM_PATH` unset.
- The generated Ninja command contained `-isystem /opt/rocm/include`, and the probe returned `ROCPRIM_COMPILE_OK 1` on gfx950.


Qubitium:
@valarLip  Ran into this bug a second time today. The core issue is `hipcc` from python pkg (not the hipcc installed by apt) only looks in the python venv local for rocm dev libs but there is no official amd repo or package that actually installs rocm dev libs to the pyhton local dev, they are always installed to /opt/rocm path. This is very strange to me. 

> hipcc may come from the Python ROCm SDK 

image

PR #5174, opened 2026-09-01
Rebase onto #5142, and stop the validator reporting coverage it did not obtain

jhinpan:
Rebased onto #5142's current head, plus four fixes found by *running* this skill on a real
kernel PR rather than by reading it.

## The case

ROCm/aiter#4538 rewrites the gfx950 FP8 MQA logits kernel. At `a8d46dcb0` the validator says:

```
verdict=PASS  findings=2  exit 0        (93 seconds)
```

#4538 asserts at `num_heads=16` on gfx950 (`_auto_variant` sends every `H <= 32` to a tag whose
builder needs `H % 32 == 0`). The pre-PR base computes `H=16` correctly, so it is a regression;
the PR already ships MFMA16 variants that work there, so it is a selection bug. Its performance
claim also went unmeasured.

Nothing in that report is false. Every number in it is weaker than it reads:

| stage | published | rested on |
|---|---|---|
| `correctness_repo_tests` | `pass`, `executed: 1` | 56 graded cases; `1` is a constant meaning "the process ran" — and exactly what the target produces when it returns early on an unsupported arch |
| `correctness_s1_grid` | `pass` | all three cells were already in the target's own `--shapes` default |
| `execution_receipt` | `pass` | one file for two head runs; the grid run overwrote the repo run |
| `arch_coverage` | `gfx950: runtime` | basis `script-exit-zero-with-output` — the process, not the kernel |
| `perf` | `skip`, "base has nothing to time against" | untrue: the entry point is signature-identical on base |

## The fixes

- **A grid duplicating the target's own defaults is not a control.** Cells are compared against
  the flag's declared default; `grid_independence` reports it and a passing duplicate becomes
  `skip`. A duplicate that *fails* keeps its `fail`.
- **`--axis NAME=FLAG:v1;v2`** reaches knobs the single shape tuple cannot — #4538's head count
  lives on `--num-heads`. Same burden of proof as `--shape-arg`: declared in the target's
  argparse **and** observed refusing an invalid value. Failing either, it is dropped and *named*.
- **A PR adding its own bench target still has a baseline.** The target is transplanted into the
  base tree to time the pre-PR implementation through the same harness, gated on
  `--perf-control-column` reproducing across the two trees. Two bugs blocked this: `scrape_perf`
  matched **0 of 56** rows (unlabeled measurement columns like `flydsl rel` were treated as row
  identity), and a control-rejected stage still published the ratio it had rejected.
- **A `test*` function pytest cannot collect is not a pytest target.** aiter's dominant
  `op_tests` convention is a script whose worker is named `test_<op>(m, d, dtype)`; pytest
  collects it, cannot supply the parameters, and the validator published *"the PR adds this test
  target and it fails on head"* against a working target. `--runner` settles what the classifier
  still gets wrong. Also: a route behind `functools.wraps` was never observed, so aiter's whole
  `@compile_ops` family matched nothing — the probe now resolves routes to code objects.
- **Credential-free execution.** `env VAR=... cmd` *adds* to the inherited environment;
  reproduced with an unmerged PR's target printing `ANTHROPIC_AUTH_TOKEN` and `OPENAI_API_KEY`
  into a stage log. The target now runs under `env -i` plus an allowlist.
- **The exit code came from `--out`**, so a run killed before `finish_report` exited on the
  *previous* run's verdict. It now comes from this run's own `$WORK`.

## Before and after — same head, same base, same host

| | before | identical flags | with the new flags |
|---|---|---|---|
| verdict / exit | **PASS / 0** | **INCONCLUSIVE / 2** | **BLOCK / 1** |
| `correctness_s1_grid` | `pass` | `skip` (not independent) | `fail` at `s_q=2048 s_k=777 nh=16` |
| `observed_work` | — | 56 (repo) / 13 (grid) | same |
| `perf` | `skip`, untrue reason | `skip`, names the missing flag | **`pass`** |

Measured: `flydsl us` median **1.29× faster on head** over 56 matched rows, with the untouched
`triton us` control reproducing within 0.2 %. (`clean_logits=0` only — base allocates the output
with `torch.full` on the other path, so the target's interception never fires there.)

## The merge, in four decisions

- **`scan_index_width.py` — took #5142's.** Both sides independently rewrote D9 as an AST pass.
  Measured on ground truth they are equivalent: on aiter#4978 both report exactly the four lines
  #5132 later fixed, and both report 0 on #5132. They differ only in noise (296 vs 286). With
  detection tied, the tie-breaker is review cost.
- **Grid channel — kept this branch's**, because it has a third channel (`--shape-argnames`) that
  #5142 does not, and `grid_channel_reason` already expresses what `cli-flag-unusable` was for.
- **`run_pytest`'s second argument — took #5142's** (grid value, not `VAR=value`).
- **The perf stage — taken whole from #5142**, which this branch predates.

Merged structurally, not textually: concatenating the test files put #5142's 13 perf tests inside
this branch's `IndexScannerTests`. One regression the merge introduced was caught by this branch's
own tests — dropping `"${shape_plugin[@]}"` silently switched the pytest channel off.

## Held out

| case | expected | observed |
|---|---|---|
| **#4885** HSTU bwd | BLOCK, PR-attributed | **BLOCK**, same cause |
| **#5025** jdbmm bwd | no false blocker | **INCONCLUSIVE**, accurate reason, every finding `note` |
| **#5172** decode TopK | truthful | right verdict, **wrong reason** — fixed |
| **#5081** SiTUv2 quant | truthful | **false blocker** — fixed |
| **FlyDSL mutants** | m0 PASS; m1,m3 correctness; m2 test_policy | all four, expected stage |

Both fresh cases found real defects in this branch, and both are fixed here.

## What this does *not* do

The `H=16` blocker was found by hand first; `--axis` is what let the validator **express** it.
The tool does not decide which configurations are worth trying. What it found on its own is all
evidence-layer: that the grid was a duplicate, that 56 cases were reported as 1, that the perf
claim was measurable after all, that #5081's target never ran under the runner chosen for it.
The kernel-semantic findings on #4538 (strided Q/KV returning wrong results, FNUZ operands
mis-scaled 2×/4×, 28 of 30 variants uncovered) all came from manual review, not from this tool.

## Verification

`pytest .claude/skills/validate-kernel-pr/tests/test_validator.py -q` → **88 passed** (55 from
this branch, 50 from #5142, 18 new; each of the 18 verified to fail against `a8d46dcb0`).
#4538 was rerun on the merged tree and reaches the same conclusion as before the merge.

Known limits: shape capture needs the route's own frame to bind the names, so a
`(*args, **kwargs)` wrapper attests execution but not shapes; a receipt still cannot say which
kernel *variant* ran; `--axis` reaches script targets only; held-out results are comparable only
under exclusive GPU access; gfx942 is unverifiable on a gfx950-only host. Still not a merge gate.


jhinpan:
Rebased onto `dd6ff6a71` — this branch had moved since I forked from #5142's head, so the first push was computed against a stale base and showed as conflicting. Nine commits replayed cleanly; suite still 55 green on the rebased tree.

One commit added on top of the nine: the auto-validation block introduced here forwards `REVIEW_SHAPE_ENV` and `REVIEW_SHAPE_ARG` but not the parametrization channel, so the channel that reaches this repository's dominant test shape was unreachable from the caller that was just wired up. `REVIEW_SHAPE_ARGNAMES` now forwards too.

Final confirmation was taken on a frozen tree (md5 checked before and after the runs), because two round-3 measurements were taken while I was still editing and do not describe a single artifact — those are flagged in the description and should not be quoted. On the frozen tree: #5141's exact false-BLOCK invocation yields `INCONCLUSIVE` with 0 blockers and a reason naming the real refusal, its own 138 tests still passing; #4885 still BLOCKs on the `flydsl.expr.buffer_ops` import.

zufayu:
Dirty again after #5142 moved at 15:38. I've asked there to freeze that branch until this lands.

**Two real defects this PR doesn't touch:**

1. `run_script_with_probe.py:54-70` catches `SystemExit` only. An uncaught `AssertionError` — the standard shape of aiter's `op_tests/*.py` — leaves `status` at `0`, and `finally` writes the receipt before the exception propagates. Receipt says `pytest_exitstatus: 0` on a red run.
2. `validate_pr.sh:760` derives `kernel_files_changed` from `git diff --name-only HEAD`, and the script has no `git add`/`--intent-to-add`. Files `git apply` adds as untracked are invisible, so a new-kernel PR reads as touching no kernel code.

**Two I couldn't settle — already covered?**
- Multi-route `--expected-route`: probe writes `",".join(sorted(...))`; `validate_receipt()` wants `receipt["route"] == expected_route` *and* `expected_route in symbols` (a list). Two routes can never satisfy the second.
- `execution-receipt.json` is `$WORK/$phase/execution-receipt.json` (`:1049-1051`), shared per phase — the grid-probe run overwrites the head-repo receipt.

`cross_namespace` reasoning is on #5142: that line is dead on the current base and this PR is what activates it.

PR #5175, opened 2026-09-01
[Triton/Gluon] fix triton 3.8 regression for rmsnorm, cap blocked BLOCK_SIZE, raise forward occupancy

nidal567:
## Motivation for PR

**TLDR: 
Result: on Triton 3.8 this patch is faster than Triton 3.4 on 79 of the 82 measurable shapes,
median −26.4 %, and closes all three shapes that regressed on the 3.8 upgrade.**

129 shapes measured on Triton 3.4.0 and 3.8.0: benchmark defaults, tester shapes, config-bucket boundaries, every `"rmsnorm"` entry in `model_shapes.json`. Two problems:

- **3 shapes regress when moving 3.4 -> 3.8:**

| shape | 3.4 | 3.8 | |
|---|---|---|---|
| `rms_norm` backward, M=873 N=1245 | 13.347 µs | 23.877 µs | **+78.9 %** |
| `rmsnorm2d_fwd_with_smoothquant`, M=256 N=40960 | 62.623 µs | 75.535 µs | **+20.6 %** |
| `rmsnorm2d_fwd_with_add_smoothquant`, M=256 N=40960 | 83.251 µs | 88.835 µs | **+6.7 %** |

- **Forward kernels launch one workgroup per CU.** `num_programs()` caps the grid at `get_num_sms()`, limiting the resident workgroups available to cover the per-row reductions each forward kernel runs per loop iteration. `rmsnorm2d_fwd_with_smoothquant` M=16384 N=1536 ran at **1.44 TB/s, ~18 % of an 8 TB/s peak**; the plain path was already at a median 71 %. N=1536 is DeepSeek-R1 / Kimi-K2 Thinking.

These are two different compiler stories and are not claimed to be one:

| cause | 3.4 | 3.8 |
|---|---|---|
| blocked-path spilling | present, 243 spills | present, 312 spills |
| forward under-subscription | present | present |
| backward reduce | absent (fix worth −4.0 %) | **the +78.9 %** |

## Technical Details

`rmsnorm` ships no config file (no JSON under `aiter/ops/triton/configs/`), so the wrapper's launch heuristic is its configuration surface. Three knobs.

### 1. Cap `BLOCK_SIZE` at 8192 in the blocked path

Blocked holds `BLOCK_SIZE` elements per program live as fp32 for `x`, again for `g`, and with smoothquant a third time for `x_scale`. At 65536 bytes and 4 warps: 32768 / 256 lanes = **128 VGPR/lane per array, ~384 for the three** -- before reduction state, addresses and temporaries. Both compilers hit the 512-VGPR file and spill.

| `_quant_rms_norm_kernel`, M=256 N=40960 bf16, smoothquant | VGPR | spills |
|---|---|---|
| 3.4, `BLOCK_SIZE=32768` | 512 (max) | **243** |
| 3.8, `BLOCK_SIZE=32768` | 512 (max) | **312** |
| **8192 (this PR)** | 214 | **0** |

8192 over the locally-faster 4096: at N=65536 that is 16 loop iterations instead of 8 -- 666.150 µs vs 498.429 µs same-session, **+33.6 %**. Enumerating every (N, element size) pair leaves the blocked/single-pass selection unchanged, so single-pass shapes and `dg_tmp_rows()` are not affected.

### 2. `num_warps=8` on `_rmsnorm_bwd_dg_reduce_triton`

The reduce launches exactly `cdiv(N, BLOCK_SIZE_N=64)` workgroups -- at the regressing shape, **20 workgroups on 256 CUs**. `num_warps` is the only knob that adds parallelism. Both values measured on 3.8, 5 reps, median, dg-reduce component isolated; workgroup counts read from the dispatch records:

| N | workgroups | CU occupancy | `N % 64` | nw4 | nw8 | delta |
|---|---|---|---|---|---|---|
| 1245 | 20 | 8 % | 29 | 16.417 | 4.452 | **−72.9 %** |
| 1536 | 24 | 9 % | **0** | 22.312 | 8.255 | **−63.0 %** |
| 4096, M=128 | 64 | 25 % | 0 | 3.950 | 2.750 | −30.4 % |
| 4096, M=1 | 64 | 25 % | 0 | 2.960 | 2.654 | −10.3 % |
| 4096, M=4096 | 64 | 25 % | 0 | 3.959 | 3.586 | −9.4 % |
| 8192 | 128 | 50 % | 0 | 3.942 | 3.958 | +0.4 % |
| 40960 | 640 | 100 % | 0 | 7.027 | 7.299 | +3.9 % |

**Not gated on the masked tail.** `num_warps = 8 if n_cols % BLOCK_SIZE_N else 4` was tested and is wrong: N=1536 is exactly aligned and still wants 8 warps by **−63.0 %**. `delta` tracks CU occupancy, not alignment. The only shape preferring 4 warps is already at 100 % occupancy, by +3.9 % on a component that is 24 % of that op -- **+0.9 % end to end**, inside the floor. A grid-based gate (`cdiv(N,64) < get_num_sms()`) recovers that 0.9% and is a one-line change

### 3. `num_programs_fwd()`: forward grid 4x, gated on `BLOCK_SIZE <= 8192`

Gated on `BLOCK_SIZE` because that is the per-program register footprint; the win exists only while several workgroups stay resident per CU. Measured directly, oversubscribe 4x vs off, median of 3 interleaved reps. Only shapes where the config actually changes are counted -- for M <= 256, `min(M, 256) == min(M, 1024)` and the two arms are byte-identical:

| selected `BLOCK_SIZE` | shapes | median | worst |
|---|---|---|---|
| 128 | 2 | **−23.7 %** | −0.0 % |
| 512 | 2 | **−23.9 %** | +3.2 % |
| 1024 | 1 | +1.4 % | +1.4 % |
| 2048 | 5 | **−41.7 %** | −29.6 % |
| 4096 | 5 | **−7.4 %** | −2.4 % |
| 8192 | 5 | −1.5 % | +0.3 % |
| **16384** | 3 | **+1.6 %** | **+12.4 %** |

16384 is the only group with a positive median, hence the gate. No non-control shape at BLOCK_SIZE=32768 exists in the manifest, so that end of the gate is untested. `num_warps` and `waves_per_eu` were screened and rejected (`num_warps=8` costs M=16384 N=8192 +10.4 %). **Not applied to backward**: `dg_tmp_rows()` ties the backward scratch buffer and its reduction row count to `num_programs`, so raising it there costs +4.9 % / +12.4 %. `large_m_small_n` computes its own grid and calls neither helper.

**The heavy add+smoothquant epilogue does pay a real cost, and excluding it is still worse.** The four `fwd_add_sq` shapes whose config actually changes, oversubscribe off vs 4x, 5 interleaved reps, median (the other three have M <= 256 and are byte-identical between arms):

| shape | off | 4x | delta | |
|---|---|---|---|---|
| M=16384 N=1536 | 77.891 | 44.840 | **−42.4 %** | |
| M=4096 N=4096 | 38.931 | 36.358 | −6.6 % | |
| M=873 N=1245 | 10.394 | 9.970 | −4.1 % | |
| **M=4096 N=8192** | **68.938** | **74.633** | **+8.3 %** | the one real cost |

M=4096 N=8192 genuinely loses 8.3 % to the wider grid -- it is the heaviest epilogue at the largest `BLOCK_SIZE` the gate still admits. It is kept because excluding `fwd_add_sq` from oversubscription costs **+18.3 %** across the entry point's real shapes, and the shape is still faster than 3.4 (−1.7 %). `_quant_fused_add_rmsnorm_kernel` compiles identically in both arms (186 VGPR, 0 spills), so this is the 4x grid's L2 working set, not codegen.

## Test Plan

* `pytest op_tests/triton_tests/normalization/test_rmsnorm.py`.
* All 129 shapes vs the tester's own `run_torch` reference at its own tolerances (bf16/fp16 1e-2, fp32 1e-4; int8 `atol=1, rtol=0`, scales 1e-3).
* `rocprofv3 --kernel-trace`. Arms measured back-to-back per shape with the order rotated per repetition, so session drift hits them equally. **200 retained dispatches** per run after warmup and 10 % chronological trimming at each end, grabbing means of total reps then median number. Backward sums its two kernels.
* **Negative control** = a shape whose launch configuration is byte-identical between two 3.8 arms, so any measured difference is pure run-to-run noise. 53 of 129 qualify; their |delta| sets the resolution floor: median 0.8 %, **p90 +/-8.9 % (median) / +/-4.9 % (min)**.
* Regression rule, 3.4 as reference: `regression` iff `3.8 > 1.005 · 3.4 + 0.5 µs`.
* `black --check`, `ruff check` (0.16.0, the CI pin), `py_compile`.

## Test Results

**Triton 3.8 with this patch vs Triton 3.4**, median, 82 shapes where 3.4 ≥ 4 µs:

| entry point | shapes | median vs 3.4 | faster |
|---|---|---|---|
| `rmsnorm2d_fwd_with_dynamicquant` | 5 | **−34.3 %** | 5/5 |
| `rms_norm` (forward) | 49 | **−28.5 %** | 46/49 |
| `rmsnorm2d_fwd_with_add_dynamicquant` | 5 | **−28.3 %** | 5/5 |
| `rmsnorm2d_fwd_with_add` | 5 | **−23.6 %** | 5/5 |
| `rmsnorm2d_fwd_with_smoothquant` | 5 | **−19.7 %** | 5/5 |
| `rms_norm` backward | 7 | **−19.2 %** | 7/7 |
| `rmsnorm2d_fwd_with_add_smoothquant` | 6 | **−16.3 %** | 6/6 |
| **all** | **82** | **−26.4 %** | **79/82** |

Over all 129 including sub-4 µs shapes: median −17.8 %, faster on 124/129.

**Reproducible regressions vs 3.4: 3 -> 0.** All three base regressions are fixed under median, min and mean alike. One shape flags under median only -- `rms_norm` M=16384 N=2048 at +3.1 %, which reverses to -0.5 % under min, sits inside the +/-4.9 % control floor, and is +0.8 % against untouched 3.8, so the patch does not move it. The untouched-3.8 column is the problem being fixed; flagged counts exceed real ones because negative controls trip the threshold on noise alone.

| aggregation | untouched 3.8 flagged | of which controls | **real** | this PR flagged | **real** |
|---|---|---|---|---|---|
| median | 5 | 2 | **3** | 2 | **1** |
| min | 4 | 1 | **3** | 0 | **0** |
| mean | 5 | 2 | **3** | 0 | **0** |

The three, all now faster than 3.4:

```
backward     M=873  N=1245    13.347 -> 11.000   −17.6%   (was +78.9% on 3.8)
smoothquant  M=256  N=40960   62.623 -> 30.001   −52.1%   (was +20.6%)
add+smooth   M=256  N=40960   83.251 -> 37.336   −55.2%   (was  +6.7%)
```

Five shapes read slower than 3.4. Three are negative controls -- including `rms_norm` M=256 N=16384 **fp32**, which read 8.061 vs 6.532 µs between byte-identical configurations and ships nowhere (the benchmark is bf16-only, `model_shapes.json` carries no dtype, the tester has no (256, 16384)) -- and one is sub-4 µs. The only measurable one is M=16384 N=2048, above.

The one shape the patch genuinely costs is not in that count because it still beats 3.4: `rmsnorm2d_fwd_with_add_smoothquant` M=4096 N=8192, +6.6 % vs untouched 3.8 and −1.7 % vs 3.4
(Technical Details §3).

**Triton 3.4.** Each knob measured on 3.4 as well, counting only shapes where that knob applies:

| knob | shape | 3.4 stock | 3.4 patched | delta |
|---|---|---|---|---|
| blocked cap 8192 | `fwd_add_sq` M=256 N=40960 | 83.120 | 37.988 | **−54.3 %** |
| blocked cap 8192 | `fwd_sq` M=256 N=40960 | 62.920 | 30.980 | **−50.8 %** |
| forward 4x | `fwd_sq` M=16384 N=1536 | 67.611 | 31.729 | **−53.1 %** |
| forward 4x | `rms_norm` M=4096 N=1280 | 15.206 | 6.803 | **−55.3 %** |
| forward 4x | `rms_norm` M=8192 N=65536 | 581.159 | 529.475 | −8.9 % |
| dg `num_warps=8` | backward M=873 N=1245 | 13.095 | 12.576 | −4.0 % |

The blocked cap and the forward grid are op-level wins on both compilers. The reduce change is worth −4.0 % on 3.4 and −72.9 % on 3.8 -- that one is the 3.8-specific part.

#### Correctness:
`380 passed, 369 warnings in 64.84s (0:01:04)`

#### Architecture coverage: 
All measurements are gfx950. Gating by architecture instead of landing the constants globally is a reasonable request and easy to add **should it be required in a future PR**.

#### Upstream sync 
This branch is rebased on a `main` that changed the rmsnorm kernel file after the measurements were taken: config-aware `repr`s were added to the normalization kernels, and a new large-M/small-N backward specialization landed. The specialization is selected only when `M > 8192 and N <= 1024`; **no shape in this 129-shape suite selects it**, so the benchmarked execution paths and every performance conclusion are unchanged. That specialization has its own dg-reduce launch site, which this PR does not modify. It postdates these measurements and no measured shape exercises it, so gating it here would not be evidence-backed.

## Submission Checklist


aiter/ops/triton/normalization/rmsnorm.py
```diff
@@ -24,8 +24,17 @@ def num_programs(x):
     return min(x.shape[0], get_num_sms())
 
 
+# Cap blocked path at 8192 (fastest) to avoid severe VGPR spilling.
+blocked_max_block_size = 8192
```
brunomazzottiamd, line 28: `blocked_max_block_size` is used only in scope of `block_size()` function. Please make it a local variable instead of a global variable.

aiter/ops/triton/normalization/rmsnorm.py
```diff
@@ -24,8 +24,17 @@ def num_programs(x):
...
+
+
 def block_size(x):
-    return min(65536 // x.element_size(), triton.next_power_of_2(x.shape[1]))
```
brunomazzottiamd, line 28: Please keep the old block size selection logic alive for architectures other than `gfx950`. You have retuned RMSNorm for `gfx950`, the outcome for other architectures is unknow, it's better to keep this aspect unchanged. As far as I can remember, a lot of tuning was done for `gfx942` - it can be irrelevant today, due to compiler evolution, the correct approach would be to retune for `gfx942` instead of assuming that best `gfx950` parameters would fit.

I'm thinking about something like this:

```python
if arch == "gfx950":
    # new block size logic
else:
    # old block size logic
    # TODO: retune to account for updated compiler
```

aiter/ops/triton/normalization/rmsnorm.py
```diff
@@ -36,6 +45,18 @@ def dg_tmp_rows(x):
     return x.shape[0] if use_blocked(x) else num_programs(x)
 
 
+# num_programs() caps the grid at one workgroup per CU, limiting resident workgroups
+# available to hide per-row reduction latency. Oversubscribe 4x, except large blocks.
+fwd_program_oversub = 4
```
brunomazzottiamd, line 50: `fwd_program_oversub` and `fwd_oversub_max_block_size ` are used only in scope of `num_programs_fwd()` function. Please make them local variables instead of a global variables.

aiter/ops/triton/normalization/rmsnorm.py
```diff
@@ -36,6 +45,18 @@ def dg_tmp_rows(x):
...
+fwd_oversub_max_block_size = 8192
+
+
+def num_programs_fwd(x):
```
brunomazzottiamd, line 54: Please keep the old number of programs selection logic alive for architectures other than `gfx950`. You have retuned RMSNorm for `gfx950`, the outcome for other architectures is unknow, it's better to keep this aspect unchanged. As far as I can remember, a lot of tuning was done for `gfx942` - it can be irrelevant today, due to compiler evolution, the correct approach would be to retune for `gfx942` instead of assuming that best `gfx950` parameters would fit.

I'm thinking about something like this:

```python
if arch == "gfx950":
    # new num. programs logic
else:
    # old num. programs logic
    # TODO: retune to account for updated compiler
    return num_programs(...)
```

brunomazzottiamd: Add a comment to `num_programs()` function explaining that it was previously used for forward and backward operators, and now it's just for backward.

Add a comment to `num_programs_fwd()` function explaining that it has replaced `num_programs()` for forward operators in `gfx950`, and it fallbacks to the old behavior (i.e. `num_programs()`) for other architectures.

aiter/ops/triton/normalization/rmsnorm.py
```diff
@@ -152,6 +173,9 @@ def _rmsnorm_backward(dz, x, gamma, rsigma):
             dg_tmp.shape[1],
             BLOCK_SIZE_M=128,
             BLOCK_SIZE_N=64,
+            # 8 warps, not 4: this reduce launches only cdiv(N, 64) workgroups,
+            # so small N starves the CUs (N=1245: 20 WGs on 256, 16.4 -> 4.5us).
+            num_warps=8,
```
brunomazzottiamd, line 178: Pick `num_warps=8` just for `gfx950`, leave it unchanged for other architectures. Do not hardcode the compiler default here since it may change in the future. Do not pass `num_warps` if we're in a GPU other than `gfx950`, pass `num_warps=8` otherwise.

The comment is also specific to `gfx950` since it mentions 256 CUs. I won't add kernel timings to comments, performance can change a lot due to compiler changes...

brunomazzottiamd (changes requested):

brunomazzottiamd (approved):
LGTM! Let's wait for CI to pass before merging.

aiter/ops/triton/normalization/rmsnorm.py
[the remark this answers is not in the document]
nidal567: get_normalization_config() does exist and is already used by RMSNorm, but there's no point moving these values into it in this PR. Adding a new RMSNorm config family for these values would require unmeasured configs for the other architectures and/or introduce fallback logic. The main goal here is to retune gfx950-only as arch-gated fix and preserve old behavior otherwise

aiter/ops/triton/normalization/rmsnorm.py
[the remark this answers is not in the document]
nidal567: This would broaden the change if included and conflicts with the current requirement to pass num_warps=8 only on gfx950 while leaving other architectures at the compiler default. Therefore, no need to introduce this into this PR.

brunomazzottiamd (approved):
LGTM!

The following failures in **Triton Tests (MI300X) / Shard 5** are unrelated to this PR:

```text
=========================== short test summary info ============================
FAILED op_tests/triton_tests/chunk_delta_attn/test_flash_kda.py::test_tuner_keeps_the_two_schedules_apart
FAILED op_tests/triton_tests/chunk_delta_attn/test_flash_kda.py::test_published_k2_schedules_can_split_their_tile
=== 2 failed, 3006 passed, 2931 skipped, 1317 warnings in 1187.16s (0:19:47) ===
```

I'll address this in the scope of another PR.

PR #5177, opened 2026-09-01
[Triton/Gluon] Add FP8 block-wise quantization kernels

WuLei-AMD:
## Summary

  Adds three Triton kernel families for FP8 quantization on AMD GPUs, each with a
  public wrapper under `aiter/ops/triton/quant/`:

  - **MXFP8 convert** (`convert_to_mxfp8` / `convert_from_mxfp8`): block-scaled
    (e8m0) MXFP8 quant/dequant with optional stochastic rounding and a gfx950 ASM
    fast path. Supports 1-D and 2-D block scaling and e4m3fn / e5m2 outputs.
    `requant_fp8_row_to_col`, `_segment_m`): per-block FP8 quant for activations
    and weights, a fused dual row+col pass for activation gradients, a row→col
    requant that skips the BF16 roundtrip, and a segmented (grouped/MoE) variant.
  - **fast_transpose** (`fast_transpose_2d`): tiled LDS-based 2-D transpose that
    replaces the `t().contiguous()` full `aten::copy_`; works for any dtype
    including FP8 (e4m3 / e5m2 / fnuz).
  
  ## Why

  These are the kernel primitives the blockwise-FP8 weight-gradient path and MXFP8
  activation quant need. Keeping them in AITER lets the dispatch layer call them
  through the standard wrapper path instead of reaching into private kernels.
  
  ## Correctness notes
  
  - Pointer arithmetic is widened to int64 where a flattened index can exceed 2^31
    — batched weight offset `bid * M * N` and segmented `m_range_start * N` — matching
    the other blockwise kernels (a stacked MoE weight past 2^31 elements would
    otherwise wrap silently).
  - MXFP8 convert asserts M/N tile alignment (the kernel loads full tiles without
    masking) and auto-selects the ASM path only on gfx950 — the `v_cvt_scalef32_*`
    instructions are gfx950-only, so other archs use the portable path.
  
  ## Test plan
  
  - [ ] `op_tests/triton_tests/quant/test_quant_fp8_blockwise.py` — blockwise /
        weight (batched, exercises int64 offset) / act_grad / requant / segment_m;
        dequant-vs-input within FP8 tolerance
  - [ ] `op_tests/triton_tests/quant/test_convert_mxfp8.py` — MXFP8 roundtrip
        (e4m3fn / e5m2 × 1-D/2-D block × fp32/bf16) + alignment guard
  - [ ] `op_tests/triton_tests/quant/test_fast_transpose.py` — transpose across
        dtypes/shapes incl. non-power-of-2 and FP8

  Verified on MI308X (gfx942); the MXFP8 ASM path requires gfx950.


vgokhale:
1) Need repr decorators for all kernels
2) Add benchmarks

vgokhale:
This PR has a lot of duplicates. 

quant_fp8_blockwise_for_weight_kernel
quant_fp8_blockwise_for_act_grad_kernel
quant_fp8_blockwise_segment_m_kernel
requant_fp8_row_to_col_kernel

can be merged into quant_fp8_blockwise_kernel. We don't need separate kernels. And quant_fp8_blockwise_kernel  can be merged into _dynamic_per_token_quant_fp8_i8_kernel . Similarly  for the other quant kernels.

_transpose_2d_kernel _generate_randval  and _compute_m_range  are the only ones that really need to be new. Can you please make separate PRs for chaging existing kernels vs adding new ones?

WuLei-AMD:
> This PR has a lot of duplicates.
> 
> quant_fp8_blockwise_for_weight_kernel quant_fp8_blockwise_for_act_grad_kernel quant_fp8_blockwise_segment_m_kernel requant_fp8_row_to_col_kernel
> 
> can be merged into quant_fp8_blockwise_kernel. We don't need separate kernels. And quant_fp8_blockwise_kernel can be merged into _dynamic_per_token_quant_fp8_i8_kernel . Similarly for the other quant kernels.
> 
> _transpose_2d_kernel _generate_randval and _compute_m_range are the only ones that really need to be new. Can you please make separate PRs for chaging existing kernels vs adding new ones?



Thanks — looked into the blockwise ↔ per-token merge. _dynamic_per_token_quant_fp8_i8_kernel produces one scale per row ([M], 1-D whole-row load); blockwise produces one scale per BLOCK group ([M, N/BLOCK], 2-D tiled). Per-token is the
  special case of blockwise (BLOCK == N), so it's subsumed, not a peer. Merging them means different optimal tiling/scale-layout and different downstream GEMMs collapsed behind a mode flag — a net loss, and it couples a backbone kernel to a
  new path. The shared abs-max→scale→quantize math is already factored into _compute_scale_and_quant, which seems like the right reuse level.

  Agree on the rest:
  - I'll split the PR into existing-kernel changes vs new kernels (_transpose_2d_kernel, _generate_randval, _compute_m_range).
  - I'll consolidate within the blockwise family where I/O matches, keeping requant/segment_m separate (FP8-in / segmented I/O genuinely differ).

  Can we align on the exact merge boundary before I restructure?

WuLei-AMD:
Another split subPR in https://github.com/ROCm/aiter/pull/5203

vgokhale:
Re-reviewed after the rebase. The three things I flagged last round as rule violations are mostly handled — thanks for turning these around quickly.

**Landed:**
- `repr=` on all four launchable kernels via `make_kernel_repr`, with the device helpers correctly left out. The note about omitting the float `FP8_MAX` from the repr keys so it can't put a `.` in the trace name is a good catch.
- `bench_quant_fp8_blockwise.py` — one provider per kernel, shared `get_caller_name_no_ext`, ms + GB/s. Matches the existing bench pattern.
- `for_act_grad` folded into `quant_fp8_blockwise_kernel` behind `DUAL: tl.constexpr` so both orientations come off one tile load. Five launchable kernels down to four, public API unchanged. Nice.
- Input validation (`_check_block_fp8`, ndim/contiguity asserts) plus three negative tests, and the `quant/__init__.py` export.

Dropping mxfp8 and fast-transpose out of scope also resolves the largest duplication concern for this PR. Where are those landing? The `downcast_to_mxfp` question from the last round still applies to them, so it'd be good to settle it before they come back.

Three things I'd like resolved before merge, then the rest is follow-up.

### Blocking

**1. `_FP8_MAX` disagrees with the tree's existing FP8 bound.**
`quant/quant_fp8_blockwise.py:24` uses `torch.finfo(torch.float8_e4m3fnuz).max` = **240.0**. `quant/fused_fp8_quant.py::get_fp8_min_max_bounds()` deliberately returns **±224.0** for fnuz (matching vLLM). We now have two saturation points for the same dtype in the same package — and `quant/__init__.py` in this PR re-exports `get_fp8_min_max_bounds` alongside these functions. `_check_block_fp8` also validates against 240, so the divergence gets locked into the public contract. Please either reuse the helper or add a comment saying why 240 is correct here and 224 is not.

**2. The int64 offset fix is incomplete within the file.**
`_triton_kernels/quant/quant_fp8_blockwise.py:159` correctly casts (`tl.cast(pid_m, tl.int64) * N`), but the same pattern at **:76**, **:95** and **:268** still multiplies two int32s before the int64 `offs_n`/`offs_k` promotes the sum. Same expression, four sites, one fixed. The threshold is high (M/128 · N > 2^31), so this is consistency rather than a live bug — but it's worth closing since the file already went through an int64 pass.

**3. No launch parameters on any of the five kernel launches.**
`quant/quant_fp8_blockwise.py:79, 139, 187, 234, 291` pass no `num_warps` / `num_stages` / `waves_per_eu`, so every kernel runs at Triton's default 4 warps while loading a `[128, 128]` tile promoted to fp32 — 16384 elements over 256 threads is 64 live fp32 values per lane, which should be spilling. For comparison, `quant/fused_fp8_quant.py` scales `num_warps` from 1 to 16 with block size. Now that the benchmark exists, could you post before/after for a couple of `num_warps` values on the `blockwise` and `weight` providers?

**4. Output dtype is hardcoded to `float8_e4m3fnuz`.** That's the gfx942 flavour; gfx950 wants OCP `float8_e4m3fn`. A `quant_dtype` parameter (defaulting to today's behaviour) would keep this usable as the target arch moves.

**5. `_compute_scale_and_quant` is bypassed by two of the four kernels.** `:200-201` and `:255-256` reimplement the amax → floor → scale → clamp sequence inline instead of calling the helper. Worth routing both through it so the epsilon and clamp semantics can't drift.

**6. The `1e-4` amax floor is absolute** (`:28`, `:201`, `:256`). A block whose true amax is ~1e-6 gets floored to 1e-4 and ends up using ~1% of the FP8 range. That's most likely to bite `for_act_grad`, since small-magnitude gradient blocks are common. A relative floor, or just special-casing amax == 0, would avoid it. No test covers a small-magnitude block today.

**7. Non-finite input.** `amax = Inf` gives `scale = FP8_MAX/Inf = 0` and `x * 0 = NaN`; the `tl.clamp` doesn't catch it.

**8. `_compute_m_range` (`:102`) does an `O(batch_size)` linear scan per program.** A binary search over `scales_seg_indptr` would be cheaper at high expert/segment counts. Also a candidate to move into a shared util — MoE code wants the same lookup.

**9. Minor:** the non-DUAL launch at `:79` passes `x_fp8` / `scales` as placeholders for the unused column pointers. It's safe today because `DUAL` is constexpr and the stores get pruned, but if that guard ever breaks it silently corrupts the primary output. `None` would fault instead.

WuLei-AMD:
> Re-reviewed after the rebase. The three things I flagged last round as rule violations are mostly handled — thanks for turning these around quickly.
> 
> **Landed:**
> 
> * `repr=` on all four launchable kernels via `make_kernel_repr`, with the device helpers correctly left out. The note about omitting the float `FP8_MAX` from the repr keys so it can't put a `.` in the trace name is a good catch.
> * `bench_quant_fp8_blockwise.py` — one provider per kernel, shared `get_caller_name_no_ext`, ms + GB/s. Matches the existing bench pattern.
> * `for_act_grad` folded into `quant_fp8_blockwise_kernel` behind `DUAL: tl.constexpr` so both orientations come off one tile load. Five launchable kernels down to four, public API unchanged. Nice.
> * Input validation (`_check_block_fp8`, ndim/contiguity asserts) plus three negative tests, and the `quant/__init__.py` export.
> 
> Dropping mxfp8 and fast-transpose out of scope also resolves the largest duplication concern for this PR. Where are those landing? The `downcast_to_mxfp` question from the last round still applies to them, so it'd be good to settle it before they come back.
> 
> Three things I'd like resolved before merge, then the rest is follow-up.
> 
> ### Blocking
> **1. `_FP8_MAX` disagrees with the tree's existing FP8 bound.** `quant/quant_fp8_blockwise.py:24` uses `torch.finfo(torch.float8_e4m3fnuz).max` = **240.0**. `quant/fused_fp8_quant.py::get_fp8_min_max_bounds()` deliberately returns **±224.0** for fnuz (matching vLLM). We now have two saturation points for the same dtype in the same package — and `quant/__init__.py` in this PR re-exports `get_fp8_min_max_bounds` alongside these functions. `_check_block_fp8` also validates against 240, so the divergence gets locked into the public contract. Please either reuse the helper or add a comment saying why 240 is correct here and 224 is not.
> 
> **2. The int64 offset fix is incomplete within the file.** `_triton_kernels/quant/quant_fp8_blockwise.py:159` correctly casts (`tl.cast(pid_m, tl.int64) * N`), but the same pattern at **:76**, **:95** and **:268** still multiplies two int32s before the int64 `offs_n`/`offs_k` promotes the sum. Same expression, four sites, one fixed. The threshold is high (M/128 · N > 2^31), so this is consistency rather than a live bug — but it's worth closing since the file already went through an int64 pass.
> 
> **3. No launch parameters on any of the five kernel launches.** `quant/quant_fp8_blockwise.py:79, 139, 187, 234, 291` pass no `num_warps` / `num_stages` / `waves_per_eu`, so every kernel runs at Triton's default 4 warps while loading a `[128, 128]` tile promoted to fp32 — 16384 elements over 256 threads is 64 live fp32 values per lane, which should be spilling. For comparison, `quant/fused_fp8_quant.py` scales `num_warps` from 1 to 16 with block size. Now that the benchmark exists, could you post before/after for a couple of `num_warps` values on the `blockwise` and `weight` providers?
> 
> **4. Output dtype is hardcoded to `float8_e4m3fnuz`.** That's the gfx942 flavour; gfx950 wants OCP `float8_e4m3fn`. A `quant_dtype` parameter (defaulting to today's behaviour) would keep this usable as the target arch moves.
> 
> **5. `_compute_scale_and_quant` is bypassed by two of the four kernels.** `:200-201` and `:255-256` reimplement the amax → floor → scale → clamp sequence inline instead of calling the helper. Worth routing both through it so the epsilon and clamp semantics can't drift.
> 
> **6. The `1e-4` amax floor is absolute** (`:28`, `:201`, `:256`). A block whose true amax is ~1e-6 gets floored to 1e-4 and ends up using ~1% of the FP8 range. That's most likely to bite `for_act_grad`, since small-magnitude gradient blocks are common. A relative floor, or just special-casing amax == 0, would avoid it. No test covers a small-magnitude block today.
> 
> **7. Non-finite input.** `amax = Inf` gives `scale = FP8_MAX/Inf = 0` and `x * 0 = NaN`; the `tl.clamp` doesn't catch it.
> 
> **8. `_compute_m_range` (`:102`) does an `O(batch_size)` linear scan per program.** A binary search over `scales_seg_indptr` would be cheaper at high expert/segment counts. Also a candidate to move into a shared util — MoE code wants the same lookup.
> 
> **9. Minor:** the non-DUAL launch at `:79` passes `x_fp8` / `scales` as placeholders for the unused column pointers. It's safe today because `DUAL` is constexpr and the stores get pruned, but if that guard ever breaks it silently corrupts the primary output. `None` would fault instead.

 Thanks for the thorough review. All 9 items addressed:
  
   1. _FP8_MAX 240 vs 224
   Added a comment explaining the divergence: fused_fp8_quant returns 224 to match the vLLM per-tensor/per-token convention; blockwise uses per-block scales so the full hardware maximum (240 = torch.finfo(float8_e4m3fnuz).max) is appropriate and we can use it without the conservative headroom.
  
   2. int64 incomplete
   Fixed the two remaining pid_m * N sites (AXIS=0 scale store and DUAL col-scale store) with tl.cast(pid_m, tl.int64).
  
   3. No launch parameters
   Added _launch_params(block_size) helper → num_warps = min(16, max(1, block_size² // 4096)) + waves_per_eu=2, num_stages=2. Benchmarked on MI308X (bf16, block_size=128):
  
   | Shape     | kernel    | nw=2  | nw=4  | nw=8  | nw=16 |
   |-----------|-----------|-------|-------|-------|-------|
   | 4096×7168 | blockwise | 0.080 | 0.072 | 0.077 | 0.070 |
   | 4096×7168 | weight    | 0.055 | 0.056 | 0.060 | 0.064 |
   | 4096×8192 | blockwise | 0.082 | 0.085 | 0.088 | 0.084 |
   | 4096×8192 | weight    | 0.070 | 0.072 | 0.078 | 0.083 |
   | 8192×8192 | blockwise | 0.160 | 0.164 | 0.174 | 0.163 |
   | 8192×8192 | weight    | 0.136 | 0.135 | 0.149 | 0.156 |
  
  num_warps=4 is the best stable choice (weight kernel degrades monotonically above 4).
  
   4. quant_dtype parameter
   Added quant_dtype: torch.dtype = torch.float8_e4m3fnuz to all 5 wrappers. Output tensors use quant_dtype; _check_block_fp8 now validates fp8_max against torch.finfo(quant_dtype).max so gfx950 / OCP float8_e4m3fn works without any further changes.
  
   5. _compute_scale_and_quant bypass
   requant_fp8_row_to_col_kernel now routes through the helper (axis=0). quant_fp8_blockwise_for_weight_kernel keeps its scalar (global) amax — _compute_scale_and_quant requires a 1-D axis — but the inline code is updated to match the helper's Inf clamp and floor  semantics exactly, with a comment pointing to the helper.
  
   6 & 7. amax floor and non-finite input
   Updated _compute_scale_and_quant: tl.minimum(amax, FP8_MAX) before dividing (Inf → FP8_MAX → saturation via clamp, not NaN); floor changed from 1e-4 to 1e-30 (prevents division by zero on all-zero blocks without destroying the dynamic range of small-magnitude blocks).
  
   8. _compute_m_range O(batch_size) scan
   Replaced with a while-loop binary search (O(log batch_size)). Also extracted to _triton_kernels/common/segment_tile.py as _find_segment_tile_range so future MoE segment-dispatch kernels can import it directly.
  
   9. non-DUAL placeholder
   Expanded the comment to make the constexpr dead-code elimination guarantee explicit — Triton prunes the col stores before codegen when DUAL=False, so the aliased pointers are never dereferenced.

Dewei-Wang-sh (approved):

PR #5137, opened 2026-08-31
[Triton/Gluon] gemm A16w16 tune for disable amdgpu-trackers

yanxuer-999:
## Motivation

Restore two triton 3.8 performance regressions on gfx950 by cherry-picking validated tuning fixes: paged-attention GQA decode (llama3, ~-18%) and GEMM-A16W16 at N=1280/K=8192 (tracker-off VGPR spill).

## Technical Details

pa_decode.py: set waves_per_eu=2 on the two v1 w_dot decode kernels (non-quant + per-token FP8), so regalloc gets the full 256-VGPR budget instead of triton 3.8's inflated 202-VGPR/occupancy-2 state (s_waitcnt lgkmcnt(0) 72→50, s_nop 36→9); v2 w_dot is intentionally left unchanged because waves_per_eu=2 there spills and regresses.

GEMM-A16W16-N=1280-K=8192.json: new gfx950 config lowering waves_per_eu (M_LEQ_32 8→4, M_LEQ_256 6→4) to remove the tracker-off VGPR spill on this shape.

## Test Result

llama3 pa_decode 8B/70B/405B recovered +43/+43/+42% (deepseek-V3/wo_dot unaffected at 0%); GEMM-A16W16 N=1280 problem M sizes recovered +242/+151/+157% (M=32/192/256) with no regression on untuned shapes (<2% spread).

PR #5181, opened 2026-09-01
Revert "[Triton/Gluon] gemm A16w16 tune for disable amdgpu-trackers"

Boss2002n:
Reverts ROCm/aiter#5137

Please use config jsons to change params like this


yanxuer-999:
Hi @Boss2002n, thanks for the review and for flagging this.

Can we merge a new [pr5200](https://github.com/ROCm/aiter/pull/5200) fixes the hardcode issue instead of reverting #5137? PR5137 bundles two independent changes, and only one of them is the hardcoding you called out. To address the hardcoding properly, I've opened a new PR that moves the `pa_decode` tuning out of the source and into the config tree
Pr 5200 Verified on MI350X (gfx950): config is picked up (`arch=gfx950`), the compiled kernel carries `"amdgpu-waves-per-eu"="2,2"` only when the config is present, and `test_pa_decode.py` GQA cases pass (120/120). cc @Dewei-Wang-sh

PR #5183, opened 2026-09-01
[Config] [Perf] Add tuned bf16 GEMM configs for Qwen3.8-2.4T-A95B on gfx950

mjkvaak-amd:
### What this adds

One tuned table, `aiter/configs/model_configs/qwen3_8_2p4t_a95b_bf16_tuned_gemm.csv`, with **75 rows** covering the seven bf16 GEMM families Qwen3.8-2.4T-A95B executes that AITER has no tuned entry for. Tuned on **gfx950, cu_num=256** (MI355X), TP8.

The matching `qwen3_8_2p4t_a95b_bf16_untuned_gemm.csv` shape list is included so the table can be re-tuned without  reconstructing the shapes by hand. The `get_config_file` glob skips any name containing `untuned`, so it is documentation and does not affect lookup.

| N | K | rows | Module |
| --- | --- | --- | --- |
| 4608 | 8192 | 12 | Gated DeltaNet `in_proj_qkvz`, sharded TP8 |
| 32 | 8192 | 12 | Gated DeltaNet `in_proj_ba`, sharded TP8 |
| 31040 | 8192 | 6 | `lm_head`, vocab 248320 sharded TP8 |
| 512 | 8192 | 6 | MoE router, `M>=512` only (see TTFT below) |
| 1 | 8192 | 15 | shared-expert gate (unfused) |
| 8192 | 256 | 12 | MoE down-projection |
| 513 | 8192 | 12 | fused shared-expert gate, 512 routed + 1 shared (for FSE quantized model)|

### Why these shapes

They are the shapes real serving runs logged as misses:

```
[aiter] shape is M:64, N:4608, K:8192 dtype='torch.bfloat16' ... not found tuned
        config in /tmp/aiter_configs/bf16_tuned_gemm.csv, will use default config!
```

Two sweeps on two checkpoints contributed: an 18-case MoE-MXFP4 sweep (1,618 distinct shapes) and an 18-case Quark-MXFP4 sweep (custom, non-public checkpoint; 461 distinct shapes). Every miss in both falls into the seven families above. 65 of the model's 92 layers are `linear_attention`, so the two GDN projections run on nearly every forward pass.

### Why 75 rows covers thousands of shapes

The lookup tries `gl in [None, 0, 1]`: exact M, then the two padded buckets from `get_padded_m`. Tuning the power-of-two bucket representatives serves every M that pads onto them.

Verified against a full production-shaped benchmark, not just the tuning sweep. A run covering ISL/OSL of 8k/1k, 60k/1k and 128k/1k logged **6,127 distinct (M, N, K)** shapes, all in one key variant (`torch.bfloat16` in and out, `bias=False`, `scaleAB=False`, `bpreshuffle=False`). Replaying AITER's lookup against every one of them:

**6,070 of 6,127 served**, by 59 rows. No family outside the seven above appeared, and the thousand-plus distinct M values per family collapse onto a dozen rows each:

| N | K | distinct observed M | rows |
| --- | --- | --- | --- |
| 1 | 8192 | 1,185 | 15 |
| 4608 | 8192 | 1,182 | 12 |
| 8192 | 256 | 1,182 | 12 |
| 512 | 8192 | 1,179 | 6 (57 shapes below M=512 left to the fallback, deliberately) |
| 32 | 8192 | 1,167 | 12 |
| 31040 | 8192 | 232 | 6 |

The 57 unserved shapes are the one deliberate gap: MoE-router GEMMs below M=512, excluded because tuned entries there measurably hurt TTFT. They keep AITER's current behavior, so nothing regresses relative to today.

We collected the untuned shapes at `max_num_batched_tokens=16384` and `max_num_seqs=256`. Raising either past those values is a configuration change that would put shapes outside the tuned table.

### Per-shape gain over the current fallback

Best speedup over AITER's `torch`/rocBLAS fallback, per family, measured from the tuner's own profile output and restricted to numerically-clean candidates (below):

| family | shapes beating the fallback | best |
| --- | --- | --- |
| N=1, K=8192 | 15/15 | **3.16x** at M=1 |
| N=32, K=8192 | 12/12 | 1.38x at M=128 |
| N=512, K=8192 | 6/6 | 1.28x at M=1024 |
| N=513, K=8192 | 12/12 | 1.37x at M=8 |
| N=4608, K=8192 | 11/12 | 1.56x at M=256 |
| N=8192, K=256 | 12/12 | 1.49x at M=256 |
| N=31040, K=8192 | 6/6 | 1.14x at M=64 |

### Every winner is numerically exact

**All 75 rows have `err_ratio` 0.0000** against the reference.

This is worth stating because the tuner's default `--errRatio 0.05` does not give you that. Selecting on speed alone, `flydsl` split-k kernels won six N=512 shapes at `err_ratio` 0.015–0.030, while `hipblaslt`, `opus`, `torch` and `triton` all reached 0.000 on the identical shape. Those winners were re-picked as the fastest candidate within `err_ratio <= 0.005`.

### End-to-end effect

Serving A/B under vLLM on the public `amd/Qwen3.8-2.4T-A95B-Quark-MXFP4`, TP8, three repeats per case, arms differing only by this table. 

**Geometric mean output throughput +2.25%**, mean TPOT +2.23%, mean TTFT +1.38%
(all concurrencies clean in both arms):

| Conc | Output tok/s | Mean TTFT | Mean TPOT |
| --- | --- | --- | --- |
| 4 | 251.7 → 251.5 (−0.1%) | 316.0 → 319.6 (−1.1%) | −0.0% |
| 16 | 671.2 → 700.7 (+4.4%) | 830.8 → 835.4 (−0.6%) | +4.4% |
| 32 | 1,049.1 → 1,079.6 (+2.9%) | 1,209.6 → 1,159.9 (+4.1%) | +2.8% |
| 64 | 1,660.0 → 1,696.1 (+2.2%) | 1,695.2 → 1,663.6 (+1.9%) | +2.1% |
| 128 | 2,462.4 → 2,509.4 (+1.9%) | 2,564.7 → 2,501.2 (+2.5%) | +1.8% |

**N=513 gate rows**, measured separately on the fused shared-expert path:
**+0.4%** geometric mean, TPOT −0.3% to −1.3%, directionally consistent across six
cases.

### Accuracy

GSM8K, 5-shot, all 1319 problems, greedy, on `amd/Qwen3.8-2.4T-A95B-Quark-MXFP4`, compared per-problem against the stock arm on the same vLLM build (`v0.27.2rc1.dev150+g311b3513a`) with the same eval settings:

| | baseline | tuned | delta | McNemar p |
| strict-match | 0.9333 | 0.9265 | −0.0068 | 0.391 |
| flexible-extract | 0.9318 | 0.9272 | −0.0045 | 0.594 |

**No detectable accuracy change.**

---

Co-authored-by: Cursor


aiter/configs/model_configs/qwen3_8_2p4t_a95b_bf16_tuned_gemm.csv
```diff
@@ -0,0 +1,76 @@
+gfx,cu_num,M,N,K,bias,dtype,outdtype,scaleAB,bpreshuffle,libtype,solidx,splitK,us,kernelName,err_ratio,tflops,bw
+gfx950,256,1,1,8192,False,torch.bfloat16,torch.bfloat16,False,False,skinny,2,0,3.2932,sol2,0.0,0.0,9.95
+gfx950,256,2,1,8192,False,torch.bfloat16,torch.bfloat16,False,False,skinny,2,0,4.0563,sol2,0.0,0.01,12.12
+gfx950,256,4,1,8192,False,torch.bfloat16,torch.bfloat16,False,False,skinny,2,0,5.6019,sol2,0.0,0.01,14.63
+gfx950,256,8,1,8192,False,torch.bfloat16,torch.bfloat16,False,False,hipblaslt,439633,0,6.4201,Cijk_Alik_Bljk_BBS_BH_Bias_HA_S_SAV_UserArgs_MT16x16x512_MI16x16x1_SN_LDSB0_AFC0_AFEM1_AFEM1_ASEM1_CLR1_CADS0_DTLA1_DTLB1_DTVA0_DTVB0_EPS0_FDSI0_GRPM1_GRVWA8_GRVWB8_GSU0_GSUAMB_GLS0_ISA950_IU1_K1_LDSTI0_LBSPPA1024_LBSPPB1024_LBSPPM0_LPA16_LPB16_LPM0_LRVW8_LWPMn1_MIAV0_MIWT1_1_MO40_NTn1_NTA7_NTB2_NTC2_NTD0_NTM0_NEPBS4_NLCA1_NLCB1_ONLL1_PGR2_PLR1_PKA1_SIA3_SS0_SPO1_SRVW0_SSO0_SVW1_SK3_SKFTR0_SKXCCM0_TLDS1_ULSGRO0_USL1_UIOFGRO0_USFGRO1_VSn1_VWA1_VWB1_WSGRA0_WSGRB0_WS64_WG16_4_4,0.0,0.02,22.97
```
yifehuan, line 5: Please exclude hipblaslt from the checked-in tuned configs. This table contains 63 hipBLASLt entries whose solution indices are library-version dependent and should not be shipped in the shared configs. Please reselect the best candidates from the remaining backends, retain the default fallback where appropriate, and update the performance results.

mjkvaak-amd: @yifehuan The hipBLASLt (and Opus) -free table is 15 rows and results in+0.35% uplift, against the 75-row table's +2.25%. I will close the PR as a dud.

PR #5191, opened 2026-09-01
[HIP] fix(sampling): drop unused out_idx workspace in topk_renorm_from_probs

peizhang56:
## Motivation

`top_k_renorm_probs` writes out of bounds on any sampler batch that mixes a request setting `top_k` with one that does not. This crashes vLLM serving Kimi-K3 on MI355X with `Memory access fault by GPU node-N`, one fault per TP rank. No unusual flags are needed - two ordinary concurrent requests are enough.

At a production vocab the same write usually does *not* fault: the overrun is ~640 KB, which typically lands inside the allocator's 2 MB rounding slack and corrupts silently instead. The write happens on **every** mixed-`top_k` batch; only the crash is probabilistic. An absence of reported crashes is therefore not evidence that the bug is absent.

## Technical Details

The op takes `(probs, maybe_top_k_arr, top_k_val)`. Per the convention this repo's own tests use (`_to_tensor_scalar_tuple` in `op_tests/test_sampling.py`), passing a per-row tensor sends `top_k_val = 0`, so the host wrapper falls back to a hardcoded `max_k`:

```cpp
int max_k = top_k_val;                                            // sampling.cuh:1414
if (max_k <= 0) max_k = 50;                                       // :1415
size_t out_idx_bytes = (size_t)batch_size * max_k * sizeof(int);  // :1426
```

`radix_topk_one_block_kernel` then takes its stride from that host scalar but its bound from the per-row array:

```cpp
const IdxT k_actual = (top_k_arr != nullptr) ? top_k_arr[batch_id] : k;  // :458
if (out_idx) out_idx += batch_id * k;                                   // :474  stride = max_k = 50
...
out_idx[i] = i < row_len ? i + rowStart : -1;   // bounded by k_actual  // :480
```

Any row with `k > 50` overruns its 50-int slot. vLLM encodes "top_k disabled" as `k = vocab_size`, so such a row takes the `row_len <= k_actual` branch at `:477` and writes `vocab_size` ints - 163840 for Kimi-K3, ~640 KB - into a 200-byte slot, running off the ee
nd of the workspace.

The fix is to stop allocating and passing that buffer. `out_idx` has no consumer on this path: the only downstream reader is `TopKRenormWriteOnlyKernel(probs, pivot_buf, norm_buf, renormed_probs, vocab_size)` at `:1450`, which takes no indices. Every `out_idx` store is already null-guarded, and `WRITE_TOPK_VALUES` is `false` here, so the value store nested inside the `if (out_idx)` guard at `:481` is compiled out regardless - no writes are lost. `max_k` still drives `k_actual` correctly on the scalar path, which is unaffected.

Present since #3217.

## Test Plan

Two tests added to `op_tests/test_sampling.py`:

- `test_top_k_renorm_probs_per_row_k` - per-row tensor path, mixing `k=10`, `k=100` and `k=vocab_size`. The existing `test_top_k_renorm_probs` only ever passes a Python `int`, taking the scalar path where `max_k == k_actual` and the indexing is exactly in bounds. Thh
e tensor overload had no coverage at all, which is why CI never caught this.
- `test_top_k_renorm_probs_mixed_batch_does_not_fault` - regression guard for `k = [50, vocab_size]`. Runs in a child process, because the workspace is a per-process `static` allocated on first call, so a process that has already called the op is a spent trial. Usess
 a 1M vocab to put the overrun ~4 MB past the end, clearing the 2 MB slack so it faults deterministically rather than flaking.

Reproduction note: `not_built()` (`csrc/cpp_itfs/utils.py:297`) keys the JIT cache on `lib.so` existing, and the folder name is the md5 of the kwargs - empty for this op - so editing `sampling.cuh` silently reuses a stale binary. Use a fresh `AITER_ROOT_DIR` whee
n comparing before and after.

## Test Result

MI355X (gfx950), ROCm 7.2, torch 2.10:

- Regression guard, unpatched: **5/5** fresh processes abort with `Memory access fault by GPU node-N` at a 2 MB-aligned address. Patched: **6/6** pass.
- At vocab 163840 the unpatched build survived, confirming the silent-corruption case described above.
- Full `op_tests/test_sampling.py` on the patched build: **358 passed, 4 skipped**.

## Submission Checklist

### DIFF
--- a/csrc/cpp_itfs/sampling/sampling.cuh
+++ b/csrc/cpp_itfs/sampling/sampling.cuh
@@ -1423,8 +1423,9 @@ static void topk_renorm_from_probs(
 
     size_t pivot_bytes      = (size_t)batch_size * sizeof(float);
     size_t normalizer_bytes = (size_t)batch_size * sizeof(float);
-    size_t out_idx_bytes    = (size_t)batch_size * max_k * sizeof(int);
-    size_t total_bytes = radix_buf_size + pivot_bytes + normalizer_bytes + out_idx_bytes;
+    // No out_idx: unused here, and unsafe -- the kernel strides it by max_k but
+    // bounds writes by the per-row k, so k > max_k ran past the workspace.
+    size_t total_bytes = radix_buf_size + pivot_bytes + normalizer_bytes;
 
     static void* s_workspace = nullptr;
     static size_t s_workspace_size = 0;
@@ -1437,12 +1438,11 @@ static void topk_renorm_from_probs(
     char* ptr = static_cast<char*>(s_workspace);
     void*  radix_buf  = ptr;                             ptr += radix_buf_size;
     float* pivot_buf  = reinterpret_cast<float*>(ptr);   ptr += pivot_bytes;
-    float* norm_buf   = reinterpret_cast<float*>(ptr);   ptr += normalizer_bytes;
-    int*   out_idx    = reinterpret_cast<int*>(ptr);
+    float* norm_buf   = reinterpret_cast<float*>(ptr);
 
     radix_topk::standalone_stable_radix_10bits<float, int, false>(
         radix_buf, radix_buf_size, probs, batch_size, (int64_t)vocab_size,
-        nullptr, nullptr, max_k, nullptr, out_idx, true, stream, 0,
+        nullptr, nullptr, max_k, nullptr, nullptr, true, stream, 0,
         pivot_buf, norm_buf, top_k_arr);
 
     constexpr int BT = 1024;
--- a/op_tests/test_sampling.py
+++ b/op_tests/test_sampling.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
 
+import subprocess
+import sys
 import warnings
 
 import pytest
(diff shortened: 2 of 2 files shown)


peizhang56:
Reproduce steps:

1. GPU: 8xMI355x
2. Docker: rocm/atom-dev:vllm-kimi-k3-20260807
3. Model: Kimi-K3
4. Server: [k3_server_crash.sh](https://github.com/user-attachments/files/31712677/k3_server_crash.sh)
5. Client:  [repro_topk_crash.sh](https://github.com/user-attachments/files/31712697/repro_topk_crash.sh)

sunway513:
@valarLip @zufayu — Peng asked me to flag this one for review on his behalf.

This PR is currently the blocker for ATOM's `main` CI, and it has been open since Sep 1 with all checks green and no review yet.

**Evidence that this is the same bug ATOM is hitting.** The `gpt-oss-120b (2 GPUs)` accuracy job on ROCm/ATOM `main` fails intermittently with exactly this signature:

```
[t=80s] GPU fault detected (1 signals) - exiting 2
Memory access fault by GPU node-2 (Agent handle: 0x1aa5ad20) on address 0x7ea8bd6ba000.
Reason: Write access to a read-only page.
```

Across three recent runs the node and address differ each time, and the reason alternates between `Unknown` and `Write access to a read-only page`:

- run 33771135354 (`412f5bfe5`) - node-2, `0x7ea8bd6ba000`, write to read-only page
- run 33751437102 (`f95ef3ec3`) - node-3, `0x7f559fa0a000`, Unknown
- run 33738805397 (`e4e473f8c`) - node-2, `0x7e74d0a4d000`, Unknown

That non-determinism matches the description in this PR: the overrun lands in allocator slack most of the time and only sometimes faults. The workload fits too. lm_eval drives the server at `num_concurrent=65`, so mixed-`top_k` batches are close to guaranteed, and gpt-oss-120b has a 201088-token vocab, which is the "production vocab" case where the ~640 KB overrun usually corrupts silently rather than faulting.

**Why the silent-corruption half matters as much as the crash.** On the same job, the runs that do not fault still land below their accuracy gate: 0.8537 and 0.8620 against a 0.87 threshold on Sep 2, where the one clean run that day scored 0.8749. We have not proven those are caused by this write, and there is a second candidate on the ATOM side that we are testing separately, but if the corruption path is live at this vocab size then any accuracy number measured on a mixed-`top_k` batch is suspect.

**On the fix itself.** Dropping `out_idx` rather than growing the workspace looks right: the buffer is sized by the host-side `max_k` (50 on the per-row-tensor path, since `top_k_val = 0`) while the kernel bounds writes by the per-row `k`, so enlarging it would only move the cliff. Removing an argument this function does not use closes the hole instead of widening it. The added per-row cases at k = 100 and k = vocab_size cover exactly the rows that overran, and isolating them in a subprocess is the right call given the workspace is a function-local `static`.

No rush intended on the substance, but ATOM `main` stays red until something lands here, so a review would unblock more than this one repo. Happy to help validate on MI355X if that speeds it up.

sunway513:
Correction to my earlier comment: I attributed ATOM's `gpt-oss-120b (2 GPUs)` failure to this PR, and that attribution was wrong. Apologies for the noise.

What I missed. I matched on the fault signature alone, and `Memory access fault by GPU node-N ... Reason: Unknown / Write access to a read-only page` is generic to any out-of-bounds write, not specific to this bug. I even noted in my own reasoning that this PR should affect the 1-GPU and 2-GPU gpt-oss configs equally, while the data shows the 1-GPU config passing 191 of 192 runs and only the 2-GPU config failing. I did not follow that contradiction up before commenting.

What settled it. I applied this PR's kernel change to an ATOM container on MI355X and reran the failing job. Two things came out of that:

- `top_k_renorm_probs` is compiled on demand through `csrc/cpp_itfs`, and `/root/.aiter/build` contained no `top_k_renorm_probs` artifact before or after the run, only `pa_ps_*`. The op is not on ATOM's gpt-oss path at all, so this patch could not have changed that job's behaviour.
- The run reached 361/1319 with the patch versus 85/1319 without it, but with the op never built, that is run-to-run variance of a probabilistic fault, not an effect of the fix. This PR's own description makes that point: the write happens every time, only the crash is probabilistic.

The actual cause looks like #4827, fixed by #4839. A negative expert id passes an upper-bound-only check in `MoeSortingMultiPhaseKernel_P0_v1` and stores at a negative offset. That explains the asymmetry this PR could not: the 2-GPU config runs `--enable-expert-parallel`, the 1-GPU config does not, and the fault only ever appears with EP on. The last log line before each fault is `[aiter] [fused_moe] ...`, not sampling.

None of this is a finding against this PR. The bug it fixes is real and the fix reads correct to me for the reasons in my earlier comment, which still stand. It simply is not what is holding ATOM's CI red, so please do not treat ATOM as a gate on merging it.

I will verify #4839 on the same box and report there.

peizhang56:
Hi @junhaha666 could you please help to take a look?

PR #5202, opened 2026-09-02
[Bugfix] Make `_fold_seqlen_indptr` cudagraph-safe (avoid scalar H2D copy)

micah-wil:
## Motivation

This PR fixes a bug in `_fold_seqlen_indptr` (introduced in #4964) in which the function is not cudagraph safe as it can trigger a H2D transfer.

## Technical Details

`_fold_seqlen_indptr` does:
```
out = torch.empty(..., device=indptr.device)
out[0] = 0
out[1:] = torch.cumsum(folded_lens, dim=0).to(indptr.dtype)
```

`out[0] = 0` is unsafe during cuda graph capture as it can be lowered to an H2D copy, resulting in `operation not permitted when stream is capturing`.

## Test Plan

This was first exposed in vLLM CI when running the following test on MI300 `HIP_VISIBLE_DEVICES=0 pytest -v -s tests/v1/e2e/spec_decode/eagle/test_eagle_correctness.py::test_eagle_correctness_light[ROCM_AITER_FA-deepseek_eagle]` 
https://buildkite.com/vllm/amd-ci/builds/12434/list?jid=01a05932-e4b1-4cd7-90df-6cefc7be92f9&tab=output#L1820
```
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]   File "/usr/local/lib/python3.12/dist-packages/vllm/_aiter_ops.py", line 2731, in mla_decode_fwd
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     torch.ops.vllm.rocm_aiter_mla_decode_fwd(
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]   File "/usr/local/lib/python3.12/dist-packages/torch/_ops.py", line 1275, in __call__
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     return self._op(*args, **kwargs)
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]            ^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]   File "/usr/local/lib/python3.12/dist-packages/vllm/_aiter_ops.py", line 604, in _rocm_aiter_mla_decode_fwd_impl
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     mla_decode_fwd(
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]   File "/usr/local/lib/python3.12/dist-packages/aiter/mla.py", line 862, in mla_decode_fwd
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     qo_indptr = _fold_seqlen_indptr(qo_indptr, fold_factor)
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]   File "/usr/local/lib/python3.12/dist-packages/aiter/mla.py", line 410, in _fold_seqlen_indptr
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     out[0] = 0
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374]     ~~~^^^
(EngineCore pid=7775) ERROR 08-31 19:15:40 [core.py:1374] RuntimeError: Cannot copy between CPU and CUDA tensors during CUDA graph capture unless the CPU tensor is pinned. Please use tensor.pin_memory() or allocate the tensor with pin_memory=True.
```

## Test Result

Test passes with this PR:
```
============ 1 passed, 14 warnings in 696.39s (0:11:36) ==================
```

### DIFF
--- a/aiter/mla.py
+++ b/aiter/mla.py
@@ -402,13 +402,8 @@ def _fold_seqlen_indptr(indptr, fold_factor):
     """Repeat each batch's seqlen ``fold_factor`` times (head-folding pseudo-batches)."""
     lens = indptr[1:] - indptr[:-1]
     folded_lens = lens.repeat_interleave(fold_factor)
-    out = torch.empty(
-        indptr.shape[0] + (fold_factor - 1) * (indptr.shape[0] - 1),
-        dtype=indptr.dtype,
-        device=indptr.device,
-    )
-    out[0] = 0
-    out[1:] = torch.cumsum(folded_lens, dim=0).to(indptr.dtype)
+    cumsum = torch.cumsum(folded_lens, dim=0).to(indptr.dtype)
+    out = torch.nn.functional.pad(cumsum, (1, 0), value=0)
     return out
 
 
--- a/op_tests/test_mla_persistent.py
+++ b/op_tests/test_mla_persistent.py
@@ -93,6 +93,72 @@ def check_support(dtype, kv_dtype, nhead):
     return not (dtype == dtypes.bf16 and nhead == 32 and get_gfx() == "gfx942")
 
 
+def test_fold_seqlen_indptr_cuda_graph_capture(fold_factor, num_batches=(1, 8, 37)):
+    """Regression test: _fold_seqlen_indptr must be safe to capture into a
+    CUDA graph.
+
+    mla_decode_fwd's nhead-folding branch (nhead in range(32, 128+1, 16),
+    persistent_mode) calls _fold_seqlen_indptr to build the folded qo/kv
+    indptrs. It used to build its output with:
+        out = torch.empty(...)
+        out[0] = 0
+        out[1:] = torch.cumsum(...)
+    `out[0] = 0` assigns a Python scalar into a CUDA tensor, which lowers
+    to an unpinned host->device copy -- illegal while a CUDA graph is
+    being captured. Real callers (e.g. vLLM) capture this exact branch
+    during CUDA-graph warmup/memory-profiling, so this crashed in
+    production despite passing eager-mode-only accuracy tests.
+
+    This captures the helper directly (with the documented side-stream
+    warmup prerequisite for torch.cuda.graph()), replays with DIFFERENT
+    indptr contents than were captured, and checks the replayed result
+    against an eager-mode reference -- following the same pattern as
+    test_moe_sorting_flydsl_cuda_graph_capture (itself a regression test
+    for an analogous .item()/host-copy capture-safety bug).
+    """
+    device = "cuda"
+    dtype = torch.int32  # matches qo_indptr/kv_indptr dtype used in practice
+
+    def make_indptr(batch_size, seed):
+        g = torch.Generator(device=device).manual_seed(seed)
+        seqlens = torch.randint(0, 4096, (batch_size,), device=device, generator=g)
+        indptr = torch.zeros(batch_size + 1, dtype=dtype, device=device)
+        indptr[1:] = torch.cumsum(seqlens, dim=0).to(dtype)
+        return indptr
+
+    for batch_size in num_batches:
+        capture_indptr = make_indptr(batch_size, seed=0)
+
+        # Standard torch.cuda.graph() prerequisite: warm up on a side stream.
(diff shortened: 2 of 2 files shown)


AndreasKaratzas:
Can we ensure any missing CI tests are also added for this issue? they can ofc be vibe coded.

micah-wil:
@AndreasKaratzas good call, done.

zufayu:
The fix is right. I checked `old` vs `new` on 30 combinations (`batch ∈ {0,1,2,8,37,129}` × `fold_factor ∈ {1,2,3,8,16}`) — identical shape, dtype and values, including the `batch=0` edge where `cumsum` is empty. `F.pad` keeps everything on device and the scalar H2D is gone.

**The only hard blocker is one line of formatting.** Black wants the two adjacent string literals joined:

```diff
     aiter.logger.info(
-        "_fold_seqlen_indptr cuda-graph capture/replay: all passed "
-        "(fold_factor=%s)",
+        "_fold_seqlen_indptr cuda-graph capture/replay: all passed " "(fold_factor=%s)",
```

Worth stating the knock-on: `check-signal` reports `Pre-checks did not pass ... FAILED: Check Code Style with Black`, so 27 of 38 checks skipped. **The new test has never executed on a runner.** The red has nothing to do with correctness, but the green hasn't been earned yet either.

**On the test itself — most of it cannot fail.** Measured on gfx942 (ROCm 7.0, torch 2.9.1), capturing the pre-fix implementation:

```
  b=1   ff=3   old-> HIP error: operation not permitted when stream is capturing   new-> ok
  b=1   ff=8   old-> HIP error: operation not permitted when stream is capturing   new-> ok
  b=8   ff=3   old-> HIP error: operation not permitted when stream is capturing   new-> ok
  b=8   ff=8   old-> HIP error: operation not permitted when stream is capturing   new-> ok
  b=37  ff=3   old-> HIP error: operation not permitted when stream is capturing   new-> ok
  b=37  ff=8   old-> HIP error: operation not permitted when stream is capturing   new-> ok
```

Every combination raises the same error at the same place, so `num_batches=(1, 8, 37)` crossed with the two fold factors buys six graph captures where one would prove the same thing. The axes are orthogonal to the defect — `out[0] = 0` is a constant copy, unaffected by batch size or fold factor.

The replay half is dead for the same reason: the exception fires *inside* `with torch.cuda.graph(g):`, so control flow never reaches `graph.replay()` or the `torch.equal(out, ref)` comparison. It was carried over from `test_moe_sorting_flydsl_cuda_graph_capture`, where it earns its place — that bug was `.item()` baking a capture-time count into the graph, so replaying with a different value is exactly the thing under test. Here there is no analogous failure mode.

**Structure.** The call site sits at module top level, while the test it cites as its model keeps everything inside `main()` (`test_moe_sorting.py`: `main()` at 486, `parse_args()` at 588, the capture call at 620, guard at 669). `test_mla_persistent.py` has no `__main__` guard at all — `args = parser.parse_args()` is bare at 2117 — so any `import` of this module now runs argparse against the caller's `sys.argv` and performs six graph captures as a side effect. Pre-existing, but this raises the cost of it.

Minor: `test_fold_seqlen_indptr_cuda_graph_capture` starts with `test_` and takes a required positional. Fine under `python3 <file>`, a fixture error the moment pytest collects it.

Comments are 33% of the added block (20 docstring + 7 `#` against 43 code lines), and the docstring restates the bug history that the PR description already carries, traceback included.

**Suggested shape:** one capture, one shape, pass if it doesn't raise — about ten lines — and move the call inside a function rather than running it at import. The capture-safety check is the right instinct; aiter's op_tests are eager-only, which is why this had to surface in vLLM CI instead of here.

micah-wil:
@zufayu Fixed, thanks!

zufayu (approved):

PR #5219, opened 2026-09-02
[Config] DSv4 TP8 a8w8 blockscale bpreshuffle: add Flash and Pro gfx950 rows

kevin-mii:
## Summary
- add DeepSeek-V4-Flash-0731 TP8 gfx950 A8W8 blockscale bpreshuffle rows: N,K = 512/4096 (19 M), 4096/12288 (19 M), 1536/4096 (3 M), 8192/1024 (3 M) -- the M values not already covered
- add DeepSeek-V4-Pro-0813 TP8 gfx950 rows for N=7168 K=21504, M = 1..16384 (27 rows)

## Why
The Flash TP8 dense shapes and the Pro TP8 shape had no entry at these M values, so every call fell back to the untuned default config (`not found tuned config ... will use default config` on every boot). Tuned with `csrc/ck_gemm_a8w8_blockscale/gemm_a8w8_blockscale_tune.py --preshuffle` on 8x MI350X (gfx950). Rows are appended to the existing `dsv4_*` file, not a new one, since `AITER_CONFIG.get_config_file()` merges every `model_configs/*<name>*.csv` match.

## Validation
- checked CSV schema consistency (13 fields, same header)
- all 71 rows errRatio 0.0; no (M,N,K) key collides with existing rows (keys already present in the file were dropped from the Flash set)
- served end-to-end with SGLang on 8x MI350X TP8: Flash bs=1 decode 256 -> 273 tok/s at landing; Pro: the per-boot fallback warnings for N=7168 K=21504 disappear, TPOT unchanged; GSM8K unchanged on both


yzhou103:
Could you also update the untuned shapes in untuned csv? Thanks.

PR #5221, opened 2026-09-03
Expose the fused_moe activation dtype resolution for dispatch dtype selection

Duyi-Wang:
Exposes the activation-dtype resolution `fused_moe()` already does internally, so
a caller can produce activations in the dtype the kernel consumes instead of one
it has to convert.

`fused_moe()` picks `q_dtype_a` from the quant type, weight dtype, activation,
gate mode, arch and token count, then converts `hidden_states` when they arrive
as something else. That decision was not reachable from outside. This lifts it
into `resolve_activation_dtype()` and calls it from `_fused_moe_impl`.

## Why

SGLang's MoRI EP fixes its dispatch format when the communication buffers are
sized -- before the first token -- and derives it from the expert weight dtype
alone. For MXFP4 weights that says nothing about the kernel: the same weights run
a16w4, a8w4 or a4w4. On DeepSeek-V4-Pro the MoE is a8w4, so the guessed fp4 costs
an `upscale_mxfp4` plus a re-quantize on every layer, in both prefill and decode.
Nothing errors; it is just slower. Driving the format from this resolution takes
gsm8k 200q from **100.0 s to 62-67 s** on MI355X / TP8 + EP8, accuracy unchanged.

## `M` is optional

Such a caller has no `M`: one format serves every batch that follows. Omitting it
returns `None` wherever the answer would have needed M, and the caller falls back
to bf16, which `fused_moe()` accepts from any path. Since M >= 0, a bound of 0
still settles its branch without M.

The distinction matters because the two directions are not symmetric: bf16 where
fp8 was wanted costs a quantize on the receive side, while fp8 where bf16 was
wanted raises `Unsupported kernel config for moe heuristic dispatch` -- or at
small M returns garbage without raising.

## No behavior change

Identical to the inlined resolution on 414720 combinations of quant type, weight
dtype, activation, gate mode, arch, M and the relevant environment variables.


zufayu:
> **Automated advisory review — not a human sign-off, and not a merge gate.**
> Produced by this repo's own `review-pr` skill (`.claude/skills/review-pr`) run against
> this PR. Treat every line below as a lead to check, not as a conclusion.

@yadaish — you maintain `aiter/fused_moe.py`, so flagging this for you.

Two things to weigh it by. The three findings were attacked by the same reviewer that
wrote them and by nobody else, which is what *not independently refuted* on the card
means. And the equivalence check behind the first one was run rather than taken on trust:
base and head transcribed into a differential harness, 186624 combinations, zero
mismatches. Anything marked `[unverified]` was not checked.

For calibration: measured over 250 PRs, roughly **3 in 4** of this skill's findings survive
an independent reader told to assume they are false. The fourth costs you the minutes it
takes to dismiss it. Its false-*clearance* rate — how often it says nothing when something
is wrong — has never been measured, so silence from it means nothing.

---

## ROCm/aiter PR #5221 — Expose the fused_moe activation dtype resolution for dispatch dtype selection

**Lifts the activation-dtype heuristic that `fused_moe()` computed inline into a public `resolve_activation_dtype()`, so a caller that builds the activations itself — SGLang's MoRI EP sizing its all-to-all buffers — can emit the dtype the kernel wants instead of one `fused_moe()` has to convert.**

Review (advisory): ⚠️ NEEDS WORK — not independently refuted
Validation (deterministic): NOT RUN — the PR changes runtime code (`aiter/fused_moe.py`, executed on every MoE call) but ships no test target, so triage had nothing to run
Perf (advisory): NOT RUN — `REVIEW_AUTO_VALIDATE=0` and the PR ships no benchmark entry point; an environment gap, and base-vs-head timing would compare an identical dispatch decision on both sides (see below)

⚠️ The new public `resolve_activation_dtype()` in `aiter/fused_moe.py` has no test, and the one thing about it that is new — `_bound_split()` returning `None` when `M` is omitted — has no in-tree caller at all, because `_fused_moe_impl` always passes `M = topk_ids.shape[0]`. Meanwhile `op_tests/test_moe_2stage.py:969` (`_runtime_situv2_mxfp4_q_dtype_a`) and `:991` (`_runtime_swiglu_mxfp4_q_dtype_a`), whose docstrings say they "Mirror fused_moe's ... routing" and which are live oracles at lines 233, 363, 883 and 893, were left as hand copies rather than pointed at the new single source of truth. Impact: if a later edit makes another branch `M`-dependent without routing it through `_bound_split`, `resolve_activation_dtype(M=None)` silently returns a concrete-but-wrong dtype instead of `None`; MoRI EP then sizes fp8 buffers where bf16 was needed, which this PR's own description says raises `Unsupported kernel config for moe heuristic dispatch` or returns garbage at small M — and the test that should catch it is a copy of the code it is testing. The extraction itself is sound: I transcribed both sides into a differential harness and swept 186624 combinations of quant type x `w1.dtype` x activation x gate mode x M x `hidden_states.dtype` x `a1_scale` x gfx x env with 0 mismatches, and every non-`None` M-omitted answer matched the base answer for every M. [verified] **Author must** add a table test that calls `resolve_activation_dtype()` directly, pinning the `M=None` contract, and switch `op_tests/test_moe_2stage.py:969/:991` to call it instead of re-deriving the routing.

📝 `os.environ.get("AITER_FORCE_A8W4", "0") in ("1")` on the new gfx1250 override in `aiter/fused_moe.py` is a substring test against the string `"1"`, not tuple membership — `("1")` is a string. Concrete trigger: with `AITER_FORCE_A8W4=""` exported (set-but-empty, e.g. `export AITER_FORCE_A8W4=` in a launch script), `"" in "1"` is `True`, so gfx1250 silently forces fp8 activations; `AITER_FORCE_A8W4="11"` is `False`. It is carried verbatim from the deleted block, not introduced here, so this is a note, not a regression — but the diff re-authors the line, and the test mirror at `op_tests/test_moe_2stage.py:977` spells the same check `== "1"`, so implementation and oracle already disagree on that input. [verified] **Author must** change it to `== "1"` while the line is being moved, so the public resolver and its test mirror agree.

📝 The headline number -- gsm8k 200q from 100.0 s to 62-67 s on MI355X TP8+EP8 -- is [unverified]: the claim ships no script, no ROCm/SGLang version, no commit and no checkpoint, and the code that produces the win is in SGLang, not in this diff, so nothing in this repo reproduces it. The equivalence claim's "414720 combinations" is likewise not traceable (`grep -rn 414720` over the merge target returns nothing), though I reproduced the property independently over 186624 combinations. Relatedly, the new `resolve_activation_dtype()` in `aiter/fused_moe.py` is inert inside aiter -- its only in-tree caller, `_fused_moe_impl`, gets exactly the answer it already computed inline -- so 100% of the PR's value depends on an SGLang MoRI EP change, and `pr_meta.json` shows no linked consumer PR and no URL in the body. Impact: the API in `aiter/fused_moe.py` could land, drift and never be exercised, and no reviewer can check the number that justifies it. [verified] **Reviewer should ask** for the MoRI EP PR link and the repro command plus ROCm/SGLang commits behind the 100.0 s to 62-67 s measurement.

coderfeli (approved):

PR #5247, opened 2026-09-03
[Triton/Gluon] [CI] Revert " Add fused KDA decode kernel (conv1d + recurrence + gated RMSNorm)"

gbyu-amd:
Reverts ROCm/aiter#4712


zejunchen-zejun:
close as author will fix the kernel

PR #5259, opened 2026-09-03
[Triton/Gluon] Clean up MoE elementwise kernels

vgokhale:


aiter/ops/triton/_triton_kernels/quant/quant.py
```diff
@@ -19,24 +21,47 @@ def _static_per_tensor_quant_fp8_i8_kernel(
...
     scale = tl.load(scale_in_ptr)
-    scale_recip = 1 / scale
-
-    qx = (x * scale_recip).to(qx_ptr.dtype.element_ty)
-
-    tl.store(qx_ptr + offs, qx, mask=mask)
+    # FAST_CONVERT multiplies by the reciprocal, one v_mul per element, instead of
+    # a correctly-rounded division. Not equivalent: a reciprocal half an fp32 ulp
+    # off can land on the far side of an fp8 rounding boundary.
+    if FAST_CONVERT:
```
lburzawa, line 55: Could we check asm for both converts? IIRC on AMD backend they both result in precision division, not sure if this is Claude optimizing for NV backend. For fast division/convert we could use this function https://github.com/triton-lang/triton/blob/691a6deec4ee15960408f957a973e11528294017/python/triton/language/extra/libdevice.py#L77.

vgokhale: You're right that both lower to the precision division sequence on AMD — same `v_div_scale_f32` / `v_rcp_f32` / `v_div_fmas_f32` / `v_div_fixup_f32`, nothing NV-specific. But the two aren't identical, because of *where* that sequence ends up.

With `x * (1 / scale)` the numerator is a literal `1.0` and `scale` is a uniform scalar loaded once, so LLVM hoists the whole division out of the vectorized body. With `x / scale` the numerator is per-element, so it stays inside and replicates.

VGPR arithmetic emitted at `BLOCK_M=1, BLOCK_N=4096` (what the launcher picks for wide rows = 16 elements/lane, gfx950):

| | `FAST_CONVERT=True` | `FAST_CONVERT=False` |
|---|---|---|
| `v_rcp_f32` | 1 | 16 |
| `v_div_scale_f32` | 2 | 32 |
| `v_fma_f32` / `v_fmac_f32` | 3 / 2 | 48 / 32 |
| `v_div_fmas_f32` | 1 | 16 |
| `v_div_fixup_f32` | 1 | 16 |
| `v_mul_f32` | 1 | 16 |
| `v_pk_mul_f32` | **8** | 0 |
| whole kernel | **92 instrs** | **254 instrs** |

So per element: `True` does half a packed multiply (8 `v_pk_mul_f32` covering 16 values), plus one hoisted division for the whole program. `False` does ~11 unpackable instructions per element. The division-sequence count tracks elements/lane exactly — 1 vs 8 at `BLOCK_N=512`, 1 vs 16 at `BLOCK_N=4096`.

Wall clock on an idle MI355X, `False/True`: 1.00–1.04x where the kernel is bandwidth-saturated (4.4–4.8 TB/s), but **1.21–1.23x** on narrow rows (16384x512, 32768x256) where it isn't.

On `fast_dividef` — it is available (`triton.language.extra.libdevice.fast_dividef` dispatches to HIP) and emits a bare `v_rcp_f32` with no `div_scale`/`div_fmas`/`div_fixup`. But it measures 0.958–1.040x against `FAST_CONVERT=True`, i.e. noise: the reciprocal form is already one multiply per element, so all `fast_dividef` saves is that single hoisted sequence amortized over 4096 elements. It does cost accuracy — 142 vs 64 differing fp8 e4m3 elements against exact division over 8,744,960 elements x 305 scales. Worth using where the divisor is genuinely per-element (the smoothquant kernels divide by a per-row scale), not for a uniform scalar.

Given all that, I think your first instinct lands, just in the other direction: rather than keep the flag, drop it and always use the reciprocal form. That removes the knob *and* takes the faster path everywhere. The only thing it costs is that `downcast_to_static_fp8` stops being bit-identical to the kernel it replaced — 64 elements in 8.7M shift by one fp8 ulp. Happy to make that change if you agree.

vgokhale: Here is the actual gfx950 asm for that region, unedited apart from stripping comment columns.

**At 1 element/lane** (`BLOCK_M=1, BLOCK_N=256`) they are the same sequence, and this is the case that matches your recollection:

```
--- FAST_CONVERT=True (14 instrs) ---        --- FAST_CONVERT=False (13 instrs) ---
v_div_scale_f32 v4, s[2:3], v0, v0, 1.0      v_div_scale_f32 v2, s[0:1], v5, v5, v0
v_rcp_f32_e32   v5, v4                       v_rcp_f32_e32   v4, v2
v_div_scale_f32 v6, vcc, 1.0, v0, 1.0        v_div_scale_f32 v6, vcc, v0, v5, v0
v_fma_f32       v7, -v4, v5, 1.0             v_fma_f32       v7, -v2, v4, 1.0
v_fmac_f32_e32  v5, v7, v5                   v_fmac_f32_e32  v4, v7, v4
v_mul_f32_e32   v7, v6, v5                   v_mul_f32_e32   v7, v6, v4
v_fma_f32       v8, -v4, v7, v6              v_fma_f32       v8, -v2, v7, v6
v_fmac_f32_e32  v7, v8, v5                   v_fmac_f32_e32  v7, v8, v4
v_fma_f32       v4, -v4, v7, v6              v_fma_f32       v2, -v2, v7, v6
v_div_fmas_f32  v4, v4, v5, v7               v_div_fmas_f32  v2, v2, v4, v7
v_div_fixup_f32 v0, v4, v0, 1.0              v_div_fixup_f32 v0, v2, v5, v0
v_mul_f32_e32   v0, v0, v1                   (x folded in as the numerator)
v_cvt_scalef32_pk_fp8_f32 v0, v0, s0, 1.0    v_cvt_scalef32_pk_fp8_f32 v0, v0, s0, 1.0
v_cvt_scalef32_pk_fp8_f32 v0, s0, v0, ...    v_cvt_scalef32_pk_fp8_f32 v0, s0, v0, ...
```

Identical instruction for instruction. The only difference is the numerator operand — `1.0` vs `v0` — and `True` then needs one extra `v_mul_f32` to apply the reciprocal, so here it is actually **one instruction more expensive**.

**At 16 elements/lane** (`BLOCK_M=1, BLOCK_N=4096`, what the launcher picks for wide rows) the operand choice decides whether the sequence is loop-invariant:

```
--- FAST_CONVERT=True (27 instrs) ---
v_div_scale_f32 v19, s[2:3], v18, v18, 1.0     <- one division, on the uniform scale
v_rcp_f32_e32   v21, v19
v_div_scale_f32 v23, vcc, 1.0, v18, 1.0
v_fma_f32       v8,  -v19, v21, 1.0
v_fmac_f32_e32  v21, v8, v21
v_mul_f32_e32   v24, v23, v21
v_fma_f32       v25, -v19, v24, v23
v_fmac_f32_e32  v24, v25, v21
v_fma_f32       v6,  -v19, v24, v23
v_div_fmas_f32  v6,  v6, v21, v24
v_div_fixup_f32 v6,  v6, v18, 1.0              <- v6 = 1/scale, reused below
v_pk_mul_f32 v[8:9],   v[6:7], v[8:9]   op_sel_hi:[0,1]   <- all 16 elements,
v_pk_mul_f32 v[18:19], v[6:7], v[0:1]   op_sel_hi:[0,1]      2 per instruction
v_pk_mul_f32 v[2:3],   v[6:7], v[2:3]   op_sel_hi:[0,1]
v_pk_mul_f32 v[4:5],   v[6:7], v[4:5]   op_sel_hi:[0,1]
v_pk_mul_f32 v[10:11], v[6:7], v[10:11] op_sel_hi:[0,1]
v_pk_mul_f32 v[12:13], v[6:7], v[12:13] op_sel_hi:[0,1]
v_pk_mul_f32 v[14:15], v[6:7], v[14:15] op_sel_hi:[0,1]
v_pk_mul_f32 v[6:7],   v[6:7], v[16:17] op_sel_hi:[0,1]
+ 8x v_cvt_scalef32_pk_fp8_f32
```

```
--- FAST_CONVERT=False (184 instrs) ---
v_div_scale_f32 v19, s[2:3], v12, v12, v0      <- numerator is per-element data,
v_div_scale_f32 v21, s[2:3], v12, v12, v8         so nothing hoists; 5 divisions
v_rcp_f32_e32   v30, v19                          get software-pipelined together
v_div_scale_f32 v23, s[4:5], v12, v12, v1
v_rcp_f32_e32   v31, v21
v_div_scale_f32 v25, s[8:9], v12, v12, v11
v_rcp_f32_e32   v32, v23
v_div_scale_f32 v27, s[10:11], v12, v12, v2
v_rcp_f32_e32   v33, v25
v_rcp_f32_e32   v34, v27
... interleaved refinement of those five ...
v_div_fixup_f32 v0,  v19, v12, v0
v_div_fixup_f32 v8,  v19, v12, v8
v_div_fixup_f32 v1,  v19, v12, v1
v_div_fixup_f32 v11, v19, v12, v11
v_div_fixup_f32 v2,  v19, v12, v2

                                               <- then the remaining 11 elements,
v_fma_f32       v19, -v29, v35, 1.0               one full sequence each, e.g.:
v_fmac_f32_e32  v35, v19, v35
v_div_scale_f32 v19, vcc, v13, v12, v13
v_mul_f32_e32   v20, v19, v35
v_fma_f32       v21, -v29, v20, v19
v_fmac_f32_e32  v20, v21, v35
v_fma_f32       v19, -v29, v20, v19
v_div_fmas_f32  v19, v19, v35, v20
v_div_fixup_f32 v13, v19, v12, v13
                                               ... x11, elided for length ...
+ 8x v_cvt_scalef32_pk_fp8_f32
```

Totals for the whole kernel: 92 instructions vs 254, with 1 `v_rcp_f32`/`v_div_fixup_f32` vs 16 of each. Zero `v_pk_mul_f32` on the `False` side, since the division sequence cannot be packed.

So the sequence really is the same on both paths — your read was right. What differs is that `1.0 / scale` depends only on a uniform scalar, so LLVM lifts it out and leaves one packed multiply per two elements, whereas `x / scale` has to run it on every data VGPR.

vgokhale: So @lburzawa I think in summary looks like it is still doing NR on the final result with FAST_CONVERT off vs on the scale reciprocal with it on.

lburzawa (approved):

azaidy (approved):
LGTM!

PR #5261, opened 2026-09-03
[Triton/Gluon] split K correct OOB afp4wfp4

afriedri:
## Summary 

`gemm_afp4wfp4_preshuffle` passes `K` in **element** units to the split-K setup, but the kernel's pointer arithmetic and its bounds guard are in **packed-byte** units. Every split therefore covers twice its intended slice of K, and the upper half of the splits read entirely past the end of the activation tensor — a GPU page fault (`Memory access fault ... Reason: Unknown`) on gfx950.

This only fires when `NUM_KSPLIT > 1`. **13 of the 26 tuned gfx950 configs for this op request split-K at small M**, so it is reachable on common decode-shaped GEMMs rather than an exotic corner.

Fix: use the packed-byte `K` at the three sites that pair with `SPLITK_BLOCK_SIZE // 2`, matching what the non-preshuffled `gemm_afp4wfp4` path already does.

## Testing

Standalone harness on gfx950 (MI355X), ROCm 7.2.3, replicating vLLM's MXFP4 call sequence (`per_1x32_f4_quant_hip` → `gemm_afp4wfp4_preshuffle`) across the four Llama-3.3-70B linear shapes × M ∈ {1, 16, 32, 64}:

- **before:** faults on 5/16 combinations — those where tuned config `NUM_KSPLIT > 1`
- **after:** 16/16 pass, and the split-K result is **bit-identical** to the same shape forced to `NUM_KSPLIT = 1` (max abs diff 0.0).

End-to-end: `vllm serve amd/Llama-3.3-70B-Instruct-MXFP4-Preview` on gfx950 faulted during cudagraph capture at batch size 64 — the first captured size that routes to this kernel, since vLLM sends `M <= 64` to the preshuffle path and everything above to the ASM `gemm_a4w4`. With this patch the server starts cleanly.


afriedri:
Added gfx950 runtime validation and a regression test in 89eceec6e.

**gfx950 validation** (MI355X, ROCm 7.2.3, this branch):

- New `test_gemm_afp4_wfp4_preshuffle_splitk`: **18/18 pass** with the fix; with the `K_bytes` change reverted the same test dies with `Memory access fault by GPU node-9 ... Reason: Unknown` (core dumped), so it does fail closed on the bug.
- Full `test_gemm_afp4wfp4.py`: **250 passed, 290 skipped**, no regressions.
- End-to-end `vllm serve amd/Llama-3.3-70B-Instruct-MXFP4-Preview` on gfx950 starts cleanly; it faulted during cudagraph capture at batch size 64 before.

k50112113 (approved):
LGTM, thanks for the catch

PR #5262, opened 2026-09-03
[HIP] [CK] perf(tuner): build the result frame once instead of concatenating per row

ThomasNing:
## Problem

`result_to_df` grows its dataframe one row at a time:

```python
temp = pd.DataFrame(key_dict)
resultdf = temp if resultdf.empty else pd.concat([resultdf, temp], ignore_index=True)
```

That is O(n²) in time and allocation. It only matters at scale — but at scale it matters a lot: with `-o2` (profile every candidate), a 42-shape × ~600-candidate `gemm_a8w8_bpreshuffle_tune.py --libtype all` run spends **hours in this loop after all GPU work has finished**. The GPUs sit idle, the process burns one core at 100%, and the tuner looks hung. We hit this on a real MI350X tuning campaign and initially misdiagnosed it as a wedged tuner.

## Fix

Accumulate the row dicts and concatenate once at the end. No behavioural change.

Applied to `GemmCommonTuner.result_to_df` and the three tuners that carry their own copy of the loop (`ck_gemm_a8w8_bpreshuffle`, `gemm_a16w16`, `ck_gemm_a8w8_blockscale`).

## Evidence

Measured inside a ROCm 7.2.4 image with the pandas it ships, same machine, synthetic rows matching the real schema:

| rows | before | after | speedup |
|---|---|---|---|
| 4000 | 6.26 s | 0.88 s | **7×** |

Output frames compare equal (`df.equals(df2)` → True). The gap widens with row count — our production runs produced 19k–52k rows.

## Notes

Found while tuning GLM-5.2 (gfx950). Several other tuner-scalability issues surfaced in the same campaign (results only written at the end so an interrupted run loses everything; the JIT build counting against the per-task `--timeout`; one infeasible instance failing every TU in the ninja module; JIT baton liveness comparing pid+hostname, which is ambiguous under `--network host`). I'm happy to file those separately as issues or PRs if useful — this PR is deliberately the smallest, most clearly-safe one.


ThomasNing:
Review is clean with no open threads on `b0b6242c9`. The remaining CI entries are fork-workflow approvals, not executed test failures. Maintainer: please approve the [Checks run](https://github.com/ROCm/aiter/actions/runs/33818793159) and [Aiter Test run](https://github.com/ROCm/aiter/actions/runs/33818793801).

valarLip (dismissed):

Issue #5293, opened 2026-09-05
[Perf] pa_decode_sparse: BLOCK_K=64 is suboptimal above the CU count — up to +43 % on gfx950 with a load-dependent choice


stefanskiasan:
`pa_decode_sparse` hardcodes `BLOCK_K = 64` (attention/pa_decode_sparse.py, gfx950 gluon driver). On MI355X this is only the right choice while the launch fits on the CUs — above that, `BLOCK_K = 32` is faster by up to **43 %**, with identical numerics.

## Measurements

MI355X (gfx950), 64 heads, head_dim 512 (448 NoPE + 64 RoPE), packed fp8 cache, top-k 2048 per query, `has_invalid=False`. Times are the minimum of 5 rounds × 15 iterations.

| queries | K=64 | K=32 | better | delta |
|---|---|---|---|---|
| 1–16 | 62–64 µs | 62–65 µs | K=64 | 0.4–1.2 % |
| 24 | 78.4 | 72.8 | **K=32** | 7.6 % |
| 28 / 30 / 32 | 73–77 | 90–96 | K=64 | 19.2 % |
| 34 / 36 | 122–131 | 103–116 | **K=32** | 12.9–18.6 % |
| 48 | 121.7 | 110.3 | **K=32** | 11.1 % |
| 60 / 64 | 131–134 | 165 | K=64 | 19.0–19.9 % |
| 68 / 72 | 204–207 | 170–193 | **K=32** | 6.0–21.6 % |
| 96 | 218.9 | 193.1 | **K=32** | 13.0 % |
| **128** | 264.5 | **186.9** | **K=32** | **41.5 %** |
| 192 | 403.2 | 354.2 | **K=32** | 13.8 % |
| 256 | 552.1 | 399.5 | **K=32** | 38.2 % |

Max abs deviation against a torch reference is 0.000061 for **both** variants — this is purely a scheduling effect, not a numerics tradeoff.

## Why it is not monotonic in `num_queries`

The switch tracks the number of CTAs, not the query count:

```
ctas = num_queries * num_splits * heads_blocks
```

`num_splits` comes from `_decode_num_splits`, so e.g. 32 queries can launch fewer CTAs than 34. Sorting the same data by `ctas` makes it monotonic, and the threshold lands exactly on the CU count (`get_num_sms()` = 256 here):

```
BLOCK_K = 32 if ctas > get_num_sms() else 64
```

[...]

Issue #2416, opened 2026-03-22
SDPA decode on gfx950: <6% HBM utilization, needs PA decode kernel + tuning


sunway513:
## Description

PyTorch SDPA decode (seq_q=1) on MI355X achieves only 0.3-6% HBM bandwidth utilization, resulting in 0.10x performance vs B300 on worst cases. Production decode should use Paged Attention, not vanilla SDPA.

## Data

| Model | HQ | HK | seq_k | BW Achieved | MI355X Peak | Utilization |
|-------|---:|---:|------:|------------:|------------:|-------------|
| Llama3-405B TP=8 | 16 | 1 | 8192 | 14 GB/s | 5300 GB/s | **0.3%** |
| Gemma2-9B TP=1 | 16 | 8 | 8192 | 160 GB/s | 5300 GB/s | 3% |
| DeepSeek-R1 TP=8 | 16 | 16 | 8192 | 230 GB/s | 5300 GB/s | 4% |
| Kimi-K2.5 TP=1 | 64 | 64 | 32768 | 330 GB/s | 5300 GB/s | 6% |

## Root Cause

1. SDPA uses dense FlashAttention for decode — should use Paged Attention (`pa_decode.py`)
2. GQA `repeat_interleave` creates unnecessary memory copies for HK < HQ
3. No gfx950-specific decode attention tuning (missing PA_DECODE config, LEANATTN for gfx950)
4. MHA path (HQ=HK) is particularly slow (see #2415)

## Recommendations

1. Add `gfx950-PA_DECODE` tuning config
2. Add `gfx950-LEANATTN` config (only gfx942 exists)
3. Consider decode-specific tile sizes in MHA config (BLOCK_M=1 or 4 for seq_q=1)

## Environment

- GPU: AMD Instinct MI355X (gfx950)
- PyTorch: 2.8.0+rocm7.2

cc @sunway513

PR #5294, opened 2026-09-05
[Triton/Gluon] [gfx950] pa_decode_sparse: pick BLOCK_K by occupancy, not a constant (up to +43 %)

stefanskiasan:
Fixes the fixed `BLOCK_K = 64` in the gfx950 sparse-decode driver, as described in #5293.

**What changes:** `BLOCK_K` is chosen after `num_splits` is known — 32 when the launch produces more CTAs than there are CUs, 64 otherwise. One comparison, no extra work.

**Why the grid size and not the query count:** the two are not proportional, because `num_splits` varies. 34 queries can launch more CTAs than 64 do, which is why the fastest tile width is not monotonic in `num_queries`. Sorted by CTA count it is monotonic, and the crossover sits exactly at `get_num_sms()`.

**Why it helps:** this kernel is neither bandwidth- nor compute-bound — on MI355X it reaches ~10 % of peak HBM bandwidth (even with fully contiguous indices, so it is not the gather pattern) and ~7 % of peak flops. It is latency-bound, so occupancy is what matters. The narrow tile uses 32 KB LDS instead of 64 KB and lets several CTAs share a CU.

**Measurements** (MI355X/gfx950, 64 heads, head_dim 512 = 448 NoPE + 64 RoPE, packed fp8 cache, top-k 2048, minimum of 5 rounds × 15 iterations):

| queries | fixed K=64 | this patch | delta | vs per-point optimum |
|---|---|---|---|---|
| 16 | 63.2 µs | 62.6 µs | +0.9 % | −0.9 % |
| 24 | 78.3 | 72.9 | +7.4 % | +0.4 % |
| 32 | 73.4 | 73.4 | +0.1 % | −0.1 % |
| 48 | 121.3 | 109.7 | +10.6 % | +0.0 % |
| 64 | 131.0 | 131.1 | −0.1 % | +0.1 % |
| 96 | 218.0 | 193.8 | +12.5 % | +0.1 % |
| **128** | 264.6 | **185.4** | **+42.7 %** | +0.2 % |
| 192 | 402.8 | 355.2 | +13.4 % | +0.0 % |
| 256 | 550.2 | 392.4 | +40.2 % | −2.6 % |

Never worse than −0.6 % anywhere on the curve. **Numerics are unchanged** — max abs deviation against a torch reference is 0.000061 for both tile widths.

The rule also transfers unchanged to a 4-bit packed cache variant we tested locally (128 queries: 289.9 → 220.2 µs), so it is not specific to the fp8 layout.

**Things I measured that did not help**, in case it saves someone the search: `GSPT` 8 or 32 instead of 16 (−14 % / −45 %), `waves_per_eu` 1/2/4 (−4 % to −69 %), cache modifier `.cs` instead of `.cg` (does not compile), and forcing `kv_splits` (the automatic choice is already optimal at 64+ queries; more splits cost up to 45 %).

Happy to re-measure on other shapes — different head counts, top-k values, or a bf16 cache — if that would help review.

### DIFF
--- a/aiter/ops/triton/attention/pa_decode_sparse.py
+++ b/aiter/ops/triton/attention/pa_decode_sparse.py
@@ -484,6 +484,8 @@ def _pa_decode_sparse_gfx950_gluon(
 
     # Tuned launch config (gfx950 / MI355), inlined. BLOCK_M = heads per MFMA M-tile;
     # BLOCK_K = KV tile; num_warps = BLOCK_K // 16 (warps tile the dot-N, MFMA N=16).
+    # BLOCK_K starts at 64 and may be narrowed once the grid size is known --
+    # see the occupancy note after num_splits below.
     BLOCK_M, BLOCK_K, MFMA_K, waves_per_eu = 16, 64, 16, 0
     num_warps = BLOCK_K // 16
     NOPE_DIM, ROPE_DIM = 448, 64
@@ -562,6 +564,17 @@ def _pa_decode_sparse_gfx950_gluon(
         num_splits = _decode_num_splits(
             num_queries, heads_blocks, avg_main, avg_extra, BLOCK_K
         )
+
+    # Occupancy: while every CTA still gets a CU of its own, the wider KV tile
+    # (4 warps, 64 KB LDS) does more work per CU and wins. Once the grid needs
+    # more than one wave, the narrower tile (2 warps, 32 KB LDS) packs several
+    # CTAs per CU and hides the gather latency, which dominates this kernel --
+    # it reaches only ~10 % of peak HBM bandwidth even with contiguous indices.
+    # Measured on MI355X: predicts the faster tile on 21/21 load points, worth
+    # up to +43 % at 128 queries (264.5 -> 186.9 us) with identical numerics.
+    if num_queries * num_splits * heads_blocks > get_num_sms():
+        BLOCK_K = 32
+        num_warps = BLOCK_K // 16
 
     if num_splits > 1:
         part_m = torch.empty(


stefanskiasan:
Some context I found after opening this, which may help place the change: **this is the measure #2416 recommended in March, applied to the sparse decode kernel.**

[#2416](https://github.com/ROCm/aiter/issues/2416) documents decode attention on gfx950 reaching **0.3–6 % of HBM bandwidth** across Llama3-405B, Gemma2-9B, DeepSeek-R1 and Kimi-K2.5, and lists as root cause *"No gfx950-specific decode attention tuning"* with the recommendation *"Consider decode-specific tile sizes"*. That recommendation was never carried out for `pa_decode_sparse` — `BLOCK_K` has been a constant since.

Two notes on how our numbers relate to that issue:

- #2416 measures against 5300 GB/s as the MI355X peak, which is the MI300X figure; MI355X carries 8 TB/s HBM3E. Our 840 GB/s is therefore ~16 % of the reference used there, or ~10.5 % of the actual peak — the same order of magnitude, at the upper end of what that issue reports.
- We reach that ceiling **even with fully contiguous indices** (measured: random 754 GB/s, sorted 822, clustered 842, contiguous 834 GB/s at 128 queries). So it is not the gather pattern, and the compute side is idle too at ~7 % of peak flops. That leaves occupancy, which is what this patch addresses.

The LDS numbers back the mechanism: gfx950 has 160 KB addressable LDS per workgroup (aiter's own `_LDS_CAP_BYTES` has `gfx950: 163840`, and LLVM sets `FeatureAddressableLocalMemorySize163840` for `FeatureISAVersion9_5_Common`). At `BLOCK_K = 64` the KV tile alone is `[64, 512]` bf16 = 64 KB, so only two CTAs fit per CU; halving it to 32 KB lets five fit, which is what recovers the latency hiding.

This also lines up with AMD's own guidance in the [ATOM decode blog post](https://rocm.blogs.amd.com/software-tools-optimization/atom-high-interactivity-mi355/README.html) (2026-09-03), which names *"a small-M GEMM or single-query attention operation may not produce enough workgroups to occupy all compute units"* as a root cause and prescribes *"smaller tile heights"* for small-M paths — there applied to GEMM, here to sparse attention decode.

aiter/ops/triton/attention/pa_decode_sparse.py
```diff
@@ -563,6 +565,17 @@ def _pa_decode_sparse_gfx950_gluon(
             num_queries, heads_blocks, avg_main, avg_extra, BLOCK_K
         )
 
+    # Occupancy: while every CTA still gets a CU of its own, the wider KV tile
+    # (4 warps, 64 KB LDS) does more work per CU and wins. Once the grid needs
+    # more than one wave, the narrower tile (2 warps, 32 KB LDS) packs several
+    # CTAs per CU and hides the gather latency, which dominates this kernel --
+    # it reaches only ~10 % of peak HBM bandwidth even with contiguous indices.
+    # Measured on MI355X: predicts the faster tile on 21/21 load points, worth
+    # up to +43 % at 128 queries (264.5 -> 186.9 us) with identical numerics.
```
azaidy: Please reduce the comment size

azaidy (changes requested):
@stefanskiasan 
1. Please rebase
2. Please address copilot comments
3. Would you be open to adding a benchmark file so that your perf benefit can be reproduced by other? Please use the same format as other benchmark files that exist in op_tests/op_benchmarks/triton

PR #5300, opened 2026-09-07
[HIP] [CK] [FlyDSL] Low-M MXFP4 fused-MoE: padded-row launch bound, SiTUv2 stage1 fusion, BM16 inline-sort

XiaobingSuper:
# Low-M MXFP4 fused-MoE: padded-row launch bound, SiTUv2 stage1 fusion, BM16 inline-sort

## Motivation

Decode-shaped fused-MoE calls (`M <= 64`, one HIP-Graph replay per token step) are
latency bound, not throughput bound. The generic MXFP4 two-stage path pays a fixed
cost that does not shrink with `M`: an `E`-wide sort whose padded extent is dominated
by `E * block_m`, and a quant launch grid sized for that extent.

For Kimi-K3 A4W4 SiTUv2 (`E=896`, `topk=16`, `H=3584`, TP8 local `I=384`) at `M=1`
roughly 98% of the quant workgroups are empty, and the frontend (sort + zero + quant)
is a fixed ~8 us on top of ~12 us of GEMM.

This PR contains three independent, generic changes. Everything is driven by ordinary
tuned-config metadata: no path registry, no `path` CSV column, no model-name dispatch,
no M-threshold environment variable.

## 1. Padded-row quant launch bound (all paths)

For `R = M * topk` route slots, `A = min(E, R)` occupied blocks and sort block `BM`,
writing each positive per-expert row count as `c_i = 1 + x_i` gives

```text
sum(ceil(c_i / BM)) = a + sum(floor(x_i / BM)) <= a + floor((R - a) / BM)
```

which is non-decreasing in both `a` and `R`, so `BM * (A + floor((R - A) / BM))` is a
distribution-independent upper bound on padded sorted rows.

`fused_dynamic_mx_quant_moe_sort_hip_bounded` takes `min(legacy_extent, bound)`, so a
caller supplying a trustworthy host-side expert bound can only shrink the grid, never
grow it. Callers that omit the bound keep the legacy extent. Buffer allocation, layout
and ABI are unchanged.

- `moe_quant_padded_rows_upper_bound` (`csrc/include/quant.h`), constexpr with
  static_asserts.
- `routing_num_experts` threaded through `aiter/fused_moe.py` to the quant call sites.

## 2. SiTUv2 fused into the mxmoe-port stage1 epilogue

The mxmoe-port stage1 (`flydsl_mxmoe_g1_a4w4_*`) only implemented SiLU, so a SiTUv2
model could only use the generic `flydsl_moe1_*` family. Kimi-K3 A4W4 was on that
family, which has no BM16 inline-quant variant -- so §3 was unreachable for it, and
more generally for any SiTUv2 model.

Stage1 now takes the activation as a compile-time parameter:

```text
silu:    g * sigmoid(g) * u
situv2:  beta * tanh(g/beta) * sigmoid(g) * linear_beta * tanh(u/linear_beta)
```

`tanh` is evaluated as `sign(x) * (1 - exp2(-2*log2e*|x|)) / (1 + exp2(-2*log2e*|x|))`,
reusing the `exp2`/`rcp` transcendentals the SiLU epilogue already issued, so it is
computed in-register on the stage1 accumulators with no extra pass over the
intermediate. `beta` and `linear_beta` are runtime f32 kernel scalars, not compile
keys, so re-tuning betas does not recompile; the activation itself is a compile key and
is part of both the kernel name suffix and the stage1 AOT job key.

## 3. BM16 inline-sort two-stage (low `M`)

For small `M` the sort is still worth doing, but at `block_m=16` rather than 32, with
activation quantization folded into stage1 so the frontend collapses to a single fused
sort/zero kernel. Three kernels per call instead of five: fused sort/zero, BM16 stage1
with inline quant, BM16 atomic stage2.

At `M=64`, 1024 routes spread over 618 active experts give a mean of 1.7 rows per
expert, so `ceil(c_i/BM) = 1` for `BM=16` and `BM=32` alike: the same block count with
half the GEMM row work. Weight re-reads only begin once some expert exceeds `BM` rows.

Selection is purely from metadata (`run_1stage=0`, `flat=0`, `output_aux`,
`fuse_quant=fp4`, an `inline_quant` stage1 name, and a stage2 whose `sort_block_m`
matches stage1 `BM`) — `_is_mxfp4_inline_sort`, no name allowlist.

Stage1 inline quant re-quantizes the `BM` rows of every block, so its cost grows with
padded rows while the saving (two frontend launches) is fixed. Whether it wins for a
given shape is a tuning question, not a fixed threshold.

## Dispatch and fallback

Validated twice, and neither ever raises:

- **Static**, in `get_2stage_cfgs`: `_is_mxfp4_inline_sort` inspects the tuned metadata
  (`run_1stage`, `flat`, `output_aux`, `fuse_quant`, the stage1 inline-quant kernel name
  and the matching stage2 `sort_block_m`). An inapplicable tuned row is dropped and the
  lookup falls through to the default heuristics.
- **Runtime**, in `_fused_moe_impl`: `_mxfp4_inline_sort_unsupported` checks only what
  the lookup key cannot see -- gate mode, hidden / intermediate padding, EP, bias,
  explicit activation scales, stage2 scatter, and whether a codegen'd BM16 aux instance
  exists. If the selected fast row cannot run this invocation, metadata is re-resolved
  with the fast path disabled.

## Tuning

Both are candidates in the existing tuner, so a new model picks them up with no code
change:

```
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py --mxfp4-flydsl \
    --untune_file <shapes>.csv --tune_file <model>_a4w4_tuned_fmoe.csv
```

`Mxfp4FlydslTuner` now also sweeps both stage1 XCD swizzles. The Kimi-K3 tuned config
rows for `token=1..64` were re-tuned in this candidate space; they move from the
`flydsl_moe1_*` family to `flydsl_mxmoe_g1_a4w4_16x256x256_f16in_nt[_xcd4]`.

## Measured result

Kimi-K3 A4W4 SiTUv2, `E=896`, `topk=16`, `H=3584`, TP8 local `I=384`, BF16 in/out,
MXFP4 per-1x32, gfx950 / MI355X. Compared against the same tree with the pre-existing
tuned rows, under HIP Graph, 9 alternating AB/BA rounds of 300 replays each, over 7
route distributions.

| M | geomean | worst |
|---|---|---|
| 1 | 1.164x | 1.153x |
| 2 | 1.222x | 1.174x |
| 4 | 1.290x | 1.163x |
| 8 | 1.136x | 1.113x |
| 16 | 1.158x | 1.124x |
| 17 | 1.122x | 1.062x |
| 24 | 1.104x | 1.061x |
| 32 | 1.092x | 1.054x |
| 40 | 1.094x | 1.052x |
| 48 | 1.089x | 1.034x |
| 64 | 1.078x | 1.026x |

Overall geomean 1.139x over 77 scenarios, i.e. 12.2% end-to-end latency reduction
across the `M <= 64` matrix, with no scenario below 1.026x. Cosine similarity against
a BF16 oracle is unchanged (0.9552 vs 0.9551 at `M=64`).

## Follow-up (not in this PR)

A sortless `M == 1` executor: at `M == 1` every route slot is a distinct
`(token, expert)` pair, so binding one route per workgroup and accumulating into a
pre-zeroed output with atomics removes the sort entirely. That is a separate change
and will be sent as a stacked PR.

## Verification

On gfx950 / MI355X: the bound is exact on small partitions and bounded quant matches
the legacy launch bit-for-bit; activation is part of the stage1 AOT cache key, so silu
and situv2 rows get distinct entries; the SiTUv2 BM16 graph output holds its cosine
similarity against a BF16 oracle; every AOT job for the shipped MXFP4
configs compiles (388/388); EP and unsupported invocations fall back to legacy instead
of raising.


aiter/fused_moe.py
```diff
@@ -227,6 +233,137 @@ def _return_output(ret, output):
...
+        return f"unsupported gate mode {gate_mode}"
+    if hidden_pad or intermediate_pad:
+        return "hidden/intermediate padding"
+    if hidden_states.dtype != dtypes.bf16:
+        return "activations must be bf16"
+    if not (getattr(w1, "is_shuffled", False) and getattr(w2, "is_shuffled", False)):
+        return "weights are not both preshuffled"
+    for name, value in (
+        ("num_local_tokens", num_local_tokens),
+        ("bias1", bias1),
```
Bernard-Liu, line 330: > bias2 is missing from this unsupported-option check. MXMOE metadata leaves stage2_has_bias=False, so a call that only provides bias2 keeps the inline-sort path but
  > never forwards the bias to GEMM2, silently producing incorrect output. Please reject bias2 here and fall back, or explicitly support and forward the stage-2 bias.

aiter/fused_moe.py
```diff
@@ -227,6 +233,137 @@ def _return_output(ret, output):
...
+    except (AttributeError, KeyError, TypeError, ValueError):
+        return False
+
+
+@functools.cache
+def _mxfp4_aux_instance_supported(experts, topk, hidden, block_m, zero_init):
+    return _mxfp4_moe_sort_internal_is_supported(
+        experts, topk, hidden, block_m, zero_init
+    )
+
```
Bernard-Liu, line 285: > expert_mask is not checked explicitly here. EP currently falls back only incidentally because the raw top-k includes the fake slot and usually misses the code-generated
  > aux instance. If a matching instance is added later, this config could pass validation and then hit the hard NotImplementedError in the output_aux path. Please pass
  > expert_mask into this validator and reject EP explicitly.

aiter/fused_moe.py
Bernard-Liu, line 2570: > This activation-agnostic fallback is unsafe for MXMOE stage1, where the activation is a compile-time variant. A Gelu lookup can reuse a Situv2-tuned row, but only
  > Situv2 explicitly sets act="situv2"; other activations fall through to the default SiLU kernel, silently producing incorrect output and potentially missing the AOT
  > cache. Please exclude MXMOE rows from this fallback or validate and bind the requested activation explicitly.

aiter/configs/model_configs/kimik3_a4w4_tuned_fmoe.csv
```diff
@@ -1,12 +1,12 @@
 gfx,cu_num,token,model_dim,inter_dim,expert,topk,act_type,dtype,q_dtype_a,q_dtype_w,q_type,use_g1u1,doweight_stage1,block_m,ksplit,us1,kernelName1,err1,us2,kernelName2,err2,us,run_1stage,xbf16,flat,tflops,bw,_tag
-gfx950,256,1,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,10.4812,flydsl_moe1_afp4_wfp4_bf16_t32x32x256_w4_kw2_fp4,0.1%,9.5512,flydsl_moe2_layout_afp4_wfp4_bf16_t32x128x128_atomic_sbm32,0.0%,20.0324,0,0,0,6.6,184670.18,
-gfx950,256,2,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,32,0,12.2582,flydsl_moe1_afp4_wfp4_bf16_t32x32x256_w2_kw2_fp4,0.1%,11.1954,flydsl_moe2_layout_afp4_wfp4_bf16_t32x128x128_atomic_sbm32,0.0%,23.4536,0,0,0,11.27,157732.61,
+gfx950,256,1,3584,384,896,16,ActivationType.Situv2,torch.bfloat16,torch.float4_e2m1fn_x2,torch.float4_e2m1fn_x2,QuantType.per_1x32,1,0,16,0,20.2004,flydsl_mxmoe_g1_a4w4_16x256x256_f16in_nt,1.2%,0,flydsl_moe2_layout_afp4_wfp4_bf16_t16x128x128_atomic_sbm16,1.2%,20.2004,0,0,0,6.54,183134.34,
```
Bernard-Liu: > These low-M rows were generated before Mxfp4FlydslTuner correctly handled SiTUv2. The later tuner fix changed the executed activation from SiLU to SiTUv2, but this CSV
  > was not regenerated, so the selected kernels and the recorded us/err values describe the old SiLU tuning run rather than the current production path. Please re-tune
  > these rows with the fixed tuner, or provide equivalent validation showing that the winners and metrics remain valid for SiTUv2.

aiter/fused_moe.py
XiaobingSuper: Good catch. Fixed in acff9a289: bias2 is now rejected by both the static has_stage2_bias check and the runtime inline-sort validator, so the call falls back before GEMM2 instead of silently dropping the bias.

aiter/fused_moe.py
XiaobingSuper: Agreed. Fixed in acff9a289: MXFP4/MXMOE rows are excluded from the activation-agnostic fallback table, so SiLU/SiTUv2 require an exact activation match and Gelu can no longer reuse a SiTUv2 row while executing the default SiLU kernel.

aiter/fused_moe.py
XiaobingSuper: Agreed. Fixed in acff9a289: expert_mask/is_ep is now an explicit unsupported condition in both config selection and runtime validation. Inline-sort falls back before the output_aux path instead of relying on a missing TOPK=17 aux instance.

aiter/configs/model_configs/kimik3_a4w4_tuned_fmoe.csv
XiaobingSuper: Agreed. The seven existing low-M keys (1/2/4/8/16/32/64) were re-tuned with the final SiTUv2 implementation and beta=4/linear_beta=25 in f33e66775; no new shape was added and token=3 was left unchanged. M4/M8/M16 changed XCD selection, while M1/M2/M32/M64 kept the same kernels with refreshed metrics. Public fused_moe smoke for M1/M16/M64 confirmed the expected three-kernel path with no fallback.

Bernard-Liu (approved):

coderfeli (approved):

PR #5301, opened 2026-09-07
[FlyDSL] [Kernel] gfx950 FP8 sparse MLA prefill and decode (FlyDSL)

JohnQinAMD:
## Motivation

DSA-style sparse attention on gfx950  FlyDSL kernels for both prefill and decode phases, plus the split reducer decode needs.

## Kernels

`flydsl_sparse_mla_prefill` takes `q_nope` and `q_rope` separately: the caller already holds them split and the kernel reads them into different LDS regions, so concatenating first would copy the larger half for nothing. Below 96 tokens it splits the topk range across CTAs, because one CTA covers one token and a short prefill otherwise leaves most of the device idle. The partials it emits are the layout the sparse-decode reducer already consumes. Above that the grid fills the device on its own and a `const_expr` selects the single-split path, which emits the same IR as before split-K. The 96-token bound is the shared decode reducer's supported scope, not a tuning constant.

Both entry points take 1 to 16 heads. The MFMA tile is 16 rows wide, so the head count is a runtime kernel argument rather than a padded staging copy: rows past it re-read head 0, in bounds and dropped by the store predicate. GLM-5.2 has 64 heads, so 16 is TP4 and 8 is TP8, and refusing the latter would have excluded the configuration the end-to-end runs used. The kernel takes the same time at 8 heads as at 16, because KV traffic does not scale with the head count.

`flydsl_sparse_mla_decode` adds a split-K reducer. `sparse_mla_decode_workspace_shape` sizes its scratch, so a caller can allocate once and reuse across layers.

`mla_reduce_kernels.py` and `kernels/mla_reduce.py` gain the split-reduce entry points decode needs. One pre-existing path does change: `_resolve_actual_max_splits`, which `flydsl_mla_reduce_v1` calls, used to answer a graph-capture lookup straight from a pointer-keyed cache. A freed buffer hands its address to the next tensor, so a recycled key could return a stale split bound. Entries now carry a weak reference to the storage and a capture-time hit is honoured only while that storage is alive at the same base; a recycled key reports a miss, which the launch path already rejects with a message asking the caller to warm the buffer. Eager behaviour is unchanged.

## Measurements

MI355X gfx950, torch `2.9.1+rocm7.2.0`, tilelang `0.1.7.post3`. fp8 e4m3 `q` and KV, fp32 softmax and accumulation, bf16 out, `P` requantised to fp8 for the PV MFMA. `q` reaches TileLang already in fp8, so the baseline is not charged for a conversion it would not do in service.

One harness throughout: 200k-row KV pool, uniformly random topk, minimum of five HIP-graph replays. SNR is against an fp32 oracle on the pre-quantisation inputs, so it includes the fp8 input error and is the same oracle for every kernel. That error dominates, which is why TileLang, Triton and FlyDSL land within 0.1 dB of one another. The inputs are Gaussian, so the softmax over the 2048 selected keys is close to uniform -- the worst case for quantising `P`, and not what real top-k selection produces. On scores with realistic structure the same kernel measures several dB higher and the split count stops mattering: 16 splits scores +0.04 to +0.30 dB against 1 split there, against -1.0 dB on these inputs. Read the absolute figures as a floor, not as production accuracy. Triton is sgl-project/sglang#30575, run unmodified, using that PR's `triton_sparse_mla_fwd` for prefill and `triton_sparse_mla_decode` for decode. aiter's own sparse-MLA ASM prefill is gfx1250-only, so `mla_decode_fwd` appears in the decode table only.

### Prefill

| `T` | TileLang | Triton | FlyDSL | vs TileLang | vs Triton |
| 1 | 13.52 us | 10.94 us | 7.82 us | 1.73x | 1.40x |
| 4 | 14.56 us | 11.48 us | 8.20 us | 1.78x | 1.40x |
| 16 | 20.25 us | 13.96 us | 10.42 us | 1.94x | 1.34x |
| 64 | 34.02 us | 34.30 us | 20.91 us | 1.63x | 1.64x |
| 96 | 41.90 us | 55.77 us | 25.66 us | 1.63x | 2.17x |
| 128 | 56.82 us | 58.46 us | 54.66 us | 1.04x | 1.07x |
| 256 | 102.43 us | 87.48 us | 59.30 us | 1.73x | 1.48x |
| 512 | 193.90 us | 103.66 us | 98.15 us | 1.98x | 1.06x |
| 1024 | 356.01 us | 200.32 us | 208.63 us | 1.71x | 0.96x |
| 2048 | 610.52 us | 391.42 us | 380.40 us | 1.60x | 1.03x |
| 4096 | 1185.05 us | 764.04 us | 752.28 us | 1.58x | 1.02x |
| 8192 | 2312.95 us | 1474.37 us | 1492.33 us | 1.55x | 0.99x |
| 16384 | 4562.58 us | 2901.64 us | 2968.91 us | 1.54x | 0.98x |
| 32768 | 9030.53 us | 5775.87 us | 5924.40 us | 1.52x | 0.97x |

14/14 against TileLang. Against Triton: ahead below 512, level within 2-4% from 2048 up where both are bandwidth-bound, 4% behind at 1024. Before split-K the first five rows against Triton were 0.24x, 0.25x, 0.26x, 0.64x and 0.61x. SNR spans 27.46-28.57 dB with the three kernels within 0.7 dB of each other.

### Decode

Sizes speculative decode at c<=16 produces -- draft `1,2,4,8,16` and verify `6,12,24,48,96`. aiter is `mla_decode_fwd`, each decode token issued as one request whose page table is its topk, with unity fp32 `q_scale`/`kv_scale` so the numerics match bare fp8.

| `seq` | TileLang | Triton | aiter ASM | FlyDSL | vs TL | vs TR | vs aiter |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 13.51 us | 8.91 us | 15.92 us | 6.47 us | 2.09x | 1.38x | 2.46x |
| 2 | 13.96 us | 9.09 us | 15.74 us | 6.72 us | 2.08x | 1.35x | 2.34x |
| 4 | 14.54 us | 9.57 us | 15.24 us | 7.23 us | 2.01x | 1.32x | 2.11x |
| 6 | 15.03 us | 10.68 us | 16.75 us | 7.76 us | 1.94x | 1.38x | 2.16x |
| 8 | 15.69 us | 10.35 us | 13.38 us | 8.30 us | 1.89x | 1.25x | 1.61x |
| 12 | 18.68 us | 11.05 us | 17.28 us | 9.63 us | 1.94x | 1.15x | 1.79x |
| 16 | 20.20 us | 11.88 us | 18.03 us | 10.36 us | 1.95x | 1.15x | 1.74x |
| 24 | 23.19 us | 15.56 us | 20.19 us | 12.97 us | 1.79x | 1.20x | 1.56x |
| 48 | 28.19 us | 24.52 us | 25.11 us | 19.78 us | 1.43x | 1.24x | 1.27x |
| 96 | 42.11 us | 39.54 us | 39.44 us | 32.87 us | 1.28x | 1.20x | 1.20x |

30/30, and decode output is bitwise deterministic across repeated runs at every size. Triton is the closest of the three.

aiter is about 1.6 dB more accurate on these inputs, and it is worth saying where that comes from, because it is not a scale tensor. Modelling the same arithmetic one step at a time: FP8 inputs with everything else exact give 31.8 dB; quantising `P` to FP8 for the PV MFMA costs 2.75 dB and lands at 29.06, which is where aiter measures. Quantising `P` against a running maximum rather than the row's final one costs a further 0.36 dB, and doing that once per split costs 1.25 dB more, putting TileLang, Triton and FlyDSL at a modelled 27.45 against a measured 27.54. So aiter is not doing something extra -- it sits at the FP8-`P` limit, and pays about 2.4x the time for it. The gap is also specific to these inputs: FP8 `P` against BF16 `P` is worth 3.92 dB here, 2.63 dB at a 1.2-bit score spread, and 0.02 dB at 12.6 bits.
27.3-28.2 dB for Triton -- which is consistent with it reading a scale tensor
per operand that the others do not. Its seq=96 SNR was not reproducible across runs, so it is left out rather than quoted.

### Sensitivity to the KV pool

The prefill margin over TileLang depends on gather locality, and the two kernels do not react to it the same way:

| `T`=1024, pool | TileLang | FlyDSL | speedup |
|---:|---:|---:|---:|
| 2048 rows (1.2 MB) | 303.96 us | 119.35 us | 2.55x |
| 8192 rows (4.7 MB) | 337.19 us | 136.06 us | 2.48x |
| 32768 rows (18.9 MB) | 352.42 us | 166.44 us | 2.12x |
| 131072 rows (75.5 MB) | 355.63 us | 200.13 us | 1.78x |
| 200000 rows (115.2 MB) | 356.08 us | 203.55 us | 1.75x |

TileLang moves 17% across that range, FlyDSL 72%: FlyDSL is near the bandwidth roofline (6.4 TB/s of effective KV read at `T`=4096 against a 5.15 TB/s streaming-copy ceiling) and pays in full when locality degrades, while TileLang is not bandwidth-bound at these sizes. `T`=4096 behaves the same way, 2.61x down to 1.58x. Since a token reaches only its own sequence's KV, the figure tracks context length: roughly 2.0-2.1x at 32k, 1.6-1.8x at 128k. The tables above use the conservative end.

### Known gaps

FlyDSL trails Triton by 2-4% around `T`=768-1024 and I do not have the mechanism. It is not occupancy: the kernel sits at 152 VGPRs, which pins it to 3 CTAs per CU independently of LDS, and padding LDS from 44.5 KB to 69.5 KB or sweeping `waves_per_eu` over 1-4 changes nothing at any `T`.

## Tests

`op_tests/flydsl_tests/`, four files, 35 tests: accepted shapes, the rejection path for every wrong dtype/shape/arch/contiguity case, the split scratch's capacity and its identity under reuse, and the compile-cache domain.

```
$ python3 -m pytest op_tests/flydsl_tests/test_flydsl_sparse_mla_prefill.py \
    op_tests/flydsl_tests/test_flydsl_sparse_mla_decode.py \
    op_tests/flydsl_tests/test_flydsl_split_cache_identity.py \
    op_tests/flydsl_tests/test_flydsl_compile_cache_domain.py -q
35 passed, 2 warnings in 9.56s
```

Black and Ruff 0.16 clean on all changed files.

## Note for reviewers

#5167 also adds a `flydsl_sparse_mla_prefill`, written independently, with a
different signature -- single concatenated `q` plus `indptr`, `attn_sink` and `split_kv`, BF16, `D`=512 -- where this one takes `q_nope`/`q_rope` separately, fp8, `D`=576, and has no varlen or sink support. They are not drop-in replacements and whichever lands second collides on the name in `aiter/ops/flydsl/__init__.py`. Worth deciding which shape the namespace should carry before either merges; I am happy to adapt.

### DIFF
--- a/aiter/ops/flydsl/__init__.py
+++ b/aiter/ops/flydsl/__init__.py
@@ -66,6 +66,18 @@ _LAZY_IMPORTS = {
         ".kernels.qk_norm_rope_quant",
         "flydsl_qk_norm_rope_quant",
     ),
+    "flydsl_sparse_mla_decode": (
+        ".sparse_mla_decode_kernels",
+        "flydsl_sparse_mla_decode",
+    ),
+    "flydsl_sparse_mla_prefill": (
+        ".kernels.sparse_mla_prefill",
+        "flydsl_sparse_mla_prefill",
+    ),
+    "sparse_mla_decode_workspace_shape": (
+        ".sparse_mla_decode_kernels",
+        "sparse_mla_decode_workspace_shape",
+    ),
 }
 
 __all__ = [
@@ -84,6 +96,9 @@ __all__ = [
     "flydsl_pa_mqa_logits_fp4_varqlen",
     "flydsl_preshuffle_gemm_a8",
     "flydsl_qk_norm_rope_quant",
+    "flydsl_sparse_mla_decode",
+    "flydsl_sparse_mla_prefill",
+    "sparse_mla_decode_workspace_shape",
 ]
 
 
--- a/aiter/ops/flydsl/kernels/mla_reduce.py
+++ b/aiter/ops/flydsl/kernels/mla_reduce.py
@@ -134,6 +134,26 @@ def _load_partial_out(buf, row, head_idx, tid, vec):
     return elems
 
 
+def _narrow_numeric_t(name: str):
+    """Resolve a 2-byte partial-output element type."""
+    if name in ("bf16", "bfloat16"):
+        return fx.BFloat16
+    if name in ("fp16", "f16", "float16", "half"):
+        return fx.Float16
+    raise ValueError(f"Unsupported partial dtype: {name}")
+
+
+def _load_partial_out_narrow(buf, row, head_idx, tid, vec, numeric_t):
+    """Load VEC packed 2-byte partials and widen them to fp32 registers."""
+    atom = _out_copy_atom(vec, numeric_t)
+    reg_layout = fx.make_layout(vec, 1)
+    row_tiled = fx.logical_divide(fx.slice(buf, (row, head_idx, None)), reg_layout)
+    frag = fx.make_rmem_tensor(reg_layout, numeric_t)
+    fx.copy_atom_call(atom, fx.slice(row_tiled, (None, tid)), frag)
+    v = frag.load()
+    return [v[i].to(fx.Float32) for i in fx.range_constexpr(vec)]
+
+
 def _out_copy_atom(vec: int, out_numeric_t):
     """Widest legal buffer copy atom for a VEC-element output fragment.
 
@@ -253,6 +273,10 @@ def compile_mla_reduce(
     grp_m64: int = 8,
     m64_hi_thr: int = 8,
     m64_hi_grp: int = 16,
+    partial_dtype: str = "fp32",
+    lse_base2: bool = False,
+    num_threads: int = NUM_THREADS,
+    dv_split: int = 1,
 ):
     """Compile an MLA reduce kernel for fixed (H, Dv, out_dtype, tier).
 
@@ -279,11 +303,34 @@ def compile_mla_reduce(
     ``m64_hi_grp`` pipeline once ``n_splits`` exceeds ``m64_hi_thr``, without
     changing CUDA-graph topology. Set ``m64_hi_thr=0`` to disable it.
     """
-    assert (
-        Dv % NUM_THREADS == 0
-    ), f"Dv ({Dv}) must be divisible by NUM_THREADS ({NUM_THREADS})"
-    assert tier in _TIER_NLSE, f"bad tier {tier}"
-    VEC = Dv // NUM_THREADS
+    NT = int(num_threads)
+    DVS = int(dv_split)
+    if Dv % (NT * DVS) != 0:
+        raise ValueError(f"Dv ({Dv}) must be divisible by NT*DVS ({NT * DVS})")
+    if tier not in _TIER_NLSE:
+        raise ValueError(f"bad tier {tier}")
+    VEC = Dv // (NT * DVS)
+    partial_is_narrow = partial_dtype != "fp32"
+    partial_numeric_t = (
+        _narrow_numeric_t(partial_dtype) if partial_is_narrow else fx.Float32
+    )
+    if partial_is_narrow and VEC not in (1, 2, 4, 8):
+        raise ValueError(f"narrow partial needs VEC in 1/2/4/8, got {VEC}")
+
+    if lse_base2:
+
+        def _xexp(x):
+            return fx.rocdl.exp2(T.f32, fx.Float32(x).ir_value())
+
+        def _xlog(x):
+            return fly_math.log2(x, fastmath=fm_fast)
+
+    else:
+        _xexp = _exp
+
+        def _xlog(x):
+            return fly_math.log(x, fastmath=fm_fast)
+
     is_runtime_tier = tier == Tier.ALL
     is_massive = tier != Tier.SIMPLE
     if tier == Tier.MLDS:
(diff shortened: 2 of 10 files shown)


coderfeli:
why compare with tilelang? do you have perf of triton or hip? @JohnQinAMD

JohnQinAMD:
> why compare with tilelang? do you have perf of triton or hip? @JohnQinAMD

@coderfeli 
TileLang was the starting point because it is the current only implementation covering both phases on gfx950 for glm5.2 and is what sglang's ROCm path picks by default. The description now carries Triton and aiter as well, all three under same test setup. 


Triton is sgl-project/sglang#30575, which is going to be merged to sglang-- `triton_sparse_mla_fwd` for prefill, `triton_sparse_mla_decode` for decode.

HIP: aiter's mla_decode_fwd is the ASM column in the decode table. For prefill there is a gfx950 path — pa_sparse_prefill_fp8_opus — but it is a different operator rather than a different implementation of this one: D=448+64 with the NoPE half in MXFP8 (32-element blocks, E8M0 exponents packed into the row) and bf16 RoPE, against this kernel's D=512+64 with both halves in plain FP8. There is no shared input a like-for-like run could take. mla_sparse_prefill_fp8_asm is gfx1250-only.

| | vs TileLang | vs Triton | vs aiter ASM |
|---|---|---|---|
| prefill `T`=1..96 | 1.63-1.94x | 1.27-2.17x | — |
| prefill `T`=256..512 | 1.73-1.98x | 1.06-1.48x | — |
| prefill `T`>=2048 | 1.52-1.60x | 0.97-1.03x | — |
| decode `seq`=1..24 | 1.79-2.09x | 1.15-1.38x | 1.56-2.46x |
| decode `seq`=48..96 | 1.28-1.42x | 1.20-1.24x | 1.20-1.27x |

The full sweep result has been updated in the PR body

PR #5303, opened 2026-09-07
[FlyDSL] replace ptr_rsrc and buffer ops low level api use

coderfeli:
clean all vector.py usage.
reduce buffer ops usage.
clean scf, vector in moe.
all same perf.


XingerZhu:
LGTM （test moe ut on a07-3）: 
image
image

PR #5306, opened 2026-09-07
[FlyDSL] Remove the unused gfx1250 d192 FMHA sibling kernel

coderfeli:
## Summary

Deletes the hand-tuned, assembly-mirroring **d192 FMHA sibling** for gfx1250 and its private helper modules. Net **−7747 lines**.

The clean-DSL **m32x8** kernel has been the default for that shape for a while and is ~16% faster (per the existing in-tree comment), so the sibling never runs in a production path. This removes it and collapses the routing to the single `flash_attn_varlen_m32x8` call.


ruanjm (approved):
LGTM.

PR #5308, opened 2026-09-07
[skills] validate-kernel-pr: prose for judgement, code for the ledger

zhiding512:
`validate-kernel-pr` mixed target-selection judgement with execution bookkeeping. This PR makes the caller declare the target and runner, while code records what actually ran, which route executed, what was measured, and the resulting verdict.

- Move report writing, GPU identity, target preparation and performance accounting into testable Python helpers. The shell keeps process launch and the GPU lock; `report.py` owns verdict and exit-code derivation.
- Remove shape-grid injection and its source-based coverage guesses. Execution receipts retain route and optional shape observations. The schema accepts route-only evidence, including an empty shape list; it still requires observed route execution for PASS.
- Discover benchmarks shipped by the PR and existing repository benchmarks that import changed modules. Time both when both resolve, including a benchmark already selected for correctness, and deduplicate overlapping targets. A correctness run cannot replace a base/head timing comparison. The worst attributable measurement gates; skipped measurements keep their reasons and provenance.
- For transplanted baselines, require a nonblank control-column name. Prefer an exact match, accept a unique substring, and decline ambiguous names or a control that moved beyond tolerance.
- Construct the environment for runtime compatibility imports and build-identity imports as well as correctness and timing runs. All use the same allowlist and private cache/home paths.
- Integrate with current main's split `review-pr` workflow. `fetch.sh` forwards `REVIEW_RUNNER`, its reason, route/shape observations and explicit perf settings; an undeclared file runner is reported before claiming a GPU. Its consumer accepts valid PASS, INCONCLUSIVE and no-target evidence under the new contract. P6 reads all measurements without duplicating the validator's missing-perf finding.

Validation: **189 CPU-only tests passed** (186 seconds), plus Bash syntax, Black, Ruff and `git diff --check` against current main. Regressions exercise the complete `fetch.sh -> validate_pr.sh -> report consumer` path for pytest, scripts named `test_*.py`, missing runners and missing routes; a synthetic 2x slowdown that previously escaped timing; and import-environment canaries for both aiter and FlyDSL. Current main's exception-receipt and untracked-file checks are preserved.

The tests verify tooling behavior and do not claim live-GPU performance or kernel correctness. Removing grid injection also removes its pinned GPU-mutant replay. Route receipts establish execution; deciding whether the selected target covers the changed behavior and whether its shapes are adequate remains the reviewer's responsibility.


zufayu:
This PR should be rebased onto the latest `main` first, as the `review-pr` execution logic has moved from `SKILL.md` to `fetch.sh`.

After rebasing, there are two integration issues to address:

1. **`review-pr` does not propagate the runner selection to `validate-kernel-pr`.**

   This PR makes runner selection caller-owned: a plain file target defaults to `runner=none` unless the caller passes `--runner` or uses an explicit `::node` selector. However, the `review-pr` auto-validation path currently invokes the validator with only `--target "$AUTO_TARGET"`.

   Because the automatically selected target is a file path, validation will select `runner=none`, skip the test, and remain `INCONCLUSIVE`, even when an idle GPU is available. The caller should pass an explicit runner declaration, such as `--runner pytest`, together with a checkable runner reason.

2. **The pytest failure-attribution path reads `observed_work`, but pytest stats do not provide that field.**

   `validate_pr.sh` queries `observed_work` from the pytest stats, while `validate_evidence.py pytest-stats` does not emit it. `target_run.py stats-field` directly indexes the requested key, resulting in a `KeyError`.

   This can bypass the intended protection against attributing a grid failure that never reached the routed kernel and may incorrectly produce a deterministic blocker. `observed_work` should either be derived from the execution receipt or handled explicitly when the field is unavailable.

The `observed_work` issue was also identified in the existing Copilot review, but it still appears to be present at the current PR head. Please rebase, migrate the `review-pr` integration changes to `fetch.sh`, fix both issues, and add an end-to-end test covering auto-selected pytest validation with explicit runner propagation.

zufayu:
A self-contained defect found in the skill (worth reporting upstream)
On the first run, the verdict was NEEDS_WORK, while correctness/receipt/perf were all skipped. The reason was "base did not leave a clean worktree".
Root cause identified by GLM:
`validate_pr.sh` sets `PYTHONDONTWRITEBYTECODE=1` for test invocations (lines 731/765/815/824). However, this env var was missing when importing `pick-idle-gpu.py` via `importlib.exec_module` (lines 446/629).
Consequently, the validator itself wrote the following file into the worktree:
`.claude/skills/validate-kernel-pr/__pycache__/pick-idle-gpu.cpython-312.pyc`
This file was flagged as dirty by the built-in ignored-artifact check.
After adding `-e PYTHONDONTWRITEBYTECODE=1`, the second run passed all stages normally.
This is a genuine bug where **the validator pollutes the target worktree being validated**.
Fix: Inject the same environment variable for those two Python invocations.
(Modifying `validate_pr.sh` directly would dirty the worktree, so the env var must be injected externally.)

zufayu:

**Refactors the validate-kernel-pr skill so the caller declares the target and runner while new python tools keep the ledger (report.py verdict and exit-code derivation, gpu_probe.py identity, target_run.py receipts and env isolation, scrape_perf.py harness detection and perf accounting), drops the shape-grid injection layer, adds benchmark discovery (PR-shipped plus repo-owned) with transplanted baselines gated by a control column, and updates review-pr's fetch.sh/SKILL.md to forward runner and perf declarations and to accept route-less reports.**

🔴 [verified] scrape_perf.py:870-917 — the transplant attribution gate never excludes the control column from the gated measurement, and its tolerance (PERF_CONTROL_TOL, default 0.10) is wider than the regression threshold (0.95): when the untouched control column drifts 8% across the two trees, attribute() accepts the comparison ("reproduced within 8.2%") while compare() (scrape_perf.py:412) has already picked that same control column as worst_column — reproduced with a 4-row table (kernel column at ratio 1.0, reference column at 0.9176): the stage reports fail with a should-fix "head is slower than base … reference us", flipping an innocent PR to a deterministic NEEDS_WORK and exit 1 on a column the patch does not touch, the exact false regression this stage's own prose promises can never happen. **Author must** exclude the control column from the gated columns (or recompute the verdict after attribution and skip when no subject column remains) and derive PERF_CONTROL_TOL against the threshold rather than leaving it twice as wide.
⚠️ [verified] scrape_perf.py:902 — attribute() reads columns[match].median_ratio without checking the eligible flag compare() itself sets: a control column with fewer usable samples than PERF_MIN_ROWS (marked eligible: false, "cannot set the verdict") still authorizes a cross-tree transplant — reproduced with a 4-row table whose control column has 2 samples ("reproduced within 0.0%"), so an unattributable comparison can gate a verdict on a median the comparison refuses to let decide anything. **Author must** require columns[match].eligible before trusting the control ratio.
⚠️ [verified] report.py:257-263 — a schema violation is recorded as a finding and downgraded only when the verdict is PASS; a violating BLOCK or NEEDS_WORK report is still copied to --out with the violation only on stderr, and the review-pr consumer validates every supplied report against report_schema.json and rejects it wholesale — so a genuine deterministic blocker is discarded and the review aborts with no trace of the malformation in the published report; no test covers the non-PASS path. **Author must** record the violation as a finding on every verdict and publish a schema-valid degraded report so a run's evidence survives its own malformation.
⚠️ [verified] .claude/skills/validate-kernel-pr/tests/replay_mutants.sh and tests/mutants/ deleted — the rewritten probe delivery (run_script_with_probe), transplant baselines and bench discovery now carry no real-kernel seeded-defect evidence: the synthetic suite seeds defects only in fixture targets, and the skill's own bar (everything shipped has been observed failing on a seeded defect) is unmet, which SKILL.md itself calls a real gap. **Author must** rebuild at least one replay case the target's own shapes can reach (the tolerance-widening mutant needs no grid, as SKILL.md notes) before this rewrite replaces the only end-to-end proof the skill ever had.
📝 [verified] scrape_perf.py:957 vs 974-980 — when the control gate rejects a transplant comparison, perf_stage pops median_ratio/worst_column/regressed_rows so a skip "must not carry the numbers it rejected", but stage["columns"], set before the pop, still publishes every column's median_ratio beside that skip — the exact "regression that was merely not acted on" reading the pop exists to prevent. **Author must** drop or annotate columns on a control-rejected skip too.

PR #5313, opened 2026-09-07
[FlyDSL] [gfx1250] Speed up the grouped-MoE route, quant, and gather-reduce epilogues

lalala-sh:
Three changes to the gfx1250 grouped-MoE auxiliary kernels, found by profiling
DSv4 TP4 prefill (7168/768, E=384, topk 6) where they had grown to ~30% of the
layer while the GEMMs took the rest.

## Grid-stride the contiguous psum row remap

`moe_contiguous_psum_remap` ran at `grid=(1,1,1)`. The scan it fuses is over E
and costs nothing, but the remap that follows walks every (token, topk) route,
so one workgroup was resolving ~200K routes on a single CU.

Every block now re-runs the cheap scan into its own LDS start table, so the
grid-strided remap resolves `starts[expert]` from LDS with no cross-block
barrier; only block 0 writes the global outputs. This is the treatment
`moe_contiguous_psum_remap_ep` already had. A dedicated `MAX_REMAP_EXPERTS`
table (8KB of the 320KB budget) lets the chunked, carry-based scan stay, so E
is not capped at the 512-lane scan width and the single-block variant is
deleted rather than kept as a wide-E fallback.

    psum kernel   4096 tokens    30.2us -> 3.8us   (7.9x)
                 32768 tokens   227.7us -> 5.1us   (45x)

## LDS route cache + one flat resource in gather-reduce

Cherry-picked from `wjx/opt_small_ops`. The topk route rows and weights are
block-uniform, so caching them in LDS drops 12 redundant per-thread loads, and
addressing `grouped_out` through one flat buffer resource takes the per-row
descriptor construction out of the load loop.

    gather_reduce 16384 tokens   209.0us -> 143.5us  (1.46x, 7.87 -> 11.46 TB/s)

## Quantize each token once instead of once per route

The topk=6 stage1 input prep quantized the same hidden row once per route, so a
token's activation was read and converted six times. A new warp-per-token
kernel reads the row once, computes the MX payload and e8m0 scale once, then
emits six destinations. It is dispatched from
`flydsl_moe_fused_quant_preshuffle` when topk==6 and tokens >= 4096;
`AITER_FLYDSL_TOKEN_MULTIDEST_QUANT=0` falls back to the route-ksplit path.

Two things this needs to be worth it:

- The 16-row-interleaved a1 scale makes the quant pass write 4 B per cache
  line. Storing it row-major instead moves the interleave into gemm1's
  LDS->register read, which is free (`row_major_ascale`; ~12us off quant at 16k
  tokens, gemm1 unchanged). `AITER_FLYDSL_ROWMAJOR_ASCALE=0` reverts.
- Row-major then lets a row's e8m0 leave as one dwordx4 per 8 MX blocks instead
  of a byte per block.

Hidden rows are staged through LDS by TDM in chunks, each chunk issuing the next
one's DMA before converting the current one. A warp whose token is past the end
takes zero-length destination descriptors rather than a branch, so the staging
barriers stay CTA-uniform for ragged token counts.

    quant kernel  16384 tokens    609 MB in 46.6us  (13.1 TB/s)

That is close to the ceiling for this access pattern, which is worth recording
because it says where the remaining headroom is not. Measured separately on the
same shape: reads sustain 16.4 TB/s, writes 12.4 TB/s, and the two are strictly
additive (14.2 + 30.2 = 44.4us predicted vs 44.3us measured), so the pattern's
floor is 44.4us and the quant math itself costs only ~2us. Replacing the payload
stores with TDM 2D or TDM gather-store made it slower, not faster (11.9 vs 13.3
TB/s), and a plain linear dwordx4 stream tops out at 13.5 TB/s -- the existing
`buffer_store` path is already at 99% of that. The only lever left is writing
fewer bytes: 352 of the 374 MB written are six copies of the same 59 MB, which
would need gemm1 to gather A by row instead of reading it linearly.

Also sets `num_buffers=3` for the a4w4 gemm1 rows at 16384/32768 tokens.

## Test plan

- [x] `test_flydsl_grouped_gemm_gfx1250.py --scenario verify`, a8w4 and a4w4,
      E=384 and E=896 (the latter only reached the single-block psum before)
- [x] Prefill sweep 4096/8192/16384/32768, accuracy unchanged
      (E=384 a8w4 `logits_diff` bit-identical at 4.1555e-06)
- [x] Dropped-route regression above, on a NaN-poisoned row 0
- [x] Same-session interleaved A/B for both kernels; `gather_reduce` reads
      identically in both arms of the psum A/B, `moe_route` in both arms of the
      gather A/B
- [x] Quant path cross-check: multidest+row-major, multidest+interleaved, and
      the old route-ksplit path give bit-identical `logits_diff` on a4w4/a8w4 at
      4096 and 16384 tokens
- [x] Token sweep 1/7/64/1023/1024/2048/4095/4096/4097/4103/8192/12289/16384/
      16385, a4w4 and a8w4 -- covers the 4096 dispatch boundary and ragged
      counts that exercise the zero-length descriptors
- [x] silu / swiglu / situv2, with and without bias
- [x] Full tuned-CSV sweep: 85/85 pass, worst `logits_diff` 9.4e-05 (gate 0.01)
- [ ] EP / multi-GPU (`test_mega_moe_gfx1250.py`) -- not run, single GPU here.
      Worth a look since the NaN fix is on the EP dropped-route path.

Measured on one gfx1250, SPX/NPS1.


coderfeli:
Looks good. e2e validated?

lalala-sh:
> Looks good. e2e validated?

Not Yet.

junhaha666 (approved):

PR #5321, opened 2026-09-08
[Triton/Gluon] [Kimi-K3][ROCm] Add merged MoE front

jiacao-amd:
## Summary

Adds a merged MoE-front kernel for Kimi-K3 on gfx950, plus tuned GEMM
configurations and numerical tests for it.

Kimi-K3's MoE front is three separate projections of the same `x[M, 7168]`:
shared gate/up `7168 -> 1536`, router logits `7168 -> 896`, and the routed
latent `7168 -> 3584`. This kernel combines them into one
`BF16 x BF16 -> FP32` GEMM (`7168 -> 6016`) followed by a Triton split/SiTU
epilogue that emits the BF16 shared SiTU intermediate, the FP32 router logits,
and the BF16 routed latent. Routed experts are untouched and still use the
existing AITER MXFP4 backend.

## Supported shapes

```text
1..16, 32, 48, 64, 80, 96, 112, 128, 192, 512, 1024, 1536, 2048
```

`32..192` is the MTP decode bucket. A pure-decode step submits
`num_seqs * (1 + num_speculative_tokens)` tokens, so with the usual
`num_speculative_tokens=3` any server past four concurrent sequences asks for a
count above 16. The gate itself lives in the companion vLLM change; what this
PR adds for those shapes is numerical coverage, since the decode test
previously stopped at `m=16`.

`kimik3_bf16_tuned_gemm.csv` gains FP32-output entries at `N=6016, K=7168,
gfx950, cu_num=256` for `M = 7, 14, 16, 512, 1024, 1536, 2048`. Config lookup
is keyed on output dtype and pads `M` up to the next multiple of 16 below 256,
so every `M` in `1..16` resolves to the `M=16` entry. The `32..192` bucket has
no FP32-output entry yet and therefore runs on torch's default hipBLASLt
heuristic; the end-to-end numbers below already include that fallback, so
tuning those shapes is remaining headroom rather than a prerequisite.

`M=4096` and `M=8192` are intentionally excluded after controlled retesting
showed regressions.

## Kernel results

Single-GPU gfx950, complete front, CUDA Graph:

| M | Speedup vs native front |
|---|---:|
| 1..16 | 1.645x – 2.140x |
| 512 | 1.224x |
| 1024 | 1.185x |
| 1536 | 1.162x |
| 2048 | 1.129x |
| 4096 | 0.960x, excluded |
| 8192 | 0.882x, excluded |

All tested shapes pass numerical checks, including exact router top-16
selection.

## End-to-end results

MI355X, TP8, DSpark MTP (`num_speculative_tokens=3`), via the companion vLLM
change. Both arms of each comparison ran back to back on one node with
identical server arguments, KV configuration and client invocation.

### 8K/1K fixed sequence length

ISL 8192, OSL 1024, prefix caching disabled, 610 requests per arm, zero
failures.

| Concurrency | Decode M | Output throughput | Median TPOT |
|---:|---:|---|---|
| 1 | 4 | 130.1 -> 135.9 tok/s, **+4.43%** | 7.15 -> 6.83 ms, **-4.53%** |
| 4 | 16 | 369.9 -> 380.5 tok/s, **+2.87%** | 9.25 -> 8.97 ms, **-2.98%** |
| 8 | 32 | 582.8 -> 606.5 tok/s, **+4.07%** | 11.99 -> 11.31 ms, **-5.66%** |
| 16 | 64 | 799.8 -> 814.3 tok/s, **+1.82%** | 18.00 -> 17.65 ms, **-1.98%** |
| 32 | 128 | 1030.2 -> 1044.7 tok/s, **+1.41%** | 28.95 -> 28.55 ms, **-1.37%** |

Median TTFT moves by at most 0.58% in any cell: prefill in this workload is
8130-8190 tokens per step and never matches a supported shape, so the gain is
decode-only. Repeating two of these rows independently bounds their
reproducibility at 0.13-0.80 pp.

### AgenticX trace replay, 3600 s per cell

| Metric | C1 | C4 |
|---|---|---|
| Total throughput | 10212 -> 10533 tok/s, **+3.14%** | 19211 -> 19617 tok/s, **+2.11%** |
| Output throughput | 84.77 -> 87.75 tok/s, **+3.52%** | 139.20 -> 146.61 tok/s, **+5.33%** |
| P90 TPOT | 8.92 -> 8.47 ms, **-5.03%** | 13.52 -> 13.36 ms, **-1.19%** |
| P90 interactivity | 112.1 -> 118.1 tok/s, **+5.29%** | 73.9 -> 74.8 tok/s, **+1.20%** |

Read the total-throughput row with care: it does not measure this change.
Input tokens are over 99% of all tokens in this trace and 97.6% of them hit the
prefix cache, so total throughput is close to (input tokens handed to this arm)
/ wall clock. Trace replay assigns sessions dynamically and the two arms did
not get an identical token mix -- the optimized arm received 3.04% more input
tokens at C1 and 2.81% more at C4. Dividing the measured total-throughput delta
by each arm's input-token and duration ratio leaves a residual of +0.00% at C1
and +0.02% at C4, so that row is token mix and nothing else.

Output throughput and P90 TPOT are the rows that carry the effect: both are
normalized per output token, so neither depends on how much input the arm
happened to receive. Note that they disagree in shape with the total row at C4,
where output throughput is +5.33% against a total of +2.11% -- the smaller
number there is the noisier one, not the more conservative one.

## Accuracy

GSM8K, full 1319-question test split, 5-shot, greedy. This is the check the
numerical tests structurally cannot make: they compare the fused kernel against
a `torch.mm` reference built from the same merged weight, so they are blind to
the merged front slicing its router logits out of a `7168 -> 6016` accumulator
where the native front runs a separate `7168 -> 896` GEMM. Both are
BF16 x BF16 -> FP32, but the wider shape picks a different solution with a
different accumulation order, and that rounding difference can flip a borderline
top-16 expert selection.

Both arms ran back to back on one node at `max_num_seqs=8` with
`num_speculative_tokens=3` and block rejection sampling, so a saturated decode
step submits 32 tokens and the merged front is the decode path for effectively
the whole run.

| Filter | Baseline | Merged |
|---|---:|---:|
| strict-match | 0.9659 | 0.9719 |
| flexible-extract | 0.9651 | 0.9712 |

Both arms answer the same questions, so the paired view is the informative one:
1295 of 1319 answers agree, 16 flip to correct and 8 flip to incorrect, an exact
two-sided sign test on the 24 discordant pairs giving p = 0.15. The merged front
is therefore indistinguishable from the native front on accuracy. The +0.6 pp is
run-to-run nondeterminism, not a claimed improvement.

## Validation

- Ruff (`ruff==0.16.0`, the version `pre-checks.yaml` pins), `ruff format` and
  `git diff --check`: passed. The two `EXE001` findings that remain in the
  repository are in unrelated flydsl files and predate this PR.
- gfx950 AITER tests: `17 passed`.
- Dependent vLLM Kimi-K3 tests: `59 passed`.

Companion vLLM PR: https://github.com/vllm-project/vllm/pull/55811


jiacao-amd:
### GSM8K accuracy A/B

Adding the accuracy evidence this PR was missing. The numerical tests compare
the fused kernel against a `torch.mm` reference built from the same merged
weight, which is self-consistent and therefore blind to the one failure mode
specific to this change: the merged front emits router logits out of an FP32
accumulator where the native front uses a separate BF16 gate GEMM, and that
rounding difference can flip a borderline top-16 expert selection. Only an
end-to-end score can see it.

Full 1319-question GSM8K test split, 5-shot, greedy. Both arms ran back to back
on one node from the same image, at `max_num_seqs=8` with
`num_speculative_tokens=3` and block rejection sampling, so a saturated decode
step submits 32 tokens and the merged front is the decode path for effectively
the whole run.

| Filter | Baseline | Merged |
|---|---:|---:|
| strict-match | 0.9659 | 0.9719 |
| flexible-extract | 0.9651 | 0.9712 |

Both arms answer the same questions, so the paired view is the informative one
rather than the unpaired +/- 0.005 binomial band: 1295 of 1319 answers agree, 16
flip to correct and 8 flip to incorrect, and an exact two-sided sign test on the
24 discordant pairs gives p = 0.15.

The merged front is indistinguishable from the native front on accuracy. The
+0.6 pp is run-to-run nondeterminism, not a claimed improvement.

aiter/ops/triton/kimi_k3_moe_front.py
```diff
@@ -0,0 +1,346 @@
...
+    )
+
+
+def _large_m_front_gemm(
+    hidden_states: torch.Tensor,
+    merged_weight: torch.Tensor,
+    front_out: torch.Tensor,
+) -> None:
+    config = _large_m_front_gemm_config(hidden_states.shape[0])
+    if config["libtype"] == "hipblaslt":
```
rahulbatra85: This code doesn't belong in this file. Probably in the Kimi-K3 code which is meant to call this Triton kernel

ops/triton folder should have code related to Triton op wrapper only(should be no hipblast lt code in here at all) and the any Triton kernels should go in ops/triton/_triton_kernels

aiter/ops/triton/kimi_k3_moe_front.py
```diff
@@ -0,0 +1,346 @@
...
+        out_dtype=torch.float32,
+    )
+
+
+@triton.jit
+def _tanh(x):
+    return 2.0 * tl.sigmoid(2.0 * x) - 1.0
+
+
+@triton.jit
```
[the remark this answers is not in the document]
rahulbatra85: This code doesn't belong in this file. 

Only op wrapper code should be in ops/triton. The actual kernel should go in ops/triton/_triton_kernels

aiter/ops/triton/kimi_k3_moe_front.py
```diff
@@ -0,0 +1,346 @@
...
+        raise ValueError("All Kimi-K3 front weights must be on the same device")
+
+    return torch.cat(
+        (shared_gate_up, router, routed_down),
+        dim=0,
+    ).contiguous()
+
+
+@functools.lru_cache(maxsize=1)
+def _initialize_hipblaslt() -> None:
```
rahulbatra85: This code should not be here at all. Don't mix hipblas lt code with Triton code.

aiter/ops/triton/kimi_k3_moe_front.py
```diff
@@ -0,0 +1,346 @@
...
+        TILE=tile,
+        SHARED_TILES=shared_tiles,
+        ROUTER_TILES=router_tiles,
+        TOTAL_TILES=total_tiles,
+        num_warps=num_warps,
+    )
+    return shared_out, router_out, routed_out
+
+
+def kimi_k3_moe_front_large_m_bf16(
```
rahulbatra85: Since this ends up calling non-Triton kernel, remove this.

aiter/ops/triton/kimi_k3_moe_front.py
```diff
@@ -0,0 +1,346 @@
...
+_HIDDEN = 7168
+_SHARED_GATE_UP = 1536
+_SHARED_INTERMEDIATE = 768
+_NUM_EXPERTS = 896
+_ROUTED_LATENT = 3584
+_FRONT = _SHARED_GATE_UP + _NUM_EXPERTS + _ROUTED_LATENT
+_DEFAULT_TILE = 512
+
+
+def merge_kimi_k3_moe_front_weights(
```
rahulbatra85: Why is this function in this file? This is specific to the Kimi-K3. Keep this code where the Kimi-K3 model code exists

rahulbatra85 (changes requested):

aiter/ops/triton/kimi_k3_moe_front.py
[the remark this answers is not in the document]
jiacao-amd: Done, relocate the entry kernel.

aiter/ops/triton/kimi_k3_moe_front.py
jiacao-amd: Done, The public Triton wrapper now lives in aiter/ops/triton/moe/moe_front_epilogue.py, and the kernel implementation is under aiter/ops/triton/_triton_kernels/moe/moe_front_epilogue.py.

aiter/ops/triton/kimi_k3_moe_front.py
jiacao-amd: removed

aiter/ops/triton/kimi_k3_moe_front.py
jiacao-amd: removed. The weight-merging logic is model-specific and will be kept in the Kimi-K3 model implementation in the companion vLLM PR

aiter/ops/triton/kimi_k3_moe_front.py
jiacao-amd: removed

aiter/ops/triton/_triton_kernels/moe/moe_front_epilogue.py
```diff
@@ -0,0 +1,91 @@
...
+        "SHARED_TILES",
+        "ROUTER_TILES",
+        "TOTAL_TILES",
+        "NUM_WARPS",
+    ],
+)
+
+
+@triton.jit
+def _tanh(x):
```
rahulbatra85, line 24: Can you use the one from here aiter/aiter/ops/triton/_triton_kernels/activation.py

rahulbatra85:
@jiacao-amd  Is this kernel simply projecting the activations(ie x) coming into the MoE layer with some non-linearity like _tanh? If, yes, can you name it a bit differently to reflect that. Also, I understand that this is specific to Kimi-K3, but don't put Kimi specific shape information in there. Make it generic.

aiter/ops/triton/_triton_kernels/moe/moe_front_epilogue.py
jiacao-amd: removed and use func in _triton_kernels/activation.py

jiacao-amd:
> @jiacao-amd Is this kernel simply projecting the activations(ie x) coming into the MoE layer with some non-linearity like _tanh? If, yes, can you name it a bit differently to reflect that. Also, I understand that this is specific to Kimi-K3, but don't put Kimi specific shape information in there. Make it generic.

No, this kernel does not project x. The projection has already been done before this kernel. 
For naming, I renamed it to moe_situ_epilogue, also removed the hard-coded Kimi-K3 dimensions and made the dimensions caller-provided.

PR #5324, opened 2026-09-08
[FlyDSL] fix(flydsl): zero GDR decode graph padding output

junna2016:
## Motivation

The FlyDSL GDR decode kernel uses negative `read_indices` and `write_indices`
as sentinels for CUDA/HIP Graph padding rows. Previously, the kernel skipped
those rows without writing their output. If a caller allocated `out` with
`torch.empty`, the padding rows retained undefined device-memory contents.

Initializing the complete output with `torch.zeros` in the caller avoids the
correctness issue, but adds a separate memset operation to every decode Graph
replay. This fixed overhead is especially significant for low-batch decode and
can erase much of the fused-kernel performance improvement.

This PR makes the fused kernel itself define the padding-row behavior so callers
can safely use an uninitialized output buffer without an additional kernel.

## Technical Details

- Add a negative-index branch to the FlyDSL GDR decode kernel.
- For a valid pair of state indices, preserve the existing behavior:
  - read the recurrent state;
  - compute the GDR decode result;
  - write the output and destination state.
- If either state index is negative:
  - write positive zero to every element of that batch row's output;
  - do not read from or write to the recurrent-state pool.
- Perform the padding write inside the existing fused kernel. This introduces no
  additional kernel launch or full-output memset.
- Expose `flydsl_gdr_decode.zeroes_invalid_output = True` as a capability marker.
  Integrations can use this marker to select `torch.empty` while retaining a
  safe `torch.zeros` fallback for older AITER versions.

The zero-store work is confined to workgroups assigned to padding rows. The
valid-request path and its generated kernel computation remain unchanged.

## Test Plan

- Add a mixed valid/padding test with:
  - BF16 query/key/value and gate inputs;
  - FP32 recurrent state;
  - `read_indices = [1, -1, 3, -1]`;
  - `write_indices = [2, -1, 4, -1]`;
  - an output buffer prefilled with NaNs.
- Verify that:
  - all valid output rows are written;
  - every padding output bit represents positive zero;
  - state-pool slots not targeted by valid requests remain bit-identical.
- Run the existing strided-input and independent read/write-index parameterized
  coverage.
- Compare the fused operator against the original two-kernel Triton decode path
  under HIP Graph replay.
- Validate Qwen3.5-4B Graph-on inference using fixed 2K/8K prompts and a
  concurrent mixed valid/padding workload.

### Correctness

- AITER focused tests: **15 passed**.
- Valid-row maximum differences against the Triton reference:
  - output: `1.2207e-4`;
  - recurrent state: `2.3842e-7`.
- Before the fix, every one of the 4096 elements in each NaN-prefilled padding
  row remained NaN, confirming that those outputs were undefined.
- With three concurrent model requests, the old `torch.empty` plus skipped-row
  behavior corrupted 2/3 responses. After this fix, all 3/3 responses were
  normal and complete.
- Sequential model comparison after the fix:
  - 2K prompts: 8/8 matched the safe zero-initialized path;
  - 8K prompts: 20/20 matched the safe zero-initialized path.

### Performance

Qwen3.5-4B production decode shape: `T=1`, `KH=16`, `VH=32`, `K=V=128`,
BF16 inputs, FP32 recurrent state, HIP Graph replay.

The Triton baseline below is the original RTP-LLM decode implementation. Its
Graph contains RTP-LLM's separate `fused_gdn_gating_kernel` and
`fused_recurrent_gated_delta_rule_fwd_kernel`; it is not an AITER Triton
baseline. The FlyDSL result uses AITER's single fused GDR decode kernel.

| Batch | RTP-LLM Triton baseline | AITER FlyDSL with fused padding zero | Speedup |
|---:|---:|---:|---:|
| 1 | 25.000 us | 19.400 us | 1.289x |
| 64 | 149.041 us | 129.641 us | 1.150x |
| 128 | 296.562 us | 224.522 us | 1.321x |
| 256 | 584.204 us | 447.223 us | 1.306x |
| 512 | 1115.566 us | 884.045 us | 1.262x |

In a Qwen3.5-4B 8K-input/100-output, concurrency-1 model benchmark, mean TPOT
improved from `6.7522 ms` to `6.5496 ms` (**3.00% reduction**).

## Submission Checklist


xytpai:
> **From agent — Recommendation: Request changes**
>
> The kernel change looks correct in the configurations I tested: 17 focused tests and 32 additional graph-replay checks passed on gfx950. I did not find a new kernel correctness defect.
>
> Before merging, please wire the new padding regression into the standard CI execution path. The current script-based entry point does not execute it, leaving the original bug unprotected in that job.
>
> Once that is addressed and CI passes, I would recommend approval. The additional asymmetric-index and valid-row coverage comments are non-blocking suggestions.

xytpai (approved):

coderfeli (approved):

PR #5334, opened 2026-09-08
[HIP] [Perf] topk_gating: add E=512 prefill_n dispatch for softmax

mjkvaak-amd:
prefill_n already instantiates E=64/128/256/384 (TPW=1) and E=640 (TPW=1/2). E=512 was the gap, so Qwen3.8 (512 experts,  top-10, softmax) currently falls through to the shared-memory fallback.

Add `_DISPATCH_PREFILL_N_KERNEL(512, 1)`, gated to `num_tokens<=1024` the same way E=640 is gated to 2048. No new kernel.

gfx950 / MI355X, E=512, softmax, need_renorm=true, stock smem vs this dispatch:

| topk | dtype | T<=1024 | T=2048 / 4096 control cases (gate off) |
|------|-------|---------|--------------------------|
| 10   | bf16  | -40 to -42% | -1.3% / +2.3% |
| 2..16 | bf16/fp32 | -36 to -51% | within ~5% |

Routing ids match; weight error <= 3e-08. TPW=2 not added (AITER's E=640 TPW=2 path is `topk<=8`, so it would not apply at k=10).

Tested: `python3 op_tests/test_moe_topk_gating.py --num-experts 512 --topk 2,8,10,16 --num-tokens 1,16,256,1024,2048,4096 -d bf16 fp32 --section softmax`


aiter/ops/topk.py
```diff
@@ -28,6 +28,9 @@ def topk_gating_fwd(
 
 
 _VALID_SCORE_FUNCS = {"sqrtsoftplus", "sigmoid", "softmax"}
+# Downstream (vLLM) refuses the softmax fast path unless this is True:
+# older AITER accepted need_renorm for softmax but silently forced it off.
+TOPK_GATING_SOFTMAX_RENORMALIZES = True
```
yzhou103: TOPK_GATING_SOFTMAX_RENORMALIZES — the pattern is fine, #5143 established it. But could you link the vLLM PR that reads it? There's no consumer in this repo, and once it's public it can't be removed. Two nits if it stays: export it explicitly
  in aiter/__init__.py like #5143 does (it currently arrives via from .ops.topk import *, which breaks silently if topk.py ever gets an __all__), and consider TOPK_GATING_SUPPORTS_SOFTMAX_RENORM to match the existing <OP>_SUPPORTS_<FEATURE>
  convention.

mjkvaak-amd: Thanks for the review. The linked consumer is https://github.com/vllm-project/vllm/pull/56226 (draft, gated on this flag).

Renamed to `TOPK_GATING_SUPPORTS_SOFTMAX_RENORM` and re-exported it explicitly from `aiter/__init__.py` the same way #5143 does.

mjkvaak-amd:
@yzhou103 : there are two failing triton tests that don't appear to be connected with this PR. Otherwise the CI looks green.

yzhou103 (dismissed):

aiter/ops/topk.py
```diff
@@ -28,6 +28,10 @@ def topk_gating_fwd(
 
 
 _VALID_SCORE_FUNCS = {"sqrtsoftplus", "sigmoid", "softmax"}
+# Downstream (vLLM) refuses the softmax fast path unless this is True:
+# older AITER accepted need_renorm for softmax but silently forced it off.
+# https://github.com/vllm-project/vllm/pull/56226
+TOPK_GATING_SUPPORTS_SOFTMAX_RENORM = True
```
valarLip: remove this one

aiter/__init__.py
```diff
@@ -122,6 +122,7 @@ def getLogger():
     from .ops.communication import *
     from .ops.rope import *
     from .ops.topk import *
+    from .ops.topk import TOPK_GATING_SUPPORTS_SOFTMAX_RENORM  # noqa: F401
```
valarLip: .

valarLip (dismissed):

zufayu:

**Adds a compile-time branch to the topk_gating kernel selector so 512-expert softmax routing with at most 1024 tokens uses the existing register-only prefill_n kernel instead of falling through to the shared-memory fallback, which is the path Qwen3.8-style routers (E=512, top-10, softmax) took before this change.**

Validation (deterministic): NOT RUN — PR changes runtime code but ships no test target (auto-validation disabled), and the changed path (the csrc/include/topk_gating_kernels.cuh dispatch, executed at run time) makes the missing target a finding in its own right, reported below; correctness rests on the author's manual sweep of the existing parameterized op_tests/test_moe_topk_gating.py (E=512, k=2/8/10/16, T=1..4096, bf16/fp32, softmax: routing ids match, weight err <= 3e-08), which I independently re-ran head-side on gfx942 — all 96 configs pass vs the torch reference (err=0, max weight err 5.96e-08)
Perf (advisory): MEASURED — E=512 softmax, k=2/8/10/16, bf16/fp32, T=1/16/256/1024/2048/4096 (96 configs, 101 iters each after 2 warmup), base vs head back to back on one idle MI300X (gfx942): head 1.22-2.35x faster at T<=1024 (Qwen3.8 shape k=10 bf16 T=1: base 13.3us vs head 7.1us, -46%) and 0.997-1.003x at T>1024 where both sides run the identical smem path (run-to-run noise floor); reproduces the author's gfx950 direction and magnitude (-36 to -51% inside the gate)

⚠️ [verified] The E=512 prefill_n branch added at csrc/include/topk_gating_kernels.cuh:1439 is unreachable by CI: op_tests/test_moe_topk_gating.py's default --num-experts sweep is [64,128,256,384] and .github/scripts/aiter_test.sh runs every op_tests file with no arguments, so the gfx942/gfx950 CI loops never enter the new dispatch — and the PR ships no test of its own, so a later edit breaking the E=512 path (gate, template, or launch) would pass CI green. **Author must** add 512 (and 640, which has the same gap) to the test's default --num-experts sweep so CI exercises the branch.

📝 [inferred] The new branch in csrc/include/topk_gating_kernels.cuh carries no arch guard, so on gfx1250 (wave32) it fires with VPT=16 — the kernel family's own documented effectiveness boundary ("Effective when VPT <= 16") — while every number behind the gate was measured on gfx950. If the doubled per-thread register footprint (row_chunk[16]+row_orig[16]+is_nan[16]) spills on RDNA4, E=512 softmax at T<=1024 could regress there vs the smem fallback. **Reviewer should ask** whether gfx1250 needs a measurement or a gate before 512-expert models run on it.

mjkvaak-amd:
> 
> **Adds a compile-time branch to the topk_gating kernel selector so 512-expert softmax routing with at most 1024 tokens uses the existing register-only prefill_n kernel instead of falling through to the shared-memory fallback, which is the path Qwen3.8-style routers (E=512, top-10, softmax) took before this change.**
> 
> Review (advisory): ⚠️ NEEDS WORK Validation (deterministic): NOT RUN — PR changes runtime code but ships no test target (auto-validation disabled), and the changed path (the csrc/include/topk_gating_kernels.cuh dispatch, executed at run time) makes the missing target a finding in its own right, reported below; correctness rests on the author's manual sweep of the existing parameterized op_tests/test_moe_topk_gating.py (E=512, k=2/8/10/16, T=1..4096, bf16/fp32, softmax: routing ids match, weight err <= 3e-08), which I independently re-ran head-side on gfx942 — all 96 configs pass vs the torch reference (err=0, max weight err 5.96e-08) Perf (advisory): MEASURED — E=512 softmax, k=2/8/10/16, bf16/fp32, T=1/16/256/1024/2048/4096 (96 configs, 101 iters each after 2 warmup), base vs head back to back on one idle MI300X (gfx942): head 1.22-2.35x faster at T<=1024 (Qwen3.8 shape k=10 bf16 T=1: base 13.3us vs head 7.1us, -46%) and 0.997-1.003x at T>1024 where both sides run the identical smem path (run-to-run noise floor); reproduces the author's gfx950 direction and magnitude (-36 to -51% inside the gate)
> 
> ⚠️ [verified] The E=512 prefill_n branch added at csrc/include/topk_gating_kernels.cuh:1439 is unreachable by CI: op_tests/test_moe_topk_gating.py's default --num-experts sweep is [64,128,256,384] and .github/scripts/aiter_test.sh runs every op_tests file with no arguments, so the gfx942/gfx950 CI loops never enter the new dispatch — and the PR ships no test of its own, so a later edit breaking the E=512 path (gate, template, or launch) would pass CI green. **Author must** add 512 (and 640, which has the same gap) to the test's default --num-experts sweep so CI exercises the branch.
> 
> 📝 [inferred] The new branch in csrc/include/topk_gating_kernels.cuh carries no arch guard, so on gfx1250 (wave32) it fires with VPT=16 — the kernel family's own documented effectiveness boundary ("Effective when VPT <= 16") — while every number behind the gate was measured on gfx950. If the doubled per-thread register footprint (row_chunk[16]+row_orig[16]+is_nan[16]) spills on RDNA4, E=512 softmax at T<=1024 could regress there vs the smem fallback. **Reviewer should ask** whether gfx1250 needs a measurement or a gate before 512-expert models run on it.


Thank you for validating the proposed fix for MI300x. To avoid the spill risk for gfx1250 / wave32 (VPT = 512/32 = 16), we could tighten the gate while still keeping gfx942 & gfx950:

```c
if(num_tokens <= 1024 && get_warp_size_func() != 32)
{
    _DISPATCH_PREFILL_N_KERNEL(512, 1)
}
```

Alternatively, we can use `get_gpu_arch()` to gate this to gfx942 and gfx950 only. What do you think @zufayu?

I also added `-num-experts=[..., 512, 640]` in the test, so a no-arg CI run hits both the new E=512 branch (at T≤1024) and the existing E=640 gap you called out.

zufayu:
ci failed

zufayu (approved):

zufayu:
Hi @valarLip, the requested changes have been addressed and the related threads are resolved. Could you please re-review the latest version?

PR #5340, opened 2026-09-08
[HIP] [CI] [JIT] [Feature] Add gfx:cu_num build targets so one build serves multiple SKUs of an arch

RElbers:
## Motivation

AITER selects tuned GEMM kernels at build time using `(gfx, cu_num, ...)`. GPUs sharing an architecture can have different CU counts. For example, `gfx950` with 128 CUs on MI350P and 256 CUs on MI350X/MI355X. `GPU_ARCHS` and a single global `CU_NUM` cannot request both configurations in one build. This PR adds `AITER_GPU_TARGETS="gfx950:128;gfx950:256"` so the existing code-generation path can include tuned kernels for both targets. It provides the target-selection mechanism; corresponding tuning data is still required.

`AITER_GPU_TARGETS` is optional. Builds that leave it unset continue to use `GPU_ARCHS`, `CU_NUM`, or native detection, but this PR also changes some of their behavior:

- For explicitly named `GPU_ARCHS` targets, omitting `CU_NUM` previously selected each architecture's default CU count; it now uses the live device's count for the matching architecture, retaining defaults for other targets or when detection fails.
- For multiple named architectures, `get_gfx()` previously selected the last entry; it now selects the detected architecture if included, otherwise the lexicographically greatest name, making the fallback independent of list order.
- Explicit `GPU_ARCHS` builds previously accepted zero and negative `CU_NUM` overrides; now only positive integers are accepted.
- `_parse_gpu_archs_env` previously preserved architecture-name case; it now normalizes names to lowercase, so entries such as `GFX942` resolve as `gfx942`.

## Build target resolution

| `AITER_GPU_TARGETS` | `GPU_ARCHS` | `CU_NUM` | Build targets |
| --- | --- | --- | --- |
| `gfx950:128;gfx950:256` | ignored | ignored | `gfx950:128`, `gfx950:256` |
| `gfx942;gfx950:128` | ignored | ignored | `gfx942:304`, `gfx950:128` |
| unset | `gfx942;gfx950` | unset | `gfx942:304`, `gfx950:256`[^cu] |
| unset | `gfx942;gfx950` | `80` | `gfx942:80`, `gfx950:80` |
| unset | `gfx950` | unset, no GPU | `gfx950:256` |
| unset | unset or `native` | unset | live architecture at its live CU count |
| unset | unset or `native` | `80` | live architecture at `80` |
| unset | unset or `native` | any, no GPU | error |

[^cu]: The architecture matching the live device takes that device's CU count.

`CU_NUM` applies to every architecture named in `GPU_ARCHS`. A bare `AITER_GPU_TARGETS` entry takes its
architecture's default count. When `AITER_GPU_TARGETS` is set, a `GPU_ARCHS` naming a different architecture set warns. A `CU_NUM` matching none of the targets warns.

## Implementation/Changes

- **Target parsing and selection.** `build_targets.py` parses `AITER_GPU_TARGETS` into `(gfx, cu_num)` pairs, validates architecture names and positive CU counts, and removes exact duplicates. Entries without a CU count use the architecture default. These pairs replace `GPU_ARCHS`/`CU_NUM` for build-target selection and feed the existing CSV filtering, so two CU counts of one architecture can both contribute tuned rows.

- **Architecture validation and gfx1250 flags.** `jit/core.py` and `cpp_itfs/utils.py` derive compiler architecture lists from the new targets when set. Both validators use `KNOWN_GFX` instead of separate allowlists. The gfx1250 compiler flags and cluster-launch define check membership in `get_gfx_list()` instead of reading `GPU_ARCHS` directly; cluster launch still requires ROCm 7.0 or later.

- **Template-library cache keys.** C++ template-library cache paths now include the sorted, deduplicated architecture set. Different sets use separate caches; `gfx950;gfx942`, `gfx942;gfx950`, and equivalent lists with duplicates share one. Paged-attention AOT temporary files and locks use the same key. Existing template libraries rebuild under the new paths on first use. The main JIT module cache is unchanged and does not distinguish CU counts.

- **HSACO and MLA assembly paths.** HSACO cache reads and default writes use `get_gfx_runtime()`, as does MLA assembly lookup. This replaces paths based on the raw `GPU_ARCHS` value, which could contain multiple architecture names.

- **Warnings and documentation.** Build-target resolution reports when no live CU count can be read, and warns when `CU_NUM` requests a count absent from `AITER_GPU_TARGETS`. CSV filtering warns about targets with no rows for the selected library type; those targets continue to use default kernels. The installation guide documents the new variable and that `CU_NUM` still overrides runtime lookup.

- **Release and nightly workflows.** The reusable release workflow accepts a `gpu_targets` input and passes it into the build containers as `AITER_GPU_TARGETS`. Automated release and default nightly builds specify `gfx942:304;gfx950:256`, so their target CU counts do not depend on the build host.

## Out of scope

OPUS and FlyDSL AOT target-filtering integration and adding the gfx950:128 tuning rows is a separate effort and is out of scope here. This is the `AITER_GPU_TARGETS` mechanism only. Both are addressed in the follow-ups below.

## Test Plan

- Extend GEMM code-generation tests to cover multiple CU counts for one architecture, target deduplication, default counts, environment precedence, and reporting targets without tuned rows.
- Add GPU target-resolution tests with the live GPU probes mocked. Cover live CU detection for a binned part, order-independent architecture selection, gfx1250 compiler flags, and cache separation.
- Run the existing GEMM dispatch/code-generation and JIT-cache transaction regression suites.

## Test Result

Verified on MI300X and MI355X :

- `python op_tests/test_gemm_codegen.py`: 64 checks passed, 0 failed.
- `python -m pytest -q op_tests/test_gpu_target_resolution.py op_tests/test_jit_cache_transaction.py`: 56 tests passed, 0 failed.

## Follow-up PRs

 * #5668, to key the opus MXFP8 BMM lookup on CU count.
 * #5342 to filter FlyDSL GEMM AOT jobs by build target.


samremes:
## Review (SKU expression / non-comm enablement)

This is the build-time half of [SILOTIGER-971](https://amd.atlassian.net/browse/SILOTIGER-971) / [SILOTIGER-978](https://amd.atlassian.net/browse/SILOTIGER-978): `AITER_GPU_TARGETS=gfx950:128;gfx950:256` is the right knob. Not adding `gfx950:128` rows here is correct ([SILOTIGER-972](https://amd.atlassian.net/browse/SILOTIGER-972) comes after dispatch is safe). Runtime XCC identity stays in #5182; this PR should key tuning on CU count.

The exact `(gfx, cu_num)` filter and the release pin `gfx942:304;gfx950:256` match the plan. The `get_gfx()` last-entry-wins fix on multi-arch wheels is necessary.

**Request changes** on one P1 before merge:

1. **Runtime arch still is not the active HIP device.** `_detect_native` takes the first `gfx` token from `rocminfo`. For a multi-arch wheel this replaces last-entry-wins with first-`rocminfo`-wins. On a mixed host that is not the device that will launch. Resolve from the current HIP device (same direction as #5182's per-device XCC/PCI queries). Add a test where the first enumerated arch differs from the active device.

Also worth fixing in this PR (P2):

2. **Bare `GPU_ARCHS` live-CU probe:** a failed `get_cu_num()` / `get_gfx_runtime()` falls through a broad `except` to the full-SKU table default with no warning. That is indistinguishable from a GPU-less cross-build and can bake 256-CU identity on a 128-CU box.

3. **Precedence vs `CU_NUM`.** Implementation lets `AITER_GPU_TARGETS` win over `CU_NUM`; `docs/installation.rst` only says it overrides `GPU_ARCHS`. Same-arch conflict `AITER_GPU_TARGETS=gfx950:256` + `CU_NUM=128` does not warn.

4. **Tests miss the new validation surface.** Unknown-arch still goes through `GPU_ARCHS`; no case for malformed `AITER_GPU_TARGETS`, live-CU substitution, probe failure, or the unmatched-target warning after `libtype` filtering.

Merge order: land this (after the P1) before #5341 / #5342.

[SILOTIGER-971]: https://amd.atlassian.net/browse/SILOTIGER-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SILOTIGER-978]: https://amd.atlassian.net/browse/SILOTIGER-978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SILOTIGER-972]: https://amd.atlassian.net/browse/SILOTIGER-972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #4735, opened 2026-08-13
[Bug]: gfx950 a16w16 bf16 tuned CSV hardcoded to cu_num=256, crashes serving when corrected (same pattern as #4634)


sscargal:
Related: #4634 describes what looks like the same two-part defect (CU-count-hardcoded tuned CSV +
CSV/kernel-catalog registration drift) on `gfx942`/MI300X for `a8w8_blockscale_bpreshuffle`. This
issue reproduces the same pattern independently on `gfx950`/MI350P for `a16w16` bf16 GEMM, filed
separately since the architecture and op family differ — happy for a maintainer to close this as a
duplicate of #4634 if you'd rather track it there, or keep it separate if the `a16w16`/gfx950
specifics are useful on their own.

### Environment

- GPU: AMD Instinct MI350P, `gcnArchName=gfx950`, `NUM_COMPUTE_UNITS=128` (`cu_num=128`)
- Container image: `rocm/atom-dev:latest`, digest `sha256:381fddcbba6e0c43a6a8f7008be2b0b3c0b817d77223b513a661a6d449c2bf90`
- AITER commit (in-image, `aiter/`): `65d3e921a8283606dee74b000bb0da084ab10251` (2026-08-05)
- `torch.__version__`: `2.10.0+rocm7.2.4.git3d3aa833` (build tag; `rocprofv3 --version` on the same host reports ROCm 7.14.0 — noting both as observed, not reconciled)
- Model: `amd/Kimi-K2-Instruct-0905-MXFP4`, served via ATOM at TP8 (8 GPUs)
- Env vars in effect: `AITER_QUICK_REDUCE_QUANTIZATION=INT4`, `AITER_MXFP4_INTERMEDIATE=1`, `AITER_AR_1STAGE_MAX_KB=512`, `AITER_USE_FLYDSL_MOE_SORTING=1`

### Bug 1: `cu_num` mismatch hides the entire tuned file on this hardware

`aiter/configs/model_configs/kimik2_bf16_tuned_gemm.csv` has 96 `gfx950` rows, **all hardcoded at
`cu_num=256`** (matches MI355X; 31 additional rows in the same file are `gfx942`-class). This
host's MI350P is also `gfx950`, but reports `cu_num=128` — since the tuned-config lookup
(`get_GEMM_A16W16_config()`) keys on `(gfx, cu_num, M, N, K)` (mirroring `get_CKGEMM_config()`'s
[...]

PR #5341, opened 2026-09-08
[HIP] [OPUS] [JIT] [Bugfix] Stop an unbaked tuned kernel id from killing the rank, and key the lookup on CU count

RElbers:
Stops a tuned row naming a kernel id this build does not contain from killing the rank. The id dispatcher now raises a distinct type, `tuned_gemm` catches it and falls back to torch, and the shape is remembered so a serving loop pays for the miss once. That is a prerequisite for loading a 128-CU CSV in serving at all (ROCm/aiter#4735 bug 2). Keys the opus a16w16 GEMM lookup on CU count, and starts dispatching gfx1250's tuned split-K kernels, which the build already contained but nothing ever selected.

Stacked on #5340, which introduces `AITER_GPU_TARGETS`: build targets written as `gfx:cu_num` pairs, as in `AITER_GPU_TARGETS="gfx950:128;gfx950:256"`, and the build and runtime target resolution behind them. The opus codegen here resolves its rows and its compile set through those targets, which is what lets one build carry a tuned winner per SKU.

## Motivation

**The lookup key had no CU count.** Tuned rows are keyed `(gfx, cu_num, ...)`, but the generated C++ tables were keyed `(M, N, K)` per arch. A shape tuned on two SKUs of one arch kept whichever row was read last, and the other SKU silently fell back to its heuristic. One build could never hold both. No checked-in CSV carries two CU counts for one arch today, so this is a fix ahead of the need.

**gfx1250's split-K winners were unreachable.** `opus_gemm.cu` resolved every gfx1250 shape through `opus_dispatch_a16w16_gfx1250<fp32_t>`, and rows bucket into tables by output dtype. Every checked-in gfx1250 opus row is bf16-output, so the fp32 table it consulted was empty and the 150 baked split-K winners never dispatched through it. The 147 `.co` winners did dispatch, through the separate `.co` lookup ahead of it.

**A tuned row naming a kernel id the build lacks killed the rank.** The CSV ships every target's winners; a build bakes only its own, so the two disagree whenever a package is used on a target it was not built for.

## Changes

One shared key type and search (`csrc/opus_gemm/include/opus_gemm_lookup_entry.cuh`) replaces three per-arch comparator copies. The ladder is exact CU, then a legacy `cu_num=0` row, then a miss so the caller reaches its heuristic, which is what the Python lookup already does when no row matches the live count. `allow_fallback` gating keeps a count-less `.co` row from shadowing an exact-CU split-K winner, which is the rule `opus_gemm.cu` depends on.

Codegen filters the emitted rows and the compile set through one target resolution, so the kid set and the lookup tables cannot disagree about which arch the build serves. The generated `OPUS_GFX1250_WS_PARTIAL_BYTES` is static-asserted against the bf16 workspace the host allocates, so a baked kid storing a wider partial fails the build rather than overrunning the buffer.

## Scope

1. **gfx1250 split-K shapes move from heuristic to tuned, and the `.co` shapes become CU-conditional.** No CI job exercises gfx1250 (`GPU_ARCH_LIST` in `aiter-test.yaml` is `gfx942;gfx950`). The A/B is cheap for the split-K half, since the old path is today's heuristic. The `.co` half is the exposure: that lookup was shape-only and always hit, it now goes through the CU-aware search, and every `.co` row is stamped 256, so on silicon reporting any other count those 147 shapes fall to the heuristic instead.
2. **`aiter/tuned_gemm.py` is the one non-opus file here, and it is the live serving route.** `tuned_gemm.opus_gemm` called the id dispatcher unguarded, so an unbaked id escaped and killed the rank. The dispatcher now raises `unbaked_kernel_error`, registered through pybind as `UnbakedKernelError`; `try_opus_gemm_a16w16_tune` catches it by type, returns False, and `tuned_gemm` falls back to `torch_gemm`, which is what the same function already does when opus is unavailable. A split-K workspace OOM is a `RuntimeError` too but not that type, so it still propagates. The `(id, splitK)` is remembered, because the dispatcher writes to stderr before it raises and a serving loop pays for that once.

## Testing

`op_tests/test_gemm_codegen.py`, `op_tests/test_opus_serving_fallback.py` and `op_tests/opus/test_opus_basic` all pass, the last building and running on a host with no GPU. One new codegen case, `test_opus_bakes_both_skus`, covers per-SKU baking, both SKUs keeping their own winner, and mixed-schema CSVs with a NaN `cu_num`. The C++ cases cover exact-CU selection and the fallback ladder, including the miss that ends it, and `test_opus_serving_fallback.py` covers each serving-degradation path. `ruff` and `black` are clean.


samremes:
## Review (opus lookup consumer / non-comm enablement)

This is the opus consumer of #5340 for [SILOTIGER-971](https://amd.atlassian.net/browse/SILOTIGER-971) / [SILOTIGER-976](https://amd.atlassian.net/browse/SILOTIGER-976). CU belongs in the lookup key, the shared `opus_gemm_lookup_entry` type is the right cleanup, and the unbaked-kid guard is real value: four keys in `dsv4_bf16_tuned_gemm.csv` already resolve to gfx1250 kid 21177 on any 256-CU device, which a gfx950 build does not contain. Keep this draft until #5340 lands.

**Request changes** on the lookup ladder before merge (P1 vs SILOTIGER-971):

The last rung returns the shape's **lowest baked `cu_num`** to any device (`opus_lookup_find` with `allow_fallback=true`). On `AITER_GPU_TARGETS=gfx950:128;gfx950:256`, a 128-CU part with only a 256-CU row for that shape **gets the 256-CU winner**. SILOTIGER-971 says never select a `cu_num=256` row on 128 CU; miss must go to the default/heuristic kernel. Python `lookup_tuned` already keys on live CU with no fallback, so C++ and Python disagree. The `cu_num=0` "legacy" rung is also dead: codegen sorts `(M,N,K,cu_num)` ascending, so a zero row *is* `it`. Outcome to reach: exact CU or `cu_num=0` only; otherwise heuristic. If cross-SKU fallback is deliberate, it needs an opt-in and a log that names the substituted count.

Also P2:

1. **SILOTIGER-976 is still open on a gfx1250-only build.** The guard matches `"tune lookup table"`. The three `dispatch unavailable for this arch` `AITER_CHECK`s do not carry that substring, so a gfx950 kid selected on a gfx1250-only wheel still kills the rank. Root cause: runtime CSV key omits `gfx`, and gfx950/gfx1250 opus rows share `cu_num=256` (the four colliding keys above). Put `gfx` in `_KEY_COLUMNS` / `_key_from_runtime`, and signal "kid not in this build" structurally rather than by message text.

2. **gfx1250 `cu_num` is unconfirmed.** Opus rows and `GFX_CU_NUM_MAP` say 256; `bf16_tuned_gemm.csv` triton rows say 96. If the device reports 96, the exact-CU rung misses and gfx1250 "reachability" depends on the P1 rung. Confirm `multiProcessorCount` and stamp one count.

3. **Split-K workspace guard skips the fuse family** (`a16w16_clusterlaunch_tdm_splitk_fuse`), which is the launcher with no runtime `element_size` check. Cover every baked gfx1250 kid that consumes that host workspace.

4. **`_filter_opus_df_for_targets` matches `gfx` two ways** (`filter_tune_df` exact vs `str.lower()` on mixed schema). Normalize once.

P3 worth a look: per-device CU cache in `opus_get_device_cu_num` is now load-bearing; serving `tuned_gemm.opus_gemm` degrades to `torch_gemm` while `gemm_a16w16_opus` still has the opus heuristic; `_arch.py` swallows `AITER_GPU_TARGETS` typos as "unsupported device".

## Series (with #5182 / #5340 / #5342)

| PR | Role vs SILOTIGER-971 |
|---|---|
| #5182 | Live XCD for swizzle (SILOTIGER-975). Kernel `NUM_XCDS=8` remains SILOTIGER-977. |
| #5340 | Build can name `gfx950:128;gfx950:256` (SILOTIGER-978 mechanism). Fix active-HIP-device `get_gfx()` first. |
| #5341 | Opus lookup/miss path (this PR). Last rung fights SILOTIGER-971; miss-path substring fights SILOTIGER-976. |
| #5342 | FlyDSL GEMM AOT filter. Packaging should use the same resolved `(gfx, cu_num)` as CK, not arch-wide `GPU_ARCHS`. |

Not in these PRs: 128-CU retune ([SILOTIGER-972](https://amd.atlassian.net/browse/SILOTIGER-972)), launch heuristics (977), collectives ([SILOTIGER-974](https://amd.atlassian.net/browse/SILOTIGER-974)).

[SILOTIGER-971]: https://amd.atlassian.net/browse/SILOTIGER-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SILOTIGER-976]: https://amd.atlassian.net/browse/SILOTIGER-976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SILOTIGER-972]: https://amd.atlassian.net/browse/SILOTIGER-972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

zufayu:

**Makes the opus id-dispatchers raise a dedicated UnbakedKernelError that tuned_gemm catches and answers with a remembered torch fallback, and re-keys the generated opus a16w16 lookup tables on (M, N, K, CU count) so one build can carry a tuned winner per SKU of one arch.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0); the added target op_tests/test_opus_serving_fallback.py was run by hand on the full head tree (3 passed), op_tests/opus/test_opus_basic was built and run on this gfx942 host (lookup tests passed), and the pybind exception chain was verified with a minimal TU built against csrc/include/aiter_hip_common.h, but no head-matched report exists
Perf (advisory): NOT RUN — the target ships no benchmark entry point, and the perf-relevant changes are gfx1250-only while this host is gfx942/304 where no checked-in opus row matches (the gfx942 opus rows are all 80-CU), so the tuned path cannot be hand-timed here

⚠️ [verified] The fix covers the a16w16 route only: aiter/ops/gemm_op_a8w8.py (unchanged by this PR) still dispatches libtype=="opus" rows straight into opus_gemm_a8w8_blockscale_bpreshuffle_tune with no catch, while this PR made the gfx942 a8w8 id dispatcher (opus_a8w8_tune_dispatch_gfx942, csrc/opus_gemm/include/gfx942/opus_gemm_arch_gfx942.cuh) raise the new unbaked_kernel_error — so a tuned a8w8 row naming a kid this build lacks still propagates uncaught out of gemm_a8w8_blockscale_bpreshuffle, the same rank-killing escape the PR title says it stops. No checked-in a8w8 CSV carries an opus row today (verified: 0 opus rows in all six a8w8_blockscale_bpreshuffle CSVs), so the exposure is latent until the first a8w8 opus CSV ships. **Author must** either route the a8w8 opus branch through the same remember-and-fall-back wrapper (falling back to the ck path), or state on the PR that the a8w8 route is deliberately out of scope.
⚠️ [verified] The whole gfx1250 dispatch ladder in csrc/opus_gemm/opus_gemm.cu is rewired (exact-CU .co, then exact-CU split-K, then legacy .co, then heuristic) and the .co lookup becomes CU-conditional, but no CI job builds or runs opus on gfx1250 (.github/workflows/aiter-test.yaml pins GPU_ARCH_LIST to "gfx942;gfx950") and the PR reports no hardware run; the PR body itself discloses that every .co row is stamped cu_num=256, so on gfx1250 silicon reporting any other CU count the 147 .co shapes fall to the heuristic, and the 150 split-K winners that just became reachable have never served a request anywhere. **Reviewer should ask** for one gfx1250-device run of the tuned shapes, or confirmation that every supported gfx1250 SKU reports 256 CUs, before merge.
⚠️ [verified] No base-vs-head timing exists for a dispatch change whose gfx1250 half is perf-motivated (150 winners moving heuristic to tuned, plus the .co exposure above): the added test target ships no benchmark entry point, auto-validation is disabled, and this gfx942/304 host matches no checked-in opus row, so the change also cannot be hand-timed here. **Author must** attach a gfx1250 (and ideally a multi-SKU gfx950) before/after timing for the tuned shapes, or mark the perf expectation unverified on the PR.
📝 [verified] test_opus_serving_fallback.py fakes the exception type (_FakeUnbaked stands in for the pybind-registered UnbakedKernelError), so no CI test exercises the real chain — C++ unbaked_kernel_error, pybind translation, Python isinstance catch; a future rename of the module or the exception would pass CI green. The chain was verified out-of-band here with a minimal pybind TU (the registered type isinstance-checks as RuntimeError and as itself; a plain runtime_error does not), but that verification is not in the repo. **Reviewer should ask** for one GPU test that triggers a real unbaked id through the built module.
📝 [verified] aiter/tuned_gemm.py is a Tier-2 backbone file whose documented failure mode is exactly "silent fallback to slow path", and this PR converts a rank-fatal unbaked id into a one-time-warning torch fallback there; the PR author has no commits to that file (top committers yzhou103, valarLip). **Reviewer should ask** the de-facto owner to acknowledge the crash-to-fallback semantics change on the PR before merge.

RElbers:
#4961 made most of this obsolete

PR #5342, opened 2026-09-08
[FlyDSL] [Feature] Precompile only the GEMM kernels the wheel is built for

RElbers:
Stops a released wheel from precompiling FlyDSL GEMM kernels for architectures it is not built for. The packaging path compiled every job in the tuned CSVs regardless of the wheel's targets; it now filters by build target, the way the standalone entry point already did.

Stacked on #5340, which introduces `AITER_GPU_TARGETS`: build targets written as `gfx:cu_num` pairs. The filter here resolves through that machinery, so packaging selects AOT jobs the way the rest of the build selects tuned rows.

## Motivation

`setup.py` → `run_aot()` → `_collect_aot_jobs_for()` called `collect_aot_jobs()` with no target filter, so approximately 220 gfx1250 kernels were built into every gfx942/gfx950 wheel, which never reaches them. The same gap shows up on MI350P: FlyDSL dispatch keys on `(gfx, cu_num)` from the live device, and every tuned FlyDSL row records 256 CUs, so on a 128-CU part every lookup misses, yet a `gfx950:128` build was precompiling all of those kernels anyway. It now selects none of them, and once rows are tuned for 128 CUs it will select exactly those. 

### Scope

GEMM only. Each other `OpKind` reads its own CSVs and keeps only one of the two fields a target needs: MoE and chunk_gdn_h carry `cu_num` and recover an arch later through `cu_num_to_arch`, grouped MoE carries `gfx` alone, mxfp4 MoE hardcodes `FLYDSL_GPU_ARCH` to gfx950, and mega MoE builds its jobs without a CSV. Matching those jobs against a `(gfx, cu_num)` pair would drop every one. Filtering them means widening each `parse_csv` to carry both fields, plus a `gfx` column for the mxfp4 model CSVs that lack one, which is out of scope here.

## Technical Details

**1. Packaging now resolves targets the way CK does; the standalone entry point still does not.** CK's codegen resolves `GPU_ARCHS` and `CU_NUM` into exact `(gfx, cu_num)` pairs and keeps only the tuned rows matching them. FlyDSL packaging did none of that, baking every row in the CSVs whatever the build targeted. It now resolves the same pairs, so both halves of a build select for the same SKU.

The standalone entry point is unchanged: it still matches on architecture alone and ignores `CU_NUM`. Only that divergence is now visible. Under `GPU_ARCHS=gfx950 CU_NUM=128` packaging selects nothing, since every FlyDSL GEMM row records 256 CUs, while the CLI selects every gfx950 job. Neither result is wrong on its own, and an AOT miss only falls back to a runtime JIT compile, but one environment now means two different things depending on which entry point reads it. 

**2. An unknown arch now fails the build.** `GPU_ARCHS=gfx9999` raises, aborting the wheel build where packaging previously ignored the variable and baked everything. Intended, but it is a hard failure on a free-form `workflow_dispatch` input. Every in-tree caller passes a valid value and `gpu_archs` has a default, so no automated path can hit it.

**3. The filter reads `AITER_GPU_TARGETS` and `GPU_ARCHS` only.** `ARCH` stays on the standalone entry point: build images export it as a machine type such as `x86_64`, and an unrecognised value raises, so routing packaging through it would kill the wheel build on those hosts.

**4. Uppercase now matches where it previously matched nothing.** Case-folding means `GFX950` selects every gfx950 job where it used to select none. On the packaging path that is a narrowing, since it baked every job in the CSVs before reading no target variable at all; on the standalone entry point it is the opposite, since an unmatched arch there selected nothing and fell back to runtime JIT.

### Also in here

A release build gains one warning line naming gfx942, correct since no `libtype=flydsl` GEMM rows exist for it. `parse_csv` no longer raises on an empty `cu_num` cell; it parses as 0, which the filter treats as a wildcard. No shipped CSV trips this today.

## Test Plan

`op_tests/test_gemm_codegen.py`, covering the packaging collector end to end, partial target coverage, legacy rows with no `gfx`, both separator forms, and each rejected `GPU_ARCHS` input.

## Test Result

Passes; `ruff` and `black` clean. An AOT miss is not a failure, it falls back to a runtime JIT compile, so `AITER_GPU_TARGETS=gfx950:128` selecting 0 jobs today is safe.

### DIFF
--- a/aiter/aot/flydsl/README.md
+++ b/aiter/aot/flydsl/README.md
@@ -79,14 +79,17 @@ python -m aiter.aot.flydsl.chunk_gdn_h --csv /path/to/tuned.csv
 | `AITER_FLYDSL_AOT_TIMEOUT` | Per-kernel wall-clock cap (seconds). A worker stuck *alive* past this is killed (and retried); `0` disables. | `1200` |
 | `AITER_FLYDSL_AOT_MAX_RETRIES` | Retries for a worker that **died abnormally** (OOM-kill / segfault / timeout-kill). A clean compile error is never retried. `0` disables. | `2` |
 | `AITER_CONFIGS` | Resolves the default CSV lookup path (same as the runtime JIT) | repo built-in |
-| `ARCH` / `GPU_ARCHS` | **Banner/logging only** — printed as the "Target arch" line. Does **not** control the compiled target. | auto-detect |
+| `AITER_GPU_TARGETS` / `ARCH` / `GPU_ARCHS` | Filter **GEMM** AOT jobs by target; other kinds are unfiltered. The first one set wins, in that order. Unset builds every job. | unset |
 
-> **About the compile target arch.** The arch each kernel is actually compiled
-> for is derived per-job from the CSV's `cu_num` column (`cu_num_to_arch(...)`)
-> and applied internally via `FLYDSL_GPU_ARCH`. That internal var is overwritten
-> for every job, so setting `ARCH` / `GPU_ARCHS` / `FLYDSL_GPU_ARCH` in your shell
-> does **not** change what gets built. To cross-compile, edit the `cu_num`
-> column in the CSV.
+> **Compile target.** Each job is compiled for the arch in its `gfx` field, or
+> from the CSV `cu_num` via `cu_num_to_arch(...)` when `gfx` is empty. AOT
+> applies that through `FLYDSL_GPU_ARCH` and overwrites it per job, so setting
+> `FLYDSL_GPU_ARCH` in the shell does not change what gets built.
+
+`AITER_GPU_TARGETS=gfx950:128` is the exact `(gfx, cu_num)` form on both
+paths. `ARCH` always selects every GEMM job of that arch. `GPU_ARCHS` does
+that on the CLI (`CU_NUM` is ignored); packaging (`setup.py`) pairs it with
+`CU_NUM` into exact SKU pairs instead.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


samremes:
## Review (FlyDSL GEMM AOT consumer / non-comm enablement)

This is the packaging consumer of #5340 for [SILOTIGER-978](https://amd.atlassian.net/browse/SILOTIGER-978). Runtime lookup already keys FlyDSL GEMM on `(gfx, cu_num, …)`, so a 128-CU process does not pick a 256-CU tuned row; AOT miss falls back to JIT. GEMM-only is the right scope while other `OpKind`s record half a target. Keep this draft until #5340 lands; rebase first (`docs/autotuning_pipeline.md` already conflicts with the current #5340 head).

On the second-opinion item (`GPU_ARCHS` + `CU_NUM` vs `AITER_GPU_TARGETS`): **use the same resolved `(gfx, cu_num)` pairs as the rest of the build for packaging.** Arch-wide `GPU_ARCHS=gfx950 CU_NUM=128` currently keeps all 3,135 gfx950 FlyDSL GEMM jobs (all 256-CU in this tree) and does not warn that exact `gfx950:128` coverage is zero. Runtime is still safe; the cost is dead AOT data and a hidden coverage gap. CLI arch-wide filtering can stay a separate mode.

Also:

- `aiter/aot/flydsl/README.md` still says `ARCH` / `GPU_ARCHS` are logging only. This PR makes them filter GEMM jobs. Update that README with GEMM-only, precedence, exact vs arch-wide, empty-`cu_num` wildcard, unknown-arch hard fail, and which op kinds stay unfiltered.
- Tests construct `{cu_num: 0}` and only check GEMM. Add a real empty-cell CSV through `parse_csv`, and one job each for the unfiltered kinds through `_filter_collected_aot_jobs`.

I ran the codegen tests (82 function checks / 11 pytest cases). No full `PREBUILD_KERNELS=1` wheel.

[SILOTIGER-978]: https://amd.atlassian.net/browse/SILOTIGER-978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5363, opened 2026-09-09
[FlyDSL] Add flydsl radix topk one block kernel

lirui927:
## Motivation

This PR adds a FlyDSL one-block radix per-row TopK kernel and wires it into the default dispatch on gfx950, and gfx1250, replacing the HIP one-block path in those cases.

Goals:

- Prefill (variable-length rows described by `row_starts` / `row_ends`) and decode (`seq_lens` / `next_n`) share the same kernel.
- Support any positive integer `k`, optional values output, and deterministic stable mode.
- Keep short-row semantics: when a row has fewer than `k` live elements, pad indices with `-1` and values with `-inf`.
- Replace only the HIP one-block coverage. Existing multi-block paths and other fallbacks stay unchanged, which limits the risk of turning this on by default.

### One-block radix TopK

New file: `aiter/ops/flydsl/kernels/topk/radix_topk_one_block.py`.

- One workgroup per row.
- Short rows (physical width `<= 4096`) cache sorted keys in LDS. Long rows use a radix histogram plus candidate compaction, and fall back to a global-memory rescan if candidates overflow LDS.
- If `row_len <= k`, skip radix selection and copy/pad the output directly.
- gfx1250 uses wave32; gfx950 use wave64.
- Block size depends on short-row width and row count: 256 threads for short rows with `rows <= 256`, otherwise 1024.
- Stable output is sorted by index ascending (smaller index wins ties). Values always stay aligned with the corresponding indices.

### Dispatch

- Prefill: on the non-multi-block branch, try FlyDSL one-block first; fall back to HIP if `_is_flydsl_radix_topk_one_block_supported` returns false.
- Decode: insert the same FlyDSL one-block attempt before the existing path.
- Supported arches are `_FLYDSL_TOPK_ONE_BLOCK_ARCHES = ("gfx950", "gfx1250")`.
- Dispatch reads only physical `logits.shape[1]` and the row count, not device-side row bounds, so it is CUDAGraph-safe.
- This PR does not change the existing `topk_use_mulblocks` multi-block path or per-arch HIP fallbacks.
- New prefill kill-switch; decode reuses the existing one:

```bash
AITER_DISABLE_FLYDSL_TOPK_PREFILL=1   # added in this PR
AITER_DISABLE_FLYDSL_TOPK_DECODE=1    # existing
```

### Layout cleanup

TopK files are grouped into their own subpackages. In this PR these moves are renames only:

- `aiter/ops/flydsl/kernels/topk/` — six kernel modules.
- `aiter/ops/flydsl/topk/` — three launch wrappers.

### Optimizations for concentrated value distributions

Real logits often span a narrow range, so radix bins on float bits collapse onto a few buckets. That causes atomic contention on the first-pass histogram, and more first-pass survivors than the LDS candidate buffer can hold. This PR addresses both:

- **Wider first radix digit.** A finer first pass cuts both surviving candidates and the number of lanes hitting the same bin. At the same LDS cost, this is cheaper than replicating a coarser first pass. Later digits keep the same bin counts, so the candidate buffer is unaffected.
- **Grow the candidate buffer from leftover LDS.** If candidates overflow the buffer, pass 3 and the final emit rescan the full row in global memory (2 global passes become 4). More concentrated distributions hit this more often. The candidate buffer now takes the remaining LDS budget instead of a fixed 4096 entries.
- **Scatter pass-1 atomics by lane.** Histogram replicas are chosen by lane, not by wave: when one bin owns most of a row, a whole wave no longer serializes onto one address. Wave-index replicas only split waves that were already concurrent, which does not help.
- **In-wave reduction for same-bin vectors.** If the four keys from one vector load share the same first-pass digit, ballot-merge the counts in the wave and issue a single atomic add. More concentrated data means neighboring elements share a bin more often.

These changes affect performance only; output semantics are unchanged.

## Correctness

`op_tests/test_topk_per_row.py` now calls the one-block kernel directly (bypassing dispatch). On supported arches, each prefill shape also runs the four stable × write_values combinations, so dispatch changes cannot hide a kernel regression.

Coverage:

- Prefill and decode.
- Full rows, padded rows, and non-zero row starts.
- `next_n=1` and MTP decode.
- Stable and unstable.
- Indices-only and indices+values.
- Random inputs and several degrees of value ties.
- Workspace reuse and CUDAGraph replay.

## How to test

```bash
python -m op_tests.test_topk_per_row
python -m op_tests.test_topk_per_row_stable
```

### Method

- Latency after CUDAGraph capture: 3 warmups, 5 alternating HIP / FlyDSL rounds, median. Both HIP and FlyDSL use one-block.
- Full rows, padding, and uneven logits (narrow, single-peak, few discrete values).
- M (rows): 1, 2, 16, 32, 52, 64, 128, 512, 1024, 4096, 16384.
- N (width): 128, 4096, 4097, 16384, 32768, 65536, 98304, 131072, 164401, 200000.
- K: 128, 512, 2048, 4096.

### GFX1250

**Conclusion: default FlyDSL one-block is faster overall, with no systematic regression.**  
775 cases, geomean **1.41×**; 731 faster, 44 slower.

| Scenario | n | Geomean speedup | Faster |
|---|---:|---:|---:|
| All | 775 | **1.41×** | 731/775 |
| Prefill | 377 | **1.40×** | 355/377 |
| Decode | 398 | **1.42×** | 376/398 |
| Uneven distribution | 214 | **1.36×** | 210/214 |

Representative points (k=2048, stable+values, full rows):

| rows × width | Prefill | Decode |
|---|---:|---:|
| 1 × 4096 | 1.78× | 1.82× |
| 16 × 65536 | 1.51× | 1.37× |
| 64 × 65536 | 1.51× | 1.48× |
| 128 × 131072 | 1.40× | 1.38× |
| 16384 × 4096 | 2.51× | 2.78× |
| 16384 × 200000 | 1.18× | 1.29× |

Of the 44 slower cases, 42 differ by at most 5 µs; the median relative regression is 3.3%. The remaining 2 cases regress by 6.6%–6.9%.

### GFX950

**Conclusion: default FlyDSL one-block is faster overall, with no systematic regression.**  
775 cases, geomean **1.44×**; 741 faster, 34 slower (worst about 0.88×, ~5.5 µs slower).

| Scenario | n | Geomean speedup | Faster |
|---|---:|---:|---:|
| All | 775 | **1.44×** | 741/775 |
| Prefill | 377 | **1.44×** | 360/377 |
| Decode | 398 | **1.44×** | 381/398 |
| Uneven distribution | 214 | **1.40×** | 195/214 |

Representative points (k=2048, stable+values, full rows):

| rows × width | Prefill | Decode |
|---|---:|---:|
| 1 × 4096 | 1.65× | 1.75× |
| 16 × 65536 | 1.68× | 1.49× |
| 64 × 65536 | 1.67× | 1.69× |
| 128 × 131072 | 1.68× | 1.73× |
| 16384 × 4096 | 2.30× | 2.35× |
| 16384 × 200000 | 1.52× | 1.53× |

Slower cases are mostly large batches that also write values; most are only a few percent slower, typically about 5 µs.

## Checklist

- [x] I have read the contributing guide https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests


aiter/ops/flydsl/kernels/radix_topk_one_block_gfx1250.py
```diff
@@ -0,0 +1,1368 @@
...
+        width: fx.Int32,
+        next_n: fx.Int32,
+        rows_m: fx.Int32,
+        stream: fx.Stream,
+    ):
+        radix_topk_one_block_gfx1250_kernel(
+            input, row_starts, row_ends, indices, values, width, next_n
+        ).launch(grid=(rows_m, 1, 1), block=(block_threads, 1, 1), stream=stream)
+
+    return launch_radix_topk_one_block_gfx1250
```
coderfeli: no need to be 1250 specific?

coderfeli:
LGTM

valarLip (approved):

PR #5370, opened 2026-09-09
[HIP] [FlyDSL] [JIT] Support Conv3d for Qwen-Image / Wan2.1 VAE on MI35X

huizzhan:
## Motivation

aiter has no conv3d (only Triton conv2d). Two diffusion VAEs need one:

- **Wan2.1** encodes in 4-frame chunks and prepends 2 cached feature frames
  after the first chunk. Those frames are real, not causal-pad zeros, so the
  `T=1` `conv3d -> conv2d` identity fails. That is **67.7% of one encode's 650
  convolution calls**.
- **Qwen-Image** is the same VAE fine-tuned at `T=1`, where the identity holds,
  so it needs this kernel's 2-D path.

A GEMM library cannot absorb these shapes: N is the 96/192/384 channel ladder,
and 7 of 18 Qwen-Image shapes have `K` not a multiple of 64 or `N=3`, which
A16W16 cannot express.

### 1. Vendored implicit-GEMM convolution (`flydsl_conv_implicit`)

Same vendoring as `gemm_a16w16_gfx950`. Dispatches 1D/2D/3D off filter rank;
gathers from the source instead of im2col (~9× less activation traffic).
Keyword surface unchanged (`groups`, `padding_mode`, `splitk`, `stream`,
`tile`, `wgm`, `autotune`, `input_layout`, `output_layout`).

Three mechanical edits vs upstream: local `buffer_atomic_add` (flydsl wheel
has no `kernels/common/`); `kt` -> `k_tile` (filter temporal extent); one
`if` collapsed to `and`. 54/56 top-level definitions byte-identical.

Named `flydsl_conv_implicit` because the entry point dispatches on rank.

### 2. Offline tile tuner

The heuristic N tiles are powers of two. Against 96/192/384, two rungs sit at
**75% N occupancy**; masked B columns still issue MFMA. `TILE_N` 96/192 is
legal but was unreachable: `TILE_{M,N} % (16 * WAVE_M * WAVE_N) == 0` excludes
the eight-wave shape every shipped candidate uses.

- `conv3d_policy.py` enumerates that closed form (reject = arithmetic, not a
  compile), then prunes on N fill, waves, LDS, accumulators.
- `csrc/flydsl_conv3d/conv3d_tune.py` writes per-model CSVs under
  `aiter/configs/model_configs/` (`qwenimage_vae_*`, `wan21_vae_*`).
- `_lookup_tuned_tile` reads them. Miss / missing file / malformed row falls
  back to the unmodified heuristic.

Launch config is five integers, not a `solidx` (reordering candidates must not
silently invalidate the table).

**Tune with `--compare --update_improved`.** `mp_tuner` times across eight GPU
processes; production runs one. A bare sweep regressed three shapes by up to
4.8%. The compare gate re-times on the production path and rejects gains below
`--min_improvement_pct` (default 3%).

Also fixes `autotune=True` / `FLYDSL_CONV3D_AUTOTUNE=1`, which imported a
FlyDSL-repo path and raised `ImportError`.

### 3. A16W16 pruner fix

`GemmConfigPruner` `keep_ratio` climbs with M: at M=27556 it kept only tiles
wasting <0.44% on padding. HTI needs `block_n=256`, so N=384 (pad to 512)
dropped the whole HTI catalog (~30% loss). Cap lets a tile pad up to 30%.

### 4. Tests

`op_tests/test_flydsl_conv_implicit.py`: keyword surface plus Wan / Qwen
shapes from an input resolution (`--wan-res`, `--wan-frames`, `--qwen-res`),
traced on meta-device encodes. Benches:
`op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/bench_three_way.py` (conv vs
A16W16 vs hipBLASLt) and `bench_conv_tuned_ab.py` (tile table on vs off).

## Test Plan

```bash
PYTHONPATH=$PWD python3 op_tests/test_flydsl_conv_implicit.py
python3 -m unittest op_tests.tuning_tests.test_config_shape_collision -v

PYTHONPATH=$PWD python3 csrc/flydsl_conv3d/conv3d_tune.py \
  -i aiter/configs/model_configs/qwenimage_vae_bf16_untuned_conv3d.csv \
  -o aiter/configs/model_configs/qwenimage_vae_bf16_tuned_conv3d.csv \
  --max_configs 32 --mp 8 --compare --update_improved

PYTHONPATH=$PWD python3 op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/bench_three_way.py
PYTHONPATH=$PWD python3 op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/bench_conv_tuned_ab.py
```

All results below: gfx950 / 256 CU, bf16.

### Correctness

`test_flydsl_conv_implicit.py`: **75/75 `checkAllclose`**, `rtol=atol=2e-2`.
`flydsl err` **0 on all 33 model shapes** (the one non-zero cell is torch's
own bf16 rounding at K=27, `err_ratio` 4e-05).

`test_config_shape_collision.py`: 18/18, including `conv3d_bf16_tuned`.

Every tuned row has `err_ratio = 0.0`. Instrumenting `_lookup_tuned_tile`
over the op_test: **1236 hits / 12 distinct shapes**, all non-default tiles
(e.g. `(96, 96, 2, 3)`, `(192, 128, 2, 2)`, `(128, 192, 1, 4) wgm=8`).

### Performance (gfx950 / 256 CU, bf16)

One table for the published buckets. **vs X** is `X_ms / current CSV` (or
`torch / orig flydsl` only where current is missing). Torch and original
flydsl † are the PR-open three-way (NCHW, includes pre-transpose). hip is
`three_way.json` library-best `torch.mm` on materialised `M×K` (no im2col).
Current CSV is NDHWC (`ed7b5a5a8`). ‡ NCDHW-tuned CSV (`05f919868`). Mixed
sessions; hip/a16 traffic is not the conv gather.

| bucket | torch | orig flydsl | hip | current CSV | vs orig flydsl | vs hip | vs torch |
|---|---:|---:|---:|---:|---:|---:|---:|
| Wan2.1 `T>1`/cached, 480×832, 81f (8 shapes, 440 calls) | 505.32 ms | 310.37 ms † | — | **200.91 ms** | **1.55x** | — | **2.52x** |
| Wan2.1 same 8, 368×544, 81f | — | 128.70 ms ‡ | — | **110.43 ms** | **1.17x** | — | — |
| Qwen-Image, 1024² (10 shapes, 52 calls) | 20.37 ms | 12.00 ms † | 8.70 ms | **7.76 ms** | **1.55x** | **1.12x** | **2.63x** |
| Qwen-Image, 1024² (16 layers, 58 calls) | — | — | 10.02 ms | **8.92 ms** | — | **1.12x** | — |
| Qwen-Image, 1328² (10 shapes, 52 calls) | — | 16.13 ms ‡ | — | **12.68 ms** | **1.27x** | — | — |
| Wan2.1 resamplers + pointwise (15 shapes) | 34.67 ms | 32.07 ms † | — | — | — | — | 1.08x † |

† Three-way at PR open (NCHW in). Orig flydsl vs torch on that row is **1.63x**
(Wan 8-shape) / **1.70x** (Qwen 10-shape) / **1.08x** (resamplers). Wan conv3d
matches the Lumen 81-frame report (1.59x). Minus the two `time_conv`, the
pointwise bucket is 0.99x (stay on torch); `time_conv` alone is 1.54x
(`3x1x1`, `K=C*3`). `time_conv` is not in the tuned tables, so the 15-shape
row has no current CSV. Hip was not measured for Wan T>1 (A up to ~8 GB,
omits 27× im2col). Qwen 1328² hip exists for only 2 of 10 shapes.

CSV rows outside those buckets (same weighting):

| | resample 3×3 (6 shapes, 63 calls) | decode extras (8 shapes, 103 calls) | all CSV rows |
|---|---:|---:|---:|
| Wan 480×832 | 6.69 ms | 41.88 ms | 249.49 ms (22 shapes, 606 calls) |
| Wan 368×544 | 4.02 ms | 21.92 ms | 136.36 ms |
| Qwen 1024², 16 layers (adds 3 downs + 3 ups) | — | — | 8.92 ms (58 calls) |
| Qwen 1328², 16 layers | — | — | 14.63 ms |

#### Wan2.1 `T>1`/cached conv3d, weighted (81 frames)

| shape | calls | 480×832 `us` | 480 contrib | 368×544 `us` | 368 contrib |
|---|---:|---:|---:|---:|---:|
| `conv_in` 3→96 | 20 | 183.46 | 3.67 ms | 114.62 | 2.29 ms |
| `down_0_1` 96→96 | 80 | 1017.87 | 81.43 ms | 515.87 | 41.27 ms |
| `down_3_in` 96→192 | 20 | 469.12 | 9.38 ms | 245.18 | 4.90 ms |
| `down_3_4` 192→192 | 60 | 907.95 | 54.48 ms | 467.67 | 28.06 ms |
| `down_6_in` 192→384 | 20 | 226.62 | 4.53 ms | 128.58 | 2.57 ms |
| `down_6_7` 384→384 | 60 | 456.06 | 27.36 ms | 247.35 | 14.84 ms |
| `down_9_10_mid` 384→384 | 160 | 114.82 | 18.37 ms | 92.70 | 14.83 ms |
| `conv_out` 384→32 | 20 | 84.46 | 1.69 ms | 82.91 | 1.66 ms |
| **weighted** | **440** | | **200.91 ms** | | **110.43 ms** |

#### Qwen-Image conv2d, weighted (one encode+decode)

| shape | calls | 1024² `us` | 1024 contrib | 1328² `us` | 1328 contrib |
|---|---:|---:|---:|---:|---:|
| `3->96` | 1 | 93.65 | 0.09 ms | 139.86 | 0.14 ms |
| `96->96` | 10 | 262.65 | 2.63 ms | 439.36 | 4.39 ms |
| `96->192` | 1 | 122.02 | 0.12 ms | 195.23 | 0.20 ms |
| `192->192` | 9 | 202.00 | 1.82 ms | 343.49 | 3.09 ms |
| `192->384` | 2 | 104.61 | 0.21 ms | 173.29 | 0.35 ms |
| `384->384` (L2) | 8 | 189.34 | 1.51 ms | 319.24 | 2.55 ms |
| `384->384` (L3) | 18 | 65.50 | 1.18 ms | 91.20 | 1.64 ms |
| `384->32` | 1 | 31.14 | 0.03 ms | 32.50 | 0.03 ms |
| `16->384` | 1 | 15.92 | 0.02 ms | 18.90 | 0.02 ms |
| `96->3` | 1 | 145.26 | 0.15 ms | 262.38 | 0.26 ms |
| **weighted** | **52** | | **7.76 ms** | | **12.68 ms** |

#### Per-shape vs hipBLASLt `mm` and FlyDSL `gemm_a16w16` (same `M/N/K`)

hip / a16 from `three_way.json`. **vs hip** = `hip_us / other_us`. a16 is `—`
when `K % 64 != 0` or `N=3` (7/18 shapes). `@166²` is 1328² L3, not a 1024²
row.

| shape | calls | hip µs | a16 µs | conv CSV µs | a16 vs hip | conv vs hip |
| `3->96 @1024²` | 1 | 85.3 | — | 93.6 | — | 0.91x |
| `96->96 @1024²` | 10 | 417.7 | — | 262.6 | — | **1.59x** |
| `96->192 @512²` | 1 | 133.6 | — | 122.0 | — | **1.09x** |
| `192->192 @512²` | 9 | 190.5 | 256.1 | 202.0 | 0.74x | 0.94x |
| `192->384 @256²` | 2 | 84.9 | 99.5 | 104.6 | 0.85x | 0.81x |
| `384->384 @256²` | 8 | 150.8 | 174.8 | 189.3 | 0.86x | 0.80x |
| `384->384 @128²` | 18 | 49.1 | 56.9 | 65.5 | 0.86x | 0.75x |
| `384->32 @128²` | 1 | 23.9 | 21.4 | 31.1 | 1.11x | 0.77x |
| `16->384 @128²` | 1 | 10.0 | — | 15.9 | — | 0.63x |
| `96->3 @1024²` | 1 | 296.8 | — | 145.3 | — | **2.04x** |
| `96->96 @1025² s2` | 1 | 106.3 | — | 96.9 | — | **1.10x** |
| `192->192 @513² s2` | 1 | 52.1 | 62.3 | 66.8 | 0.84x | 0.78x |
| `384->384 @257² s2` | 1 | 49.1 | 56.9 | 68.3 | 0.86x | 0.72x |
| `384->192 @256²` | 1 | 90.9 | 113.0 | 109.0 | 0.80x | 0.83x |
| `384->192 @512²` | 1 | 342.0 | 446.0 | 417.5 | 0.77x | 0.82x |
| `192->96 @1024²` | 1 | 674.3 | 707.1 | 410.1 | 0.95x | **1.64x** |
| `384->384 @166²` | 18 | 68.3 | **81.5** | 91.2 | **0.84x** | **0.75x** |
| `96->96 @1328²` | 10 | 719.3 | — | 439.4 | — | **1.64x** |

384 / `K=3456`: conv ~0.75x of hip, a16 **0.84–0.86x**. Heuristic `@166²` was
0.67x on this hip arm. Weighted 1024² still beats hip via `96->96` and skinny
N.

#### How much of the 1.55x is the transpose

Kernel reads NDHWC only. NCDHW input pays a transpose no tile affects.
Original 12.00 / 310.37 ms include it; current CSV `us` does not.

`05f919868` NCDHW CSV = later tile + transpose. `ed7b5a5a8` NDHWC retune.
Same-layout tile-only A/B: **−0.95%** Qwen / **−0.78%** Wan unweighted, so
the CSV gap is the transpose.

| bucket | original flydsl | NCDHW CSV (kernel + transpose) | current NDHWC | kernel since the original | drop transpose | product |
| Qwen 1024², 10 shapes | 12.00 ms † | 9.71 ms | 7.76 ms | **1.24x** | **1.25x** | 1.55x |
| Qwen 1328², same 10 | — | 16.13 ms | 12.68 ms | — | **1.27x** | — |
| Wan 480×832, 8 shapes | 310.37 ms † | 235.51 ms | 200.91 ms | **1.32x** | **1.17x** | 1.55x |
| Wan 368×544, same 8 | — | 128.70 ms | 110.43 ms | — | **1.17x** | — |

Saved vs original flydsl: Qwen 4.24 ms, transpose **1.95 ms (46%)** (~half of
ln(1.55x)); Wan 8-shape 109 ms, transpose **35 ms (32%)** (~one third).

Transpose as a share of NCDHW (kernel + transpose): Qwen 1024² **20%**
(1.95 / 9.71 ms), 1328² **21%** (3.46 / 16.13 ms); Wan 480×832 **15%**
(34.6 / 235.51 ms), 368×544 **14%** (18.3 / 128.70 ms). Those eight Wan
layers have no stride-2 resample, so they sit below the 76-row retune:
**13% median / 20% summed / 50% peak** (peak = stride-2). Full CSV passes
513.1 → 409.4 ms (−20.2%).

1.55x is not all transpose: Qwen ~half kernel/tile since the three-way, half
transpose; Wan ~two-thirds kernel, one-third transpose.

### What the tile tuner added

`bench_conv_tuned_ab.py`, conv only (no NCHW→NHWC transpose), rotated inputs,
best of three, arms interleaved per shape:

**Weighted by calls per decode: 16842.5 us -> 15960.3 us, 1.055x.**

| shape | calls | heuristic | tuned | speedup |
|---|---|---|---|---|
| `384->384 @166²` | 18 | 102.58 us | 86.55 us | **1.185x** |
| `96->3 @1024²` | 1 | 150.54 | 131.28 | 1.147x |
| `96->192 @512²` | 1 | 144.36 | 126.00 | 1.146x |
| `192->192 @512²` | 9 | 233.46 | 207.42 | 1.126x |
| `384->192 @256²` | 1 | 114.15 | 105.68 | 1.080x |
| `384->192 @512²` | 1 | 454.62 | 421.68 | 1.078x |
| `16->384 @128²` | 1 | 11.35 | 10.67 | 1.063x |
| `96->96 @1024²` | 10 | 307.40 | 296.68 | 1.036x |
| `96->96 @1328²` | 10 | 522.56 | 505.48 | 1.034x |
| `192->96 @1024²` | 1 | 505.92 | 489.17 | 1.034x |

Ten of eighteen shapes hit the table; the other eight land **0.984x–1.003x**
(miss = no-op; noise ~±1.6%). Nine of those ten took a non-power-of-two N
tile.

Largest gain: `384->384 @166²` (18 calls/decode). Heuristic conv 0.67x of the
stored hipBLASLt arm; a16 **0.84x** (81.5 vs 68.3 µs); NDHWC CSV **0.75x**
(91.2 µs).

### Scope and limits

- **bf16, forward only.** Grad tensors route to torch. These VAEs run
  `no_grad`.
- Tuned tables are per-model, keyed on the full conv shape including `bias`
  (compile-time). Anything else uses the heuristic.
- Benches time only; correctness is the op_test and the tuner's `errRatio`
  gate.
- `three_way.json` predates the tile table (conv arm = heuristic). Conv arm
  reproduces within 0.6% weighted; 5.5% slower than current HEAD.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests
- [x] `pre-commit` / ruff clean on all added and modified files
- [x] Correctness verified against torch on every added shape
- [x] New config family registered in `test_config_shape_collision.py`
- [x] Runtime degrades to the existing heuristic if the tuned CSV is absent or malformed


aiter/configs/bf16_tuned_conv3d.csv
```diff
@@ -0,0 +1 @@
+gfx,cu_num,N,C,D,H,W,K,kT,kH,kW,stride_d,stride_h,stride_w,pad_d,pad_h,pad_w,dil_d,dil_h,dil_w,groups,bias,tile_m,tile_n,wave_m,wave_n,wgm,splitK,us,kernelName,err_ratio,tflops,bw
```
yzhou103: I think we can also add libtype column, we may add more other implementations. The1024x1024 csv and 1328x1328 csv should be put together in one csv

huizzhan: Refined

zhiding512:
Reviewed `aiter/aot/flydsl/conv.py`. The `dyn_hw` mechanism itself checks out — I chased the parts that would be easy to get wrong and they're right:

- `_launch.compile` appends `*extra_args`, so the 4-tensor `_dispatch` in the AOT pass is correct against both kernel signatures.
- The 14 `fx.Int64` scalars have `__cache_signature__ == (type(self),)`, so the extents really do leave the artifact key.
- `_conv_dedupe_key` mirrors `_assert_shape_agnostic` faithfully, including `grid._replace(grid_x=0, grid_z=0, grid_m=0)` and building the scatter plan from the *unblanked* grid.

The splitK rework looks like a solid fix: going through the real `_resolve_splitk` with `num_cu=` picks up the `SPLITK_MAX_STAGING_BYTES` guard, and treating an absent column as `None` rather than `1` closes a case where the AOT built an artifact the runtime would never ask for.

One substantive issue below, then some small stuff.

## The AOT artifacts are unreachable at the resolutions whose rows were dropped

The module docstring says:

> one artifact serves every resolution of the same layer, so a row covers sizes the table never listed

The *artifact* is resolution-agnostic, but the *launch config* isn't. `tuned_key` (`conv_kernels.py:845`) still carries `n, d, h, w` and `_lookup_tuned_tile` is an exact match, so an unlisted resolution misses the table, falls through to `_pick_tile`/`_pick_wgm`, and asks for a different `(tile, wgm)` — which *is* part of the compile key. That matters here because the same change drops the second resolution of each model from the tuned tables (76 rows → 38).

### Repro

AOT only the 1024x1024 row of one Qwen layer, then run both resolutions under `FLYDSL_RUNTIME_RUN_ONLY=1` (gfx950):

```
1024x1024: hits the AOT cache
1328x1328: RuntimeError: FLYDSL_RUNTIME_RUN_ONLY=1 but no usable AOT cache for launch
```

Forcing the tuned tile at the unlisted resolution makes it hit, which isolates the cause to the tile lookup rather than to `dyn_hw`:

```
1328x1328 + tile=(128,128,2,2): hits
```

```bash
head -1 aiter/configs/model_configs/qwenimage_vae_bf16_tuned_conv3d.csv > /tmp/one_row.csv
grep '^gfx950,256,1,96,1,1024,1024,96,' \
  aiter/configs/model_configs/qwenimage_vae_bf16_tuned_conv3d.csv >> /tmp/one_row.csv
FLYDSL_RUNTIME_CACHE_DIR=/tmp/convcache python -m aiter.aot.flydsl.conv --csv /tmp/one_row.csv

FLYDSL_RUNTIME_CACHE_DIR=/tmp/convcache FLYDSL_RUNTIME_RUN_ONLY=1 python -c '
import torch
from aiter.ops.flydsl import flydsl_conv_implicit
w = torch.randn(96, 96, 1, 3, 3, device="cuda", dtype=torch.bfloat16)
b = torch.randn(96, device="cuda", dtype=torch.bfloat16)
for res in (1024, 1328):
    x = torch.randn(1, 96, 1, res, res, device="cuda", dtype=torch.bfloat16)
    try:
        flydsl_conv_implicit(x, w, b, stride=1, padding=(0,1,1)); print(res, "hit")
    except Exception as e: print(res, "MISS:", type(e).__name__)
    del x; torch.cuda.empty_cache()
'
```

### How often it happens

Of the 38 dropped rows, 3 have a heuristic `(tile, wgm)` that happens to match the retained row's tuned one; 35 don't. The structural reason doesn't depend on my arithmetic: `_pick_tile` can only ever return one of four tiles (`TILE_LADDER` + `TILE_WIDE_N`), and 28 of the 38 tuned rows use a tile outside that set — `(128,192,2,2)`, `(96,96,2,3)`, `(192,128,4,1)`, and so on.

### What it costs

Ran each dropped row two ways on one idle gfx950: the tile the runtime picks today (heuristic, since the row is gone) vs the tile that row was tuned to. Median of 20 iterations after 3 warmups, `torch.cuda.Event`.

| | heuristic (today) | deleted row's tuned tile | delta |
|---|---|---|---|
| Qwen-Image VAE @ 1328x1328 (14 layers) | 5703 us | 5348 us | **+6.6%** |
| Wan2.1 VAE @ 480x832 (15 layers) | 7984 us | 7341 us | **+8.8%** |
| combined | 13687 us | 12690 us | **+7.9%** |

27 of the 29 layers I could measure are >2% slower on the heuristic tile; the worst are +14.7% (`C=384 K=192 1x332x332`), +14.3% (`C=96 K=192 1x664x664`), +12.6% (`C=192 K=192 6x242x418`) and +12.2% (`C=384 K=384 1x166x166`). Caveat: 9 of the 38 rows didn't run for me — they take the WGM-swizzle path, which needs `fx.min`, and my local flydsl is 0.3.1 against the 0.3.4.1 in `requirements.txt`. They're excluded from both columns, so the comparison is still apples to apples, but the sample isn't the full set.

So at 1328x1328 and 480x832 the VAE now (a) JITs on first call despite the AOT pass and (b) runs ~8% slower in kernel time. Worth noting the test file's own comment already describes (b) correctly — "fall back to `_pick_tile`'s heuristic at any other" — which reads at odds with the AOT docstring. The -0.70% median / -1.4% total quoted in `conv_kernels.py` is `dyn_hw` on-vs-off and doesn't cover the row removal.

### Options

1. Have the table lookup fall back to a resolution-agnostic row for the same layer when the exact match misses. That's what closes the loop `dyn_hw` opens, and it's the one that makes the docstring claim true.
2. Keep the dropped rows. Dedupe already collapses same-layer/same-tile rows, so this doesn't take the job count back to where it was.
3. At minimum, reword the docstring claim so it doesn't promise coverage the dispatch can't deliver.

## Smaller things

- `aiter/aot/flydsl/README.md` gained a `mega_moe.py` row in this PR but still doesn't mention `conv.py` at all — neither the module table nor the usage list.
- `_probe`'s docstring says only rank and dtype reach the cache key. With `dynamic_layout=True` the signature is `(type, dtype, align, use_32bit_stride, (-1,)*n, strides-with-unit-index)`, so `strides.index(1)` is in there too. The probes work because the innermost extent is `TR_VEC`, which puts the unit stride on the last axis like a real contiguous tensor; a probe with a trailing extent of 1 would put it at rank-2 and silently miss. Worth stating that invariant — and perhaps not borrowing the transpose's `TR_VEC` for it. (The `if rank > 1` ternary is also redundant: both branches give `(TR_VEC,)` at rank 1.)
- `c_padded % TR_VEC == 0` is always true — `LDG_VEC == TR_VEC == 8` and `c_padded = groups * _pad_channels(C // groups)`.
- The `ARCH`/`GPU_ARCHS` filter only exists in `main()`; `run_aot()` (the `setup.py` path) calls `collect_aot_jobs` directly and skips it. Same shape as `gemm.py`, so pre-existing rather than introduced here — but conv contributes the most jobs of any kind.

zhiding512 (dismissed):
LGTM.

yzhou103:
LGTM

coderfeli (approved):

PR #5373, opened 2026-09-09
[FlyDSL] perf(hca): the gfx1250 hca_compress optimization

XingerZhu:
…nfig table

The Phase 1 / Phase 2 split gave compress_forward's K loop runtime bounds, so it could not unroll and each iteration serialised into a load -> softmax -> load dependency chain. Splitting the cost three ways (all-sentinel plan, one active boundary, full plan) showed the kernel was spending 3.1us on a single block's latency chain against only 2.8us of throughput for all 128 boundaries, which is also why larger k_split_num_waves used to help: it only shortened the chain.

Phase 2 now covers the wave's whole K range with the Phase 1 prefix masked to -inf. A masked step is an exact no-op (score=-inf forces w_k=0 and m_new=m_old, so the accumulator is scaled by exp(0)=1 and incremented by 0), and the trip count becomes the compile-time K_PER_WAVE, so the loop unrolls and the scheduler batches all K loads. Two smaller latency fixes go with it: the state-cache `slot` load moves into the Phase 1 loop, where LICM sinks it to the preheader so waves that read no state cache never pay for that dependent round trip; and Kernel B's kv_compressed load moves above the sentinel guard, since its address needs only pid and the guard was chaining it behind the plan load.

On prefill bs=64 (16K tokens, gfx1250) compress_forward goes 19.1us -> 9.2us and the 2-kernel path 22.6us -> 12.9us, measured A/B in one process run.

hca_per_n_config_gfx1250 shipped as "Initial values -- need hardware tuning" and selected slice_size=512 above plan_capacity 1024, which does not build at all because VEC=16 exceeds what the Phase 1 f32 loader handles -- decode bs=1024 failed outright. It is retuned from measured medians and VEC=16 now fails at build time with a clear message instead of a bare assert.

Validated against the torch reference across hca_main prefill bs=1/4/16/64 and decode bs=2/64/1024 with mtp=0/3; decode bs=1024 goes from a build failure to 70.7us.

## Submission Checklist


coderfeli:
help apply the flydsl kernel code cleanup skill

junhaha666 (approved):

PR #5375, opened 2026-09-09
[Triton/Gluon] [CI] Consolidate MHA backend test parameterization

vorapolsiloai:
## Summary

- restore a single pytest entry point for fixed-length MHA and one for varlen MHA
- couple each backend to its supported head-pair/head-size matrix so pytest does not form cross-backend Cartesian products
- preserve the exact post-#5352 coverage: 450 fixed-length cases and 1,120 varlen cases

## Why

#4147 added `backend=["triton", "gluon"]` as an independent parameter on shape matrices expanded for Gluon. That made Triton run every Gluon shape too. The [initial #5352 shard-4 run](https://github.com/ROCm/aiter/actions/runs/34260549117/job/102179514759) had 14 MHA failures; every one was a Triton causal case with sequence length 2048, head size 33, and `(Q, K)` heads `(1, 1)` or `(8, 1)`.

#5352 fixed that accidental cross-product by splitting the backend tests and restoring the original Triton matrices. The shape isolation was necessary, but separate test functions were not. This PR expresses the backend distinction in the parameter data while keeping shared batches, sequence lengths, causal modes, implementation, and assertions in one test function per API.

The fifteenth failure in that CI job was the known MHA-PE case whose node ID changed after backend parameterization. Its workflow deselect was corrected separately by #5352 and is unchanged here.

## Test plan

- [x] `python3 -m compileall -q op_tests/triton_tests/attention/test_mha.py`
- [x] `black --check op_tests/triton_tests/attention/test_mha.py`
- [x] `ruff check op_tests/triton_tests/attention/test_mha.py`
- [x] `git diff --check`
- [x] validate unique backend/head tuples and exact totals (450 fixed + 1,120 varlen)
- [ ] upstream GPU CI

### DIFF
--- a/op_tests/triton_tests/attention/test_mha.py
+++ b/op_tests/triton_tests/attention/test_mha.py
@@ -30,6 +30,48 @@ logger = logging.getLogger(__name__)
 DEBUG_MODE = False
 
 
+# Keep backend-specific head domains coupled to avoid unsupported cross-products.
+def _backend_head_cases(backend, head_pairs, head_sizes):
+    return [
+        pytest.param(
+            backend,
+            num_q_heads,
+            num_k_heads,
+            head_size,
+            id=f"{backend}-{num_q_heads}-{num_k_heads}-{head_size}",
+        )
+        for num_q_heads, num_k_heads in head_pairs
+        for head_size in head_sizes
+    ]
+
+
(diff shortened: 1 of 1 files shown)


vorapolsiloai:
duplicated #5374

PR #5404, opened 2026-09-10
[CK] [FlyDSL] [Perf] Add BM16 SwiGLU path for MiniMax M3

XiaobingSuper:
## Summary

- add MiniMax-M3 A16W4 tuned/untuned configs for BF16 activation input
- dispatch profitable `M<=128` BF16 shapes to main's activation-tagged BM16
  inline-quant SwiGLU path
- keep current-main dispatch for low-M shapes where the fused path does not win
- rename the forced-FP4 config to A4W4 and retain the validated
  `(M=256, I=384)` BM16 winner
- extend the MXMOE G1 full-search tuner to cover `xcd_swizzle=1` (legacy A16W4
  GEMM1 already supported XCD1), and use route-validated XCD1 winners for
  `M=4..64, I=768`
- keep the generated fused auxiliary sort for small expert sets through the
  measured crossover, reducing M64/M128 from four graph kernels to three
- include A16W4 model configs in MXFP4 MoE AOT discovery

The underlying MXMOE kernel, parser, and runtime support from the original
version of this PR has since landed in main. This PR now adds the MiniMax-M3
configs, A16W4 AOT discovery, MXMOE XCD1 tuner coverage, the validated
small-expert sort crossover, and quantization-aware config validation.

## Config layout

- `minimax_m3_a16w4_tuned_fmoe.csv`: 14 BF16-input rows through `M=128`;
  profitable rows use `_f16in_swiglu`, while low-M losers retain main dispatch
- `minimax_m3_a16w4_untuned_fmoe.csv`: 16 reproducible BF16-input shapes through
  `M=128`
- `minimax_m3_a4w4_tuned_fmoe.csv`: renamed forced-FP4 config with one content
  change, the validated `(M=256, I=384)` BM16 winner
- `minimax_m3_a4w4_untuned_fmoe.csv`: pure rename of the forced-FP4 untuned
  config

The `_f16in` path consumes BF16 activations and quantizes inside stage1, so its
GEMM remains MXFP4 A4W4.

## Current-main performance

Paired on one MI355X (`gfx950`, 256 CUs), comparing benchmark baseline
`23965aba8` with the production/config tree published at `38835a917`.

- public `fused_moe` CUDA Graph replay, including route/sort/quant/stage1/stage2
- `GateMode.SEPARATED`, SwiGLU limit 7, no bias
- identical frozen inputs and weights on the same physical GPU
- balanced, random, and hotspot routing
- ABBA alternating main/PR order, at least eight rounds per version
- noisy and initially regressing shapes rerun at 5,000-10,000 iterations

Results:

- `M<=64`, 42 cases: `1.1223x` geomean (**10.90% lower latency**)
- `M<=256`, 54 cases: `1.1264x` geomean (**11.22% lower latency**)
- `I=768`, 27 cases: `1.1111x` geomean (**10.00% lower latency**)
- regressions greater than 2%: **0**
- worst case: `0.9839x`
- maximum timing spread: main `3.64%`, PR `3.98%`
- noisy cases above 4%: **0**

### Per-shape performance

Each row is the geomean across balanced, random, and hotspot routing. The
latencies below are measured graph-replay times; they are not the `us` values
stored in the tuned CSV.

| M | I | main (us) | PR (us) | speedup |
|---:|---:|---:|---:|---:|
| 1 | 384 | 26.215 | 26.361 | 0.9945x |
| 2 | 384 | 29.523 | 29.550 | 0.9991x |
| 4 | 384 | 36.085 | 32.369 | 1.1148x |
| 8 | 384 | 42.594 | 36.025 | 1.1824x |
| 16 | 384 | 54.747 | 44.138 | 1.2404x |
| 32 | 384 | 73.773 | 64.961 | 1.1357x |
| 64 | 384 | 79.990 | 68.947 | 1.1602x |
| 128 | 384 | 90.133 | 70.738 | 1.2742x |
| 256 | 384 | 87.214 | 72.052 | 1.2104x |
| 1 | 768 | 27.192 | 27.255 | 0.9977x |
| 2 | 768 | 30.195 | 29.879 | 1.0106x |
| 4 | 768 | 34.962 | 30.926 | 1.1305x |
| 8 | 768 | 50.163 | 44.527 | 1.1266x |
| 16 | 768 | 71.012 | 59.986 | 1.1838x |
| 32 | 768 | 95.556 | 83.017 | 1.1510x |
| 64 | 768 | 126.374 | 94.061 | 1.3435x |
| 128 | 768 | 121.490 | 110.505 | 1.0994x |
| 256 | 768 | 124.268 | 124.453 | 0.9985x |

### A4W4 content change

The A4W4 tuned CSV is a 98% rename rather than a pure rename. Its only content
change is `(M=256, I=384)`, where the BM16 fused SwiGLU path replaces main's
BM32 generic path. The measured shape geomean is `1.2104x`, and its worst
routing case is `1.1321x`.

## Correctness

- routing IDs and weights match between main and PR in all 54 cases
- the two fallback shapes `(M=1,I=384)` and `(M=2,I=384)` remain
  bit-comparable between versions
- across the 48 cases using an exact PR config, maximum PR-vs-reference cosine
  difference is `0.02295`
- current main reaches `0.21715` on the same exact-PR subset because its
  `I=384` shapes use the fallback path
- prior MiniMax-M3 GSM8K 5-shot validation: `0.934` exact match

## Validation

- separate Aiter/JIT rebuilds for main and PR source trees
- real kernel traces confirm expected dispatch
- split-config uniqueness and activation dtype-boundary checks
- A16W4 AOT discovery and CSV parsing
- all 12 MiniMax inline-SwiGLU CSV rows pass strict `test_moe_2stage.py`
  validation against the quantization-aware reference
- cross-model guard: the only additional rows whose default sort policy changes,
  DSV3 `M=32,E=256,I=256/512`, improve by `1.2375x` and `1.1593x`
  respectively versus Opus across balanced/random/hotspot routing, with every
  individual route improving
- Python syntax, Black, Ruff 0.16.0, and `git diff --check`

Raw benchmark artifacts:
`/shared_nfs/xiaobizh/pr5404-current-main-ab-20260915/`


coderfeli (approved):

aiter/fused_moe.py
```diff
@@ -79,20 +79,26 @@ def _aux_uses_opus(output_aux, block_size, routed_rows=None, num_experts=None):
...
     if output_aux != AUX_SORT_OPUS:
         return False
     if block_size != 16:
         return True
+    if _MOE_SORT_BACKEND == "fused":
+        return False
     if routed_rows is None or num_experts is None:
         return False
+    if num_experts <= 256:
+        return routed_rows >= num_experts * 12
```
binding7012, line 101: This change requires a bit of caution.

PR #5405, opened 2026-09-10
[FlyDSL] Support runtime softmax scale in gfx950 FP8 FMHA

gbyu-amd:
The gfx950 FlyDSL FP8 FMHA kernel currently fixes the attention scale to `1 / sqrt(D)`. Consumers using a different layer scale, such as YaRN, must fold the correction into a quantization descale, and the AITER dispatch rejects nondefault scales.

Add `softmax_scale=None` to the direct FP8 API, using the BF16 API's parameter name and default. Pass it as a runtime FP32 scalar and apply `softmax_scale * q_descale * k_descale` to the logits inside FMHA. Dense and varlen dispatch and the large-tensor batch-slicing path forward the scale. This needs no descale mutation or additional GPU kernel launch, and scale values do not enter the build-cache key. Output and LSE use the same scale.

The existing online softmax requires a **positive, finite** scale: it takes maxima and applies masking before scaling. The direct API rejects zero, negative and non-finite scales; dispatch returns `None` for those configurations. Supporting them would require a separate change to the softmax algorithm.

This PR targets `main` and contains only FMHA scale support, dispatch changes and tests. The review follow-up removes the redundant `head_dim_runtime` launch argument and shares softmax-scale validation between the dispatch gate and direct FP8 API.

Validation:

- Review follow-up (`a859b8271`): **24 CPU tests passed, 252 GPU tests skipped**, with HIP/ROCR/CUDA device visibility disabled. Fake-tensor tests exercise the direct API and dense/varlen dispatch for default/custom/invalid scales; launch-boundary tests check default and custom scales in both run and compile argument preparation for D=128/192. No GPU execution or GPU compilation was performed. Black, Ruff and `git diff --check` pass.
- Prior validation after the rebase onto main: Black, Ruff and `git diff --check` pass; all 252 FMHA tests collect successfully with GPUs hidden (`AITER_AOT_IMPORT=1`). Collection is not a correctness run.
- At that rebase, `git range-diff` confirmed the feature patch was unchanged; all five changed files were byte-identical to the previously GPU-validated PR head. GPU tests were not repeated during the active TP8 workload.
- Added coverage checks nondefault scales, default/explicit-scale equivalence, repeated scale changes on a cached launcher, 12-head K3 shapes, dense/varlen, GQA, split-K, fully masked rows, zero-length KV entries, graph replay, dispatch and batch slicing.
- GPU validation before the main rebase, on a free gfx950 GPU with a fresh FlyDSL cache: **80 passed, 1 skipped** (the cross-device test requires a second visible GPU). This includes the custom-scale correctness, LSE, dispatch and graph-replay tests above.
- ATOM integration after replacing descale folding with direct `softmax_scale` forwarding: **112 tests passed**.
- Cached-chunk GPU traces for Q=70, H=12, D=192, Dv=128 and KV=16,384 / 11,008 confirm one FMHA kernel per call, with no added scale-preparation launch. Warmed CUDA-graph measurements were 278.99 / 190.98 microseconds per call (7 rounds of 100 replays); these are absolute measurements, not a speedup claim for this change.


aiter/ops/flydsl/kernels/fmha_gfx950/flash_attn_fp8_gfx950.py
```diff
@@ -505,6 +505,7 @@ def _prepare(
         stride_q_n=None,
         head_dim_runtime=None,
```
binding7012, line 506: head_dim_runtime now is a dead parameter, can be removed

aiter/ops/flydsl/fmha_kernels.py
```diff
@@ -280,8 +280,8 @@ def _fp8_gfx950_supported(
     ):
         return False
     qk_hdim = q.shape[-1]
-    if softmax_scale is not None and not math.isclose(
-        softmax_scale, 1.0 / math.sqrt(qk_hdim), rel_tol=1e-6
+    if softmax_scale is not None and (
+        not math.isfinite(softmax_scale) or softmax_scale <= 0
```
binding7012, line 284: both write in this position and flash_attn_func_fp8_gfx950.py, but they are not sync, can write a helper in flash_attn_func_fp8_gfx950.py, and import that helper in this pos

aiter/ops/flydsl/kernels/flash_attn_func_fp8_gfx950.py
```diff
@@ -409,6 +414,13 @@ def flydsl_flash_attn_fp8_func(
             f"flydsl_flash_attn_fp8_func: head_dim ({D}) must be >= 64 and a multiple of 32"
         )
 
+    if softmax_scale is None:
+        softmax_scale = D**-0.5
+    if not math.isfinite(softmax_scale) or softmax_scale <= 0:
```
binding7012, line 419: see the comment in the fmha_kernel.py

binding7012:
LGTM

coderfeli (approved):

Issue #4137, opened 2026-07-08
[Regression] Triton 3.7.0 causes ~4x regression in unified_attention on gfx950


mjkvaak-amd:
### Problem Description

`aiter/.github/scripts/install_triton.sh` currently installs `triton==3.7.0` on ROCm 7.2. With the same AITER checkout, same unified attention wrapper, same Triton kernel source, same input shape, and separate clean Triton caches, `triton==3.7.0` regresses `unified_attention` from ~4.44 ms to ~17.91 ms on gfx950.

Downgrading only Triton to 3.6.0 restores the expected performance.

**Proposed Fix: **
Until the Triton 3.7 regression is resolved, pin AITER’s managed Triton install to 3.6.0: do necessary changes in https://github.com/ROCm/aiter/blob/main/.github/scripts/install_triton.sh#L77-L80 . Optionally keep the wheel download helper aligned too: changes needed in https://github.com/ROCm/aiter/blob/main/.github/scripts/download_triton_wheel.sh#L26-L31 .

### Operating System

22.04.5 LTS (Jammy Jellyfish)

### CPU

AMD EPYC 9575F 64-Core Processor

### GPU

gfx950

### ROCm Version

7.2.3

### Steps to Reproduce

Start container:
```bash
docker run --rm -it --init \
  --name triton-unified-attention-perf-regression \
  --network host \
  --device /dev/kfd --device /dev/dri \
  --group-add video --shm-size 64g \
  --cap-add SYS_PTRACE --security-opt seccomp=unconfined \
  --entrypoint "" \
  --workdir /app \
  vllm/vllm-openai-rocm:nightly-a16dbd5b8572d4128be9f10b9dcff4999b594b25 \
  bash
```

Install AITER:
```bash
git clone https://github.com/ROCm/aiter --recursive
cd aiter
pip install -r requirements.txt
python3 setup.py develop
```

[repro_aiter_unified_attention_perf.py](https://github.com/user-attachments/files/29799114/repro_aiter_unified_attention_perf.py)
[ua_debug_qwen3_5_397b_prefill.json](https://github.com/user-attachments/files/29799115/ua_debug_qwen3_5_397b_prefill.json)

Run attached repro with Triton 3.7:
```bash
export TRITON_CACHE_DIR=/tmp/triton-cache-ua-37
rm -rf "$TRITON_CACHE_DIR"
python repro_aiter_unified_attention_perf.py \
[...]

PR #5409, opened 2026-09-10
[Triton/Gluon] Unified attention: opt-in prefill configs for head sizes 129+

cpersson-amd:
## Motivation

This PR covers a Triton regression raised in https://github.com/ROCm/aiter/issues/4137: Unified attention prefill regressed on gfx950 for head sizes 129 and above. On the Qwen3.5-397B debug shape the kernel went from about 4ms on Triton 3.6 to 16ms on 3.7 and 12ms on 3.8. This is due to three separate problems:

**Register pressure.** At `BLOCK_M=128` a head size in 129-255 pads to a 256-element accumulator, and the kernel needs about 470 VGPRs (it uses no LDS at all). `waves_per_eu=2` caps the budget at 256, so the excess spills to scratch inside the KV loop.

**An LLVM scheduler regression in Triton >= 3.7.** The same config that fit on 3.6 does not on 3.7+, which is what turned this from latent to visible. TTIR and TTGIR are byte-identical across the three versions and the LLVM IR instruction mix is the same; only the scratch size differs, so it is the AMDGPU machine scheduler producing longer live ranges, not anything in Triton's own passes.

**Config shadowing.** Because the lookup gives the leftmost axis priority, `D_GEQ_256` outranks `Q_GEQ_256`, so head sizes 256+ in prefill were picking a `BLOCK_M=16` entry meant for decode. That is why `d=256` was slower than `d=192` despite doing less work per program.

The goal is to fix all three without changing the default behavior so when a fix eventually lands in Triton, this change does not end up regressing the performance. The tuned configs trade occupancy for registers against one measured shape family, so they are exposed as an opt-in rather than made the default.

## Technical Details

Two files change: the gfx950 `DEFAULT.json` and `unified_attention_utils.py`.

**An opt-in config section.** A new `attn_2d.OPT_IN` section holds the retuned prefill entries and is merged into the lookup only when `AITER_TRITON_UA_FAST_PATH=1`. With the variable unset nothing changes: the section is never consulted, and all five pre-existing sections of `DEFAULT.json` are byte-identical to before. The core fix in these entries is `waves_per_eu=1`, which lifts the VGPR budget from 256 to 512 and eliminates the spill. It is scoped to head sizes 129+ (`D_GEQ_129` / `D_GEQ_256` keys) because sizes below that never spilled, and halving their occupancy target only costs them speed.

**The AMD backend workaround.** On top of the register fix, the opt-in entries can pass `amdgpu-sched-strategy=iterative-minreg` through Triton's `llvm_fn_attrs` option. This overrides the AMDGPU machine scheduler with a register-minimising strategy, recovering a further ~25% on large caches. It is not unconditionally good, so it sits in a `LARGE_KV` sub-block behind three guards:

- **Cache size.** Under `2**31` elements the backend indexes the cache with 32-bit buffer ops instead of 64-bit global loads. Against that different code the attribute costs about 3x rather than saving 25%. The guard counts elements, so if what the backend really tests is bytes, a borderline cache is declined rather than wrongly taken.
- **Dtype.** The overrides pay off for fp8 and cost bf16 about sixfold, so only the `DT_fp8_fp8` keys carry them.
- **Triton support.** `llvm_fn_attrs` is read off the backend's own options rather than inferred from a version number, and the block is skipped with a warning where the option does not exist.

## Test Plan

A standalone harness replays the Qwen3.5-397B debug shape and swaps configs through the lookup:

- **Coverage sweep** across head sizes 64-512, both prefill and decode, small (128-block) and large (4795-block) KV caches, fp8 and bf16, with the flag both unset and set. Each run reports the matched config entry, p50 latency, and the compiled kernel's scratch (register spill) bytes.
- **Guard tests** that exercise each of the three `LARGE_KV` gates: large cache with the option supported (applies), small cache (size guard declines), and a backend without `llvm_fn_attrs` (support guard declines). Also asserts the `LARGE_KV` sub-block never leaks into the launched config.
- **Numerics check** comparing the opt-in base and the large-KV overrides for bitwise-identical output, since the overrides only move scheduling and occupancy.
- **Regression guard on the default** by diffing every pre-existing section of `DEFAULT.json` against `HEAD` to confirm the flag-off path is untouched.

## Test Result

All pre-existing `DEFAULT.json` sections are byte-identical to `HEAD`; the only addition is `attn_2d.OPT_IN`. The guard tests pass (overrides apply only on a large fp8 cache with a supporting Triton, and never leak), and the opt-in base and large-KV overrides are bitwise identical.

gfx950, fp8, prefill. Small cache is 128 blocks, large is 4795. Each cell is p50 latency in ms (±0.1-0.2 ms run to run), with kernel scratch (register spill) in bytes in parentheses; "unchanged" means the opt-in section holds no entry for that head size, so the flag has no effect.

| head size | off, small cache (ms) | off, large cache (ms) | on, small cache (ms) | on, large cache (ms) |
|---|---|---|---|---|
| 64, 128 | 2.5-2.9 | 2.5-3.1 | unchanged | unchanged |
| 160, 192, 224 | 5.2-5.3 (132 B) | 12.1-12.2 (288 B) | 5.3 (0 B) | 4.6 (20 B) |
| 256 | 10.0 | 12.5 | 5.2 (0 B) | 4.7 (20 B) |
| 512 | 24.8 | 28.9 | unchanged | unchanged |

bf16, same shape, large cache is 2400 blocks. bf16 stays on the opt-in base config, since the large-KV overrides do not apply to it:

| head size | off, small cache (ms) | off, large cache (ms) | on, small cache (ms) | on, large cache (ms) |
|---|---|---|---|---|
| 192 | 11.7 (236 B) | 31.1 (920 B) | 7.1 (0 B) | 8.7 (0 B) |
| 256 | 25.0 | 34.8 | 7.2 (0 B) | 8.7 (0 B) |

Spill drops from up to 920 B to at most 20 B on every affected head size (0 B except on the fp8 large cache, where the minreg override leaves 20 B). The `d=256` bf16 row is the shadowing fix on its own; the `d=192` row is the register fix, worth 3.6x on the large cache. Decode is untouched under both settings, as are head sizes 64, 128 and 512.

## Notes and follow-ups

- The opt-in section has four entries where two would do. `D_GEQ_256` shadows `D_GEQ_129.Q_GEQ_256`, so the config has to be repeated at each `D` bound, and again split by dtype because only fp8 takes the `LARGE_KV` block.
- These configs were tuned against one shape family. `waves_per_eu=1` trades occupancy for registers, which is why this is opt-in rather than the default.

## Submission Checklist

### DIFF
--- a/aiter/ops/triton/configs/gfx950/triton/attention/unified_attention/DEFAULT.json
+++ b/aiter/ops/triton/configs/gfx950/triton/attention/unified_attention/DEFAULT.json
@@ -131,6 +131,50 @@
       "TILE_SIZE_MAX": 64
     }
   },
+  "attn_2d.OPT_IN": {
+    "D_GEQ_256.Q_GEQ_256.DT_fp8_fp8": {
+      "BLOCK_M": 128,
+      "num_warps": 4,
+      "num_stages": 2,
+      "waves_per_eu": 1,
+      "TILE_SIZE_MIN": 64,
+      "TILE_SIZE_MAX": 64,
+      "LARGE_KV": {
+        "num_stages": 1,
+        "waves_per_eu": 2,
+        "llvm_fn_attrs": "amdgpu-sched-strategy=iterative-minreg"
+      }
+    },
+    "D_GEQ_129.Q_GEQ_256.DT_fp8_fp8": {
+      "BLOCK_M": 128,
+      "num_warps": 4,
+      "num_stages": 2,
+      "waves_per_eu": 1,
+      "TILE_SIZE_MIN": 64,
+      "TILE_SIZE_MAX": 64,
+      "LARGE_KV": {
+        "num_stages": 1,
+        "waves_per_eu": 2,
+        "llvm_fn_attrs": "amdgpu-sched-strategy=iterative-minreg"
+      }
+    },
+    "D_GEQ_256.Q_GEQ_256": {
+      "BLOCK_M": 128,
+      "num_warps": 4,
+      "num_stages": 2,
+      "waves_per_eu": 1,
+      "TILE_SIZE_MIN": 64,
+      "TILE_SIZE_MAX": 64
+    },
+    "D_GEQ_129.Q_GEQ_256": {
+      "BLOCK_M": 128,
+      "num_warps": 4,
+      "num_stages": 2,
+      "waves_per_eu": 1,
+      "TILE_SIZE_MIN": 64,
+      "TILE_SIZE_MAX": 64
+    }
+  },
   "attn_3d": {
     "D_GEQ_512": {
       "BLOCK_M": 16,
--- a/aiter/ops/triton/utils/unified_attention_utils.py
+++ b/aiter/ops/triton/utils/unified_attention_utils.py
@@ -25,15 +25,26 @@ that exists: LEQ bounds ascending, then GEQ descending, then "any". So the
 leftmost axis wins: D before Q is what makes head_size outrank max_seqlen_q.
 Dtypes fall back too: DT_fp8_fp8, then DT_fp8_any, DT_any_fp8, then "any".
 
+Because the leftmost axis wins outright, D_GEQ_256 beats D_GEQ_129.Q_GEQ_256 at
+head_size 256 despite saying less. An entry narrowing a rightward axis has to be
+repeated at every bound of a leftward one, which is why some keys differ only in
+that.
+
 A section with no axes, like reduce above, is just a config.
 
 Tile size and number of splits (segments) are derived from the following parameters:
 TILE_SIZE_MIN/MAX, and MIN_SEGMENTS/MAX_SEGMENTS/SEGMENTS_PER_CU.
+
+An "<op>.OPT_IN" section holds entries that exist only when the user opts in
+(_load); a LARGE_KV block inside one applies only above a KV cache size
+(_resolve_large_kv).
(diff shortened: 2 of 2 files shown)


cagrikymk:
@cpersson-amd Did you have a chance to look at why the register spill happens in newer triton versions?

cpersson-amd:
> @cpersson-amd Did you have a chance to look at why the register spill happens in newer triton versions?

I think the spills are due to a change in the LLVM AMDGPU scheduler. I asked Claude to look for changes in LLVM between Triton 3.6 and 3.7 that could cause this regression. This PR https://github.com/llvm/llvm-project/pull/147653 was pointed out as the possible regression point. 

This was briefly tested by running the **identical kernel IR** through Triton 3.7's pinned LLVM (`ac5dc54d`, `llc` v23.0.0git), toggling only the #147653 mechanism via its knob (`amdgpu-scheduler-pending-queue-limit`, where `0` disables the pending-queue examination):

| pending-queue-limit | ScratchSize | spill stores |
|---|---|---|
| 0 (disabled = pre-#147653) | **84 B** | 9 |
| 1 | 84 B | 9 |
| 2 | 88 B | 8 |
| 4 | 148 B | 12 |
| 8 / default | **168 B** | 13 |

This PR was also suggested as a potential fix for this issue: https://github.com/llvm/llvm-project/pull/213116.

PR #5414, opened 2026-09-10
[Triton/Gluon] One shared helper for opt-in @triton.autotune config lists

Boss2002n:
autotune_configs in utils/tuned_config_utils replace the chunk_delta_attn / gated_delta_rule copies and the attn_res/sparse flag, and the list decorators in gated_delta_rule/utils and chunk_delta_attn/gate now go through it, so no kernel benchmarks configs unless <FAMILY>_TRITON_AUTOTUNE=1.


aiter/ops/triton/utils/tuned_config_utils.py
```diff
@@ -23,6 +24,28 @@
 )
 
 
+def autotune_enabled(family: str) -> bool:
+    """``<FAMILY>_TRITON_AUTOTUNE=1`` opts a kernel family into runtime tuning; off by default."""
+    return os.getenv(f"{family}_TRITON_AUTOTUNE", "0").lower() in (
```
brunomazzottiamd: **Suggestion for hardening the code:** Add a `.strip()` here to remove trailing whitespaces from the environment variable.

aiter/ops/triton/utils/tuned_config_utils.py
```diff
@@ -23,6 +24,28 @@
...
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
+
+def autotune_configs(
```
brunomazzottiamd, line 37: **Suggestion for hardening the code:** Add an assertion to guarantee that `configs` isn't an empty list. If it's an empty list then Triton autotuner will complain and `configs[0]` will raise `IndexError`.

brunomazzottiamd (changes requested):
I have two small suggestions to improve the common utility code. Please share this PR with the affected kernel owners as well.

brunomazzottiamd (approved):

PR #5424, opened 2026-09-11
[Triton/Gluon] no CSV under pytest, no leaked default device

Boss2002n:
The test wrote run_pa_gluon_test.*.csv into cwd on every pytest run and set torch's default device
to cuda at import for the whole session


op_tests/triton_tests/test_pa_decode_gluon.py
```diff
@@ -2030,11 +2033,11 @@ def compute_column_mean(col_data):
...
     total_errors = results_df["err_gluon"].sum()
     if total_errors > 0:
-        logger.warning(
-            "\nTests failed! %d test case(s) exceeded the error threshold. ",
+        logger.info(
+            "\nTests failed! %s test case(s) exceeded the error threshold. ",
             total_errors,
         )
-        logger.warning("Please check rows with non-zero err_gluon in %s.", output_file)
+        logger.info("Please check rows with non-zero err_gluon in %s.", output_file)
```
brunomazzottiamd: This log message may be referring to a non-existing file.

I think we should do the following in line `1949`:

```python
# Unit tests only check pass/fail; only the CLI (__main__) run keeps the CSV report.
write_output_file: bool = "PYTEST_CURRENT_TEST" not in os.environ

if write_output_file:
    results_df.to_csv(output_file, index=False)
    logger.info("\nResults saved to %s", output_file)
```

And then, in line `2040`:

```python
if write_output_file:
    logger.info("Please check rows with non-zero err_gluon in %s.", output_file)
```

op_tests/triton_tests/test_pa_decode_gluon.py
```diff
@@ -1943,10 +1944,12 @@ def parse_arg_and_run_test(sample_rate0: float | None = None):
         ps_options,
     )
 
+    torch.set_default_device(prev_default_device)  # per-case runs change it; restore
```
brunomazzottiamd: Default device may not be restored with `torch.set_default_device(prev_default_device)` if `run_multi_pa_gluon_test` throws an exception or fails on an `assert`. This will leave the test process dirty, with mutated global state, and may affect other tests in the same runner shard.

We need a setup + guaranteed cleanup structure, like the context manager from [`test_mha_dao_ai.py`](https://github.com/ROCm/aiter/blob/main/op_tests/triton_tests/attention/test_mha_dao_ai.py#L22):

```python
import contextlib

@contextlib.contextmanager
def _with_default_device(device):
    prev_device = torch.get_default_device()
    try:
        torch.set_default_device(device)
        yield
    finally:
        torch.set_default_device(prev_device)

with _with_default_device("cuda"):
    run_multi_pa_gluon_test(...)
```

brunomazzottiamd (changes requested):

brunomazzottiamd (approved):

PR #5431, opened 2026-09-11
[HIP] feat(custom_ar): add torch.symm_mem transport(mori) for gfx1250

kawhil-amd:
## Summary

- **torch.symm_mem transport**: Add `torch.distributed._symmetric_memory` (backed by `mori.allocator`) as a new transport for custom all-reduce on gfx1250. Opt-in via `AITER_CUSTOM_AR_USE_SYMM_MEM=1`. Replaces manual IPC handle exchange (TCPStore / Unix socket VMM fd) with `symm_mem.rendezvous()` which handles fabric handles or POSIX fds internally. Reuses existing gfx1250 VMM-style C++ ops (pointer-list interface) — no C++ kernel changes for this part.
- **Collective P2P probe**: Before init, allocates a small buffer + rendezvous across all peers. mori's rendezvous validates P2P via `hipMemSetAccess(RW)` for each peer — all ranks agree on pass/fail, safe fallback to VMM/IPC with no hang risk.
- **world_size 6/8**: Extend gfx1250 template instantiations from {2,4} to {2,4,6,8} across all dispatch points (allreduce, allgather, reduce_scatter). Introduce `DISPATCH_NGPUS_1250` macro to unify the dispatch pattern. No algorithmic changes — kernels are simple `for(i<ngpus)` loops and `Signal`/`RankData` structs already have 8 slots.

## Test plan

- [ ] `AITER_CUSTOM_AR_USE_SYMM_MEM=1 torchrun --nproc_per_node=2 op_tests/multigpu_tests/test_custom_allreduce.py`
- [ ] `torchrun --nproc_per_node=8 op_tests/multigpu_tests/test_custom_allreduce.py` (world_size=8)
- [ ] Verify fallback: uninstall mori, set env var, confirm warning + IPC path works
- [ ] Without env var: existing IPC/VMM paths unchanged


TennyWang1223 (changes requested):
csrc/include/custom_all_reduce_gfx1250.cuh:122 — LL scratch buffer overflow at world_size 6/8

kLLMaxRanks is still 4, but this PR opens the LL fast path to world_size_ <= 8 (line ~1428). The shared meta buffer is sized meta_size() = kLLScratchOffset + llScratchBytes(), and llScratchBytes() = 2 * kLLMaxRanks * kLLPackCapacity * 16 only reserves 4 ranks of scratch. But ar_ll_gfx1250 writes at bankOffPkts + rank*slot with bankOffPkts = (flag&1) * ngpus * slot, indexed by the template ngpus. For ngpus=8 the max write offset is (2*ngpus-1)*slot = 15*slot, while only 2*kLLMaxRanks*slot = 8*slot is allocated — a ~2x out-of-bounds write into every peer's meta buffer (ngpus=6 also overflows: 11*slot > 8*slot). This corrupts adjacent device memory regardless of transport (symm_mem's meta tensor uses the same meta_size_gfx1250()).

Fix: bump kLLMaxRanks to 8 and update the stale world_size <= 4 comment on line 121.

csrc/include/custom_all_reduce_gfx1250.cuh:611 — ag_gfx1250_warpsplit_unroll4 out-of-bounds at world_size 6

tnum_gpu = 256 / ngpus doesn't divide evenly for ngpus=6 (= 42, 6*42=252). With blockDim=256, threads 252–255 compute warp_id = 6, then read ptrs[6] (valid range 0–5) and write result + 6*size — both out of bounds. Since  _SUPPORTED_WORLD_SIZES now includes 6, this all_gather kernel is broken at 6 GPUs (8 is fine: 256/8=32 divides evenly). The allreduce path (naive/LL) uses % ngpus and is unaffected.

Fix: guard the tail threads (if (warp_id >= ngpus) return;) or restrict this kernel's supported world sizes to powers of two that divide 256.

kawhil-amd:
> csrc/include/custom_all_reduce_gfx1250.cuh:122 — LL scratch buffer overflow at world_size 6/8
> 
> kLLMaxRanks is still 4, but this PR opens the LL fast path to world_size_ <= 8 (line ~1428). The shared meta buffer is sized meta_size() = kLLScratchOffset + llScratchBytes(), and llScratchBytes() = 2 * kLLMaxRanks * kLLPackCapacity * 16 only reserves 4 ranks of scratch. But ar_ll_gfx1250 writes at bankOffPkts + rank_slot with bankOffPkts = (flag&1) * ngpus * slot, indexed by the template ngpus. For ngpus=8 the max write offset is (2_ngpus-1)_slot = 15_slot, while only 2_kLLMaxRanks_slot = 8_slot is allocated — a ~2x out-of-bounds write into every peer's meta buffer (ngpus=6 also overflows: 11_slot > 8*slot). This corrupts adjacent device memory regardless of transport (symm_mem's meta tensor uses the same meta_size_gfx1250()).
> 
> Fix: bump kLLMaxRanks to 8 and update the stale world_size <= 4 comment on line 121.
> 
> csrc/include/custom_all_reduce_gfx1250.cuh:611 — ag_gfx1250_warpsplit_unroll4 out-of-bounds at world_size 6
> 
> tnum_gpu = 256 / ngpus doesn't divide evenly for ngpus=6 (= 42, 6_42=252). With blockDim=256, threads 252–255 compute warp_id = 6, then read ptrs[6] (valid range 0–5) and write result + 6_size — both out of bounds. Since _SUPPORTED_WORLD_SIZES now includes 6, this all_gather kernel is broken at 6 GPUs (8 is fine: 256/8=32 divides evenly). The allreduce path (naive/LL) uses % ngpus and is unaffected.
> 
> Fix: guard the tail threads (if (warp_id >= ngpus) return;) or restrict this kernel's supported world sizes to powers of two that divide 256.

Could you help to check it again? The issue you mentioned have corrected .

TennyWang1223 (dismissed):
LGTM

valarLip:
Reviewed this against the tree at head. Four findings I would treat as blocking;
each was traced to the line rather than inferred. Happy to post the remaining
non-blocking ones (world_size 6 silently dropped, a stream-ordered `zero_()`
before a host-side rendezvous, `get_gfx()` used for a runtime dispatch decision,
teardown/leak on the symm path, the unconditional meta-size doubling) if useful.

---

### 1. The world-size expansion is unreachable, and TP=8 now crashes at startup

`csrc/kernels/custom_all_reduce_gfx1250.cu:40` and `:67` still carry

```cpp
if (world_size > 4)
    throw std::invalid_argument("gfx1250 custom allreduce: world size > 4 is not supported");
```

Python-side the gate is now open, so on gfx1250 TP=8 the constructor passes
`_SUPPORTED_WORLD_SIZES`, sets `self.disabled = False`, and calls
`_ops_init_custom_ar` — which throws, uncaught, out of `CustomAllreduce.__init__`.
Engine startup dies. Before this PR the same config hit the graceful
"unsupported world size" warning and fell back to RCCL.

The consequence for the PR itself: the new `DISPATCH_AR` case 8 and `DISPATCH_AG`
cases 16/32 can never execute, so the headline change is dead code until these
two guards move.

### 2. `_SUPPORTED_WORLD_SIZES` is arch-agnostic, so old archs lose their fallback

`custom_all_reduce.py:766` is a `ClassVar` shared by every architecture:

```python
_SUPPORTED_WORLD_SIZES: ClassVar[list[Any]] = [2, 4, 8, 16, 32]
```

The 16/32 entries are there for gfx1250 all-gather, but the gate at `:873` is the
only thing standing between an old-arch rank and `csrc/kernels/custom_all_reduce.cu:42`:

```cpp
if (world_size > 8)
    throw std::invalid_argument("world size > 8 is not supported");
```

So MI300 at TP=16 now passes the gate, sets `disabled = False`, enters `_init_ipc`,
and throws at startup where it previously logged and fell back. The new
`should_custom_ar` `> 8` guard does not help — it gates dispatch, never
construction.

### 3. `_SymmMemBufferProxy` registers a one-time copy, so peers reduce stale data

`custom_all_reduce.py:750-761`:

```python
buf = symm_mem.empty(size, dtype=torch.uint8, device=ca.device)
buf.copy_(tensor.reshape(-1).view(torch.uint8)[:size])   # once, at registration
hdl = symm_mem.rendezvous(buf, ca.group)
return list(hdl.buffer_ptrs)
```

The caller then passes the *original* `inp.data_ptr()` as the local pointer, and
`custom_all_reduce_gfx1250.cuh:1149` (and `:1160`) binds the two differently:

```cpp
data.ptrs[i] = (i != rank_) ? (void*)all_ptrs[i] : self;
```

Slot `rank_` is the live input; every other slot is peer *j*'s frozen
registration-time copy, and nothing re-copies afterwards. On a TP=4 decode, every
all-reduce past warm-up sums fresh local activations against iteration-0 peer
values: **wrong logits, no crash, no warning** — the worst failure shape of the
four.

The pre-existing `_GFX1250BufferProxy` has the identical structure, so whatever
fixes this needs to sweep both.

### 4. Three `*_sync_latency` entry points were missed by the macro conversion

Every other `world_size_ == 2 ? 2 : 4` site in
`csrc/kernels/custom_all_reduce_gfx1250.cu` became `DISPATCH_AR_NGPUS_1250`.
These three did not — `start_sync_latency` (`:468`, dispatching at `:475`/`:478`),
`end_sync_latency` (`:482`, `:489`/`:492`) and `two_sync_latency` (`:496`,
`:503`/`:506`) still read:

```cpp
int ws = fa->world_size_;
if (ws == 2) ...<2>... else ...<4>...
```

At `world_size = 8` this launches the 4-rank barrier: each rank writes only
`sg.signals[0..3]`, ranks 4..7 never have their `Signal::start` row set, and every
rank spins forever. A permanent GPU hang rather than a throw, which is
considerably harder to diagnose than #1.

---

On #1/#2 the shape of the fix matters as much as the fix: right now a world size
the C++ side rejects is discoverable only by catching its exception. A predicate
the Python gate can ask — per-arch, returning the reason it cannot serve — would
let both the constructor and `should_custom_ar*` read the same answer instead of
each carrying its own copy of the limits.

Related, and the reason I would put the cap in one place: the `> 8` guard added to
`should_custom_ar` was not added to `should_custom_ar_bytes`, which shares
`_fits_custom_ar_size`. Putting it in `_fits_custom_ar_size` gives it to all four
predicates at once.

kawhil-amd:
@valarLip  Thanks for the thorough review. #1, #2, and #4 are fixed in the latest push (two separate commits).

Regarding **#3** (`get_external_ipc_meta` one-time copy):

This is a pre-existing pattern shared by both the VMM and symm_mem paths — not introduced by this PR. The normal allreduce/allgather hot path does not go through `get_external_ipc_meta`:

- The primary input buffer (`_symm_input_tensor`) is allocated via `symm_mem.empty()` + `rendezvous()` in `_init_symm_mem`. All ranks share the same live device memory directly. Data is copied into this buffer before each kernel launch, so peers always read fresh values.
- `get_external_ipc_meta` is only called from `register_input_buffer` / `register_output_buffer` for additional external tensors (e.g. CUDA graph capture buffers). The IPC path does not have this issue — it takes a hipIpc handle on the original tensor's physical memory, no copy involved. The VMM and symm_mem paths copy into a new shared buffer because they cannot directly export an arbitrary tensor's backing memory.
- A proper fix would require the caller to allocate external tensors on symm_mem/VMM from the start (avoiding the copy entirely), which is an allocator-level change beyond the scope of this PR.

Happy to hear if you see a scenario where the hot path actually hits `get_external_ipc_meta` — in that case this would need to be addressed here.

valarLip (changes requested):
Reviewed at `12a4486`.

Three things I chased and want to clear off the table so they don't get re-litigated: the AR kernel does **not** spill registers at `ngpus=8` (measured `ScratchSize=0` on gfx942/gfx1201 proxy targets); the symm_mem meta is not uncached memory (`_init_gfx1250` was already plain VMM cached, so that's not something this PR introduced); and `register_output_buffer` does not lose writes, because the gfx1250 kernel never dereferences `_output_dp`.

The probe-and-fall-back design is the right shape. My concerns are that the probe is rank-local, that it validates an allocation seven orders of magnitude smaller than the one it commits to, and that falling back leaves two guards already evaluated against the wrong transport.

### all2all.py breaks FlyDSL EP on gfx1250, and is unrelated to this PR

`FlyDSLAll2AllManager._init_mori_shmem` now returns early on gfx125x, but the FlyDSL EP op allocates every P2P buffer from mori's shmem heap: `flydsl_dispatch_combine_intranode_op.py` calls `mori_shmem_create_tensor` unconditionally at :438 and :696-712, `ms.shmem_barrier_all()` at :434/:573/:662, and `ms.shmem_ptr_p2p()` at :362. With `shmem_torch_process_group_init` never executed, this hits `assert ptr != 0, "mori_shmem.shmem_malloc failed"` in `mori/shmem/tensor_utils.py:46`, or libmori_shmem's "Shmem state is not initialized". So `all2all_backend=flydsl` is simply broken on gfx1250.

The `MoriAll2AllManager` half is fine — mori's `_ep_comm()` defaults to cco on gfx125x and genuinely doesn't need shmem. But the skip here keys on arch alone and doesn't follow mori's own `MORI_EP_COMM` variable, so a user setting `MORI_EP_COMM=shmem` falls back to the shmem backend while aiter has already skipped init, and hits the same assert. The FlyDSL class docstring at :128 and the assert text at :174 still say "FlyDSL still uses mori's shmem heap", which now contradicts the code — and the call-site comment that recorded the coupling was deleted in this PR.

I'd suggest splitting these two hunks into their own PR. They're a different subsystem from symm_mem custom AR, and leaving them here means that when someone bisects a gfx125x EP hang later, they land on a commit titled "symm_mem custom AR".

### The probe is rank-local, so a single failing rank deadlocks startup

`_probe_symm_mem` wraps a collective rendezvous in a bare `except Exception` and returns a rank-local boolean with no cross-rank reduction — the docstring's claim that every step is collective so no rank can hang is exactly inverted. If rank 3's `import mori.allocator` fails, or `set_backend` raises, or `symm_mem.empty(64)` OOMs, the exception is swallowed, logged at debug level, and the rank returns False and proceeds to `_init_ipc`. Ranks 0-2 return True and block forever in `_init_symm_mem`'s `symm_mem.rendezvous(self._symm_meta_tensor, self.group)`. Startup deadlock with no visible diagnostic. `all_peers_mapped` is a purely local judgement too and splits the same way. An `all_reduce(MIN)` on the probe result before branching closes it.

### The probe validates 64 bytes and commits to ~1.15 GiB, with no fallback

A successful `symm_mem.empty(64)` commits to the symm path, and then `_init_symm_mem` asks the mori symmetric heap for meta (~128 MiB) plus `max_size` (1 GiB by default) at :1043/:1049 — with no `try`/`except`. Heap exhaustion, a granularity or alignment rejection, or a rendezvous failure at that size propagates out of `CustomAllreduce.__init__` and aborts engine startup, which is precisely the scenario the probe-and-fallback design exists to prevent. By then `self.disabled` is already False, so a caller that catches the exception holds a half-initialized object with no `_pool` and no `_ptr`, and the first `all_reduce` raises `AttributeError`. Either probe at the real size, or route `_init_symm_mem` through the same fallback path.

### Two guards are evaluated against a transport that may not be the final one

The cross-node early exit is bypassed and never rechecked. At :857, `not self._same_node and not self._use_symm_mem` is false whenever `AITER_CUSTOM_AR_USE_SYMM_MEM=1`, so the group is not disabled and :932 sets `disabled=False`. On a non-gfx1250 target the probe returns False immediately at :1012, and :996-999 falls through to `_init_ipc`, which calls `hipIpcOpenMemHandle` on a handle exported by a different host — an error, or a garbage pointer. Before this PR the same combination printed "spans across nodes" and disabled cleanly. `_same_node` is stored and then never read again.

The `expandable_segments` IPC guard has the same shape. At :950-954, `not self._use_symm_mem` is false while `_use_symm_mem` is still True, so `enable_register_for_capturing` is not forced off; the probe then fails and we land in `_init_ipc`, and graph capture calls `hipIpcGetMemHandle` on an expandable-segment pointer — the exact failure that guard's comment describes, reintroducing #4174. gfx1250 is saved only incidentally, by the unconditional `enable_register_for_capturing = False` at :942. This has to be computed after the transport is finally decided.

Relatedly, `symm_mem.set_backend("MORI")` is a process-global switch flipped inside the probe and never restored when the probe fails and we fall back to VMM or IPC; `_init_symm_mem` at :1040 then sets it a second time. After any rank's probe failure, other symmetric-memory users in the same process silently get the MORI allocator. Same class of issue: `_select_ops()` is called twice, and the `mori.allocator` + `set_backend` lines are duplicated verbatim between probe and init — once those two copies drift, you probe one configuration and run another.

### world_size: the gate, the C++ check, and the actual capability disagree

The new `world_size > 8` limit lives only in `_fits_custom_ar_size`, but `custom_fused_qknorm_ar` (:2011) and `custom_fused_qknorm_ar_rope` (:2076) check only `self.disabled` and bypass it. Before this PR, ws=16/32 could not be constructed on gfx1250 at all — the deleted `gcnArchName` check raised. Now `_SUPPORTED_WORLD_SIZES_GFX1250` admits 16 and 32, `disabled` stays False, and the fused qknorm path dispatches the kernel and throws `RuntimeError` mid-forward instead of falling back as the guard's comment ("larger world sizes use RCCL") promises. The root cause is that a *dispatch capability* limit was placed inside a predicate named `_fits_custom_ar_SIZE`, so it only covers one entry point.

More broadly, admitting 16/32 advertises support that isn't there: above ws=8, `all_reduce` and `reduce_scatter` silently return None and fall to RCCL (:1378, :1491) while the full meta and `max_size` pool are still allocated, and only `all_gather` runs a custom kernel. There's no warning anywhere — the only clue is a source comment. Note the `RuntimeError` this PR deletes said "RCCL fallback is also not available on this platform", which directly contradicts the new behaviour; the PR doesn't say which one is right. I'd either warn once at init or drop 16/32 from the list until AR/RS actually support them.

On the C++ side the init check moved from `> 4` to `> aiter::kMaxNgpus`, and `kMaxNgpus = kMaxNgpusAg = 32` is far wider than the {2,4,8} that AR/RS dispatch supports — so rejection moved from construction time to the first collective. Any even ws ≤ 32 (6, 10, 12, …) constructs successfully and then hits `DISPATCH_AR_NGPUS_1250`'s default arm with `std::runtime_error("unsupported world_size N")`. The in-repo Python gate blocks this today (6 isn't in the whitelist), so it's latent rather than live — but `torch.ops.aiter.init_custom_ar_gfx1250` is a public binding that downstream vLLM/SGLang call directly. The message is misleading too: it says "world size > 32 is not supported" when the usable AR set is {2,4,8}. The check should mirror the dispatch set (`kMaxNgpusAr` for AR).

### Two unconditional memory increases, with the comments now wrong

`kLLMaxRanks` 4 → 8 doubles `llScratchBytes()` from 64 MiB to 128 MiB, and `meta_size()` applies it unconditionally at every world size — so TP=2 and TP=4 pay 64 MiB per GPU for nothing. The arithmetic is `2 * kLLMaxRanks * kLLPackCapacity(=524288) * 16 B`, i.e. 67,108,864 → 134,217,728 independent of ws, and all three init paths allocate and zero that. An 8-GPU node goes 512 MiB → 1 GiB, while at ws=2 the kernel's `bankOffPkts = (flag & 1) * ngpus * slot` touches a quarter of it. `meta_size()` not taking `world_size` is the root cause. Three comments are now inaccurate and actively misleading: `cuh:142`'s `// 32768` (the real value is 524288 — this one misled one of my own verification passes into computing the total as 8 MiB), `cuh:158-160`'s "4 MiB at the 256 KiB / 4-rank defaults", and `custom_all_reduce.py:1171`'s "Signal (~34 KB) + LL staging scratch (~4 MiB)".

Separately, `kMaxNgpus = kMaxNgpusAg = 32` is used to size `Signal`, `RankData` and `RankSignals`, though AR/RS top out at 8. `Signal` grows 34,816 B → 133,120 B, and `RankData`/`RankSignals` grow 64 B → 256 B — and those two are passed **by value on every kernel launch**. Every world size including ws=2 pays ~97 KiB of extra meta, plus 256 B of extra kernel arguments per AR/RS/AG launch, which is not free on a few-microsecond decode path. The 8 MiB rank_data arena also drops from 131072 registered tuples to 32768, while the comment at :964-968 still says "world_size is at most 8 … enough for 131072 such tuples". Only AG needs 32 lanes; the AR/RS structures should be sized by `kMaxNgpusAr`.

### AG index arithmetic overflows now that ngpus can be 32

`ag_gfx1250_scalar`'s `result[gpu_idx * size + idx]` (:488), `ag_gfx1250_naive_vec`'s `size * rank_idx + idx` (:520) and `ag_gfx1250_lastdim`'s `(ngpus * y + rank_idx) * last_dim_size + x` (:604) are all `int` and all scale with ngpus, and `custom_all_reduce_gfx1250.cu:290` narrows an `int64_t` numel to `int`. Raising the ngpus ceiling from 4 to 32 shrinks the safe output size by 8x. `should_custom_ag`'s `max_size/(ws*2)` limit protects the default pool, but `custom_all_gather` (:1589) and `all_gather_reg` (:1523) don't consult `should_custom_ag` at all — they don't even check `disabled`. At ws=32, total output elements exceed 2^31 at roughly 134 MB of bf16 input per GPU (where at ws=4 the threshold was ~1 GiB), giving negative offsets and silent out-of-bounds writes. `size_t` fixes it.

Also on that kernel: `ag_gfx1250_warpsplit_unroll4`'s `constexpr int tnum_gpu = 256 / ngpus` leaves 8 lanes at ngpus=32 — less than a wave, so "warpsplit" stops being an accurate name — and there's no `static_assert` that `blockDim == 256`.

### Benchmark-only kernels roughly double the JIT TU

Three sync-latency microbenchmark entry points were switched to the AG macro (2/4/8/16/32), but they measure the AR/RS-side synchronization primitives, whose macro tops out at 8. Together with AG's own 16/32 instantiations, template instances in the single JIT TU go from roughly 78 to 153 — nine pure-benchmark kernels get 16/32 instances compiled into the shipped module for code paths production can never reach. This module is JIT-compiled on first call, on the engine-init path of every rank, so ~2x instances shows up directly as a user-visible first-call compile wall and `.so` size. Switching those three back to the AR macro removes nine instances with no behavioural change.

While measuring that: on proxy targets the AG kernel at ngpus=32 shows 272 B/lane of `ScratchSize`, because the `ptrs[32]` pointer table gets demoted to scratch — so every peer load reads scratch first. That's the real performance cliff for AG at ws=32, and the PR carries no measured numbers at ws=8/16/32 at all.

### Lifecycle

`get_external_ipc_meta` copies the user tensor into a freshly allocated symm buffer and returns peer pointers to *that copy*, while the local key remains `inp.data_ptr()`. The C++ `register_input_buffer` stores `data.ptrs[i] = (i != rank_) ? all_ptrs[i] : self` (cuh:1149) and the kernel re-reads those peer pointers every round, so each rank sees its own data fresh and every peer's as a snapshot taken at registration — divergent results across ranks, with no error. Each call also appends a fresh full-size buffer to `ca._symm_ext_hdls` that is only released at `close()`. And `symm_mem.rendezvous` is collective but sits inside a per-rank API, so any mismatch in registration order or count between ranks deadlocks — `aiter/ops/communication.py:68` already records that the VMM version "deadlock[s] in its vmm_exchange rendezvous" for this reason. Most of this is inherited from the unchanged `_GFX1250BufferProxy` (:675) rather than introduced here, and nothing in-repo calls `register_*_buffer` today, so it's latent — but the new copy propagates it.

`close()` delattrs three handle attributes, but `_symm_meta_tensor`, `_symm_input_tensor` and the `self._pool` holding them stay alive, so the "peer mappings are torn down" in the comment doesn't happen. `_SymmMemBufferProxy` holds strong references to both tensors and back to `ca`, so after close the ~128 MiB meta and the `max_size` input and their peer mappings remain resident, plus a `ca._pool._ca is ca` cycle that defers GC. Repeated create/close in one process — multiple engines, TP reconfiguration, a test loop — leaks a full set each time until the symmetric heap is exhausted, which is exactly what the comment claims to prevent. Either drop both tensors and `_pool` in `close()`, or fix the comment.

### Smaller

The new blank line at `custom_all_reduce.py:130` fails `black --check` (I ran it: `would reformat`), and CONTRIBUTE.md requires pre-commit to pass.

There are no test changes in this PR. `test_custom_allreduce.py`'s `--tp-size choices=[2,4,6,8]` still offers 6, which is no longer in the gfx1250 list, so `-t 6` passes green while never touching custom AR at all.

On duplication: `_SymmMemBufferProxy` and `_GFX1250BufferProxy` are near-identical line for line; the two `_init_mori_shmem` methods carry a copy-pasted gfx125x guard whose docstrings have already drifted apart; and the probe-failure branch re-copies the transport `elif` chain below it. The "raw pointer transport" test `_use_symm_mem or _use_vmm` appears in four places. all2all.py introduces a third way of detecting gfx1250 (`startswith("gfx125")`), reading a different source than `_detect_gfx1250()`, which also means the `AITER_CUSTOM_AR_DISABLE_GFX1250` escape hatch — documented as end-to-end — doesn't reach all2all. And `AITER_CUSTOM_AR_USE_SYMM_MEM` has no named constant or comment the way MAX_SIZE/MIN_SIZE do, and isn't documented anywhere.

### Summary

I'd want the rank-local probe made collective and the two stale guards (cross-node, expandable_segments) recomputed after the transport is final, since between them a single unlucky rank turns into either a startup deadlock or a cross-node `hipIpcOpenMemHandle`. The 64-byte probe should validate what it's actually committing to. `world_size` 16/32 should either warn or wait. And I'd really like to see the all2all.py hunks split out — both because they break FlyDSL EP on gfx1250 as written, and because they don't belong to this change.

PR #5439, opened 2026-09-11
Add FP8 MoE intermediate option for Kimi-K3 a4w4

amd-wsung102:
## Summary

This PR adds an opt-in dataflow that moves the MoE intermediate to fp8. a4w4 quantizes twice per MoE layer, and the two tensors are not alike. The first is the MoE input, which on Kimi-K3 is a down-projected, RMSNormed latent and is Gaussian. The second is the SiTUv2 output that feeds gemm2, and it is produced by a multiplicative nonlinearity, so its tails are far heavier than those of the tensor it came from. Measured on the real MXFP4 checkpoint at layer 1, the gemm1 output has kurtosis 3.08 and a per-32-block amax/mean of 2.93 at the median and 4.60 at p99, while the intermediate it produces has kurtosis 13.30 and 5.25 / 11.05. MXFP4 handles the second case worst, because one large element inflates the e8m0 scale shared by its whole block.

That intermediate carries most of the a4w4 error. This PR adds an opt-in dataflow that keeps fp4 activations into gemm1 and moves only the intermediate to fp8, so gemm2 becomes fp8 activation against MXFP4 weight.

**Accuracy improvement:**

* 1.318x lower relative L2 of the MoE block output, averaged over all 92 Kimi-K3 MoE layers, improving every layer with no regressions.
* a4w4 moves from 37.7x to 28.6x the a16w4 error.
* fp8 captures 98% of the benefit a bf16 intermediate would give, so there is almost nothing left to gain by spending 16 bits there.

image

## Relevant Files

* `aiter/fused_moe.py` — 10 lines in `get_2stage_cfgs`, immediately before the dispatch log line so the logged kernel names reflect the arm actually selected. Two conditions gate the rewrite: the activation dtype must be fp4x2, and stage 1 must be a `flydsl_moe1_` name.

### Moving the intermediate to fp8

The whole change is a kernel-name rewrite at dispatch. `parse_flydsl_v2_gemm2_kernel` reads the activation dtype out of a `flydsl_moe2_layout_*` name and `get_2stage_cfgs` then assigns it to stage 1's `out_dtype`, which also sets `fuse_quant`. Selecting an `afp8` stage 2 therefore moves the entire intermediate path in one step: the stage-1 epilogue switches from `rocdl.cvt_scalef32_pk_fp4_f32` to `rocdl.cvt_pk_fp8_f32`, writes one byte per element instead of half a byte, and keeps the same sorted e8m0 scale buffer that `mxfp4_moe_gemm2` already consumes for either dtype. `skip_inter_quant` stays true because the name still contains `_moe2_layout_`, so no host re-quantization kernel is introduced.

### Restricting the rewrite to the mixed_moe stage-1 family

The server selects two different stage-1 families for a4w4 on the same model: `flydsl_moe1_afp4_wfp4_*` on prefill-shaped batches and `flydsl_mxmoe_g1_a4w4_*` on decode. Only the first honours `out_dtype`; the mxmoe port always emits fp4. Rewriting stage 2 to `afp8` for a mxmoe stage 1 therefore makes gemm2 read a 4-bit intermediate as 8-bit, walk past the end of the buffer, and fault. This reproduced as a memory access fault on all eight ranks during a serving run, so the rewrite now requires `kernelName1` to start with `flydsl_moe1_`. Decode keeps the fp4 intermediate until the mxmoe port learns `out_dtype`, and the accuracy numbers below are measured on a prefill-shaped configuration.

## Accuracy

Measured on MI355X, gfx950, 256 CU, against the real `moonshotai/Kimi-K3` MXFP4 checkpoint, all 96 shards. Kimi-K3 routed-expert geometry at TP=8: `model_dim=3584`, `inter_dim=384`, E=896, top-16, SiTUv2 with beta 4.0 and linear_beta 25.0, real `routed_expert_down_proj` and `routed_expert_norm`, real sigmoid noaux_tc router.

### MoE layer

| statistic over 92 layers | fp4 intermediate | fp8 intermediate |
| --- | --- | --- |
| mean rel L2 | 2.0760e-1 | 1.5746e-1 |
| median rel L2 | 2.0928e-1 | 1.5927e-1 |
| min rel L2 | 1.6664e-1 | 1.2639e-1 |
| max rel L2 | 2.2161e-1 | 1.7066e-1 |
| mean multiple of a16w4 | 37.7x | 28.6x |
| mean per-token max error | 2.2420e-1 | 1.7310e-1 |

| layer | fp4 intermediate | fp8 intermediate | gain | fp4 vs a16w4 (baseline) | fp8 vs a16w4 (baseline) |
| --- | --- | --- | --- | --- | --- |
| 1 | 1.8720e-1 | 1.3292e-1 | 1.408x | 34.4x | 24.4x |
| 10 | 2.0641e-1 | 1.5459e-1 | 1.335x | 37.5x | 28.1x |
| 20 | 2.0814e-1 | 1.5950e-1 | 1.305x | 37.8x | 28.9x |
| 30 | 2.0264e-1 | 1.4994e-1 | 1.351x | 36.9x | 27.3x |
| 40 | 1.9935e-1 | 1.4485e-1 | 1.376x | 36.2x | 26.3x |
| 50 | 1.9379e-1 | 1.4265e-1 | 1.359x | 35.7x | 26.3x |
| 60 | 2.1588e-1 | 1.6875e-1 | 1.279x | 38.9x | 30.4x |
| 70 | 2.1431e-1 | 1.6550e-1 | 1.295x | 38.7x | 29.8x |
| 80 | 2.1415e-1 | 1.6519e-1 | 1.296x | 38.6x | 29.8x |
| 92 | 2.1652e-1 | 1.6714e-1 | 1.295x | 39.1x | 30.2x |

## Performance on ATOM

Measured on MI355X, gfx950, 256 CU. Kimi-K3 routed-expert geometry at TP=8: `model_dim=3584`, `inter_dim=384`, E=896, top-16, SiTUv2, MXFP4 weights. `kv_cache_dtype=fp8`, `block-size=128`, prefix caching off. 8k/1k, 128 prompts, concurrency 128. Absolute microseconds run high because this host is capped at 1000 W against a 1400 W TBP.

### MoE layer

| token | a16w4 | a4w4 fp4 inter | a4w4 fp8 inter | fp8 cost | a4w4 speedup | fp8 speedup |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | 119.5 | 140.3 | 139.4 | -0.6% | 0.85x | 0.86x |
| 8 | 120.4 | 143.0 | 143.8 | +0.6% | 0.84x | 0.84x |
| 32 | 181.1 | 150.3 | 150.6 | +0.2% | 1.20x | 1.20x |
| 64 | 286.4 | 245.8 | 244.1 | -0.7% | 1.17x | 1.17x |
| 256 | 427.1 | 360.3 | 371.3 | +3.1% | 1.19x | 1.15x |
| 512 | 453.6 | 376.6 | 382.5 | +1.6% | 1.20x | 1.19x |
| 1024 | 574.6 | 406.7 | 406.4 | -0.1% | 1.41x | 1.41x |
| 2048 | 785.3 | 557.7 | 577.7 | +3.6% | 1.41x | 1.36x |
| 4096 | 1167.3 | 816.7 | 854.5 | +4.6% | 1.43x | 1.37x |
| 8192 | 1973.8 | 1226.1 | 1276.9 | +4.1% | 1.61x | 1.55x |
| 16384 | 3631.9 | 2021.0 | 2122.6 | +5.0% | 1.80x | 1.71x |

| Metric | fp4 intermediate | fp8 intermediate | Delta | Change |
| Mean TTFT (ms) | 31,133.82 | 31,589.15 | +455.33 | +1.46% |
| Median TTFT (ms) | 31,334.69 | 31,666.11 | +331.42 | +1.06% |
| P99 TTFT (ms) | 60,545.82 | 61,436.31 | +890.49 | +1.47% |
| Mean TPOT (ms) | 73.82 | 74.30 | +0.48 | +0.65% |
| Median TPOT (ms) | 73.62 | 74.22 | +0.60 | +0.82% |
| P99 TPOT (ms) | 102.78 | 103.59 | +0.82 | +0.80% |
| Mean ITL (ms) | 73.75 | 74.23 | +0.48 | +0.66% |
| Output throughput (tok/s) | 1,228.76 | 1,217.93 | -10.83 | -0.88% |
| Total throughput (tok/s) | 11,058.88 | 10,961.33 | -97.55 | -0.88% |
| Duration (s) | 106.67 | 107.62 | +0.95 | +0.89% |

## Submission Checklist

* [x] Look over the contributing guidelines at <https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests>.


aiter/fused_moe.py
```diff
@@ -2866,6 +2866,16 @@ def _lookup_cfg(c2s):
         _opus_a8w4.parse_stage2_config(cfg, block_m) if is_opus_cfg else None
     )
 
+    if (
```
valarLip, line 2869: let's add something like a top API level arg like quant_type_a and quant_dtype_a, default =None which means auto, otherwise use user specific quant type

valarLip (dismissed):

aiter/fused_moe.py
amd-wsung102: I have added API arg for `quant_dtype_a`, `quant_dtype_a2`, and `quant_type_a` in this commit https://github.com/ROCm/aiter/pull/5439/changes/2b85f8b2d825328da7ec2ff5a98c6e89e4ff0755. By default, they are set to None/auto.

As an example, these API can be used like this:
```
from aiter import ActivationType, QuantType, dtypes
from aiter.fused_moe import fused_moe

common = dict(
    w1_scale=w1_scale, w2_scale=w2_scale,
    quant_type=QuantType.per_1x32,
    activation=ActivationType.Situv2,
    gate_mode=GateMode.SEPARATED.value,
)

# 1. Auto, unchanged. a16w4 unless an AITER_SITUV2_* env var says otherwise.
out = fused_moe(hidden_states, w1, w2, topk_weight, topk_ids, **common)

# 2. This PR's dataflow: fp4 activations into gemm1, fp8 intermediate into gemm2.
#    No env var needed.
out = fused_moe(hidden_states, w1, w2, topk_weight, topk_ids, **common,
                quant_dtype_a=dtypes.fp4x2,
                quant_dtype_a2=dtypes.fp8)

# 3. Plain a4w4, fp4 all the way through. Replaces AITER_SITUV2_A4W4=1.
out = fused_moe(hidden_states, w1, w2, topk_weight, topk_ids, **common,
                quant_dtype_a=dtypes.fp4x2)

# 4. a8w4. Replaces AITER_SITUV2_A8W4=1.
out = fused_moe(hidden_states, w1, w2, topk_weight, topk_ids, **common,
                quant_dtype_a=dtypes.fp8)

# 5. Pin to fp4 intermediate even if AITER_SITUV2_A4W4_FP8_INTER=1 is in the
#    environment, e.g. to A/B against it from inside one process.
out = fused_moe(hidden_states, w1, w2, topk_weight, topk_ids, **common,
                quant_dtype_a=dtypes.fp4x2,
                quant_dtype_a2=dtypes.fp4x2)
```

yifehuan (approved):
LGTM

PR #3493, opened 2026-06-02
Add MiniMax-M2.7 MI325X (gfx942) tuned configs + fix fmoe tuner crash

jiejingzhangamd:
## Summary
Adds AMD **MI325X (gfx942)** tuned kernel configs for **MiniMax-M2.7** served via sglang (tp4/ep1, fp8 weights + fp8 KV), plus a fix for a crash in the fmoe tuner.

## Motivation
On MI325X (gfx942, `cu_num=304`, fp8 `torch.float8_e4m3fnuz`) the MiniMax-M2.7 decode step is GEMM-bound and was largely untuned:

- The dense **qkv_proj** GEMM (N=2048, K=3072) had **no gfx942 Triton config** → it fell back to the generic `BLOCK_SIZE_M=128`, which is **~2.3–3.3× too slow** at decode batch sizes.
- The shipped `a8w8_blockscale_tuned_fmoe_minimax-m2_5.csv` rows are keyed on `cu_num=256` / `torch.float8_e4m3fn`, but MI325X reports `cu_num=304` / `e4m3fnuz`, so **every fmoe lookup silently missed** and used the default 1-stage heuristic.

## Changes
1. **New** `aiter/ops/triton/configs/gemm/gfx942-GEMM-A8W8_BLOCKSCALE-N=2048-K=3072.json` (qkv_proj) — per-bucket Triton configs. Small-M winners use `NUM_KSPLIT=8, num_stages=3` (split the K=3072 reduction to fill the GPU). Tuned **inside a CUDA graph** — out-of-graph benchmarking hits a ~37 µs host-launch floor that masks the real kernel time.
2. **fmoe CSV** — add `cu_num=304` / `e4m3fnuz` rows for the MiniMax-M2.7 tp4 shape (model_dim=3072, inter_dim=384, expert=256, topk=8) with full decode-M coverage (1–32, incl. odd M=3,5,6,7,12). Existing rows untouched.
3. **`csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py`** — fix `ValueError: too many values to unpack` in the failed-case report (1-stage asm task tags carry extra trailing fields). This aborted the entire tuning batch and discarded all already-completed results; it triggers for shapes like `inter_dim=384` that reject all tile_n=256/512 asm kernels.

## Performance
[...]

PR #5443, opened 2026-09-11
[CK] [Bugfix][Perf] fused_moe: tune the non-temporal load hint per tuned row

ZhengGong-amd:
## What's wrong

`get_2stage_cfgs()` initialises the non-temporal load hint to `False` and assigns it only
inside the heuristic branch:

```python
use_non_temporal_load = False
if cfg is None or bypass_tuned_config:
    use_non_temporal_load = use_nt(token, topk, expert)   # heuristic branch only
else:
    ...                                                   # tuned branch: never assigned
```

A shape loses non-temporal weight loads the moment it gains a tuned CSV row, and
`AITER_USE_NT` is dead there as well, because the override lives inside `use_nt()`.

An earlier revision of this PR simply derived the flag on the tuned path too. Review pointed
out why that is not sound, and measurements agree: the tuner never passes `non_temporal_load`
(`gemm_moe_tune.py` -> `ck_moe_stage{1,2}_fwd_out`), so **every** candidate in every shipped
row was ranked with the hint off. Turning it on at runtime runs the selected kernel outside
the conditions it was selected in, and the recorded `us1`/`us2` stop describing runtime. On
gfx942 that shows up directly: the shipped `ds_v3` row for `7168x2048 e33 topk9` at token 128
(`block_m=32`) is **6.6 % slower** with the hint on, while the same shape at tokens 16 and 64
is 5.0 % / 3.5 % faster. One shape family, no constant right answer.

## Fix

Make the hint part of what gets tuned instead of something derived next to the tuned row. The
tuner sweeps both values for the quant types whose kernels read it, ranks
`(kernel, block_m, nt)` jointly by `us1 + us2`, and records the winner in a new `nt` column;
the runtime reads that column.

```diff
     else:
         block_m = cfg["block_m"]
+        # rows tuned before the nt column keep the hint off
+        nt_override = int(os.environ.get("AITER_USE_NT", "-1"))
+        if nt_override != -1:
+            use_non_temporal_load = bool(nt_override)
+        elif "nt" in cfg:
+            use_non_temporal_load = bool(int(cfg["nt"]))
```

`use_nt()` and the heuristic branch are untouched, so nothing changes for shapes without a
tuned row. Rows without the column keep `nt=False`, which is today's behaviour, so no shipped
config changes until it is retuned. `AITER_USE_NT` now overrides both paths, as it was
documented to. On the tuner side `nt` is threaded exactly like the existing `flat` / `xbf16`
columns: it rides in the candidate tag, becomes a column in the candidate table, joins the
dedup key and the stage1 x stage2 merge key (so both stages of a row agree on it), and is
written out. The sweep is gated to `per_1x128`, so every other quant type keeps its candidate
count.

## Results, MI325X (gfx942, 304 CU)

DSv3 EP shape `7168x2048 e33 topk9`, per_1x128 fp8 blockscale, preshuffled weights.

**Tuner level** — same box, same harness, the extra dimension is the only difference:

| token | without the nt sweep    | with the nt sweep               | gain   |
| ----- | ----------------------- | ------------------------------- | ------ |
| 16    | `block_m=16`, 360.34 us | `block_m=16`, `nt=1`, 329.59 us | +8.5 % |
| 64    | `block_m=32`, 375.48 us | `block_m=32`, `nt=1`, 358.88 us | +4.4 % |
| 128   | `block_m=64`, 404.58 us | `block_m=64`, `nt=1`, 395.84 us | +2.2 % |

**Runtime level** — whole `fused_moe` op, CUDA-graph replay, 200 iters after 40 warmup, 8
rotated weight sets, 3 interleaved rounds per config, median:

| token | shipped rows | retuned, no nt sweep | retuned, nt swept | vs shipped |
| ----- | ------------ | -------------------- | ----------------- | ---------- |
| 16    | 383.2 us     | 374.5 us             | 329.6 us          | +14.0 %    |
| 64    | 398.0 us     | 411.5 us             | 365.0 us          | +8.3 %     |
| 128   | 455.4 us     | 431.2 us             | 417.2 us          | +8.4 %     |

Retuning with the hint in the search is worth +3.3 % to +12.0 % over retuning without it on
the same box.

**Why this cannot be a heuristic.** The `block_m` the tuner picks and the best value of the
hint are coupled: what the hint discards is weight reuse between M-tiles of the same expert,
so it wins while an expert stays inside one tile and loses once it spans two. At token 128 on
this shape, stage2 measured alone with `rocprofv3` is 161.0 -> 191.8 us with the hint at
`block_m=32`, but 153.1 -> 135.7 us at `block_m=64`; `rocprofv3 --pmc TCC_HIT_sum TCC_MISS_sum`
attributes the whole difference to cache hit rate (42.8 % -> 28.5 %, 722 MB -> 920 MB below L2
against 2 x 484 MB of logical B reads, both arms sustaining ~4 TB/s). A predicate over token
count cannot see `block_m`, and under imbalanced routing it cannot see the tail experts that
cross into a second tile before the mean does either. Tuning it removes the guesswork.

## Results, MI355X (gfx950, 256 CU)

Second architecture, independent box. 26 shapes from three families (`4096x384 e128 topk8`,
`7168x512 e257 topk9`, `3072x1536 e256 topk8`), same machine, same harness, same
single-process settings, one card, tuned **three times per arm** at the PR's parent `c3caa55`
and at the head `c26f391`. 23 of the 26 resolve a winner on every run of both arms (the other
three are under *Caveats*). Times are the tuner's own per-candidate measurement for the
winning row; gain is median-of-three, spread is max-min within an arm.

The result separates by winning path. The 16 asm 1-stage shapes do not read the flag and do
not move: all keep `nt=0`, median gain +0.07 % over a range of -0.7 % .. +0.8 % against
per-arm spreads up to 1.4 %. The 7 CK 2-stage shapes are one family, `4096x384 e128 topk8`,
`doweight_stage1=1`, all at `block_m=16`:

| token | parent    | PR        | gain   | per-arm spread | `nt`                |
| ----- | --------- | --------- | ------ | -------------- | ------------------- |
| 1     | 38.62 us  | 38.19 us  | +1.1 % | 7.1 % / 4.6 %  | `0`/`1`, unstable   |
| 2     | 42.65 us  | 41.38 us  | +3.0 % | 6.2 % / 1.5 %  | `1`                 |
| 4     | 50.69 us  | 47.30 us  | +6.7 % | 6.2 % / 3.6 %  | `1`                 |
| 8     | 61.86 us  | 57.14 us  | +7.6 % | 1.3 % / 2.7 %  | `1`                 |
| 16    | 83.12 us  | 76.54 us  | +7.9 % | 1.4 % / 1.7 %  | `1`                 |
| 32    | 113.93 us | 105.10 us | +7.8 % | 0.3 % / 0.4 %  | `1`                 |
| 64    | 122.32 us | 113.05 us | +7.6 % | 0.2 % / 0.4 %  | `1`                 |

From token 8 up, the gain is several times either arm's spread and holds within half a point
across four shapes. At tokens 1-4 the shape runs in tens of microseconds, the arms overlap,
and token 1 does not even settle on a value across repeats — the flag genuinely does not
matter there, so nothing is claimed. Both arms pick the same kernel pair at tokens 1-16; at
tokens 32 and 64 the PR arm's winner is the `128x16x128x128` variant rather than
`256x16x128x256`, which is the ranking doing its job once `nt` is in the search space. An
isolated single-shape rerun where both arms converged on the identical kernel and `block_m`
isolates the flag on its own: 121.76 us -> 114.31 us, +6.1 % (stage1 +7.0 %, stage2 +4.5 %).

## Cost and compatibility

* CK candidates double for `per_1x128` only: 10 -> 20 per shape here, one shape 19 s -> 38 s,
  roughly 10 min for a full model file. Verified that a `per_Token` tune is unchanged.
* Rows without the column behave exactly as today. The 46 shipped blockscale 2-stage rows with
  `use_nt() == True` (38 at `cu_num` 256, 8 at `cu_num` 304) keep the hint off until someone
  retunes them; retuning shipped configs is out of scope here.
* No C++ change: `ck_moe_stage{1,2}_fwd` already take the flag and `gen_instances.py` already
  compiles both NT instantiations.

## Verification

* Runtime decision table, all five cases checked against `get_2stage_cfgs` on **both** gfx942
  and gfx950: legacy CSV (no column) -> off; `nt=0` -> off; `nt=1` -> on; legacy +
  `AITER_USE_NT=1` -> on; `nt=1` + `AITER_USE_NT=0` -> off. At the parent commit all five
  return off, which is the bug.
* Shipped-config regression, exhaustive: all **179** `cu_num=256` blockscale 2-stage rows across
  the 7 shipped CSVs (`ds_v3`, `glm5`, `minimax-m2_5`, `qwen3_235b`, `qwen3_5_397b`, and the
  untuned files) resolved through `get_2stage_cfgs` on both arms and compared field by field —
  `block_m`, `ksplit`, both kernel names, `nt_stage1`, `nt_stage2`. **Zero differences.** 119 of
  those rows are ones where `use_nt()` would return `True`; they stay off on both arms,
  confirming the heuristic was already inert on the tuned path and that this change does not
  switch it on.
* Tuner smoke test: the dispatch list shows each CK kernel twice, once per value; the candidate
  table and the tuned CSV carry `nt`; `per_Token` keeps its candidate count.
* Hand-written rows confirm the column drives the kernel: the same shipped config with `nt=1`
  reproduces +5.0 % / +3.5 % / -6.6 % at tokens 16 / 64 / 128.
* Numerics: the flag only selects `AmdBufferCoherenceEnum`, no arithmetic changes. Measured on
  gfx950, per-candidate reference error is identical between the two arms and between `nt=0` and
  `nt=1` (stage1 0.1 % in every group; stage2 within 0.1 pp).
* `op_tests/test_moe_blockscale.py` exits 0 with the same six checks; instrumenting
  `get_2stage_cfgs` shows it never enters the tuned branch (asm path only), so it does not cover
  this change.
* `ruff` clean on both files; no new findings (3 pre-existing in the tuner, 34 pre-existing E501
  in `fused_moe.py`, unchanged).

## Caveats

* The rankings above were produced with the catastrophic-mismatch check relaxed locally, because
  the tuner's reference comparison rejects blockscale stage1 candidates — asm and CK alike — on
  both of my boxes, while `per_Token` is unaffected. That blocks blockscale tuning independently
  of this change: the two files involved are untouched by this PR, and the parent commit fails
  identically on the same shapes. On gfx950 it is what leaves the three shapes above unresolved.
  The measurements are unaffected, since both arms are gated identically and the per-candidate
  errors match between them.
* Per-stage tuning of the hint is deliberately not attempted: the runtime passes one flag to both
  stages, so the merge forces stage1 and stage2 of a row to agree.

## Duplicate check

This updates my own #5443. Of the 640 open PRs, none touch `use_nt` / `non_temporal_load` in
`aiter/fused_moe.py`: #5117 is MXFP6 non-temporal *stores* in GEMM, #2947 is a gfx942 CK stage2
OOB fix, #3493 adds gfx942 tuned rows (12 more `cu_num=304` blockscale 2-stage rows, all of which
this PR leaves at `nt=0` until retuned).

---


ZhengGong-amd:
@yzhou103 Could you please run the CI again, the pr has been updated.

ZhengGong-amd:
**MI325X (gfx942, 304 CU)**

Coverage first, since the shipped configs barely reach this arch: 8 blockscale 2-stage
rows at `cu_num=304` have `use_nt()` true, all DSv3 EP `7168x2048 e33 topk9`, tokens
1-128. #3493 adds 12 more (MiniMax-M2.5/M2.7 `3072x384 e256 topk8`, tokens 1-32).
Everything else on gfx942 is either `run_1stage=1` asm, which takes no `nt`, or not
blockscale.

Whole `fused_moe` op on those rows, CUDA-graph replay, 200 iters after 40 warmup, 8
rotated weight sets, 3 rounds per arm interleaved round by round, median:

| token | block_m | rows/expert | before    | after     |
| ----- | ------- | ----------- | --------- | --------- |
| 1     | 16      | 0           | 116.5 us  | 113.8 us  |
| 2     | 16      | 0           | 201.5 us  | 192.7 us  |
| 4     | 16      | 1           | 279.6 us  | 270.3 us  |
| 8     | 16      | 2           | 335.4 us  | 323.3 us  |
| 16    | 16      | 4           | 360.5 us  | 340.9 us  |
| 32    | 16      | 8           | 369.4 us  | 358.1 us  |
| 64    | 32      | 17          | 382.0 us  | 369.3 us  |
| 128   | 32      | 34          | 455.6 us  | 456.1 us  |
| sum   |         |             | 2500.5 us | 2424.5 us |

+3.0 % aggregate, +3.0..5.4 % per tier up to token 64; the token-128 tier is outside the
bound and unchanged. stage1 is the asm kernel on these rows and does not move (256.3 vs
255.8 us), so the whole delta is in the CK stage2.

Controls at token 128, three timing windows each: `after + AITER_USE_NT=0` reproduces
`before` (455/453/449 us vs 454/452/449 us), and `before + AITER_USE_NT=1` does not change
`before` (455/452/449 us) -- the override really is dead on the tuned path today, and
nothing but the `nt` flag differs between the arms.

No MI325X end-to-end number, because none is measurable with shipped configs:
Qwen3.5-122B-A10B-FP8, TP=2, `VLLM_ROCM_USE_AITER=1` reports `tuned=False` on all 25
unique MoE dispatches (`3072x512 e256 topk8` has no `cu_num=304` row), so both arms run
identical kernels and the heuristic branch already supplies `nt=True`.

ZhengGong-amd:
**Why the tuned-path flag is bounded by `block_m`**

What the hint discards is weight reuse between M-tiles of the same expert: with more than
one tile per expert, every tile re-reads that expert's whole weight slab.
`rows_per_expert < 64` stands in for "read once", but it is only equivalent to
`ceil(rows_per_expert / block_m) == 1` when `block_m == 64`.

gfx942, DSv3 EP `7168x2048 e33 topk9`, token 128 (34 rows per expert), stage2 timed alone
with `rocprofv3 --kernel-trace`, only `nt` and `block_m` varied:

| block_m | M-tiles per expert | nt=False | nt=True  |
| ------- | ------------------ | -------- | -------- |
| 32      | 2                  | 161.0 us | 191.8 us |
| 64      | 1                  | 153.1 us | 135.7 us |

Counters account for the whole delta at `block_m` 32: stage2 TCC hit rate 42.8 % ->
28.5 %, misses 5.64 M -> 7.19 M per dispatch, i.e. 722 MB -> 920 MB below L2 at 128 B/line
against 2 x 484 MB of logical B reads. Both arms sustain ~4 TB/s, so the extra 198 MB is
the extra 55.6 us.

The tuner is the second reason: `gemm_moe_tune.py` never passes `non_temporal_load` to
`ck_moe_stage{1,2}_fwd`, so every tuned row's `block_m`, kernel names and recorded
`us1`/`us2` were selected under `nt=False`. Bounding the runtime decision by `block_m`
keeps it inside the conditions the config was picked in.

Of the 46 shipped blockscale 2-stage rows with `use_nt()` true, the bound leaves 43
enabled and keeps 3 at `nt=False`, all at `rows_per_expert >= block_m`: `ds_v3`
`7168x2048 e33 topk9` token 128, and `qwen3_235b` `4096x1536 e16 topk8` token 64 and
`4096x384 e128 topk8` token 512, all `block_m=32`.

Cross-checks on synthesised tuned rows for a `3072x1024 e256 topk8` MoE, deltas vs
`nt=False`, 3 interleaved rounds: with `block_m` forced to 16, token 256 (8 rows/expert)
keeps +3.7 % and tokens 512 and 1024 fall outside the bound; with the tuner-shaped 32/64
rows, tokens 256/512/1024 stay enabled at +3.0/+1.5/+2.0 %. Same shape and token as the
table above but `block_m=64` also stays enabled, at +1.8 %.

Making `nt` a tuned column would drop the heuristic for tuned shapes entirely -- on the
DSv3 shape `block_m=64 + nt=True` (135.7 us) beats the shipped `block_m=32 + nt=False`
(161.0 us) by 16 %, so the two are coupled. Happy to follow up with that.

yzhou103:
The block_m guard avoids the known multi-M-tile regression, but it does not make the tuned configuration consistent with runtime. Every CK candidate in the tuning CSV was ranked with use_non_temporal_load=False, while this line runs the selected candidate with nt=True. Since NT changes the cache behavior, it may also change the relative ranking of kernels with different tile/wave/pipeline parameters. The benchmarks show that the already-selected kernels improve for the sampled shapes, but they do not show that those kernels remain optimal under NT, and the recorded us1/us2 values no longer describe runtime behavior. Could we either apply the same NT predicate during tuning or record nt as part of each tuned row? If that is outside this PR’s scope, I would keep automatic NT disabled for tuned rows and only restore the explicit AITER_USE_NT override.

ZhengGong-amd:
Reworked along your first option: `nt` is tuned rather than derived. The tuner sweeps both
values for `per_1x128` and ranks `(kernel, block_m, nt)` jointly -- the winner does change
once the hint is in the search, so ranking every candidate with it off and running the
winner with it on was indeed not sound -- and records the result in the row. The runtime
reads that column; rows without it keep `nt=False`, so existing tuned configs are unchanged
until they are retuned. That covers your fallback as well, and `AITER_USE_NT` works on the
tuned path again. `use_nt()` and the heuristic branch are untouched.

yzhou103:
Since new models are moving away from CK, could we first retune the affected existing shapes with FlyDSL included in the candidate set and compare it against both shipped CK and CK with the NT dimension? If FlyDSL covers these configurations and is competitive, migrating the existing rows would avoid adding a new CK-specific field and pairing logic to the shared CSV schema. We should keep CK only for shapes where FlyDSL is unsupported or materially slower

ZhengGong-amd:
> Since new models are moving away from CK, could we first retune the affected existing shapes with FlyDSL included in the candidate set and compare it against both shipped CK and CK with the NT dimension? If FlyDSL covers these configurations and is competitive, migrating the existing rows would avoid adding a new CK-specific field and pairing logic to the shared CSV schema. We should keep CK only for shapes where FlyDSL is unsupported or materially slower

FlyDSL candidates are gated to per_1x32 in gen_flydsl_2stages_task (q_type != QuantType.per_1x32 -> return), on this branch and on main, and there is no FlyDSL path for per_1x128 in the runtime either. All the rows this PR touches are per_1x128 fp8 blockscale (DSv3/R1, Qwen3-235B, GLM, MiniMax), so including FlyDSL in the candidate set for them is not something I can measure today — it would need FlyDSL blockscale support first. Is that on the roadmap? If it is, parking this until those rows can migrate is fine by me.

On the schema point: the column is no longer in the shared resultList — it is passed only to the fmoe tuner, so the mxfp4-flydsl and grouped tuners are untouched and never emit it. The stage1/stage2 pairing is one entry in the tuner's internal merge key, not part of the CSV contract. If you would still rather not have the column at all, the fallback you offered earlier stands: drop the automatic behaviour and keep only the AITER_USE_NT override, which needs no schema change.

ZhengGong-amd:
Ran the flow from `csrc/ck_gemm_moe_2stages_codegen/README.md` on gfx942 (MI325X, 304 CU), same
command and same four shapes at the PR's parent and at this branch, to check the tuning path
itself:

```bash
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py \
  -i aiter/configs/untuned_fmoe.csv -o aiter/configs/tuned_fmoe.csv -o2 profile_fmoe_all.csv
```

Shapes: `7168x2048 e33 topk9` 64 and `3072x1024 e256 topk8` 256 (per_1x128, the only quant
type this PR sweeps), `4096x14336 e8 topk2` 1024 (per_Token), `4096x1408 e128 topk6` 128
(bf16, `QuantType.No`).

| | parent | this branch |
| --- | --- | --- |
| exit / wall time | 0 / 26.0 s | 0 / 25.4 s |
| tuned / failed shapes | 4 / 0 | 4 / 0 |
| `tuned_fmoe.csv` | 621 rows, 28 cols | 621 rows, 29 cols (`nt`) |
| winning row per shape | — | identical to parent on all four |

The incremental rewrite of the existing file is the part that matters for "don't break the tune
path": all 617 pre-existing rows survive with **zero** differences in `block_m`, `ksplit`,
`kernelName1`, `kernelName2` and `run_1stage`, and pick up `nt=0`. Candidate counts are
unchanged for per_Token and `QuantType.No` (2 -> 4 only for per_1x128), and
`GroupedFmoeTuner` / `Mxfp4FlydslTuner` override both `post_process` and `result_to_csv`
without `super()`, so neither sees the column.

The sweep itself, on `7168x2048 e33 topk9` 64: `nt=1` wins at every `block_m` (`32`: 358.72 vs
375.59 us, `64`: 386.97 vs 400.47 us) and the selected row moves from the asm 1-stage kernel
(376.18 us) to CK 2-stage `block_m=32`, `nt=1` (358.72 us).

One thing worth flagging for anyone reproducing this on gfx942: at the default error tolerance
the CK 2-stage candidates never enter the ranking here at all. The gate is
`profileDF["err"] <= args.errRatio` (`gemm_moe_tune.py:4219`) with the default `errRatio` of
`0.1` in the code, and every blockscale 2-stage candidate on this box reports a stage2 cosine
err of `0.159`, so the whole group is dropped and only asm 1-stage rows remain — winner
`block_m=32`, 1-stage, 364.05 us. The numbers above therefore come from a run with `--errRatio`
raised. This reproduces identically at the parent commit, so it is not from this PR, but it
does mean a default `per_1x128` tune on MI325X can currently only ever emit 1-stage rows. (The
README documents the default as `0.5`, which does not match the code either.) I can open that
separately.

yzhou103:
it  seems --e2e_tune raises ValueError after this change

aiter/fused_moe.py
```diff
@@ -3245,6 +3245,11 @@ def _lookup_cfg(c2s):
         )
     else:
         block_m = cfg["block_m"]
+        nt_override = int(os.environ.get("AITER_USE_NT", "-1"))
+        if nt_override != -1:
+            use_non_temporal_load = bool(nt_override)
+        elif "nt" in cfg:
+            use_non_temporal_load = bool(int(cfg["nt"]))
```
yzhou103: The column may be absent (old files), empty (hand-merged / partially
      written files -> NaN), or a float ("1.0"). None of those should raise
      on the fused_moe hot path

ZhengGong-amd:
> it seems --e2e_tune raises ValueError after this change

fixed in b94475c.

`e2e_tune` builds its rows from a hard-coded results tuple instead of going through
`post_process`, so adding `nt` to `resultList` left the row one value short of
`self.columns`: `pd.DataFrame` raised `29 columns passed, passed data had 28 columns`.
It hits every fmoe quant type, not just `per_1x128`, and only after the tuning work is
done. Those rows are `flydsl_gfx942` with an empty `kernelName2` and never read the hint,
so the tuple now carries `nt=0`.

I went through the other row builders in `FmoeTuner`: the failed-shape path in
`post_process` already carries the extra `-1`, and `_write_candidate_config` passes a
dict, so both are fine; `GroupedFmoeTuner` and `Mxfp4FlydslTuner` construct their own
result lists. My wheel has no `aiter.ops.flydsl.fused_moe_gfx942`, so I could not run
`--e2e_tune` itself — this is verified on the column count (14 keys + 14 results + `nt`
= 29, against a 29-value row).

aiter/fused_moe.py
ZhengGong-amd: fixed in b94475c:

```python
try:
    use_non_temporal_load = bool(int(float(cfg["nt"])))
except (TypeError, ValueError):
    use_non_temporal_load = False
```

Checked each case on gfx942 by resolving a real tuned row through `get_2stage_cfgs`:
column absent -> `False` (the `elif` guard, and `update_config_files` fills a missing
column with 0 anyway); blank cells -> NaN, which raised `cannot convert float NaN to
integer` and is now `False`; a column pandas keeps as object -> `"1.0"` strings, which
raised `invalid literal for int()` and is now `True`. A clean float column parses as
float64 and was already fine.

`xbf16` and `flat` have the same shape but are only evaluated for `run_1stage` rows, so
I left them as they are.

yzhou103:
The clones survive the merge now that both stages are expanded. On per_1x32 every
  candidate pair is duplicated (nt=0 original + nt=1 clone with bit-identical us,
  since both copies get the same fairness adjustments), and they reach
  `best_one = profileDF.loc[profileDF["us"].idxmin()]` as an exact tie. It resolves
  correctly today only because `concat([profileDF, agnostic])` puts originals first
  and idxmin returns the first minimum -- an invariant that isn't written down
  anywhere and that any upstream sort_values would flip.

  It's also visible right now: --profile_file dumps the post-merge frame, so every
  mxfp4 candidate is listed twice with identical timings, one tagged nt=1 even
  though nt was never measured for that quant type.

  An early return keeps per_1x128 untouched and makes the per_1x32 case
  structurally impossible rather than incidentally correct:

      if not (profileDF["nt"] == 1).any():
          return profileDF

ZhengGong-amd:
Done, added the early return in 8ee715a.

yzhou103 (approved):

PR #5455, opened 2026-09-11
.agent_loop: orchestration layer driving review-pr and validate-kerne…

demonsan:
…l-pr as one gated loop

Adds .agent_loop/, a workflow-level orchestration package that runs the two review skills as a single loop: prefetch, one human break, then analysis fan-out, independent cross-model refutation, the seven review-pr gates, on-GPU validation, refutation of the executor's own findings, refold and cleanup.

It is not a third skill, which is why it sits beside .claude rather than inside it: it calls the skills' CLI surface and adds the guarantees a single pass cannot give. Every card finding is attacked by an agent that never saw the analysis, that agent writes the ledger line itself, and the ledger's provenance is checked before a card can be certified.

Six fail-closed guards, all covered by a zero-token dry-run harness (tools/dryrun.mjs, 10/10): dead analysis agent, collection short of the FIRE count, ledger provenance mismatch, gates never green, ledger rewritten behind the gate, and gates failing after the validation refold.

docs/upstream-findings.md records four reproducible defects this loop found in validate-kernel-pr itself, each of which reads as a defect in the PR under review if the tool is trusted.

Validated on ROCm/aiter#4961 across three heads on gfx950 and gfx1250; see VALIDATED-AGAINST.md for the pinned CLI surface and the measurements.

### DIFF
--- a/.agent_loop/.gitignore
+++ b/.agent_loop/.gitignore
@@ -1,0 +1,7 @@
+﻿# installation-specific and never committed
+config/loop.json
+*.key
+id_*
+*.pem
+__pycache__/
+*.pyc
--- a/.agent_loop/README.md
+++ b/.agent_loop/README.md
@@ -1,0 +1,83 @@
+# .agent_loop
+
+An orchestration layer that drives the two review skills in this repo — `.claude/skills/review-pr`
+and `.claude/skills/validate-kernel-pr` — as one repeatable loop, so that a PR goes from a number
+to a gated verdict card plus deterministic on-GPU evidence without a human holding it by the hand.
+
+It is **not** a third skill. The skills decide *what* a review means; this decides *who runs what,
+in which order, and what may not be believed*. It calls their CLI surface and nothing else.
+
+```
+[once per host] environment probe → human confirms → env record
+[per PR]  phase P (prefetch): verify env + skill Step 1 fetch        ← cheap, 2 agents
+             ↓  ★ the single human break: targets, models, budget — asked once
+          phase R: analysis fan-out → independent refutation → card → seven gates
+                   → on-GPU validation → refute the executor → refold → cleanup   ← no stops
+```
+
+## Why a loop rather than one review pass
+
+Three things only a loop can do, each of which happened during the runs this was built from:
+
+- **It re-reviews the same PR as it moves.** When the base advances, a file the previous head added
+  and this head removed is invisible in the diff — it reads as if it never existed. Phase P reports
+  `git log <prev-head>..<head>` for exactly this reason.
+- **It puts the tools on trial too.** A validator finding is a claim, not a fact. Stage S6b refutes
+  the executor's own findings; on the runs behind this package that withdrew four false ones,
+  including a "comparison tolerance widened" verdict that was an artifact of positional matching.
+- **It makes the review record falsifiable.** Every finding on the card was attacked by a fresh
+  agent that never saw the reasoning behind it, and that agent — not the card's author — wrote the
+  ledger line recording the outcome.
+
+## Quick start
+
+```bash
+cp config/loop.example.json config/loop.json && $EDITOR config/loop.json
+
+node tools/dryrun.mjs            # zero-token: proves the guards fire. 10/10 or do not proceed.
+```
+
+Then, from an agent harness that provides a `workflow` tool (the scripts are workflow script
+bodies, not standalone programs):
+
+1. run `workflows/prefetch.workflow.js` with `args` built from your config plus `{ pr, sinceHead }`;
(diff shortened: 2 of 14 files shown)


zufayu:

**Adds a self-contained .agent_loop/ package (no aiter/ file touched) that runs the two review skills as one gated loop — prefetch, a single human break, analysis fan-out, per-finding independent refutation, the seven triage.py gates, on-GPU validation, refutation of the validator's own findings, refold and cleanup — plus four inspection tools, a 10-scenario zero-token dry-run harness, config and docs.**

Validation (deterministic): NOT RUN — the PR changes runtime code but ships no test target (triage); the changed paths are .agent_loop/ orchestration tooling that nothing in the serving path loads, so the absent target is not itself a defect — the package's own zero-token harness (tools/dryrun.mjs) was run for this review and passes 10/10
Perf (advisory): NOT RUN — no benchmark entry point: the changed paths are orchestration tooling with no runtime perf surface (triage basis: no perf target for the same reason there is no test target)

⚠️ The S4/S5 gate-green decision greps a formatter agent's prose for the substring `ALLGREEN`, so a failing gate report that contains the word negated reads as green: with 3 findings, a gate-check response "gate 3 exited 1 … I cannot report ALLGREEN" satisfies both checks (substring match true, `\b3\b` matches "gate 3"), reproduced by running the two regexes from run.workflow.js on that string — and a hallucinated or lazy ALLGREEN ships an ungated card through validation and the refold the same way. This contradicts README.md's guarantee that the seven gates are "judged by the skill's own exit codes, never by a model's say-so" and loop.md's termination table: the workflow body never sees an exit code, every verdict passes through a model's transcription, and the dry-run covers a missing digit but not a negated mention. **Author must** have the gate-runner write one structured line per gate (e.g. `GATE3:1`) to a file and count those seven lines in the workflow body instead of grepping prose for a word. [verified]

⚠️ The "independent" refuter can read the worker's reasoning: the refute prompt's context block points the refuter at the same scratch directory ("its artifacts are at $WORK — read them, never re-derive") where answers.txt, verdicts_*.txt and findings_raw.txt — the analysis the refuter is told it "has NOT seen" — sit with no prohibition, so the independence README.md guarantees ("an agent that did not see the analysis") is prompt-asserted, not isolated. A refuter that lists the directory anchors on the worker's own reasoning, degrading cross-model refutation (the 40% vs 18% kill-rate gap the package exists to keep) toward same-reasoning review, and known-limitations.md documents the cross-round sibling-directory leak while this same-round leak goes unrecorded. **Author must** move the S2 outputs out of the refuter's reach (a separate directory, or an explicit prohibition in the refute prompt) and record the residual exposure in known-limitations.md. [verified]

⚠️ resume reuses stale refutations for re-derived findings: refuters are skipped by positional F-number when their file already exists, but S2 always re-runs first (known-limitations.md: "analysis outputs … are always redone"), so the findings list is re-derived and can shift in content while the F-file count stays equal or grows — the previous attempt's F01.md verdict then attaches to a different finding 1, and the ledger provenance check passes because it compares file and line counts (FILES=REFUT=INDEP), never content. A resumed round can certify a card whose refutation ledger describes findings that no longer exist, with no gate any the wiser. **Author must** key the skip on content — the refuter already restates the finding in F<nn>.md, so compare that restatement (or hash the finding into the filename) instead of trusting the positional number. [verified]

⚠️ The FIRE-count floor fails open when its counter dies: minFindings is derived from a formatter agent's `FIRE=<n>` line with a silent 0 fallback and no null check — unlike the prefetch phase, which aborts when its env or fetch agent returns nothing — so a dead counter degrades the collection floor to max(1, 0) = 1 and a collector that lost findings passes the exact guard ("a collection short of the FIRE count aborts the run") that exists to catch it. The counter is a formatter agent, the class loop.md documents returning nothing 31% of the time in one round, and the dry-run covers a short collection but not a dead counter. **Author must** abort when the count agent returns nothing or no FIRE= line, the same way the env and fetch agents already do. [verified]

⚠️ A cleanup that reports failure in prose counts as success: S8 computes cleanupFailed only from agents that returned nothing at all, so an agent that reports "ssh refused, could not clean" in prose — exactly what its prompt demands ("SAY SO AND STOP") — returns non-empty text and leaves cleanupFailed at 0, an actively wrong signal in the run's return value. loop.md §8.6 states "a failed cleanup is a failed round", and even known-limitations.md's softer description ("the run logs it and still returns") overstates the code, which logs nothing for a prose-reported failure; the next round then discovers the dirty remote as a worktree-removal permission failure. **Author must** make the cleanup agent write a machine-checkable verdict line (e.g. CLEAN:OK / CLEAN:FAILED) and treat FAILED as the failed round the contract promises. [verified]

PR #5461, opened 2026-09-11
[FlyDSL] One stage and two-stage ring all-reduce kernels for PCIe interconnects

vpietila-amd:
## Motivation

PR #4970 introduced a two-stage int4 quick reduce FlyDS kernel that improved all-reduce throughput over the existing Aiter two-stage/RCCL kernels. Current Aiter all-reduce kernels are written primarily for the xGMI interconnect. This PR optimizes the FlyDSL quick reduce int4 kernel for PCIe interconnects (MI350P) and adds two new kernels: a one-stage lossless and a two-stage ring kernel with int4/int6 codecs. On MI350P, they offer better performance than the existing Aiter production kernels (RCCL/cross-device-reduce)

Category | Speedup TP=2 | Speedup TP=4 | Speedup TP=8
-- | -- | -- | --
Decode (new kernels are exact) | 1.40x-2.27x (avg: 1.62x) | 1.45x-5.60x (avg: 2.25x) | 2.36x-4.60x (avg: 3.18x)
Prefill (new kernels use quantization) | 2.42x-3.54x (avg: 2.94x) | 2.50x-3.89x (avg: 3.08x) | 2.43x-4.84x (avg: 3.57x)

## Technical Details

**Kernels** (`aiter/ops/flydsl/kernels/`)
- `one_shot_allreduce.py`: exact one-stage push all-reduce. bf16 in and out, fp32 accumulation in a fixed rank order. The IPC inbox is double-buffered by color parity.
- `quick_allreduce_int4_ring.py`: two-stage ring schedule, 2(N-1) single-destination hops, with a wire format (int4/int6) chosen separately for each lap. At TP8 the reduce-scatter lap defaults to INT6 so the ring holds the 18 dB SQNR floor.
- `quick_allreduce_codec.py` and `quick_allreduce_shared.py`: the INT4, INT6 and fp16 wire formats and the transport shared by the mesh and ring schedules. The #4970 mesh kernel is refactored onto them.

**Host wrappers** (`aiter/ops/flydsl/`)
- `QuickAllReduceInt4` gains an `algorithm="mesh"|"ring"` choice, per-lap codecs, inbox memory modes, and payload-size super-tile ladders.
- `OneShotAllReduce` is a new host wrapper for the one-stage kernel. It picks a per-fabric rung (`ONESHOT_LADDER`) by payload size.
- Fabric detection (`has_xgmi_peer_links`) reads KFD link types. PCIe hosts get a fine-grained inbox, because uncached peer writes serialize per destination on PCIe. On xGMI we use un-cached IPC memory.

**Production dispatch**, opt-in with `AITER_FLY_AR=1` - the existing behavior is unchanged when the env variable is unset/disabled.
- The custom-all-reduce slot is fronted by the exact one-shot for plain bf16 up to the policy ceiling.
- The quick-reduce slot, under `AITER_QUICK_REDUCE_QUANTIZATION=INT4`, dispatches bf16 payloads in the policy window to the FlyDSL mesh or ring instance.
- The windows come from `allreduce_policy.FAMILY_POLICY`, keyed on (fabric, world size). They can be overridden with `AITER_FLY_AR_ONESHOT_MAX_BYTES` and `AITER_FLY_AR_MESH_MAX_BYTES`.
- `CudaCommunicator.destroy()` now closes the all-reduce slots it owns. That releases the IPC handles before the process group goes away.
- Fabric detection is per host. The assumption that a node is uniformly either xGMI or PCIe.

## Test Plan

Added option to use fp16 codec over the wire. This ensures lossless transport and allows to verify the transport correctness using a bitwise equality/all close against a reference implementation. The new one-stage kernel is exact by construction. Added/modified the following test suites that run under the CI
-  `test_flydsl_quick_allreduce_int4.py`: Added tests for the ring algorithm, added transport correctness tests for both mesh and ring algorithm.
- `test_reuse_identical_rank_groups.py`: Added tests to ensure that `CudaCommunicator.destroy()` now closes the all-reduce slots it owns.
- `test_flydsl_one_shot_allreduce.py`: New test suite for the one-stage exact all-reduce kernel.

All tests should pass on CI to verify the code changes form this PR.

Created a separate benchmarking suite for communication collective under `op_tests/op_benchmarks/flydsl`
- `bench_comm_allreduce.py`: Comprehensive benchmarking suite for Aiter's all-reduce collectives. Can be extended to include the fused all-reduce variants.
- `transferbench_roofline.py`: Optional roofloine analysis using TransferBench. Used in the benchmarking script with additional `--roofline` flag to provide wire-efficiency estimate of the different all-reduce implementations.

The benchmarking suite is used to tune the instance selection as well to produce performance claims in this PR. The suite can be used to tune any all-reduce implementations, not just the FlyDSL ones.   ‎

## Submission Checklist


samremes:
# Agent review: AITER PR #5461 — FlyDSL ring all-reduce

**Reviewed:** https://github.com/ROCm/aiter/pull/5461
**Title:** FlyDSL one-stage and two-stage ring all-reduce kernels for PCIe
**Local worktree:** `/home/samremes/dev/aiter-pr-5461-flydsl-comm-kernels`
**Base:** `samremes/flydsl-qr-int4` @ `3312a2b66c5ad91429df7c14cda5abd48c6e8100`
**Head:** `c8e09bc30ab90bc4c7e0e994431d29e83afbe806`
**Dirty files in this worktree:** none
**Review date:** 2026-09-14
**Skills applied:** GPU communication kernels, FlyDSL kernel testing, AITER op test, and FlyDSL kernel review guidance

This document is a read-only review of `3312a2b66..c8e09bc30`. I ran the policy and codec tests. I ran the TP2 one-shot script on the attached gfx950 system. I compiled four ring variants for gfx950 and inspected their resource metadata. I did not run a ring collective or a performance benchmark. A later pass audited the test and benchmark surface against the AITER op-test contract and the CI runner scripts.

## Verdict

Changes are required before merge. One dispatch mismatch raises on an input that the new production selector accepts. A stray gitlink also breaks recursive submodule setup and must come out of the PR. The test surface is the third area that needs work: it is large, it diverges from the AITER op-test contract, the standard CI path executes almost none of it, and one new script under `op_tests/multigpu_tests/` fails the multi-GPU job on every runner. The payload-before-publish order, current-stream launch, ring slot mapping, and sampled persistent-grid residency are coherent.

## Review context

| Item | Value |
|---|---|
| Change class | New kernels, performance policy, public AITER integration, and benchmark support |
| Targets | gfx942 and gfx950, wave64; xGMI and PCIe policies |
| Dtypes | bf16 input/output; fp32 exact one-shot accumulation; packed fp16 quantized accumulation; INT4, INT6, or fp16 wire payload |
| Shape contract | Engine wrappers require contiguous, non-overlapping, 16-byte-aligned buffers with byte size divisible by 16 |
| Collective scope | One node, world sizes 2, 4, and 8 |
| Fallback | Production dispatch falls through when the FlyDSL selector returns false. Direct engine calls raise outside their size and input contracts. |
| Declared FlyDSL pin | `flydsl==0.3.2` |
| Imported FlyDSL | 0.3.2 from `/opt/venv/lib/python3.14/site-packages/flydsl` |
| Required init | Each `_StEngine` zeroes its IPC inbox once during construction. It initializes one color per possible block to 1 once. Kernels then advance nonzero colors and use parity slots. No per-call fill is required. |
| Init in reported timings | No. Constructor allocation, zeroing, handle exchange, and warm-up are outside the kernel timing. This setup is one-time for the communicator lifetime. |
| Retained workspace | Default exact mode keeps only one-shot inboxes. Their TP2/TP4/TP8 sizes are approximately 8.03/1.02/16.06 MiB per rank. PCIe fast mode keeps all reachable ladder engines: approximately 38.49/117.77/154.81 MiB per rank. |
| New test surface | 4 test files, about 150 collected pytest cases, about 31 multi-process spawns in the quick-allreduce suite alone |
| New non-test surface under `op_tests/` | 3 scripts, 3960 lines, plus 3 shape CSVs |
| Diff vs base | 23 files, +9688 / −471 |

Changed product files:
- `aiter/dist/device_communicators/communicator_cuda.py` adds production dispatch and shared-engine lifetime wiring.
- `aiter/dist/device_communicators/flydsl_all_reduce.py` owns family selection and production eligibility.
- `aiter/ops/flydsl/allreduce_policy.py` defines fabric, world-size, accuracy, and payload-size policy.
- `aiter/ops/flydsl/kernels/one_shot_allreduce.py` implements the exact push one-shot kernel.
- `aiter/ops/flydsl/kernels/quick_allreduce_int4_ring.py` implements the ring reduce-scatter and all-gather protocol.
- `aiter/ops/flydsl/kernels/quick_allreduce_codec.py` implements INT4, INT6, and fp16 wire formats.
- `aiter/ops/flydsl/kernels/quick_allreduce_int4.py` and `quick_allreduce_shared.py` refactor the mesh transport and shared cache protocol.
- `aiter/ops/flydsl/one_shot_allreduce.py`, `quick_allreduce_int4.py`, and `quick_allreduce_int4_ipc.py` own validation, allocation, launch, and cleanup.

Changed test and support files (see the P2 testing finding):
- `op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py` extends the existing canonical suite for the quantized all-reduce.
- `op_tests/multigpu_tests/test_flydsl_one_shot_allreduce.py` adds a suite for the new exact one-shot operation.
- `op_tests/flydsl_tests/test_flydsl_allreduce_policy.py` and `test_flydsl_quick_allreduce_codec.py` add two parallel suites for internals.
- `op_tests/multigpu_tests/bench_comm_allreduce.py`, `fit_allreduce_policy.py`, and `transferbench_roofline.py` add benchmark, fitting, and roofline tooling.
- `op_tests/multigpu_tests/shapes/ar_sweep_*.csv` add sweep shape lists.

The kernels launch on the current PyTorch stream when the caller does not give a stream. The persistent schedules clamp their grid before they size inboxes. The ring samples compiled for gfx950 had no spills or scratch. TP2/ST32 used 63 VGPRs and 4608 B LDS. TP4/ST32 used 59 VGPRs and 2304 B LDS. TP8/ST16 and TP8/ST32 with INT6 reduce-scatter used 127 VGPRs and 1664 B LDS. The TP8 result permits four 256-thread workgroups per CU, which matches the conservative fallback used by `clamp_grid_cap`.

The communication protocol drains payload VMEM, joins the workgroup, writes back cacheable inbox data, and only then publishes the flag. Consumers poll with cache-bypassing loads, join the workgroup, and acquire before payload reads. The super-tile loops drain LDS reads before reuse. The manual ring slot formulas state their i32 units and localize the exception to layout coalescing at unit extents. The surrounding HBM, thread, fanout, color, and mesh wire ownership uses FlyDSL layouts and tiled copies.

## Findings

Findings are ordered by severity. Each finding has a location, a trigger, a reason, and a proposed fix.

### P1 — The production selector accepts layouts that both engines reject

**Where:** `aiter/dist/device_communicators/flydsl_all_reduce.py` lines 208-225, `FlyDSLAllReduce.should_fly_all_reduce` (weak-contiguity eligibility); lines 227-232, `fly_all_reduce` (unconditional selected-engine call); `aiter/ops/flydsl/one_shot_allreduce.py` lines 211-245, `OneShotAllReduce._check_payload` (strict contiguity); `aiter/ops/flydsl/quick_allreduce_int4.py` lines 795-828, `QuickAllReduceInt4._check_payload` (strict contiguity)

**Trigger:** Enable the new production path and pass a non-contiguous tensor whose view consumes the full remaining storage. A transpose of a contiguous allocation is one example. The payload must otherwise be in the selected family window and meet the dtype and 16-byte-size checks.

**Reason:** The selector imports `is_weak_contiguous`. That helper returns true when the storage tail equals the logical byte count, even when `Tensor.is_contiguous()` is false. Both selected engines then require `inp.is_contiguous()` and `out.is_contiguous()`. A local probe with a bf16 transpose reported `should_fly_all_reduce=True`; both `_check_payload` methods raised `ValueError`. `CudaCommunicator.all_reduce` has already committed to this path, so the exception replaces the intended fallback.

**Proposed fix:** Make the selector and engine contract identical. One sufficient approach is to require strict contiguity in `should_fly_all_reduce`. Another sufficient approach is to support the accepted weak-contiguous mapping in both engines and in output allocation. Add a production-dispatch test with a full-storage transpose and assert either a successful result or a clean fallback.

### P1 — The accidental `.wavescope-src` gitlink must be removed

**Where:** `3rdparty/.wavescope-src` line 1 (new mode-160000 gitlink added by this PR); `.gitmodules` lines 1-3 (the only mapping is for `3rdparty/composable_kernel`); `setup.py` lines 227-233, the source-build submodule requirement

**Trigger:** Clone the repository recursively, or run `git submodule sync && git submodule update --init --recursive` before a source build.

**Reason:** Git sees `3rdparty/.wavescope-src` as a submodule, but `.gitmodules` has no path or URL for it. Both `git submodule init` and `git submodule status` fail locally with a missing mapping error. This failure blocks normal recursive source setup before the PR code can build. The entry is a local profiling-tool checkout that was staged by accident. It is not a dependency of the all-reduce kernels, nothing in the build or the runtime references it, and the dot-prefixed name shows it was never meant to be part of the source tree.

**Proposed fix:** Remove the gitlink from the PR so `3rdparty/` returns to its base state, and keep the local checkout out of the index (for example through a personal exclude file). Adding a `.gitmodules` entry is not an acceptable alternative here, because it would publish a developer tool as a build dependency of AITER.

### P2 — The test surface diverges from the AITER op-test contract, the standard CI path executes almost none of it, and one new script fails the multi-GPU job

**Where:**
- `op_tests/multigpu_tests/fit_allreduce_policy.py` lines 534-539, `main` (required positional `csv`), lines 757-758 (`__main__` calls `main()` unconditionally)
- `op_tests/multigpu_tests/bench_comm_allreduce.py` lines 2055-2058, `main` (arch gate only), lines 676-678, `L_SHAPE` (13 default shapes), line 2126 (`--iters` default 101), line 1045 (`assert sqnr >= cand.sqnr_floor` inside the timed loop), line 1120 (`assert ca_comm is not None`)
- `op_tests/multigpu_tests/transferbench_roofline.py` lines 623-716, `_self_test` (about 90 assertions on a text parser, reachable only through `--self-test`), lines 768-777 (deliberate self-skip when the third-party binary is missing)
- `op_tests/flydsl_tests/test_flydsl_allreduce_policy.py` lines 1-8 (module docstring, "No GPU required"), lines 62-71, `test_oneshot_ceiling_shrinks_with_world_size` (asserts the xGMI rows are flat), lines 206-222, `test_byte_overrides` (asserts `oneshot_max == (48 << 10)`), lines 254-256 (`pytest.main` entry point)
- `op_tests/flydsl_tests/test_flydsl_quick_allreduce_codec.py` lines 45-57 (imports `_codec_quant`, `_codec_load`, `_codec_dequant`, `_scale_from_word`), lines 82-187, `make_codec_roundtrip_kernel` (a second FlyDSL kernel written inside the test file), lines 336-420 (`degenerate`, ceiling, and sub-floor cases), lines 440-484 (`_resolve` codec-selection unit tests)
- `op_tests/multigpu_tests/test_flydsl_quick_allreduce_int4.py` lines 584-607 and 743-808 (`sqnr`, `bit_exact`, and `allclose` families, each parametrized over `mesh` and `ring`), lines 647-674 (fill cases that repeat the codec suite)
- `op_tests/multigpu_tests/test_flydsl_one_shot_allreduce.py` lines 346-386, `main` (PASS/FAIL print, no `@benchmark`, no `run_perftest`, no `checkAllclose`, no markdown table), lines 351-356 (`-tp` is one value with `choices`, not a swept list)
- `.github/scripts/aiter_test.sh` lines 23-25 (multi-GPU discovery is a recursive `find` over every `*.py`), lines 29-30 (single-GPU discovery is `find op_tests -maxdepth 1`), line 85 (`timeout 60m python3 "$file"`), lines 37-46 (`skip_tests`)
- `.github/scripts/split_tests.sh` line 53 (`find op_tests -maxdepth 1 -name 'test_*.py'`)

**Trigger:** Run the repository's standard single-GPU and multi-GPU test jobs with no `AITER_TEST` override.

**Reason:** Four separate problems share one root cause: the new tests were designed as a pytest estate, while AITER's CI runs `python3 <file>` over a fixed, shallow file list.

1. *A new script makes the multi-GPU job fail.* The multi-GPU runner executes every `*.py` under `op_tests/multigpu_tests/`, recursively, as a script. `fit_allreduce_policy.py` takes a required positional `csv` argument, so a bare `python3` invocation exits with status 2 and the job reports a failed file. I confirmed the argparse contract directly: a parser with `add_argument("csv", nargs="+")` raises `SystemExit(2)` on an empty argument list. `transferbench_roofline.py` already documents this exact runner behavior at lines 768-772 and self-skips; `fit_allreduce_policy.py` does not, and neither is in `skip_tests`.
2. *A 2405-line benchmark becomes a CI test.* The same runner executes `bench_comm_allreduce.py`. Its only guard is the architecture allow-list, so on a supported runner it sweeps 13 shapes at TP4 across every applicable candidate at 101 timed iterations, after JIT-building each FlyDSL ladder engine. It also asserts an SQNR floor per candidate inside the timed loop, which turns a tuning tool into a correctness gate whose failure message points at a benchmark rather than at an operation. These three scripts are tooling, not tests, and `op_tests/multigpu_tests/` is a directory where everything is executed.
3. *The behavioral coverage is not reachable from CI.* Neither `aiter_test.sh` nor `split_tests.sh` descends into `op_tests/flydsl_tests/`, so the 32 policy cases and the 36 codec cases never run; the FlyDSL tests CI does run are the top-level `op_tests/test_flydsl_*.py` files listed in `split_tests.sh` lines 89-172. The two multi-GPU files do run, but as scripts: the quick-allreduce `main()` sweeps `mesh` only at TP8 over three shapes, and the one-shot `main()` runs TP2 only. None of the 46 quick-allreduce pytest cases and none of the 36 one-shot pytest cases execute. The principal new ring protocol, the TP8 INT6 reduce-scatter default, and every codec edge case can regress while CI stays green.
4. *The suites themselves depart from the op-test contract.* `test_flydsl_one_shot_allreduce.py` has no `@benchmark` function, no `run_perftest`, no `checkAllclose`, no TFLOPS or TB-s columns, and no markdown summary; its `-tp` flag is a single behavior selector instead of a swept list, so `itertools.product` never applies. `test_flydsl_allreduce_policy.py` is a CPU-only structural suite with no kernel and no oracle, and it couples to shipped tuned values: line 211 asserts `oneshot_max == (48 << 10)`, and lines 62-71 assert that the xGMI rows are flat with a comment stating the test is meant to fail once a real fit replaces them. Retuning is a normal event and must not turn a suite red. `test_flydsl_quick_allreduce_codec.py` builds a second FlyDSL kernel inside the test file to drive private codec helpers, which makes the wire format a product under test rather than checking the operation through its public wrapper.

On volume: the estate is disproportionate. About 150 collected cases and about 31 multi-process spawns cover an operation family whose distinct behaviors are the two schedules, three wire formats, three world sizes, the partial-tile and super-tile paths, and a handful of E4M3 edge cases. Much of the count is the same case restated. The `allclose` family at lines 781-808 reruns the shapes of the `bit_exact` family at lines 743-779 with a weaker gate. The E4M3 edge cases appear twice, as codec cases at lines 336-420 and as `_FILLS` cases at lines 647-674. `mesh` and `ring` are crossed with every world size in three separate families, although the ring is the schedule this PR introduces and the mesh is already covered at base. A smaller estate that CI actually executes would protect more behavior than this one does.

Two minor items belong with this finding. `bench_comm_allreduce.py` line 131 refers to `op_tests/multigpu_tests/shapes/README.md`, which the PR does not add. The one-shot suite writes rank logs to fixed `/tmp/flydsl_one_shot_allreduce_tp{N}_rank{R}.log` paths (lines 260-263), which collide between concurrent runs on a shared runner.

**Proposed fix:** The outcome to reach is one canonical, CI-executed suite per new operation, with the tooling moved out of the executed directory. One sufficient approach:

- Move `bench_comm_allreduce.py`, `fit_allreduce_policy.py`, and `transferbench_roofline.py` out of `op_tests/multigpu_tests/` into a location the runner does not execute. If they must stay, add all three to `skip_tests` in `aiter_test.sh` and give each a no-argument path that exits 0, as `transferbench_roofline.py` already does.
- Keep `test_flydsl_quick_allreduce_int4.py` as the canonical suite for the quantized all-reduce and reshape `test_flydsl_one_shot_allreduce.py` into the standard form: a module-level reference, a `@benchmark` function timed with `run_perftest` and checked with `checkAllclose`, swept argparse lists, the `get_gfx()` allow-list in `main()`, and a markdown summary table.
- Fold the parts of the two `op_tests/flydsl_tests/` suites that protect real behavior into those canonical files, and delete the rest. Worth keeping: the codec sub-floor, ceiling-saturation, and zero-group cases, expressed through the public wrapper against the fp32 oracle, where the `_FILLS` mechanism already reaches them; and a policy check for family selection, precedence, and fallback built on a small synthetic policy map. Worth dropping: assertions on shipped threshold constants, the xGMI flatness tripwire, the structural checks that restate `FamilyPolicy.__post_init__`, the in-test FlyDSL codec kernel and its private-symbol imports, and the codec-resolution unit tests once the same resolution is observed through the wrapper.
- Reduce the remaining matrices to distinct behaviors: one gate per transport shape rather than a bit-exact and an allclose pass over the same shapes, and the full `mesh` cross-product only where the mesh path actually changed.
- Make the default script sweep of each canonical file small enough to finish inside the 60-minute per-file budget while still touching each supported world size, both schedules, and the partial-tile and super-tile paths.

### P3 — Residual: fabric detection is host-global, not group-specific

**Where:** `aiter/ops/flydsl/quick_allreduce_int4.py` lines 295-333, `has_xgmi_peer_links` (scans every KFD link on the host); lines 336-350, `_resolve_inbox_flags` (selects one memory mode from that result); `aiter/dist/device_communicators/flydsl_all_reduce.py` lines 118-123, `FlyDSLAllReduce.__init__` (selects the policy from the same result)

**Trigger:** A host that mixes PCIe-connected and xGMI-connected GPU pairs, where the process group uses only the PCIe-connected devices.

**Reason:** The detector returns true after it finds any xGMI link on the host. It does not receive the selected device or the process group, so a PCIe-only group on such a host would select the xGMI policy and an uncached inbox, and the source documents that uncached peer writes collapse on PCIe. This is recorded as an accepted assumption rather than a defect: deployed single-node hosts are uniformly PCIe or uniformly xGMI, so the mixed topology this depends on is outside the expected deployment.

**Proposed fix:** No change is required for merge. If the assumption is ever in doubt, classify only the links between the devices in the process group, for example by gathering each rank's device identity over the CPU group, mapping those devices to KFD nodes, and inspecting the peer links the collective uses. Recording the assumption near `has_xgmi_peer_links` would make it visible to the next reader.

## Verification gaps

- `PYTHONPATH=. pytest -q op_tests/flydsl_tests/test_flydsl_allreduce_policy.py` passed: 32 tests.
- `PYTHONPATH=. pytest -q op_tests/flydsl_tests/test_flydsl_quick_allreduce_codec.py` passed on gfx950: 36 tests.
- `PYTHONPATH=. FLYDSL_RUNTIME_ENABLE_CACHE=0 python3 op_tests/multigpu_tests/test_flydsl_one_shot_allreduce.py -tp 2` passed on two gfx950 GPUs.
- Compile-only succeeded for gfx950 ring TP2/ST32 INT4, TP4/ST32 INT4, and TP8/ST16/ST32 INT6-reduce-scatter plus INT4-all-gather. The sampled kernels had no VGPR or SGPR spills, scratch, or AGPR use.
- `python3 -m py_compile` passed for the changed product Python modules. `git diff --check 3312a2b66..HEAD` passed.
- I confirmed the weak-contiguous mismatch with a real gfx950 tensor and both engine validators. I did not run a full distributed fallback reproduction.
- For the CI finding I read `aiter_test.sh` and `split_tests.sh` and confirmed the argparse contract of `fit_allreduce_policy.py` in isolation. A bare `python3 op_tests/multigpu_tests/fit_allreduce_policy.py` in this shell stopped earlier, at `import aiter`, because AITER is not installed in this environment. I did not reproduce the failure on a runner with AITER installed, and I did not run the multi-GPU CI job or `bench_comm_allreduce.py` end to end.
- I did not run ring runtime correctness, run-ahead skew, non-default-stream, graph-capture, communicator-reuse, constructor-failure, AOT, or run-only tests.
- I did not run gfx942 runtime or compile-only checks. The gfx942 target and its resource assumptions remain unverified in this pass.
- I did not reproduce the submitted performance claims or policy thresholds. I did not run the long multi-GPU benchmark or fitting scripts.

PR #5468, opened 2026-09-11
[Config] Retune Kimi-K3 a8w4 fp8 route-out stage2

amd-wsung102:
## Summary

This PR fixes an inconsistency in the Kimi-K3 a8w4 SiTUv2 tuned config: at prefill token counts the two `inter_dim` families disagree about which stage-2 epilogue to use, and the `inter_dim=384` family — the production TP=8 shape — is on the slower one.

* `inter_dim=512` already selects the opus fp8 route-out epilogue (`opus_moe2_layout_afp8_wfp4_fp8_t64x256x256_sbm64_*`) at tokens 2048 through 16384.
* `inter_dim=384` selects it at token 2048 and then reverts to the bf16 reduce epilogue (`flydsl_moe2_layout_afp8_wfp4_bf16_t64x256x128_reduce_sbm64_sp3201`) for 4096 and above.

The top-k combine writes a `[token, topk, model_dim]` scratch buffer and reduces it. At long context that buffer, not the reduction arithmetic, is the cost — it is 16 slots × 3584 wide per token. Storing each element as fp8 + e8m0 instead of bf16 roughly halves that traffic. This PR moves the four `inter_dim=384` rows at tokens 4096–32768 onto the same kernel `inter_dim=512` already ships, so the epilogue choice is monotonic in token count for both families.

**Performance improvement:**

* E2E Kimi-K3 at 8k/1k, TP=8 — 2.87% lower median TTFT, 1.52% higher total throughput.
* Kernel-level 1.02–1.11× on the MoE GEMM pair at prefill, from 1.10–1.25× on gemm2 alone.
* Layer-wise 1.06–1.10× on the whole MoE layer at tokens 4096–32768.

## Relevant Files

* `aiter/configs/model_configs/kimik3_a8w4_tuned_fmoe.csv` — `kernelName2` swapped from `flydsl_moe2_layout_afp8_wfp4_bf16_t64x256x128_reduce_sbm64_sp3201` to `opus_moe2_layout_afp8_wfp4_fp8_t64x256x256_sbm64_rbn3584` on the four `inter_dim=384`, `token>=4096` rows; `us1`/`us2`/`us`/`tflops`/`bw` re-measured for those rows. No other column and no other row changes.

### Prefill — fp8 route-out epilogue on `inter_dim=384`, `token >= 4096`

* The kernel is not new and needs no new validation path. `opus_moe2_layout_afp8_wfp4_fp8_t64x256x256_sbm64_rbn3584` is the exact name the `inter_dim=512` rows at token 8192 and 16384 already ship, and the `_rbn3072` sibling is what `inter_dim=384` already uses at token 2048. This PR removes a gap, it does not introduce a format.

* The win is in gemm2, and it grows with token count, which is what a buffer-traffic saving should do. Same-host A/B of the two epilogues at their own configs, `block_m=64` in both arms:

  | token | gemm2 reduce (µs) | gemm2 route-out (µs) | gemm2 speedup |
  | ----- | ----------------- | -------------------- | ------------- |
  | 4096  | 409.23            | 371.22               | 1.102×        |
  | 8192  | 700.59            | 569.73               | 1.230×        |
  | 16384 | 1264.35           | 1026.44              | 1.232×        |
  | 32768 | 2464.93           | 1975.12              | 1.248×        |

  gemm1 moves by 0.0–2.3% across the same runs, consistent with it being untouched.

* The band edge is measured, not assumed. The route-out loses below the crossover: forced onto the `token<=1024` rows at `inter_dim=384` it is 0.961× at token 512 and 0.916× at token 1024 against the shipped atomic epilogue. Those rows are left alone. This matches the expected mechanism — the quantized buffer only pays once buffer traffic dominates, which it does not at small M where the atomic path is launch-bound.

* The atomic → reduce crossover itself was re-checked and left alone. A first measurement suggested a bf16 reduce beat the atomic epilogue at token 512 by 1.078×, but it did not reproduce: three repeats gave 1.005×, 1.018× and 0.989×, and the kernel-level A/B disagreed in sign with the wall-clock one (0.968× vs 1.032× at token 1024). The effect is inside run-to-run noise, so the existing atomic-through-1024 dispatch is correct and unchanged.

* The change was verified to actually take effect after the edit, not just to measure well in a forced A/B. With the new CSV in place the server selects `opus_moe2_layout_afp8_wfp4_fp8_t64x256x256_sbm64_rbn3584`, and forcing the old reduce kernel back on top of it is 0.927× at token 8192.

## Performance on ATOM

Measured on MI355X (gfx950, 256 CU). Kimi-K3 routed-expert geometry at TP=8: `model_dim=3584`, `inter_dim=384`, E=896, top-16, SiTUv2, fp8 × MXFP4, `AITER_SITUV2_A8W4=1`. `kv_cache_dtype=fp8`, `block-size=128`, prefix caching off. 8k/1k, 128 prompts, concurrency 128.

### Kernel-level

| token | block\_m | gemm1 before → after | gemm2 before → after | gemm1 + gemm2   | speedup |
| ----- | -------- | -------------------- | -------------------- | --------------- | ------- |
| 4096  | 64→64    | 504.07 → 521.11      | 409.23 → 371.22      | 913.30 → 892.33 | 1.023×  |
| 8192  | 64→64    | 725.81 → 737.17      | 700.59 → 569.73      | 1426.40 → 1306.90 | 1.091×  |
| 16384 | 64→64    | 1178.33 → 1178.93    | 1264.35 → 1026.44    | 2442.69 → 2205.37 | 1.108×  |
| 32768 | 64→64    | 2065.48 → 2112.97    | 2464.93 → 1975.12    | 4530.41 → 4088.09 | 1.108×  |

### MoE layer

| token | before (µs) | after (µs) | speedup | bucket changed |
| ----- | ----------- | ---------- | ------- | -------------- |
| 512   | 431.67      | 431.67     | 1.000×  | no — control   |
| 1024  | 458.28      | 458.28     | 1.000×  | no — control   |
| 4096  | 1034.06     | 978.85     | 1.056×  | yes            |
| 8192  | 1552.94     | 1426.80    | 1.088×  | yes            |
| 16384 | 2618.02     | 2386.49    | 1.097×  | yes            |
| 32768 | 4782.25     | 4365.08    | 1.096×  | yes            |

| Metric                    | main      | kimi\_k3\_a8w4\_fp8\_routeout | Delta      | Change  |
| ------------------------- | --------- | ----------------------------- | ---------- | ------- |
| Mean TTFT (ms)            | 32,804.92 | 31,948.10                     | \-856.82   | \-2.61% |
| Median TTFT (ms)          | 32,987.37 | 32,039.31                     | \-948.06   | \-2.87% |
| P99 TTFT (ms)             | 63,808.21 | 62,181.06                     | \-1,627.15 | \-2.55% |
| Mean TPOT (ms)            | 75.96     | 75.19                         | \-0.77     | \-1.01% |
| Median TPOT (ms)          | 75.78     | 75.10                         | \-0.68     | \-0.90% |
| P99 TPOT (ms)             | 106.49    | 104.84                        | \-1.65     | \-1.55% |
| Mean ITL (ms)             | 75.89     | 75.12                         | \-0.77     | \-1.01% |
| Output throughput (tok/s) | 1,185.79  | 1,203.76                      | +17.97     | +1.52%  |
| Total throughput (tok/s)  | 10,672.07 | 10,833.81                     | +161.74    | +1.52%  |
| Duration (s)              | 110.54    | 108.89                        | \-1.65     | \-1.49% |

## Accuracy

* gemm1 is bit-identical. `kernelName1` and `block_m` are unchanged in all four rows, and the fp8 activation quantization and fp8 intermediate are untouched.
* The affected buckets are prefill only (`token >= 4096`). Every `token <= 2048` row, and every `inter_dim=512` row, is byte-identical.

## Submission Checklist

- [x] Look over the contributing guidelines at <https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests>.


yifehuan:
LGTM, it could be merged when you resolve the conflict and pass the ci workflow

yifehuan (approved):
lgtm

PR #5472, opened 2026-09-12
[Triton/Gluon] Keep automatic W4A16 backend selection quiet

andyluo7:
## Summary

- keep `backend=None` as silent architecture-based auto-selection;
- preserve the existing warning and Triton fallback when `backend="gluon"`
  is explicitly requested on an unsupported architecture; and
- add focused regression coverage for automatic and explicit selection.

The A16W4 MoE Gluon implementation currently supports gfx1250. On gfx942 and
gfx950, automatic selection is therefore expected to choose Triton. The old
selector treated `None` like an explicit Gluon request and emitted a warning
for every GEMM invocation, which produces substantial log noise in vLLM's
DeepSeek-V4.1-Flash path.

This addresses the AITER-side cause of vllm-project/vllm#56540 without changing
kernel selection or kernel code.

ROCm/aiter#5471 appeared while this patch was being prepared and overlaps this
change. This version centralizes backend resolution, avoids process-global
warning state, and includes committed regression tests. If maintainers prefer
#5471, the tests and selection split here can be reused there.

## Validation

- `ruff==0.16.0` on both changed files
- `black --check` on both changed files
- Python bytecode compilation on both changed files
- `git diff --check`

The focused tests do not launch a GPU kernel. GPU execution was not available
locally; kernel dispatch and kernel arguments are unchanged.

AI assistance was used for investigation and implementation. Human review is
requested before merging.

### DIFF
--- a/aiter/ops/triton/moe/moe_op_gemm_a16w4.py
+++ b/aiter/ops/triton/moe/moe_op_gemm_a16w4.py
@@ -30,6 +30,27 @@ def _is_gluon_available():
         return False
 
 
+def _resolve_backend(backend: str | None) -> str:
+    """Resolve automatic selection without warning on the expected fallback."""
+    if backend is None:
+        return "gluon" if _is_gluon_available() else "triton"
+
+    backend = backend.lower()
+    assert backend in (
+        "triton",
+        "gluon",
+    ), f"Unknown backend '{backend}', must be 'triton' or 'gluon'"
+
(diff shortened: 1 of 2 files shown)


andyluo7:
Closing as superseded by #5471, which has merged the AITER-side fix. This branch also contains focused backend-selection regression tests; I can submit those separately if maintainers would like them.

PR #5475, opened 2026-09-12
[FlyDSL] gather_kv_b_proj: lift the 4 GiB cache limit, and let callers ask before committing

valarLip:
Two things about `aiter/ops/flydsl/gather_kv_b_proj` and its contract with the caller. They are separate commits and could be split if you prefer; the second touches the guards the first moved, so it is stacked rather than parallel.

### Motivation

The op refused a KV cache whose byte span reached `2**31`. DeepSeek-R1-0528 tp4 reaches it at 4,458,592 pages, so every MLA prefill that hit the guard took down the model runner:

```
ValueError: [FlyDSL gather_kv_b_proj] num_blocks=4458592 x 576 overflows 32-bit buffer indexing
[atom] AsyncIOProcManager(ModelRunner): [ModelRunner3/4] proc died unexpectedly (exitcode=1)
```

Three distinct 32-bit edges sat behind that one message, and only the third is real:

1. **Host ABI.** FlyDSL packs memref *shapes* as i32 (`jit_argument.py`), and the op handed the cache over flattened, so `shape[0] = num_blocks * 576` broke the C-ABI marshalling before any kernel ran — `struct.error: 'i' format requires ...`. The cache now goes in as a pointer plus `num_blocks`, and the extent only ever widens in 64 bits.
2. **The `idx * 576` gather offset.** Measured on gfx950: the raw-buffer voffset is read unsigned, so the i32 wrap addresses correctly all the way to 4 GiB. The `2**31` guard was one bit conservative and the arithmetic needed no change at all.
3. **`num_records`, a 32-bit byte count.** This one is real, and a gather cannot escape it the way #4473's tile scan does: an SRsrc base has to stay uniform across the wave, and a gathered row is per-lane.

### Technical details

Past 4 GiB the A path drops the descriptor for `rocdl.global.load.lds` — the same direct-to-LDS DMA, driven by a per-lane 64-bit pointer instead of an SRsrc. It is selected as a compile-time `wide_index` variant from the cache's own extent, so a cache that fits a descriptor keeps the descriptor.

Two things the copy atom did implicitly and the wide form does by hand: lane `i` lands at `lds_base + i*16`, and the K offset rides in the address, there being no `soffset` operand. That last point is the inverse of #4473's warning — with no descriptor there is no second origin to mismatch, so folding it in is unconditionally right. The narrow path still passes `soffset`, and now says so in a comment, because a future rebase there would have to move it into the base or read zeros silently.

Both forms are kept. They measure the same, so the split is not for speed: the descriptor form keeps the hardware bounds check, which the wide form gives up (an out-of-range `kv_indices` entry faults instead of returning zeros).

### Test plan

`op_tests/test_flydsl_gather_kv_b_proj.py` stopped at `num_blocks=3_000_000`, just under the 3,728,270 boundary, which is why this was never caught. Added cases brackets the switch at one row either side of 4 GiB, plus one well past it and the R1 shape:

| num_blocks | span | path | |
|---|---|---|---|
| 4,458,592 | 2.39 GiB | descriptor | PASS — the shape CI died on |
| 7,456,540 | 4.00 GiB, the widest a descriptor spans | descriptor | PASS |
| 7,456,541 | 4.00 GiB, the narrowest it does not | `global_load_lds` | PASS |
| 9,000,000 | 4.83 GiB | `global_load_lds` | PASS |

Each zero-fills the cache outside the gathered rows, so a misaddressed load can only come back as zeros — `_make_case` tiles its content every 4096 rows, which would let a wrong address return something plausible. `checkAllclose` only warns, so these assert on cosine.

### Test result

```
                 M = 2048   8192   16384
narrow (us)         16.06  33.65   58.90     base 16.07 / 33.69 / 58.54
wide / narrow       1.007x 1.001x  0.995x
```

The wide path costs nothing measurable, and the descriptor path is unchanged.

### Motivation

The op refuses a good deal more than it accepts — non-gfx950, `page_size > 1`, bf16 caches, unquantized and MXFP4 weights, head dims other than 128 — and it refuses by raising. A caller holding a fallback has no way to find out before committing.

ATOM therefore gated on "did the import succeed", and in one CI run three accuracy jobs died mid-forward on shapes `aiter.ops.triton.gather_kv_b_proj` — same arguments, positional drop-in — serves without complaint:

```
ValueError: [FlyDSL gather_kv_b_proj] an unquantized weight (kv_proj_scale=None) is not supported
ValueError: [FlyDSL gather_kv_b_proj] k_buffer must be torch.float8_e4m3fn (OCP e4m3), got torch.bfloat16
```

### Technical details

Adds `gather_kv_b_proj_flydsl_supported`. The guards move into `_unsupported_reason` / `_config_reason`, which return the reason as a string; the op raises it and the predicate compares it against None. One copy, two outlets — a second copy written for the predicate would drift, and drift reads as "declines a shape it handles" or, worse, "accepts one it does not".

Everything the predicate looks at is fixed once the weights are loaded and the cache is allocated, so a caller asks once and keeps the answer. What stays inside the op is per-call — `num_tokens` against the workspace, `kv_indices` length — and is a caller bug rather than a shape this backend declines, so it still raises.

### Test plan

`33 passed` (was 25). Beyond the size cases above:

- the predicate and the op never disagree, on an accepted config and on two rejected ones;
- the three shapes ATOM actually hands this op that only the Triton one covers — bf16 cache, `kv_proj_scale=None`, MXFP4 weight — are each declined by the predicate *and* raised by the op.

The ATOM-side change that consumes this is separate. Without it nothing behaves differently; with an older aiter, ATOM's import falls into its existing `except` and the FlyDSL gather simply stays off.

---

Reviewed against gfx950 hardware for every number quoted. `ruff` and `black` clean.

### DIFF
--- a/aiter/ops/flydsl/kernels/gather_gemm_8wave.py
+++ b/aiter/ops/flydsl/kernels/gather_gemm_8wave.py
@@ -5,8 +5,10 @@
 
 import flydsl.compiler as flyc
 import flydsl.expr as fx
-from flydsl._mlir import ir as _ir
-from flydsl._mlir.dialects import llvm as _llvm_d
+
+# Raw dialect, and the only one here: `global_load_lds` has no fx wrapper (the
+# expr/rocdl surface stops at the buffer form). Pointers still go through
+# `fx.to_llvm_ptr`, so no address space is hardcoded.
 from flydsl._mlir.dialects import rocdl as _rocdl_d
 from flydsl.expr import T, const_expr, range_constexpr, rocdl
 from flydsl.expr.rocdl import cvt_pk_f32_fp8
@@ -33,17 +35,12 @@ N_WAVES = 8  # 512 threads
 BLOCK_K = 128  # fixed by MFMA_Scale(16, 16, 128)
 
 
-def _ir_ptr(space: int):
-    """``!llvm.ptr<space>`` -- the raw ROCDL ops take these, not ``!fly.ptr``."""
-    return _ir.Type.parse(f"!llvm.ptr<{space}>")
-
-
 def _raw(v):
     return v.ir_value() if hasattr(v, "ir_value") else v
 
 
 def _gather_a_offsets(
-    kvi_div, m_base, lane_id, wave_id, n_rounds, half_row_off, kv_base=None
+    kvi_div, m_base, lane_id, wave_id, n_rounds, half_row_off, wide=False
 ):
     """Per-lane A sources: gathered kv_cache row + the stock XOR-swizzled column.
 
@@ -58,10 +55,10 @@ def _gather_a_offsets(
     invariant across rounds and halves -- one VGPR for the whole tile rather
     than re-emitting ``swizzle_128``'s divide chain per step.
 
-    Element offsets by default; with ``kv_base``, the cache's address, absolute
-    64-bit byte addresses. Row and column widen together here because they are
-    added here: split across two functions the two widths could disagree, and
-    an i32 sum of a 64-bit address is not an error, just a wrong row.
+    Offsets are in fp8 elements; ``wide`` makes them 64-bit, for a cache the
+    32-bit form can no longer reach. Row and column widen together here because
+    they are added here: split across two functions the widths could disagree,
+    and an i32 sum against a 64-bit row is not an error, just a wrong row.
     """
     atom = fx.make_copy_atom(fx.rocdl.BufferCopy32b(), fx.Int32)
     regs = []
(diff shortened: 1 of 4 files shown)


aiter/ops/flydsl/kernels/gather_gemm_8wave.py
```diff
@@ -47,20 +70,59 @@ def _load_row_bases(kvi_div, m_base, lane_id, wave_id, n_rounds, half_row_off):
...
+        for step in range_constexpr(self.n_load_steps):
+            step_off = self.wave_id * self.LDS_BYTES_PER_WAVE_STEP + step * (
+                self.n_waves * self.LDS_BYTES_PER_WAVE_STEP
+            )
+            lds_off = (
+                fx.Int32(fx.ptrtoint(lds_dst.ptr))
+                + fx.Int32(step_off)
+                + self.lane_id * 16
+            )
+            _rocdl_d.global_load_lds(
```
coderfeli, line 118: kernel code cleanup skill to clean raw dialects?

PR #5482, opened 2026-09-13
[FlyDSL] optimize mxfp4 moe kernels perf

binding7012:
## Motivation

optimize flydsl mxfp4 moe kernels perf

## Test Result

image


aiter/fused_moe.py
```diff
@@ -3663,6 +3669,37 @@ def get_block_m() -> int:
         and use_g1u1
         and not doweight_stage1
     )
+    _mxmoe_fallback_ok = (
```
Bernard-Liu, line 3672: > Please exclude EP from both this fallback and unsupported tuned MXMOE configurations. Both can select
> output_aux='opus', causing NotImplementedError when expert_mask is provided. The newly selected Kimi-K3 rows at  tokens 4096/16384/32768 bypass this fallback, so adding not is_ep here alone would leave those EP calls broken.

aiter/fused_moe.py
```diff
@@ -3663,6 +3669,37 @@ def get_block_m() -> int:
...
+        dtype in [dtypes.bf16, dtypes.fp16]
```
Bernard-Liu, line 3673: > This condition accepts FP16, but the fallback selects a BF16-only GEMM2. An untuned SiTUv2/a4w4 call with FP16 output (M=128, H=3072, I=512, E=128, topk=4) raises TypeError:
  > FlyDSL v2 GEMM2 supports only torch.bfloat16 output, got torch.float16. Please restrict this fallback to BF16 or add FP16 support to the selected backend.

aiter/ops/flydsl/mxfp4_kname.py
```diff
@@ -30,6 +30,7 @@
         (64, True, False),
         (64, False, False),
         (128, False, False),
+        (160, False, False),
```
Bernard-Liu, line 33: > Adding BM160 here breaks default --mxfp4-flydsl tuning: _g1_matches_m_est() has no BM160 rule and raises ValueError: Missing GEMM1 pruning rule for BM160. Full search avoids this exception but still filters out all BM160 pairs because stage2 enumeration and pairing were not updated. Please add the pruning rule and support the required GEMM1/GEMM2 tile combinations.

aiter/ops/flydsl/kernels/mxmoe_dispatcher.py
```diff
@@ -129,25 +129,32 @@ def compile_gemm2_a4w4_port(
...
     if BN not in (128, 256, 512) or BK not in (128, 256):
         raise AssertionError(
             "mxfp4_moe_gemm2 supports only "
             f"(BN in {{128,256,512}}, BK in {{128,256}}); got (BN={BN}, BK={BK})"
         )
-    if SBM % BM != 0:
-        raise AssertionError(f"SBM ({SBM}) must be a multiple of BM ({BM})")
+    if SBM < BM:
+        raise AssertionError(f"SBM ({SBM}) must be >= BM ({BM})")
+    if SBM % BM and (SBM % 32 or BM % 32):
```
Bernard-Liu, line 154: > I haven’t reproduced a GPU memory error with the current tests, but there may be an issue with tightly sized allocations. For SBM=160, BM=64, max_sorted=160, the last tile covers rows 128–191. _write_i does not clamp its sorted_weights loads, and the A buffer bound is rounded up to 192 rows. Masking invalid output rows may hide these reads from numerical checks. Could we use the actual allocation bounds, clamp the tail loads, and add a focused boundary test?

PR #5490, opened 2026-09-14
[Triton/Gluon] [JIT] fix: resolve read-only tuning config failures and Python 3.10 JIT import errors

junna2016:
## Motivation

Fix two issues encountered while integrating AITER into RTP-LLM:

1. MoE configuration loading fails in read-only installations, such as Bazel CI runfiles, because duplicate-resolution logic attempts to overwrite packaged CSV files.
2. Importing AITER from source on Python 3.10.9 can fail with `ValueError: No function definition found for kernel` when Triton/Gluon inspects functions with inline-lambda heuristic decorators.

### Resolve configuration collisions without modifying package files

Update `AITER_CONFIG.update_config_files()` to resolve duplicate configurations in memory:

- Preserve the existing selection rule: retain the lowest-`us` entry, with stable ordering for ties.
- Write only the merged result to the existing cache destination.
- Never overwrite source configuration files.
- Continue on the first invocation instead of raising an exception requesting a rerun.
- Preserve the error for duplicate configurations without a `us` column.

Update existing configuration tests and remove the obsolete source-rewriting `--fix` workflow.

### Fix Python 3.10 source inspection during JIT initialization

Replace the affected inline-lambda heuristics with named functions and `functools.partial` in the gfx1250 Gluon and Triton MXFP4 kernels.

On Python 3.10.9, source inspection can truncate these decorated functions before their `def` statement, causing JIT initialization to fail. The revised decorators avoid this behavior without globally patching Python's `inspect` module.

The heuristic conditions and GPU kernel bodies are unchanged.

## Test Plan

Run directly from the AITER source checkout:

```bash
PYTHONPATH="$PWD" python3 -m unittest \
  op_tests.test_config_merge_readonly \
  op_tests.tuning_tests.test_config_shape_collision \
  op_tests.test_mxfp4_jit_import \
  -v
```

Coverage includes:

- Read-only source CSVs and verification that their contents remain unchanged.
- Lowest-latency selection, stable tie handling, distinct tags, and repeated calls.
- Existing configuration-family merge tests.
- Importing the affected modules and inspecting decorated function source.
- Heuristic results for aligned and unaligned dimensions.

Run Black and Ruff checks on the modified files.

## Test Result

Validated in a Python 3.10.9 / Triton 3.8.0 ROCm environment:

- AITER imports successfully directly from source.
- All **19 tests passed**.
- Black, Ruff, and `git diff --check` passed.

No GPU kernel bodies were changed. This validation does not include gfx1250 device execution, kernel performance measurements, or Python 3.11/3.12 testing. The original RTP-LLM FP8 MoE CI target should be rerun with a wheel containing these fixes.

## Submission Checklist


aiter/jit/core.py
```diff
@@ -420,29 +420,17 @@ def update_config_files(self, file_path: str, merge_name: str):
                     .index
                 )
 
-                saved_files = []
```
yzhou103, line 423: The read-only install crash is only possible if the installed configs already contain duplicate shapes — which means the source branch itself is dirty. The RuntimeError is intentional: it serves as the enforcement gate during CI build and unit tests to keep main free of duplicates. If main is clean, read-only installs will never hit this code path.

junna2016: Thanks for clarifying that the RuntimeError intentionally enforces duplicate-free configs in CI.

We encountered this issue in RTP-LLM using an AITER wheel built and provided by Zhiquan Chen (陈智泉) from AMD. We have not confirmed which source branch or additional config changes were used to build that wheel. Could you help check with him to confirm the wheel's source and packaging process?

I agree that the duplicate-detection gate should be preserved. The read-only installation exposed the issue as a PermissionError because the current path attempts to rewrite the installed config files before raising the intended RuntimeError. We can revise the fix to report the duplicates without modifying the source files, while keeping the CI enforcement behavior.

> The read-only install crash is only possible if the installed configs already contain duplicate shapes — which means the source branch itself is dirty. The RuntimeError is intentional: it serves as the enforcement gate during CI build and unit tests to keep main free of duplicates. If main is clean, read-only installs will never hit this code path.

PR #5493, opened 2026-09-14
feat(mori-ep): add the low-latency kernels and a cached handle set

JiaoliangYu:
`MoriAll2AllManager` could only create the existing `IntraNode` and `InterNodeV1` EP kernels. Callers that needed `AsyncLL` or `InterNodeV1LL` had to construct `EpDispatchCombineOp` themselves, duplicating kernel selection and configuration outside the communicator.

This PR makes kernel selection depend on the requested latency mode and the topology measured from the process group:

- regular, intra-node: `IntraNode`
- regular, inter-node: `InterNodeV1`
- low latency, intra-node: `AsyncLL`
- low latency, inter-node: `InterNodeV1LL`

`low_latency` defaults to `False`, so existing callers retain their previous kernel, configuration, and cache identity. Low-latency launch geometry follows the WideEP reference values and remains configurable through `MORI_EP_WARP_PER_BLOCK`, `MORI_EP_BLOCK_NUM`, and `MORI_EP_RDMA_BLOCK_NUM`. `num_qp_per_pe=2` is added only for low-latency kernels.

The handle cache supports multiple reusable instances of one configuration through:

```python
get_handle(kwargs, index=0)
```

Index `0` preserves the existing shared-handle behavior. Callers with concurrent operations select another index for each additional in-flight slot. The index participates in the cache key but is not passed into the MoRI configuration. The same indexed-cache API is exposed by the FlyDSL manager for consistency, while `create_handle()` remains available for deliberately uncached FlyDSL ops.

ROCm/ATOM#2228 is the consumer integration. It passes the requested latency mode to the manager and requests two indexed handles when TBO is enabled, one per concurrent ubatch; otherwise it requests only index `0`.

Validation:

- Existing default kernel rows and geometry remain unchanged.
- Indexed-cache smoke coverage confirms that `(kwargs, index)` controls identity, index `0` preserves the old singleton behavior, and negative indexes are rejected.
- Verified against MoRI `1.2.3.dev20260824`: both low-latency kernel types exist, and split `dispatch_recv`/`combine_recv` is supported only by `AsyncLL`.
- Style, repository dependency, signal, wheel, and MI300X OPUS checks pass on the current PR head.


aiter/dist/device_communicators/all2all.py
```diff
@@ -112,6 +137,25 @@ def get_handle(self, kwargs):
         )
         return handle
 
+    def get_handles(self, kwargs, count: int):
```
ZhangLirong-amd, line 140: Can we unify get_handle and get_handles into one index-based method?

JiaoliangYu: make sense !

ZhangLirong-amd (dismissed):

ZhangLirong-amd (approved):

PR #5494, opened 2026-09-14
[FlyDSL] Widen the FlyDSL gather + kv_b_proj backend past a 128+128 head

XiaobingSuper:
Lets the fused gather + `kv_b_proj` backend serve a head whose k and v dims differ, which is what unblocks GLM-5.2 (192+256). Before this it required `qk_nope_head_dim == v_head_dim == 128` and raised otherwise, so ATOM had to run GLM-5.2 with `ATOM_USE_FLYDSL_GATHER_KV_B_PROJ=0` and fall back to the Triton op.

## What the constraint actually was

Not a tile boundary. The k/v split is at the **B LDS half** boundary -- `c00/c10` come from half 0 and land in `k_prefix`, `c01/c11` from half 1 and land in `v_prefix`. The real requirement is that each half is a multiple of 64 columns and the split falls on the half boundary.

Both halves are therefore padded to `round_up(max(nope, v_dim), 64)`. The weight overlap is read-only, and the k half's dead tail is dropped by the store predicate `StoreC.store` already implements. A 128+128 head computes 128 and is unchanged by construction.

## Why a wide head was slow, and what fixed it

Padding was not the cost. Single-workgroup latency tracks `BLOCK_N` -- 4.7, 6.4, 8.3, 12.7 us at 128, 256, 384, 512 -- and a true 256+256 head costs 12.7 against the padded 192+256's 13.1. At 512 columns the tile fills all 160 KiB of LDS, so one workgroup resides per CU with nothing to hide its latency, and `BLOCK_M` is capped at 128; at low m the grid does not cover the GPU at all.

So a head too wide to hold a 256-row tile is now covered by two workgroups, split **by column**: each tile takes `LDS_BLOCK_N` columns of both outputs and keeps the half 0 -> k, half 1 -> v routing, so which tensor a store targets stays a compile-time fact. Splitting by output instead would make it a runtime choice, which FlyDSL's AST rewriter cannot lift out of a dynamic `if`.

`block_m` now also reads the row count -- the 128-row tile is faster while its grid still covers the GPU in one pass -- and `xcd_swizzle` was picking its tile count from a hardcoded 256-row tile, which is wrong whenever `block_m` is 128.

The software pipeline is untouched: no change to `wait_barrier`, `vmcnt`, `SAFE_VMCNT` or `N_LDS_ROUNDS`.

## Relation to #5475

#5475 landed first and changed the same two files, so this is rebased on top of it rather than beside it.

- **Guards.** Both relaxations live in `_config_reason`, the callback #5475 made canonical, so the op and `gather_kv_b_proj_flydsl_supported` cannot disagree: `nope % 16 == 0 and v_dim % 16 == 0` replaces the `== 128` equality, and the block-scale rejection below sits next to it. `_config_reason` gained `head_tiles` and `per_row_scale`; `_unsupported_reason` fills both from the tensors it already has. 192+256 with a per-output-row scale reports **supported**, so ATOM routes to it; the unsupported combinations report a reason instead of raising, which is what lets ATOM fall back.
- `_config_reason` now defaults `block_m` to `None` rather than 256, because that is the op's default and the op then picks a tile that fits. Checking LDS against 256 would decline a head the op would serve at 128.
- **Cache ABI.** `wide_index` and the pointer + `num_blocks` signature are taken as-is; the `num_blocks` ceiling this PR used to carry is deleted, since past `_BUFFER_SPAN_MAX` the kernel addresses the cache in 64 bits per lane instead of through the descriptor.
- **The two variants are orthogonal, and tested together.** `wide_index` changes the A path (`_WideG2SLoader`, `global_load_lds`) and `_RopeCopy`'s *source*; `head_tiles` changes the B path offsets, the epilogue's column routing and `_RopeCopy`'s *destination*. `head_tiles` leaves `N_LDS_STEPS_A`, `N_LDS_STEPS_B` and `SAFE_VMCNT` at the unsplit kernel's values -- a split head changes which columns a workgroup owns, not how many loads it issues -- so the vmcnt hand-count #5475 depends on is untouched. The descriptor-span test gained a GLM row at 9,000,000 blocks (4.83 GiB) so a split head and the wide A path run together against the float32 reference.

## Performance

Device time against the Triton op under CUDA-graph replay, so the host floor does not mask it. Re-measured on this PR's head after the rebase; the whole table is from one run, not spliced.

192+256, 16 heads, on the tensors a live GLM-5.2 server hands the op (weight 7168x512 fp8, per-output-row scale 7168x1):

| m | triton us | this PR us | speedup | off best pin |
|---:|---:|---:|---:|---:|
| 128 | 9.6 | 6.8 | 1.40x | -0.2% |
| 256 | 9.9 | 7.8 | 1.27x | 0.0% |
| 512 | 11.0 | 8.8 | 1.24x | 0.1% |
| 929 | 19.7 | 11.9 | 1.66x | 5.1% |
| 1024 | 19.9 | 11.8 | 1.69x | 2.1% |
| 1536 | 28.2 | 15.6 | 1.81x | 3.8% |
| 2048 | 36.9 | 17.2 | 2.15x | 2.4% |
| 4096 | 127.7 | 29.2 | 4.38x | -0.1% |
| 8192 | 271.4 | 53.5 | 5.07x | 0.6% |
| 16384 | 565.2 | 102.7 | 5.50x | -1.6% |
| 32768 | 1067.7 | 218.6 | 4.88x | 3.3% |

Below m=512 this shape used to be 0.68-0.71x, i.e. slower than Triton. The last column is the default config against the best of the eight `(block_m, xcd_swizzle)` pins at that m; negative means the default won. Re-timing one identical config spreads about 3%, so anything inside that is noise.

128+128 keeps one tile per head -- splitting it measured worse at high m -- and gains only from the row-aware tile:

| m | 128 | 512 | 929 | 1536 | 2048 | 4096 | 16384 |
| this PR | 0.86x | 0.84x | 1.31x | 1.64x | 1.54x | 1.91x | 1.18x |
| before | 0.57x | 0.59x | 0.98x | 1.30x | 1.28x | 1.95x | 1.23x |

From m=4096 up the row count resolves to the same `BLOCK_M=256` the old code pinned, so those columns run the same kernel and differ only by noise. The real gain is the m=929..2048 band. Below m=512 this shape is still short of the Triton op, just far less so than before.

## A latent miscompute, now rejected

A 128x128 block scale indexes scale row `head*2 + half`, which assumes each half is exactly one 128-row scale block. A 448-wide head makes one block straddle the k/v split, which the one-scalar-per-half epilogue cannot represent -- measured cosine 0.78. Block scale is now rejected for unequal head dims, with a test asserting the predicate and the op agree. Per-output-row scale, which is what an online-quantized checkpoint produces, is unaffected.

## Correctness

`op_tests/test_flydsl_gather_kv_b_proj.py`: **46 passed**. Every correctness test now runs both head shapes, including the Triton cross-check and rope completeness at all three `BLOCK_M`.

Two bitwise baselines, both taken against this PR's actual base (`e471f2c35`) rather than carried over from an earlier one:

- **128+128 is byte-identical to upstream itself.** 16/16 configurations, SHA-256 over the raw outputs, against the same base commit with this PR's changes absent: both scale modes, both weight layouts, the row tail, m=1, duplicate slots, non-unit k_scale, 12 and 16 heads, `BLOCK_M` 128/256/384.
- **The rebase moved nothing.** 24/24 configurations bitwise identical on int16 views between the pre-rebase branch and the rebased branch -- 512/1000/2048/8192 rows x 12/16 heads x `BLOCK_M` 128/256, the default path, row-major weight, the block-scale epilogue, and GLM 192+256. So absorbing #5475's pointer + `num_blocks` ABI and guard restructuring changed no output byte.

At 192+256 the result matches the Triton op to cosine 1.0000000 (k) / 0.9999999 (v) with max abs diff 1.5e-05, and a float32 reference to 0.9999987 -- exactly what Triton scores against that same reference. The rope tail is bitwise equal.

End to end: GLM-5.2-MXFP4, TP4 on MI355X, served through ATOM (which routes on `gather_kv_b_proj_flydsl_supported`), 5-shot GSM8K over the full 1319 questions, the two legs differing only in `ATOM_USE_FLYDSL_GATHER_KV_B_PROJ`:

| leg | gsm8k exact_match | stderr |
|---|---|---|
| this backend | 0.9469 | 0.0062 |
| Triton op | 0.9484 | 0.0061 |

Two questions out of 1319, a quarter of the combined standard error. Both legs log about 100,540 gather calls per rank at `head_dims=(192, 256)`, `n_heads=16`, so the compared path is the one under test, and on the flydsl leg the predicate answered supported for the live tensors rather than being bypassed.

## Notes

`xcd_swizzle`'s ladder was calibrated for one tile per head, and a split head doubles the N tile count, so for `head_tiles > 1` it collapses to a single row cutoff: `0` at or below 896 rows, `4` above. The constant was chosen by worst-case regret against the best pinned config over m=128..32768, not fitted per band, so there is one cutoff rather than a second ladder to keep calibrated. Measured residual is at most 5.1%, at m=929, and from m=4096 up the default is the best pin within noise. The `head_tiles == 1` ladder is untouched. `_default_xcd` also takes `block_m` now, so its tile count matches the tile actually launched instead of a hardcoded 256.

Each distinct `(block_m, xcd_swizzle)` compiles its own kernel, so a 192+256 deployment now spans two instead of one; a server crossing the boundary pays one extra JIT compile and then hits the cache. The choice is made on the host before launch, so a captured graph pins whatever config its capture-time row count selected.

`N_LDS_ROUNDS` takes the max of the A and B step counts, so it over-issues `kv_indices` loads when they differ -- 3 wasted dword loads per lane instead of 1 at `BLOCK_M=128`. They are clamped by the descriptor and consumed before any G2S load. Pre-existing, slightly worse here.


coderfeli (approved):
LGTM

PR #5500, opened 2026-09-14
[Config] [Tune] Add GLM-5.3-Flash a8w8 blockscale fused-MoE configs for gfx942

jin-amd:
## What

Adds two files under `aiter/configs/model_configs/`:

- `a8w8_blockscale_tuned_fmoe_glm5_3_flash.csv` — five tuned rows for the routed fused-MoE shape GLM-5.3-Flash dispatches on gfx942 / cu_num 304.
- `a8w8_blockscale_untuned_fmoe_glm5_3_flash.csv` — the tuner input, all fifteen token buckets rather than the five that were kept, so the search space behind the numbers below is reproducible.

Kernel-time attribution on a TP4 serving run put the routed MoE at 15.1% of prefill forward time against 2.1% for the a8w8_blockscale GEMMs, and the MoE was still running default kernels, so this is where the remaining headroom is.

## Shape

```
model_dim 4096, inter_dim 512, expert 288, topk 8, Silu,
bf16 activations, fp8 e4m3fnuz per_1x128 blockscale, g1u1, no doweight_stage1
```

`inter_dim` is 512 because `moe_intermediate_size` 2048 is sharded over TP4. Tuning aiter's power-of-two token ladder covers every token count, since the lookup rounds to these buckets.

## Environment

Both tuning runs used the same container and the same aiter working tree,
mounted at `/aiter`.

| | |
|---|---|
| GPU | 8× AMD Instinct MI325X, device `0x74a5` |
| Arch / CU | `gfx942`, `cu_num=304` — the table key |
| Driver | `amdgpu` 6.12.12 |
| ROCm (host) | 6.4.2-120 |
| Container | `amdsiloai/vllm:vllm-openai-rocm-glm5.3-flash-mi325-08092026` |
| ROCm (container) | 7.2.3, HIP 7.2.53211 |
| PyTorch | 2.12.0+git6bbd260 |
| vLLM | 0.28.1rc1.dev278+g73029d424 |
| aiter | this branch — `d2bb0945e` at the main tuning run, `76cd9af0e` at the small-batch run, `024bae1` as submitted |
| Parallelism | `--mp 8`, all eight GPUs, node verified idle before each run |

`gfx942` alone is ambiguous — MI300X shares it — so the identifying key is
`cu_num=304` together with the part above.

The aiter tree was bind-mounted rather than installed, so the commits above are
the state of the branch at each run rather than a guarantee the working tree was
clean. The tuner, its inputs and its outputs are all in this branch.

## Commands

All three are `csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py`, run inside the
container above with the aiter tree at `/aiter`:

```bash
docker run --rm --device /dev/dri --device /dev/kfd --group-add video \
  --shm-size 32G --ipc host \
  -v <aiter>:/aiter -w /aiter -e PYTHONPATH=/aiter \
  -v <logdir>/compare:/tmp/aiter_compare \
  -e TUNE_ONLY=asm,cktile \
  --entrypoint bash \
  amdsiloai/vllm:vllm-openai-rocm-glm5.3-flash-mi325-08092026 -c "<cmd>"
```

**1. Tuning, all fifteen buckets** — produced the kept rows:

```bash
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py \
  -i aiter/configs/untuned_fmoe_glm53flash.csv \
  -o aiter/configs/tuned_fmoe_glm53flash_gfx942.csv \
  --mp 8 --timeout 180 --compare --update_improved
```

`--compare` benchmarks the production operator before and after tuning;
`--update_improved` then writes only shapes that improved by at least
`--min_improvement_pct` (default 3%). That gate is what rejected tokens
64/128/256/512.

**2. Verification** — re-runs the production operator against the tuned CSV:

```bash
AITER_CONFIG_FMOE=/aiter/aiter/configs/tuned_fmoe_glm53flash_gfx942.csv \
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py \
  -i aiter/configs/untuned_fmoe_glm53flash.csv \
  --run_config aiter/configs/tuned_fmoe_glm53flash_gfx942.csv --mp 8
```

**3. Tokens 1–32, default vs tuned** — the measurement that rejected them. Two
arms of `--run_config` differing only in whether a CSV is supplied, over a
six-row input (tokens 1–32):

```bash
# arm A - default kernels, no CSV
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py \
  -i untuned_small.csv --run_config --mp 8

# arm B - tuned kernels
python3 csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py \
  -i untuned_small.csv --run_config tuned_small.csv --mp 8
```

Run twice; the two runs agree within 1%.

## Results

End-to-end operator time from `--compare`, tuned against the default kernel:

| token | default (µs) | tuned (µs) | speedup |
|------:|-------------:|-----------:|--------:|
|  1024 |       542.18 |     524.53 |   1.03x |
|  2048 |      1002.53 |     670.24 |   1.50x |
|  4096 |      1864.42 |     868.68 |   2.15x |
|  8192 |      2812.08 |    1524.90 |   1.84x |
| 16384 |      4336.83 |    2764.15 |   1.57x |

Time-weighted speedup across the five is **1.66x** (10.56 ms -> 6.35 ms summed), median per-shape 1.57x. The gain concentrates at large token counts, so it matters most for prefill and large batched decode.

All five rows select `block_m` 64 and a 64x256 stage-1 kernel: `fmoe_bf16_blockscaleFp8_g1u1_vs_ps_silu_64x256` at tokens 1024, 4096, 8192 and 16384, and `fmoe_bf16_blockscaleFp8_g1u1_vs_silu_64x256` at token 2048. Stage 2 is unused; every row runs single-stage.

## Why only five of fifteen buckets

The other ten are deliberately absent, because the default kernel is equal or faster there and falling back is the better choice.

**Tokens 64, 128, 256, 512** were rejected by the `--update_improved` gate on two separate tuning runs, at -0.57%/-0.20%, -9.75%/-5.26%, +1.23%/-0.07% and -0.80%/+0.33%.

**Tokens 1 through 32** were rejected after a dedicated measurement. The tuner had marked them `no_baseline` and admitted them on the post-run alone, because the pre-run benchmark returned no timing for them. Benchmarking the default path directly (`--run_config` with no CSV) against the tuned kernels shows the tuned candidates are slower at five of the six:

| token | default (µs) | tuned (µs) | speedup |
|------:|-------------:|-----------:|--------:|
|     1 |        38.70 |      58.02 |   0.67x |
|     2 |        51.55 |      61.83 |   0.83x |
|     4 |        73.88 |      74.33 |   0.99x |
|     8 |       113.85 |     111.26 |   1.02x |
|    16 |       173.23 |     179.94 |   0.96x |
|    32 |       255.15 |     295.34 |   0.87x |

Reproduced over two runs agreeing within 1%. Only token 8 improves at all, by 2.3%, under the 3% threshold. Shipping these would have regressed the small-batch decode path by up to 1.5x.

## Verification

Keys are disjoint from the 35 `tuned_fmoe` tables **in this repo**, so
`get_config_file()` merges this file upstream with no duplicate-shape conflict.
Verified by loading `AITER_CONFIG_FMOE_FILE` with this file in `model_configs/`
and confirming all five rows survive the merge. The untuned file is excluded by
that glob, which skips any name containing `untuned`.

**That guarantee does not extend past this repo, and it is worth stating.** The
glob at `jit/core.py:439-444` accepts any `*tuned_fmoe*.csv` in
`model_configs/`, so a downstream image shipping its own table for this shape
is merged alongside this one. Our ROCm/vLLM nightly does exactly that
(`tuned_fmoe_glm53flash_gfx942.csv`), and the two agree on all 14 shape keys at
all five tokens.

The dedup that would normally protect against this does **not** fire, because
`_tag` is part of the dedup key: a file without the column is filled with `0`
while this one carries `""`, the keys therefore differ, and both rows survive
for the same shape with nothing deciding which the lookup returns. Confirmed by
running the merge over the two tables — 10 rows, 5 shapes, 0 duplicates
detected. It is not a "faster row wins" tie-break; there is no tie-break.

So the rule for this shape is **one table, not two**: whoever adds the second
must retire the first. Flagging it here because the next person to hit it will
have no way to tell from a server log.


zufayu:

**Adds two gfx942 data files for the GLM-5.3-Flash routed fused-MoE shape (model_dim 4096, inter_dim 512, 288 experts, topk 8, fp8 e4m3fnuz per-1x128 blockscale): a five-row tuned table that selects single-stage 64x256 asm kernels at token buckets 1024-16384, plus the fifteen-bucket tuner-input ladder it was tuned from.**

Validation (deterministic): NOT RUN — data-only config PR: triage found no test target (target=null; the diff adds only a tuned-config table and a tuner-input CSV under aiter/configs/model_configs/), so there is nothing a test target could have covered
Perf (advisory): NOT RUN — no benchmark target ships with a data-only CSV PR; the PR's own default-vs-tuned µs tables (20 token buckets, two runs agreeing within 1%) are author-measured on one box and not independently reproduced

⚠️ [verified] The tuning provenance behind aiter/configs/model_configs/a8w8_blockscale_tuned_fmoe_glm5_3_flash.csv cannot be reconstructed: the PR names the tool flags (--compare, --update_improved, --run_config) and TP4/gfx942, but ships no exact command lines, no ROCm version, no GPU model, and no aiter commit for the runs that produced the five kept rows and the fifteen rejections. Anyone re-running the tuner later cannot tell whether a changed number is a real kernel regression or a different environment. **Author must** add the exact tune/compare commands and environment (ROCm version, GPU, driver, aiter commit) to the PR or a checked-in tuning record.

jin-amd:
> 
> **Adds two gfx942 data files for the GLM-5.3-Flash routed fused-MoE shape (model_dim 4096, inter_dim 512, 288 experts, topk 8, fp8 e4m3fnuz per-1x128 blockscale): a five-row tuned table that selects single-stage 64x256 asm kernels at token buckets 1024-16384, plus the fifteen-bucket tuner-input ladder it was tuned from.**
> 
> Review (advisory): ⚠️ NEEDS WORK Validation (deterministic): NOT RUN — data-only config PR: triage found no test target (target=null; the diff adds only a tuned-config table and a tuner-input CSV under aiter/configs/model_configs/), so there is nothing a test target could have covered Perf (advisory): NOT RUN — no benchmark target ships with a data-only CSV PR; the PR's own default-vs-tuned µs tables (20 token buckets, two runs agreeing within 1%) are author-measured on one box and not independently reproduced
> 
> ⚠️ [verified] The tuning provenance behind aiter/configs/model_configs/a8w8_blockscale_tuned_fmoe_glm5_3_flash.csv cannot be reconstructed: the PR names the tool flags (--compare, --update_improved, --run_config) and TP4/gfx942, but ships no exact command lines, no ROCm version, no GPU model, and no aiter commit for the runs that produced the five kept rows and the fifteen rejections. Anyone re-running the tuner later cannot tell whether a changed number is a real kernel regression or a different environment. **Author must** add the exact tune/compare commands and environment (ROCm version, GPU, driver, aiter commit) to the PR or a checked-in tuning record.

Thanks @zufayu  — provenance was the same ask in each, so it's now in the description: new **Environment** and **Commands** sections (exact tuner invocations, container, GPU/driver/ROCm, and the aiter commit for each run), plus a rewritten **Verification**.

**On "not independently reproduced"** — now confirmed by a different method. A paired vLLM serving sweep rather than the tuner: 131k in / 1k out, 7 concurrencies, 20 requests each, TP4, one image holding both tables with `AITER_CONFIG_FMOE` selecting per arm, control first — and the control arm reproduces the shipping image's kernel selection exactly. Removing the tokens 1–32 rows gives **+1.17% output throughput winning 7/7 concurrencies, +3.58% at concurrency 2, +1.31% TPOT**, GSM8K 0.9727 in both arms. On "one box": repeat runs of a fixed config here agree to **±0.48%** per row over eight repeats, so the low-concurrency points are clear of variance. The gain also sits exactly where the dropped buckets were being selected and decays to ~0 above them — the shape the µs tables predict.

**A correction to my own description.** I claimed keys are disjoint from all 35 existing `tuned_fmoe` tables. True in this repo — but the glob takes any `*tuned_fmoe*.csv`, and our downstream nightly ships `tuned_fmoe_glm53flash_gfx942.csv` for this same shape, agreeing on all 14 shape keys at all five tokens. I expected the `us` tie-break to handle that; it doesn't. `_tag` is part of the dedup key, so a file without the column fills to `0` while this one carries `""` — **no duplicate is detected and both rows survive**, with nothing deciding which the lookup returns (verified over the two tables: 10 rows, 5 shapes, 0 duplicates). So the rule for this shape is one table, not two, and I'll retire the downstream one when this lands.

**On the `NOT RUN` validation stage** — agreed that a data-only CSV gives a test target nothing to exercise. If it would be welcome, I'd follow up with a schema check over `model_configs/`: valid columns against the untuned sibling, no duplicate shape keys within a file, and no cross-file collisions on the keys the merge actually uses. It would have caught the above automatically.

PR #5501, opened 2026-09-14
[Triton/Gluon] Fix gfx1250 Gluon MLA cache launch args

amd-ruitang3:
The gfx1250 fused MLA KV-cache wrapper passed an extra positional argument that is not present in the Gluon kernel signature.

This shifted launch arguments and caused `k_scale_ptr` to be bound twice. Remove the unused extra argument so the wrapper matches the kernel signature.

Validation:
- `python -m py_compile aiter/ops/triton/fusions/fused_kv_cache.py`
- `git diff --check`


k50112113 (approved):
Thanks for catching this

PR #5507, opened 2026-09-14
[FlyDSL] [gfx1250] Quantize the MoE a1 scale and shuffle.

lalala-sh:
Measured on gfx1250, a4w4, 16384 tokens, 7168/768, E=384, topk=6, no bias (device time summed over the layer kernels):

  scattered interleaved write   1014 us   quant 143.1
  compact + rebuild, 1 launch    953 us   quant 75.1

The two-launch form is faster on device and the one-launch form on wall time: merging the passes saves two launches but the persistent grid costs the GEMMs after it about 30 us. This ships the single launch; logits_diff is 3.5076e-06 on all three, unchanged from the scattered write.

## Submission Checklist


zufayu:

**On the gfx1250 grouped MoE, the a1 e8m0 scale is now written as one compact row per source token and a second kernel rebuilds the 16-row-interleaved layout gemm1 reads (replacing the row-major-ascale path), while the gather-reduce epilogue switches to per-route 64-bit-based descriptors so grouped outputs past 4 GiB still address correctly.**

Validation (deterministic): NOT RUN — the PR changes runtime code but ships no test target (triage: required, target=null); auto-validation disabled (REVIEW_AUTO_VALIDATE=0) and no gfx1250 GPU on this box, so the new kernels were checked by static layout analysis and an index-math simulation only.
Perf (advisory): NOT RUN — no gfx1250 GPU here and the PR ships no benchmark entry point as a test target; the claimed 1014→953 us layer / 143.1→75.1 us quant (gfx1250, a4w4, 16384 tokens, 7168/768, E=384, topk=6) is unverified.

🔴 [verified] The new compact-path gate `_use_fused_quant_preshuffle` (grouped_moe_gfx1250.py:301) checks token count, topk and geometry but not the pre-existing `AITER_FLYDSL_TOKEN_MULTIDEST_QUANT` kill-switch that the actual dispatch inside `flydsl_moe_fused_quant_preshuffle` checks (moe_kernels.py:3229). With that env var set to 0 on a ≥1024-token (2048 when wmma_rep=1) non-EP gfx1250 a4w4 batch, the caller sizes `out_scale` as a compact (token_num, model_dim//32) buffer (grouped_moe_gfx1250.py:698) but the dispatch silently falls back to the route-indexed kernel, which writes the full 16-row-interleaved scale for all token_num*topk routed rows through an unbounded byte view (moe_fused_route_quant_scatter.py:1691) — contiguous_m ≈ topk*token_num + E*tile_m, so at topk=6 the stores land ~6x past the allocation: silent heap corruption with no crash, on exactly the rollback path an operator reaches for when the quant path misbehaves. **Author must** make the gate and the dispatch agree — check the env var in `_use_fused_quant_preshuffle`, or assert `fused_preshuffle_out is None` on every branch that did not take the compact path.

⚠️ [verified] The compact-quant + scale-rebuild pair ships with no collectible test and no CI path that can execute it: no test file is touched, the pk8 requirement keeps it off the gfx942/gfx950 default CI, and the PR carries no ci:* label — the only correctness evidence is a self-reported `logits_diff 3.5076e-06` with no harness named, and the gate additionally accepts quant_mode="fp8" (a8w4 activations) that the body's a4w4-only measurement never covered. **Author must** add a test pinning the compact+rebuild scale buffer against the scattered path (or a torch reference) and name the harness that produced the logits_diff; **Reviewer should ask** which ci:* label actually runs gfx1250 hardware.

⚠️ [verified] The PR body says "This ships the single launch" and labels the 953 us row "compact + rebuild, 1 launch", but the shipped code launches two kernels (compact quant, then scale rebuild — moe_kernels.py:3245-3270), and the rebuild docstring explicitly rules out fusing them ("it cannot start until every token is quantized", moe_fused_route_quant_scatter.py:2470). The 953 us / 75.1 us numbers therefore cannot be mapped unambiguously to the shipped implementation. **Author must** reconcile the body's launch-count wording with the code and state which form the measured numbers describe.

⚠️ [verified] The new "quant_a1" bench hook (grouped_moe_gfx1250.py:857) re-launches `flydsl_moe_fused_quant_preshuffle` without `row_to_token`/`fused_preshuffle_out`, so when the compact path is active the hook times the old scattered kernel instead — a tuning run would report ~143 us for a production path that costs ~75 us, with nothing flagging the mismatch. **Author must** thread the compact buffers through the bench partial (or add a separate entry) so the label matches the kernel actually measured.

⚠️ [verified] `build_moe_token_multidest_compact_quant_module` and `build_moe_scale_rebuild_module` (moe_fused_route_quant_scatter.py:2176, 2459) are newly public in a module `aiter/aot/flydsl/grouped_moe.py` imports, but no AOT hunk teaches the pre-compiler the pair, so every (feat_dim, wmma_rep, topk) combination JIT-compiles on first invocation — mitigated by the sibling token-multidest builder already sitting outside AOT's enumeration, but the ≥1024-token default path now routes through the unenumerated pair. **Reviewer should ask** whether AOT should pre-compile the compact pair now that the large-batch default depends on it.

junhaha666 (approved):

PR #5513, opened 2026-09-14
[CI] Add Some vLLM lm_eval Nightly Tests

micah-wil:
This adds a nightly job which pulls the latest `vllm/vllm-openai-rocm:nightly` and installs AITER ToT main (with `PREBUILD_KERNELS=0` to spot-check some models' `lm_eval` scores. I think this could help as an additional signal to at least catch model breakages. This does not include any e2e performance benchmarking, which will eventually be covered by an additional nightly job in vllm's `perf-eval`.


zufayu:
## ROCm/aiter PR #5513 — [CI] Add Some vLLM lm_eval Nightly Tests

**Adds a daily scheduled job that serves 10 model configs from a vLLM ROCm image with AITER ToT installed over the bundled copy, then fails the run when the lm_eval gsm8k score drops below a threshold.**

⚠️ [verified] .github/workflows/vllm-nightly-integration.yaml:16 sets `BASE_IMAGE: vllm/vllm-openai-rocm:latest`, but the PR body says the job "pulls the latest `vllm/vllm-openai-rocm:nightly`" — on Docker Hub `latest` is byte-identical to release v0.29.0 (digest sha256:e5e47f6a, cut 2026-09-09) while the floating `nightly` tag is rebuilt per commit (updated 2026-09-17). Testing a frozen release means a vLLM-main regression only reaches this job when the next release moves `latest`, so the nightly signal the PR exists for arrives weeks late and a red run cannot be bisected against vLLM commits. **Author must** point BASE_IMAGE at `:nightly`, or correct the PR body if `:latest` was intentional.

⚠️ [verified] All 10 matrix entries gate on the same THRESHOLD=0.90 (.github/scripts/vllm_nightly_accuracy.sh:8 default; the workflow never passes `-e THRESHOLD` and the matrix has no threshold field), and no reference score is recorded for any model — the repo's own convention (kimi-downstream.yaml) documents a reference run (gsm8k 0.9409, aiter ef114b0d4, 2026-05-30) and sets its threshold below it (0.92). A model whose true gsm8k 5-shot flexible-extract sits near or below 0.90 turns the nightly permanently red until the signal is ignored, a loose 0.90 for stronger models misses real regressions, and when the job does fire there is no baseline to judge the drop against. **Author must** wire a per-model threshold into the matrix and record the reference score each threshold came from.

⚠️ [verified] Nothing asserts AITER is actually engaged by the served process: vLLM's integration degrades silently (vllm/_aiter_ops.py `if_aiter_supported` returns None and falls back to non-aiter paths when the toggle is unread or aiter is unsupported), and this env family demonstrably drifts (VLLM_ROCM_USE_AITER_MOE_SITUV2_A8W4 is already a deprecated alias on vLLM main). If the nightly image stops honoring VLLM_ROCM_USE_AITER (.github/workflows/vllm-nightly-integration.yaml:135), all 10 jobs serve through vanilla vLLM and stay green while testing zero AITER code; the run also never records the vLLM version or image digest (the sibling kimi_vllm_accuracy.sh prints both vllm.__version__ and the aiter pip version), so even a red run against a floating tag is not reproducible. **Author must** add a post-install `python -c "import aiter"` smoke check in the serve environment and echo vllm.__version__ plus the pulled image digest into the log.

📝 [verified] The Kimi-K3 entry sets VLLM_ROCM_USE_AITER_MOE_SITUV2_A8W4=1 (.github/workflows/vllm-nightly-integration.yaml:82), which vLLM main documents as a deprecated alias for VLLM_ROCM_USE_AITER_MOE_SITUV2 that "does not select a8w4 kernels" (vllm/envs.py:1259-1266) — it still enables the a4w4 SiTUv2 path today, but new CI wired to a deprecated name breaks silently the day the alias is removed. **Author must** switch to VLLM_ROCM_USE_AITER_MOE_SITUV2.

📝 [verified] The install step runs `pip install "lm-eval[api]"` unpinned in a fresh container every night (.github/workflows/vllm-nightly-integration.yaml:154), so each run scores with whatever lm-eval PyPI served that morning — an lm-eval release that changes gsm8k prompting or flexible-extract scoring moves all 10 scores overnight and shows up as a red nightly misattributed to AITER/vLLM. **Author must** pin the lm-eval version so harness drift cannot masquerade as a regression.

PR #5519, opened 2026-09-15
[FlyDSL] fix: select FMoE GEMM2 A addressing by stored buffer size

Bernard-Liu:
<html>
<body>
<html><head></head><body><h2>Motivation</h2><p>GEMM2 buffer addressing cannot represent an A buffer of 4 GiB or larger, which can produce incorrect results for large-token MoE workloads.</p><p>Select the addressing mode from the actual A tensor size so that large inputs use 64-bit global addressing while smaller inputs retain the existing buffer-load path.</p><h2>Technical Details</h2><ul><li>Add <code dir="ltr">requires_flydsl_stage2_global_a()</code> and share it between runtime and AOT. The decision uses <code dir="ltr">numel() * element_size() &gt;= (1 &lt;&lt; 32)</code>, correctly accounting for packed FP4 storage, declared padding, and contiguous views.</li><li>Propagate the internal <code dir="ltr">use_global_a</code> compiler parameter through FMoE and FHMoE. Direct compiler callers retain the safe default of <code dir="ltr">True</code>.</li><li>Select sync buffer/global loads and async buffer-to-LDS/global-to-LDS operations at compile time. Construct the A descriptor only for the buffer variant.</li><li>Distinguish both variants in compilation caches and generated symbols through <code dir="ltr">_abuffer</code> and <code dir="ltr">_aglobal</code>.</li><li>Use FakeTensor metadata to compile the required AOT variant without allocating a large GPU buffer. Keep AOT consistent with runtime for FP8 <code dir="ltr">persist_m</code> adjustment and automatic reduce selection for large outputs.</li><li>Preserve public runtime APIs, GPU launch arguments, CSV schemas, and the existing Graph warmup/capture/replay workflow.</li></ul><h2>Test Plan</h2><p>Validation was performed on <code dir="ltr">gfx950</code> with FlyDSL <code dir="ltr">0.3.2</code>.</p><ul><li>Check FP8 and packed FP4 inputs below, at, and above the 4 GiB boundary where representable.</li><li>Exercise ordinary MoE/FHMoE, sync/async loads, atomic/reduce output, and small contiguous views whose backing-storage offsets exceed 4 GiB.</li><li>Validate eager execution, Graph replay, and replay after updating input data and routing.</li><li>Check host metadata decisions and runtime/AOT parameter propagation, then compile small and exact-4-GiB FP8/FP4 AOT configurations.</li><li>Recompile 11 configurations with both addressing modes before and after rebasing onto current main; compare complete ISA, <code dir="ltr">.text</code> bytes, and code-object resources.</li><li>Run full two-stage FMoE with a local tuned CSV using token=262144, H=7168, I=3072, E=24, top-k=6, <code dir="ltr">swiglu_limit=10</code>, and an FP32 reference. This configuration has a 4.5 GiB FP8 A buffer.</li><li>Run Black, Ruff, Python compilation, and <code dir="ltr">git diff --check</code> on the changed files.</li></ul><h2>Test Result</h2><div><div><div>
Validation | Result
-- | --
Stage2 correctness and Graph checks | 64/64 passed on the final head; the same 64 passed before rebase
Compilation and ISA comparison | 44 builds succeeded; all 22 corresponding ISA, .text, and resource comparisons matched
Global variant descriptor checks | All 11 global variants omitted the A descriptor
Host metadata and parameter propagation | 24/24 passed
Actual AOT compilation | 4/4 passed without initializing the Torch GPU runtime
Changed-file formatting and static checks | Passed

</div></div></div><p>The full-FMoE run selected <code dir="ltr">_aglobal</code> and reported no NaNs. The final-head <code dir="ltr">logits_diff</code> was <code dir="ltr">7.9042e-6</code>, satisfying the agreed <code dir="ltr">&lt;= 0.01</code> criterion. Elementwise mismatch was 13.3% on both revisions; this result does not claim elementwise equality.</p><p>Standards and Spec reviews found no outstanding issues. The full CSV performance matrix was not rerun after rebase.</p></body></html>
</body>
</html>


op_tests/test_flydsl_moe_large_buffer.py
```diff
@@ -0,0 +1,362 @@
+# SPDX-License-Identifier: MIT
```
coderfeli: dont add such test

binding7012:
LGTM

coderfeli (approved):

zufayu:

**Selects the FlyDSL MoE stage2 (GEMM2) A-activation addressing mode at compile time from the A tensor's stored size — A at or above 4 GiB builds an `_aglobal` variant that addresses A through 64-bit global pointers (sync `ptr_load` plus async `global_load_lds`), smaller A keeps the existing 32-bit buffer-descriptor path — and threads the same `use_global_a` decision through the runtime, FHMoE, and the AOT precompiler.**

Validation (deterministic): NOT RUN — the PR changes runtime code but ships no test target (auto-validation disabled, REVIEW_AUTO_VALIDATE=0)
Perf (advisory): NOT RUN — no perf target for the same reason there is no test target: the PR changes runtime code but ships no test target

⚠️ [verified] The final head ships zero tests for the `_aglobal` variants this diff adds to aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage_common.py (`load_x` / `dma_x_tile_to_lds`) and for the new `requires_flydsl_stage2_global_a` in aiter/ops/flydsl/moe_kernels.py: the boundary test that existed earlier in this PR (`op_tests/test_flydsl_moe_large_buffer.py`, covered by the first bot review on commit b7cbdc9 as 1 of 7 files) is absent from the final 6-file diff, and existing CI (`op_tests/flydsl_tests/test_flydsl_moe.py`, `op_tests/test_fhmoe.py`) only reaches the < 4 GiB `_abuffer` variant. An A buffer at or above 2^32 bytes (about 233k tokens at topk=6, I=3072, fp8 — reachable in long-context prefill) is exactly the regime that produced silently wrong GEMM2 results, so a future refactor of `load_x`/`dma_x_tile_to_lds` could break the `_aglobal` path with no CI signal, and the PR body's 64/64 and `logits_diff 7.9042e-6` numbers have no committed reproducer. **Author must** restore a committed regression test for the boundary — a >= 4 GiB A run where the GPU can represent it, or at minimum host-side checks that `requires_flydsl_stage2_global_a` and the `use_global_a` propagation select `_aglobal`/`_abuffer` on both sides of 2^32 — or state on the PR why it cannot be committed.

⚠️ [verified] Stage1 has the same unfixed bug this PR fixes for stage2: `compile_mixed_moe_gemm1_common` still builds `x_rsrc = ptr_rsrc(arg_x, x_nbytes_i32)` unconditionally (mixed_moe_gemm_2stage_common.py:602, no global variant anywhere in the stage1 emitter) and gathers A through `buffer_copy_gmem16_dwordx4` at :761. Stage1's A is the [tokens, model_dim] input, which crosses 4 GiB at about 599k fp8 tokens (H=7168; about 1.2M for fp4), so the "large-token MoE workloads" named in the motivation still get silently wrong stage1 output above that — this PR makes the full FMoE correct only for tokens between roughly 233k and 599k. **Author must** extend the same stored-size addressing selection to stage1's A, or state the supported stage1 token bound on the PR.

⚠️ [verified] The bf16-A stage2 path is not covered: `compile_flydsl_moe_stage2` (moe_kernels.py:759) dispatches a_dtype=bf16 to `compile_gemm2_a16w4_port`, which stages A through a 32-bit buffer view (`moe_2stage_a16wmix/utils.py` `make_a_loader`, `_global_i32_buffer_view` plus BufferCopy atoms) with no `use_global_a` equivalent, even though this PR added the parameter to the same function's fp4/fp8 arm. bf16 A crosses 4 GiB at about 116k tokens (topk=6, I=3072) — half the fp8 threshold this PR fixes — so a W4A16 model at e.g. 150k tokens still computes wrong GEMM2 results after this PR. **Author must** confirm whether the a16w-mix path needs the same stored-size selection, or state that its token range is bounded below the limit.

📝 [verified] GEMM2's weight buffer resource (`w_rsrc = ptr_rsrc(arg_w, w_nbytes)`, mixed_moe_gemm_2stage_common.py:3511) is an unconditional 32-bit descriptor: W2 is 3.77 GB at DSv3 scale (E=257, H=7168, I=2048) and crosses the same 4 GiB limit at E >= 293, while stage1's W already carries a per-expert 64-bit re-base (:725) for exactly this reason. No shipping model crosses it today, but a larger-expert-count model would hit silently wrong W loads with no size guard. **Reviewer should ask** whether a W-side guard or per-expert re-base is planned for GEMM2.

📝 [inferred] No base-vs-head codegen comparison exists for the small-buffer path: the PR's 11-config ISA comparison ran before and after rebasing the PR (both sides carry the change), not head against the merge target, and the `_abuffer` branch of `load_x` in aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage_common.py — textually the old emitter under a `const_expr` else — is expected but not demonstrated to produce identical ISA. If the restructure (moved `x_rows`/`x_elem`/`vec16_elems` definitions, new branch nesting) shifted small-buffer codegen, every existing production shape would pay it and no measurement would catch it. **Reviewer should ask** for one `_abuffer` config's ISA and .text diff against the merge target to confirm zero codegen drift below 4 GiB.

PR #5523, opened 2026-09-15
[Flydsl] stop the A16W16 pruner from evicting every HTI config

huizzhan:
GemmConfigPruner scores every policy against one tile-efficiency threshold, and `keep_ratio` climbs toward 1 as M grows: at M=27556 it only keeps tiles that waste under 0.44% on padding. Half-tile interleaving is only expressible at block_n=256, so on N=384 -- which pads to 512 -- the gate drops the entire HTI catalog and tuning falls back to a full-tile policy. HTI is worth more than the padding it costs, so the shape loses ~30%.

Cap keep_ratio so a tile may pad up to 30% and still be considered. On gfx950 across the 17 GEMM shapes behind the Qwen-Image VAE convolutions: 6 shapes gain (up to 1.30x on 27556x384x3456), none regress, and the candidate count grows only 112 -> 144 on the worst-hit shape.

144 is a small slice of what the kernel can express -- 2968 of the 677376 configs in the raw grid survive validation for that shape -- so the gate is worth checking rather than trusting. Timing all 2968 finds nothing faster than the candidate it now admits: the winner is the same 256x256x64 stage-2 HTI tile, and the runner-up is 22% behind it.

The tuned config holds the fastest backend per shape, which for these shapes is hipBLASLt throughout.

## Submission Checklist


zufayu:

**Caps the A16W16 GEMM tuning pruner's keep_ratio at 0.7 so half-tile-interleaved block_n=256 candidates survive large-M pruning, and ships the resulting Qwen-Image VAE tuned/untuned config CSVs plus a three-way conv/GEMM/hipBLASLt benchmark.**

Validation (deterministic): NOT RUN — triage required a target but the PR ships none; the changed code (aiter/ops/flydsl/gemm_a16w16_policy.py) is tuner-time scoring and the two CSVs are a tuned-config table plus tuner input, nothing a serving-path test target could cover — the absence is not a defect.
Perf (advisory): NOT RUN — the shipped harness (op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/bench_three_way.py) requires gfx950; this review box is gfx942 and the a16w16 flydsl path is gfx950-only (environment gap, not a PR defect). The pruner-logic claims were verified GPU-free: 677376 raw / 2968 valid / 112→144 candidates on 27556×384×3456 all reproduce exactly.

🔴 [verified] aiter/configs/model_configs/qwenimage_vae_bf16_tuned_gemm.csv commits 17 libtype=hipblaslt rows with solidx 439567–441391 — the first hipblaslt data ever committed under aiter/configs (all 2564 existing gfx950 rows are asm/flydsl/opus/skinny/torch/triton), and aiter/jit/core.py:490-503 auto-merges every model_configs/*bf16_tuned_gemm*.csv into the runtime dispatch table, where a matching shape (e.g. 27556×384×3456 bf16 no-bias on gfx950/256 CU) dispatches through hipb_gemm into getAlgosFromIndex with the raw CSV index and no validation — an index space internal to the hipBLASLt build, so on a docker with a different hipBLASLt the same solidx can select a different algorithm or fail. **Author must** drop the hipblaslt rows from the committed CSV (the tuner keeps hipblaslt opt-in for exactly this reason) or gate them per hipBLASLt build.

⚠️ [verified] The comment at aiter/ops/flydsl/gemm_a16w16_policy.py:90-94 (and the PR body) claims block_n=256 is "the only width the half-tile-interleaved policy can express" and the old gate dropped "the entire HTI catalog", but the kernel's own validator accepts HTI at block_n 64/128/256 — for 27556×384×3456 there are 34 valid HTI configs and the old gate kept 8 of them; what the 0.7 cap actually does is re-admit every config with up to 30% padding waste, growing the candidate count up to 4.6× on shapes with no HTI at all (59→269 on 1048576×96×1728, 24→96 on 65536×192×1728), not the "only 112→144" the description highlights. **Author must** correct the comment and state the worst cross-shape candidate growth, since it sets retune wall time.

⚠️ [verified] op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/bench_three_way.py:223-224 guards a missing arm (fg for t_fly, rg for the ratio) but the print at 226-227 formats {t_hip:9.2f} and {t_hip / t_conv:8.3f}x unguarded — one shape where hipblaslt yields no solution or the conv profiler key never matches kills the whole sweep at the print, before rows.append and before three_way.json is written. **Author must** guard t_hip and t_conv the same way fg/rg already are.

📝 [inferred] The headline timing numbers (up to 1.30× on 27556×384×3456, "the shape loses ~30%", runner-up 22% behind) name no baseline, and the one committed artifact (op_tests/op_benchmarks/flydsl/qwenimage_vae_conv/figures/three_way_vs_hipblaslt.png) compares conv/flydsl/hipblaslt, not pruner old vs new. **Reviewer should ask** for the before/after tuning log on both sides of the one-line change — the count claims reproduce exactly, the timing claims are the unverified half.

📝 [verified] The PR is stale against the merge target: the binary PNG hunk does not apply and main has moved 24 commits, though all text hunks apply cleanly and gemm_a16w16_policy.py itself has not moved on main. **Author must** rebase so CI runs against the current tree.

PR #5525, opened 2026-09-15
[FlyDSL] gfx950 FP8 paged DSA indexer score-plus-local-TopK

samremes:
## Summary
- Add a gfx950 FlyDSL operator that scores FP8 paged DSA indexer K (`H=32`, `D=128`) and keeps exact local top-k per history split, then merges the split-local bags with a split-aware decode TopK. The public entry points are `flydsl_fp8_paged_mqa_local_topk` and `flydsl_fp8_paged_mqa_topk`.
- This is the decode `next_n=1` path only (`q` is `[B,1,32,128]`). It does not include mixed per-request speculated widths, MTP group-max / harvest, or persistent/sorted Stage B.
- Motivation: long decode still pays `O(L)` indexer scoring, but it should not allocate or write the full `[rows, max_model_len]` FP32 logits tensor (128 MiB at 32 rows × 1M; 512 MiB at 128 rows).
- K pages must be `shuffle_weight(..., layout=(16,16))` (packed or split). That matches the vLLM DSA cache for `block_size > 1`.

## Merge path

Stage A holds every retained candidate's key in registers, so it also fills the merge's pass-0 histogram on the way out. The merge then starts at pass 1. Supporting this needed a few hooks on `top_k_per_row_decode`: `split_width` and `use_split_counts` bind one chunk to one split's live prefix, `payload_indices` carries positions through the gather, `combine_histograms` folds the splits into a single histogram and zeroes each bin as the reduce consumes it (which is what leaves the workspace clean for the next call), and `precomputed_first_pass` skips the pass Stage A already did.

The existing public decode TopK path is unchanged: it builds with these hooks off and binds the two extra tensor arguments to operands it already holds.

## Serving / CUDA-graph usage

The extra intermediate is the merge workspace, not the candidate scores. Allocate it **once** outside capture; pass the same pair on every decode step. Do **not** `zero_()` between successful A→B calls. Stage A atomics into the histogram; the merge consumes it and writes every bin back to 0.

```python
from aiter.ops.flydsl.split_topk_merge import alloc_split_topk_merge_workspace

# rows = q.shape[0]
# hist: [rows, 1, 2048] int32    state: [rows, 6] int32
workspace = alloc_split_topk_merge_workspace(device, rows)  # zeros once

values, positions = flydsl_fp8_paged_mqa_topk(
    q, kv, None, weights, lens, tables, k=2048, workspace=workspace,
)
```

Two-kernel form (caller also owns the `[rows, S, k]` bags):

```python
bags = flydsl_fp8_paged_mqa_local_topk(..., workspace=workspace)
values, positions = split_topk_merge(
    *bags, k=k, precomputed_first_pass=True, workspace=workspace,
    out_scores=out_scores, out_positions=out_positions,
)
```

`num_splits == 1` has no merge and rejects `workspace`. If Stage A runs and merge does not, the histogram is dirty. Tests/benches may omit `workspace`; the wrapper then uses a process cache that zeros on first miss only.

Auto `num_splits` can pick 128; merge currently accepts at most 64 chunks. The numbers below clamp `S` to 64. Default `num_splits=None` raises on 1-row `L >= 256k`.

## Measured vs Gluon logits + decode TopK

Idle MI355X / gfx950, packed page-64 shuffled KV, `k=2048`, `next_n=1`, independent KV pages per request. CUDA-event medians, 10 warmup / 31 samples.

vLLM-shaped logits: always `[rows, max_model_len=1048576]`. Live `context_lens = L`. KV covers `L` only. Gluon and FlyDSL `top_k_per_row_decode` follow `L` (they do not scan the unused 1M tail); the allocation still fixes logits stride and TopK compile width at 1M. Compact never sees that tensor. Merge workspace and TopK outputs are persistent across timed calls.

### Compact speedup (Gluon+TopK ÷ compact). Above 1.00 compact is faster.

| L \ rows | 1 | 4 | 8 | 16 | 32 |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4k | 1.22 | 1.22 | 1.17 | 1.24 | 1.14 |
| 8k | 1.20 | 1.20 | 1.19 | 1.22 | 1.14 |
| 16k | 1.17 | 1.20 | 1.20 | 1.23 | 1.16 |
| 32k | 1.20 | 1.19 | 1.18 | 1.19 | 0.96 |
| 64k | 1.18 | 1.18 | 1.10 | 0.99 | 0.88 |
| 128k | 1.17 | 1.10 | 0.93 | 0.92 | 1.01 |
| 256k | 0.94 | 0.66 | 0.91 | 1.04 | 1.07 |
| 512k | 0.73 | 0.90 | 1.03 | 1.10 | 1.18 |
| 1M | 0.58 | 1.02 | 1.17 | 1.19 | 1.20 |

Geomean over these 45 cells: **1.07**. Compact wins 34/45.

### Raw device time, compact (µs)

| L \ rows | 1 | 4 | 8 | 16 | 32 |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4k | 58 | 60 | 62 | 62 | 62 |
| 8k | 58 | 61 | 61 | 62 | 61 |
| 16k | 60 | 60 | 61 | 62 | 63 |
| 32k | 59 | 61 | 61 | 64 | 75 |
| 64k | 60 | 62 | 65 | 76 | 108 |
| 128k | 60 | 67 | 79 | 109 | 171 |
| 256k | 76 | 113 | 113 | 172 | 267 |
| 512k | 101 | 118 | 175 | 268 | 447 |
| 1M | 140 | 188 | 269 | 447 | 836 |

### Raw device time, Gluon + decode TopK (µs)

| L \ rows | 1 | 4 | 8 | 16 | 32 |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4k | 70 | 74 | 72 | 77 | 71 |
| 8k | 70 | 73 | 72 | 75 | 70 |
| 16k | 70 | 73 | 73 | 76 | 73 |
| 32k | 70 | 72 | 72 | 76 | 72 |
| 64k | 71 | 73 | 72 | 75 | 95 |
| 128k | 71 | 73 | 74 | 100 | 172 |
| 256k | 71 | 75 | 103 | 178 | 286 |
| 512k | 74 | 107 | 180 | 295 | 526 |
| 1M | 82 | 193 | 316 | 531 | 1007 |

- Compact wins short/medium `L` at every batch (`L ≤ 32k`, usually `L ≤ 64k`) and large-batch megatoken (`8–32 × 512k–1M`; **32×1M is 836 vs 1007 µs**).
- Gluon+TopK wins skinny long decode: **1×256k–1M** (worst **1×1M: 140 vs 82**) and a mid-length band at 16–32 rows (`16×64–128k`, `32×32–64k`).
- One-row Gluon stays ~70–82 µs from 4k to 1M; compact grows with `L` after `S` caps at 64. That is the B=1 1M hole. Earlier “we win 1M” numbers were 8/16/32 rows.

PR 4221 paged logits was not re-run here. Prior matched e2e at `next_n=1` did not beat Gluon+TopK; 4221’s edge is K reuse at `next_n≥2`, and it cannot launch large independent 1M caches (signed-i32 numel).

## Test plan
- [x] `python3 op_tests/test_flydsl_fp8_paged_mqa_local_topk.py` on gfx950 (script-mode CI; edges + compact vs logits+TopK)
- [x] Caller-owned workspace stays all-zero after repeated A→B calls and CUDA-graph replay
- [x] Compact vs Gluon+TopK selected-score multisets agree within 3e-6 on the 1M-width grid
- [x] Public `flydsl_top_k_per_row_decode` still exact vs `torch.topk` at `(1, 1M, k=2048)`, `(8, 32K, k=512)`, `(4, 64K, k=128, next_n=2)`
- [ ] CI FlyDSL gfx950 job collects the new `op_tests` file


samremes:
Correctness follow-up (gfx950, independent KV, packed page-64, `k=2048`, `next_n=1`):

- The intermittent 1M failure was structural, not FP8 boundary noise. Stage A occasionally emitted duplicate split-local positions because cross-wave LDS scan subtotals crossed `s_barrier` before their writes were retired. The fix adds `s_waitcnt lgkmcnt(0)` before publication and forces a new compiled kernel identity.
- A separate deterministic failure affected row 31 at 32×1M: its packed cache crossed 4 GiB, while buffer instructions used a 32-bit byte offset from one flat descriptor. Paged loads now rebase the descriptor per physical page using 64-bit pointer arithmetic.
- Validation compares selected score multisets within the existing FP8 tolerance, while duplicate/count/range invariants remain exact.

Verification:
- 36 targeted tests pass.
- 0/1000 structural failures at each of 8×1M, 16×1M, 32×1M, and 32×512K.
- All 21 standard `run_perftest` sweep cells pass the FP32 value oracle.

Corrected device time at 1M: compact vs materialized logits + TopK is 311.4 vs 327.5 us (8 rows), 538.8 vs 569.8 us (16), and 1023.3 vs 1101.0 us (32). Compact does not win every shorter-length cell; the PR conclusions should use the corrected device-time sweep rather than the earlier wall-clock measurements.

samremes:
Performance correction (supersedes the per-page `GTensor.rebase_bytes()` implementation in my previous update):

The descriptor-rebase fix was correct but rebuilt a lane-divergent V# for every K and scale load, costing 10–25% at long lengths. Commit `3447a3e04` removes that path. K and scale now use typed global iterators with 64-bit element offsets directly; generated ISA lowers to `global_load_dwordx4` / `global_load_dword`, preserving >4 GiB correctness without per-load descriptor construction.

The pointer-load schedule also showed that the two-barrier cross-wave scan was still not sufficient (one fresh 32×1M run reproduced a duplicate). The final scan is three-phase: publish wave totals, wave 0 computes wave prefixes, all waves consume, then a trailing barrier protects reuse.

Correctness:
- 36/36 targeted tests
- full 32×1M FP32 value oracle passes
- 0/1000 structural failures at each of 8×1M, 16×1M, 32×1M, and 32×512K

Median of 7 `run_perftest` calls (µs), final vs descriptor-rebase vs knowingly incorrect pre-fix:
- 8×1M: 256.9 vs 318.7 vs 259.7
- 16×1M: 451.7 vs 540.8 vs 488.8
- 32×1M: 849.4 vs 1057.7 vs 946.4
- 32×512K: 450.7 vs 552.9 vs 489.5

So the correctness regression is removed: the final kernel is neutral or faster than the pre-fix kernel on the long shapes while retaining both structural and >4 GiB correctness. Static resources at k=2048: 128 VGPR, 98 SGPR, 60,480 B LDS, no spills/scratch.

zufayu:

**Adds a gfx950-only FlyDSL decode operator that scores FP8 paged DSA indexer K (H=32, D=128) and keeps an exact top-k per history split in an on-chip reservoir, then merges the split bags into the row's global top-k without ever materializing the [rows, L] fp32 logits tensor.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0) and this review box is gfx942 with no gfx950/MI355X; the single target op_tests/test_flydsl_fp8_paged_mqa_local_topk.py skips on every other arch. The PR reports 29 passed on gfx950; not independently reproduced here.
Perf (advisory): NOT RUN — no gfx950 GPU on this box (the target's sweep exits early on gfx942). The PR's own evidence is event medians (10 warmup / 31 samples, one idle MI355X) of head against the two incumbent arms (Gluon logits + decode TopK, and #4221 e2e), which is the meaningful comparison for a new operator but is not a base-vs-head pair and was not re-run here.

🔴 [verified] split_topk_merge's default configuration (precomputed_first_pass=False — used by merge_local_topk_candidates at fp8_paged_mqa_local_topk.py:480 and described as the normal mode in split_topk_merge.py's docstring) leaves output slots unwritten when a row's live candidates total fewer than k. The pass-0 histogram kernel clamps the selector's remaining-k to row_len (topk_per_row_decode.py:378), but row_ends is the full splits*k bag width (split_topk_merge.py:132-135, :205), so remaining-k is always k. At candidate_counts=[[32,32]] with k=128, splits=2 (a fresh 64-token request in a mixed-length batch), the 3-pass radix select never crosses remaining-k, the threshold stays 0, and the gather writes only [0,32) of each [rows,128] output — slots [32,128) keep torch.empty garbage (or stale out_scores content), and caller-supplied negative scores are dropped entirely. The e2e entry flydsl_fp8_paged_mqa_topk is safe (Stage A pre-inits remaining to min(valid_len, k) and the last reduce clears the tail), which is why the PR's 29 passing gfx950 tests never see it: no test runs the merge non-precomputed on a short row. Downstream attention reading k positions per row then gathers KV at garbage offsets — silent wrong output or crash. **Author must** clamp remaining-k to the live count (sum of candidate_counts) in the non-precomputed path — or clear the output tail there — and add a short-row regression for split_topk_merge(precomputed_first_pass=False) and merge_local_topk_candidates.

⚠️ [verified] out_scores/out_positions are accepted with no shape, dtype, device, or contiguity validation: split_topk_merge.py:197-204 binds them directly to the kernel output tensors and fp8_paged_mqa_local_topk.py:351-352, :451-452 pass them through, while the gather writes row_indices[out_pos] for every out_pos < k. An undersized buffer (out_positions=[rows,64] with k=128) writes 64 int32 past each row — silent memory corruption with no error; a non-contiguous or wrong-dtype buffer writes through wrong addresses. Every candidate input is validated; the two output buffers are not. **Author must** validate out_scores/out_positions like the candidate tensors (shape [rows,k], float32/int32, same device, contiguous).

📝 [verified] The kernel and the test oracle disagree for negative k_scales: the epilogue clamps scale to >=0 after the ReLU-weighted head sum (kernels/mqa_logits/fp8_paged_mqa_local_topk.py:527, "a negative or NaN scale must score 0") while run_torch applies scale before the ReLU (op_tests/test_flydsl_fp8_paged_mqa_local_topk.py:150-156), which is nonzero for negative scale x negative dot. No test generates negative scales (torch.rand is [0,1)) and serving scales are amax-derived (>=0), so the divergence is unreachable today. **Reviewer should ask** which semantic is the contract and assert scales >= 0 at the wrapper, or align the oracle.

📝 [verified] The float32-to-ordered-int32 key mapping now exists in two independent copies that must stay bit-identical: candidate_topk_common.py:57 f32_to_ordered_i32 (feeds Stage A's precomputed pass-0 histogram) and topk_per_row_decode.py:153 _f32_to_ord(nan_to_bottom=True) (feeds the merge's data passes and gather). They agree today, but a one-sided edit — canonicalizing -0.0, or moving NaN — silently desynchronizes the precomputed histogram from the merge's keys and mis-selects with no error. **Reviewer should ask** for one shared implementation, or a bit-pattern sweep test asserting the two mappings agree.

📝 [verified] When kv_cache is packed, a caller-supplied k_scales is silently ignored: _normalize_kv_inputs (fp8_paged_mqa_local_topk.py:53-63) returns the packed buffer's own scale view without touching the argument, so a caller migrating from the split layout that passes (q, packed_kv, scales, ...) gets the packed scales with no error even if the two differ. **Author must** raise when k_scales is not None with a packed cache, or document the discard in the signature.

PR #5529, opened 2026-09-15
[Triton/Gluon] [Perf] Shallow-copy GEMM configs instead of deep-copying them

ZhengGong-amd:
`get_gemm_config()` returns `copy.deepcopy(config)` of a bucket dict it took
from an `lru_cache`. The copy is load-bearing — callers mutate what they get
back, `compute_splitk_params()` in place and the gluon paths via
`config.pop("kernel_type")` — but every value in a GEMM bucket is a scalar, so
`dict(entry)` isolates the cached entry identically. `copy.deepcopy` takes its
atomic fast path for `int`/`str`/`bool`/`None`, so both copies hand the kernel
the *same* value objects: nothing downstream can observe the difference.

This is the rule the repo already states in `configs/CLAUDE.md` and
`.github/instructions/aiter-ops-triton.instructions.md` — "`dict(...)` for flat
entries, `copy.deepcopy` for nested ones" — and the one `get_conv_config()`
already follows, its docstring saying it is modeled on `get_gemm_config()` and
that "a deep copy only adds hot-path overhead". `get_gemm_config()` is the
shared loader for ~28 Triton GEMM kernels and runs once per launch.

## Micro

gfx950 MI355X, torch 2.12.0+git6bbd260, `d483dbf` against `d483dbf` + this
change, single tenant on an otherwise idle host. Host wall-clock on the dispatch
path, calls issued back to back without syncing; `do_bench` reports kernel time
and is blind to this. Four interleaved pairs with the order alternated, pair 0
discarded, median of the other three; 6 shapes × 4000 iterations × 9 repeats.
The trailing-`synchronize()` drain stayed at 0.005–0.014 µs/call across all 72
kept points, so the host is what was timed.

| | base | patched | | per-shape range |
|---|---|---|---|---|
| `get_gemm_config()` alone | 2.46 µs | 0.12 µs | **−95.1 %** | 92.3–95.2 % |
| `gemm_a16w16()` dispatch | 27.35 µs | 24.90 µs | **−8.7 %** | 8.3–9.9 % |
| `gemm_afp4wfp4()` dispatch | 37.42 µs | 33.87 µs | **−9.1 %** | 8.0–9.7 % |

The two GEMM families agree because each pays exactly one copy per launch: about
2.4 µs, which is 9 % of their host dispatch. The outlier shape in the first row
is M=23, which lands in a bucket whose lookup is itself slower.

## Engine startup

The one place the change has a measurable effect under vLLM's defaults. Timing
the function itself across a full `LLM()` init of gpt-oss-120b, rather than
looking for it in wall-clock startup, which weight I/O dominates:

| | calls | time in `get_gemm_config` |
|---|---|---|
| base | 51 156 | 408.8 / 397.8 ms |
| patched | 51 156 | 55.5 / 56.8 ms |

Two repeats each; the call count is identical by construction. About 345 ms per
engine start, 0.4–0.5 % of init.

## What reaches this code

Counted by instrumenting `get_gemm_config` inside the aiter tree, so
tensor-parallel workers are included, rather than inferred from config files:

| model | calls | family |
|---|---|---|
| gpt-oss-120b | 51 156 at startup, 0 per decode step | `GEMM-A16W16` |
| Kimi-Linear-48B-A3B-Instruct | 4 200 at startup, 0 per decode step | `BATCHED_GEMM-A16WFP4` |
| Qwen3-14B-FP8, MiniMax-M2.7 | 0 | — |

Under vLLM's default HIP-graph path a decode step makes no calls at all;
`--enforce-eager` on gpt-oss gives 108 per step. So the win lands on startup,
prefill, graph capture, eager execution, tuning sweeps and callers outside vLLM,
and **cannot** move graph-mode decode throughput.

## End to end

`vllm bench serve`, TP=1, ISL 2000 / OSL 150, 60 prompts, concurrency 24,
interleaved pairs with the order alternated and pair 0 discarded. Included to
show nothing regresses; by the counts above it cannot show a gain.

**Kimi-Linear-48B-A3B-Instruct**, 5 kept pairs, with an **A/A control** — the
base build against a second copy of itself, given the same separate cache roots
the real pair gets:

| | per-pair Δ tok/s | band |
|---|---|---|
| base vs patched | −3.46 −4.10 +0.76 −3.72 +0.44 | −4.1 … +0.8 % |
| base vs base (A/A) | −0.99 +0.91 +3.08 +5.02 | −1.0 … +5.0 % |

Throughput on this model is bimodal: 5 of 18 runs land near 21 500 tok/s instead
of 22 400. Three are the patched arm — and **two are an unpatched arm in the A/A
control**, which is what rules the patch out. The bands are the same width, so
this harness cannot resolve anything below about 5 % here.

TTFT does resolve, and it is the metric prefill lives in:

| | base | patched | Δ | spreads |
|---|---|---|---|---|
| Mean TTFT (ms) | 475.2 | 475.7 | +0.10 % | 0.9 % / 0.5 % |
| Median TTFT (ms) | 483.0 | 482.4 | −0.12 % | 0.6 % / 0.6 % |

**gpt-oss-120b**, 5 converged pairs (its first two are warmup — one base run came
in at 6 838 tok/s): total 28 800 → 28 866 tok/s (+0.23 %, base spread 5.1 %),
mean TPOT 7.46 → 7.43 ms, mean TTFT 461.7 → 456.9 ms.

## Accuracy

`lm_eval` GSM8K, 5-shot greedy, all 1319 items, `max_model_len=4096`,
`max_gen_toks=256`, `enable_prefix_caching=False`, `seed=1234`. `base2` is a
second run of the **unpatched** build, so its delta is the harness's noise floor.

| model | strict base | patched | Δ patch | Δ base-repeat | stderr | identical text base→patched | base→base |
| gpt-oss-120b | 0.2699 | 0.2714 | +0.0015 | +0.0197 | ±0.0122 | 2.0 % | 1.5 % |
| Qwen3-Next-80B (0 calls) | 0.8105 | 0.8249 | +0.0144 | +0.0083 | ±0.0108 | 37.1 % | 33.7 % |

On the affected model a repeat of the unpatched build moves the score 13× further
than the patch does, and in both models the patched arm reproduces base's text at
least as often as base reproduces its own. Same-verdict rate is 71.9 % and 94.6 %.
gpt-oss's absolute score is low because `strict-match` does not fit its output
format; only the within-model comparison is read here.

No measurable accuracy change on GSM8K, within the base-repeat noise floor. GSM8K
exercises one shape of one path and says nothing about long context, sampling or
multi-turn.

The end-to-end and GSM8K arms also carried two lazy-logging conversions that have
since been split out of this PR. Neither reaches those call sites in any material
number — gpt-oss never calls them, Kimi-Linear calls `dynamic_mxfp4_quant` 14
times in a whole run — so those measurements stand for the change as it ships.

## Tests

`op_tests/triton_tests/gemm/basic/test_gemm_a16w16.py` 360 passed / 648 skipped
and `test_gemm_afp4wfp4.py` 234 passed / 290 skipped, identical on both arms.

`op_tests/triton_tests/test_config_copy_contract.py` is new, and covers more than
this change. Four loaders return `dict(entry)` of an `lru_cache`d object —
`get_gemm_config`, `get_conv_config`, `get_mhc_config`, `get_mhc_post_config` —
and nothing tested that contract for any of them. A nested value in any of their
tables would be shared between the cache and the caller, and a caller mutating it
in place would poison every later lookup of that shape, with no exception raised.

Three layers: the shipped gemm, conv and mhc tables are walked for **every**
architecture rather than the one running the test; the loaders are driven to
check the isolation itself, since reading the tables alone would still pass if a
loader returned the cached object; and an inventory test fails when a new
`dict(entry)` loader appears without coverage, so this tracks the source instead
of freezing at the set that exists today.

It was run against six deliberately wrong trees, each of which takes down a
different subset — a nested value in a gemm bucket, in a conv exact-shape pin and
in an mhc bucket; a typo'd bucket key the walk does not recognise; a loader that
hands back the cached object (5 of 9); and a new uncovered shallow-copy loader.
All nine pass on the shipped tree.

It lives under `op_tests/triton_tests/` because that is what CI actually
collects — the Triton job pytest-runs that directory recursively on any PR
touching `aiter/ops/triton/**`, which this PR does.

---


vgokhale:
We discussed offline - the original data is with eager mode and the benefits are negligible with hipgraphs. I don't think the benefits outweigh the potential risks of a shallow copy. Closing.

PR #5536, opened 2026-09-15
[CI] Nightly Triton suite on MI350/MI300X with retry and auto-triage

Boss2002n:


zufayu:
## ROCm/aiter PR #5536 — [CI] Nightly Triton suite on MI350/MI300X with retry and auto-triage

**Adds a nightly 8-shard MI350+MI300X Triton workflow that retries each failed shard once (flaky vs hard-fail), comments every run on a rolling tracker issue, and on hard failures ranks suspect commits by source overlap and runs a first-parent git bisect that posts the "first bad commit" and @mentions its author.**

⚠️ [verified] `.github/scripts/triton_nightly_bisect_step.sh:32-38` classifies every bisect commit from a single pytest run, while the `test` job retries once precisely because single runs are unreliable — a flaky test that hard-fails the nightly (fails both attempts) then flakes once at an innocent midpoint and the bisect names that innocent commit as the culprit, which `.github/workflows/triton-nightly.yaml:461-466` posts as fact ("is the first bad commit", `Author: @login`) with no flakiness caveat. Both failure modes reproduced on a synthetic history: a flaky failure at an innocent commit made the bisect report that commit as first bad; a failure that never reproduces (runner/infra) made it report HEAD, i.e. the most recent innocent main commit. **Author must** require two consecutive failures (or one retry) before marking a bisect commit bad, and add a flakiness caveat to the culprit comment.

⚠️ [verified] `.github/scripts/triton_nightly_blame.py:80` gates the precise ranking on `select_triton_tests.py` exposing `TESTS`/`SRC`/`scan_imports`/`reachable` — none of which that script defines on main (it has `list_files`/`build_dependency_graph`/`find_tests_to_run` instead), so the import graph never loads and every run falls back to `aiter/ops/triton/` prefix overlap (blame.py:109). A regression that breaks Triton tests from outside that directory (`aiter/jit/core.py`, `aiter/__init__.py`, `csrc/`) ranks zero suspects and prints "look at the runner, the Triton pin or the image rather than aiter" (blame.py:125) — verified on a synthetic repo — actively pointing triage away from the real aiter culprit. `git show --name-only` also returns nothing for merge commits, though aiter is squash-merge (1 merge in the last 2000 first-parent commits), so that part is minor. **Author must** wire the graph to `select_triton_tests.py`'s real API (or widen the fallback beyond `aiter/ops/triton/`) and reword the empty result to "no commit under aiter/ops/triton/ matched" instead of directing attention off aiter.

⚠️ [verified] `.github/scripts/triton_nightly_bisect_step.sh:21` runs `git submodule update --init --recursive >/dev/null 2>&1 || true`, and aiter pins `3rdparty/composable_kernel` — during a bisect each commit pins a different submodule revision, so a transient fetch failure is silently swallowed (output discarded) and the build/tests classify that commit against the previous iteration's stale composable_kernel tree, producing a false good/bad classification and a possible false culprit with no signal in the log. **Author must** drop `|| true` and exit 125 (untestable) when the submodule update fails, so git skips the commit instead of testing the wrong tree.

⚠️ [inferred] `.github/workflows/triton-nightly.yaml:93-96` hardcodes `secrets.DOCKER_PASSWORD` and `-u rocmshared`, while every existing workflow that logs in (triton-test.yaml, aiter-test.yaml, vllm_benchmark.yaml, atom-test.yaml) uses `secrets[needs.ci_config.outputs.docker_username_secret]`/`docker_password_secret` (= DOCKERHUB_USERNAME/DOCKERHUB_PASSWORD) — `DOCKER_PASSWORD` and `rocmshared` appear nowhere else in the repo. If that secret does not exist, every shard's login fails 3x ("continuing anyway") and 16 nightly shard jobs + 2 bisect jobs pulling anonymously can hit Docker Hub rate limits, killing shards before they write a status file — the aggregate then reports "N shard(s) never reported" and the nightly is permanently red from night one. **Author must** confirm the `DOCKER_PASSWORD` secret exists and `rocmshared` is the right account, or switch to the ci-config secret indirection the rest of the repo uses.

📝 [verified] `.github/scripts/triton_nightly_bisect_step.sh:28-31` comments "Collection errors (4)" but pytest collection errors exit 2 (4 is usage error) — harmless since both map to 125, but the same mapping means a commit that breaks the failing test module's import is classified untestable rather than bad: verified on a synthetic history, a commit that both adds the test and breaks it makes every midpoint skip and `git bisect run` exits nonzero with no culprit reported, so import-breaking regressions get no attribution at all. **Author must** correct the exit-code comment and consider treating a collection error on a test file that exists at that commit as bad rather than skip.

PR #4919, opened 2026-08-21
[Triton/Gluon] [GFX950] Sparse paged attention and Sparse MLA

cagrikymk:
## Motivation

This kernel extends the existing sparse paged attention kernel (initially designed for dsv4 decode) to cover GLM5 like sparse decode/prefill as well. Also, optimizes it further for dsv4 as well.

The target is vllm, hence it consumes the default kv cache format for those models (`fp8 with tensor scale` for glm5, and `fp8 ds mla` format for DSV4.

TODO:

- Check e2e performance to make sure it matches the ubench numbers.
- [ ] dsv4
- [x] GLM 5.1

## Technical Details

Performance tuning for dsv4 + and sparse mla support for GLM like LLMs.

## Test Plan

New tests for the new path (sparse mla).

### Performance evaluation
Execution times are in us and it includes the reduce time if it exists.

#### GLM 5.2

fp8 cache with a per-tensor scale, `topk = 2048`, q pre-quantized. `shipped` is the
default configuration the launcher picks; no flags.

ASM is the existing aiter/ASM path from VLLM.

Warm: No L2 cache flushing, cold is with cache flushing to remove cache effect between runs.

H=8 is TP8, H=16 is TP4. Numbers are from MI355 + Triton 3.7.1

| shape | H | warm asm | warm gluon | warm speedup | cold asm | cold gluon | cold speedup |
| --- | --: | --: | --: | --: | --: | --: | --: |
| prefill 1 seq x 8k | 16 | 955.0 | **739.7** | **1.29x** | 1016.7 | **809.1** | **1.26x** |
| prefill 1 seq x 8k | 8 | 934.7 | **713.1** | **1.31x** | 1018.1 | **795.9** | **1.28x** |
| prefill 8 seq x 1k | 16 | 389.2 | **266.9** | **1.46x** | 425.7 | **292.8** | **1.45x** |
| prefill 8 seq x 1k | 8 | 389.9 | **256.8** | **1.52x** | 427.7 | **277.8** | **1.54x** |
| decode C=1 | 16 | 28.8 | **11.7** | **2.46x** | 30.8 | **16.6** | **1.86x** |
| decode C=1 | 8 | 29.2 | **11.8** | **2.47x** | 30.7 | **15.8** | **1.94x** |
| decode C=32 | 16 | 21.8 | **18.5** | **1.18x** | 30.2 | **28.2** | **1.07x** |
[...]

PR #5539, opened 2026-09-15
[Triton/Gluon] [gfx942] Enable sparse_mla_fwd on gfx942

jin-amd:
Enables `sparse_mla_fwd` on gfx942, on top of #4919.

The Gluon kernel needs no change to run there — the `gl.amd.cdna4` intrinsics it
uses (`async_copy`, `mfma`, `buffer_load`/`buffer_store`) all lower fine on
gfx942 under Triton 3.7.1. The arch gate was the only thing in the way, plus two
launch-config values sized against gfx950's LDS budget.

### Why the tile has to change

gfx942 (CDNA3) has **64 KB of LDS per workgroup against gfx950's 160 KB**. The
bf16 KV tile alone is `BLOCK_K * (kv_lora_rank + LDS_PAD) * 2 B`, so the gfx950
tile of 64 asks for 68 608 B and will not launch:

```
OutOfResources: out of resource: shared memory, Required: 68608, Hardware limit: 65536
```

`BLOCK_K=32` is the largest that fits.

**`num_warps` is decoupled from `BLOCK_K` in the same change.** It was derived as
`BLOCK_K // 16`, so capping the tile for LDS would have dropped num_warps from 4
to 2 as a *side effect* rather than as a tuning decision. Keeping it at 4 is
worth **1.19× → 1.55×** on the prefill shapes, so it is worth making explicit.

### `dot_precision="fp8"` is rejected on gfx942

This one is a genuine hardware limit and was the only real surprise. The fp8 dot
path feeds the cache's own code points to the matrix core, which needs **OCP
e4m3 MFMA — that arrives with CDNA4**. CDNA3 has fp8 MFMA in the fnuz encoding
only, so OCP code points read back wrong.

It was **silently wrong (rel-err 7.5e-1) rather than faulting**, which is why
this raises now, naming the alternative. `dot_precision="bf16"` dequantizes the
tile on its way into LDS and works on every arch and every cache format —
including the fp8 caches, which is what production uses here.

### gfx950 is unchanged

`_arch_block_k` → 64, `_arch_num_warps` → 4, `lds_limited` False, and
`FP8_DOT_ARCHS` contains gfx950, so every value and branch resolves exactly as
before. I do not have gfx950 hardware, so that claim is by construction rather
than by test — worth a second pair of eyes.

### Validation

On MI325X (gfx942), TP4, GLM-5.3-Flash at 131 k context. This matters on that
model because its NoPE MLA has `qk_rope_head_dim = 0`, which has no AITER path
today, so vLLM falls back to a vendored Triton gather+dot — the single largest
kernel in the model in both phases.

`op_tests/triton_tests/attention/test_sparse_mla.py`:

```
32 passed, 12 skipped in 23.61s
SKIPPED [12] fp8 matrix-core dots need OCP e4m3, absent on gfx942
```

Kernel bucket, from a real torch profile at concurrency 12 (rank 0):

| phase | bucket | Triton (vLLM) | this kernel | speedup |
|---|---|---|---|---|
| decode | per step | 3.099 ms | 0.400 ms | **7.74×** |
| prefill | per call | 10 988 µs | 10 330 µs | 1.06× |

End-to-end, 131 k in / 1024 out, same image both arms with only the dispatch
switched:

| conc | out tok/s before → after | delta |
|---|---|---|
| 2 | 80.92 → 91.36 | **+12.90%** |
| 4 | 110.96 → 121.09 | +9.13% |
| 8 | 129.61 → 137.74 | +6.27% |
| 12 | 141.29 → 147.94 | +4.71% |
| 16 | 143.24 → 149.89 | +4.64% |

**Won 7/7 rows, mean +7.14% output throughput, zero failed requests.** gsm8k
(5-shot, full 1319 questions) is neutral: 0.9719 vs 0.9712 strict-match, against
a ±0.0046 standard error.

The gain being largest at low concurrency is the signature of this kernel rather
than noise: the operator is batch-independent, so replacing it takes a roughly
constant ~2.7 ms off every decode step, which is 16.6% of TPOT at concurrency 2
but 4.9% by concurrency 16.

### Note on prefill

The 1.06× prefill figure is the honest one, measured in a profile. A
microbenchmark using `randperm` top-k indices reports 1.55× for the same shape,
but that flatters it — random indices give the Triton gather far worse locality
than production does (11.0 ms real vs 15.7 ms synthetic for the same shape),
while this kernel gathers whole 512 B rows and is locality-insensitive. **Prefill
microbenchmarks on this operator should not be trusted without a profile.**


op_tests/op_benchmarks/triton/bench_sparse_mla.py
```diff
@@ -271,8 +281,8 @@ def main():
             f"--num_tokens {over} exceeds --context {args.context}: a sequence "
             "cannot prefill more tokens than its context holds"
         )
-    if arch_info.get_arch() != "gfx950":
-        raise SystemExit(f"sparse_mla_fwd is gfx950-only (got {arch_info.get_arch()})")
+    if arch_info.get_arch() not in SUPPORTED_ARCHS:
```
frida-andersson, line 282: Did you run the default bench on gfx942 after this? This used to bail out before run_benchmark, and the fp8 series in there is unconditional, dot_precision="fp8" now raises on this arch, and triton's harness doesn't catch it, so I'd expect the first fp8 point to kill the run before it prints anything. Or maybe there's a skip I'm missing somewhere?

jin-amd: Good catch, and no — you're not missing a skip. I ran the op_tests suite on gfx942 but not the bench, which is exactly how this got through.

Reproduced on the as-submitted file with the default invocation: exit 1, ValueError: dot_precision='fp8' is not supported on gfx942 out of the first fp8 point, and no table rows at all. Widening the gate replaced a clean early exit with a traceback partway in, which is worse than what was there before.

Fixed in 7c1c89db3 by building the series list from FP8_DOT_ARCHS, with a printed note so the omission is visible rather than silent:

```
note: skipping the fp8-dot series, gfx942 has no OCP e4m3 matrix core
    phase  num_seqs  num_tokens  num_heads  context  topk  bf16 dots (Time (ms))
0  decode         1           1         16     8192  2048               0.039077
1  decode         8           1         16     8192  2048               0.042981
2  decode        64           1         16     8192  2048               0.101375
3  decode       256           1         16     8192  2048               0.289891
```
op_tests still 32 passed, 12 skipped.

PR #5543, opened 2026-09-15
[FlyDSL] Add token-major layout option to GDN prefill h kernel

johannes-graner:
Summary

Adds an optional token-major layout for the w/u inputs and the v_new output of the FlyDSL gated-delta-rule hidden-state kernel chunk_gated_delta_rule_fwd_h_flydsl_opt. New public param wu_head_major: bool = True on the host wrapper (aiter/ops/flydsl/linear_attention_prefill_kernels.py); when False, w/u are consumed token-major ([B,T,H,D]) and v_new is stored token-major ([B,T,H,V]). Default True preserves the existing head-major behavior exactly. Implemented via a compile-time const_expr branch in the device kernel (aiter/ops/flydsl/kernels/gdr_prefill/chunk_gated_delta_h.py), so the head-major path is unchanged.

Motivation: lets a consumer with native token-major GDN tensors (e.g. SGLang's linear-attention path) reach the kernel without a transpose.

Technical Details

- Head-major path is provably unchanged: emitted ISA is byte-identical between base and this branch (same VGPR/occupancy). Zero regression by construction (compile-time dead-branch elimination).
- Correctness: token-major output is bit-exact against the head-major reference on dense and production varlen shapes, including an unaligned tail.
- wu_head_major=False changes both the v_new store base/stride and the w/u input loads, which move from head-major ([B,H,T,D], token-row stride D) to token-major ([B,T,H,D], token-row stride H*D) — see chunk_gated_delta_h.py:394-400. K/V stays the unit-stride innermost dimension in both layouts, so lane-level coalescing is unchanged; the difference is inter-token-row locality (head-major advances 256 B/row for K=V=128 BF16, token-major advances H·256 B/row).
- Performance is layout- and architecture-dependent, not neutral. On gfx950 (CDNA4, the target arch) token-major is within noise to slightly faster. On gfx942 (CDNA3) token-major measures ~4.3% slower (dense T=1024) and ~6.1% slower (varlen 8×1024) than head-major, from the reduced token-row spatial locality above (a secondary cache/prefetch effect, not a coalescing failure). For a consumer with native token-major tensors, the saved host transpose can still make wu_head_major=False a net win; on gfx942 it is a trade-off to weigh against the transpose cost.

AOT / prebuild

The FlyDSL AOT enumeration (aiter/aot/flydsl/chunk_gdn_h.py) now fans out wu_contig over both values with layout-matched compile dummies, so the token-major specialization is prebuilt into the packaged cache rather than cold-JIT-compiling on the first wu_head_major=False serving-path call. Verified on gfx950: enumeration count doubles (288 → 576 jobs, 288 head-major + 288 token-major), and a fresh-process token-major call under FLYDSL_RUNTIME_RUN_ONLY=1 (which forbids compile-on-miss) hits the prebuilt cache — with an empty-cache negative control raising the expected "no usable AOT cache" error, confirming the hit depends on the prebuilt artifact and not a JIT fallback.

Test Plan

- op_tests/test_flydsl_linear_attention_prefill.py runs a direct token-major == head-major torch.equal exact assertion (on h, layout-normalized v_new, and final_state) on every case in the default sweep.
- Default run also exercises two dedicated review cases: review-unaligned-varlen (varlen T=1000, a non-multiple-of-BT=64 tail) and review-dense-b2 (dense_batch=2, token-major dense base path).
- ISA diff of the head-major path between base and this branch.

Test Result

- Bit-exact match (torch.equal) between token-major and head-major outputs on all tested shapes, including the unaligned-tail and dense_batch>1 cases, on gfx950.
- Identical emitted ISA (VGPR/occupancy) for the head-major path, confirming no regression.

Submission Checklist

- [x] Default behavior (wu_head_major=True) unchanged
- [x] New path validated bit-exact against reference (direct token-major == head-major torch.equal assertion)
- [x] Token-major specialization AOT-prebuilt (no cold-JIT serving stall)
- [x] Performance characterized per architecture (gfx950 neutral/faster; gfx942 ~4–6% slower


zufayu:

**Adds an opt-in token-major ([B,T,H,D]) memory layout for the w/u inputs and v_new output of the FlyDSL gated-delta-rule prefill hidden-state kernel, selected by a new default-preserving `wu_head_major=True` flag on the host wrapper.**

Validation (deterministic): NOT RUN — REVIEW_AUTO_VALIDATE=0 (auto-validation disabled; environment gap, not a PR defect). The single test target op_tests/test_flydsl_linear_attention_prefill.py existed; a hand-run of it on gfx942 observed err=0 on every cell for both layouts (dense T=1024; varlen 8x1024; varlen 4x1000 unaligned tail).
Perf (advisory): MEASURED — Qwen3.5-35b tp=1 on one idle gfx942, base vs head back to back, 101 iters/cell: default head-major path 60.04 -> 59.86 us (dense T=1024, -0.3%) and 450.80 -> 450.26 us (varlen 8x1024, -0.1%), flat; token-major vs head-major on head 62.46 vs 59.86 us (dense, +4.3%) and 477.85 vs 450.26 us (varlen, +6.1%).

⚠️ [verified] aiter/aot/flydsl/chunk_gdn_h.py pins "wu_contig": True in _FIXED_SWITCHES (line 67) and builds only head-major compile dummies (lines 215-218), while linear_attention_prefill_kernels.py:817 now lets the runtime request the token-major specialization (wu_head_major=False reaches the WU_CONTIGUOUS=False compile key) — that AOT file's own docstring fans out g_head_major/use_state_indices precisely because non-tuned switches fork the compiled artifact, so the first wu_head_major=False call on an AOT-cached box cold-JIT-compiles inline, a seconds-scale stall in the serving path. **Author must** fan wu_contig out in the AOT enumeration (with token-major dummy shapes) or document the option as JIT-only.
⚠️ [verified] The PR description claims "the layout only changes v_new's store base/stride" and "Performance is not affected by layout", but wu_head_major=False also switches the w/u input loads to token-major H*K/H*V strides (chunk_gated_delta_h.py:394-400), and hand-measured token-major runs +4.3% (dense, 62.46 vs 59.86 us) / +6.1% (varlen, 477.85 vs 450.26 us) slower than head-major — a consumer weighing this flag against a host transpose would be deciding on wrong numbers. **Author must** correct the description and document the token-major load-stride cost (the saved transpose may still make it a net win).
⚠️ [verified] The Test Plan/Test Result claims are not enforced by op_tests/test_flydsl_linear_attention_prefill.py: "bit-exact match between token-major and head-major" is never asserted (each layout is only checked against the fp32 reference at rtol=atol=2e-2, where checkAllclose tolerates up to 5% mismatched elements and err is recorded but never asserted), "including an unaligned tail" never runs (every catalog length — dense 1024..65536, varlen seqlens 1024..8192, totals 8192..65536 — is a multiple of BT=64 and no group sets head_seqlens), and the B>1 dense token-major base path has no catalog coverage (dense_batch is never >1); hand-run checks on gfx942 (torch.equal between layouts on B=1/2/3, tails at T=1000/257, err=0 on all cells) suggest the gap is claim-vs-test rather than a live defect. **Author must** add a direct token-major == head-major exact assertion, a non-multiple-of-64 catalog length, and a dense_batch>1 token-major case.
📝 [inferred] The motivation names SGLang's linear-attention path as the consumer, but no in-tree caller passes the new wu_head_major kwarg on the linear_attention_prefill_kernels.py wrapper (the triton chunk.py dispatch and both test files use the default) and no downstream PR is linked from the description. **Reviewer should ask** who consumes the new flag and link the SGLang-side PR (ci:sglang covers Qwen 3.5 accuracy when it lands).

johannes-graner:
Thanks @zufayu for the review!

I have addressed all verified points in PR description and commits [0f327f5](https://github.com/ROCm/aiter/pull/5543/commits/0f327f5ec4d49dc5f3832db43db9ba046174c410) and [22bd24c](https://github.com/ROCm/aiter/pull/5543/commits/22bd24c38b24e43f58f1271878ecfe68afcb9d20).

The claim of no effect on performance still holds on gfx950 (which is also the SGLang target), but I added the caveat of being a regression on gfx942.

The SGLang PR is not published yet, but another team is working on it and are currently integrating this PR. So I cannot link to the PR, but the token-major path will be consumed.

op_tests/test_flydsl_linear_attention_prefill.py
```diff
@@ -740,55 +739,115 @@ def test_chunk_gdn_prefill_h(
...
+                k, w_c, u_c, g=g_hm, **common
+            )
+            # Unsupported shapes leave the HIP cells nan rather than dropping the row.
+            if _hip_k5_supported(case):
+                candidates["hip"] = lambda: chunk_gated_delta_rule_fwd_h_hip_fn(
+                    k, w_c, u_c, g=g_hm, g_head_major=True, **common
+                )
+
+        for name, fn in candidates.items():
+            (h, vn, fs), us = run_perftest(fn)
```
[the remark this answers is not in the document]
johannes-graner: The tests are fast enough that this does not really matter.

op_tests/test_flydsl_linear_attention_prefill.py
```diff
@@ -391,6 +391,25 @@ def _k5_varlen_groups(model_name: str, hv: int) -> list[PrefillGroup]:
     *_k5_dense_groups(K5_MODELS["397b"]["label"], K5_MODELS["397b"]["Hv"]),
     *_k5_varlen_groups(K5_MODELS["35b"]["label"], K5_MODELS["35b"]["Hv"]),
     *_k5_varlen_groups(K5_MODELS["397b"]["label"], K5_MODELS["397b"]["Hv"]),
+    # Dedicated review case: varlen with a non-multiple-of-BT tail (T=1000 for BT=64).
+    PrefillGroup(
+        model_name="review-unaligned-varlen",
```
samremes: the comment and test name should be named after the actual test case, not where it came from (a review comment?).

johannes-graner: Thanks for the catch!

PR #5545, opened 2026-09-15
[Triton/Gluon] Log with % placeholders instead of f-strings

Boss2002n:
An f-string inside a logging call is built before the call, so the message is formatted whether or not the level is enabled. AITER_TRITON_LOG_LEVEL defaults to WARNING, so every _LOGGER.info(f"...") on a hot wrapper path formats a string that is then discarded.


aiter/ops/triton/comms/iris.py
```diff
@@ -109,14 +109,26 @@ def calculate_heap_size(
...
-        f"  Gather buffer: {mem_gather:,} bytes\n"
-        f"  Total (with {overhead_factor}x overhead): {total_with_overhead:,} bytes "
-        f"({total_with_overhead / (1024**3):.2f} GB)"
+        "Heap size calculation: M=%d, N=%d, dtype=%s, world_size=%d, "
+        "quant_mode=%s, all_gather=%s\n"
+        "  Input: %d bytes\n"
+        "  RS buffer: %d bytes\n"
+        "  Quant buffer: %d bytes\n"
+        "  Gather buffer: %d bytes\n"
+        "  Total (with %sx overhead): %d bytes (%.2f GB)",
```
brunomazzottiamd: `overhead_factor` is a `float`, please check function signature:

```python
def calculate_heap_size(
    M: int,
    N: int,
    dtype: "torch.dtype",
    world_size: int | None = None,
    quant_mode: str = "none",
    all_gather: bool = True,
    overhead_factor: float = 1.2,
) -> int:
```

aiter/ops/triton/gated_delta_net/gated_delta_rule.py
```diff
@@ -155,8 +155,14 @@ def fused_recurrent_gated_delta_rule(
 
     # Log operation
     _LOGGER.info(
-        f"GATED_DELTA_RULE: q={tuple(q.shape)}, k={tuple(k.shape)}, v={tuple(v.shape)}, "
-        f"scale={scale}, use_g={g is not None}, use_gk={gk is not None}, use_gv={gv is not None}"
+        "GATED_DELTA_RULE: q=%s, k=%s, v=%s, scale=%s, use_g=%s, use_gk=%s, use_gv=%s",
+        tuple(q.shape),
+        tuple(k.shape),
+        tuple(v.shape),
+        scale,
```
brunomazzottiamd, line 162: `scale` is `float | None`. I tried:

- `logger.info("%s", 1.0)` - works
- `logger.info("%s", None)` - works
- `logger.info("%f", 1.0)` - works
- `logger.info("%f", None)` - throws `TypeError: must be real number, not NoneType`

I think it's worth adding this caveat to the docs, so AI agents are better informed to do the right thing! Use `%s` if it's nullable.

aiter/ops/triton/moe/moe_gemm_mxfp8.py
```diff
@@ -55,9 +55,12 @@ def moe_gemm_mxfp8(
         Output tensor ``[total_tokens, N]``.
     """
     _LOGGER.info(
-        f"MOE_GEMM_MXFP8: lhs={tuple(lhs.shape)} rhs={tuple(rhs.shape)} "
-        f"x_scale={tuple(x_scale.shape)} w_scale={tuple(w_scale.shape)} "
-        f"block_size={quant_block_size}"
+        "MOE_GEMM_MXFP8: lhs=%s rhs=%s x_scale=%s w_scale=%s block_size=%s",
```
brunomazzottiamd: `quant_block_size` is `int`, we should use `block_size=%d`.

aiter/ops/triton/_triton_kernels/flash_attn_triton_amd/utils.py
```diff
@@ -140,7 +140,7 @@ def cu_count(self) -> int:
             num_warps=conf.pop("num_warps", 4),
         )
 except Exception as e:  # noqa: BLE001
-    logger.warning(f"FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON parse error: {e}")
+    logger.warning("FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON parse error: %s", e)
```
brunomazzottiamd: Is this really a warning or an error?

`logger.exception()` logs the message at `ERROR` level and appends the full traceback underneath it. Something like this:

```python
try:
    ...
except Exception:
    logger.exception("FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON parse error.")
```

I'm not against using the `WARNING` level, what caught my attention was the fact that we may be trying to replicate a feature that's already baked into standard `logging` module (i.e. logging exceptions).

Please evaluate at your discretion.

Source: https://docs.python.org/3/library/logging.html#logging.Logger.exception

aiter/ops/triton/utils/tuned_config_utils.py
```diff
@@ -103,14 +103,19 @@ def get_tuned_kernel_config(
         )
     except BaseException as error:  # noqa: BLE001 -- no accelerator/unreadable file
         logger.warning(
```
brunomazzottiamd: Please evaluate previous advice w.r.t. `logger.exception()` method.

brunomazzottiamd (changes requested):
I think the only real issues to be fixed are the ones from `aiter/ops/triton/comms/iris.py` and `aiter/ops/triton/moe/moe_gemm_mxfp8.py`. The other comments are suggestions to be evaluated.

PS: Satya, please do not force push, add the next set of changes as a new commit. By this way is much easier for me to approve the PR, it's a matter of checking the last commit.

brunomazzottiamd (dismissed):

brunomazzottiamd (approved):

brunomazzottiamd:
**Flash Attention Integration** CI jobs have been broken for sometime. Such failures aren't a blocker for merging this PR.

PR #5549, opened 2026-09-15
[CK] [FlyDSL] [FMoE] Enable Silu A16W4 INTERLEAVE fused_moe on gfx950

SamiAario-AMD:
## Motivation

Silu A16W4 (bf16/fp16 activations × MXFP4 weights, GateMode.INTERLEAVE) could not run decode-sized fused_moe correctly. CK2stages has no A16W4 instance (the JIT module is A=B=fp4x2), so moe_stage1_heuristic_dispatch rejected the activation. Only ksplit > 1 shapes escaped to CK-Tile; batches where get_ksplit is 0 crashed. FlyDSL stage1 also assumed GGUU (standard) W1, while INTERLEAVE Silu consumes GUGU.

## Technical Details

Silu A16W4 (`bf16`/`fp16` activations × MXFP4 weights, `GateMode.INTERLEAVE`) could not run decode-sized `fused_moe` correctly. CK2stages has no A16W4 instance (the JIT module is `A=B=fp4x2`), so `moe_stage1_heuristic_dispatch` rejected the activation. Only `ksplit > 1` shapes escaped to CK-Tile; batches where `get_ksplit` is 0 crashed. FlyDSL stage1 also assumed GGUU (`standard`) W1, while INTERLEAVE Silu consumes GUGU.

The branch:

1. **Routes Silu A16W4 `ksplit=0` through CK-Tile** (`silu_mxfp4_bf16_cktile`), same split-k rule as Swiglu A16W4, so decode B no longer hits the unsupported CK2stages heuristic.
2. **Wires INTERLEAVE to GUGU** in FlyDSL stage1 and the FMoE tuner (`w_layout=guinterleave`, `shuffle_weight_a16w4` / `shuffle_scale_a16w4` with `gate_up=True` on W1).
3. **Adds gfx950 FlyDSL tiles** in `aiter/configs/model_configs/silu_a16w4_tuned_fmoe.csv` (token 1–64 for Qwen-like 2048/512, MiniMax-like 3072/1536, DeepSeek-like 7168/2048) so those decode shapes pick FlyDSL instead of untuned CK-Tile.

## Test Plan

Tests added in op_tests/test_moe_2stage.py

## Test Result

**Env:** `flydsl_venv`, `HIP_VISIBLE_DEVICES=6`, gfx950 (MI355X).

```bash
python op_tests/test_moe_2stage.py --no-legacy --no-flydsl-csv
```

**Outcome:** exit 0.

| Check | Result |
|---|---|
| `test_silu_a16w4_interleave_tiles` dispatch | Pass — all 21 `silu_a16w4_tuned_fmoe.csv` rows match `get_2stage_cfgs` |
| Qwen decode B=1 numerics (`2048×512`, E=512, topk=10, INTERLEAVE) | Pass on harness cosine gate (`logits_diff` ≤ 1e-3). Ran `flydsl_moe1_abf16_wfp4_bf16_t16x32x128_w4_bnt0_xcd4_kw4` + matching stage2 |
| `test_output_buffer_contract` | Pass |

`checkAllclose(atol=0.01, rtol=0.01)` failed on the Qwen cell (max abs ~864, ~76% of elements). That is the usual MXFP4 abs-error pattern; the test does not fail unless cosine is also over 0.01.

Not exercised: full CSV sweep and legacy product (`--no-flydsl-csv --no-legacy`).

**Other log noise (non-failing):** 12 duplicate primary FMoE rows / 96 after disabling `act_type`; Torch `aiter::fused_moe_` output-aliasing deprecation; ROCTracer duplicate-flow; NUMA balancing on.

## Submission Checklist


coderfeli:
why need gugu here？

coderfeli (changes requested):
don't change default behavior

csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py
```diff
@@ -819,6 +819,15 @@ def run_flydsl_stage1_out(
             xcd_swizzle=kparams.get("xcd_swizzle", 0),
             bias=bias,
             k_wave=kparams.get("k_wave", 1),
+            w_layout=(
+                "guinterleave"
+                if (
+                    act_type == ActivationType.Silu
+                    and kparams.get("a_dtype") == "bf16"
+                    and kparams.get("b_dtype") == "fp4"
+                )
```
coderfeli, line 828: why change this to guinterleave?

zufayu:

**Routes Silu A16W4 (bf16/fp16 activations × MXFP4 weights, GateMode.INTERLEAVE) decode-sized fused_moe away from the rejecting CK2stages heuristic into CK-Tile split-k (untuned) or new gfx950 FlyDSL tiles (silu_a16w4_tuned_fmoe.csv), and threads a `w_layout="guinterleave"` flag through FlyDSL stage1 so INTERLEAVE models' GUGU W1 is consumed in its native layout.**

Validation (deterministic): NOT RUN — triage required the target (op_tests/test_moe_2stage.py, runtime surface in aiter/fused_moe.py + aiter/ops/flydsl/moe_kernels.py) but auto-validation is disabled (REVIEW_AUTO_VALIDATE=0) and no head-matched report was supplied; every runtime-behaviour statement below is traced to code, not to a run.
Perf (advisory): NOT RUN — perf triage required (runtime surface changed, perf claimed, target has a perftest harness) but the target is gfx950-gated and this review box is gfx942, and auto-validation is disabled; no base-vs-head timing exists, so the PR's speed claims are [inferred].

🔴 [verified] The AOT pre-compiler was not taught the new layout variant, so every row of the new CSV deterministically misses the AOT cache. The runtime dispatch derives `w_layout` from gate_mode (aiter/fused_moe.py:3164-3170), so a Silu a16w4 INTERLEAVE call on any of the 21 rows (e.g. Qwen 2048/512, E=512, topk=10, token=1) requests the kernel `gemm1_a16w4_port_a16w4_guinterleave_h2048_i512_ne512_bm16_tn32_...`, but `_precompile_a16w4_to_cache` (aiter/aot/flydsl/moe.py:905) hardcodes `w_layout="standard"` — despite its docstring promising "the cache key then matches runtime by construction" — so AOT (whose DEFAULT_CSVS merges model_configs/*tuned_fmoe*.csv) prebuilds `gemm1_a16w4_port_a16w4_h2048_...`, a variant the runtime never requests for these rows. First Silu a16w4 decode call then JIT-compiles at runtime, and under `FLYDSL_RUNTIME_RUN_ONLY=1` (AOT-only serving) it raises RuntimeError; the PR's test passes `check_aot_cache=False` and skips these rows in the CSV sweep, so nothing verifies AOT coverage. **Author must** thread `w_layout` through `_precompile_a16w4_to_cache` the same way the runtime derives it (and through `compile_flydsl_moe_stage1`, aiter/ops/flydsl/moe_kernels.py:661) and run one case under `FLYDSL_RUNTIME_RUN_ONLY=1`.

⚠️ [verified flip, inferred reachability] The `w_layout` derivation changes default behavior for existing a16w4 FlyDSL rows, not just the new Silu ones. `aiter/fused_moe.py:3164-3170` applies to every `flydsl_moe1` a16w-port dispatch: the existing kimik3_a16w4 (Situv2) and minimax_m3_a16w4 (Swiglu) tuned rows flip from the previously hardcoded "standard" (GGUU) to "guinterleave" (GUGU) the moment they are served with `gate_mode=INTERLEAVE` — silent wrong output if such a deployment exists, since W1 gate/up rows would be read from the wrong offsets. aiter's own tests run both Swiglu and Situv2 a16w4 as SEPARATED, so no test in or out of this PR would catch the flip, and I could not exhibit a serving config that passes INTERLEAVE for those rows (GPT-OSS a16w4 is CK-Tile-routed and unaffected). This is the maintainer's live CHANGES_REQUESTED ("don't change default behavior", "why need gugu here?"). **Author must** scope the guinterleave derivation to the Silu-a16w4 combination it was tuned for, or demonstrate the INTERLEAVE serving config for the affected rows.

⚠️ [verified] The headline crash fix — the CK-Tile route — ships with zero numerics coverage. The new `silu_mxfp4_bf16_cktile` gate (aiter/fused_moe.py:3276-3283) is what stops the ksplit=0 crash, but no test reaches it: `test_silu_a16w4_interleave_tiles` gates on gfx950 where every decode shape matches a tuned CSV row and goes to FlyDSL instead, the CSV sweep skips Silu a16w4 rows by dtype combo (op_tests/test_moe_2stage.py:928-936), and the legacy sweep hardcodes Swiglu for the a16w4 triple — so on gfx942, where CK-Tile is the *only* Silu a16w4 decode path, nothing validates the route (including its `post_activation_layout="auto"` → interleaved post-activation for fp4x2 weights). The comment at op_tests/test_moe_2stage.py:925-927 also overstates coverage: "Qwen/MiniMax numerics" — only the Qwen 2048/512 token=1 cell runs numerics; MiniMax (CSV err2 1.2-1.3%) and DeepSeek (err2 ~7%) rows are dispatch-pinned only. **Author must** add a numerics case that forces the CK-Tile route (non-CSV shape or tuned-config bypass) and correct the coverage comment.

⚠️ [verified] `w_layout` is derived from three different signals that disagree, and the impl silently coerces typos to "standard". The dispatch keys on `gate_mode == INTERLEAVE` (aiter/fused_moe.py:3167), `_flydsl_stage1_wrapper`'s default keys on `activation == Silu` (aiter/fused_moe.py:1840-1844), and the tuner on `act_type == Silu and a_dtype == "bf16" and b_dtype == "fp4"` (csrc/ck_gemm_moe_2stages_codegen/gemm_moe_tune.py:822-830): for a Silu a16w4 SEPARATED call the wrapper default picks guinterleave while the dispatch picks standard. `_flydsl_moe_stage1_impl` (aiter/ops/flydsl/moe_kernels.py:1552-1556) then coerces any value that is not "guinterleave" to "standard", so a mistyped `w_layout` can never reach `compile_gemm1_a16w4_port`'s assert. No current caller hits the wrong default (a16wi4 is coerced back by the `b_dtype == "fp4"` guard; the FlyDSL heuristic is unreachable for bf16 activations except Situv2), so this is latent. **Author must** derive `w_layout` from one source at every site and let unknown values raise instead of coercing.

⚠️ [verified] No downstream CI label on a dispatch change downstream serves, and the perf claim is unquantified. The new rows in aiter/configs/model_configs/silu_a16w4_tuned_fmoe.csv activate the changed aiter/fused_moe.py dispatch for Qwen-like (2048/512, E=512, topk=10), MiniMax-like (3072/1536) and DeepSeek-like (7168/2048) MXFP4 decode shapes — Qwen 3.5 and DeepSeek-R1-MXFP4 accuracy are `ci:sglang` jobs — but the PR carries only CK/FlyDSL labels, so every downstream job is skipped; the body's only perf claim ("CK-Tile also measured faster than the untuned FlyDSL a16w-mix port on DeepSeek/MiniMax/Qwen at B=1..32") carries no numbers or units (the CSV's us1/us2 are kernel-only timings with no baseline), and no base-vs-head timing exists. **Author must** add the minimal `ci:sglang` label (the DeepSeek shape is also exercised by atom/vllm, arguing for `ci:all`) and attach the CK-Tile-vs-FlyDSL numbers with units.

PR #5555, opened 2026-09-15
[FlyDSL] Add QuickAllReduce INT6 and INT5

msaffari-amd:
## Motivation

FlyDSL quick all-reduce only had an INT4 wire format (~19 dB SQNR). This adds
INT6 and INT5, so the mesh covers the full quality/wire range at FlyDSL's speed.

TP8, gfx950, bf16, HIP-graph. Decode 512x5120, prefill 32768x5120. Each row is
the fastest of an ST x grid sweep; SQNR is vs fp32 NCCL.

| Codec | Decode | Decode SQNR | Prefill | Prefill SQNR |
| INT4 | 26.3 µs | 19.2 dB | 524 µs | 19.2 dB |
| INT5 | 28.7 µs | 25.0 dB | 620 µs | 25.0 dB |
| INT6 | 31.2 µs | 30.2 dB | 712 µs | 30.2 dB |

Against HIP QuickReduce at the same bit width (HIP has no INT5):
| Codec | FlyDSL decode | HIP decode | FlyDSL prefill | HIP prefill |
| INT4 | 26.3 µs / 19.2 dB | 29.0 µs / 18.3 dB | 524 µs / 19.2 dB | 1057 µs / 18.3 dB |
| INT6 | 31.2 µs / 30.2 dB | 35.3 µs / 30.4 dB | 712 µs / 30.2 dB | 1137 µs / 30.4 dB |

## Technical Details

INT6 and INT5 are not new engines. The mesh and ring kernels were already
codec-generic, so per review this folds the INT4 and INT6 hosts and kernels into
one and drops `int4` from names that no longer describe the code:
- `quick_allreduce_int4.py` + `quick_allreduce_int6.py` -> `quick_allreduce.py`,
  class `FlyQuickAllReduce`
- `kernels/quick_allreduce_int4.py` + `..._int6.py` -> `kernels/quick_allreduce_mesh.py`
- `kernels/quick_allreduce_int4_ring.py` -> `kernels/quick_allreduce_ring.py`
- `quick_allreduce_int4_ipc.py` -> `quick_allreduce_ipc.py`
- both multi-GPU test files -> `test_flydsl_quick_allreduce.py`
The codec is then a parameter: `MESH_CODECS = ("int4", "int5", "int6", "fp16")`,
one `Codec` each in `quick_allreduce_codec.py`, and a codec-keyed
`MESH_ST_LADDER` behind `mesh_st_ladder(codec, world_size)`. Both new formats
keep INT4's nibble plane byte-identical and add a dense plane on top, so every
region stays on the 64 B fabric sector grid:
- INT5: 1024 B nibbles + 256 B 1-bit plane + 128 B scale = 1408 B, 22 sectors
- INT6: 1024 B nibbles + 512 B 2-bit plane + 128 B scale = 1664 B, 26 sectors
`_Algorithm.single_codec` marks the mesh as carrying one format across both
laps, so it does not inherit the ring's TP8 `rs_codec=int6` / `ag_codec=int4`
default. Mesh only for INT5/INT6; the ring is unchanged, and the communicator
still constructs the mesh INT4 default.

## Test Plan

`op_tests/multigpu_tests/test_flydsl_quick_allreduce.py` on 8 GPUs plus the
flydsl codec and policy tests. Mesh cases are parametrized on codec with
per-format SQNR floors (int4 18 dB, int5 24 dB, int6 25 dB) so a silent
fallback to narrower bits fails instead of passing.

## Test Result

All suites pass. Benchmarks as in the tables above.
Known, tracked separately: below TP8 on an uncached inbox the mesh can publish
the handshake flag before the payload is visible to the peer, because
`_publish` waits with `s_waitcnt vmcnt(0)` (local issue-completion, not peer
visibility). At TP2 with ST=8 and a large grid that costs INT6 ~5 dB and shows
~0.27% of elements differing between two runs of identical input; INT4 has the
same hole, smaller in dB. TP8 (the shipping configuration) is bit-stable, and
the finegrained/default inboxes do not flake.

## Submission Checklist


zufayu:

**Admits INT6 as a wire format of the FlyDSL mesh quick all-reduce — the mesh kernel was already codec-generic, so this wires the INT6 codec into MESH_CODECS with its own super-tile ladder, makes the host resolve one codec across both mesh laps, and renames the codec-generic QuickAllReduceInt4 host/kernels/tests to FlyQuickAllReduce / quick_allreduce* — lifting mesh accuracy from ~19 dB to ~30 dB SQNR at the same two-shot schedule.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0) and 4 candidate test targets with no unique one to name; this box is a single gfx942, so the TP2/4/8 collectives cannot run here — an environment gap, not a PR defect.
Perf (advisory): NOT RUN — no multi-GPU host (single gfx942 here); the PR's 30.8 µs / 731 µs table carries no attached output, and no base-vs-head timing exists.

🔴 [verified] The INT6 mesh ladder's ST=8 rung over-launches the persistent kernel on gfx950. `MESH_ST_LADDER["int6"]` (aiter/ops/flydsl/kernels/quick_allreduce_mesh.py) caps both rungs at grid 1216, which is exactly 4 workgroups/CU on a 304-CU gfx942 — and 4/CU is INT6's hard LDS ceiling, since one workgroup allocates ATOMS × 1664 B = 13,312 B of LDS (4 × 13,312 = 53,248 ≤ 64 KB; 5 × 13,312 = 66,560 > 64 KB). On gfx950 — 256 CUs per this repo's own tuned configs (aiter/configs/a8w8_bpreshuffle_tuned_gemm.csv records cu_num 256 for gfx950; any gfx950 part below 304 CUs behaves the same) — the LDS-safe grid is 4 × 256 = 1024, but `clamp_grid_cap` passes 1216 through because `_RESIDENT_WGS_PER_CU[(8,8)]=6` and `[(4,8)]=5` are INT4-measured VGPR numbers (INT4's 9,216 B LDS allows 7/CU) and the new comment's claim that INT6 "runs at these counts on gfx942/gfx950" cannot hold: 6 × 13,312 = 79,872 B exceeds the 64 KB LDS on both arches. At gfx950 + INT6 mesh (AITER_ALL_REDUCE_CODEC=INT6 or rs_codec="int6") + TP4/TP8 + payload > 38 MiB, `_pick_st` takes the ST=8 rung and `_grid_x` launches 1,216 blocks (no release fence on xGMI, so no ceil(tiles/8) batching), of which at most 1,024 can be co-resident. Does this deadlock the bid-matched color handshake, or silently leave the 192 unstarted blocks' tiles unwritten — the "over-launching a persistent kernel deadlocks" failure the file's own comment names? The PR's own prefill-calib test shape (32768×5120 = 335 MiB) sits exactly on this rung. **Author must** key the residency clamp on the codec's LDS — cap the INT6 ST=8 rung at 4 × cu_count or add codec keying to `_RESIDENT_WGS_PER_CU` — and run the TP4/TP8 INT6 mesh cases on a gfx950 part to confirm which failure mode (or safety) actually obtains.

⚠️ [verified] The patch is stale against the merge target and the rename collides with new drift. applies.txt shows 4 files no longer applying to d231ca07, and the target's aiter/ops/flydsl/one_shot_allreduce.py now imports `has_xgmi_peer_links` from `.quick_allreduce_int4` (added by 1576c1ea2) — the exact module this PR deletes — alongside a new `link=` parameter the PR's head version of that file predates. A naive conflict resolution drops the import; `OneShotAllReduce.__init__` then NameErrors, and aiter/dist/device_communicators/flydsl_all_reduce.py's try/except converts that into `_IMPORT_OK = False`: the entire FlyDSL all-reduce path silently disables instead of failing loudly, and any CI on the current head describes a tree that has moved (the policy test and bench were also restructured on the target). **Author must** rebase onto d231ca07, re-point the `has_xgmi_peer_links` import to `.quick_allreduce`, and re-run the restructured policy/bench tests.

⚠️ [verified] The headline latency claims trace to no output. The PR's table (FlyDSL INT6 ~30.2 dB, 30.8 µs decode, 731 µs prefill vs HIP INT6 ~30.4 dB, 34.7/1139 µs) has no attached benchmark output, the Test Plan and Test Result sections are unfilled template comments, bench_comm_allreduce.py ships no mesh-INT6 candidate row (only `qr_int6` and ring `rs_codec="int6"` rows), and no base-vs-head timing exists. The SQNR half I could and did check: an independent numpy re-implementation of the group-16 E4M3 codec gives 30.23 dB for the INT6 mesh at the TP8 512×5120 decode-calib shape (and 19.21 dB for INT4, matching the ~19 dB claim), and the 24.08 dB uniform-512 edge-case arithmetic reproduces exactly — but the latency advantage, the PR's reason to exist, is not checkable by a reviewer as posted. **Author must** attach the sweep output (test_flydsl_quick_allreduce.py `--codec int6`, or a new `fly_int6` bench row) backing the µs numbers, fill the Test Plan/Test Result sections, and add the mesh-INT6 row so the FlyDSL-vs-HIP comparison is reproducible from the committed bench.

📝 [verified] The public rename ships no compatibility shim. `QuickAllReduceInt4` → `FlyQuickAllReduce` and `quick_allreduce_int4` → `quick_allreduce` remove the old names from aiter/ops/flydsl/__init__.py's lazy-export table and __all__, so `from aiter.ops.flydsl import QuickAllReduceInt4` (the spelling bench_comm_allreduce.py itself used before this PR) raises at import time. Mitigating: the name is 7 days old (56b244b29, 2026-09-09), unreleased (v0.1.21 is 125 commits back), and every in-tree reference is updated. **Reviewer should ask** whether any in-flight consumer branch (ATOM or a sibling feature branch) already imports the week-old name before the rename lands.

📝 [verified] Both INT6 rungs sit at grid 1216, so one TP8 engine set costs ~281 MiB of IPC inbox heap per rank (ST=8 engine 248 MiB + ST=1 engine 32 MiB + flags) against ~21.5 MiB for the entire INT4 ladder at cap 128 — the same full-grid footprint the host's own comment says INT4's cap-128 rungs were sized to avoid ("Seeding it with the 1216 default instead built a 194 MiB inbox … on every object ever constructed"). A sweep or bench holding several INT6 engine variants live at once multiplies this, and it is uncached IPC heap. **Reviewer should ask** whether the ST=8 rung needs the full 1216 cap now that the ST=1 rung covers the sub-38 MiB range, or whether ~13× INT4's inbox per engine set is the accepted cost.

PR #5558, opened 2026-09-15
[Triton/Gluon] Fix MoE routing kernel compile failure

micah-wil:
#5038 reworked the MoE routing helpers so that _expt_data_compute_stage1 computes the per-expert tile offset and returns it as a small tensor (built from a tl.gather / tl.zeros in an if/else), which is then passed into _expt_data_compute_stage2 and _expt_data_compute_stage2_fused. Those two helpers start with an early return:
```
    n_tokens = tl.load(Hist + expt_id)
    if n_tokens == 0:
        return
    TileInfo += tile_start   # uses the passed-in tensor after the return
```
This pattern makes the routing kernels fail to compile. Both `_combined_routing` (num_tokens > 16) and `_combined_routing_fused` (num_tokens <= 16) hit it, so any MXFP4 MoE run that goes through this routing path breaks. Reproduced on gfx942 and gfx950, with Triton 3.7.1 (the default in vLLM).
The compile aborts inside Triton's ConvertTritonToTritonGPU pass:
```
INFO 09-14 23:16:38 [mxfp4.py:1997] Using AiterW4A16ExpertsMonolithic
/usr/local/lib/python3.12/dist-packages/aiter/ops/triton/_triton_kernels/moe/moe_routing/expt_data.py:89:0: error: failed to legalize unresolved materialization from () to ('i32') that remained live after conversion
/usr/local/lib/python3.12/dist-packages/aiter/ops/triton/_triton_kernels/moe/moe_routing/expt_data.py:90:30: note: see existing live user here: %5 = "tt.addptr"(%2, %3) : (!tt.ptr<i32>, i32) -> !tt.ptr<i32>
    n_tokens = tl.load(Hist + expt_id)
    ...
RuntimeError: PassManager::run failed
```

(full error log: https://buildkite.com/vllm/amd-ci/builds/12929/list?jid=01a0a157-9fe0-4e1e-b3c7-46d8d32ce602&tab=output#L1425)

This appears to fundamentally be a Triton compiler bug.

Fix: replace the early return with a guarded `if n_tokens != 0:` block, and add into a fresh local (tile_info = TileInfo + tile_start) instead of reassigning the TileInfo pointer argument in place. This keeps the value from having to live across the early-return control flow, which is what the conversion pass couldn't handle. The computation and results are unchanged; the previous design that avoided this (loading the offset from a TileStart buffer) confirms the behavior is equivalent.

Repro (vLLM):
    `pytest -v -s tests/kernels/moe/test_ocp_mx_moe.py::test_rocm_mxfp4_moe_oracle[16-256-256-8-4-AITER_TRITON_MXFP4_BF16]`


vgokhale (commented):
LGTM - perf looks neutral. ASM is identical

vgokhale (dismissed):

vgokhale (approved):

PR #5560, opened 2026-09-15
[CI] Assign a PR to everyone who committed to it

Boss2002n:
A second step on the existing title and label workflow rather than a new one: it wants the same trigger, the same permissions and the same PR, and the header already offers no-auto-title as the way to opt a PR out entirely, so that label now covers this as well.

For each changed file the step reads the commit history and counts who wrote it, then assigns the heaviest contributors.


Boss2002n:
repository variable PR_ASSIGN_COMMITTERS is set to 'true' (Settings > Secrets and variables > Actions). Merging this changes nothing until then.

zufayu:
## ROCm/aiter PR #5560 — [CI] Assign a PR to everyone who committed to it

**Adds a second step to the existing PR title/label workflow that assigns a PR to its non-bot committers (ranked by commit count, capped at ten, honouring manual unassignment), inert unless the repository variable PR_ASSIGN_COMMITTERS is 'true'.**

⚠️ [verified] The new step in .github/workflows/pr-title-tags.yaml runs inside the shared pr-title-tags-<PR> concurrency group with cancel-in-progress: true, but its if-condition excludes labeled events — a maintainer adding any ci:* label while a synchronize run is executing the step cancels that run, and the replacement labeled run skips the step, so the assignment is silently lost until the next push. **Author must** move the step to its own concurrency group or add label events to its trigger so an interleaved label event cannot cancel an in-flight assignment.
⚠️ [verified] .github/workflows/pr-title-tags.yaml adds `contents: read` "for the PR's commit list", but every call the step makes — pulls.listCommits, issues.listEvents, the assignee probe, issues.addAssignees — is covered by the pull-requests/issues scopes the workflow already grants, so the extra scope buys nothing on a privileged pull_request_target job while the PR body still claims "the same permissions". **Author must** drop contents: read or name the call that needs it, and fix the description's same-permissions claim.
📝 [verified] The new step in .github/workflows/pr-title-tags.yaml widens the no-auto-title label to also suppress assignment, but CONTRIBUTE.md:426 and .github/workflows/pr-welcome-comment.yaml:46 still document it as opting out of automatic title tagging only, and PR_ASSIGN_COMMITTERS is documented nowhere in the repo — a contributor opting out of title automation unknowingly opts out of assignment too. **Author must** update both documents to state the label's widened meaning and the variable.
📝 [verified] The PR body says "For each changed file the step reads the commit history and counts who wrote it", but the new step in .github/workflows/pr-title-tags.yaml calls pulls.listCommits once for the PR and counts PR-commit authors — its own comment says "The commits on this PR, not the history of the files it touches" — so prior authors of touched files are never considered and a many-small-commits contributor outranks the main author. **Author must** correct the description to the implemented PR-commit weighting, or implement the per-file version.
📝 [verified] The new step in .github/workflows/pr-title-tags.yaml reads the no-auto-title check, the current assignees and the slot count from context.payload.pull_request — the event-time snapshot — so a label added or an assignment made in the queue/step-1 window before the script runs is invisible, and the script can assign users to a PR that was just opted out while computing slots against a stale assignee list. **Author must** refresh the PR with pulls.get before checking the label and computing slots.

Boss2002n:
> ## ROCm/aiter PR #5560 — [CI] Assign a PR to everyone who committed to it
> **Adds a second step to the existing PR title/label workflow that assigns a PR to its non-bot committers (ranked by commit count, capped at ten, honouring manual unassignment), inert unless the repository variable PR_ASSIGN_COMMITTERS is 'true'.**
> 
> Review (advisory): ⚠️ NEEDS WORK Validation (deterministic): N/A — no runtime surface changed Perf (advisory): N/A — no runtime surface changed
> 
> ⚠️ [verified] The new step in .github/workflows/pr-title-tags.yaml runs inside the shared pr-title-tags- concurrency group with cancel-in-progress: true, but its if-condition excludes labeled events — a maintainer adding any ci:* label while a synchronize run is executing the step cancels that run, and the replacement labeled run skips the step, so the assignment is silently lost until the next push. **Author must** move the step to its own concurrency group or add label events to its trigger so an interleaved label event cannot cancel an in-flight assignment. ⚠️ [verified] .github/workflows/pr-title-tags.yaml adds `contents: read` "for the PR's commit list", but every call the step makes — pulls.listCommits, issues.listEvents, the assignee probe, issues.addAssignees — is covered by the pull-requests/issues scopes the workflow already grants, so the extra scope buys nothing on a privileged pull_request_target job while the PR body still claims "the same permissions". **Author must** drop contents: read or name the call that needs it, and fix the description's same-permissions claim. 📝 [verified] The new step in .github/workflows/pr-title-tags.yaml widens the no-auto-title label to also suppress assignment, but CONTRIBUTE.md:426 and .github/workflows/pr-welcome-comment.yaml:46 still document it as opting out of automatic title tagging only, and PR_ASSIGN_COMMITTERS is documented nowhere in the repo — a contributor opting out of title automation unknowingly opts out of assignment too. **Author must** update both documents to state the label's widened meaning and the variable. 📝 [verified] The PR body says "For each changed file the step reads the commit history and counts who wrote it", but the new step in .github/workflows/pr-title-tags.yaml calls pulls.listCommits once for the PR and counts PR-commit authors — its own comment says "The commits on this PR, not the history of the files it touches" — so prior authors of touched files are never considered and a many-small-commits contributor outranks the main author. **Author must** correct the description to the implemented PR-commit weighting, or implement the per-file version. 📝 [verified] The new step in .github/workflows/pr-title-tags.yaml reads the no-auto-title check, the current assignees and the slot count from context.payload.pull_request — the event-time snapshot — so a label added or an assignment made in the queue/step-1 window before the script runs is invisible, and the script can assign users to a PR that was just opted out while computing slots against a stale assignee list. **Author must** refresh the PR with pulls.get before checking the label and computing slots.

addressed.

zufayu:
Thanks for the thoughtful proposal and for addressing the implementation feedback. Automated PR routing is a useful direction, and the event handling and manual-override ideas here are valuable.

After considering the overall design, though, I don’t think we should merge this PR. We already have a private agent-based assignment flow, and adding another controller that derives assignees from commit authorship would create two sources of truth for the same field. It also conflates contribution with ownership, cannot represent external or unresolved committers consistently, and may consume the limited assignee slots without identifying who is actually responsible for driving the PR.

I’d prefer to keep `assignee` as a clear accountability signal—normally one DRI—and integrate any useful ideas from this PR into the existing routing flow, with reviewers handled separately. Let’s close this for now.

Boss2002n:
> Thanks for the thoughtful proposal and for addressing the implementation feedback. Automated PR routing is a useful direction, and the event handling and manual-override ideas here are valuable.
> 
> After considering the overall design, though, I don’t think we should merge this PR. We already have a private agent-based assignment flow, and adding another controller that derives assignees from commit authorship would create two sources of truth for the same field. It also conflates contribution with ownership, cannot represent external or unresolved committers consistently, and may consume the limited assignee slots without identifying who is actually responsible for driving the PR.
> 
> I’d prefer to keep `assignee` as a clear accountability signal—normally one DRI—and integrate any useful ideas from this PR into the existing routing flow, with reviewers handled separately. Let’s close this for now.

okay. I have changed the PR to address this now we wont touch the reviewers field
image

It will only do the assignee field
image

zufayu:
Thanks, the sprint-tracking context helps. I think we can keep this much simpler.

In our process, the primary reviewer is normally also responsible for driving the PR. Our internal script already selects a primary reviewer and any additional reviewers. I can update it to set the primary reviewer as the assignee when the PR does not already have one.

That gives the sprint one clear owner without adding another workflow or assigning every committer. Let’s drop this PR for now.

Boss2002n:
There is an ongoing plan with the review assignments - as of now the request was to assign the PR/issue to the person responsible for it (that's what the assignee field is for)

I agree with this cuz - there is no way to know who is working on a PR/issue without the assignment 
adding the person reviewing doesn't add up to me for this purpose. 

Is there any other way you want us to achieve this?

PR #5561, opened 2026-09-15
[Bugfix] Drain FlyDSL stage-1 LDS-DMA loads before the tile barrier

kevin-mii:
## Motivation

Upstream the FlyDSL MoE stage-1 synchronization fix currently carried by sgl-project/sglang#39186 for gfx950 DSV4.1 a8w4 execution.

The software-pipelined K loop uses a tile-derived partial VMEM wait before a workgroup barrier. That estimate does not establish that every LDS-DMA load has completed, and scale-load scheduling can interleave with the DMA group. A consumer wave can therefore read an incompletely populated LDS tile.

## Technical Details

- Drain VMEM with `s_waitcnt(0)` before the barrier in the ordinary path. The heterogeneous async path already drains VMEM.
- Add a scheduling fence immediately after the next LDS prefetch group.
- Remove the now-unused partial-wait estimate and its helper constants.
- Add gfx950 coverage for t64x128, t64x256, and t128x128 stage-1 tiles: compare async loads against synchronous loads, poison/rewrite the output across 50 eager launches, and check 20 graph replays for exact repeatability.

This is based on current `main` at `0138f88b2`; it does not depend on SGLang code.

## Test Plan

```bash
PYTHONPATH=. HIP_VISIBLE_DEVICES=0 TRITON_HIP_USE_ASYNC_COPY=0 \
FLYDSL_RUNTIME_CACHE_DIR=/tmp/aiter-lds-fresh \
python -m pytest op_tests/test_flydsl_moe_lds_dma.py -q

black --check aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage_common.py op_tests/test_flydsl_moe_lds_dma.py
ruff check aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage_common.py op_tests/test_flydsl_moe_lds_dma.py
```

## Test Result

- MI350X VF, gfx950, ROCm 7.2.4, PyTorch 2.11.0+rocm7.2: **3 passed** with a fresh FlyDSL cache. Formatting and lint pass.
- The same test also passes on unpatched current main with a separate fresh cache. It exercises the synchronization paths but does **not** reproduce the previously reported intermittent failure on this compiler/device setup.
- `TRITON_HIP_USE_ASYNC_COPY=0` works around a local Triton failure in supporting kernels; the tested FlyDSL path explicitly uses `use_async_copy=True`.
- Not executed: full-model accuracy, other architectures, ROCm 10, or an isolated performance comparison of this additional wait. Keeping this PR in draft while the reproduction and performance impact are reviewed.

## Submission Checklist

- [x] Reviewed AITER contribution and title conventions.
- [x] Signed-off commits; no new dependencies.
- [ ] Reproduce the intermittent failure on an affected toolchain.
- [ ] Measure the additional wait across other affected MoE tile/dtype combinations.


zufayu:

**Replaces the tile-derived partial VMEM wait before the MoE stage-1 K-loop barrier with a full s_waitcnt(0) drain and adds a scheduling fence after the LDS-DMA prefetch group, so a consumer wave can no longer read an incompletely populated LDS tile.**

Validation (deterministic): NOT RUN — auto-validation disabled (REVIEW_AUTO_VALIDATE=0), and the single test target op_tests/test_flydsl_moe_lds_dma.py is gfx950-only while this box is gfx942, so the changed sync path cannot be executed here; verification is static analysis of the emitted wait/fence sequence only.
Perf (advisory): NOT RUN — the triaged target op_tests/test_flydsl_moe_lds_dma.py exposes no benchmark entry point and no gfx950 is available on this box; the PR ships no timing of its own and leaves "Measure the additional wait" unchecked.

⚠️ [inferred] The partial-VMEM-wait-before-barrier idiom that this PR's own new comment declares unsafe ("Partial VMEM waits can leave LDS-DMA loads crossing the barrier", aiter/ops/flydsl/kernels/mixed_moe_gemm_2stage_common.py:1530) remains live in sibling FlyDSL kernels on the same arch: aiter/ops/flydsl/kernels/preshuffle_gemm.py:561/590/625 (s_waitcnt(num_b_loads) before gpu.barrier with dma_a_to_lds in the same pipeline), aiter/ops/flydsl/kernels/gemm_a16w16_gfx950.py:1111/1135/1155 (wait_vmcnt_and_barrier(half_ldg_b_iters + half_ldg_a_iters) around cdna4 BufferLoadAsyncLDS), aiter/ops/flydsl/kernels/mega_moe/gemm1.py:154 (wait_lds_barrier(NUM_ACC_N*_PACK + NUM_B_SCALE) with a_gather.prefetch_to_lds), aiter/ops/flydsl/kernels/fmha_gfx950/pipeline.py:38 and hstu_attention_fwd.py:782. If instruction scheduling can interleave the counted loads into the DMA group in this kernel, the same mechanism can do it there — silent wrong output for those kernels' consumers. **Author must** either apply the same full drain to the sites that guard LDS-DMA consumption or state for each site why its counted group cannot be interleaved.

⚠️ [verified] The new test cannot reproduce the race it is named for: the author's own Test Result states it "also passes on unpatched current main" on their toolchain (MI350X VF, gfx950, ROCm 7.2.4), so the fix's behavioral effect is unvalidated on every toolchain where the race was actually reported (upstream sgl-project/sglang#39186) — the 50 eager launches and 20 graph replays only catch the race if it happens to fire during those runs. **Author must** reproduce the intermittent failure on the affected toolchain (or a stress configuration of it) before merge, or add a structural assertion on the compiled kernel (asm dump: LDS-DMA group contiguous and a full drain before each s_barrier) so the test can fail statically.

⚠️ [inferred] The full drain trades the partial wait's memory-latency overlap for certainty and nobody measured the price: per half-iteration the ~6-10 in-flight post-DMA B/scale VMEM loads (the old count for t64x128/t128x128/t64x256 a8w4 interleave) must now complete before the barrier, and the added sched_barrier(0) further constrains scheduling around the DMA group; no base-vs-head timing exists. The cost concentrates on the small-count configs — fp8-B tiles whose old count was >= 16 already wrapped the raw bitfield to vmcnt(count%16), a near-full drain, and see almost no delta. **Author must** measure base vs head on the affected a8w4 tile/dtype combinations on gfx950 before taking the PR out of draft.

📝 [inferred] The async-vs-sync oracle in op_tests/test_flydsl_moe_lds_dma.py carries rtol=atol=0.02, but both paths load identical quantized values through the same MFMA sequence, so bitwise equality is the expected result and the 2% slack could mask a small systematic async-path discrepancy (the repeat checks are already exact, rtol=0/atol=0). **Reviewer should ask** whether the async-vs-sync comparison also passes at rtol=0/atol=0, and if not, why.

PR #5571, opened 2026-09-16
[CK] Yzhou/fmoe runcfg gatemode verify

yzhou103:
## Submission Checklist


zufayu:
## [ROCm/aiter] PR #5571 — [CK] Yzhou/fmoe runcfg gatemode verify

**Centralizes the runtime dispatch knobs fused_moe's tuned-config lookup depends on (gate mode, w1 gate/up weight layout, SiTUv2 env selectors and betas) into one fmoe_runtime_policy shared by the tuner's --run_config benchmark and test_moe_2stage.py, so hand-built weights dispatch to the tuned config they were tuned for.**

Validation (deterministic): NOT RUN — triage found 2 candidate test targets (op_tests/test_moe_2stage.py, op_tests/tuning_tests/test_tune_pipeline.py) and no unique added one, so no target was named, and auto-validation is disabled (REVIEW_AUTO_VALIDATE=0) — environment gap, not a PR defect
Perf (advisory): NOT RUN — no unique perf target for the same reason (2 candidates, none named) and auto-validation disabled; the PR claims no numbers — environment gap, not a PR defect

⚠️ [verified] The merged reference branch in gemm_moe_tune.py claims "Mirrors op_tests/test_moe_2stage.py", but the test quantizes a8w4 reference activations to fp8 (per_1x32_f8_scale_f8_quant) while the tuner keeps bf16 and lets logits_diff absorb the quant error, so a8w4 verification (gptoss_fp8fp4, kimik3_a8w4 rows) is strictly weaker than the op-test it claims to mirror — an activation-quantization bug the test would catch can hide inside cos_tol=0.01. **Author must** quantize the a8w4 reference like the op-test does, or correct the comment and state why the looser reference is acceptable.
⚠️ [verified] fmoe_runtime_env in fused_moe.py applies policy.env additively and never clears the opposing selector, while fused_moe checks AITER_SITUV2_A8W4 before AITER_SITUV2_A4W4 — with AITER_SITUV2_A8W4=1 exported in the environment, every Situv2 a4w4 row dispatches as a8w4 against GGUU-shuffled weights and the benchmark fails mysteriously; test_moe_2stage.py already does the right thing (override_env with AITER_SITUV2_A4W4=None). **Author must** clear both SiTUv2 selectors inside fmoe_runtime_env, not just set the wanted one.
📝 [verified] The PR description is an unfilled template — Motivation, Technical Details, Test Plan and Test Result are all template comments — so the changed verification path ships no test evidence, and a single --run_config sweep over the merged tuned_fmoe CSVs would have exercised the a8w4-reference and env-selector paths above. **Reviewer should ask** for the output of op_tests/tuning_tests/test_run_config.py (fmoe family) and op_tests/tuning_tests/test_tune_pipeline.py before merge.

PR #5572, opened 2026-09-16
[Triton/Gluon] fix: avoid Python 3.10 source inspection failures in MXFP4 JIT imports

junna2016:
## Motivation

While integrating AITER into RTP-LLM, `import aiter` failed during MXFP4 kernel JIT initialization:

```text
ValueError: No function definition found for kernel
```

The import chain eagerly loads MXFP4 Gluon/Triton kernels, including gfx1250 kernels even on MI308. On the affected Python 3.10 environment, inline lambda expressions in heuristic decorators can cause `inspect.getsource()` to return a truncated source block without the subsequent function definition. Triton/Gluon JIT initialization then fails while locating the function definition.

## Technical Details

- Replace inline lambda heuristics with named predicates and `functools.partial` in the gfx1250 Gluon and Triton MXFP4 kernel modules.
- Cover all five affected decorated kernels, including the reduction/activation path whose alignment condition uses `BLOCK_SIZE_N1 * NUM_ITER`.
- Preserve the heuristic conditions, kernel signatures, and GPU kernel bodies.
- Do not patch Python's `inspect` module, alter Triton, suppress import errors, or change architecture dispatch.
- Add an import/source-inspection regression test covering all five kernels and 1,620 alignment predicate checks, including aligned/unaligned dimensions, distinct block sizes, and `NUM_ITER=1/3`.

## Test Plan

Run the import-compatibility regression test from the AITER repository root, using Python 3.10 with Triton/Gluon installed and without any source-inspection monkey patch:

```bash
env -u PYTHONPATH python3 -m unittest op_tests.test_mxfp4_jit_import -v

```

`Mxfp4JitImportTest.test_decorated_kernels_and_heuristics` verifies:

- Both affected kernel modules import successfully, exercising Triton/Gluon JIT initialization.
- `inspect.getsource()` includes the function definition for each of the five decorated kernels.
- All heuristic predicates match the original alignment conditions for aligned and unaligned dimensions and distinct block sizes.
- The middle kernel, `_fused_reduce_act_mul_and_dynamic_mxfp4_quant_kernel`, retains its `BLOCK_SIZE_N1 * NUM_ITER` alignment condition with `NUM_ITER=1` and `3`.

This is an import and host-side heuristic regression test; it does not require a gfx1250 GPU or launch GPU kernels.

## Test Result

- Source import and the regression test passed on Python **3.10.9** and **3.10.12** with Triton **3.8.0** in the ROCm development container.
- One unittest passed, explicitly covering all five decorated kernels and all 1,620 predicate comparisons; no GPU kernel launches are needed by this test.
- AST comparison against the base commit confirmed that all **11 existing function bodies** across the two kernel modules are unchanged.
- No gfx1250 hardware execution, cross-Python-3.11/3.12 validation, or kernel-performance benchmark is claimed. Existing unrelated `ResourceWarning` messages from AITER initialization were observed.

## Submission Checklist


Dewei-Wang-sh (dismissed):

zufayu:

**Replaces the inline-lambda heuristics of five MXFP4 quant kernels with partial-applied named predicates (verified semantically identical, kernel bodies untouched) so Python 3.10's inspect cannot truncate the decorated source and break `import aiter` at JIT initialization, plus a host-side regression test.**

Validation (deterministic): NOT RUN — REVIEW_AUTO_VALIDATE=0 (auto-validation disabled); the target op_tests/test_mxfp4_jit_import.py existed but no head-matched report was produced, and hand-run evidence below is advisory only
Perf (advisory): NOT RUN — perf stage never ran (auto-validation disabled); the triaged target op_tests/test_mxfp4_jit_import.py exposes no benchmark entry point and the repo's existing harness op_tests/op_benchmarks/triton/bench_quant_mxfp4.py was not run; GPU-side timing is unchanged by construction (kernel bodies AST-identical, heuristic values identical)

🔴 [verified] aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py and its gfx1250 Gluon twin rewire the heuristics of the five MXFP4 quant kernels behind aiter/ops/quant.py's fused_rms_mxfp4_quant / fused_reduce_act_mul_and_mxfp4_quant / fused_reduce_rms_mxfp4_quant — the activation-quant ops of the MXFP4 models the CI guide maps to ci:sglang (DeepSeek-R1-MXFP4) and ci:atom/ci:vllm (GPT-OSS-120B) — but the PR carries no ci:* label, so every downstream job is skipped and a quant-path break would pass aiter-only CI green and surface only after merge. **Reviewer should ask** for ci:all (or at least ci:sglang + ci:atom + ci:vllm) and require it green before merge.

⚠️ [verified] The claimed root cause does not reproduce on a clean Python 3.10: with stock Triton 3.8.0 on Python 3.10.12 (inspect.py byte-identical to upstream v3.10.12) both merge-target kernel modules import cleanly and inspect.getsource returns the full block for all five kernels on the pre-fix code, and the new test op_tests/test_mxfp4_jit_import.py passes on the pre-fix code on both 3.10.12+Triton 3.8.0 and 3.12+Triton 3.7.0 — while PR CI runs it only on the Python 3.12 image (the 3.10 matrix entry builds wheels on main only), so the test cannot catch the regression it targets in any environment CI runs; if the real trigger in the RTP-LLM container was something else (the test plan's own "without any source-inspection monkey patch" wording suggests one exists), this change may not cure the user's crash. **Author must** state the exact reproducing environment (container image, triton fork commit, any inspect patch) or wire a CI lane that actually reproduces the original failure.

⚠️ [verified] No base-vs-head timing exists for the changed launch path: Triton 3.8.0's Heuristics.run evaluates every heuristic value on each kernel launch, so the lambda→partial swap in aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py adds a Python call per launch on the MXFP4 quant hot path, and neither the PR nor the disabled perf stage measured it — the triaged target has no benchmark entry point and the existing op_tests/op_benchmarks/triton/bench_quant_mxfp4.py was not run; the GPU binary is unchanged by construction (bodies AST-identical, constexpr values identical), so only host-side launch overhead is unmeasured. **Author must** attach one base-vs-head run of bench_quant_mxfp4.py or state that the per-launch host delta is accepted.

📝 [verified] The alignment predicate _even_m_n now exists twice with diverging signatures — aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py:11 takes num_iter=None while the gfx1250 Gluon copy omits it — so a future predicate correction applied to one copy silently diverges the Triton and Gluon quant paths, and a wrong EVEN_M_N value selects the wrong masking branch with no error. **Author must** hoist the predicate into a shared util (e.g. aiter/ops/triton/utils) and import it from both kernel modules.

⚠️ [verified] op_tests/test_mxfp4_jit_import.py breaks the repo test convention — it uses unittest.TestCase (10 of 285 op_tests files; 145 use pytest) with a hand-rolled itertools.product((32,33,48), repeat=4) god-test, while its own sibling for these same MXFP4 quant kernels, op_tests/triton_tests/quant/test_fused_mxfp4_quant.py, is pytest + parametrize; a single nested test method cannot name which case failed, and its expected value re-derives the divisor with the same branch logic as _even_m_n, so a symmetric formula error passes both sides. **Author must** rewrite it as pytest + parametrize matching test_fused_mxfp4_quant.py and replace the computed expected with hardcoded input→truth pairs.

junna2016:
> 
> **Replaces the inline-lambda heuristics of five MXFP4 quant kernels with partial-applied named predicates (verified semantically identical, kernel bodies untouched) so Python 3.10's inspect cannot truncate the decorated source and break `import aiter` at JIT initialization, plus a host-side regression test.**
> 
> Review (advisory): 🔴 HIGH RISK Validation (deterministic): NOT RUN — REVIEW_AUTO_VALIDATE=0 (auto-validation disabled); the target op_tests/test_mxfp4_jit_import.py existed but no head-matched report was produced, and hand-run evidence below is advisory only Perf (advisory): NOT RUN — perf stage never ran (auto-validation disabled); the triaged target op_tests/test_mxfp4_jit_import.py exposes no benchmark entry point and the repo's existing harness op_tests/op_benchmarks/triton/bench_quant_mxfp4.py was not run; GPU-side timing is unchanged by construction (kernel bodies AST-identical, heuristic values identical)
> 
> 🔴 [verified] aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py and its gfx1250 Gluon twin rewire the heuristics of the five MXFP4 quant kernels behind aiter/ops/quant.py's fused_rms_mxfp4_quant / fused_reduce_act_mul_and_mxfp4_quant / fused_reduce_rms_mxfp4_quant — the activation-quant ops of the MXFP4 models the CI guide maps to ci:sglang (DeepSeek-R1-MXFP4) and ci:atom/ci:vllm (GPT-OSS-120B) — but the PR carries no ci:* label, so every downstream job is skipped and a quant-path break would pass aiter-only CI green and surface only after merge. **Reviewer should ask** for ci:all (or at least ci:sglang + ci:atom + ci:vllm) and require it green before merge.
> 
> ⚠️ [verified] The claimed root cause does not reproduce on a clean Python 3.10: with stock Triton 3.8.0 on Python 3.10.12 (inspect.py byte-identical to upstream v3.10.12) both merge-target kernel modules import cleanly and inspect.getsource returns the full block for all five kernels on the pre-fix code, and the new test op_tests/test_mxfp4_jit_import.py passes on the pre-fix code on both 3.10.12+Triton 3.8.0 and 3.12+Triton 3.7.0 — while PR CI runs it only on the Python 3.12 image (the 3.10 matrix entry builds wheels on main only), so the test cannot catch the regression it targets in any environment CI runs; if the real trigger in the RTP-LLM container was something else (the test plan's own "without any source-inspection monkey patch" wording suggests one exists), this change may not cure the user's crash. **Author must** state the exact reproducing environment (container image, triton fork commit, any inspect patch) or wire a CI lane that actually reproduces the original failure.
> 
> ⚠️ [verified] No base-vs-head timing exists for the changed launch path: Triton 3.8.0's Heuristics.run evaluates every heuristic value on each kernel launch, so the lambda→partial swap in aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py adds a Python call per launch on the MXFP4 quant hot path, and neither the PR nor the disabled perf stage measured it — the triaged target has no benchmark entry point and the existing op_tests/op_benchmarks/triton/bench_quant_mxfp4.py was not run; the GPU binary is unchanged by construction (bodies AST-identical, constexpr values identical), so only host-side launch overhead is unmeasured. **Author must** attach one base-vs-head run of bench_quant_mxfp4.py or state that the per-launch host delta is accepted.
> 
> 📝 [verified] The alignment predicate _even_m_n now exists twice with diverging signatures — aiter/ops/triton/_triton_kernels/quant/fused_mxfp4_quant.py:11 takes num_iter=None while the gfx1250 Gluon copy omits it — so a future predicate correction applied to one copy silently diverges the Triton and Gluon quant paths, and a wrong EVEN_M_N value selects the wrong masking branch with no error. **Author must** hoist the predicate into a shared util (e.g. aiter/ops/triton/utils) and import it from both kernel modules.
> 
> ⚠️ [verified] op_tests/test_mxfp4_jit_import.py breaks the repo test convention — it uses unittest.TestCase (10 of 285 op_tests files; 145 use pytest) with a hand-rolled itertools.product((32,33,48), repeat=4) god-test, while its own sibling for these same MXFP4 quant kernels, op_tests/triton_tests/quant/test_fused_mxfp4_quant.py, is pytest + parametrize; a single nested test method cannot name which case failed, and its expected value re-derives the divisor with the same branch logic as _even_m_n, so a symmetric formula error passes both sides. **Author must** rewrite it as pytest + parametrize matching test_fused_mxfp4_quant.py and replace the computed expected with hardcoded input→truth pairs.

Thanks for the detailed review. Addressed in latest commit.
Root cause: You are correct that the original code imports successfully on Python 3.10.12. Our failing RTP-LLM Bazel environment actually selects CPython 3.10.9. Its inspect.BlockFinder exhibits the nested-parentheses decorator parsing issue tracked in CPython gh-83035. We corrected the previously overbroad “Python 3.10” description.

We verified both interpreters’ inspect.py files against the official CPython release files. With the same AMD Triton 3.8.0+amd.rocm7.2.0.git111ff227, the original AITER source fails on 3.10.9 and passes on 3.10.12; the fixed source passes on both. A pure-stdlib reproducer also demonstrates the issue under python -I -S, without AITER, Triton, or any source-inspection monkey patch. Exact environment details and reproduction steps are documented in docs/mxfp4_jit_import_compatibility.md.

Code and tests: Both kernel modules now share one alignment helper. The regression tests use pytest parameterization with hardcoded input-to-boolean expectations. An additional AST compatibility check catches reintroduced inline lambdas even on newer Python versions. All 60 tests pass on both 3.10.9 and 3.10.12.

Performance: We no longer claim zero host overhead. A paired host-only benchmark of actual Triton Heuristics.run, with a no-op downstream launch, measured approximately +0.31–1.01 µs per call across the five kernels. This is not GPU or end-to-end timing; the GPU kernel bodies remain unchanged.

Downstream CI validation remains outstanding. Could a maintainer enable ci:all, or at least ci:sglang, ci:atom, and ci:vllm, before merge?

zufayu:
add  more ci test to verify

Dewei-Wang-sh (dismissed):

zufayu:
coud we del or merged test_mxfp4_jit_import.py to op_tests/triton_tests/quant/

junna2016:
> coud we del or merged test_mxfp4_jit_import.py to op_tests/triton_tests/quant/

Done

zufayu (dismissed):

zufayu:
[Triton Test / Triton Tests (MI35X) / Shard 7 (pull_request)](https://github.com/ROCm/aiter/actions/runs/35506677321/job/106099535906?pr=5572)
Triton Test / Triton Tests (MI35X) / Shard 7 (pull_request)Failing after 35m

zufayu:
@vgokhale could we merged it without all ci passed?

valarLip (approved):

PR #5574, opened 2026-09-16
[Triton/Gluon] Create benchmark for `fused_add_rmsnorm_pad.py`

nidal567:
## Motivation

`fused_add_rmsnorm_pad` ships a gfx942/gfx950 Triton kernel and a gfx1250 Gluon kernel but has no benchmark, so there is no way to put a metric on the numbers without a driver. Therefore, a benchmark file must be made.

The op is also not covered by the model benchmarking tool because it is absent from `model_shapes.json` and `KERNEL_DICT` (commented out in `bench_models.py`), so it cannot be reached through `bench_models.py`.

## Technical Details

Shape-driven benchmark following a neighbouring kernels `bench_rmsnorm.py` layout:
- same `get_x_vals`/`run_benchmark`/`parse_args`/`main` structure, 
- same `triton.testing.Benchmark` + `perf_report` + `do_bench(warmup=25, rep=100)`, and, 
- it reuses the tester's `generate_inputs` rather than duplicating input construction.

Organized around what the kernel actually specializes on. The Triton kernel has exactly two constexprs, `HAS_RES` and `BLOCK_SIZE_N`, and launches one program per row, so the sweep varies `M`, `N`, `--add-residual` and `--pad-to-multiple`, where,

```
N_out        = cdiv(N, pad_to_multiple) * pad_to_multiple   (else N)
BLOCK_SIZE_N = next_power_of_2(N_out)
```

15 default shapes, each grouped differently:

| category | shapes | source |
|---|---|---|
| M scaling at a fixed compiled variant | (1 … 16384, 2880) | `BLOCK_SIZE_N` held constant |
| tester widths, different `BLOCK_SIZE_N` buckets | (8192, 4/16/320/640) | `test_fused_add_rmsnorm_pad.py::parametrize("N", …)` |
| production width | N = 2880 | in tester list, but also GPT-OSS 120B's hidden size in `model_shapes.json` |
| aligned powers of two (without padding) | (8192, 512/1024/2048/4096) | isolates the norm from the pad |

#### Memory accounting
reflects the padding asymmetry. Reads are `N` wide, the write is `N_out` wide; `HAS_RES` adds one read and one write. `n_out` is computed exactly as the wrapper computes it. FLOPs are computed as: `flops = 4 * M * N + (M * N if add_residual else 0)`

#### Backend pinned to Triton
The wrapper defaults to Gluon on gfx1250 and Triton elsewhere, so an unpinned benchmark would measure a different kernel per architecture. Pinning is a **no-op** on **gfx950/gfx942** which I verified by kernel trace, `backend=None` and `backend="triton"` dispatch the identical kernel and takes effect only on gfx1250, where it keeps the numbers comparable with every other target.

No model mode: `--model`/`-M`/`--model-configs` are deliberately absent rather than inherited, since nothing associates this op with a model. A `model_benchmarking_shapes` stub is left as a TODO for if there requires future support for this kernel.

## Test Plan

- Every CLI path exercised: default sweep, `--shape`, all metrics and dtypes, `--add-residual`, `--pad-to-multiple`, `-o`
- Kernel trace to confirm the pinned backend dispatches `_fused_add_rmsnorm_pad` and never the Gluon kernel
- `rocprofv3 --kernel-trace` kernel time
- `black` and `ruff` checks, as well as `pytest` check

## Test Result

Default sweep, bandwidth, gfx950:

```
       M       N   GB/s              M       N   GB/s
       1    2880    2.6           8192       4   14.0
      32    2880   56.9           8192      16   57.7
     256    2880  445.7           8192     320 1056.2
    1024    2880 1677.0           8192     640 1902.7
    4096    2880 3712.9           8192     512 1612.6
    8192    2880 4590.6           8192    1024 2755.4
   16384    2880 5174.2           8192    2048 4199.9
                                  8192    4096 5101.1
```

Cross-check against `rocprofv3` kernel timing at `M=8192 N=2880`: benchmark 21.6 µs vs kernel 18.9 µs. The *2.7* µs gap is the wrapper's per-call `torch.empty` allocations, which `do_bench` includes and a kernel trace does not.
**note**: This is no longer the case. This happens with triton's default `do_bench` which clears L2 cache before starting the kernel event. Every event starts with a cleared cache, and this runtime entry is what causes the slight gap we see.

A larger gap (6.5 µs) with `--add-residual`, which allocates two tensors.

Backend pin verified as a no-op on gfx950: `backend=None` and `backend="triton"` both dispatch `_fused_add_rmsnorm_pad`, 3 dispatches each.

### Regression this makes measurable

Not addressed by this PR. This will be filed separately as a compiler issue with ATT traces.

| shape | Triton 3.4 | Triton 3.8 | Δ |
|---|---|---|---|
| M=8192 N=2880 | 17.559 µs | 18.863 µs | **+7.4 %** |
| M=8192 N=2880, pad=256 | 18.889 µs | 19.389 µs | **+2.6 %** |

`rocprofv3 --kernel-trace`, 250 iters + 25 warmup, with single-average number (10% off head and 10% off tail, average the middle 80%). This is not reachable by any launch parameter - `num_warps`, `waves_per_eu` and `BLOCK_SIZE_N` were all tested and every candidate was worse; `num_stages` is a no-op on this kernel.

#### Correction test
pytest - `840 passed, 420 skipped, 109 warnings in 8.82s`

## Submission Checklist


op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
```diff
@@ -0,0 +1,180 @@
...
+from op_tests.op_benchmarks.triton.utils.benchmark_utils import (
+    get_caller_name_no_ext,
+    print_vgpr,
+)
+from op_tests.triton_tests.normalization.test_fused_add_rmsnorm_pad import (
+    generate_inputs,
+)
+
+
+def model_benchmarking_shapes(args):
```
brunomazzottiamd: Why define this function? It's a brandy new source file and there's no caller in it. What are the supposed callers?

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
```diff
@@ -0,0 +1,180 @@
...
+        x_vals_list = [list(args.shape)]
+    else:
+        x_vals_list = [list(shape) for shape in get_x_vals()]
+
+    if args.metric == "time":
+        ylabel = "Time_(ms)"
+    elif args.metric == "bandwidth":
+        ylabel = "Bandwidth_(GB/s)"
+    elif args.metric == "throughput":
+        ylabel = "Throughput_(TFLOPS)"
```
brunomazzottiamd, line 52: Why compute TFLOPS for a memory bound kernel? Is this useful?

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
```diff
@@ -0,0 +1,180 @@
...
+        mem_read = M * N * es + N * es
+        mem_write = M * n_out * es
+        if add_residual:
+            mem_read += M * N * es
+            mem_write += M * N * es
+        mem = mem_read + mem_write
+
+        flops = 4 * M * N + (M * N if add_residual else 0)
+
+        ms = triton.testing.do_bench(fn, warmup=25, rep=100)
```
brunomazzottiamd, line 104: Please double check with @azaidy, but I think we should rely on `do_bench_cudagraph` from now on.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
```diff
@@ -0,0 +1,180 @@
...
+        line_arg="unit",
+        line_vals=line_vals,
+        line_names=line_names,
+        styles=[("green", "-")],
+        ylabel=ylabel,
+        plot_name=get_caller_name_no_ext(),
+        args={"metric": args.metric},
+    )
+
+    # Gluon implementation is gfx1250, we pin Triton on backend here otherwise
```
brunomazzottiamd: Why do we need to pin the backend to `"triton"`? I think this benchmark script can be used with the Gluon kernel too! It's useful!

I'd add a `--backend` CLI argument that may assume the following values: `None` (the default value), `"triton"` and `"gluon"`. If it isn't specified (`None`) then the arch default will be selected. If you pass `"gluon"` in archs other than `gfx1250` then the wrapper will do this:
```python
    if backend == "gluon":
        if get_arch() == "gfx1250":
            # omitted for brevity - launches Gluon kernel
        else:
            _LOGGER.warning(
                "Gluon kernel is only supported on gfx1250, using triton kernel instead"
            )
            backend = "triton"
```

I think this design is perfectly fine.

brunomazzottiamd (commented):
I'm sharing some comments. They are open questions and small suggestions, no blockers at all! Address them at your own judgment, I can approve it tomorrow once all discussion are resolved for good.

brunomazzottiamd:
@nidal567, we're trying to fix the error from **Triton Test / Triton Tests (MI300X) / Shard 5** in https://github.com/ROCm/aiter/pull/5594 and / or https://github.com/ROCm/aiter/pull/5590 (PR 5590 probably is the best way to implement the fix).

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
nidal567: The intent was to keep it as a TODO placeholder. Other kernel benchmarks have these for the "--model" flags, but for this one in particular, it doesn't have a place in the model_benchmarking_tool or `KERNEL_DICT` in `aiter/op_tests/op_benchmarks/triton/model_benchmarking_tool/bench_models.py`.

I can remove this since this is essentially noise.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
nidal567: Every `--metric`-carrying benchmark in the Normalization kernels family offers the same metric of TFLOPs as an option. If we remove this one, it would be considered the "odd one out". I used the family `rmsnorm` as a blueprint when creating this benchmark. That's why I kept it there, but I do understand it doesn't offer much value here

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
brunomazzottiamd: Please remove it.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
brunomazzottiamd: OK, let's keep TFLOPS throughput metric in order to maintain code base uniformity. Please consider this discussion resolved.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
nidal567: In the latest commit, it is removed :)

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
brunomazzottiamd: Implemented by last pushed commit.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
brunomazzottiamd: Implemented by last pushed commit.

brunomazzottiamd:
> @nidal567, we're trying to fix the error from **Triton Test / Triton Tests (MI300X) / Shard 5** in #5594 and / or #5590 (PR 5590 probably is the best way to implement the fix).

@nidal567, this issue was already resolved by https://github.com/ROCm/aiter/pull/5590. Please rebase on top of `main` before your next push.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
brunomazzottiamd: We should use `do_bench` for now. Our team will figure out a proper methodology later with CUDA graph and rotating buffers.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
```diff
@@ -0,0 +1,183 @@
...
+    c_dtype = str_to_torch_dtype[args.dtype]
+
+    @triton.testing.perf_report([benchmark])
+    def bench_fused_add_rmsnorm_pad(M, N, metric, **kwargs):
+        x, weight, res = generate_inputs(M, N, add_residual, c_dtype)
+        eps = 1e-6
+
+        fn = lambda: fused_add_rmsnorm_pad(
+            x, weight, eps, res, pad_to_multiple, backend
+        )
```
brunomazzottiamd, line 87: **Suggestion:** Use named optional arguments here, for the sake of readability.

```python
fn = lambda: fused_add_rmsnorm_pad(
    x,
    weight,
    eps,
    res=res,
    x_pad_to_multiple=pad_to_multiple,
    backend=backend,
)
```

brunomazzottiamd (commented):
I have one final suggestion. You mention that `torch.empty` calls can take 2.7 µs or 6.5 µs, it's a considerable amount of time compared to 18.9 µs from the kernel itself.

The current wrapper function has this signature:

```python
def fused_add_rmsnorm_pad(
    x: torch.Tensor,
    weight: torch.Tensor,
    epsilon: float,
    res: torch.Tensor = None,
    x_pad_to_multiple: int = 0,
    backend: str | None = None,
):
```

We can pass pre-allocated output tensors to the wrapper without introducing API breaking changes:

```python
def fused_add_rmsnorm_pad(
    x: torch.Tensor,
    weight: torch.Tensor,
    epsilon: float,
    res: torch.Tensor = None,  # <- wrong type annotation here, should be torch.Tensor | None
    x_pad_to_multiple: int = 0,
    backend: str | None = None,
    out: torch.Tensor | None = None,
    res_out: torch.Tensor | None = None,
):
```

And conditionally allocate output tensors:

```python
if out is None:
    out = torch.empty((M, N_out), dtype=x.dtype, device=x.device)
else
    assert out.shape == (M, N_out), f"Expecting pre-allocated out with shape {(M, N_out)}, got {tuple(out.shape)}."
    assert out.dtype == x.dtype, f"Expecting pre-allocated out with data type {x.dtype)}, got {out.dtype}."
    assert out.device == x.device,, f"Expecting pre-allocated out on {x.device} device, got {out.device}."
    
...

if res_out is None:
    res_out = torch.empty((M, N), dtype=res.dtype, device=res.device)
else:
    # do the respective assertions, ommited for brevity...
```

Finally you can pre-allocate the output tensors in `bench_fused_add_rmsnorm_pad` and pass they along to `fused_add_rmsnorm_pad`, to get rid of `torch.empty` overhead and decrease the gap to `rocprofv3`.

op_tests/op_benchmarks/triton/bench_fused_add_rmsnorm_pad.py
nidal567: This change has been reflected!

brunomazzottiamd (approved):
LGTM!

brunomazzottiamd:
> I have one final suggestion. You mention that `torch.empty` calls can take 2.7 µs or 6.5 µs, it's a considerable amount of time compared to 18.9 µs from the kernel itself.
> 
> The current wrapper function has this signature:
> 
> ```python
> def fused_add_rmsnorm_pad(
>     x: torch.Tensor,
>     weight: torch.Tensor,
>     epsilon: float,
>     res: torch.Tensor = None,
>     x_pad_to_multiple: int = 0,
>     backend: str | None = None,
> ):
> ```
> 
> We can pass pre-allocated output tensors to the wrapper without introducing API breaking changes:
> 
> ```python
> def fused_add_rmsnorm_pad(
>     x: torch.Tensor,
>     weight: torch.Tensor,
>     epsilon: float,
>     res: torch.Tensor = None,  # <- wrong type annotation here, should be torch.Tensor | None
>     x_pad_to_multiple: int = 0,
>     backend: str | None = None,
>     out: torch.Tensor | None = None,
>     res_out: torch.Tensor | None = None,
> ):
> ```
> 
> And conditionally allocate output tensors:
> 
> ```python
> if out is None:
>     out = torch.empty((M, N_out), dtype=x.dtype, device=x.device)
> else
>     assert out.shape == (M, N_out), f"Expecting pre-allocated out with shape {(M, N_out)}, got {tuple(out.shape)}."
>     assert out.dtype == x.dtype, f"Expecting pre-allocated out with data type {x.dtype)}, got {out.dtype}."
>     assert out.device == x.device,, f"Expecting pre-allocated out on {x.device} device, got {out.device}."
>     
> ...
> 
> if res_out is None:
>     res_out = torch.empty((M, N), dtype=res.dtype, device=res.device)
> else:
>     # do the respective assertions, ommited for brevity...
> ```
> 
> Finally you can pre-allocate the output tensors in `bench_fused_add_rmsnorm_pad` and pass they along to `fused_add_rmsnorm_pad`, to get rid of `torch.empty` overhead and decrease the gap to `rocprofv3`.

Empirical investigations proved that the overhead w.r.t. `rocprofv3` isn't caused by calls to `torch.empty`.

PR #5576, opened 2026-09-16
[HIP] Fix a missing block_reduce barrier and rewrite indexer_qk_rope_quant_and_cache around the wave

XiaobingSuper:
Two commits against one file, `csrc/kernels/cache_kernels.cu` (+376 / −151).

## `#1` Fix a missing barrier between `block_reduce` rounds

`block_reduce` stages its cross-wave partials through one block-scoped
`__shared__` buffer and its only barrier sits *after* that write, so two calls of
the same instantiation have to be separated by the caller. Three call sites did
not do that:

- `reshape_and_cache_with_block_quant_kernel`, K amax then V amax
- `reshape_and_cache_with_block_quant_kernel_for_asmpa`, same pair
- `indexer_qk_rope_quant_and_cache_kernel`, the K layernorm mean then variance

In the indexer this silently corrupts `kv_cache`: a raced variance gives a wrong
`inv_std`, so the stored K row does not match its input. Nothing asserts, and it
is occupancy-dependent — it does not reproduce at small launches, which is why
the existing tests never caught it.

One `__syncthreads()` per site, 9 lines total. Cost is within noise (every point
moves less than 0.3%). This commit can be cherry-picked on its own.

## `#2` Rewrite `indexer_qk_rope_quant_and_cache` around the wave

The old kernel launched one 128-thread block per (token, head) and moved 2 bytes
per lane, with a shared-memory round trip and four `__syncthreads()` per block.

- **One wave per block**, `HEAD_DIM / THREADS` dims per lane, so every reduction
  is wave-level and the reduction staging buffer and its barriers are gone.
- **Vectorized q loads**: a lane reads 16 bytes of a q row in one `dwordx4`,
  16 lanes cover a head row, a wave covers 4. FP4 packing loses its lane shuffle
  as a side effect, because both nibbles of an output byte now live in one lane.
- **RoPE pairs exchange in-wave** on the wide instantiation. The narrow one keeps
  the shared-memory exchange, because lane shuffles are the more expensive of the
  two when the launch is small.
- The lane width is a template parameter chosen at launch from `num_tokens` —
  one code path, two instantiations, nothing inspected at run time.

Four invariants that were previously only prose are now `static_assert`s:
`THREADS == ROPE_DIM`, `ROPE_DIM % VEC == 0`,
`!SHARED_ROPE || HEADS_PER_BLOCK == 1`, and
`LANES_PER_HEAD % LANES_PER_GROUP == 0`. Each one guards a silent failure that
would follow from changing a constant.

### Prefill, medians of 3 independent processes

`n_heads=32`

| tokens | fp8 before | after | speedup | fp4 before | after | speedup |
|---|---|---|---|---|---|---|
| 2048 | 33.51 us | 23.88 | 1.40x | 36.31 | 27.66 | 1.31x |
| 2560 | 41.92 | 24.12 | 1.74x | 44.63 | 27.68 | 1.61x |
| 3072 | 52.02 | 24.09 | 2.16x | 53.89 | 27.67 | 1.95x |
| 4096 | 69.99 | 28.70 | 2.44x | 74.54 | 27.78 | 2.68x |
| 5120 | 86.58 | 34.68 | 2.50x | 92.38 | 30.69 | 3.01x |
| 6144 | 102.96 | 40.27 | 2.56x | 110.03 | 35.91 | 3.06x |
| 8192 | 136.48 | 51.84 | 2.63x | 145.53 | 46.98 | 3.10x |
| 10240 | 169.68 | 64.84 | 2.62x | 181.71 | 56.49 | 3.22x |
| 12288 | 203.48 | 77.58 | 2.62x | 217.81 | 67.75 | 3.22x |
| 16384 | 273.39 | 99.97 | **2.73x** | 291.93 | 89.08 | **3.28x** |

`n_heads=64`

| tokens | fp8 before | after | speedup | fp4 before | after | speedup |
|---|---|---|---|---|---|---|
| 2048 | 69.16 us | 28.36 | 2.44x | 71.75 | 27.76 | 2.58x |
| 4096 | 134.20 | 51.63 | 2.60x | 140.21 | 44.78 | 3.13x |
| 8192 | 265.50 | 97.82 | 2.71x | 277.99 | 84.80 | 3.28x |
| 12288 | 462.22 | 158.38 | 2.92x | 417.50 | 126.41 | 3.30x |
| 16384 | 628.86 | 213.00 | **2.95x** | 650.69 | 191.51 | **3.40x** |

Below about 4096 tokens at `n_heads=32` the per-launch cost has not amortized and
the kernel is not memory-bound, which is why 2048 shows 1.40x and 16384 shows
2.73x. Baselines were measured in the same batch, on the same node, with the same
protocol as the after column.

### Decode, 1-64 tokens

No regression: all 24 configurations (both modes x Q-only and KV-append x
1/8/32/64 tokens) land between **1.01x and 1.16x**, measured under graph replay.

### Where the width switch changes over

Both instantiations forced, fp8, medians of 3 processes:

| tokens | narrow | wide | wide/narrow | selected |
|---|---|---|---|---|
| 64 | 3.595 us | 4.023 | 1.119x | narrow |
| 128 | 3.716 | 4.372 | 1.177x | narrow |
| 256 | 4.462 | 4.755 | 1.066x | narrow |
| 512 | 6.477 | 5.837 | 0.901x | wide |
| 1024 | 10.087 | 8.822 | 0.875x | wide |
| 2048 | 24.530 | 24.082 | 0.982x | wide |

Narrow wins at every size at or below 256 and wide at every size above it, so the
threshold is 256. Re-measured after the decode fix, since that made the narrow
instantiation faster and could have moved the crossover.

## Output equivalence

384 output tensors compared byte for byte over the full ABI surface, three
repeats, covering both instantiations, with a control confirming the reference is
deterministic.

- `q_out`, `q_scale_out`, `weights_out`: **0 differences**.
- **All FP4 output byte-identical**, including `kv_cache` and `kv_cache_scale`.
- fp8 `kv_cache`: one byte in 4 of 384 tensors, only at 16384 x 64, `14 -> 15` in
  e4m3 = 1 ULP. One K element in 2.1 M, on dim 100, which never rotates.

Four candidate causes for that byte were each tested by building and counting
bytes, and each is excluded: sum association in both directions, and FMA
contraction in the layernorm, in the K RoPE line and in the store multiply.

It changes nothing the indexer selects. Both `kv_cache` versions were
dequantized, the MQA logits formed, and the selected index sets compared:
**0 of 512 queries differ** at k = 64, 256, 512, 1024 and 2048, across all four
affected configurations.

## Testing

`op_tests/test_indexer_qk_rope_quant_and_cache.py --mode fp8 fp4 e2e` passes on
gfx950 for both instantiations. No test files are modified by this PR.


zufayu:

**Inserts the missing __syncthreads between the back-to-back K/V amax block_reduce rounds in the two reshape_and_cache block-quant kernels, and rewrites indexer_qk_rope_quant_and_cache from a 128-thread block per (token, head) into a single-wave 64-thread block with vectorized q loads, in-wave RoPE partner exchange, and a launch-time narrow/wide instantiation switch at 256 tokens.**

Validation (deterministic): NOT RUN — triage found no test target shipped by the PR (runtime kernel change, no test file in the diff); auto-validation disabled (REVIEW_AUTO_VALIDATE=0)
Perf (advisory): NOT RUN — no perf target triaged for the same reason (PR ships no test target); the PR's own tables (prefill 1.31–3.40x, decode 1.01–1.16x, medians of 3 processes) are the only measurement and were not re-run

⚠️ [verified] The narrow/wide launch switch at `csrc/kernels/cache_kernels.cu:4630` (`INDEXER_NARROW_MAX_TOKENS = 256`) routes every call with num_tokens > 256 to the wide instantiation — in-wave `__shfl_xor` RoPE, 4 heads per block, 16-byte q loads — but `op_tests/test_indexer_qk_rope_quant_and_cache.py` defaults to num_tokens [8, 32] and CI invokes it with no arguments (`.github/scripts/aiter_test.sh` builds `test_cmd=(timeout 60m python3 "$file")`), so no CI leg ever executes the wide path that every prefill speedup in the PR (1.31–3.40x) comes from. A lane-mapping error in the wide path would silently corrupt q_out/kv_cache for every prefill above 256 tokens with a green CI. **Author must** extend the test's default `-n` sweep past 256 (e.g. add 512 and 4096) so both instantiations run in CI.

⚠️ [verified] The new `static_assert(THREADS <= WARP_SIZE)` at `csrc/kernels/cache_kernels.cu:1636` hard-fails on wave32 archs: `WARP_SIZE` is 32 wherever `__GFX9__` is undefined (`csrc/include/aiter_hip_common.h:497`), gfx1250 is an allowed JIT arch (`aiter/jit/core.py:604`), and module_cache compiles cache_kernels.cu with no arch gate (`aiter/jit/optCompilerConfig.json`) — so a gfx1250 build of module_cache now fails to compile, taking down every cache op (reshape_and_cache, concat_and_cache_mla, …), where the old 128-thread kernel compiled (block_reduce waves=4 path). The compile failure is inferred from the macro definitions — this box's ROCm 7.0.0 rejects the gfx1250 target ID, so it could not be built here. **Reviewer should ask** whether gfx1250 builds of module_cache matter; if they do, gate the wave-level kernel to GFX9 and keep a wave32 fallback.

⚠️ [verified] The rewritten kernel in `csrc/kernels/cache_kernels.cu` (`indexer_qk_rope_quant_and_cache_kernel` and its `INDEXER_NARROW_MAX_TOKENS = 256` launch switch) has no base-vs-head timing independent of the author: auto-validation is disabled and the PR ships no test target, so the prefill, decode, and crossover tables are author-measured medians of 3 with no re-run, and the 256-token crossover constant rests on them. **Author must** run the existing test's perftest harness (`op_tests/test_indexer_qk_rope_quant_and_cache.py` carries one) on base and head back to back on one GPU, or have a maintainer do so, before the crossover and the decode no-regression claim are trusted.

📝 [verified] The barrier fix ships no test that can fail without it: the race is occupancy-dependent — the PR body itself states it does not reproduce at small launches, which is why the existing tests never caught it — so the current suite passes with or without the `__syncthreads`. The fix itself — the `__syncthreads` between the K and V amax rounds in `csrc/kernels/cache_kernels.cu` — is correct by construction (block_reduce's only barrier sits between its smem write and read, `csrc/include/hip_reduce.h:394-418`; an ISA-level check of the two-call pattern on gfx942 confirms no `s_barrier` orders round 1's read against round 2's write without it), but nothing in the repo would catch a regression of it. **Author must** add a repeated K/V-amax comparison for reshape_and_cache_with_block_quant at high occupancy, or state why a deterministic test of the race is not feasible.

📝 [verified] The rewrite narrows accepted q inputs: the old kernel honored `q_stride_d` (any stride along head_dim), while the new launcher rejects q with `stride(2) != 1`, strides not multiples of the vector width, or an unaligned pointer (`csrc/kernels/cache_kernels.cu:4625-4641`). The failure is loud (AITER_CHECK), but a downstream caller passing a dim-strided or sliced q — no in-repo caller exists to check, consumers are SGLang/ATOM-side — would newly throw where it previously worked. **Reviewer should ask** whether any downstream consumer passes a non-contiguous q; if none does, the check is pure hardening.

XiaobingSuper:
Thanks — two of these turned into real fixes, pushed as `c64343039`.

**Wide instantiation never ran in CI** — fixed. The test's default `-n` now includes 512, so the fp8 sweep goes from 16 to 24 configs with 8 on the wide path, on both gfx942 and gfx950. You were right that every prefill number came from a path CI never touched.

**wave32 build break** — fixed and verified on real gfx1250 hardware (ROCm 10.1.0, `warpSize` 32). Unpatched reproduces exactly what you predicted: `static assertion failed due to requirement '64 <= WARP_SIZE'` → `Error building extension 'module_cache'`, no `.so` at all. Patched builds in 11.5 s and passes 32 of 32 correctness configs, 16 of them on the wide instantiation.

One trap worth passing on, since it would catch anyone doing this generalization: making `THREADS = WARP_SIZE` a *template argument* compiles cleanly and then fails at first launch. `WARP_SIZE`'s `constexpr operator int()` answers 64 to any constant-evaluated read, including the host pass of a wave32 build, so the host registers `probe_kernel<64>` while the device binary only has `probe_kernel<32>` — `Cannot find Symbol with name: _Z12probe_kernelILi64EEvPi`. The wave size therefore never reaches a template argument: the kernel stays templated on `bool NARROW`, the device pass derives `THREADS` from `WARP_SIZE` in the body, and the host recomputes the launch shape from the runtime `warpSize`. Instantiation count is unchanged.

wave64 is byte-for-byte unchanged by this commit: 384 tensors, 0 differences on every output including `kv_cache`, top-k identical for 512 queries at every k, perf flat at 0.993–1.030 with a canary bracketing the run.

**Independent timing** — fair as a process point. For what it is worth the base and head numbers were taken in one batch on one `--exclusive` node, three independent processes per point with medians, a calibrated canary re-measured before and after every job (runs whose canary moved were discarded), and the committed kernel sha256-matched against the tree the numbers came from. The perftest harness is in the op test if a maintainer wants to re-run it.

**Barrier fix ships no failing test** — correct, and I should be precise rather than claim infeasibility. It *is* feasible: FP4 at 16384 x 64 with 40 repetitions detects the race 93% of the time in about 4 s, and an earlier revision of this PR carried exactly that test. It was dropped because commit 2 removes `block_reduce` from the indexer entirely, so the construct that test protected no longer exists there. What remains uncovered is the `reshape_and_cache` pair, which still uses the pattern — a pre-existing gap this PR does not close. Happy to add the test back if you would rather have it.

**Narrowed q inputs** — checked the one consumer I can see: ATOM passes a plain `[T, H, 128]` tensor with no slicing or transpose, so the checks are pure hardening there. I cannot speak for the SGLang side.

Adjacent to your wave32 point: FP4 on gfx1250 is refused with the right message and no silent zero-writes, but `AITER_CHECK` aborts rather than raising, because `check_fail` only throws under a thread-local opt-in. That is pre-existing and untouched here, and it is why I left `SUPPORTED_GFX` alone — adding gfx1250 would make the test abort when it reaches fp4 mode, which would need a third change to the test file. The gfx1250 runs above were driven from a script outside the repo with that gate bypassed.

yzhou103 (approved):
tested  perf ,no drop

PR #5583, opened 2026-09-16
[FlyDSL] [CI] [a16w4] Support MXFP4 weights on gfx942 (CDNA3)

MHYangAMD:
## Motivation

`moe_2stage_a16wmix` cannot run with MXFP4 weights on gfx942 (CDNA3) at all — it does not
even assemble. `make_b_loader`'s `upconvert_b` upconverts the weights with
`v_cvt_scalef32_pk_bf16_fp4`, which is CDNA4-only:

```
error: instruction not supported on this GPU (gfx942): v_cvt_scalef32_pk_bf16_fp4
        v_cvt_scalef32_pk_bf16_fp4 v47, v56, v90 op_sel:[1,0,0]
```

1792 occurrences, and it is the **only** unsupported instruction in the kernel. Everything
else in the pipeline is already arch-gated and fine on CDNA3 — the `use_k16` K=16 MFMA
path, the int4 dequant one line above the fp4 branch, the `dma_via_vgpr` A-tile staging.
So the weight upconvert is the single thing standing between this kernel and CDNA3.

This is architectural rather than SKU-specific: the instruction is absent from all of
CDNA3, so every gfx942 part is affected (MI300X, MI300A, MI308X, MI325X). CDNA3 also has
no fp4 MFMA, so the weights genuinely have to be decoded to bf16 before the MFMA rather
than fed in directly as on CDNA4.

The goal is to make the existing a16w4 MXFP4 path work on gfx942, with the gfx950 path
bit-unchanged. `op_tests/flydsl_tests/test_flydsl_moe_a16wfp4.py` is already gated for
`get_gfx() in ("gfx942", "gfx950")`, so no test gating changes here — that test is
supposed to run on CDNA3 already and only fails because the kernel will not build.

## Technical Details

`_fp4_nibble_to_bf16x8` in `moe_2stage_a16wmix/utils.py` is the CDNA3 counterpart of the
`_int4_nibble_to_bf16x8` already in that file, and `upconvert_b` selects it the same way
it selects the int4 one. It is a `v_perm_b32` byte lookup rather than per-element
arithmetic, because for E2M1 it can be:

- Every E2M1 magnitude is exactly representable in bf16, so the 16-entry table reduces to
  a byte permute over two constant pools — one for the bf16 high bytes, one for the low
  bytes — selected by `nibble & 7`.
- The sign is the nibble's bit 3, shifted into bf16 bit 15.
- The E8M0 scale is a power of two, so applying it leaves the result exactly
  representable and the f32 -> bf16 step is a truncation, not a round-to-nearest.
- E2M1 has no NaN encoding, so no NaN handling is needed.

The result is no reconstruction through f32, no compares or selects for the subnormal
cases, and no full `cvt`.

**Arch gating.** The selection is an `is_gfx942` flag, following the
`is_gfx942 = str(gpu_arch).startswith("gfx942")` / `const_expr(is_gfx942)` idiom already
used in `preshuffle_gemm.py`. It is computed inside `make_b_loader` because, as that
factory's docstring notes, FlyDSL hashes its source for the cache key and not nested
helpers.

The arch arrives as a new `rocm_arch` parameter rather than being derived from `use_k16`.
`use_k16` means "K=16 MFMA", which is only correlated with CDNA3 — forcing the K=16 path
on a gfx950 box to validate it must not also switch the weight decode. `rocm_arch` is
threaded explicitly alongside `use_k16` (`compile_gemm{1,2}_a16w4_port` -> the local
`_use_k16` alias -> `_gemm{1,2}_body_a16w4` -> `make_b_loader`) rather than resolved
inside the builder, so `compile_gemm{1,2}_a16w4_port`'s `functools.cache` stays keyed on
it and two targets cannot share a cached kernel. It has no default at the compile entry
point on purpose — a silently-wrong arch produces a kernel that does not assemble — so
both callers are updated: `moe_2stage_a16wmix/__init__.py` and `ops/flydsl/moe_kernels.py`.

gfx950 is unaffected: the new branch is `const_expr`, so there is no runtime check and the
emitted gfx950 ISA is unchanged.

`+100 / -0` across 5 files, no deletions.

Related: [ROCm/FlyDSL#424](https://github.com/ROCm/FlyDSL/pull/424) and
[#431](https://github.com/ROCm/FlyDSL/pull/431) introduced these kernels as gfx950 work;
[vllm-project/vllm#50319](https://github.com/vllm-project/vllm/pull/50319) asked for
exactly this gfx942 MXFP4-bf16 MoE capability and had to route around its absence.

## Test Plan

Environment: gfx942 (MI308X), ROCm 7.14, flydsl 0.3.2.

1. Assemble `compile_gemm1_a16w4_port` for the Kimi-K3 `3584x512 E896 topk16` geometry and
   count unsupported instructions, before and after, natively on gfx942 and with
   `ARCH=gfx950` to confirm the CDNA4 path is untouched.
2. Numerics: because this container's prebuilt `module_aiter_core.so` predates aiter main
   and raises `ActivationType has no attribute 'GeluTanh'` in `aiter/ops/quant.py`,
   `op_tests/flydsl_tests/test_flydsl_moe_a16wfp4.py` could not be run here. The decode
   was instead validated through `ROCm/FlyDSL`'s `tests/kernels/test_moe_gemm.py`
   (`test_a16w4_moe_e2e`, strict cosine gate against a torch reference), with the decode
   block lifted verbatim from this change — byte-identical, 3919 bytes, sha256 prefix
   `8cda818e15177887`.
3. The same FlyDSL selection with only the kernel change reverted, to confirm the tests
   are not passing vacuously.

## Test Result

| check | before | after |
|---|---|---|
| gfx942 assembly, unsupported instructions | 1792 | **0** |
| `ARCH=gfx950` assembly, unsupported instructions | 0 | 0 |
| FlyDSL `test_a16w4_moe_e2e` (a16w4 selection), gfx942 | 3 failed, 2 passed | **5 passed** |

The `before` column of row 3 is the load-bearing one: the three failures are exactly the
fp4 cases (`small-a16w4`, `kimi512-a16w4`, `native_layout[small]`), while the two that
still pass are the `a16w16` bf16-weight variants, which never reach the decode. So the
gate is neither vacuous nor over-broad.

Not claimed here: no gfx950 measurements or behavioural claims — this box is CDNA3 only.
Since the new path sits behind `const_expr(is_gfx942)` and the gfx950 ISA is unchanged,
gfx950 behaviour cannot move, but a gfx950 CI run to confirm that would be welcome. The
aiter-side e2e test is also still unrun for the environment reason above; it needs a
container whose `module_aiter_core` matches aiter main.

Out of scope, happy to file separately: `test_flydsl_moe_a16wfp4.py` covers
`inter_dim` 512 and 384, but FlyDSL's `test_moe_gemm.py` requires `inter_dim % 256 == 0`,
so the two suites do not cover the same shapes for this kernel.

## Submission Checklist


MHYangAMD:
python3 -m pytest op_tests/flydsl_tests/test_flydsl_moe_a16wfp4.py -v

Test result:

w/o this patch
============================= test session starts ==============================
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-1-3584-512] FAILED   [  8%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-1-3584-384] FAILED   [ 16%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-16-3584-512] FAILED  [ 25%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-16-3584-384] FAILED  [ 33%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-128-3584-512] FAILED [ 41%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-128-3584-384] FAILED [ 50%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-1-3584-512] FAILED  [ 58%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-1-3584-384] FAILED  [ 66%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-16-3584-512] FAILED [ 75%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-16-3584-384] FAILED [ 83%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-128-3584-512] FAILED[ 91%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-128-3584-384] FAILED[100%]
=================================== FAILURES ===================================

w/ this patch
============================= test session starts ==============================
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-1-3584-512] PASSED   [  8%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-1-3584-384] PASSED   [ 16%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-16-3584-512] PASSED  [ 25%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-16-3584-384] PASSED  [ 33%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-128-3584-512] PASSED [ 41%]
test_flydsl_a16wfp4_situv2_e2e[1.0-1.0-128-3584-384] PASSED [ 50%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-1-3584-512] PASSED  [ 58%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-1-3584-384] PASSED  [ 66%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-16-3584-512] PASSED [ 75%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-16-3584-384] PASSED [ 83%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-128-3584-512] PASSED[ 91%]
test_flydsl_a16wfp4_situv2_e2e[4.0-25.0-128-3584-384] PASSED[100%]
============================= 12 passed in 38.94s ==============================

zufayu:

**Adds a gfx942 (CDNA3) weight-decode path to the FlyDSL a16wmix MoE kernels: MXFP4 (E2M1) weights are decoded to bf16 through v_perm_b32 byte-lookup constant pools with the E8M0 group scale applied in f32, replacing the CDNA4-only v_cvt_scalef32_pk_bf16_fp4 upconvert that made the kernel impossible to assemble on gfx942.**

Validation (deterministic): NOT RUN — the PR changes runtime code but ships no test target (auto-validation disabled, REVIEW_AUTO_VALIDATE=0); the changed path is executed at run time and is covered by the existing op_tests/flydsl_tests/test_flydsl_moe_a16wfp4.py, which is gated gfx942+gfx950 and runs on the MI300X CI shards
Perf (advisory): MEASURED — head-only on this MI300X (gfx942); base cannot be timed because it does not assemble (v_cvt_scalef32_pk_bf16_fp4 unsupported, reproduced on this box). fused_moe a16w4 SiTUv2 end-to-end (E=896, topk=16, model_dim=3584, inter_dim=512): token=1 0.319 ms/iter, token=128 4.883 ms/iter, 20 iters after 5 warmup. No base-vs-head ratio can exist on gfx942 by construction; on gfx950 the emitted ISA is unchanged (the new branch is const_expr'd out at trace time)

⚠️ [verified] The new gfx942 decode in aiter/ops/flydsl/kernels/moe_2stage_a16wmix/utils.py (_fp4_nibble_to_bf16x8 behind the is_gfx942 branch of upconvert_b, ≈14 v_perm_b32/integer/f32 ops per 8 weights vs 4 cvt_scalef32 on gfx950) has no benchmark entry point — op_tests/op_benchmarks/triton/bench_moe_gemm_a16w4.py benches the Triton moe_gemm_a16w4, not this 2-stage kernel — so the cost of the path this PR enables cannot be measured with the repo's own tools; the perf-line numbers above are e2e wall time from a hand-written script, not a harness, and have no base to compare against on gfx942. **Author must** add a bench entry for the a16wmix 2-stage path (or post kernel-level gfx942 characterization numbers) so the path is measurable in-repo.
📝 [verified] The new gfx942 decode path in aiter/ops/flydsl/kernels/moe_2stage_a16wmix/utils.py was hand-verified on this MI300X box against merge target 22d2c7c9: without the PR, op_tests/flydsl_tests/test_flydsl_moe_a16wfp4.py fails to assemble with exactly the error the PR description quotes (v_cvt_scalef32_pk_bf16_fp4 not supported on gfx942); with the PR applied it passes 12/12 in 173 s, exercising the decode end-to-end through fused_moe → compile_flydsl_moe_stage1/2 → both gemm1 and gemm2, and the decode itself is bit-exact against an E2M1×E8M0 reference by exhaustive emulation (all 256 raw bytes × 9 scale bytes × 8 elements, 18432/18432) under v_perm_b32 semantics measured on this box. **Reviewer should ask** the author to link the gfx942 CI run of that test on the PR.
📝 [verified] The enablement inherits the a16wmix A-path 32-bit addressing limit flagged in #5519: gemm1.py:148-149 builds the activation row base as token_id × K/2 dwords in fx.Int32 and the byte offset as token_id × K × 2, so a bf16 activation with tokens × K × 2 bytes ≥ 2^31 silently overflows (≈116k tokens at K=9216). The limit is pre-existing, identical on gfx950, and untouched by this diff — but before this PR the a16wmix path could not run on gfx942 at all, so it is newly reachable there. **Reviewer should ask** whether the gfx942 support should document the token bound or reference #5519 now that the path is reachable.

op_tests/op_benchmarks/flydsl/bench_moe_a16wmix_2stage.py
```diff
@@ -0,0 +1,669 @@
+# SPDX-License-Identifier: MIT
```
coderfeli: We don't need this bench

MHYangAMD: Removed.

coderfeli (approved):

PR #5186, opened 2026-09-01
[CK] [CI] [JIT] Mi300a enablement

afanfa:
## Motivation

Add initial support for MI300A

## Technical Details

This PR has three parts: 1) improve the device detection for MI300A; 2) add encoding for fp8 e4m3fnuz ; 3) Add MI300A (gfx942, cu_num=228) configs for GEMM and fused MoE

## Test Plan

New tests for the newly added features have been added. The test suite has been run

## Test Result

No regressions or failures on the newly added tests.

PR #3324, opened 2026-05-24
ck_gemm_a8w8_blockscale: add tuned configs for Qwen3-VL-235B, Qwen3-VL-32B, Qwen3.6-27B (gfx942)

eppaneamd:
## Motivation

Add `a8w8_blockscale_tuned_gemm` CSVs for three models across TP=2/4/8 on gfx942 (MI300X, cu_num=304).

## Changes

3 CSVs in `aiter/configs/model_configs/` (one per model, covering all TP variants):

| File | Model | Shapes |
|---|---|---:|
| `a8w8_blockscale_tuned_gemm_qwen3_vl_235b.csv` | Qwen3-VL-235B-A22B-Instruct-FP8 | 4,086 |
| `a8w8_blockscale_tuned_gemm_qwen3_vl_32b.csv` | Qwen3-VL-32B-Instruct-FP8 | 8,172 |
| `a8w8_blockscale_tuned_gemm_qwen36_27b.csv` | Qwen3.6-27B-FP8 | 9,534 |

Each file contains TP=2/4/8 shapes concatenated. (M,N,K) tuples are disjoint across TP values — N differs per TP — so no deduplication was needed. Per-TP files replaced per reviewer feedback.

## Test results

**Environment:**
- Hardware: AMD MI300X (gfx942, 304 CUs)
- AITER: 0.1.13.post1
- Metric: sum of per-shape execution time (µs) across all shapes in the CSV, `ck` kernel, `ck_preshuffle=False`

| Model | TP | Shapes | Default (s) | Tuned (s) | Avg speedup |
|---|---|---:|---:|---:|---:|
| Qwen3-VL-235B-A22B-Instruct-FP8 | 8 | 1,362 | 2,021 | 651 | **2.91×** |
| Qwen3-VL-235B-A22B-Instruct-FP8 | 4 | 1,362 | 3,556 | 1,100 | **3.06×** |
| Qwen3-VL-235B-A22B-Instruct-FP8 | 2 | 1,362 | 6,962 | 2,116 | **3.14×** |
| Qwen3-VL-32B-Instruct-FP8 | 8 | 2,724 | 13,031 | 3,793 | **3.27×** |
| Qwen3-VL-32B-Instruct-FP8 | 4 | 2,724 | 23,635 | 7,274 | **3.13×** |
| Qwen3-VL-32B-Instruct-FP8 | 2 | 2,724 | — | — | — |
[...]

PR #5585, opened 2026-09-16
[Config] Add Qwen3.8-27B TP1 a8w8 blockscale GEMM tunings for gfx942

phambinhfin:
## What this adds

**641 tuned `a8w8_blockscale` rows in one file** — 249 decode (M ≤ 512) and 392 prefill (M 907…65536) — covering the five GEMM widths Qwen3.8-27B drives at **TP1** on gfx942 / `cu_num=304`: `down` 5120×17408, `out_proj` 5120×6144, `in_proj` 16384×5120, `qkv` 14336×5120, `gate_up` 34816×5120.

The accompanying untuned CSV is the shape list these were tuned from. It was extracted from real vLLM server logs rather than generated as an M ladder, so the M grid is what the scheduler actually requests.

#3324 tuned this model family at TP=2/4/8 only, so the TP1 widths were never covered and every call fell back to the default — **275 `not found tuned config` messages per profile run, now 4.**

## Results — decode

This config against AITER's own default, measured in a real vLLM serving run — Qwen3.8-27B-FP8, 128K context, max-concurrency 1, one MI325X — with byte-identical images and the config bind-mount as the only variable, three interleaved repeats per arm on an idle node.

| | AITER default | this PR | |
| --- | --- | --- | --- |
| decode step | 15.444 ms | **12.449 ms** | **1.24×** |

Per-shape decode GEMM against the default, µs/call: `down` 2.33×, `out_proj` 1.51×, `in_proj` 1.12×, `qkv` 1.00×, `gate_up` 1.00×. Decode-step spread across repeats was 0.037 ms (default) and 0.012 ms (this PR). **No decode shape regresses.**

## Method and validation

- Tuned with `csrc/ck_gemm_a8w8_blockscale/gemm_a8w8_blockscale_tune.py --libtype ck --splitK`. Decode rows were re-timed with an interleaved sweep-then-verify pass, because the tuner's `min()` over short samples is noisy on near-ties. Only rows beating the default by ≥3% were kept, following #5421.
- `op_tests/tuning_tests/test_csv_validation.py` (15 tests) and `test_config_shape_collision.py` (17 tests) both pass.
- **Zero duplicate `(gfx, cu_num, M, N, K)` keys** — against all 12 other sources in the runtime merge set on `main` (40,892 rows), and against the three open PRs that touch those same files (#5602, #5278, #5186), none of which adds a gfx942/304 row at these widths. No pre-existing rows for these five `(N, K)` pairs. `update_config_files` raises on duplicates. Nearest related open work is #5586, which routes *untuned* blockscale shapes to triton; these rows make the same shapes tuned, so the two do not collide.
- **Kernel parity**, not a model evaluation: max relative Frobenius error against an fp32 reference is 4.11e-3, against bf16's `2^-8` = 3.9e-3 rounding floor. Every row has tuner `errRatio` ≤ 0.0002.

## Scope and caveats

- **Only the decode rows have in-situ evidence.** The prefill rows carry tuner measurements only, for the reason above.
- **MI300X is also gfx942 / 304 CUs**, so these rows are selected there too and cannot be scoped out without changing the lookup key. Measurement here is MI325X only. #3324's own environment section records its configs as measured on MI300X, so gfx942/304 rows already move in both directions.
- **Two rows pin AITER's default kernel** (`qkv` and `gate_up` at M=1), after the serving trace measured the tuned picks at 0.94× and 0.98× in situ — the 1.00× entries above. Pinned rather than deleted: `getPaddedM(1, N, K, 0) == 16`, so a deleted M=1 row lands on the M=16 row, not the default.
- **Six decode shapes ship no row** — `qkv` and `in_proj` at M=24 and M=32, `out_proj` at M=104 and M=112 — near-ties that did not clear the 3% bar and fall back.
- **`gate_up` ships 72 prefill rows, not 80.** The 8 missing ones are shapes where `M*N > INT32_MAX` and the CK kernel faults during tuning. At runtime `getPaddedM` collapses M > 8192 to 8192 at these widths, so they inherit the tuned M=8192 row rather than going untuned.
- One model, one context length, one concurrency point, one node.


phambinhfin:
@zufayu Could you please review the PR when you have time, thanks

PR #5586, opened 2026-09-16
[Perf] Route untuned a8w8 blockscale GEMMs to triton above a per-arch M

ZhengGong-amd:
`gemm_a8w8_blockscale` picks its CK instance from a table keyed on `(gfx, cu_num, M, N, K)`. When a shape is absent, `get_CKGEMM_config` returns `None` and the dispatcher calls `gemm_a8w8_blockscale_ck` with no `kernelName` and no `splitK` — one fixed tile shape at every M. That instance cannot fill the MFMA pipeline at prefill-sized M: on gfx950 it holds 385–421 TFLOP/s from M=512 all the way to M=16384, while the triton kernel, which buckets its config by M, reaches 1041.

Untuned is the common case rather than a corner. On `gfx950/cu_num=256` the shipped table has no row for any of Qwen3-14B-FP8's four linear shapes, and of the 115 distinct `(N, K)` that one host's fp8 `weight_block_size [128,128]` checkpoints ask for across TP ∈ {1, 2, 4, 8}, 88 have no tuned row. Per-model tuning closes shapes one checkpoint at a time — #5485, #5343, #5585, #4816 and #5219 are all that — but the set is not bounded by anything aiter controls. On this entry point a tuned row cannot select triton at all: the `libtype` dispatch here accepts `ck` and `cktile` and asserts otherwise, unlike `gemm_a8w8_blockscale_bpreshuffle`, which honours `libtype == "triton"`.

So when no tuned config exists and M is at or above a per-arch threshold, dispatch to the triton kernel rather than the default CK instance. Triton loses at decode-sized M, so the threshold is measured rather than derived: 384 on gfx950 (MI355X, cu_num=256) and 2048 on gfx942 (MI325X, cu_num=304). An arch absent from the table keeps the CK fallback at every M, so gfx1250, gfx1201 (#4322) and anything new are unchanged by construction.

The triton call that the unsupported-arch fallback already made is factored out and reused; it takes CK's inputs as they are apart from the `uint8` → `fp8` view.

Scope: the non-preshuffled entry point only. `gemm_a8w8_blockscale_bpreshuffle`, the asm and FlyDSL paths are untouched, a tuned shape takes the branch it took before, and `get_CKGEMM_config` was already on this path and is cached, so no lookup is added. This composes with per-shape tuning rather than competing with it — the predicate stops firing for a shape the moment that shape gets a row.

`op_tests/test_gemm_a8w8_blockscale_fallback.py` pins the dispatch rather than the speed: untuned takes triton at and above the threshold and CK below it, a tuned shape takes CK at every M, and an arch with no threshold never routes. The tuned config and the arch are injected rather than read from the shipped CSV, so tuning one of these shapes cannot make a case vacuous, and the arch gate is covered for gfx940/941/1250 without that hardware.

Measurements for both arches are in a comment below.

---


ZhengGong-amd:
## Measurements

gfx950 numbers are from this host: MI355X, `cu_num=256`, one card, vLLM 0.29.0, torch 2.12.0+git6bbd260. gfx942 is 2× MI325X, `cu_num=304`. The gfx950 substrate carries this repo's `.py` with the host's compiled extensions; `gemm_a8w8_blockscale` is byte-identical between it and `ba5f08ea26`, so the patch applied there is the patch in this PR.

### Dispatch and correctness

Instrumenting which kernel the op actually reaches, over the four Qwen3-14B-FP8 shapes at M ∈ {128, 256, 320, 383, 384, 512, 1024, 2047, 2048, 4096}: before the patch every one of the 40 points takes CK. After it, every point below 384 takes CK and every point at or above it takes triton — the boundary is exact, and a tuned shape (`2048 x 3072` at M=1024, read out of the shipped table for this exact card) stays on CK in both arms.

The two kernels agree to **≤ 1 bf16 ULP** on every shape, and both sit the same distance from an fp32 reference, so the difference is output rounding rather than one kernel being less accurate:

| shape (N x K) | M | max abs output | CK vs fp32 | triton vs fp32 | CK vs triton | in bf16 ULPs |
|---|---|---|---|---|---|---|
| `7168 x 5120` | 2048 | 31.1 | 1.54e-03 | 1.54e-03 | 1.25e-01 | 1.00 |
| `5120 x 5120` | 384 | 31.4 | 1.52e-03 | 1.52e-03 | 1.25e-01 | 1.00 |
| `34816 x 5120` | 384 | 33.8 | 1.51e-03 | 1.51e-03 | 1.25e-01 | 0.50 |
| `5120 x 17408` | 2048 | 94.8 | 1.64e-03 | 1.64e-03 | 5.00e-01 | 1.00 |

GSM8K on gfx942, full 1319 samples, 5-shot greedy, Qwen3-14B-FP8: strict 0.8650 → 0.8628, flexible 0.8681 → 0.8650, against ±0.94 points of standard error.

### Micro, gfx950

CK's untuned fallback against triton, both kernels timed back to back in one process on the same tensors, arm order alternated across 7 repeats of 100 iterations after 50 warmup calls. Greater than 1.00 means triton wins; bold marks a point this PR routes and loses.

| shape (N x K) | model | M=128 | M=256 | M=320 | M=384 | M=512 | M=1024 | M=2048 | M=4096 | M=8192 | M=16384 |
|---|---|---|---|---|---|---|---|---|---|---|---|
| `7168 x 5120` | Qwen3-14B-FP8 qkv | 0.87 | 0.40 | 1.99 | 2.32 | 2.29 | 2.17 | 2.19 | 2.19 | 2.50 | 2.28 |
| `5120 x 5120` | Qwen3-14B-FP8 o | 0.87 | 0.31 | 1.69 | 1.69 | 1.97 | 1.57 | 1.57 | 2.08 | 2.48 | 2.48 |
| `34816 x 5120` | Qwen3-14B-FP8 gate_up | 1.08 | 0.76 | 2.23 | 2.17 | 2.19 | 2.10 | 2.17 | 2.26 | 2.29 | 2.31 |
| `5120 x 17408` | Qwen3-14B-FP8 down | 0.81 | 0.30 | 1.77 | 1.78 | 2.12 | 1.65 | 1.65 | 2.19 | 2.40 | 2.39 |
| `32768 x 1024` | DSV4-Flash kv_b | 0.74 | 1.10 | 1.59 | 1.88 | 2.00 | 1.75 | 1.76 | 1.77 | 1.63 | 1.63 |
| `6144 x 512` | GLM-5.3 TP4 | 0.21 | 0.21 | 0.23 | **0.27** | **0.30** | **0.51** | **0.89** | 1.26 | 1.25 | 1.25 |
| `1024 x 6144` | GLM-5.3 TP4 | 0.56 | 0.13 | 0.56 | **0.57** | **0.57** | **0.38** | **0.68** | 1.27 | 2.42 | 2.43 |

Median over the 49 routed points is 1.97×, and **8 of them regress**. Five of the seven shapes stay above 1.00 from M=320 up, which is where 384 comes from — the cliff at M=256 is aiter's generic triton config, which has a `BLOCK_SIZE_M=256` bucket covering M ∈ (128, 256] and returns to 64 above it. The two narrow shapes do not cross until M=4096. The 21 points below the threshold execute unchanged code in both arms and are an in-run control.

That losing band is fixable here rather than by the threshold: the generic block-scale config has only `M_LEQ_512` and `any` above M=256, so M=768–1024 gets a tile sized for much larger M. Adding buckets would raise the floor for every consumer, and would let the threshold come down.

### End to end, gfx950

Qwen3-14B-FP8, TP=1, ISL 2048 / OSL 128, 600 prompts, `VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_LINEAR=1`. **vLLM is pristine upstream in both arms** — the only difference between them is this diff, which is the point of putting the routing here. Pair 0 is discarded as warmup and arm order alternates.

| pair | order | before | after | delta |
|---|---|---|---|---|
| 0 | orig → patched | 13240 tok/s | 22072 tok/s | +66.7 % (discarded) |
| 1 | patched → orig | 13331 tok/s | 22097 tok/s | +65.8 % |
| 2 | orig → patched | 13334 tok/s | 23807 tok/s | +78.5 % |
| 3 | patched → orig | 13333 tok/s | 24204 tok/s | +81.5 % |
| **median** | | | | **+78.5 %** |

The baseline is 13331/13334/13333 — **0.02 % spread** — while the patched arm climbs 22.1 → 23.8 → 24.2k as triton's autotune cache warms, so +65.8 % is the conservative reading and the ranges are disjoint by a wide margin. All four models' four shapes are untuned on this card, so this is close to the ceiling: 13.3k tok/s is about 350 TFLOP/s of linear work against the 385–421 the table above measures for CK, i.e. these four GEMMs are most of baseline step time.

### Carried over from vllm-project/vllm#56812

The same dispatch decision, measured at vLLM's call site instead of here, before these reviewers asked for it to live in aiter. Same predicate, same thresholds, same two kernels — its gfx950 Qwen3-14B median was +74.0 %, which the +78.5 % above reproduces within the warmup spread.

Four models on gfx950, with the affected share taken from the shapes observed at the op rather than derived from `config.json`:

| model | architecture | affected | e2e | before-spread | after-spread |
|---|---|---|---|---|---|
| `Qwen3-14B-FP8` | dense | 100 % of linear FLOPs | +74.0 % | 1.3 % | 5.2 % |
| `GLM-5.3` | `glm_moe_dsa`, TP=4 | 6 untuned `(N, K)` | +14.0 % | 0.2 % | 0.1 % |
| `DeepSeek-V4-Flash-0731` | `deepseek_v4`, MoE+MLA | 80 % of linear FLOPs | +4.5 % | 0.5 % | 0.5 % |
| `Qwen3-Next-80B-A3B-Instruct-FP8` | `qwen3_next` | **0 %** — aiter ships its CSV | −0.5…+1.4 % | 2.6 % | 1.9 % |

The last row is the useful one: every shape it issues is CK-tuned, so the predicate returns false everywhere and the run is a *provable* no-op rather than one that merely measured flat. Its baseline was re-run five times alone on one card to pin the noise floor at 2.6 %, which is the figure to judge the rows above against.

GLM-5.3 is also the reason the threshold is not per-shape. At TP=4 it routes both of the shapes the sweep above finds worst, `6144 x 512` and `1024 x 6144`, and still measures +14.0 % with 0.1–0.2 % arm spreads. A shape rule drafted from isolated ratios would have excluded exactly those and lost the win.

### gfx942, and why 2048

Measured both thresholds on 2× MI325X rather than picking one:

| config | threshold 384 | threshold 2048 |
|---|---|---|
| Qwen3-14B-FP8, long prefill | +11.7 % | **+14.1 %** |
| Qwen3.5-122B-A10B-FP8, long prefill | +4.9 % | **+7.9 %** |
| Qwen3-14B-FP8, short prefill | +6.2 % | flat |
| Qwen3.5-122B-A10B-FP8, short prefill | **−1.1 %** | flat |

2048 is better on both long-prefill configs by 2.4–3.0 points and removes the reproducible −1.1 % regression on the MoE short-prefill case. What it gives up is the +6.2 % on 14B short prefill, where M never reaches 2048 and nothing routes. That trade looks right: the short-prefill gain appears on one model and is negative on the other, which says 384 was taking mid-M shapes CK should have kept, while the long-prefill gain is positive on both models and 2048 makes it larger. Nothing between the two has been measured; 1024 is the obvious next candidate if the threshold is worth tuning further rather than just making safe.

CK does not flatline on gfx942 the way it does on gfx950 — 250–300 TFLOP/s across M against 418 flat — and triton tops out near 330, so the ceiling there is ~1.15× rather than 2.35×. Method for that batch: 3 pairs per config, pair 0 discarded, arm order alternated, and a null arm with the threshold set to 10⁹ so it provably never routes, which shows ≤0.05 % between arms — the harness is unbiased and the structural change alone costs nothing.

### Two limits worth stating

The crossover is not monotonic. Across 384…2048 on gfx942, six of ten shapes dip below 1.0 and come back: `5120 x 17408` runs 1.73, 1.36, 0.98, 1.27, 1.02, 0.90, 1.14. So "a threshold above which every shape is safe" has no solution, and both thresholds here are chosen because they win end to end, not because every shape is below 1.0.

Isolated ratios also do not extrapolate. `5120 x 17408` at M=1024 is 1.27× in isolation but 0.99 when swept over a realistic 40-layer 10.7 GB working set where weights are cold per layer. An isolated harness keeps one weight set cache-resident and overstates wins and losses alike, which is why the e2e tables rather than the micro table are what the thresholds were picked on.

simondanielsson (commented):
Nice! Should we add a UT to ensure we retain this invariant of falling back to triton if untuned and `m` is above the threshold?

ZhengGong-amd:
Added, in `op_tests/test_gemm_a8w8_blockscale_fallback.py`.

It pins which kernel the dispatcher reaches rather than how fast it is, since the arithmetic is not the part that will break. Both kernels are replaced with recorders and the tuned config and the arch are injected rather than read from the shipped CSV, so the file runs in about 5 seconds with no GEMM and no triton JIT — and tuning one of these shapes later cannot quietly turn a case vacuous, which matters given how often the tuned tables grow.

Four invariants, of which the last two are the ones I think earn their keep:

- untuned and `m >= threshold` → triton, at exactly the threshold and above;
- untuned and `m < threshold` → CK;
- **tuned → CK at every M**, which pins the threshold check *after* `get_CKGEMM_config` rather than before it;
- **an arch with no measured threshold never routes**, covered for gfx940/941/1250 by injecting the arch, so the fail-safe is tested without that hardware.

Plus one real-execution case asserting the kernel the fallback picks agrees with the one it replaces.

A test that passes on the patch and fails on `main` has only shown it depends on *something* I changed, so I ran it against four plausible-but-wrong versions of the dispatch. Each fails a different subset, which is what makes the set worth keeping:

| dispatch version | what fails |
|---|---|
| shipped | — (11 pass) |
| `m > min_m` instead of `>=` | only the case at exactly the threshold |
| threshold checked *before* `get_CKGEMM_config` | only the two tuned-shape cases |
| `.get(get_gfx(), 384)` instead of `.get(get_gfx())` | only the three no-threshold-arch cases |
| no threshold, route every untuned shape | the below-threshold and no-threshold-arch cases |

What it deliberately does not cover is whether 384 and 2048 are the *right* numbers. Those are a measurement, not an invariant, so the test reads them from the table instead of hardcoding them and skips on an arch that has no entry.

ZhengGong-amd:
> Nice! Should we add a UT to ensure we retain this invariant of falling back to triton if untuned and `m` is above the threshold?

UT added

ZhengGong-amd:
## Two more models end to end on gfx950

MoE+MLA and MoE-DSA, to go with the dense case above. Same conditions: MI355X, `cu_num=256`, **vLLM pristine upstream in both arms** so the only difference is this diff, ISL 2048 / OSL 128, 600 prompts, pair 0 discarded as warmup, arm order alternated. Single tenant on the host for all 16 runs, and both checkpoints were copied to local disk first so NFS is out of the loop.

**DeepSeek-V4-Flash-0731** — `deepseek_v4`, MoE + MLA, TP=1 on one card, `--kv-cache-dtype fp8`. About 80 % of its linear FLOPs are untuned block-scale, but that share is itself small next to the MoE expert GEMMs.

| pair | order | before | after | delta |
|---|---|---|---|---|
| 0 | orig → patched | 18177 tok/s | 20644 tok/s | +13.6 % (discarded) |
| 1 | patched → orig | 20036 | 21037 | +5.0 % |
| 2 | orig → patched | 19930 | 20629 | +3.5 % |
| 3 | patched → orig | 19928 | 21223 | +6.5 % |
| **median** | | | | **+5.0 %** |

Baseline spread 0.54 % (20036 / 19930 / 19928), patched 2.82 %, ranges disjoint.

**GLM-5.3** — `glm_moe_dsa`, TP=4 across four cards, 6 untuned `(N, K)` at this split.

| pair | order | before | after | delta |
|---|---|---|---|---|
| 0 | orig → patched | 9288 tok/s | 9553 tok/s | +2.9 % (discarded) |
| 1 | patched → orig | 9929 | 11406 | +14.9 % |
| 2 | orig → patched | 10065 | 11152 | +10.8 % |
| 3 | patched → orig | 10175 | 11260 | +10.7 % |
| **median** | | | | **+10.8 %** |

Spreads 2.45 % and 2.26 %, ranges disjoint by ~1000 tok/s.

Worth flagging that the GLM baseline is still warming across the three pairs — 9929 → 10065 → 10175, +2.5 % — while the patched arm is flat, so the delta shrinks with pair index. The +14.0 % quoted for this model in vllm-project/vllm#56812 came from two pairs, both early, on a contended host; +10.8 % over three pairs on a quiet one is the better number and I would rather correct it here than carry the larger one forward.

### Where that leaves the gfx950 picture

All three measured with vLLM pristine, so these are the aiter-side change on its own:

| model | architecture | affected | aiter-side (this PR) | vLLM-side (#56812) |
|---|---|---|---|---|
| `Qwen3-14B-FP8` | dense | 100 % of linear FLOPs | **+78.5 %** | +74.0 % |
| `GLM-5.3` | `glm_moe_dsa`, TP=4 | 6 untuned `(N, K)` | **+10.8 %** | +14.0 % (2 pairs, contended) |
| `DeepSeek-V4-Flash-0731` | `deepseek_v4`, MoE+MLA | 80 % of linear FLOPs | **+5.0 %** | +4.5 % |
| `Qwen3-Next-80B-A3B-Instruct-FP8` | `qwen3_next` | **0 %** — aiter ships its CSV | not re-run | −0.5…+1.4 %, 2.6 % noise floor |

The ordering is the point rather than any single number: the gain tracks how much of the model's linear work has no tuned row, and the control that structurally cannot reach the path sits on zero. GLM-5.3 is also still the case against a per-shape rule — at TP=4 it routes both shapes the micro sweep finds worst, `6144 x 512` and `1024 x 6144`, and gains anyway.

yifehuan (approved):
LGTM

ZhengGong-amd:
## GSM8K accuracy, gfx950

**No model moves more than 0.0061 strict-match, every delta is inside one standard error, and in two of the four a second run of the *unpatched* build moves as far or further than the patch does.**

`lm_eval 0.4.13`, `--tasks gsm8k --num_fewshot 5 --batch_size auto`, greedy, all 1319 items, `max_model_len=4096`, `max_gen_toks=256`, `enable_prefix_caching=False`, `seed=1234`. vLLM `0.29.0`, torch `2.12.0+git6bbd260`, MI355X `cu_num=256`, `VLLM_ROCM_USE_AITER=1 VLLM_ROCM_USE_AITER_LINEAR=1`. The arms are one tree differing in exactly one file — `diff -r` verified, and the patched tree checked AST-identical to this PR's head — with separate `VLLM_CACHE_ROOT` and `TORCHINDUCTOR_CACHE_DIR` per arm, prompts identical on all 1319 items, and runs strictly serialised on an otherwise idle host.

`base2` is a symlink to the base tree, so its delta is the harness's own noise floor rather than anything about the patch.

| model | untuned (N,K) | calls rerouted | strict base | strict patched | Δ strict | Δ base-repeat | stderr | Δ flexible |
|---|---|---|---|---|---|---|---|---|
| `Qwen3-14B-FP8` | 4 | 492 | 0.8886 | 0.8863 | −0.0023 | −0.0015 | ±0.0087 | −0.0076 |
| `GLM-5.3` (TP=4) | 6 | 2832 | 0.9227 | 0.9242 | +0.0015 | +0.0000 | ±0.0074 | +0.0008 |
| `DeepSeek-V4-Flash-0731` | 1 | 101 | 0.9318 | 0.9325 | +0.0008 | +0.0000 | ±0.0069 | +0.0000 |
| `Qwen3-Next-80B-A3B-Instruct-FP8` | 3 | 9 | 0.8143 | 0.8203 | +0.0061 | −0.0038 | ±0.0107 | +0.0099 |

The largest movement, +0.0061, is on the model the patch barely touches — 9 rerouted calls against Qwen3-14B's 492 — and its base-repeat moves −0.0038 on its own. That ordering is the signature of noise rather than effect: if the routing were hurting accuracy the deltas would track the number of rerouted calls, and they do the opposite.

### A correction to the "provable no-op" claim

I have been describing `Qwen3-Next-80B-A3B-Instruct-FP8` as 0 % affected and a *provable* no-op, here and in vllm-project/vllm#56812. **That is wrong and I should have checked it rather than inherited it.** Instrumenting the shapes aiter actually reports as untuned during a run, it has three `(N, K)` with no tuned row on this entry point — `(1024, 2048)`, `(2048, 512)`, `(2048, 4096)` — and all three are reached at M = 8320…10395, well above the gfx950 threshold. So it is rerouted 9 times in a full GSM8K run.

What it actually is, then, is the *least* affected model by a factor of about 55, which is why it measures flat. "Measured flat with a tiny affected surface" is a weaker claim than "structurally cannot reach the path", and it is the accurate one. The genuinely provable statements about scope are the ones about entry points and architectures: `gemm_a8w8_blockscale_bpreshuffle` is untouched, which the same instrumentation confirms — DeepSeek-V4-Flash drives `(1536, 4096)` untuned through the preshuffle path 143 times and this PR reroutes none of them — and an arch absent from the threshold table never routes at any M.

The reachability counts above also confirm two things I had asserted from `config.json`: GLM-5.3 really does have exactly 6 untuned `(N, K)` at TP=4, and DeepSeek-V4-Flash's is `(32768, 1024)` rather than the `(33792, 4096)` that deriving it from config predicts.

### Per-sample agreement

Sharper than the aggregate, and it needs the same control, because vLLM's batching makes bf16 reductions batch-shape dependent — two runs of the *same* build already disagree on individual generations.

| model | identical text base→patched | identical text base→base | same verdict base→patched |
|---|---|---|---|
| `Qwen3-14B-FP8` | 45.0 % | **94.8 %** | 96.0 % |
| `GLM-5.3` | 26.3 % | 27.5 % | 95.6 % |
| `DeepSeek-V4-Flash-0731` | 36.2 % | 37.1 % | 95.2 % |
| `Qwen3-Next-80B-A3B-Instruct-FP8` | 38.1 % | 35.4 % | 95.1 % |

Three of the four are indistinguishable from their own base repeat — for `Qwen3-Next` the patched arm is *closer* to base than base's own repeat is. `Qwen3-14B-FP8` is the one real signal: 45.0 % against a 94.8 % base-repeat, so the patch does change its generations, which is expected when every dense linear in the model reroutes at every prefill above M=520. Even there the score moves −0.0023 against ±0.0087 of standard error, and 96.0 % of items keep the same verdict.

### What this does not cover

5-shot greedy short-answer arithmetic exercises one shape of one path. It says nothing about long context, sampling, or multi-turn behaviour. The claim is "no measurable accuracy change on GSM8K, within the base-repeat noise floor" — not bit-exactness, which the ≤ 1 bf16 ULP kernel difference already rules out.

PR #5596, opened 2026-09-16
[CI] Enable 2P1D cases for vllm DI CI tests and update vllm container to use latest v0.29.0 release version.

lcskrishna:
## Motivation

- **Enable vLLM v0.29.0**: `BASE_IMAGE` → `vllm/vllm-openai-rocm:v0.29.0` and
  `VLLM_BASELINE_SHA` → `98dff2a81d747d1dba01a47f939f48c3526d4206`. The
  `select_vllm_cases` checkout now reads the env var instead of repeating the
  SHA, so the image and the harness tree can't drift apart.
- **Add 2P1D coverage**: upstream's `pipeline-disagg.yaml` only ships 1P1D and
  2P2D, so each 2P1D case is derived from its 2P2D counterpart by rewriting
  `yD=2 NODES=4` → `yD=1 NODES=3`. Both the rewrite and the relabel assert they
  matched, so an upstream reformat fails loudly instead of silently dropping
  coverage. Selected cases go 10 → 15 (5 per topology).
- **Allow 3-node jobs**: the node-pool guard accepted only 2 or 4. No harness
  change is needed — `run-slurm-disagg-test.sh` already patches
  `#SBATCH --nodes` at submit time.
- **Manual subset runs**: new `workflow_dispatch` inputs `models` and
  `topologies` (both empty by default, so PR and cron runs are unaffected).
  Unknown values are rejected with the available list rather than producing an
  empty, silently-green matrix.
- **Router image**: `vllm/vllm-router:nightly`. The previous digest pin is
  orphaned — no current tag points at it — and that repo publishes no release
  tags, only nightlies.

## Submission Checklist


lcskrishna:
Tested offline on the CI cluster with models - DSV3, DSR1-FP4, & Kimi K2.6 FP4.

gyohuangxin (approved):

PR #5605, opened 2026-09-16
[Build] prebuild: add the missing nmask/nlse bf16 mha_varlen_fwd variant

Arist12:
`PREBUILD_KERNELS=1` prebuilds five of the six bf16 `mha_varlen_fwd` variants that the dense (non-sparse) attention path can ask for. `req_md_names` in `setup.py` hardcodes the list, and it covers mask/nmask crossed with lse/nlse except for one combination: no-mask **and** no-LSE. That is the plainest case of the six.

The consequence is not a missing feature but a startup hang. The variant is still perfectly buildable, so it gets JIT-compiled on first use instead — and in a tensor-parallel server every rank reaches it at the same moment. On an 8-GPU node, eight SGLang schedulers start compiling the same kernel concurrently and the engine's readiness probe times out before any of them finishes.

## How we found it

Bringing up GLM-5.2 744B LoRA RL on MI355X (8x gfx950) under SGLang. GLM-5.2's DSA indexer only goes sparse above `index_topk` (2048) tokens, so a short rollout runs dense attention through `mha_varlen_fwd` with no causal mask and no returned LSE — exactly the combination that is not prebuilt. Engine startup then hung, and the log names the module:

```
(CommandActor pid=59282) [aiter] [pid=63232] start build
    [mha_varlen_fwd_bf16_nlogits_nbias_nmask_nlse_ndropout_nskip_nqscale]
(CommandActor pid=59284) [aiter] [pid=63264] waiting for baton release at
    .../build/lock_mha_varlen_fwd_bf16_nlogits_nbias_nmask_nlse_ndropout_nskip_nqscale
(CommandActor pid=59285) [aiter] [pid=63258] waiting for baton release at
    .../build/lock_mha_varlen_fwd_bf16_nlogits_nbias_nmask_nlse_ndropout_nskip_nqscale
```

One rank builds it; the other ranks serialise behind that module's `FileBaton`, and the engine's readiness probe times out first. Prebuilding this one variant out-of-band (~27 s) and caching the `.so` made startup reliable, which is what pointed at the prebuild list rather than at the kernel.

## Evidence

On `rocm/sgl-dev` (`v0.5.16-rocm720-mi35x`, and reproduced on `v0.5.15.post1`):

```
$ ls /sgl-workspace/aiter/aiter/jit/ | grep -c mha_varlen_fwd
5
```

The five present are exactly the five in `req_md_names`; the no-mask/no-LSE variant is the one absent.

`get_mha_varlen_prebuild_variants_by_names` already accepts the name and returns a well-formed recipe, identical in shape to a shipped one — so this is a list omission and not a missing recipe:

```
SHIPPED: 1 variant(s); blob_gen_cmd=True
MISSING: 1 variant(s); blob_gen_cmd=True
```

## Scope

One more variant in the prebuild set. No kernel, recipe or build-system change, and no effect on `PREBUILD_KERNELS=0`, `2` or `3`.

## What we did not verify

We did not run a full base-image rebuild to watch the `.so` land, because `PREBUILD_KERNELS=1 setup.py build_ext` rebuilds the whole module set (one module alone took 980 s in our timing) and that belongs in the image builder's CI rather than in our environment. What is verified is that the variant is absent from the shipped set, that aiter's own recipe lookup accepts the name and returns a valid build entry, and that the kernel builds — the runtime JIT builds it today, which is the behaviour this change moves to build time.


Arist12:
@amd-ruitang3 hi, could you help review this when you got a time? thanks!

PR #5606, opened 2026-09-17
[Triton/Gluon] Add fused Qwen3-Next GDN prefill (gfx950 Gluon) with fp8 group-quant epilogue

karverma-amd:
A single fused op for the Qwen3-Next gated delta-net *prefill* chain (causal_conv1d_split_qkv -> fused gating -> fp32 chunked delta scan -> gated RMSNorm -> per-head group-128 FP8 quant), replacing the four-kernel path with a tight set of Gluon launches that fuse the work intra-kernel and drop the intermediate HBM traffic. The FP8 activations a block-FP8 out_proj consumes are emitted by the epilogue launch (no separate quant kernel).

Support is reported, not asserted: fused_gdn_prefill_qkvz_supported gates on gfx950, Triton >= 3.8 (Gluon dialect), and an in-range (tokens, batch)/dtype contract, so callers fall back to the four-kernel chain instead of crashing. Prefill spans a wide token range, so it dispatches one of four autotuned M-tile schedules. The Gluon/HIP tiles import lazily on dispatch, keeping the module and the support predicate importable on any platform.

Tests: pure-torch fp32 sequential-scan reference (deliberately not the chunked algorithm) at the kernel's bf16 rounding boundaries; 9/9 pass on MI355X (gfx950, Triton 3.8). Includes an op benchmark.

### Motivation
  The Qwen3-Next GDN (gated delta-net) **prefill** is a four-kernel chain today:
  `causal_conv1d_split_qkv` → fused gating → the fp32 chunked delta-rule scan →
  gated RMSNorm + per-head group-128 FP8 quant. Four launches means the large
  intermediate tensors (split qkv, conv output, gated core) round-trip through HBM
  between every step.

  This PR adds a single fused Gluon op, `fused_gdn_prefill_qkvz`, that folds that
  work intra-kernel: split + conv + SiLU + QK-norm + gating-prep in one launch, the
  fp32 chunked scan, and a gated-RMSNorm + group-128 FP8-quant **epilogue**. The win
  is the intra-launch fusion and the dropped intermediate HBM traffic. The FP8
  activations a block-FP8 `out_proj` consumes are emitted by the epilogue launch, so
  there is no separate quantization kernel.

  This is the **prefill sibling of the decode kernel** (ROCm/aiter#5520): same
  Qwen3-Next GDN math, but it consumes ragged variable-length batches
  (`cu_seqlens` + `has_initial_state`) and runs the fp32 chunked scan instead of the
  single-token recurrent update.

  ### Design
  - **Reported, not asserted.** `fused_gdn_prefill_qkvz_supported(...) -> (bool, reason)`
    gates on **gfx950** (CDNA buffer addressing / inline asm / explicit register
    layouts) and **Triton ≥ 3.8** (the tiles are the Gluon dialect; some ROCm 3.7
    builds import `triton.experimental.gluon` but fail to compile it, so the gate is a
    version check, not just an import probe). Callers fall back to the four-kernel
    chain instead of crashing.
  - **Lazy tile import** on the dispatch path, so the module and the support predicate
    stay importable on any platform (older Triton, non-gfx950, CUDA).
  - Prefill spans a wide token range, so it dispatches one of **four autotuned M-tile
    schedules** over the covered region (`tokens ∈ [1024, 16384]`, `batch ∈ [1, 64]`).
  - Constraints inherited from the epilogue: `head_k_dim == head_v_dim == 128`, group
    size 128, `num_v_heads == 2·num_k_heads`.

  ```python
  (normalized_bf16, conv_state, delta_state, quantized_fp8, scales) = fused_gdn_prefill_qkvz(
      projected_qkvz, projected_ba, conv_state, delta_state,
      cache_indices, cu_seqlens, has_initial_state,
      conv_weight, conv_bias, a_log, dt_bias, norm_weight,
      scale=head_k_dim ** -0.5, eps=norm.eps,
  )   # conv_state / delta_state advanced in place
  ```

  ### Files changed (9 files, +4655)
  | File | + | What |
  |------|---|------|
  | `aiter/ops/triton/gated_delta_net/fused_gdn_prefill_qkvz.py` | 308 | Public op + `_supported()` gate + tile
  dispatch (`_select_tile_key`) |
  | `aiter/ops/triton/_gluon_kernels/gfx950/gated_delta_net/fused_gdn_prefill_qkvz/` | 3761 | 4 Gluon M-tile schedules
  (`_tile_m1024_3071`, `_tile_m3072_16384`, `_tile_m12289_16384_b1_5`, `_tile_m12289_16384_b6_15`) + package `__init__`
  |
  | `aiter/ops/triton/gated_delta_net/__init__.py` | 6 | Export `fused_gdn_prefill_qkvz` / `_supported` |
  | `op_tests/triton_tests/test_fused_gdn_prefill_qkvz.py` | 416 | Correctness vs pure-torch fp32 reference |
  | `op_tests/op_benchmarks/triton/bench_fused_gdn_prefill_qkvz.py` | 164 | Latency benchmark |

  ### Accuracy — run command
  Correctness is checked against a **pure-torch fp32 sequential-scan reference**
  (deliberately *not* the chunked algorithm the kernel uses, so agreement is
  meaningful), matched to the kernel's bf16 rounding boundaries. Needs a gfx950 +
  Triton ≥ 3.8 ROCm environment.

  ```bash
  # from the aiter checkout, on an MI350-series (gfx950) host:
  python3 -m pytest -q op_tests/triton_tests/test_fused_gdn_prefill_qkvz.py
  ```
  Result on MI355X (Triton 3.8): **9 passed** — all 4 M-tiles compile+run+match the
  reference, both `has_initial_state` paths, determinism, and the unsupported-reason
  path. Numerics: fp8 activation ~1.3–7% (fp8 level), `conv_state` exact, recurrent
  state ~0.4%.

  ### Perf

  **Kernel vs the four-kernel chain it replaces** (MI355X, num_k=4 / num_v=8 /
  head=128; latency ms, median; speedup of the fused kernel):

  | shape k:v:b:s |     m | 4-kernel chain (ms) | fused (ms) | speedup |
  |---------------|------:|--------------------:|-----------:|--------:|
  | 4:8:1:4096    |  4096 |               0.691 |      0.222 |   3.1×  |
  | 4:8:2:4096    |  8192 |               0.834 |      0.302 |   2.8×  |
  | 4:8:4:2048    |  8192 |               0.782 |      0.287 |   2.7×  |
  | 4:8:1:8192    |  8192 |               0.867 |      0.324 |   2.7×  |
  | 4:8:1:2048    |  2048 |               0.616 |      0.215 |   2.9×  |
  | 4:8:2:1024    |  2048 |               0.677 |      0.176 |   3.8×  |
  | 4:8:8:1024    |  8192 |               0.744 |      0.209 |   3.6×  |
  | 4:8:2:8192    | 16384 |               1.072 |      0.509 |   2.1×  |

  (2.1–3.8× faster; widest at small-seq/large-batch where launch overhead dominates,
  narrowest at m=16384 where it is compute-bound.)

  **End-to-end through sglang** (Qwen3-Next-80B FP8, MI355X, TP4), prefill-focused
  (ISL 6144 / OSL 8) so TTFT surfaces the prefill win:

  | metric | stock | fused | Δ |
  |--------|------:|------:|---|
  | Median TTFT (ms) | 647.9 | 471.6 | **−27%** |
  | Input throughput (tok/s) | 92.8k | 125.4k | **+35%** |
  | GSM8K | 0.955 | 0.950 | parity |

  A balanced 1k/1k concurrency sweep (1→32) is TTFT-neutral-to-better (−12–16% at low
  concurrency, −2–6% at high), decode-neutral (TPOT identical), GSM8K parity
  (0.953 / 0.950). The sglang consumer wiring is a companion PR.


karverma-amd:
Thanks for the review — rebased onto current `main` (now mergeable) and worked through the findings. On-device validated on MI355X (Triton 3.8): the op tests pass 9/9 across all four M-tiles, `black` + `ruff` clean.

**Addressed**
- **Support-gate dispatch-safety** (the main cluster): `fused_gdn_prefill_qkvz_supported` now validates the full contract the fixed-stride tiles assume, so a mismatched call falls back instead of reading out of bounds — exact baked topology (k=4/v=8, head_dim 128), packed projection widths (`[T,3072]`/`[T,16]`), `conv_weight [2048,4]` and `conv_bias [2048]` shapes, contiguity + single-device for every consumed tensor, and `a_log`/`dt_bias`/`norm_weight` shape+dtype. The three parameter checks are keyword-only so the existing `(…, conv_bias, quant_dtype)` positional contract is unchanged.
- **Import-safety**: architecture detection is imported and called lazily inside the gate and any detection failure is reported as unsupported, so the module stays importable on CPU-only / non-ROCm workers.
- **`make_kernel_repr`**: added to every launchable Gluon entry across the four tiles (including the variable-grid / selection-launched ones) with its constexpr schedule keys.
- **torch-free kernel modules**: moved the torch/triton host orchestration out of each `_gluon_kernels` tile into a per-tile launcher next to the wrapper (`_gdn_prefill_launch_<key>.py`); the tiles are now Gluon-only. `_load_tile` imports only the dispatched tile's launcher, preserving the lazy/import-safe behavior.
- **Conventions**: absolute imports for the new tile package; module-level CUDA-skip marker on the test; corrected the "single launch" docstring and added a full Args/Returns/state-contract section. The four tile modules were also run through Black (they predated the repo's Black pre-check), which accounts for the bulk of the tile diff — no logic change.

**Intentionally not changed (following the nearest sibling rather than the suggestion)**

- **Tuning in JSON**: the closest sibling, `_gluon_kernels/gfx950/chunk_delta_attn/flash_kda_k1.py`, keeps its schedule constants in-source (`_NUM_WARPS`, launch params) rather than in the config tree, so I kept these tiles consistent with it. Happy to externalize if you'd prefer a different convention here.
- **Test directory**: the sibling decode test `op_tests/triton_tests/test_fused_kda_decode_unified.py` is flat under `triton_tests/` (not in a `gated_delta_net/` subdir), so I left this test flat to match.

vgokhale:
# Review: PR #5606 — Add fused Qwen3-Next GDN prefill (gfx950 Gluon) with fp8 group-quant epilogue

Thanks for this. I went through all 13 files. Overall the kernel quality, safety engineering, and test strength are high; my substantive concerns are structural (rules 4 and 5) and are things I'd like an explicit maintainer decision on rather than a silent pass. Details below, organized by the repo's kernel rules first, then the broader points.

## Rule-by-rule

**1. repr decorator — pass.** Every *launched* Gluon kernel across all four tiles carries a config-aware `@gluon.jit(repr=_<name>_repr)`; I cross-checked the `[grid](...)` launch sites against the repr'd kernels and they match. The bare `@gluon.jit` device helpers (`_add`, `_inverse`, `_matrix_product`, `_load_packed_operand`, …) are inlined and correctly omit repr.

**2. unit tests — pass (strong).** `test_fused_gdn_prefill_qkvz.py` validates against an *independent* pure-torch fp32 sequential-scan reference (deliberately not the chunked algorithm — the right call for catching a bug in the algorithm under test), with fail-hard `torch.testing.assert_close` on norm / conv / delta and the dequantized fp8 activations, parametrized to hit all four tiles and asserting each dispatches to the correct tile. Determinism (bitwise, `atol=0`) and a support-gate test are included. One ask: the two heavy 12289-token tiles are slow-marked — please confirm CI actually runs them so they don't rot.

**3. benchmarks — pass.** `bench_fused_gdn_prefill_qkvz.py` is present with a proper bytes-moved model. It would be stronger if it also timed the four-kernel baseline this replaces, since the entire justification is dropped intermediate HBM traffic — a side-by-side would make the win concrete.

**4. right folder + config for tuning — partial.** Folder placement is correct. But there is no config JSON: "tuning" is four frozen tile files selected by a hardcoded `_select_tile_key(tokens, batch)`, so this does not use the config-driven tuning system (`configs/<arch>/…/DEFAULT.json`) the rule points at. I recognize the per-band tuning here is *structural* (different algorithms/schedules), not just tile-size params, so it genuinely can't be reduced to JSON the way a GEMM tile can — but it's still a deviation worth acknowledging explicitly.

**5. no duplicate kernels — the main tension.** Two parts:
- *Fused vs. the chain it replaces:* fine. The general `causal_conv1d_split_qkv` / `fused_gdn_gating` / `chunk_gated_delta_rule` / `gated_rmsnorm` kernels remain for fallback; this adds a specialized fused path (like flash-attn vs. unfused attention), not a duplicate.
- *Internal:* the four tile files (907 + 1235 + 821 + 1425 ≈ 4,400 lines) are four implementations of the *same* GDN-prefill math. To be fair, they are **not** copy-paste — each has entirely different function names and decompositions (`_four_diagonal_inverses`/`_propagate_segments` vs `_build_group_maps`/`_recurrence_compact` vs `_solve_diagonal_blocks`…), i.e. distinct byte-faithful tuned schedules ported per token-band. So it isn't lazy duplication. But under the rule's wording ("same or almost same functionality"), four independent implementations of identical math is exactly the case it's meant to flag. See the structural note below.

**6. no local activation helpers — soft pass.** SiLU / sigmoid / softplus are inlined via Gluon primitives (`gl.div_rn(x, 1+libdevice.exp(-x))`, `1/(1+gl.exp(-x))`, `gl.where(av<=20, gl.log(1+gl.exp(av)), av)`), not defined as local reusable functions, so there's nothing to move to utils — and the existing `activation.py` helpers are triton (`tl.`), unusable from `@gluon.jit`. Mild note: the same activation math is re-inlined in each of the four tiles, which compounds the point above and argues for a shared Gluon activation-utils module down the line.

**7. succinct comments — pass.** No oversized (>4-line) inline comment blocks in any tile file. Module/function docstrings are detailed but justified for a kernel this intricate.

## Broader assessment

**Engineering quality is high.** Support is reported, not asserted (`fused_gdn_prefill_qkvz_supported` gates gfx950 + Triton ≥ 3.8 + an exact shape/dtype contract); Gluon/HIP imports are lazy so the module stays importable on any platform; arch detection is guarded so it can't raise at import; tile modules are strictly torch-free with host orchestration split into separate launchers; and the hardcoded Qwen3-Next topology is protected by a rejecting gate rather than reading out of bounds. This is careful, production-minded work.

**Two things I'd like resolved before merge:**

1. **The four-implementation maintenance surface (rules 4 + 5 together).** 4,400 lines of four *structurally different* implementations of the same delta-scan math is a large, risky surface: a math fix or spec change has to be re-derived four ways, correctness argued four times, and the variants can silently drift. The perf case for freezing per-band tuned schedules is real, but could you share (a) roughly how much perf is won versus a single parametrized Gluon kernel that accepts some loss, and (b) how the four variants will be kept in sync as the GDN math evolves? I'd like this to be a conscious, documented decision rather than an implicit one.

2. **It isn't wired into a caller yet, and the fallback is caller-enforced.** `fused_gdn_prefill_qkvz` is exported, but nothing in-tree calls it, and on an unsupported shape the wrapper *raises* `ValueError` — the docstring pushes the "check `_supported`, else use the four-kernel chain" responsibility onto the caller. So "replaces the four-kernel path" is aspirational for now (no in-repo switch), and a future caller that skips the gate will crash rather than fall back. That's a reasonable scoping (kernel PR separate from the integration), but please state it explicitly and confirm the consuming integration will use the gate.

**Specialization, eyes-open.** ~6,800 lines serving exactly one model config (K_HEADS=4, V_HEADS=8, HEAD_DIM=128, 2048 conv channels, 3072-wide qkvz). Any topology change needs new kernels. A deliberate, safely-gated perf trade — just noting this is a single-shape asset, not a general GDN prefill.

## Verdict

Rules 1, 2, 3, 6, 7 pass — 2 and the safety engineering notably well. Rule 4 is a partial deviation (tuning baked into source + hardcoded dispatch, no config JSON) and rule 5 is genuinely stressed by the four same-math implementations. Both are defensible for a perf-critical, ported Gluon fused op, but both deserve an explicit call rather than a silent pass. I'm in favor of the kernel on quality and test strength, conditioned on: (a) a stated rationale for four frozen variants vs. one parametrized kernel plus a sync/ownership plan; (b) confirmation the two slow-marked tiles run in CI; and (c) clarity that the fused path isn't yet adopted in-tree and the fallback is the caller's responsibility.

karverma-amd:
Thanks — addressed all three in the latest push. Summary below; code/doc changes in
`fused_gdn_prefill_qkvz.py` (module docstring) and `test_fused_gdn_prefill_qkvz.py`.

## 1. Four-tile maintenance surface

**(b) How they stay in sync.** The four schedules aren't four hand-derived rewrites
of the delta-scan math — they're ported autotuner output (the Artemis MI355 v1
kernel-pack), one specialization per token/batch band. They share a single contract:
for any covered shape the result must match one `ref_gdn_prefill` reference, and
`test_matches_reference` parametrizes **every** tile against it. A math/spec change
lands in that one reference; CI then fails any tile that drifts. So the source of
truth is the reference, not four parallel edits — I've written this into the module
docstring ("Keeping the four tiles in sync"). If a future math change is large enough
that re-porting four tiles is a real burden, the honest fix is to re-run the
upstream autotuner and re-import, which is how the four were produced in the first
place.

**(a) Perf vs a single parametrized kernel — measured.** First, correctness: the four
tiles are numerically **interchangeable** — running any tile off its band matches the
dispatched tile to a norm-relative error of ~1e-5…7e-5 (bf16/fp8 accumulation noise from
a different chunk schedule, well inside the tolerance the op tests already accept). So a
single kernel across the whole range is *viable*; the split is a perf choice, not a
correctness one. Then I forced the most general tile (`m3072_16384`, already the
dispatch's wide-region fallback) across the sweep and timed it against the per-band
dispatch on identical inputs (gfx950, CUDA-graph replay, median of 5, two runs, <1%
spread):

```
  b  seqlen  tokens          dispatch  disp_us single_us  loss%   norm-rel vs dispatch
  1    1024    1024        m1024_3071    91.7     74.9    -18.5    1.3e-05
  1    2048    2048        m1024_3071   141.6    105.3    -25.5    7.5e-05
  1    3072    3072       m3072_16384   129.3    129.3     0.0    (same tile)
  3    2048    6144       m3072_16384   203.4    203.3    -0.1    (same tile)
  1    8192    8192       m3072_16384   272.4    271.7    -0.2    (same tile)
  1   12288   12288       m3072_16384   378.6    378.9     0.0    (same tile)
  1   16384   16384 m12289_16384_b1_5   455.4    486.1    +6.7    3.7e-05
  4    3072   12288       m3072_16384   384.7    382.7    -0.5    (same tile)
  8    1600   12800m12289_16384_b6_15   302.0    292.3    -3.2    3.3e-05
  3    4096   12288       m3072_16384   364.5    366.4    +0.5    (same tile)
```

Reading it:

* **Most of the covered space is already one kernel.** `m3072_16384` is what the
  dispatch selects for the whole 3072–12288 band and for 12289+ at batch ≥16 — 6 of the
  10 sampled shapes. So "four kernels" is really *one general kernel + three edge
  specializations*; there's no dispatch cost over the bulk of the range.
* **The specialization win is bounded: ≤ ~7%.** Forcing the general tile onto the top of
  the range costs +6.7% at 16384 (vs `m12289_16384_b1_5`) and +3.2% at the batched 12800
  case. That is the perf the extra tiles actually buy on these equal-length shapes.
* **But the split still earns its place on *ragged* prefill, which is why I'm not
  collapsing it.** A deeper sweep with variable-length seqlens (realistic prefill, not
  equal-split) flips the winner: `m3072_16384` wins the ragged mid-range by up to ~19%,
  while `m12289_16384_b1_5` wins equal-length high-token. No single tile is universally
  best, and the dispatch keys only on `(tokens, batch)`, not raggedness — so a
  blanket single kernel would regress real ragged workloads, not just cost ≤7% at the top.

So the honest tradeoff: the tiles are interchangeable for correctness, and consolidating
to one would shed ~3 files but (i) give up ≤7% at the high-token equal-length points and
(ii) regress ragged mid-range by up to ~19%, which the dispatch can't recover because it
can't see raggedness. That's why I kept the autotuner's per-band split. If you'd still
prefer a single parametrized kernel that eats that loss for a smaller surface, it's a
clean follow-up (the interchangeability result means it's safe to try) — happy to, but
it's a design change, not a correctness fix.

## 2. Slow-marked tiles rotting

Confirmed they run — and I removed the misleading mark. aiter CI invokes
`pytest -v op_tests/triton_tests/` with only explicit per-case `--deselect`; there is
no `-m "not slow"` filter anywhere, and `pytest.mark.slow` isn't used by any other test
in the repo, so the mark deselected nothing and just implied these might be skipped.
The two 12289-token cases are now unconditional params, which is what actually
guarantees CI executes them. Updated the comment to say so explicitly.

## 3. Integration clarity / ValueError vs fallback

Made the contract explicit in the module docstring ("Not adopted in-tree yet"):
aiter's own GDN prefill path does not call this op; it's for an external caller
(sglang) to opt into, and the fallback to the four-kernel chain is the **caller's**
responsibility. The intended flow is probe `fused_gdn_prefill_qkvz_supported(...)` →
run the fused op on `True`, run the existing chain on `(False, reason)`. The
`ValueError` in `fused_gdn_prefill_qkvz` only fires for a caller that skipped that
probe — it flags misuse, it is not a shape-dispatch fallback — so a supported-gated
caller never hits it.

vgokhale:
# Follow-up on the four-tile structure (rule 5)

Thanks for the sync-note and CI changes — the slow-tile coverage and the fallback/in-tree contract are both genuinely addressed. I want to sharpen my earlier note on the four tile schedules, because I framed it wrong the first time and the actual ask is much cheaper than "rewrite it as one kernel."

## What I'm *not* asking for

- Not fewer lines. The four schedules are byte-faithful ported autotuner output; keep the bodies verbatim, keep the per-band codegen, keep the perf.
- Not merging the algorithms into one function. They're structurally different (I checked — only `_multiply` and `_update_conv_state` names overlap across tiles; everything else is unique), and hand-weaving four generated programs together would risk drifting from the byte-faithful port.

So the perf/codegen concerns I raised before don't apply, and I withdraw the "this can't be config-driven / four files are forced" framing — that was too strong.

## What rule 5 is actually about

The point of "no duplicate kernels" isn't LOC — it's a **single canonical place to look** for a given op. Today "the fused GDN prefill kernel" is spread across eight files: four tile bodies in the `fused_gdn_prefill_qkvz/` subpackage, plus four separate host launchers (`_gdn_prefill_launch_m*.py`) sitting next to the wrapper. To understand or change the op you have to know to open eight files across two directories.

The dispatch doesn't need that fragmentation. The selector is already `constexpr`-able (or config-JSON-able): a `TILE: gl.constexpr` branch — or the existing `_select_tile_key` picker — chooses the schedule with the untaken branches dead-code-eliminated at JIT time, so each specialization compiles identically to today. Zero perf cost.

## The cheap fix

Consolidate, don't merge:

- Put the four tile bodies in **one** tile module (each schedule stays its own set of functions, verbatim).
- Put the four host launchers in **one** launcher module.
- Keep `_select_tile_key` as the picker, now dispatching among functions in those modules.

Same LOC, same kernels, same perf — but "the kernel" and "its host orchestration" each become one file where all four schedules sit side by side. That's the "single place to look" the rule is after, and it also naturally restores the rule-4 story (the schedule becomes a config/constexpr selection rather than a hardcoded four-file layout).

karverma-amd:
> # Follow-up on the four-tile structure (rule 5)
> Thanks for the sync-note and CI changes — the slow-tile coverage and the fallback/in-tree contract are both genuinely addressed. I want to sharpen my earlier note on the four tile schedules, because I framed it wrong the first time and the actual ask is much cheaper than "rewrite it as one kernel."
> 
> ## What I'm _not_ asking for
> * Not fewer lines. The four schedules are byte-faithful ported autotuner output; keep the bodies verbatim, keep the per-band codegen, keep the perf.
> * Not merging the algorithms into one function. They're structurally different (I checked — only `_multiply` and `_update_conv_state` names overlap across tiles; everything else is unique), and hand-weaving four generated programs together would risk drifting from the byte-faithful port.
> 
> So the perf/codegen concerns I raised before don't apply, and I withdraw the "this can't be config-driven / four files are forced" framing — that was too strong.
> 
> ## What rule 5 is actually about
> The point of "no duplicate kernels" isn't LOC — it's a **single canonical place to look** for a given op. Today "the fused GDN prefill kernel" is spread across eight files: four tile bodies in the `fused_gdn_prefill_qkvz/` subpackage, plus four separate host launchers (`_gdn_prefill_launch_m*.py`) sitting next to the wrapper. To understand or change the op you have to know to open eight files across two directories.
> 
> The dispatch doesn't need that fragmentation. The selector is already `constexpr`-able (or config-JSON-able): a `TILE: gl.constexpr` branch — or the existing `_select_tile_key` picker — chooses the schedule with the untaken branches dead-code-eliminated at JIT time, so each specialization compiles identically to today. Zero perf cost.
> 
> ## The cheap fix
> Consolidate, don't merge:
> 
> * Put the four tile bodies in **one** tile module (each schedule stays its own set of functions, verbatim).
> * Put the four host launchers in **one** launcher module.
> * Keep `_select_tile_key` as the picker, now dispatching among functions in those modules.
> 
> Same LOC, same kernels, same perf — but "the kernel" and "its host orchestration" each become one file where all four schedules sit side by side. That's the "single place to look" the rule is after, and it also naturally restores the rule-4 story (the schedule becomes a config/constexpr selection rather than a hardcoded four-file layout).

---
  Done  @vgokhale  — consolidated in 43d13def3.

  - One tile module: _gluon_kernels/gfx950/gated_delta_net/fused_gdn_prefill_qkvz.py (the four _tile_*.py + the package
  __init__ collapsed into it, same dotted import path).
  - One launcher module: gated_delta_net/_gdn_prefill_launch.py (the four _gdn_prefill_launch_<key>.py collapsed into
  it).

  All four schedules now sit side by side in each file, clearly banner-delimited. _select_tile_key is unchanged and
  still the dispatcher; _load_tile just imports the single launcher module and picks the per-key entry. Kernels and
  host bodies are byte-identical to before, so same kernels / same perf — net −93 lines, only from dropping the
  duplicated per-file headers/docstrings/precision-config blocks.

  The one place I couldn't keep names literally verbatim is where a shared namespace forced a collision — flagging so
  it's visible in review:
  - Tiles: m12289_16384_b6_15's _multiply and _update_conv_state (+ its _repr) get a _b6_15 suffix (they collided with
  b1_5 / m1024_3071); every other schedule keeps its names bare and unchanged.
  - Launchers: the identical FP8PrecisionConfig + FP8_E4M3_FN are hoisted once (with a _PrecisionConfig alias so
  m1024_3071's body stays verbatim); each schedule's public entry becomes gdn_prefill_group_fp8_quant_<key> so all four
  coexist.

  Every other line of every schedule is unchanged. Re-validated on MI355X (gfx950, Triton 3.8): all four schedules
  dispatch to the right entry and match ref_gdn_prefill across the full test_fused_gdn_prefill_qkvz.py case set,
  including the renamed b6_15.

vgokhale:
# Follow-up: fold `_gdn_prefill_launch.py` into the wrapper

The four-tile consolidation is great — that resolved the "scattered across eight files" concern cleanly. One structural thing remains: `aiter/ops/triton/gated_delta_net/_gdn_prefill_launch.py` is an extra file outside the canonical layout, which is **kernel** (`_gluon_kernels/gfx950/...`) + **wrapper** (`ops/triton/...`) + **unit test** + **benchmark**. The launcher is a fifth file — it's host (torch/triton) orchestration, i.e. wrapper-layer code, split out of the wrapper.

## Why it was split, and why it doesn't need to be

The reason for the split is import-safety: the wrapper must import and run `fused_gdn_prefill_qkvz_supported()` on any platform (Triton < 3.8, non-gfx950, CUDA) so callers can probe the gate and fall back, but the launch code imports the Gluon kernel module at top level, which isn't importable without `triton.experimental.gluon`. So the Gluon import has to be deferred to dispatch time.

That's a real constraint, but it needs a *lazy import boundary*, not a separate file — and this is already a solved problem in the repo.

## Precedent: the bf16 GEMM (`gemm/basic/gemm_a16w16.py`)

It does gluon-on-gfx1250 with a triton fallback, entirely inside the wrapper, with no launcher module:

```python
_GLUON_SUPPORTED_ARCHS = ("gfx1250",)
def _is_gluon_available():
    return any(a in get_arch() for a in _GLUON_SUPPORTED_ARCHS)
...
backend = "gluon" if _is_gluon_available() else "triton"
...
if backend == "gluon":
    from aiter.ops.triton._gluon_kernels.gfx1250.gemm.basic.gemm_a16w16_persistent import (...)  # lazy, inside the branch
    ...  # buffer allocs, config unpack, splitk, kernel launch — all inline in the wrapper
```

So the pattern is: kernels in `_gluon_kernels/gfx<arch>/…`; the wrapper branches on `get_arch()`, and on the supported arch it **lazily imports the Gluon kernels inside the branch** and does the buffer-alloc + launch **inline**. The wrapper stays import-safe on non-gluon platforms because the Gluon import only fires when the gluon branch runs. There is no `*_launch` file.

That gives the exact import-safety guarantee `_gdn_prefill_launch.py` was created for, without the extra file.

## Suggested change

Fold the launcher into the wrapper following `gemm_a16w16`:

- Gate the gluon path on `get_arch() == "gfx950"` (mirroring `_is_gluon_available()`), keeping the existing `_select_tile_key` as the per-(tokens, batch) picker.
- Move the four `gdn_prefill_group_fp8_quant_<key>` launch routines into the wrapper module as helper functions, each doing its Gluon import lazily inside the function (as the wrapper already does for `_arch_supported`).
- Delete `_gdn_prefill_launch.py`.

The only difference from `gemm_a16w16` is volume — this op's orchestration is ~1,250 lines (four schedules × a 6-launch pipeline) versus ~60 lines for the single-launch GEMM — so the wrapper grows to ~1,700 lines. That's fine; the launch routines stay as separate helper functions in the same module, selected by the arch/`_select_tile_key` branch. End state is the canonical four files — kernel (gluon/gfx950) + wrapper (dispatch + inline launch helpers) + test + bench — matching how the bf16 GEMM already handles gluon-vs-triton.

PR #5619, opened 2026-09-17
[Triton/Gluon] fused KDA decode: order conv_state shifts after the tap loads

Boss2002n:
The conv1d taps read  and the shift then writes that same set of positions, in place
Both decode kernels now hoist all three shifts behind the tap loads and bracket them with tl.debug_barrier().


aiter/ops/triton/_triton_kernels/gated_delta_rule/decode/fused_conv_recurrent_norm_unified.py
```diff
@@ -301,6 +305,7 @@ def _fused_kda_decode_unified_kernel(
                     p_csv + j * stride_cs_pos, tl.load(p_csv + (j + 1) * stride_cs_pos)
                 )
             tl.store(p_csv + (W - 2) * stride_cs_pos, b_x_v.to(p_csv.dtype.element_ty))
+        tl.debug_barrier()
```
omuhamma: I dont think you need this barrier, the stores above are not accessed by the loads below

omuhamma (dismissed):

aiter/ops/triton/_triton_kernels/gated_delta_rule/decode/fused_conv_recurrent_norm_unified.py
Boss2002n: nah uh, we need it for the next iters

omuhamma (approved):

azaidy (approved):
LGTM!

PR #5622, opened 2026-09-17
[FlyDSL] Bind comm-fused GEMM2 to the sorted-inter layout used by stage2

charlieguo1106:
## Summary
- Comm-fused GEMM2 always remapped packed sort IDs to route-major `token * topk + slot`. That is wrong when ordinary stage2 is a `flydsl_moe2_layout_*` kernel: `fused_moe` then sets `skip_inter_quant` and stage1 writes sorted-row intermediates, so GEMM2 was reading the wrong A rows.
- Host now sets `sorted_input` from the ordinary stage2 kernel name (`"_moe2_layout_" in kernelName`, same predicate as `skip_inter_quant`). The producer drops the row resolver in that case, and the dispatcher allows a composed GEMM2 without a resolver so it uses natural sorted rows.
- Tests and the tuner build matching stage2 inputs so DSV4-Pro TP8 buckets that mix layouts (M=1/4/8 sorted, M=2/16 route-major) stay consistent.

## Test plan
- [ ] `black --check` and `ruff check` on the six changed files
- [ ] Layout predicate vs `get_2stage_cfgs(...).skip_inter_quant` for M=1,2,4,8,16 on DSV4-Pro TP8 (`7168/384/384/6`) — M=1/4/8 sorted, M=2/16 route-major
- [ ] 8-GPU `op_tests/multigpu_tests/test_comm_fused_moe.py` per token bucket (1, 2, 4, 8, 16), uniform + skew, eager + graph, plus the M=3→4 `fused_moe` padding path
- [ ] On base, M=4 ordinary stage2 still fails (`finite=False`, `max_abs=nan`); on this branch it passes

## Note
Same-process multi-bucket runs can still HSA-abort (`0x1016`). That is pre-existing (v2-layout ordinary stage2 vs a second comm-fused runner) and is out of scope here. Run those tests one token bucket per process.

## Result

main vs cguo/fix_dsv4 at M=1/2/4/8 — **_same speed._**

comm-fused Stage2 latency (µs, max over 8 ranks) — 100 warm-up + 500 timed iterations per run, 3–5 runs per cell. Two data groups per configuration: group 1 = eager, group 2 = graph, each for both routes.

How to read it. M=2 is the yardstick: there both branches execute byte-identical code, and it still spans −1.9% … +6.5% (min-based). The sorted buckets M=1/4/8 span −0.6% … +6.7% — the same band. No configuration shows the branch measurably slower than main.

Don't read the −15.7% / −8.7% at M=1 skew as the fix being faster: those come from main-side runs spiking high. Both sides spike occasionally (main M=8 skew eager over five runs: 27.6 / 27.6 / 68.4 / 26.8 / 27.4; branch M=1 uniform eager: 27.5 / 40.1 / 27.8), with the ordinary path normal in those same runs — a straggler-rank artifact of max over 8 ranks, independent of branch. Hence the min column is the more robust one.


aiter/ops/flydsl/comm_fused_moe_host.py
```diff
@@ -407,6 +407,14 @@ def _barrier(tensor, flat_base, ready_offset, tp_size, stream) -> None:
     )
 
 
+def _bind_inter_layout(runner, ordinary_stage2) -> None:
+    keywords = getattr(ordinary_stage2, "keywords", None) or {}
+    name = keywords.get("kernelName") or keywords.get("kernelName2") or ""
+    sorted_input = "_moe2_layout_" in str(name)
```
binding7012, line 413: reuse skip_inter_quant ?

yifehuan (commented):
LGTM, and please run the TP8 performance benchmarks for M=1, 2, 4, and 8, and add the results to the PR description? A before/after table with the selected
  intermediate layout and latency would help confirm that this fix does not introduce a performance regression.

charlieguo1106:
Result updated in the first comment.

coderfeli (approved):

PR #5423, opened 2026-09-11
[Triton/Gluon] Route test print() output through the aiter logger

Boss2002n:
print()s across 14 test files flooded `pytest -s`; they now go through aiter.logger
test_fav3_sage/test_mha also stop configuring the root logger at DEBUG.

PR #2886, opened 2026-04-23
[TRITON] Conv Kernels First Commit to AITER

saeid-rostami:
Adds a Triton conv2d library targeted at AMD RDNA GPUs, plus a
  correctness + benchmark harness that compares against PyTorch / MIOpen. 

  ## Motivation

  PyTorch on AMD goes through MIOpen, whose hand-tuned solvers cover some
  dtype × layout × architecture combinations well and others poorly — bf16 in
  particular falls back to direct/GEMM solvers on RDNA4 that are noticeably
  slower at large channel counts. Most modern checkpoints (LLMs, diffusion VAEs)
  ship in bf16, so the gap matters.

  This op takes the opposite approach: a single set of Triton kernels that runs
  **fp16 and bf16 through the same code path**, supports **NCHW and NHWC**, and gets reasonable performance across
  the full matrix without per-architecture hand tuning.

  ## What's added

  **Library** (`aiter/ops/triton/conv/`):
  - `conv2d.py` — public API + shape-driven router
  - `_launch.py` — grid setup + `_select_3x3_method` heuristic
  - `_prepack.py` — LRU-cached weight/input repack
  - `_utils.py` — shape math, dtype/activation enums, tolerance model

  **Kernels** (`aiter/ops/triton/_triton_kernels/conv/`), five families:
  | Family | When it runs |
  |---|---|
  | 1×1 GEMM | `R==1, S==1` |
  | 3×3 cblocked (NCHW) | 3×3, channel-blocked input for coalesced loads |
  | 3×3 NHWC | 3×3 with channels-last input — no input repack |
  | Winograd F(4×4, 3×3) | 3×3, stride=1, dilation=1, `C ≥ 512`, `K ≥ 512`, enough output tiles |
  | General | anything else (5×5, 7×7, dilated, strided) |

  **Test/bench harness** (`op_tests/triton_tests/conv/`):
  - `cli.py` — `--test-mode {edge,random,stability,activations,models,all}`
  - `suite.py` — correctness checking + bench accumulation + result tables
  - `bench.py` — timing + `precompute_miopen_solvers` (subprocess + `MIOPEN_LOG_LEVEL=6`
    to label each PyTorch baseline row with the MIOpen solver it picked)
  - `test_edge.py` / `test_fuzz.py` / `test_models.py` — shape sources
  - `test_pytest.py` — parametrized over fp16/bf16 × nchw/nhwc
  - `_registry.py` — single source of truth for kernel methods (used by CLI,
[...]

PR #5640, opened 2026-09-17
[Triton/Gluon] Add Triton-based Conv3D kernels

saeid-rostami:
## Motivation

This PR adds  Triton Conv3D implementation with optimized
routing for common video-model and VAE workloads.

The public API supports FP16 and BF16 inputs, NCDHW and channels-last-3D
execution, optional bias, asymmetric stride/padding/dilation, and activations.
## Technical Details

The Conv3D router selects between the following kernel families:

- Specialized 1x1x1 Conv3D
- NCDHWc 3x3x3 Conv3D with fused activation packing
- Native NDHWC 3x3x3 Conv3D
- General im2col-free Conv3D
- Winograd F(4x4, 3x3) over the H/W axes
- Winograd with an NCDHWc input path

The implementation also adds:

- Conv3D weight-prepack and Winograd-filter caches
- Layout-preserving NCDHW and channels-last-3D behavior
- Shape-driven production routing
- Complete launch keys containing all three spatial axes
- Tuned gfx1201 launch configurations
- Configuration-table coverage for gfx1200, gfx942, and gfx950, initially
  seeded from the gfx1201 values
- Public export through `aiter.ops.triton`
- Real-model-derived benchmark shape coverage for:
  - Wan2.2 A14B VAE
  - Wan2.2 TI2V-5B VAE
  - LTX-Video 0.9.5 VAE
  - InternVideo2 and Qwen-family video patch embeddings

The benchmark reports per-layer correctness, selected Triton kernels,
kernel-only and repack-inclusive performance, aggregate throughput, and
call-count-weighted encoder/decoder totals. It exits nonzero when a requested
shape errors or fails correctness.

## Test Plan

The Conv3D test suite covers:

- FP16 and BF16
- NCDHW and channels-last-3D layouts
- Every direct Conv3D route
- Optional bias
- `none`, `relu`, `relu6`, and `gelu` activations
- Asymmetric dimensions, stride, padding, and dilation
- Batched, odd-channel, and tail-masking shapes
- Winograd transform and output correctness
- Output dtype preservation
- Weight-cache reuse, invalidation, and LRU eviction
- Actual production-router dispatch
- Benchmark database, reporting, and CLI behavior
- Invalid argument and unsupported configuration handling

## Test Result

Tests were run on an AMD Radeon AI PRO R9700 (`gfx1201`) after clearing the
Triton compilation cache.

```text
Conv3D: 174 passed in 223.69s
```

## Submission Checklist


op_tests/triton_tests/conv/_helpers.py
```diff
@@ -136,10 +136,10 @@ def _direct_3x3_guard(R, S, stride, dilation, C):
 
 
 def _wino_guard(R, S, stride, dilation, C):
-    # _is_winograd_eligible signature varies by upstream — keep the flag tight
-    from aiter.ops.triton.conv._utils import _is_winograd_eligible
+    # _is_winograd_2d_eligible signature varies by upstream — keep the flag tight
+    from aiter.ops.triton.conv._utils import _is_winograd_2d_eligible
 
-    return _is_winograd_eligible(R, S, stride, dilation, C)
+    return _is_winograd_2d_eligible(R, S, stride, dilation, C)
```
brunomazzottiamd: This piece of code seems weird to me for the following reasons:

1. It does a `import` at the function scope. Is it too expensive to do the `import` at module scope? Why are we importing just at calling time? If the import is cheap let's move it to the module scope because some Python internal lookups (to check if a module is already imported) just add up at function calling site.
2. `_wino_guard` has the same signature of `_is_winograd_2d_eligible` and `_wino_guard` calls `_is_winograd_2d_eligible` passing the arguments as-is. Why do we need this indirection? Can we can `_is_winograd_2d_eligible` directly?

aiter/ops/triton/__init__.py
```diff
@@ -109,6 +109,7 @@ def _warn_if_deprecated(name, new_path):
     "fused_gemm_a8w8_blockscale_split_cat": "gemm.fused.fused_gemm_a8w8_blockscale_split_cat",
     # Conv modules (conv/)
```
brunomazzottiamd: To be honest, I don't think this is necessary for both convolution modules (2D and 3D). `_BACKWARD_COMPAT_MAP` was introduced 8 months ago by https://github.com/ROCm/aiter/pull/1638. Conv2D was introduced 2 months ago by https://github.com/ROCm/aiter/pull/2886. `_BACKWARD_COMPAT_MAP` is a way of keeping AITER callers that still rely on the old package structure working. The preferable way is to have callers importing `conv.conv2d` and `conv.conv3d`.

Do you know who's using your AITER convolutions and how the imports are being done? We shouldn't add Conv3D here, and if possible, remove Conv2D. Only remove Conv2D if you are sure no external caller will be affected. If external callers are going to be affected, please work with them to port their code to the current package structure so we can remove Conv2D in the future.

aiter/ops/triton/conv/README.md
```diff
@@ -118,34 +118,56 @@ and seed under both backends, only VAE convs swapped to Triton): max diff
...
+  `"reflect"`, `"replicate"`, and `"circular"` padding modes to PyTorch/MIOpen.
 - Inputs must be `fp16` or `bf16`.
 - Forward only (no backward / training).
 
 ---
 
 ## Reproducing the tests and benchmarks
 
-Run from the AITER repo root (`/app/aiter` in this tree, or `PYTHONPATH=/app/aiter`).
+Run these commands from the AITER repository root (`/app/aiter` in this tree).
```
brunomazzottiamd: What is this `/app/aiter` absolute path? We don't have any `app` folder in AITER code base (confirmed with `find . -type d -name app` executed at AITER root).

This seems to be a local setup leaking to AITER docs, it's probably AI-driven.

Please double check if it makes sense in the broader context of `README.md`, otherwise I think we can remove this section.

aiter/ops/triton/conv/README.md
```diff
@@ -179,63 +247,26 @@ Cross-axis flags:
 --no-bias                                     # bench the bias=None code path
 --miopen-solvers                              # detect MIOpen solver names (sweep mode; ~60-120s subprocess)
 --show-kernel-name                            # include routed kernel name in single-shape output
+--model MODEL                                 # select a model workload by name substring
```
brunomazzottiamd, line 245: **Double check before merging:** Do the command line docs reflect the real code?

aiter/ops/triton/conv/README.md
```diff
@@ -179,63 +247,26 @@ Cross-axis flags:
...
-
-## Documentation
-
-- **[`DESIGN.md`](DESIGN.md)** — architecture, per-kernel deep-dive, full
-  Winograd F(4,3) derivation (G/Bᵀ/Aᵀ matrices, 361× amplification
-  analysis, why Winograd is disabled for `C < 4`), the
-  `_select_3x3_method` heuristic, memory layouts and repacking,
-  numerical model, extension guide.
+```
+--dtype {fp16,bf16}                           # default fp16
```
brunomazzottiamd, line 253: **Double check before merging:** Do the command line docs reflect the real code?

aiter/ops/triton/conv/README.md
```diff
@@ -244,20 +275,36 @@ files are shipped with the kernels.
 ```
 aiter/ops/triton/conv/                Kernel library
```
brunomazzottiamd, line 274: **Double check before merging:** Does the filesystem org docs reflect the real filesystem org?

aiter/ops/triton/conv/_prepack.py
```diff
@@ -62,11 +68,27 @@ def put(self, key, src, item):
...
 _PACK_CACHE_WINOGRAD_F4X3 = _LRUPackCache()
 
+# The 3-D packed layouts use independent cache instances to prevent collisions
+# when source tensor metadata matches a cache entry for another packed format.
+_PACK_CACHE_3D_GENERAL = _LRUPackCache()
+_PACK_CACHE_3D_3X3X3 = _LRUPackCache()
+_PACK_CACHE_3D_WINOGRAD_HW = _LRUPackCache()
+
+
+def clear_conv3d_weight_pack_caches():
```
brunomazzottiamd: Why do we have a function to clean up Conv3D weight packs and don't have the same feature for Conv2D weight packs? What's the caller of `clear_conv3d_weight_pack_caches`?

Please evaluate the real need for this feature. If if makes sense, can it be applicable do Conv2D as well?

I think the 2D and 3D implementations should be "symmetrical" (I couldn't find a better word for it...) to keep the code base internally consistent.

aiter/ops/triton/conv/_utils.py
```diff
@@ -60,20 +82,135 @@ def _is_3x3_conv(R, S):
...
+    dd, dh, dw = dilation
+    OD = (D + 2 * pd - dd * (T - 1) - 1) // sd + 1
+    P = (H + 2 * ph - dh * (R - 1) - 1) // sh + 1
+    Q = (W + 2 * pw - dw * (S - 1) - 1) // sw + 1
+    return OD, P, Q
+
+
+def _conv3d_dims(x, w_oidhw, stride, padding, dilation):
+    """Validate NCDHW/OIDHW tensors and return all convolution dimensions."""
+    assert x.is_cuda and w_oidhw.is_cuda
```
brunomazzottiamd: Are you sure this assertion makes sense in the scope of `_conv3d_dims`? As far as I can see, the functions inspects tensor shapes and would work perfectly fine for CPU tensors as well. I'm not against this device check, just wondering if it's the best place to do it...

op_tests/triton_tests/conv/test_conv2d.py
```diff
@@ -131,6 +134,25 @@ def test_cross_method(dtype):
     _assert_suite(suite)
 
 
+@pytest.mark.parametrize("layout", ["nchw", "nhwc"])
+def test_scalar_parameters_and_noncontiguous_input(layout):
+    """Conv2D accepts scalar parameters and materializes sliced inputs."""
+    torch.manual_seed(0)
+    x_base = torch.randn(1, 32, 12, 18, device="cuda", dtype=torch.float16)
+    x = x_base[..., ::2]
+    assert not x.is_contiguous()
```
brunomazzottiamd: Please add messages to your assert statements, use `assert <condition>, "Optional error message"` form. The form with the error message improves debug ability upon test failures.

brunomazzottiamd (commented):
I'm publishing a partial review with some comments.

I'll check the main part in another review round because this PR is too big. I still have to check the following files:

- [ ] `aiter/ops/triton/_triton_kernels/conv/conv3d_1x1x1.py`
- [ ] `aiter/ops/triton/_triton_kernels/conv/conv3d_3x3x3.py`
- [ ] `aiter/ops/triton/_triton_kernels/conv/conv3d_general.py`
- [ ] `aiter/ops/triton/_triton_kernels/conv/conv3d_winograd_hw_f4x3.py`
- [ ] `aiter/ops/triton/_triton_kernels/conv/ncdhw_to_cblocked.py`
- [x] `aiter/ops/triton/conv/conv3d.py`
- [ ] `op_tests/op_benchmarks/triton/bench_conv3d.py`
- [x] `op_tests/triton_tests/conv/test_conv3d.py`

I didn't review any config file, i.e. `aiter/ops/triton/configs/*`, and simply ignored them all.

Next time you must submit smaller PRs that are reviewable. You don't need to throw everything at once into AITER code base, it's a matter of structuring your changes in a logical way with smaller change sets. For instance, you can file a PR with utils and core changes, then one PR for each kernel flavor, then other PRs with the performance tuning and optimizations per GPU arch. Feel free to brainstorm alternative organizations with Claude or your favorite AI-assistant. Check GitHub stacked PRs features, it's very useful: https://docs.github.com/en/pull-requests/get-started/about-stacked-prs.

Please do not force push future changes, squashing everything into the last commit. Publish future changes as distinct commits because this makes further review way easier!

brunomazzottiamd:
@saeid-rostami, as far as I can see your PRing from your fork. Try to publish your feature branch to ROCm repo and then PR the feature branch to `main` branch in the same ROCm repo.

Lots of Triton CI test shards are failing with the following error:

```text
docker: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit.
```

This happens because CI isn't logging to Docker Hub, with CI team credentials. This step is skipped when PRing from a fork:

https://github.com/ROCm/aiter/blob/main/.github/workflows/triton-test.yaml#L183

```yaml
      - name: Docker login
        if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
        env:
          DOCKER_USERNAME: ${{ secrets[needs.ci_config.outputs.docker_username_secret] }}
          DOCKER_PASSWORD: ${{ secrets[needs.ci_config.outputs.docker_password_secret] }}
```

brunomazzottiamd:
> @saeid-rostami, as far as I can see your PRing from your fork. Try to publish your feature branch to ROCm repo and then PR the feature branch to `main` branch in the same ROCm repo.
> 
> Lots of Triton CI test shards are failing with the following error:
> 
> ```
> docker: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit.
> ```
> 
> This happens because CI isn't logging to Docker Hub, with CI team credentials. This step is skipped when PRing from a fork:
> 
> https://github.com/ROCm/aiter/blob/main/.github/workflows/triton-test.yaml#L183
> 
> ```yaml
>       - name: Docker login
>         if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.head.repo.fork }}
>         env:
>           DOCKER_USERNAME: ${{ secrets[needs.ci_config.outputs.docker_username_secret] }}
>           DOCKER_PASSWORD: ${{ secrets[needs.ci_config.outputs.docker_password_secret] }}
> ```

Postponing this since it requires opening a new PR. We can move on to a new branch / PR at ROCm fork if Docker rate limiting becomes a real issue in the future.

aiter/ops/triton/conv/conv3d.py
```diff
@@ -0,0 +1,671 @@
...
+    """
+    if not _is_winograd_hw_3d_eligible(3, 3, 3, stride, dilation, C):
+        return "general" if C < block_c else "cblocked"
+
+    OD, P, Q = _out_dhw(D, H, W, 3, 3, 3, stride, padding, dilation)
+    tile_h = (P + 3) // 4
+    tile_w = (Q + 3) // 4
+    tiles = N * OD * tile_h * tile_w
+
+    if not _is_amd_wave32():
```
brunomazzottiamd, line 85: **I may be wrong**, but I have a feeling that `_is_amd_wave32()` function is used only in `aiter/ops/triton/conv/conv3d.py`. If this is the case, please define `_is_amd_wave32()` in this file to reduce scope. Putting a function in other module just makes sense if we have more than one caller IMHO.

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -0,0 +1,1638 @@
...
+
+from ._helpers import (
+    ALL_SUPPORTED_ARCHS,
+    _winograd_tolerances,
+    apply_activation,
+    dynamic_conv_tolerances,
+)
+
+_current_arch = get_arch()
+if _current_arch not in ALL_SUPPORTED_ARCHS:
```
brunomazzottiamd, line 76: Please make sure we can run these unit tests in `gfx942` and `gfx950` AITER CI runners. They will be a guarantee that future commits won't break Conv3D correctness! Remember that we don't have RDNA runners!

You don't need to tune the kernels for CDNA, just pick any reasonable block size and related meta parameters - be conservative in this regard. I'll be happy if everything is green in CDNA CI (no perf promises at all).

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -0,0 +1,1638 @@
...
+    assert torch.isfinite(y).all()
+    assert torch.isfinite(ref).all()
+
+    if is_winograd:
+        rtol, atol = _winograd_tolerances(dtype, K_red)
+    else:
+        rtol, atol = dynamic_conv_tolerances(dtype, K_red)
+    y32 = y.float()
+    ref32 = ref.float()
+    torch.testing.assert_close(y32, ref32, rtol=rtol, atol=atol)
```
brunomazzottiamd, line 124: Please add the `msg` argument to `torch.testing.assert_close`, it can improve debugging ability when a test fails. You can do something like this:

```python
expected = torch.tensor([1.0, 2.0, 3.0])
actual = torch.tensor([1.0, 4.0, 5.0])

# If msg is a callable, it can be used to augment the generated message with
# extra information
torch.testing.assert_close(
    actual, expected, msg=lambda msg: f"Header\n\n{msg}\n\nFooter"
)

# AssertionError: Header
# 
# Tensor-likes are not close!
# 
# Mismatched elements: 2 / 3 (66.7%)
# Greatest absolute difference: 2.0 at index (1,) (up to 1e-05 allowed)
# Greatest relative difference: 1.0 at index (1,) (up to 1.3e-06 allowed)

# Footer
```

Source: https://docs.pytorch.org/docs/2.14/testing.html

You can provide more context about the tensors being compared, the operand and anything else that helps debugging efforts.

PS: Please do it for all Conv3D tests. You can file Conv2D related  changes in the scope of another PR - this PR already is big enough.

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -0,0 +1,1638 @@
...
+        )
+        reference_scale = (
+            scale_tensor.abs().max().clamp_min(torch.finfo(torch.float32).tiny)
+        )
+        reference_norm = torch.linalg.vector_norm(ref32).clamp_min(
+            torch.finfo(torch.float32).tiny
+        )
+        normalized_max_error = error.max() / reference_scale
+        relative_l2_error = torch.linalg.vector_norm(error) / reference_norm
+        assert normalized_max_error < _WINOGRAD_NORMALIZED_MAX_ERROR
```
brunomazzottiamd: Add error messages to Python `assert` statements. This helps debugging when a test fails.

PS: Please do it for all Conv3D tests. You can file Conv2D related changes, for what's already merged into `main` branch, in the scope of another enhancement PR - this PR already is big enough.

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -0,0 +1,1638 @@
...
+from aiter.ops.triton.utils.conv_config_utils import (
+    format_prepack_shape_key_3d,
+    format_shape_key_3d,
+)
+from op_tests.op_benchmarks.triton import bench_conv3d
+from op_tests.op_benchmarks.triton.bench_conv3d import (
+    _DEFAULT_MODEL,
+    EDGE_CASE_SHAPES,
+    _format_single_shape_line,
+    _get_miopen_solver,
```
brunomazzottiamd, line 62: Do we need MIOpen to run this test suite? What's the purpose of having MIOpen here?

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -0,0 +1,1638 @@
...
+    conv3d_winograd_hw_f4x3,
+    conv3d_winograd_hw_f4x3_cblocked,
+)
+from aiter.ops.triton.utils import conv_config_utils
+from aiter.ops.triton.utils._triton.arch_info import get_arch
+from aiter.ops.triton.utils.conv_config_utils import (
+    format_prepack_shape_key_3d,
+    format_shape_key_3d,
+)
+from op_tests.op_benchmarks.triton import bench_conv3d
```
brunomazzottiamd, line 57: Why are you importing the benchmark module in the unit test? Are you trying to test the benchmarking tool behavior? If this is the case, I'd suggest splitting this test file in two:
 
1. `test_conv3d.py` - test Conv3D operator and related kernels
2. `test_bench_conv3d.py` or `test_conv3d_bench.py` - test benchmarking tool behavior

If you are trying to import some shared logic, I'd suggest putting it in a third module that can be imported by the test and by the benchmark tool as well.

brunomazzottiamd (commented):
More comments on the Conv3D wrapper function and Conv3D unit tests.

op_tests/triton_tests/conv/test_conv3d.py
```diff
@@ -398,6 +399,46 @@ def test_weight_prepack_cache_uses_lru_eviction(monkeypatch):
...
+    [
+        (
+            clear_conv2d_weight_pack_caches,
+            ("_PACK_CACHE", "_PACK_CACHE_3x3", "_PACK_CACHE_WINOGRAD_F4X3"),
+            (
+                "_PACK_CACHE_3D_GENERAL",
+                "_PACK_CACHE_3D_3X3X3",
+                "_PACK_CACHE_3D_WINOGRAD_HW",
+            ),
+        ),
```
brunomazzottiamd: Please do not mix Conv2D aspects in Conv3D test suite. Break this parameterization in two and add the related Conv2D aspects to `op_tests/triton_tests/conv/test_conv2d.py`. If the code is identical then add the duplicated part to a test commons / utils / helpers module and call it twice (in `test_conv2d.py` and `test_conv3d.py`).

aiter/ops/triton/conv/conv3d.py
saeid-rostami: I checked the call sites. _is_amd_wave32() is used by both conv2d.py and conv3d.py for architecture-specific route selection, so I kept it in the shared _utils.py module to avoid duplicating the logic.

op_tests/triton_tests/conv/test_conv3d.py
saeid-rostami: No I removed benchmark related stuff from unit test file.

aiter/ops/triton/conv/README.md
saeid-rostami: Checked and fixed the missing parts/

op_tests/triton_tests/conv/test_conv3d.py
saeid-rostami: I think this is because this PR coming from my fork and PR cannot access the Rocm CI, I'll fix this part when all the reviews done.

brunomazzottiamd: PLMK when you're able to push to ROCm fork, then I can give it a quick try locally in some CDNA hosts.

brunomazzottiamd:
I hid all the GitHub Copilot comments. They are very verbose, repetitive, noisy... They just add cognitive pollution to the PR, making it hard to find real productive discussions being held by the author and reviewer. It's the worst kind of AI-slop. I deeply dislike GitHub Copilot comments and would like to have it running **JUST ONCE** before merging (as requested by the people collaborating in this PR) instead of running at every commit that's pushed.

FYI: @Boss2002n + @azaidy

op_tests/op_benchmarks/triton/bench_conv3d.py
```diff
@@ -0,0 +1,1374 @@
...
+            unique.append(key)
+    if not unique:
+        return
+
+    dtype_str = {
+        torch.float16: "torch.float16",
+        torch.bfloat16: "torch.bfloat16",
+    }.get(dtype, "torch.float16")
+    lines = [
+        "import os, sys",
```
brunomazzottiamd: What's the purpose of this function? As far as I can see, it's generating Python code at runtime! This is very hard to maintain in the long run! It smells bad IMHO...

Is it possible to replace this dynamic code generation with regular Python code?

op_tests/op_benchmarks/triton/bench_conv3d.py
```diff
@@ -0,0 +1,1374 @@
...
+        "--conv_layout",
+        "--conv-layout",
+        type=str.lower,
+        choices=["ncdhw", "ndhwc"],
+        default="ncdhw",
+    )
+    parser.add_argument(
+        "--metric",
+        type=str.lower,
+        choices=["time", "throughput"],
```
brunomazzottiamd: Is it possible to implement bandwidth metric (GB/s)?

op_tests/op_benchmarks/triton/bench_conv3d.py
```diff
@@ -0,0 +1,1374 @@
...
+def main(argv: list[str] | None = None) -> None:
+    args = parse_args(argv)
+    if args.single_shape:
+        run_single_shape(args)
+    else:
+        run_sweep(args)
+
+
+if __name__ == "__main__":
+    main()
```
brunomazzottiamd: **General comment about `op_tests/op_benchmarks/triton/bench_conv3d.py`:**

It's very convoluted and complex. 1K+ lines of code for a benchmark may be too much IMHO. I'm not against doing what needs to be done, but 1K+ lines smell bad. Is this benchmark AI generated? Please try to simplify the code and make it more maintainable in the long run. We have very complex kernels like MHA, with lots of implementations and options - in such case the benchmark is complex, it's unavoidable. I'm not a convolution expert and might be wrong. Please evaluate my simplification suggestion according to your expertise.

For instance, most of our benchmarks have less than 0.5K lines of code:

```text
~/Projects/aiter/op_tests/op_benchmarks/triton $ wc -l bench_*.py | sort | head -40
    77 bench_fast_transpose.py
    77 bench_quant_mxfp8.py
    93 bench_moe_gemm_per_token.py
   102 bench_moe_gemm_mxfp8.py
   104 bench_cross_entropy.py
   109 bench_moe_wgrad.py
   115 bench_dsv4_indexer.py
   118 bench_gemm_a16w8_blockscale.py
   122 bench_gemm_a16w16_gated.py
   124 bench_quant_fp8_blockwise.py
   130 bench_fused_gemm_a8w8_blockscale_mul_add.py
   131 bench_sparse_mla_dsv4_train.py
   132 bench_fused_gemm_a8w8_blockscale_a16w16.py
   136 bench_fused_gemm_commons.py
   139 bench_gemm_afp4wfp4_pre_quant_atomic.py
   148 bench_fused_gemm_afp4wfp4_mul_add.py
   150 bench_fused_gemm_afp4wfp4_a16w16.py
   151 bench_fp8_mqa_logits.py
   152 bench_mhc_dsv4.py
   155 bench_gemm_a8w8_per_token_scale.py
   155 bench_gemm_a8wfp4.py
   157 bench_batched_gemm_a8w8.py
   157 bench_fused_sigmoid_mul.py
   171 bench_batched_gemm_afp4wfp4.py
   171 bench_fused_rmsnorm_add.py
   176 bench_batched_gemm_a16wfp4.py
   176 bench_gated_delta_rule_snapshot_dtype.py
   185 bench_gemm_afp4wfp4.py
   187 bench_quant_mxfp4.py
   188 bench_batched_gemm_bf16.py
   188 bench_fused_add_rmsnorm_pad.py
   188 bench_gemm_a8w8.py
   200 bench_batched_gemm_a8w8_a_per_token_group_prequant_w_per_batched_tensor_quant.py
   217 bench_gemm_a8w8_blockscale.py
   218 bench_chunk_delta_attn.py
   223 bench_mla_decode.py
   224 bench_gemm_a16w16_gating.py
   227 bench_fused_kda_decode.py
   239 bench_flash_kda.py
   242 bench_gemm_afp8wfp8.py
```

Take a look at other benchmarks, for some inspiration.

brunomazzottiamd (commented):
I checked all files. Generally speaking the kernels look good! I'm publishing some comments about other pieces of code. Let's resolve every pending discussion, get CI passing and then we can merge.

brunomazzottiamd:
@saeid-rostami, I've noticed that `op_tests/triton_tests/conv/_helpers.py` have some conditional `print` statements. Generally speaking, this is a bad practice and we're trying to get rid of it. Can you please replace these `print` statements with proper logging calls **in the scope of another PR**? Check https://github.com/ROCm/aiter/pull/5423 to get some inspiration and feel free to add me and / or @Boss2002n as reviewers!

Boss2002n:
@saeid-rostami Can we incorporate Bruno's feedback and move each kernel into a different PR. It will make reviewing it easier and faster

brunomazzottiamd:
> @saeid-rostami Can we incorporate Bruno's feedback and move each kernel into a different PR. It will make reviewing it easier and faster

We don't need to do this for this PR, let's move in this direction in the future.

brunomazzottiamd (changes requested):
`ncdhw_to_cblocked` kernel is character-identical to the existing `nchw_to_cblocked`. The only change is the runtime argument HW renamed to DHW. One copy should be deleted and the other adjusted / reused for Conv3D purposes.

Image

brunomazzottiamd (changes requested):
**Conv3D Winograd F(4,3) transform math is a verbatim copy of Conv2D**

`aiter/ops/triton/_triton_kernels/conv/conv3d_winograd_hw_f4x3.py` documents this itself:

```15:17:aiter/ops/triton/_triton_kernels/conv/conv3d_winograd_hw_f4x3.py
Winograd runs only on H,W (large dims); depth stays a direct fp32 sum, so the
numerical amplification is the 2D F(4,3) level (~361x), not cubed. The transform
math is copied verbatim from ``conv_3x3_winograd_f4x3.py``.
```

That copy is real. The 2.5D path reuses the same 2D F(4,3) `BᵀdB` / `AᵀMA` arithmetic and only changes indexing for the extra depth axis. The matching kernels are:

| Conv3D (`conv3d_winograd_hw_f4x3.py`) | Conv2D (`conv_3x3_winograd_f4x3.py`) |
|---|---|
| `_winograd_hw_f4x3_input_transform_kernel` | `_winograd_f4x3_input_transform_kernel` (and the cblocked twin `_winograd_f4x3_cblocked_input_transform_kernel`) |
| `_winograd_hw_f4x3_output_transform_kernel` | `_winograd_f4x3_output_transform_kernel` |

I compared the formula lines: **108/108 input** (`d**` fp32 casts, `t**` column transform, `v**` row transform) and **76/76 output** (`m**` loads, `s**` / `y**` `Aᵀ` transforms) are identical. Around that sits the same 6×6 load unroll, 36 `V` stores, bias/act, and 4×4 store unroll - about **450 lines** of duplicated transform body. The 3-tap depth GEMM (`_winograd_hw_f4x3_batched_gemm_kernel`) is the part that is actually new; that one is fine as its own kernel.

Please share the F(4,3) input/output transform with Conv2D (helpers or a shared module both kernels call) instead of keeping a second copy.

brunomazzottiamd (changes requested):
The same 6x3 F(4,3) `G` literal is in `_prepack.py` twice, with identical entries:

1. **`prepack_winograd_filter_f4x3`** (Conv2D) - lines 185–196
2. **`prepack_winograd_hw_filter_f4x3`** (Conv3D) - lines 311–322

Both are:

```text
[[ 1/4,  0,    0  ],
 [-1/6, -1/6, -1/6],
 [-1/6,  1/6, -1/6],
 [ 1/24, 1/12, 1/6],
 [ 1/24,-1/12, 1/6],
 [ 0,    0,    1  ]]
```

Get rid of the duplicate.

---

I have a feeling that this PR ships a lot of duplicated code. Please use you AI-agent to detect duplicates and refactor the code, to promote some code reuse.

PR #5649, opened 2026-09-17
[Triton/Gluon] [Config] gfx942: fix unified-attention prefill config for head_size >= 512 (Gemma-4 full-attention layers)

zhumengzhiren:
## Summary

Adds one gfx942 `attn_2d` entry so that prefill on layers with `head_size >= 512` (Gemma-4's ten full-attention layers: head_dim 512, 32 q / 4 kv heads) stops running on configs that were never right for that shape:

```json
"D_GEQ_512.Q_GEQ_256": {"BLOCK_M": 64, "num_warps": 4, "num_stages": 2, "waves_per_eu": 2, "TILE_SIZE_MIN": 16, "TILE_SIZE_MAX": 16}
```

Measured on MI325X (gfx942), fp8 q + fp8 paged KV, block 64, one causal sequence, cache flushed before every call, median of 7, same kernel body for every row (config is the only variable):

| tokens | pre-#4044 default (M128 / T64) | current table (M16 / T32) | this entry (M64 / T16) |
|---|---|---|---|
| 4096 | 7.80 ms | 2.95 ms (2.64x) | 2.09 ms (3.74x) |
| 16384 | 115.8 ms | 44.7 ms (2.59x) | 26.0 ms (4.45x) |

End to end in vLLM (`ROCM_AITER_UNIFIED_ATTN`, gemma-4-31B FP8-block, 16k-token cold prefill): the ten layers go from 107 to 24.7 ms each, prefill wall time -32%, prefill throughput +47% in a same-GPU A/B. Output relative error vs. the current config 6e-3 (bf16 output noise); lm_eval IFEval / GSM8K unchanged within noise.

## What is wrong today, version by version

| version | config selected for Gemma-4 fp8 prefill on gfx942 | 16k tokens, per layer | why |
|---|---|---|---|
| before #4044 (<= v0.1.19) | head-256 defaults, BLOCK_M 128 / TILE 64 | 115.8 ms | no head-512 branch. At head 512 the K and V tiles are 32 KB each, so one workgroup owns the CU's whole 64 KB LDS, one workgroup per CU, ~6% MFU |
| #4044 (v0.1.19.post1+) | head >= 512 branch: TILE 16 / stages 2 / warps 4, but BLOCK_M 16, and placed before the fp8 clamp `TILE_SIZE = max(32, TILE_SIZE)` -> fp8 gets TILE 32 | 44.7 ms | (1) the clamp belongs to the pre-shuffled A8W8 KV branch (`block_size >= 32` assert there); applied to plain fp8 at head 512 it pushes LDS back to 32 KB per stage. (2) BLOCK_M 16 = 2 query positions per program at nqpkv 8: each K/V tile is re-read 8x and the QK^T M-dimension starves the MFMA units |
| #5088 (v0.1.21+, JSON tables) | fp8: `D_GEQ_512.DT_fp8_fp8` = BLOCK_M 16 / TILE_SIZE_MIN 32; bf16: `D_GEQ_512` = BLOCK_M 16 / TILE 16 | fp8 44.7 ms; bf16 entry 198 ms | the clamp survived as `TILE_SIZE_MIN 32` on the fp8 keys; the lookup resolves the D axis before Q, so `Q_GEQ_256` (BLOCK_M 128) is unreachable for head >= 256 (#5063 reports the same). The bf16 entry M16/T16 is 0.58x of the pre-#4044 default |

## Why these values, and why the entry is conditional

The head-512 slowdown is an LDS occupancy cliff, not an MFMA-shape problem. LDS per workgroup scales with TILE x head_size x bytes x (K + V) x stages; the default fills the 64 KB per CU exactly at head 512. TILE 16 brings it to 16 KB per stage so several workgroups co-reside and loads overlap compute. BLOCK_M has to stay large enough for K/V reuse and the QK^T M-dimension, while the fp32 accumulator (BLOCK_M x 512) grows with it; M64 / T16 / stages 2 / warps 4 was the measured optimum.

The same tile on other head sizes goes the other way, hence the `D_GEQ_512` condition instead of touching any default:

| shape (fp8, 16k) | default | TILE 16 | ratio |
|---|---|---|---|
| head 512 | 115.6 ms | 26.0 ms | 4.44x |
| head 256 | 12.10 ms | 11.61 ms | 1.04x |
| head 128 | 3.45 ms | 7.88 ms | 0.44x |

The key is deliberately dtype-agnostic: at head 512 the small tile is needed regardless of dtype, and `TILE_SIZE_MIN 32` on the `DT_fp8_fp8` keys is exactly what blocks it for fp8 today. `Q_GEQ_256` restricts it to prefill; the `Q_LEQ_1` decode entries are untouched.

## Relation to #5063

#5063 adds the same key with BLOCK_M 128 for the same reason (the D-axis shadowing). M128/T16 and M64/T16 have not been compared side by side on gfx942; happy to run that on MI325X and converge on one entry. Only gfx942 is changed here (measured); the gfx950 table has the same structure and may deserve the same look.

## Testing

- Kernel bake-off and vLLM end-to-end numbers above; raw logs available on request.
- The key uses the same axes and form as the existing `D_GEQ_512.Q_LEQ_1*` entries, so it is matched by the same lookup path (`explain("attn_2d", ...)` prints the selected entry; with this change a (D=512, Q>=256) prefill should report `D_GEQ_512.Q_GEQ_256`).

## Side notes (not changed here)

- `TILE_SIZE_MIN 32` on `D_GEQ_512.DT_fp8_fp8` comes from the pre-shuffled-KV assert; whether the plain fp8 decode path still needs it is worth a look.
- Resolving the D axis before Q means any `Q_GEQ_*` entry is unreachable for head sizes that have a D key; composite keys work around it, the ordering is the root cause.


zhumengzhiren:
Closing in favor of #5063.

@mustafayildirim's sweep on current main (MI325X, triton 3.8.0, Gemma-4 32q/4kv, 16k tokens) shows M128/T16 with 8 warps at 18.9 ms vs this entry's 25.7 ms for d512 fp8 prefill (current table: 43.6 ms), and two defects in this PR as written: the dtype-agnostic key regresses bf16 (93 vs 70 ms at 16k), and on the pre-shuffled A8W8 KV path the TILE=block_size override pushes M64 over the LDS budget (69632 > 65536).

The findings carry over to #5063: the head-512 slowdown is an LDS occupancy cliff (the default fills the 64 KB per CU at head 512), the fix must stay conditional on head size (the same tile regresses head 128 2.3x), and the fp8 TILE_SIZE_MIN 32 is what blocked TILE 16 for fp8 at head 512. End-to-end on vLLM (gemma-4-31B FP8-block, 16k cold prefill): the ten full-attention layers 107 -> 24.7 ms each with M64/T16, prefill wall -32%, output rel. err 6e-3, lm_eval unchanged.

mustafayildirim:
Ran your entry side-by-side against alternatives on MI325X (isolated kernel bench, same methodology as yours: same kernel body, config swapped via `select_2d_config`, Gemma-4 per-rank shapes 32q/4kv, page 64, median of `do_bench`), on both triton 3.7.1 and the CI pin 3.8.0. Your M64/T16/w4 numbers reproduce exactly (25.7 vs your 26.0 ms at 16k fp8) ✅ — but two findings:

**1. The dtype-agnostic key regresses bf16 on CI triton 3.8.0.** Q outranks DT in the schema (`["D", "Q", "DT"]`), so `D_GEQ_512.Q_GEQ_256` (no DT component) takes precedence over `D_GEQ_512.DT_fp8_fp8` for fp8 too — and the same M64/T16/**w4** config applied to bf16 is a regression: 93.0 ms vs 69.9 ms table baseline at 16k (285 vs 210 at 28k). The fp8-tuned warp count doesn't transfer to bf16; bf16 wants w8 at this tile (38.8 ms).

**2. fp8 can go further with w8 at M128.** At fp8, M128/T16/**w8** compiles with 0 spills (238 VGPR, LDS exactly 65536B — the Q tile is 128×512×1B and fits since only K/V stages double-buffer) and runs 18.9 ms vs your entry's 25.7 — 26% faster still. The spill problem that killed BLOCK_M 128 in #5063 was specifically **w4/w2**, not M128 itself: with 8 warps the register file divides far enough that nothing spills.

**3. Pre-shuffled A8W8 KV:** with `shuffled_kv_cache=True` the wrapper pins `TILE_SIZE = block_size` (64), and M64/T64 needs 69632B LDS > 65536 — compile error. So a dtype-agnostic Q_GEQ_256 key that also matches the pre-shuffled path would hard-crash that configuration at head 512. Worth a guard or a SHUF-scoped entry before merge.

I've posted the dtype-split superset as #5650 (fp8: M128/T16/w8; bf16: M64/T16/w8; plus d256 sliding entries and Q_GEQ_1024 binding — measured crossover vs the D-only entries is between 512 and 1024 tokens). Happy to converge on one PR — yours has the e2e vLLM numbers, mine has the dtype split, the Q bound and the pre-shuffled guard; whichever you and @cagrikymk prefer as the landing vehicle.

mustafayildirim:
Thanks for closing this out and for carrying the analysis forward — one pointer: **#5063 is also closed** (both of its entries spill on gfx942; the corrected successor is **#5650**, same branch of work, dtype-split entries + Q_GEQ_1024 binding + the pre-shuffled-KV guard you flagged).

Your e2e numbers are exactly the missing piece for #5650: I've only run isolated kernel benches there (d512 fp8: table 43.6 → M64/T16/w4 25.7 → **M128/T16/w8 18.9 ms**, 0 spills). Your M64/T16 e2e (107 → 24.7 ms/layer, prefill −32%) validates the family; if your e2e harness is reproducible, running it once with the #5650 fp8 entry (M128/T16/w8) would confirm the isolated 26% delta survives end-to-end — happy to help set that up, or if you'd rather land M64/T16 as the safer measured-e2e config, that's a one-line change to #5650's fp8 entry and I'm fine with it.

PR #5652, opened 2026-09-18
[HIP] [Perf] Fold shared-expert gate GEMV into topk_softmax kernel

Emmanuel0612:
Add fuse-gate mode to topk_softmax so the shared-expert sigmoid gate is computed inside the routed softmax top-k kernel, exposing the new hidden_states/gate_weight/shared_expert_scale/shared_expert_base args through the pybind registration. Adds an op_test with a perf+correctness sweep.

## Submission Checklist


yzhou103:
Hi I'd like to understand the baseline this is being compared against, because downstream the shared-expert gate already appears to be folded into the router GEMM. 
 In Qwen3-Next — a gated shared expert with sigmoid scoring, i.e. exactly this PR's target
  shape — the router projection is a *merged* linear:

  ```python
  self.gate = MergedReplicatedLinear(
      config.hidden_size,
      [config.num_experts, config.n_shared_experts],   # routed | shared, one weight
      bias=False, quant_config=None, ...)
  ...
  logits = self.gate(hidden_states)   # single GEMM -> [M, E + S]
  router_logits = logits              # trailing columns are the shared gate logit
```
 1. Which model/config can't use the merged-gate approach, and why? The gate is
  unquantized bf16 (quant_config=None), so concatenating the [S, H] rows at load time
  seems straightforward, and the infrastructure is already generic.
  2. Could you share measurements against the merged-gate baseline — rather than against
  a standalone GEMV — over a realistic range of M? The test added here only reports the
  fused timing, with no baseline column.

  If there's no case the merged gate can't cover, I'd rather not add a second way to do this:
  it's a new API mode, a new weight-plumbing path, and it moves work off the matrix cores.

  Happy to go through the kernel-level findings (static LDS sizing, vector-load alignment vs.
  the tail loop, missing validation) once the motivation is settled — but they're moot if the
  merged gate already covers thi

**Comment by Emmanuel0612** [comment]
> Hi I'd like to understand the baseline this is being compared against, because downstream the shared-expert gate already appears to be folded into the router GEMM. In Qwen3-Next — a gated shared expert with sigmoid scoring, i.e. exactly this PR's target shape — the router projection is a _merged_ linear:
> 
> ```python
> self.gate = MergedReplicatedLinear(
>     config.hidden_size,
>     [config.num_experts, config.n_shared_experts],   # routed | shared, one weight
>     bias=False, quant_config=None, ...)
> ...
> logits = self.gate(hidden_states)   # single GEMM -> [M, E + S]
> router_logits = logits              # trailing columns are the shared gate logit
> ```
> 
> 1. Which model/config can't use the merged-gate approach, and why? The gate is
>    unquantized bf16 (quant_config=None), so concatenating the [S, H] rows at load time
>    seems straightforward, and the infrastructure is already generic.
> 2. Could you share measurements against the merged-gate baseline — rather than against
>    a standalone GEMV — over a realistic range of M? The test added here only reports the
>    fused timing, with no baseline column.
> 
> If there's no case the merged gate can't cover, I'd rather not add a second way to do this: it's a new API mode, a new weight-plumbing path, and it moves work off the matrix cores.
> 
> Happy to go through the kernel-level findings (static LDS sizing, vector-load alignment vs. the tail loop, missing validation) once the motivation is settled — but they're moot if the merged gate already covers thi

Hi, thank you for the careful read.
Let me explain why this PR exist.

1) Our target is Qwen3.5-397B-A17B-MXFP4. 
    The router gate is MXFP4 (fp4) but shared_expert_gate is in the quant exclude list (bf16, all 60 layers), so they can't share one GEMM without re-quantizing the bf16 gate to fp4 and changing calibrated numerics.

2) Since the merged gate isn't available for this recipe, the fair comparison is the separate bf16 GEMV + append. Please have a check of following tables.
without folding:
image
with folding:
image

3) We also restructured per your first note: this is now a SEPARATE op
(`topk_softmax_fused_shared_gate`) with its own kernel. The existing
`topk_softmax` and its kernel are reverted to base verbatim, so non-fuse-gate
callers carry NO shared-gate LDS allocation (the earlier static `s_gate` penalty
is gone from the shared kernel).

**Comment by yzhou103** [comment]
Thanks! please look at these:
 2. Alignment is still unchecked (UB). is_contiguous() actually makes this concrete:
  with a contiguous hidden_states, hs_row = base + row * H, so any H not a multiple of
  ELTS_PER_LDG makes every row after the first misaligned for the 64B AccessType load
  (same for gw_row at s >= 1). Please either AITER_CHECK(H * sizeof(DTYPE) % BYTES_PER_LDG == 0)
  and drop the tail loop, or make the tail path actually handle misalignment.

  3. Empty shared_expert_scoring_func is still a silent no-op. The check is
  if(!empty()) AITER_CHECK(== "sigmoid"), so "" passes, SCORING_FUNC resolves to NONE,
  and no shared weights or ids are written at all — uninitialized output, no error. For this
  op the scoring function is mandatory; please require it (and num_shared_experts > 0).

  4. The duplication has already diverged. The header says "Bodies are verbatim copies
  with renamed identifiers", and in the copy the fuse branch got float score = 0.0f; while
  the legacy branch still has the uninitialized float score;. That legacy branch is also
  dead code here — the host always passes a non-null gate_weight — so it's an unreachable
  path carrying an unfixed bug. Suggest deleting the legacy branch (plus the now-redundant
  shared_gate_weight != nullptr and shared_expert_base >= 0 runtime checks) from the new
  kernel, and factoring the shared top-k body (argmax + multithread_reduce + clear-to--INFINITY
  - renorm) into a __device__ helper in a header so the two copies can't drift further.

  5. token_expert_indices width is now documented incorrectly. The header says
  [num_tokens, topk + num_shared_experts], but the kernel writes
  source_rows[k * thread_row + k_idx] with a hardcoded stride of topk. With a
  topk + S wide buffer the writes land in the wrong columns. Either thread a stride
  through, or document it as [num_tokens, topk].

  6. Still open from the previous round:
  - indices[...] = NUM_EXPERTS (the out-of-range marker) equals shared_expert_base in the
  intended usage, so a filtered routed slot reads back as shared expert 0.
  - applySharedExpertSigmoidFusedGate still does a serial for(h = 0; h < H; ++h) per
  thread on the non-power-of-2 path. Given this op requires a gate, I'd just
  AITER_CHECK that path unsupported rather than ship a GEMV that's slower than the
  unfused version.
  - Test coverage is still E=512 / H=4096 / S in {1,2} — the risky configs (S=8,
  H > 4096 where the LDS staging silently falls back, H not divisible by
  ELTS_PER_LDG, fp32) are all untested.
  - The file is named test_*.py but contains only bench_* functions, so pytest collects
  it and runs nothing. Either rename to bench_*.py or add a real parametrized test.

**Comment by Emmanuel0612** [comment]
> Thanks! please look at these: 2. Alignment is still unchecked (UB). is_contiguous() actually makes this concrete: with a contiguous hidden_states, hs_row = base + row * H, so any H not a multiple of ELTS_PER_LDG makes every row after the first misaligned for the 64B AccessType load (same for gw_row at s >= 1). Please either AITER_CHECK(H * sizeof(DTYPE) % BYTES_PER_LDG == 0) and drop the tail loop, or make the tail path actually handle misalignment.
> 
> 3. Empty shared_expert_scoring_func is still a silent no-op. The check is
>    if(!empty()) AITER_CHECK(== "sigmoid"), so "" passes, SCORING_FUNC resolves to NONE,
>    and no shared weights or ids are written at all — uninitialized output, no error. For this
>    op the scoring function is mandatory; please require it (and num_shared_experts > 0).
> 4. The duplication has already diverged. The header says "Bodies are verbatim copies
>    with renamed identifiers", and in the copy the fuse branch got float score = 0.0f; while
>    the legacy branch still has the uninitialized float score;. That legacy branch is also
>    dead code here — the host always passes a non-null gate_weight — so it's an unreachable
>    path carrying an unfixed bug. Suggest deleting the legacy branch (plus the now-redundant
>    shared_gate_weight != nullptr and shared_expert_base >= 0 runtime checks) from the new
>    kernel, and factoring the shared top-k body (argmax + multithread_reduce + clear-to--INFINITY
> 
> * renorm) into a **device** helper in a header so the two copies can't drift further.
> 
> 5. token_expert_indices width is now documented incorrectly. The header says
>    [num_tokens, topk + num_shared_experts], but the kernel writes
>    source_rows[k * thread_row + k_idx] with a hardcoded stride of topk. With a
>    topk + S wide buffer the writes land in the wrong columns. Either thread a stride
>    through, or document it as [num_tokens, topk].
> 6. Still open from the previous round:
> 
> * indices[...] = NUM_EXPERTS (the out-of-range marker) equals shared_expert_base in the
>   intended usage, so a filtered routed slot reads back as shared expert 0.
> * applySharedExpertSigmoidFusedGate still does a serial for(h = 0; h < H; ++h) per
>   thread on the non-power-of-2 path. Given this op requires a gate, I'd just
>   AITER_CHECK that path unsupported rather than ship a GEMV that's slower than the
>   unfused version.
> * Test coverage is still E=512 / H=4096 / S in {1,2} — the risky configs (S=8,
>   H > 4096 where the LDS staging silently falls back, H not divisible by
>   ELTS_PER_LDG, fp32) are all untested.
> * The file is named test__.py but contains only bench__ functions, so pytest collects
>   it and runs nothing. Either rename to bench_*.py or add a real parametrized test.

Thanks for the second review. Every point is fixed in the new op and kernel. Here is what changed,

2) The host op now checks that each row's size in bytes (hidden size times the data-type size) is a multiple of 64. If it is not, the op raises an error. Because of this check, the hidden size is always an exact multiple of the vector width, so we also removed the old scalar "tail" loop — with the check in place it could never run.

3) The op now always requires two things — at least one shared expert (num_shared_experts > 0) and scoring_func == "sigmoid". If either is missing, it raises an error instead of returning garbage.

4) We removed the dead "legacy" branch (the one with the uninitialized score) and the two runtime guards that could never be false (the null-gate check and the shared_expert_base >= 0 check). The new kernel now has only the fused path. About your suggestion to move the shared top-k code into one helper that both kernels include: we chose to keep the two kernels separate instead. The reason is the design goal from round 1 — the original topk_softmax kernel stays exactly the same as upstream, so every existing caller that does not use the shared gate keeps working with zero risk.

5) This buffer is scratch space: the op writes it, but nothing reads it back. The kernel fills it with a fixed stride of topk (routed slots only). So we now size it [num_tokens, topk] instead of [num_tokens, topk + num_shared], and fixed the header comment to match.

6) 
* The routed code marks a "skip this slot" entry with a special number. That number used to be num_experts, which is exactly the id of the first shared expert — so a skipped slot could be read back as shared expert 0. We changed the marker to num_experts + num_shared_experts, which is outside the range of all real ids, so it can no longer be confused with any expert.
* That path used a slow, one-element-at-a-time GEMV that was actually slower than not fusing at all. Instead of shipping it, the op now rejects non-power-of-2 expert counts with a clear error, and we deleted that slow code.
*  The test file used to contain only benchmark functions, so pytest collected the file but ran no real test. We added a real test (test_correctness) that compares the fused op and the unfused baseline against a reference for several shapes: 1, 2, and 8 shared experts; float32 inputs; and a larger hidden size that triggers the "read the gate from global memory" fallback. We also added two tests that confirm the new checks actually fire — one for a misaligned hidden size, one for a non-power-of-2 expert count.

File: aiter/ops/moe_op.py [hunk 1]
```diff
@@ -66,6 +66,68 @@ def topk_softmax(
...
+    # written by the kernel. Output buffers are width topk + num_shared_experts.
+    num_routing_experts = gating_output.shape[-1]
+    num_tokens = gating_output.numel() // num_routing_experts
+    is_pow_2 = (
+        num_routing_experts != 0
+        and (num_routing_experts & (num_routing_experts - 1)) == 0
+    )
+    needs_workspace = (not is_pow_2) or num_routing_experts > 256
+    workspace_size = num_tokens * num_routing_experts if needs_workspace else 0
+    softmax_workspace = torch.empty(
```
yzhou103, line 112: Its only consumer was the non-power-of-2 two-pass path (`moeSoftmax` writes row
  probabilities, `moeTopK` reads them back), which this round replaced with
  `AITER_CHECK(false, ...)`. Grepping the fused path, `softmax_workspace` appears only in
  the signatures — it is never read.

  The Python wrapper still carries the old sizing logic, though:

  ```python
  needs_workspace = (not is_pow_2) or num_routing_experts > 256
  workspace_size = num_tokens * num_routing_experts if needs_workspace else 0
  softmax_workspace = torch.empty(workspace_size, ...)
```

  With the target E = 512 this hits the > 256 branch, so every call allocates
  num_tokens * 512 * 4 bytes and immediately discards them — 16MB per call at M=8192,
  per layer. Please drop the parameter from the new op entirely (C++ entry, moe_op.h,
  the py::arg list, and the wrapper); the non-power-of-2 path it existed for is gone

yzhou103:
1. `bench_unfused` is not a representative baseline, so the reported speedup is inflated.**

  The timed region is ~9 kernels, and most of them are artifacts of the torch plumbing
  rather than of "not fusing":

  ```python
  topk_softmax(r_w, r_i, r_tei, gating, renorm)                            # 1  necessary
  shared_w = torch.sigmoid(hs.float() @ gate_weight.float().t()) * scale   # 2 upcasts + mm + sigmoid + mul
  w[:, :topk] = r_w;   w[:, topk:] = shared_w                              # 2 copies
  ids[:, :topk] = r_i; ids[:, topk:] = shared_ids                          # 2 copies
 ```
  - The two .float() casts materialize an [M, H] fp32 temporary (128MB at M=8192,
  H=4096). A bf16 F.linear already accumulates in fp32 on MFMA, which is what the fused
  kernel does — so the upcast also makes the baseline more accurate than the op under
  test, which is the wrong way round.
  - The two routed copies are avoidable. The kernel indexes its outputs through
  topk_weights.stride(0) / topk_indices.stride(0), so w[:, :topk] and ids[:, :topk]
  can be passed directly as views and the routed columns get written in place. That
  strided-output path is already exercised in-repo by op_tests/test_moeTopkSoftmax.py,
  which calls the base op with torch.empty_strided buffers.
  - shared_ids depends only on (num_routing_experts, num_shared_experts), not on the
  input, so it belongs outside the timed region.

  Suggested rewrite (~4 kernels, and the sigmoid/scale/store could be folded further):

  w   = torch.empty(tokens, total, dtype=dtypes.fp32)
  ids = torch.empty(tokens, total, dtype=dtypes.i32)
  tei = torch.empty(tokens, topk,  dtype=dtypes.i32)
  w_routed, ids_routed = w[:, :topk], ids[:, :topk]      # views, written in place
  ids[:, topk:] = base + torch.arange(num_shared, dtype=dtypes.i32)   # input-independent
  logit = torch.empty(tokens, num_shared, dtype=dtype)   # reused, not reallocated

  def fn():
      topk_softmax(w_routed, ids_routed, tei, gating, renorm)
      torch.nn.functional.linear(hs, gate_weight, out=logit)   # bf16, no upcast
      torch.sigmoid(logit, out=logit)
      w[:, topk:] = logit * scale

  2. The M sweep is decode-only. --tokens defaults to [1, 4, 17, 64]. The in-kernel
  GEMV runs on VALU with THREADS_PER_ROW = 16 lanes per row and adds a full [M, H] HBM
  pass, so the interesting question is at the other end. Please extend to
  [1, 4, 16, 64, 256, 1024, 4096, 16384] so we can see whether there's a crossover, and
  where it is.

  3. Missing axes in test_correctness. It currently pins tokens=64, num_experts=512, topk=10, scale=1.0, renorm=True, so several branches are untested:

  - need_renorm=False — a separate epilogue, currently never executed under pytest.
  - shared_expert_scale != 1.0 — applied in the kernel; would pass unnoticed if dropped.
  - num_experts=256 — at E < 512 the launcher picks BYTES_PER_LDG = 32, i.e. a
  different ELTS_PER_LDG and a different vector width in the GEMV. Only the 64B path is
  covered today.
  - Token-count edge cases: M = 1, and an M that isn't a multiple of ROWS_PER_CTA
  (8 at E=512) such as M = 65, to cover the tail block. M = 0 currently yields
  num_blocks == 0 and a failed launch — worth deciding whether that should be a guarded
  no-op.

PR #5663, opened 2026-09-18
[Triton/Gluon] Fix fp8 KV correctness reference in bench_unified_attention

### LINKED ISSUE #5661: bench_unified_attention.py: fp8 KV -test reference is mis-scaled (false correctness failures)
## Summary
`op_tests/op_benchmarks/triton/bench_unified_attention.py`'s `-test` correctness check reports **false failures for every fp8 KV shape**, while bf16 passes. Root cause: the reference is fed the **original (un-quantized)** K/V tensors together with the fp8 **descale** factors.

## Details
In the `-test` block, the kernel is called with the fp8-quantized tensors (`k_fp8`, `v_fp8`) plus `k_descale`/`v_descale`, so it reconstructs `k_fp8 * k_descale ≈ K`. But `ref_paged_attn(...)` is called with the original `inputs["key_cache"]`/`inputs["value_cache"]` **and** the same descales. `ref_paged_attn` (in `op_tests/triton_tests/attention/test_unified_attention.py`) unconditionally does `key_cache = key_cache * k_descale` / `value_cache = value_cache * v_descale`, so the reference applies an extra descale to full-precision tensors, producing a wrong reference. Softmax is not invariant to a scale on K, so the outputs diverge structurally.

## Repro (MI325 / gfx942)
```
python op_tests/op_benchmarks/triton/bench_unified_attention.py -b 1 -hq 32 -hk 8 -d 128 -sq 512 -sk 512 -equal_seqlens -fp8_kv -test
```
→ `FAIL ... max_diff=3.87` (19.9% mismatched). The same shape in bf16 (drop `-fp8_kv`) passes with `max_diff=0.0156`.

## Fix
Pass the quantized tensors to the reference when fp8 is enabled (so `ref * descale` reconstructs the original), e.g. use `inputs["k_fp8"]`/`inputs["v_fp8"]`/`inputs["q_fp8"]` in the `ref_paged_attn` call. With that change all shapes (incl. head_size 512 GQA 32/4 and head_size 256 GQA 32/16) pass at `max_diff=0.0156`.

## Impact
The kernel is correct; only the test harness is wrong. But the broken check hides real fp8 regressions and blocks using `-test` as a correctness gate when tuning fp8 configs.

mpashkovskii:
## Purpose
`op_tests/op_benchmarks/triton/bench_unified_attention.py -test` reports false correctness failures for **every** fp8 KV shape (bf16 passes). The `-test` path calls `ref_paged_attn` with the original, un-quantized `key_cache`/`value_cache` **and** the fp8 descale factors. `ref_paged_attn` then does `key_cache = key_cache * k_descale` on full-precision tensors, so the reference diverges structurally from the kernel (softmax is not scale-invariant in K). The kernel itself is correct; only the harness reference is wrong. Fixes #5661.

## Change
Pass the quantized tensors (`q_fp8`/`k_fp8`/`v_fp8`) to `ref_paged_attn` when fp8 is enabled, so the reference reconstructs K/V via `* descale`, matching what the kernel consumes. No kernel code touched.

## Test Plan / Result (MI325, gfx942)
```
python op_tests/op_benchmarks/triton/bench_unified_attention.py -b 1 -hq 32 -hk 8 -d 128 -sq 512 -sk 512 -equal_seqlens -fp8_kv -test
```
- Before: `FAIL ... max_diff=3.87` (19.9% mismatched).
- After: `PASS ... max_diff=0.0156`.
- Also verified PASS after the fix for head_size 512 (GQA 32/4) and head_size 256 (GQA 32/16), the Gemma-4 shapes.

## Notes
AI assistance (Claude Code) was used to find and fix this. Not a duplicate — no open PR addresses the bench fp8 reference. This is a test-harness-only change; it unblocks using `-test` as a correctness gate when tuning fp8 configs.


azaidy:
@mpashkovskii can you check code style. please match the black/ruff versions that the CI is using

cagrikymk:
@mpashkovskii from the changes perspective, LGTM!
As stated, please address and styling issues

PR #5676, opened 2026-09-18
[Triton/Gluon] Create benchmark for fused_rmsnorm_add_bench

nidal567:
## Motivation

`fused_rmsnorm_add.py` ships a Triton kernel and a gfx1250 gluon kernel, but it doesn't have a benchmark within Triton, so there is no baseline to measure numbers. Because it's been added post Triton 3.4, it does not have any performance history at all. 

The op is not covered by the model benchmarking tool either -- it is absent from both `model_shapes.json` and `KERNEL_DICT`, so it cannot be reached through `bench_models.py`.

## Technical Details

Shape-driven benchmark following a similar family layout:

- same get_x_vals/run_benchmark/parse_args/main structure,
- same triton.testing.Benchmark + perf_report + do_bench(warmup=25, rep=100), and,
- it reuses the tester's generate_inputs rather than duplicating input construction.

It has a smaller flag set compared to previous additions of benchmarks ([PR#5574](https://github.com/ROCm/aiter/pull/5574)) since there is no `--backend` (Gluon on gfx1250 and Triton otherwise). No `--pad-to-multiple`, `--quant`, or `--backward` as op has none of those paths. It does keep `add_residual`.

15 default shapes: 

| group | shapes | why |
|---|---|---|
| M scaling at a fixed width | (1 / 1024 / 16384, 4096) | `grid=(M,)`, so M is the occupancy knob |
| `BLOCK_SIZE_N` buckets | (8192, 4 / 128 / 512 / 2048 / 8192 / 16384), (256, 32768) | the 32 floor up to the widest row |
| model hidden sizes | (8192, 2880 / 7168), (4096, 5120), (16380, 1536) | `model_shapes.json` rmsnorm entries |
| divisibility | (8192, 4095) | `N % 16 != 0`, (below) |

N = 4, 128, 512, 1536, 2880, 4096, 5120, 7168, 8192 and 16384 are all rmsnorm widths in this repo's `model_shapes.json` or the tester's own `get_vals()`; (16380, 1536) is a tester shape.

Bandwidth is the meaningful figure here, and that's the default metric for this kernel. 
Memory accounting is similar to ([PR#5574](https://github.com/ROCm/aiter/pull/5574)) with the exclusion of the output due to the padding in PR#5574. 

## Test Plan

- Every CLI path exercised: default sweep, `--shape`, all three metrics, all three dtypes, `--add-residual`, `-o`, and both `--shape` guards
- Byte model cross-checked against `rocprofv3 --kernel-trace` on four configurations
- pytest correctness test
- black, ruff checks

## Test Result

Default sweep, bandwidth, gfx950, bf16:

```
       M       N    GB/s  (no res)    GB/s (--add-residual)
       1    4096         3.9                    6.1
    1024    4096      2205.5                 3179.2
   16384    4096      5369.3                 5621.3
    8192       4        13.7                   24.5
    8192     128       460.7                  844.6
    8192     512      1645.7                 2682.8
    8192    2048      4164.8                 4599.2
    8192    8192      5165.8                 5572.6
    8192   16384      5412.8                 5638.2
     256   32768      3216.7                 4258.1
    8192    2880      4805.6                 5132.6
    8192    7168      5279.1                 5470.7
    4096    5120      4425.7                 4866.8
   16380    1536      4758.7                 5032.2
    8192    4095      1717.2                 1943.3
```

`test_fused_rmsnorm_add.py`: 32 passed. `black` and `ruff`: clean

### Byte model cross-check

`rocprofv3 --kernel-trace`, 250 iterations + 25 warmup, against the model above:

| shape | bytes | kernel time | kernel GB/s | % of ~8 TB/s |
|---|---|---|---|---|
| 8192 x 8192 | 268.45 MB | 42.825 us | 6268.5 | 78.4 % |
| 8192 x 8192, residual | 536.89 MB | 80.032 us | 6708.4 | 83.9 % |
| 8192 x 4096 | 134.23 MB | 22.069 us | 6082.0 | 76.0 % |

The residual and non-residual paths land within 7 % of each other, which is what validates the `HAS_RES` branch of the accounting -- had the `out_res1` write been dropped from the model, the residual row would read ~25 % low. `do_bench` reads above kernel time (27.4 us vs 22.1 us at 8192 x 4096) because it flushes L2 before every timed iteration. An upcoming  `do_bench_cudagraph` can fix and account for this difference.

### One finding the benchmark makes visible

`(8192, 4095)` is in the default list because `N % 16 != 0` costs **3.4x**:

| shape | `BLOCK_SIZE_N` | kernel time | kernel GB/s |
|---|---|---|---|
| 8192 x 4096 | 4096 | 22.069 us | 6082.0 |
| 8192 x 4095 | 4096 | 74.462 us | 1802.2 |

Same kernel, same `BLOCK_SIZE_N`, same byte count to within 0.02 % -- the only difference is that Triton's divisibility-by-16 specialization on `N` and the row strides does not apply. This is *not* row-stride byte alignment: `N = 4088` gives a 16-byte-aligned row stride (8176 B) and is still in the slow tier, while `N = 4080` is fast. Across 13 sampled widths the fast/slow split tracks `N % 16 == 0` exactly.

Not addressed by this PR. The benchmark exists so it can be tracked for future changes or comparisons.

### Register growth

The uncapped `BLOCK_SIZE_N` shows up as register pressure rather than spilling (`FIRST_INPUT_RES=0`, `num_warps=4`):

| `BLOCK_SIZE_N` | 32 | 4096 | 8192 | 16384 | 32768 |
|---|---|---|---|---|---|
| VGPR | 12 | 43 | 63 | 154 | 219 |
| spills | 0 | 0 | 0 | 0 | 0 |

There are no spills at any bucket, and the widest rows are not slow -- `(8192, 16384)` is the fastest shape in the sweep. The cost is occupancy, which this streaming kernel tolerates.

## Submission Checklist


brunomazzottiamd (approved):
LGTM!

PR #5688, opened 2026-09-19
[Config] configs(glm5.2): o_proj switches to CK Tile at M=1024, not M=8192

ThomasNing:
`configs/model_configs/a8w8_blockscale_bpreshuffle_tuned_gemm_glm5.2.csv` keeps CK instances for `o_proj` (N=6144, K=4096) all the way to M=4096 and only moves to `a8w8_blockscale_cktile_192x256x128` at M=8192.

Measured on MI350X (gfx950, 256 CU), **CK Tile is already ahead from M=1024** — three rows earlier.

## Result

| M | shipped | candidate | gain | resolution floor |
|---:|---:|---:|---:|---:|
| 1024 | 54.739 µs `ck` | **49.978 µs** `cktile` | **+8.70%** | 5.98% |
| 2048 | 100.818 µs `ck` | **87.913 µs** `cktile` | **+12.80%** | 3.15% |
| 4096 | 184.560 µs `ck` | **148.773 µs** `cktile` | **+19.39%** | 3.10% |

`o_proj` runs in every one of this model's 78 layers, so this is a prefill and large-batch win on a shape that is always hot.

## The boundary is measured, not assumed

This is **not** "CK Tile is better than CK". Below M=1024 this shape dispatches to the asm kernel, and CK Tile loses there badly:

| M | shipped | CK Tile forced | |
|---:|---:|---:|---|
| 256 | 20.092 µs `asm` | 46.028 µs | **−129%** |
| 512 | 33.206 µs `asm` | 46.434 µs | **−39.8%** |

So the claim is narrower and more useful: **the switch is three rungs late.** Rows below M=1024 are untouched.

## Protocol

- Production entry point, not a microbenchmark.
- Arms **interleaved**, 9 repeats × 100 iterations, compared by **median**. Single-shot margins on CK candidates here are misordered, not merely noisy.
- Correctness checked against a dequantized fp32 reference on the first repeat of every arm.
- A margin is only called a win if it clears **both arms' spreads and the control delta** — the floor column above.

### Negative control

Forcing the kernel the heuristic already picks must not change the timing. If it does, the instrument is measuring something other than the kernel.

```
# stage 1 control: M=8192 unforced 70.348 (spread 3.7%) vs forced-same 70.356 (spread 1.6%) -> +0.01%
# stage 1 control: M=8192 unforced 68.737 (spread 9.8%) vs forced-same 70.191 (spread 4.0%) -> +2.11%
```

+0.01% on the low-M run and +2.11% on the high-M run. Both runs used the larger of that and the arm spreads as the bar. Without this arm the M=1024 row would not have been distinguishable from noise on the first attempt; with it, +8.70% against a 5.98% floor is.

## Scope

Three rows. No kernel, no dispatch logic, no ABI change. The table stays uniquely keyed on `(gfx, cu_num, M, N, K)` and its schema is unchanged.

Environment: ROCm 7.2.4, torch 2.10.0+rocm7.2.4, MI350X 8×gfx950/256 CU, quiet box.


ThomasNing:
**Correcting the M=1024 number in the table above, downward.**

I built a reproduce-or-fail gate for this change and ran it repeatedly. M=2048 and M=4096 hold. **M=1024 does not hold at +8.70%** — its gain estimate fell every time I added repeats, which is what a margin flattered by noise looks like:

| run | repeats × iters | M=1024 gain | floor | verdict |
|---|---|---:|---:|---|
| original (in the table above) | 9 × 100 | +8.70% | 5.98% | KEEP |
| gate run 1 | 9 × 100 | +7.99% | 9.47% | TIE |
| **deep** | **25 × 300** | **+6.55%** | 4.98% | KEEP |
| gate run 2 | 9 × 100 | +6.50% | 7.58% | TIE |

**~6.5% is the honest figure for M=1024**, not 8.7%. It is a real win — it clears its floor when measured with enough repeats, and it never once measured negative — but it sits close to that shape's own run-to-run spread, so at modest repeat counts it lands TIE about as often as KEEP.

The other two rows were stable across all four runs:

| M | gains observed | honest figure |
|---:|---|---:|
| 1024 | +8.70, +7.99, +6.55, +6.50 | **~+6.5%** |
| 2048 | +12.80, +10.85, +12.97 | **~+12.9%** |
| 4096 | +19.39, +19.04, +18.35 | **~+18.6%** |

Nothing about the change itself is affected — the same three rows, the same kernel, the same direction, and the boundary below M=1024 is unchanged (CK Tile still loses to asm by 129% at M=256 and 40% at M=512). Only the headline number for the smallest of the three rows was optimistic, and I would rather correct it here than have someone find it later.

If you would prefer the M=1024 row dropped and only 2048/4096 changed, say so and I will cut it — the two large rows carry almost all of the benefit and are not close to their floors.

ThomasNing:
**Dropping the M=1024 row. Cross-box measurement says it is a wash, not a win.**

I ran the same gate on **MI355X** — same `gfx950`, same 256 CU, so the *same tuned row serves both machines*. It disagrees with MI350X at M=1024:

| M | MI350X | MI355X | |
|---:|---:|---:|---|
| 1024 | ~+6.5% | **−1.81%** | **opposite signs** |
| 2048 | ~+12.9% | +8.21% | wins on both |
| 4096 | ~+18.6% | +13.88% | wins on both |

MI355X, pinned image via enroot, stage-1 control at **+0.88%**:

```
# stage 1 control: M=8192 unforced 57.330 (3.7%) vs forced-same 57.835 (4.7%) -> +0.88%
M=1024   shipped   45.738 (4.0%)  row   46.568 (1.5%)   -1.81%  floor 3.98%  TIE
M=2048   shipped   85.447 (2.2%)  row   78.433 (1.7%)   +8.21%  floor 2.16%  KEEP
M=4096   shipped  150.778 (2.8%)  row  129.849 (2.9%)  +13.88%  floor 2.93%  KEEP
```

A row keyed on `(gfx, cu_num)` cannot be right on one box and wrong on the other, and −1.81% on the second machine is not something to ship on the strength of +6.5% on the first. Both numbers sit near their floors; the honest reading is that **M=1024 is a wash**, and its apparent win on MI350X was that machine's noise — which is also why its estimate kept falling as I added repeats (+8.70 → +7.99 → +6.55 → +6.50).

**The branch now changes two rows, not three.** M=2048 and M=4096 win on both boxes, by margins that clear every floor measured (2.16–4.53%), across four runs on MI350X and one on MI355X. They carry nearly all of the benefit.

Everything else is unchanged: same protocol, and the boundary below is still measured — at M=512 and M=256 this shape dispatches to asm, where CK Tile loses by 40% and 129%.

Apologies for the churn on this one. The original three-row table went out before I had the second box, and I would rather land two rows I can defend on both machines than three where one flips sign.

aiter/configs/model_configs/a8w8_blockscale_bpreshuffle_tuned_gemm_glm5.2.csv
```diff
@@ -1,7 +1,7 @@
 gfx,cu_num,M,N,K,libtype,kernelId,splitK,us,kernelName,tflops,bw,errRatio
 gfx950,256,1024,6144,4096,ck,14,0,44.0612,a8w8_blockscale_bpreshuffle_1x128x128_256x64x128x128_16x16_16x16_8x32x1_8x32x1_1x32x1x8_8_2x1_intrawave_v1,1169.73,951.93,0.0002
-gfx950,256,2048,6144,4096,ck,13,0,74.66,a8w8_blockscale_bpreshuffle_1x128x128_256x64x256x128_16x16_16x16_8x32x1_8x32x1_1x32x1x8_8_2x1_intrawave_v1,1380.65,786.5,0.0
-gfx950,256,4096,6144,4096,ck,13,0,131.1005,a8w8_blockscale_bpreshuffle_1x128x128_256x64x256x128_16x16_16x16_8x32x1_8x32x1_1x32x1x8_8_2x1_intrawave_v1,1572.52,703.85,0.0
+gfx950,256,2048,6144,4096,cktile,29,0,87.9130,a8w8_blockscale_cktile_192x256x128_4x2x1_16x16x128_intrawave_0x1x0_4,1172.5,667.9,0.0
```
yzhou103, line 3: The new tuned column is slower than old one ?

PR #5698, opened 2026-09-20
[HIP] Lift the 2 GiB limit on mla_reduce_v1 final output and LSE

kaiyang-1:
`mla_reduce_v1` rejected any call whose final output or final LSE spanned 2 GiB or more. The limit was real: both stores went through a buffer descriptor based at the tensor pointer with a 32-bit signed byte offset carrying the full `(seq, head)` index, so a larger tensor would have silently wrapped. It is easy to reach — bf16 output at `nhead=128` and `batch=16384` is exactly 2 GiB, so `-n 128,1 -b 16384` failed on every context length, on both the opus and asm decode paths.

`seq_idx` and `head_idx` are wave-uniform in all four store sites, so they now go into the 64-bit descriptor base and only the per-lane element offset stays in the 32-bit voffset. This is the same treatment the partial-output buffer already gets a few lines above, and it brings the buffer-store paths in line with `mla_reduce_v1_impl_*` massive's plain-pointer variant, which was already 64-bit. With nothing large left in the voffset, both host-side span checks and the `kMaxSpanBytes` constant are gone.

## Validation

The four previously rejected configurations now run and match the torch reference on both paths:

```
PYTHONPATH=. python3 op_tests/test_mla_persistent.py -d bf16 -kvd bf16 -n 128,1 -b 16384 -c 224 2048
```

## No regression

Folding the offset into the base means the descriptor is rebuilt per sequence instead of hoisted out of the loop, so the cost was measured on the shape where the reduce carries the most weight relative to the decode (`-n 96,4 -c 256`, over a thousand partials against a ~30 us kernel), five runs each, comparing medians:

| | before | after | delta |
|---|---:|---:|---:|
| `-b 32` | 31.186 us | 31.190 us | +0.01% |
| `-b 64` | 36.027 us | 35.958 us | -0.19% |

Both directions sit inside the run-to-run spread and the two distributions overlap almost entirely. A single-shot comparison had suggested up to +1.2%, which the repeats did not reproduce.


ruanjm (approved):
LGTM

shengnxu (approved):

PR #5705, opened 2026-09-20
[Perf] Tune DeepSeek V4.1 mHC large-M split-K on gfx950

jhinpan:
## Purpose

`get_mhc_pre_splitk_large_m` currently selects `(split_k=8, tile_k=64)` from
M=8192 for every eligible gfx950 shape. For the unpacked DeepSeek-V4.1
four-stream H=5120 projection, that policy regresses the lower and upper
large-M buckets; it wins only after the generic policy falls to two-way
split-K and before eight-way split overhead takes over.

This change keeps `(8, 64)` only for M=13312–14592 for that exact geometry.
Packed BF16 weights, other hidden sizes, other architectures, and every shape
outside the measured interval retain the existing policy.

## MI355X results

Controlled fresh-process A-B-B-A measurements used the public delayed-pre
wrapper with identical inputs, warmup, output allocation, stream, and JIT
cache:

- M=13312: 1.0594x device-mean and 1.0575x wall-mean speedup.
- M=14336: 1.0355x device-mean and 1.0344x wall-mean speedup.

Boundary sweeps selected the interval: M<=12544 and M>=14848 are faster on the
generic policy. The test now asserts both interval boundaries, the packed
weight path, and the H=7168 path.

The gfx950 mHC pre/post/post-pre boundary suite completed at M=8192, 13312,
14592, 14848, and 16384 with zero reported `hip_err`.


jhinpan:
Closing after tracing the production call path. This policy is currently reached only through `force_fused=True`; the repository caller is the mHC op test, while vLLM uses the default `force_fused=False` path. A paired vLLM integration showed component gains but no stable end-to-end improvement, so this is not a production-relevant optimization in its current form. The measurements and boundary results are preserved locally as no-go evidence.

PR #5632, opened 2026-09-17
[HIP] Add register-resident grouped topk for single-group MoE.

lirui927:
## Motivation

When `biased_grouped_topk` / `grouped_topk` run with no group filter (`topk_group == num_expert_group`, typically G=1), expert counts are in the hundreds to thousands and topk is in the teens to 32. Decode and small prefill launch one row per token. The existing HIP kernel stores scores in LDS and then runs sequential argmax for `topk` rounds, rescanning the full row each time, so latency grows with topk.

This PR keeps the same APIs but takes a register path: scores stay in VGPRs, and a lane-max pivot plus warp compact picks the topk, skipping the repeated LDS argmax.

## Technical Details

Only `csrc/kernels/topk_softmax_kernels_group.cu` changes:

- New kernel `topk_reg_kernel`: 1 token = 1 wave64; each lane owns `E/64` experts. Comparison scores live in VGPRs; output weights (sigmoid or softmax) are written to LDS and gathered.
- Lane-max pivot + warp compact selects topk; falls back to sequential argmax if there are too many candidates. Numerics match the original LDS kernel.
- `biased_grouped_topk` (sigmoid+bias) and `grouped_topk` (softmax or unbiased sigmoid) share `LAUNCHER_TOPK_REG`; a missed gate falls back to LDS. Biased remains sigmoid-only, matching the existing API.
- Even row stride uses vec2 loads. Odd stride misaligns the row pointer for vec2, so the same `(2L, 2L+1)` pair is loaded as scalars.

Host gates (all must hold):

- `topk_grp == num_expert_group` (no group filter)
- wave64 only (wave32 untested)
- `num_experts % 64 == 0`,  max: 2048， now ues whitelist {128, 192, 256, 384, 512, 896, 1024, 2048}
- topk in [4, 32]

dtypes: fp32 / fp16 / bf16. `gating_output` / `correction_bias` share a dtype; `topk_weights` stay fp32.

## Test Plan

No new test file. Reuse `op_tests/test_moeTopkSoftmax.py`: G=1 hits the register path; a group filter still hits LDS.

## Test Result

`test_moeTopkSoftmax.py` biased grouped topk (including the 896-expert strided case) **passed**, `err_aiter=0`.

### gfx950

gfx950, bf16, `topk_reg_kernel` vs `grouped_topk_kernel` (µs):

**biased**

| token | E=128 k=4 | E=128 k=32 | E=896 k=16 | E=2048 k=4 | E=2048 k=32 |
[...]

PR #5726, opened 2026-09-21
[HIP] Fix grouped topk tie handling for GLM-5.2 accuracy

lirui927:
## Motivation

PR [#5632](https://github.com/ROCm/aiter/pull/5632) added a register-resident
kernel for the single-group Top-K path. However, its output did not exactly
match the original `grouped_topk_kernel`: ties at the Top-K cutoff could change
the selected expert set, and the selected experts and weights could be emitted
in a different order. This changed downstream floating-point accumulation and
caused an accuracy regression on GLM-5.2.

This PR makes the register-resident kernel match `grouped_topk_kernel` exactly,
including selected expert IDs, output order, and weights, while retaining the
register-resident fast path.

## Technical details

- Retain the register-resident fast path introduced in #5632.
- Emit experts in the same score order as the legacy kernel.
- When multiple experts have equal scores at the Top-K selection boundary, use
  the deterministic priority and comparison rules of the legacy kernel,
  including the register-resident fallback path.
- Accumulate the renormalization denominator in the same serial order as the
  legacy implementation, producing bitwise-equal weights.
- Handle NaN/`-Inf` inputs without producing invalid or duplicate expert IDs.

There are no API, dispatch-gate, dtype, or supported-shape changes.

## Test plan

No new test file is added. Compatibility cases are added to the existing
`op_tests/test_moeTopkSoftmax.py`.

They cover the GLM-5.2 router shape (`biased`, BF16, `E=256`, `topk=8`, `G=1`)
with random inputs, all-equal scores, and a 16-way cutoff tie. The register
output is compared directly with the legacy LDS output, including expert order
and bitwise-equal weights.

### gfx950

gfx950, bf16, `topk_reg_kernel` vs `grouped_topk_kernel (LDS)` (µs):

| token | E=128 k=4 | E=128 k=32 | E=256 k=8 (GLM) | E=896 k=16 | E=2048 k=4 | E=2048 k=32 |
|------:|---|---|---|---|---|---|
| 1 | 2.64 vs 2.62 (1.0x) | 6.57 vs 11.86 (1.8x) | 3.40 vs 3.72 (1.1x) | 5.26 vs 10.67 (2.0x) | 5.06 vs 5.96 (1.2x) | 12.46 vs 27.67 (2.2x) |
| 32 | 3.78 vs 4.31 (1.1x) | 7.78 vs 12.81 (1.6x) | 4.56 vs 5.35 (1.2x) | 6.54 vs 12.98 (2.0x) | 6.14 vs 8.84 (1.4x) | 14.05 vs 30.07 (2.1x) |
| 1024 | 4.00 vs 4.52 (1.1x) | 8.14 vs 12.83 (1.6x) | 4.90 vs 5.75 (1.2x) | 7.08 vs 12.85 (1.8x) | 7.27 vs 11.37 (1.6x) | 13.74 vs 33.62 (2.4x) |
| 4096 | 5.29 vs 6.14 (1.2x) | 10.76 vs 18.27 (1.7x) | 6.99 vs 8.23 (1.2x) | 10.37 vs 19.73 (1.9x) | 10.84 vs 22.02 (2.0x) | 18.20 vs 82.20 (4.5x) |

gfx950, bf16, `PR` vs `main (this kernel before)` (µs):

| token | E=128 k=4 | E=128 k=32 | E=256 k=8 (GLM) | E=896 k=16 | E=2048 k=4 | E=2048 k=32 |
|------:|---|---|---|---|---|---|
| 1 | 2.33 → 2.64 (+13%) | 2.34 → 6.57 (+181%) | 2.50 → 3.40 (+36%) | 3.26 → 5.26 (+61%) | 4.78 → 5.06 (+6%) | 7.25 → 12.46 (+72%) |
| 32 | 4.06 → 3.78 (-7%) | 4.07 → 7.78 (+91%) | 4.24 → 4.56 (+7%) | 5.09 → 6.54 (+28%) | 6.48 → 6.14 (-5%) | 8.93 → 14.05 (+57%) |
| 1024 | 4.29 → 4.00 (-7%) | 4.29 → 8.14 (+90%) | 4.57 → 4.90 (+7%) | 5.59 → 7.08 (+27%) | 7.56 → 7.27 (-4%) | 10.14 → 13.74 (+36%) |
| 4096 | 5.39 → 5.29 (-2%) | 5.40 → 10.76 (+99%) | 5.73 → 6.99 (+22%) | 7.76 → 10.37 (+34%) | 10.26 → 10.84 (+6%) | 14.47 → 18.20 (+26%) |

Compared with the previous unordered implementation, the current kernel preserves deterministic legacy ordering for accuracy at the cost of lower performance.

## GLM-5.2 end-to-end accuracy
image

## Submission Checklist


csrc/kernels/topk_softmax_kernels_group.cu
```diff
@@ -886,16 +947,25 @@ __global__ void topk_reg_kernel(DTYPE_I* __restrict__ gating_output,
...
-        warpReduceMax(max_val, max_idx);
+        using kvp = aiter::KeyValuePair<int, float>;
+        const kvp local{max_idx, max_val};
+        auto arg_max = [](kvp a, kvp b) {
+            return topk_candidate_better(a.value, a.key, b.value, b.key) ? a : b;
+        };
+        const kvp winner = wave_reduce<kvp, decltype(arg_max), WARP_SIZE, true>(local, arg_max);
+        max_val          = winner.value;
+        max_idx          = winner.key;
+        if(max_idx < 0)
```
zufayu: When no finite candidates remain, `max_idx = k` may pick an expert that was already selected. For example, if expert 2 is the only finite entry, it can be emitted again when `k == 2`. Could we choose an unused fallback and add a NaN/`-Inf` test?

op_tests/test_moeTopkSoftmax.py
```diff
@@ -782,6 +837,21 @@ def test_topk_softmax_shared_experts(
     "moeTopkSoftmax_reg_biased_grouped_topk summary (markdown):\n%s", df_md
 )
 
+# Re-run the register-resident path for each existing case to verify stable
+# expert selection and weights across repeated executions.
+stable_cases = [
```
zufayu, line 1024: The new order-sensitive test only covers biased sigmoid with renormalization enabled. Since this code is also used by softmax and unbiased sigmoid, could we add the same legacy-vs-register check for those paths?

zufayu (commented):
A few correctness and validation questions.

csrc/kernels/topk_softmax_kernels_group.cu
```diff
@@ -306,6 +306,98 @@ __inline__ __device__ void warpReduceMax(float& val_o, int& idx)
...
+
+__device__ __forceinline__ bool
+topk_candidate_better(float a_score, int a_id, float b_score, int b_id)
+{
+    if(a_score != b_score)
+        return a_score > b_score;
+    return legacy_topk_tie_rank(a_id) < legacy_topk_tie_rank(b_id);
+}
+
+__device__ __forceinline__ void emit_legacy_ordered_topk(const float* cand_val,
```
zufayu, line 341: These numbers compare the new register kernel with the legacy kernel. Could we also include before/after register-kernel numbers to make sure this change does not slow down the fast path?

csrc/kernels/topk_softmax_kernels_group.cu
lirui927: update

op_tests/test_moeTopkSoftmax.py
lirui927: added

csrc/kernels/topk_softmax_kernels_group.cu
lirui927: added performance test, compared with the previous unordered implementation, the current kernel preserves deterministic legacy ordering for accuracy at the cost of lower performance.  It remains substantially faster than the legacy LDS kernel for most shapes, especially at larger `topk`.

valarLip (changes requested):
Reviewed at `1283b6bb6d`.

First, the part I want to be explicit about because it's the hard part of this PR and it holds up: I independently simulated the gfx9 wave64 DPP ladder in `wave_reduce` and confirmed `legacy_topk_lane_rank` is a bijection on [0,63] and reproduces the legacy tie winner exactly — zero mismatches across all 2016 lane pairs and 4000 random tie subsets. The pivot lower-bound argument, candidate-set completeness, the `(score desc, tie_rank asc)` total order, the exclusive-scan compaction, the softmax lane-strided sum matching `grouped_topk_kernel`, and the `expert+4` padding trick in the test (which does not perturb the legacy lane/round mapping) are all sound. I also chased two suspicions to ground and they're **not** problems: the non-gfx9 tie order is unreachable today because device `WARP_SIZE` is 32 off gfx9 so the host `reg_lanes == 64` gate can't be satisfied, and NaN cannot reach `cand_val` because `dev_max_<float>` is `__builtin_fmaxf`, so the pivot stays finite and NaN fails both filters.

With that said, I think there's one real correctness divergence, one sweep gap, and a performance cost that may be avoidable.

### The fast path selects `-INFINITY` experts and gives them real weight

The fallback (`v > -INFINITY`, line 956) and the legacy kernel (`tmp[i] > max_val`, line 677) both treat `-inf` as unselectable. `emit_legacy_ordered_topk` does not, so the same row gets two different answers depending only on `n_cand`.

Concretely: biased, `num_experts=128` (EPL=2), `topk=8`. Experts 0-2 have finite bias; experts 3..63 have `correction_bias = -inf`, so `score = sigmoid + (-inf) = -INF`; experts 64..127 have NaN gating. Lanes 32..63 own only NaN, so `lmax = fmaxf(-inf, NaN) = -INF`, and only lanes 0 and 1 have a finite `lmax` — the pivot comes out `-INFINITY`. `n_gt = 3`, `n_eq = 61`, so `n_cand = 64`, which passes the new gate at line 911. Inside the emit loop, for `i = 3..7` the target is `-INFINITY` and `const bool match = lane < n_cand && !selected && cv == target;` is **true** for the `-inf` candidates. Five masked-out experts get emitted with `winner_weight = sig_scores[winner_id]` — their pre-bias sigmoid, ~0.5 each — which then dominates the renormalized output.

Move one NaN expert so `n_cand` becomes 65 and the identical row takes the fallback instead and emits inert zero-weight slots. The PR's own `one_finite`/`all_nan` cases never reach this path, since `n_cand = 256 > 64` there.

### The divide-by-zero guard was added in one of four places

The PR adds `total > 0.0f` (line 388) and `sum > 0.0f` (line 999) in `topk_reg_kernel`, but leaves the same unguarded renormalization in the kernels it claims bit-identity with: `grouped_topk_kernel` line 715 `sum = routed_scaling_factor / sum;`, and `grouped_topk_opt_sort_kernel` lines 1461 and 1490 `__builtin_amdgcn_rcpf(sum)`. `grep -n 'rcpf(sum)\|routed_scaling_factor / sum'` returns four hits; one was fixed.

A router row whose top-k logits are all below roughly -89 makes every sigmoid exactly `+0.0f`, so `sum == 0`. Routed to the LDS path — `num_experts % 64 != 0`, an EPL outside the whitelist, or `topk < 4` — it still produces inf/NaN weights that NaN the whole token. Worth noting the guard is also a silent departure from the PR's own bitwise-equality claim (`total > 0.0f` is false for `-0.0f` and NaN too), and the new test routes exactly the two patterns that could trigger it away from the legacy comparison, so nothing catches it.

### Guards that hold only by coincidence, with nothing recording why

`(matches & (matches - 1)) == 0` is true for an empty mask as well as a single-bit one, and the empty case falls silently to `winner_lane = 0` (line 361, mirrored at 371). `if(lane == winner_lane) selected = true;` then re-sets `selected` on an already-consumed lane, and `topk_ids` contains the same expert twice — the exact corruption class this PR exists to eliminate. That's unreachable today only because of the new `n_cand >= topk` gate and because NaN can't reach `cand_val`; neither is asserted or documented in the helper, which is a free function whose caller is 550 lines away. And since `ci` is -1 for lanes past `n_cand`, `sig_scores[winner_id]` at line 376 would be an out-of-bounds LDS read if `n_cand` ever hit 0.

`return (legacy_topk_lane_rank(lane) << 5) | local_pos;` (line 325) reserves exactly 5 bits for `local_pos`, which saturates at exactly 31 for the gate's maximum `num_experts` — zero margin, and no `static_assert` tying it to `kMaxExpertsPerLane` 1200 lines away. `AITER_TOPK_REG_LAUNCH_EPLS` is documented at line 45 as user-overridable; add `X(64)` or raise `kMaxExpertsPerLane` for `num_experts=4096` and `local_pos` reaches `(15<<2)|3 = 63`, ORing bit 5 into the lane-rank field. `tie_rank` stops being injective, `legacy_topk_tie_rank(ci) == rank` matches two lanes, and `__ballot` picks the wrong one — wrong tie order and possible duplicate ids, with no compile error and no test at E=4096.

The synthetic-id branch's `__builtin_ctz(fallback_source)` and `1u << max_idx` (line 977) rest on an unasserted `topk <= 32`, and the comment at line 974 cites `num_experts >= 64` rather than the launcher gate at line 1552 that actually makes it safe (`topk <= reg_lanes / 2`). Widen that gate to `topk <= reg_lanes` — a natural-looking change, since `topk_ids` has 64 lanes available — and round 32 onward computes `fallback_source == 0`; `__builtin_ctz(0)` is UB (gfx9's `v_ffbl` returns -1), `max_idx` becomes -1, `if(max_idx < 32)` then evaluates `1u << -1` (also UB), and `topk_ids` receives -1: a negative expert index fed straight into the downstream MoE scatter. The comment as written doesn't warn the next editor which invariant is load-bearing.

### The cost of the ordering guarantee

`emit_legacy_ordered_topk` reintroduces what the pivot fast path existed to remove. The deleted comment said the pivot path "replaces topk wave argmax rounds (each a 6-stage dependent bpermute chain) with two wave sorts". The new `for(int i = 0; i < topk; ++i)` loop does, per round: `__shfl(sorted_cv, i)` → `__ballot` → on any tie (which is exactly the GLM / `sixteen_way_tie` case this PR targets) a full 6-stage `wave_reduce<int>` → a second `__ballot` → `readlane` → LDS load, all serially dependent through `selected`. `legacy_topk_tie_rank(ci)` is loop-invariant in `ci` and is recomputed at both line 367 and line 370 every iteration. Separately, the softmax path (866-881) now stores all `num_experts` values to LDS, fences, re-reads them lane-strided, then stores them again, where it previously accumulated `tsum` in a register.

The PR's table shows the register kernel slower than main on every shape measured — +24% on the GLM E=256/topk=8 shape, +112% to +173% at E=128/topk=32. I realize the regression was acknowledged in review, but I don't think it's forced: packing a single 64-bit key (score bits high, `legacy_topk_tie_rank` low), sorting it with the existing `warp_bitonic_merge_sort_to_reg`, and scattering all topk winners in one `cumsum_topk_with_pivot` pass preserves the same total order without the serial rounds — which is structurally what the pre-PR code did.

On the same kernel there's also a free occupancy regression: the launcher still sizes LDS for the deleted `final_idx` buffer. Line 1555 allocates `num_experts*sizeof(float) + reg_lanes*sizeof(float) + 2*reg_lanes*sizeof(int)`, but the kernel now indexes only `sig_scores[num_experts] + cand_val[64] + cand_idx[64]`. The diff removed `int* final_idx = cand_idx + WARP_SIZE;` without narrowing the allocation, so 256 bytes of dynamic LDS per block are reserved and never touched (at E=256: 1792 B allocated, 1536 B used). On a one-wave-per-block kernel LDS is the occupancy limiter. Should be `reg_lanes * sizeof(float) + reg_lanes * sizeof(int)`.

### An ordering hazard held by an accidental dependency

The new softmax LDS round trip is asymmetrically fenced. The write at 868 is followed by `__builtin_amdgcn_wave_barrier()` before the strided read at 874-875, but the second write to those same LDS words at line 880 has no barrier before it. The write-after-read is covered only by a data dependency: the stored value `v /= tsum` transitively depends on every lane's load through the DPP `wave_reduce`. Break that — hoist `tsum`, short-circuit on `tsum != 1.0f`, cache it, or split normalization into its own loop — and the WAR becomes real: lanes read post-normalization values into the sum and a subset of experts gets normalized twice, silently. The surrounding comments document the barrier discipline for the write-then-read direction only.

### Naming and scope of the equivalence claim

When `!has_valid_candidate`, the synthetic filler id is assigned back into `max_idx` (line 980), which then no longer means "index of the max" — and line 986's `for_each_owned([&](float& v, int id) { v = id == max_idx ? -INFINITY : v; })` destroys the score of an expert in [0,31] that was never a winner. Harmless today only because `has_valid_candidate` is monotone (scores are only retired, never restored, and `v > -INFINITY` excludes both `-inf` and NaN) — an invariant stated nowhere. Any future two-phase fallback, per-group re-enable, or NaN-tolerant guard would find that expert already destroyed and drop a genuine top-k winner. It's also a wasted wave-wide compare-and-select per synthetic round, on rows where by definition nothing is left to retire. Renaming to `emit_idx`, or skipping the retire when the candidate is synthetic, would fix both.

The rewritten header at line 730 now reads "Single-group top-k that keeps scores in VGPRs and preserves legacy ordering", but the fallback deliberately does *not* match `grouped_topk_kernel` on degenerate rows, and the comment no longer says so. Traced against the test's own `one_finite` pattern (bias `-inf` except `bias[2]=0`, gating 0): legacy emits ids `[2,1,2,3,4,5,6,7]` — a duplicate id 2, because round k keeps its `int max_idx = k` default — with weight 0.3125 in all eight slots, while the new reg path emits eight unique ids with weights `[2.5,0,0,...]`. For all-NaN rows legacy propagates NaN weights and the reg path emits 0.0. The divergence is an improvement; the comment just shouldn't promise unconditional bit-exactness when the test only asserts it for the random / all_equal / sixteen_way_tie patterns.

### Test coverage

The new call convention isn't exercised. Every `stable_cases` entry uses `token=1` and a plain contiguous `torch.empty((token, topk))`, so `stride_tk` and the multi-block grid are dead code in the only bitwise-exact test in the file — and pointer rebasing is precisely what this PR introduces (`emit_legacy_ordered_topk(..., topk_weights + token_idx * stride_tk, topk_ids + token_idx * stride_tk, ...)` at line 932, with the helper then writing `topk_weights[lane]` off that pointer). At `token=1` with a contiguous output, `token_idx * stride_tk` is identically 0, so a wrong stride, a missing stride, or row-to-row interference in the new wave-level emit is invisible. Every other test in this file uses `torch.empty_strided((token, topk), (topk + 10, 1))` (lines 293-294, 340-341, 502-503) because production callers pass a strided slice, and the adjacent reg driver uses `reg_tokens = [1, 128]`.

The bitwise assertions run on unseeded data. `gating`/`bias` come from `torch.randn` and `torch.manual_seed` appears nowhere in the file, while the four "random" cases assert exact bit equality of ids *and* weights between two kernels that resolve ties by completely different mechanisms (`legacy_topk_tie_rank` over a wave sort vs the DPP `warpReduceMax` chain). bf16 gating has 8 mantissa bits, so near-ties and exact fp32 score ties over 256-2048 experts aren't rare. If the tie table is ever wrong for some lane pair, the failure shows up on maybe one CI run in N, on a row nobody can regenerate; a re-run goes green and it gets written off as a flake. A fixed seed costs nothing and makes the exact-match contract enforceable.

`return {"err": 0, "us": us}` (line 473) hard-codes the error column. Every other driver in this file returns a real metric (`ret["err"] = checkAllclose(...)` at 129, 336, 518, 634); the summary table at line 906 prints `err=0` for all 17 rows regardless of what was compared, and the only real checks are bare `assert` statements at 432-435 that vanish under `python -O`. The same driver also ignores `args.iters`/`args.warmup` — the adjacent block at 869-870 threads them through — so 17 correctness-only cases each run `run_perftest`'s default 101 timed iterations with `num_rotate_args=0`, triggering a device-memory-profiling probe plus up to 100 `copy.deepcopy` of the arg tuple per case, inside a profiler context, against the 36 s budget in `.github/scripts/split_tests.sh`.

Finally, `assert mode == "biased" and need_renorm` (line 432) is vacuously true — all three degenerate entries at 901-903 hard-code `('biased', True)` — and it papers over the fact that the degenerate patterns run only for biased+renorm. `has_valid_candidate`, `unused_from_k`, the `selected_low_mask` bookkeeping and the `sum > 0.0f` guard all live outside `if constexpr(isBiased)` and are separately instantiated for `isSoftmax=true` and for `isBiased=false/isSoftmax=false`. An all-NaN row under `mode='softmax'` is a genuinely different path (`lmax` comes out `-inf`, `expf(-inf - -inf)` is NaN, `tsum` is NaN), and `need_renorm=False` takes the `w *= routed_scaling_factor` branch. Two further traps: nothing asserts the reg path was actually dispatched, so trimming `AITER_TOPK_REG_LAUNCH_EPLS` or raising `AITER_TOPK_REG_MIN_TOPK` (the header invites both) turns every `assert_close` into `grouped_topk_kernel(E)` vs `grouped_topk_kernel(E+4)` — a tautology that passes while testing nothing; and on the sigmoid branch the `-inf` padding becomes a *finite* reference score (`sigmoid(-inf) == 0.0`), i.e. legitimately selectable, so any future pattern whose real scores reach 0 makes the reference emit ids `>= expert` that the reg kernel can never produce. The file already knows how to pin this kind of assumption — see `assert gating_odd.stride(0) % 2 == 1` at line 915.

### Smaller

The new code introduces this TU's first unqualified `uint64_t` (359, 370) and `uint32_t` (950, 977-979) without adding `<cstdint>`, in the very hunk that reorders the include block. That block has only `<cfloat>` and `<hip/hip_runtime.h>`, so both types resolve transitively through `hip_runtime.h` / `aiter_hip_common.h`; a ROCm header refactor or another reorder of the block this PR already touched breaks `module_moe_asm` with `unknown type name 'uint32_t'` — at JIT time, on every MoE routing call, not at a build step CI would catch.

Relatedly, a `static_assert(defined(__GFX9__) && WARP_SIZE == 64)` in the helper would be worth adding. `legacy_topk_lane_rank` transcribes the `#if defined(__GFX9__)` branch of `wave_reduce` (hip_reduce.h:270-308); the `#else` branch (`row_xmask:7/15` + `permlanex16`, no >32 stage, no broadcast) yields a different tie order. Unreachable today, as noted at the top — but `hip_reduce.h` is a shared header with nothing recording that its DPP ctrl constants are now an ABI for this kernel, and a build error is a much better outcome than a silent wrong answer.

A few things I noticed but am not asking for: the fallback open-codes `warpReduceMax`'s body (line 963) and re-declares min/max lambdas duplicating `aiter::Min` / `aiter::Max` / `aiter::ArgMax`, whose header comment says they exist precisely so tie/NaN conventions aren't re-derived per call site; the untouched `grouped_topk_kernel` has a pre-existing OOB LDS write (`int max_idx = NUM_GRP;` at 641 with a strict `>`, so an all-`-inf` group score set writes `group_scores[NUM_GRP]`, which in the biased LDS layout aliases `sig_scores[0]`); the new fallback body is instantiated across 8 EPLs x 3 modes x 2 rowVec2 variants x every dtype and now carries two `legacy_topk_tie_rank` evaluations inside a fully-unrolled EPL=32 loop plus a word-wise 8-byte `wave_reduce`, inflating `module_moe_asm` JIT build wall for a path the code documents as never taken; and `aiter.biased_grouped_topk()` (aiter/ops/topk.py:191) still switches to `moe_fused_gate` above `cu_num*212` tokens, so the newly-advertised ordering is scoped to the HIP path (not an issue for GLM's G=1 shapes, where `num_experts // num_expert_group > 32` forces the HIP path anyway).

### Summary

The tie table is the hard part and it's right. What I'd want before merge is the `-inf` divergence closed so the fast path and the fallback agree, the other three renormalization sites guarded in the same change, and a look at whether the packed-key single-sort formulation recovers the 24%-173%. The unasserted invariants around `matches == 0`, the 5-bit `local_pos` field and `topk <= 32` are each a one-line `static_assert` or comment and would make this much safer to edit next year.

lirui927:
> Reviewed at `1283b6bb6d`.
> 
> First, the part I want to be explicit about because it's the hard part of this PR and it holds up: I independently simulated the gfx9 wave64 DPP ladder in `wave_reduce` and confirmed `legacy_topk_lane_rank` is a bijection on [0,63] and reproduces the legacy tie winner exactly — zero mismatches across all 2016 lane pairs and 4000 random tie subsets. The pivot lower-bound argument, candidate-set completeness, the `(score desc, tie_rank asc)` total order, the exclusive-scan compaction, the softmax lane-strided sum matching `grouped_topk_kernel`, and the `expert+4` padding trick in the test (which does not perturb the legacy lane/round mapping) are all sound. I also chased two suspicions to ground and they're **not** problems: the non-gfx9 tie order is unreachable today because device `WARP_SIZE` is 32 off gfx9 so the host `reg_lanes == 64` gate can't be satisfied, and NaN cannot reach `cand_val` because `dev_max_<float>` is `__builtin_fmaxf`, so the pivot stays finite and NaN fails both filters.
> 
> With that said, I think there's one real correctness divergence, one sweep gap, and a performance cost that may be avoidable.
> 
> ### The fast path selects `-INFINITY` experts and gives them real weight
> The fallback (`v > -INFINITY`, line 956) and the legacy kernel (`tmp[i] > max_val`, line 677) both treat `-inf` as unselectable. `emit_legacy_ordered_topk` does not, so the same row gets two different answers depending only on `n_cand`.
> 
> Concretely: biased, `num_experts=128` (EPL=2), `topk=8`. Experts 0-2 have finite bias; experts 3..63 have `correction_bias = -inf`, so `score = sigmoid + (-inf) = -INF`; experts 64..127 have NaN gating. Lanes 32..63 own only NaN, so `lmax = fmaxf(-inf, NaN) = -INF`, and only lanes 0 and 1 have a finite `lmax` — the pivot comes out `-INFINITY`. `n_gt = 3`, `n_eq = 61`, so `n_cand = 64`, which passes the new gate at line 911. Inside the emit loop, for `i = 3..7` the target is `-INFINITY` and `const bool match = lane < n_cand && !selected && cv == target;` is **true** for the `-inf` candidates. Five masked-out experts get emitted with `winner_weight = sig_scores[winner_id]` — their pre-bias sigmoid, ~0.5 each — which then dominates the renormalized output.
> 
> Move one NaN expert so `n_cand` becomes 65 and the identical row takes the fallback instead and emits inert zero-weight slots. The PR's own `one_finite`/`all_nan` cases never reach this path, since `n_cand = 256 > 64` there.
> 
> ### The divide-by-zero guard was added in one of four places
> The PR adds `total > 0.0f` (line 388) and `sum > 0.0f` (line 999) in `topk_reg_kernel`, but leaves the same unguarded renormalization in the kernels it claims bit-identity with: `grouped_topk_kernel` line 715 `sum = routed_scaling_factor / sum;`, and `grouped_topk_opt_sort_kernel` lines 1461 and 1490 `__builtin_amdgcn_rcpf(sum)`. `grep -n 'rcpf(sum)\|routed_scaling_factor / sum'` returns four hits; one was fixed.
> 
> A router row whose top-k logits are all below roughly -89 makes every sigmoid exactly `+0.0f`, so `sum == 0`. Routed to the LDS path — `num_experts % 64 != 0`, an EPL outside the whitelist, or `topk < 4` — it still produces inf/NaN weights that NaN the whole token. Worth noting the guard is also a silent departure from the PR's own bitwise-equality claim (`total > 0.0f` is false for `-0.0f` and NaN too), and the new test routes exactly the two patterns that could trigger it away from the legacy comparison, so nothing catches it.
> 
> ### Guards that hold only by coincidence, with nothing recording why
> `(matches & (matches - 1)) == 0` is true for an empty mask as well as a single-bit one, and the empty case falls silently to `winner_lane = 0` (line 361, mirrored at 371). `if(lane == winner_lane) selected = true;` then re-sets `selected` on an already-consumed lane, and `topk_ids` contains the same expert twice — the exact corruption class this PR exists to eliminate. That's unreachable today only because of the new `n_cand >= topk` gate and because NaN can't reach `cand_val`; neither is asserted or documented in the helper, which is a free function whose caller is 550 lines away. And since `ci` is -1 for lanes past `n_cand`, `sig_scores[winner_id]` at line 376 would be an out-of-bounds LDS read if `n_cand` ever hit 0.
> 
> `return (legacy_topk_lane_rank(lane) << 5) | local_pos;` (line 325) reserves exactly 5 bits for `local_pos`, which saturates at exactly 31 for the gate's maximum `num_experts` — zero margin, and no `static_assert` tying it to `kMaxExpertsPerLane` 1200 lines away. `AITER_TOPK_REG_LAUNCH_EPLS` is documented at line 45 as user-overridable; add `X(64)` or raise `kMaxExpertsPerLane` for `num_experts=4096` and `local_pos` reaches `(15<<2)|3 = 63`, ORing bit 5 into the lane-rank field. `tie_rank` stops being injective, `legacy_topk_tie_rank(ci) == rank` matches two lanes, and `__ballot` picks the wrong one — wrong tie order and possible duplicate ids, with no compile error and no test at E=4096.
> 
> The synthetic-id branch's `__builtin_ctz(fallback_source)` and `1u << max_idx` (line 977) rest on an unasserted `topk <= 32`, and the comment at line 974 cites `num_experts >= 64` rather than the launcher gate at line 1552 that actually makes it safe (`topk <= reg_lanes / 2`). Widen that gate to `topk <= reg_lanes` — a natural-looking change, since `topk_ids` has 64 lanes available — and round 32 onward computes `fallback_source == 0`; `__builtin_ctz(0)` is UB (gfx9's `v_ffbl` returns -1), `max_idx` becomes -1, `if(max_idx < 32)` then evaluates `1u << -1` (also UB), and `topk_ids` receives -1: a negative expert index fed straight into the downstream MoE scatter. The comment as written doesn't warn the next editor which invariant is load-bearing.
> 
> ### The cost of the ordering guarantee
> `emit_legacy_ordered_topk` reintroduces what the pivot fast path existed to remove. The deleted comment said the pivot path "replaces topk wave argmax rounds (each a 6-stage dependent bpermute chain) with two wave sorts". The new `for(int i = 0; i < topk; ++i)` loop does, per round: `__shfl(sorted_cv, i)` → `__ballot` → on any tie (which is exactly the GLM / `sixteen_way_tie` case this PR targets) a full 6-stage `wave_reduce<int>` → a second `__ballot` → `readlane` → LDS load, all serially dependent through `selected`. `legacy_topk_tie_rank(ci)` is loop-invariant in `ci` and is recomputed at both line 367 and line 370 every iteration. Separately, the softmax path (866-881) now stores all `num_experts` values to LDS, fences, re-reads them lane-strided, then stores them again, where it previously accumulated `tsum` in a register.
> 
> The PR's table shows the register kernel slower than main on every shape measured — +24% on the GLM E=256/topk=8 shape, +112% to +173% at E=128/topk=32. I realize the regression was acknowledged in review, but I don't think it's forced: packing a single 64-bit key (score bits high, `legacy_topk_tie_rank` low), sorting it with the existing `warp_bitonic_merge_sort_to_reg`, and scattering all topk winners in one `cumsum_topk_with_pivot` pass preserves the same total order without the serial rounds — which is structurally what the pre-PR code did.
> 
> On the same kernel there's also a free occupancy regression: the launcher still sizes LDS for the deleted `final_idx` buffer. Line 1555 allocates `num_experts*sizeof(float) + reg_lanes*sizeof(float) + 2*reg_lanes*sizeof(int)`, but the kernel now indexes only `sig_scores[num_experts] + cand_val[64] + cand_idx[64]`. The diff removed `int* final_idx = cand_idx + WARP_SIZE;` without narrowing the allocation, so 256 bytes of dynamic LDS per block are reserved and never touched (at E=256: 1792 B allocated, 1536 B used). On a one-wave-per-block kernel LDS is the occupancy limiter. Should be `reg_lanes * sizeof(float) + reg_lanes * sizeof(int)`.
> 
> ### An ordering hazard held by an accidental dependency
> The new softmax LDS round trip is asymmetrically fenced. The write at 868 is followed by `__builtin_amdgcn_wave_barrier()` before the strided read at 874-875, but the second write to those same LDS words at line 880 has no barrier before it. The write-after-read is covered only by a data dependency: the stored value `v /= tsum` transitively depends on every lane's load through the DPP `wave_reduce`. Break that — hoist `tsum`, short-circuit on `tsum != 1.0f`, cache it, or split normalization into its own loop — and the WAR becomes real: lanes read post-normalization values into the sum and a subset of experts gets normalized twice, silently. The surrounding comments document the barrier discipline for the write-then-read direction only.
> 
> ### Naming and scope of the equivalence claim
> When `!has_valid_candidate`, the synthetic filler id is assigned back into `max_idx` (line 980), which then no longer means "index of the max" — and line 986's `for_each_owned([&](float& v, int id) { v = id == max_idx ? -INFINITY : v; })` destroys the score of an expert in [0,31] that was never a winner. Harmless today only because `has_valid_candidate` is monotone (scores are only retired, never restored, and `v > -INFINITY` excludes both `-inf` and NaN) — an invariant stated nowhere. Any future two-phase fallback, per-group re-enable, or NaN-tolerant guard would find that expert already destroyed and drop a genuine top-k winner. It's also a wasted wave-wide compare-and-select per synthetic round, on rows where by definition nothing is left to retire. Renaming to `emit_idx`, or skipping the retire when the candidate is synthetic, would fix both.
> 
> The rewritten header at line 730 now reads "Single-group top-k that keeps scores in VGPRs and preserves legacy ordering", but the fallback deliberately does _not_ match `grouped_topk_kernel` on degenerate rows, and the comment no longer says so. Traced against the test's own `one_finite` pattern (bias `-inf` except `bias[2]=0`, gating 0): legacy emits ids `[2,1,2,3,4,5,6,7]` — a duplicate id 2, because round k keeps its `int max_idx = k` default — with weight 0.3125 in all eight slots, while the new reg path emits eight unique ids with weights `[2.5,0,0,...]`. For all-NaN rows legacy propagates NaN weights and the reg path emits 0.0. The divergence is an improvement; the comment just shouldn't promise unconditional bit-exactness when the test only asserts it for the random / all_equal / sixteen_way_tie patterns.
> 
> ### Test coverage
> The new call convention isn't exercised. Every `stable_cases` entry uses `token=1` and a plain contiguous `torch.empty((token, topk))`, so `stride_tk` and the multi-block grid are dead code in the only bitwise-exact test in the file — and pointer rebasing is precisely what this PR introduces (`emit_legacy_ordered_topk(..., topk_weights + token_idx * stride_tk, topk_ids + token_idx * stride_tk, ...)` at line 932, with the helper then writing `topk_weights[lane]` off that pointer). At `token=1` with a contiguous output, `token_idx * stride_tk` is identically 0, so a wrong stride, a missing stride, or row-to-row interference in the new wave-level emit is invisible. Every other test in this file uses `torch.empty_strided((token, topk), (topk + 10, 1))` (lines 293-294, 340-341, 502-503) because production callers pass a strided slice, and the adjacent reg driver uses `reg_tokens = [1, 128]`.
> 
> The bitwise assertions run on unseeded data. `gating`/`bias` come from `torch.randn` and `torch.manual_seed` appears nowhere in the file, while the four "random" cases assert exact bit equality of ids _and_ weights between two kernels that resolve ties by completely different mechanisms (`legacy_topk_tie_rank` over a wave sort vs the DPP `warpReduceMax` chain). bf16 gating has 8 mantissa bits, so near-ties and exact fp32 score ties over 256-2048 experts aren't rare. If the tie table is ever wrong for some lane pair, the failure shows up on maybe one CI run in N, on a row nobody can regenerate; a re-run goes green and it gets written off as a flake. A fixed seed costs nothing and makes the exact-match contract enforceable.
> 
> `return {"err": 0, "us": us}` (line 473) hard-codes the error column. Every other driver in this file returns a real metric (`ret["err"] = checkAllclose(...)` at 129, 336, 518, 634); the summary table at line 906 prints `err=0` for all 17 rows regardless of what was compared, and the only real checks are bare `assert` statements at 432-435 that vanish under `python -O`. The same driver also ignores `args.iters`/`args.warmup` — the adjacent block at 869-870 threads them through — so 17 correctness-only cases each run `run_perftest`'s default 101 timed iterations with `num_rotate_args=0`, triggering a device-memory-profiling probe plus up to 100 `copy.deepcopy` of the arg tuple per case, inside a profiler context, against the 36 s budget in `.github/scripts/split_tests.sh`.
> 
> Finally, `assert mode == "biased" and need_renorm` (line 432) is vacuously true — all three degenerate entries at 901-903 hard-code `('biased', True)` — and it papers over the fact that the degenerate patterns run only for biased+renorm. `has_valid_candidate`, `unused_from_k`, the `selected_low_mask` bookkeeping and the `sum > 0.0f` guard all live outside `if constexpr(isBiased)` and are separately instantiated for `isSoftmax=true` and for `isBiased=false/isSoftmax=false`. An all-NaN row under `mode='softmax'` is a genuinely different path (`lmax` comes out `-inf`, `expf(-inf - -inf)` is NaN, `tsum` is NaN), and `need_renorm=False` takes the `w *= routed_scaling_factor` branch. Two further traps: nothing asserts the reg path was actually dispatched, so trimming `AITER_TOPK_REG_LAUNCH_EPLS` or raising `AITER_TOPK_REG_MIN_TOPK` (the header invites both) turns every `assert_close` into `grouped_topk_kernel(E)` vs `grouped_topk_kernel(E+4)` — a tautology that passes while testing nothing; and on the sigmoid branch the `-inf` padding becomes a _finite_ reference score (`sigmoid(-inf) == 0.0`), i.e. legitimately selectable, so any future pattern whose real scores reach 0 makes the reference emit ids `>= expert` that the reg kernel can never produce. The file already knows how to pin this kind of assumption — see `assert gating_odd.stride(0) % 2 == 1` at line 915.
> 
> ### Smaller
> The new code introduces this TU's first unqualified `uint64_t` (359, 370) and `uint32_t` (950, 977-979) without adding `<cstdint>`, in the very hunk that reorders the include block. That block has only `<cfloat>` and `<hip/hip_runtime.h>`, so both types resolve transitively through `hip_runtime.h` / `aiter_hip_common.h`; a ROCm header refactor or another reorder of the block this PR already touched breaks `module_moe_asm` with `unknown type name 'uint32_t'` — at JIT time, on every MoE routing call, not at a build step CI would catch.
> 
> Relatedly, a `static_assert(defined(__GFX9__) && WARP_SIZE == 64)` in the helper would be worth adding. `legacy_topk_lane_rank` transcribes the `#if defined(__GFX9__)` branch of `wave_reduce` (hip_reduce.h:270-308); the `#else` branch (`row_xmask:7/15` + `permlanex16`, no >32 stage, no broadcast) yields a different tie order. Unreachable today, as noted at the top — but `hip_reduce.h` is a shared header with nothing recording that its DPP ctrl constants are now an ABI for this kernel, and a build error is a much better outcome than a silent wrong answer.
> 
> A few things I noticed but am not asking for: the fallback open-codes `warpReduceMax`'s body (line 963) and re-declares min/max lambdas duplicating `aiter::Min` / `aiter::Max` / `aiter::ArgMax`, whose header comment says they exist precisely so tie/NaN conventions aren't re-derived per call site; the untouched `grouped_topk_kernel` has a pre-existing OOB LDS write (`int max_idx = NUM_GRP;` at 641 with a strict `>`, so an all-`-inf` group score set writes `group_scores[NUM_GRP]`, which in the biased LDS layout aliases `sig_scores[0]`); the new fallback body is instantiated across 8 EPLs x 3 modes x 2 rowVec2 variants x every dtype and now carries two `legacy_topk_tie_rank` evaluations inside a fully-unrolled EPL=32 loop plus a word-wise 8-byte `wave_reduce`, inflating `module_moe_asm` JIT build wall for a path the code documents as never taken; and `aiter.biased_grouped_topk()` (aiter/ops/topk.py:191) still switches to `moe_fused_gate` above `cu_num*212` tokens, so the newly-advertised ordering is scoped to the HIP path (not an issue for GLM's G=1 shapes, where `num_experts // num_expert_group > 32` forces the HIP path anyway).
> 
> ### Summary
> The tie table is the hard part and it's right. What I'd want before merge is the `-inf` divergence closed so the fast path and the fallback agree, the other three renormalization sites guarded in the same change, and a look at whether the packed-key single-sort formulation recovers the 24%-173%. The unasserted invariants around `matches == 0`, the 5-bit `local_pos` field and `topk <= 32` are each a one-line `static_assert` or comment and would make this much safer to edit next year.

Thanks for the detailed review. Here is a priority-ordered summary of the changes:

P0: Correctness
Fixed the -Inf fast-path divergence: the candidate fast path is now used only when pivot > -Inf.
Added zero-denominator protection to the register, legacy LDS, and optimized-sort paths to prevent Inf/NaN weights.
Degenerate register inputs now use unique valid expert IDs with zero weights.
P1: Test Coverage
Added fixed seeds, multi-token and strided outputs, renormalization on/off, and biased sigmoid, unbiased sigmoid, and softmax coverage.
Added masked scores, a single finite candidate, all-NaN inputs, and all-zero sigmoid inputs.
Replaced bare assertions and hard-coded err=0 with actual mismatch checks.
Reduced correctness cases to two iterations to limit CI runtime.
All 24 ordered/register cases and the LDS/opt-sort zero-denominator cases pass. The 2,880-case grid contains no out-of-range IDs, duplicate IDs, or non-finite weights.
P1: Performance
Implemented and tested an exact uint64 packed-key approach. It improved k=32, but regressed GLM k=8 by 6–9% and k=4 by 16–23%, so it was not retained.
Caching the tie rank was also tested but provided no performance benefit, so it was not retained.
Removed the unused 256 bytes of dynamic LDS left over from final_idx.
Used a wave-uniform pivot > -Inf gate instead of a per-expert check, avoiding the EPL=32 performance regression.
P2: Maintainability
Added <cstdint> and an EPL <= 32 constraint for the tie-rank encoding.
The existing launcher gate continues to enforce wave64 and topk <= 32; these assumptions are documented next to the helper and fallback.
Added an explicit wave barrier between the softmax LDS reads and writes.
Documented the monotonicity assumption behind synthetic retirement without adding another branch that would increase register pressure.

zufayu (approved):

zufayu:
Thanks, the latest update appears to address the main correctness concerns raised above.

One possible way to keep the scope manageable would be to address the remaining small documentation and invariant-safety items in this PR, and consider the packed-key/single-sort performance optimization separately, since it would require another round of correctness and performance validation. Just a suggestion — I’ll defer to the maintainers on whether they prefer to keep everything in this PR.

XiaobingSuper (approved):

PR #5736, opened 2026-09-21
[Triton/Gluon] [gfx950] pa_decode_sparse: bound split-K by wave count and tile budget (up to +22%)

tianxiaojiang4:
Bounds the split-K count in the gfx950 gluon sparse-decode driver from both sides: the launch stops
overshooting one wave of workgroups, and it stops cutting splits once each workgroup carries enough
KV tiles to pay for itself.

## What changes

`_decode_num_splits_occ` bounds the split count by `cta_cap = 2 * num_sms // base_wg`, which lets
`base_wg * splits` exceed `num_sms`. Capping that alone is not enough, though — past a point, cutting
splits stops removing overhead and starts removing parallelism. So the count is bounded by both:

```python
wave_cap = 1 << max(0, (num_sms // base_wg).bit_length() - 1)
return max(1, min(cta_cap, tiles, max(wave_cap, tiles // 4)))
```

`tiles // 4` is not a new constant — it is the split floor the `base_wg >= num_sms` branch of this
same function already uses. Only the `base_wg < num_sms` branch changes.

## Why both bounds are needed

Every split costs a fixed amount — its own partial write and its own Q re-read, both paid **inside the
attention kernel** — and that cost does not depend on how much KV the split actually covers. So the
tradeoff is per-split fixed cost against parallelism, and where it balances depends on how many KV
tiles a workgroup has to amortise it over. That is set by top-k, which is why one bound cannot do it.

An earlier revision of this PR had only `wave_cap`. Measured over 24 shapes, that is a **net
regression**:

| | wins | losses | worst | best | mean |
|---|---|---|---|---|---|
| `wave_cap` only | 6 | **15** | **-43.1 %** | +22.1 % | **-7.6 %** |
| with the `tiles // 4` floor | **10** | **0** | -1.0 % | +21.9 % | **+4.2 %** |

## Measurements

MI355X / gfx950 (256 CU), DeepSeek-V4.1 decode shape: 16 heads, head_dim 512 (448 NoPE + 64 RoPE),
packed fp8 cache, 128-slot SWA window plus a top-k list per query, 8 GiB KV pool. Median of 100
samples of 4 calls in one HIP graph. Each arm runs its own `_decode_num_splits_occ` — `kv_splits` is
not pinned — so this is what merging actually gives you. Speedup against current `main`:

| queries | top-k 512 | top-k 1024 | top-k 2048 |
|---|---|---|---|
| 6 | +0.4 % | +0.5 % | -0.2 % |
| 36 | **+15.6 %** | **+21.9 %** | **+14.8 %** |
| 48 | **+16.3 %** | +5.7 % | +6.0 % |
| 96 | +5.5 % | +8.1 % | -0.2 % |
| 126 | +5.2 % | -0.1 % | +0.3 % |
| 150 | -1.0 % | -0.5 % | +0.0 % |
| 192 | -0.3 % | -0.2 % | +0.1 % |
| 252 | +0.4 % | +0.2 % | +2.9 % |

Against a per-point sweep of `kv_splits` over {1, 2, 3, 4, 5, 6, 8, 10, 14, 16} at every shape, the
guarded rule lands on or within 2.5 % of the measured optimum at every one of the 24 points. The
worst cell is -1.0 %.

## Tests

`test_decode_num_splits_occ_bounds` asserts the four bounds the split count is built from — at most
`tiles`, at most `cta_cap`, at least `tiles // 4`, and within one wave where the tile floor is not
what holds the count up — plus monotonicity in top-k, over 165 shapes.

`num_sms` is patched rather than read from the device, so the body allocates nothing and the bounds
are checked at partitioned CU counts (256 / 128 / 64 / 32) rather than only whatever the runner has.
The `torch.cuda.is_available()` guard matches the in-body form the other tests in this module use.

It discriminates, and the two failing arms fail for different reasons:

```
main                4 failed   cu=32 q=1 topk=8192: 64 workgroups over 32 CUs    <- the wave ceiling
wave cap, no floor  4 failed   cu=32 q=6 topk=2048: 4 splits leaves 8 tiles/WG   <- the floor
this PR             4 passed
```

Note it fails on `main` by design: the wave ceiling is behaviour this PR introduces, so asserting it
makes the contract explicit rather than incidental. The floor assertion is the regression guard, and
it is not masked by the ceiling — the un-floored version trips the floor first.

The H=16 subset of the existing kernel test is **42 passed, 10 skipped**, identical across all three.

## Numerics

Unchanged in kind: the split count only changes how many partials the fp32 combine sums, as it
already does across the existing `cta_cap` and `tiles` bounds. Checked directly — output max
absolute deviation across every split count at every shape is **<= 9.8e-04** against the single-split
result, which is fp8 accumulation order, not a behaviour change.

`op_tests/triton_tests/attention/test_pa_decode_sparse.py`, the H=16 subset: **42 passed, 10 skipped**,
identical on `main`, on the bare cap, and with the floor.

## On the mechanism

An earlier revision explained this as a partly filled second wave costing a full wave of latency. A
`rocprofv3 --kernel-trace` of the launch does not support that, and the description is corrected:

- The saving is **entirely in the attention kernel** (103 % of the delta at 48 queries, 8 -> 4
  splits); `_aiter_sparse_decode_reduce_kernel` is flat to +0.1 us. The partial write happens inside
  the attention kernel before the workgroup exits, which is where the per-split cost actually lands.
- The kernel reports `LDS_Block_Size = 0` — no static LDS — so occupancy is VGPR-limited at 128
  VGPRs, giving 4 workgroups per CU. 384 workgroups is nowhere near a wave boundary, and there is no
  partly filled second wave to pay for.

`wave_cap` is still the right shape for the upper bound; the reasoning for it is amortisation, not
wave quantisation.

## Reproducing

Measured with a standalone benchmark that calls `pa_decode_sparse` directly, independent of any
serving stack, and cross-checked against an sglang-based harness that shares no code with it. Both
agree on where the win and loss zones fall. Happy to attach the benchmark if useful.

One methodology note for anyone re-running: a first measurement process reads 3-11 % fast from cold
start, so arms must be compared back to back within one process rather than across processes.


tianxiaojiang4:
@cagrikymk @k50112113 — would either of you have time to look at this?

Tagging you two specifically: `git blame` puts 13 of the 15 lines of `_decode_num_splits_occ` on
#4919, and 199 of 200 lines of `_pa_decode_sparse_gfx950_gluon` alongside them; and the last three
merged changes to this file were approved by @k50112113.

Short version: the first revision of this PR capped split-K at one wave, which is +9..18 % at top-k
512 but **-15..26 % at top-k 2048** — 6 wins and 15 losses over 24 shapes. Flooring the cap at
`tiles // 4` makes it 10 wins, 0 losses. Measured two ways that share no code (a standalone bench
against `pa_decode_sparse` directly, and an sglang-based harness), with a bounds test that fails on
the un-floored version.

@cagrikymk, two questions where you are the only one who can settle it:

1. **Was `cta_cap = 2 * num_sms` a deliberate occupancy target?** I have been reading it as "two
   workgroups per CU", and the PR argues that is too loose below one workgroup per CU. If it encodes
   something else — a Gluon scheduling constraint, say — the upper bound here is wrong in kind and I
   would rather know before this lands.

2. **`rocprofv3 --kernel-trace` reports `LDS_Block_Size = 0` for `_sparse_mla_*`.** Taken at face
   value the kernel allocates no static LDS and is VGPR-limited at 4 workgroups/CU, which is what led
   me to drop the "partly filled wave" explanation from the description. Is that real, or does Gluon
   allocate LDS in a way the trace does not attribute? It also bears on #5294, whose BLOCK_K argument
   assumes 64 KB/workgroup.

Happy to rerun anything, add shapes, or split the floor into its own PR if you would rather take the
two bounds separately.

PR #5745, opened 2026-09-22
[HIP] Guard GroupNorm vectorization at channel boundaries

UD-mmcminn:
## Summary

- Require the per-channel element count to be divisible by four before enabling GroupNorm vectorization.
- Keep Python workspace sizing consistent with the HIP launch configuration.
- Make GroupNorm correctness failures fatal and report the worst mismatch.
- Add a regression shape where the group span is divisible by four but the channel span is not.

## Motivation

The vectorized output path applies the affine parameters selected for the vector's first element to all four values. Therefore, a vector must never cross a channel boundary.

The previous condition checked whether the complete group span was divisible by four. That could enable vectorization when an individual channel had an incompatible element count.

This is a general correctness fix and is not specific to gfx908.

## Testing

Tested on AMD Instinct MI100 (`gfx908`) with:

- FP16
- BF16
- FP32
- small tensors
- multi-batch odd spatial dimensions
- aligned channel spans
- the odd `32769` channel-span regression case

All 12 dtype/shape combinations passed with zero tolerance violations.

`git diff --check` also passes.


yzhou103:
Follow-up idea: keep the vectorized path for odd numel_per_channel

  The fix is correct, but it buys correctness by disabling vectorization for every shape whose per-channel element count isn't a multiple of 4 — including the 2D (N, C)
  case (numel_per_channel == 1), which now always falls back to the scalar path. It might be worth recovering that in a follow-up.

  The cheap way is to keep the gate at the original inner_size % 4 == 0 (which is required anyway — the loop steps by 4 and would write past the tail otherwise) and
  instead handle the straddling vector inside the kernel. The down kernel already does an idx / numel_per_channel division on every iteration, so the remainder comes
  essentially for free:

  for (uint32_t i = tid * 4; i < inner_size; i += (gridDim.x * THREADS_PER_BLOCK) * 4)
  {
      uint32_t idx = blockIdx.y * inner_size + i;
      uint32_t c   = idx / numel_per_channel;
      uint32_t off = idx - c * numel_per_channel;   // reuses the same divide
      uint32_t channel_idx = c % num_channels;

      tmp.load(x + idx);                            // load/store stay vectorized either way

      if (off + 4 <= numel_per_channel) {
          // fast path: whole vector lives in one channel, fetch w/b once
      } else {
          // slow path: resolve channel_idx per element
          #pragma unroll
          for (int k = 0; k < 4; ++k) { /* (idx + k) / numel_per_channel % num_channels */ }
      }

      vec_y.store(y + idx);
  }

  Notes:

  - The slow path fires at most once per channel boundary, i.e. roughly 4 / numel_per_channel of all iterations, so it's negligible for real spatial shapes and stays
  warp-coherent.
  - numel_per_channel == 1 takes the slow path every time, but loads and stores are still 128-bit — strictly better than today's scalar fallback. A later refinement
  could also load weights/bias as Vec<T,4> in that case, since channel_idx == idx is naturally 4-aligned there.
  - The up kernel needs no change at all (it never touches weights/bias), so it keeps vectorizing under the inner_size % 4 == 0 gate. Note that splitting the flag
  between up and down isn't really an option: both kernels share gridDim.x, and mean_acc[blockIdx.y * gridDim.x + i] binds the workspace layout to it.

  Unrelated, but while this code is being touched: uint32_t idx = blockIdx.y * inner_size + i; is 32-bit while inner_size derives from an int64_t, so it silently wraps
  past 2^32 elements. Probably worth promoting idx to int64_t/size_t in the same pass.

  Not blocking this PR — the correctness fix should land as-is

UD-mmcminn:
@yzhou103 
Agreed, I think that's a sound approach but it's a bigger change and I agree it should be a separate PR to keep things manageable. I can do some performance testing on the gfx908 to see if the added complexity in the kernel is worth the squeeze. But I can't validate on anything newer. If someone at AMD wants to send me a MI350X box :).

PR #5750, opened 2026-09-22
[Triton/Gluon] [JIT] [Feature] Add tuned native group32 A8W8 Triton GEMM for gfx950

valarLip:
DeepSeek-V4.1-Flash projections use unshuffled E4M3 weights and compact E8M0 group32 scales. This adds a gfx950 Triton backend for those operands to `gemm_a8w8_blockscale`, together with tuned configurations for TP4, TP2 and attention-DP local GEMMs.

### Implementation and dispatch

- Keep activations, weights and scales in their native formats. Support compact 32x32 or per-row 1x32 weight scales, FP32 accumulation, and BF16/FP16/FP32 output.
- Use native microscaling MFMA, packed-K execution for small M, and the existing non-atomic split-K reduction. The public interface accepts an optional native-group32 `split_k` partition count, preserving explicit caller controls while keeping configured backend selection first.
- Query `get_CKGEMM_config` before selecting a fallback. Native E8M0 operands have a separate `AITER_CONFIG_GEMM_A8W8_BLOCKSCALE_GROUP32` CSV family, so an identical M/N/K in the FP32 128x128 table cannot select an incompatible kernel. Unsupported configured backends raise instead of silently falling back.
- Add 450 DSV4.1 model-config rows: 18 geometries and 25 M buckets, all selecting `libtype=triton`. Tile, split-K and packed-variant settings remain in the existing Triton JSON configuration system. Of these rows, 262 reuse exact-shape measured timings; remaining timing fields are empty.
- Extend the shared GEMM loader with optional file-defined `M_BOUNDS`, retaining explicit-argument precedence and legacy bounds. Config copies preserve nested mutation isolation while avoiding per-scalar `deepcopy` overhead.

The new native backend requires gfx950. Existing FP32 128x128 operands retain their CK/CKTile paths. Group32 currently supports the Triton backend; configuring legacy CK/CKTile for its E8M0 operands is rejected.

### Performance

Hardware: AMD gfx950, 256 CUs; PyTorch 2.9.1 ROCm 7.1.1 and Triton 3.8.0 ROCm build. Measurements use BF16 output, CUDA graphs and at least 512 MiB of rotating weights. The baseline is the initial imported native group32 configuration in the first commit of this PR.

The sweep evaluated 49,304 candidates across 234 representative M/N/K cases and validated 603 cases, including interval boundaries. The geometric-mean speedup is **1.179x** (15.18% lower GEMM latency). Of 300 intervals checked, 252 use improved configurations and 48 retain the baseline. The largest sampled slowdown among changed configurations is 0.97%. Measurements affected by other GPU users were discarded and rerun on available cards.

| Projection | N | K | Geometric-mean speedup |
|---|---:|---:|---:|
| qkv_a | 1792 | 5120 | 1.366x |
| indexer_q | 4096 | 1280 | 1.211x |
| q_b TP4 | 8192 | 1280 | 1.061x |
| q_b TP2 | 16384 | 1280 | 1.063x |
| q_b attention-DP | 32768 | 1280 | 1.112x |
| o_b TP4 | 5120 | 2048 | 1.188x |
| o_b TP2 | 5120 | 4096 | 1.221x |
| o_b attention-DP | 5120 | 8192 | 1.121x |
| shared_up TP4 | 1152 | 5120 | 1.131x |
| shared_up TP2 | 2304 | 5120 | 1.217x |
| shared_up TP1 | 4608 | 5120 | 1.194x |
| shared_down TP4 | 5120 | 576 | 1.273x |
| shared_down TP2 | 5120 | 1152 | 1.268x |
| shared_down TP1 | 5120 | 2304 | 1.243x |
| Engram | 25600 | 6144 | 1.123x |
| draft_main | 5120 | 15360 | 1.180x |
| draft_kv_fused | 1536 | 5120 | 1.176x |
| draft_kv | 512 | 5120 | 1.146x |

These are standalone GEMM results. End-to-end serving throughput was not measured. Tuning covers M through 4096; larger M retains the original JSON fallback. Model CSV entries through M=16384 choose the backend without claiming newly tuned tiles at those larger sizes.

### Validation

- **136 AITER tests passed** on the rebased branch: native numerical contracts, column/K tails, extreme scale codes, per-row scales, split-K, output buffers, live-input graph replay, dynamic-M fullgraph compile, CSV format isolation, and configured-backend precedence. CK/CKTile dispatch-priority tests use stubs; native numerical and graph tests execute GPU kernels.
- **100 ATOM integration tests passed** with the companion migration to the public AITER interface, covering actual backend-config lookup, blockscale projections, accumulation, GPU quantization, graph replay and dynamic batched fullgraph compilation.
- All **294,912** combinations of the 18 model geometries and integer M=1..16384 resolve through the real shared loader to Triton; unknown geometries still miss the table and reach fallback.
- Six representative cases comparing dispatch before/after the configuration-priority change have bitwise-equal BF16/FP32 outputs, identical GPU kernel sequences, and graph-latency differences below 0.5%.
- Black and Ruff pass for all seven changed Python files; `git diff --check` passes.

The sweep compares sampled outputs against an independent FP64 reference and complete BF16 outputs against the baseline. BF16 outputs are bitwise equal in 384/603 cases; the maximum BF16 relative RMS difference is 1.052e-4. One pre-existing FP32 reference-boundary outlier at M=97, N=5120, K=1152 remains: maximum absolute error changes from 0.09451294 to 0.09445190. The experiment accepts that point only when candidate error does not exceed baseline error; repository test tolerances are unchanged.

```bash
# AITER, on gfx950
python -m pytest -q \
  op_tests/test_gemm_group32_interface.py \
  op_tests/test_opus_a8w8_interface.py \
  op_tests/triton_tests/gemm/basic/test_gemm_a8w8_blockscale_group32.py

# ATOM integration, with the V4.1 reference weights available
ATOM_DSV41_REFERENCE=/mnt/DeepSeek-V4.1-Flash python -m pytest -q \
  tests/model_ops/test_blockscale_packed.py \
  tests/models/deepseek_v41/test_accumulation.py \
  tests/models/deepseek_v41/test_quant_gpu.py
```


Dewei-Wang-sh (approved):
lgtm

Fangzhou-Ai:
Evaluated this downstream in vLLM for DeepSeek-V4.1-Flash on gfx950 (MI355X). The kernel looks strong — for shapes in the tuned table it beats the `gemm_a8w8_blockscale_bpreshuffle` ASM path by roughly 2–2.5x across decode-sized M, and holds parity out to M=2048.

Our attention shapes are all covered by your tuning (`N=1792-K=5120` for the fused WQA/WKV projection, `N=5120-K=8192` / `N=5120-K=2048` for the output projection), so this is **not** a regression report. But while measuring I found the fallback for *uncovered* shapes is a sharp cliff, which seems worth raising since the table can't cover everything.

To isolate tuning presence as the only variable, I compared a tuned shape against a deliberately chosen untuned near-neighbor. `N=1792, K=5120` is tuned; `N=1856, K=5120` is not, and has only 3.6% more work:

| M | N=1792 (tuned) | N=1856 (untuned) | untuned / tuned |
|---|---|---|---|
| 96 | 16.1 µs | 22.6 µs | 1.40x |
| 384 | 22.5 µs | 42.5 µs | 1.89x |
| 512 | 22.4 µs | 50.4 µs | 2.25x |
| 1024 | 29.7 µs | 88.2 µs | 2.97x |
| 2048 | 42.6 µs | 160.8 µs | 3.77x |

The same pattern holds where the untuned shape does *less* work — `N=5120, K=1152` (tuned) vs `N=5120, K=1024` (untuned, 11% fewer FLOPs):

| M | K=1152 (tuned) | K=1024 (untuned) |
|---|---|---|
| 512 | 20.7 µs | 30.2 µs |
| 1024 | 22.9 µs | 51.1 µs |
| 2048 | 30.1 µs | 92.2 µs |

Both untuned shapes are synthetic probes, not shapes we run. The point is the magnitude: at those sizes the ASM bpreshuffle path is 1.9–3.6x faster than group32, so a shape that misses the table lands slower than the path it replaces, while a covered shape wins by ~2x.

The cause looks like `DEFAULT.json` being a single decode-shaped tile applied at every M:

```json
"any": {
  "BLOCK_SIZE_M": 16, "BLOCK_SIZE_N": 32, "BLOCK_SIZE_K": 256,
  "num_warps": 2, "NUM_KSPLIT": 1, ...
}
```

`BLOCK_SIZE_M=16` with `num_warps=2` is sensible for M≤128 but leaves a lot unused at M≥512, which is where the curves diverge.

Two suggestions:

1. Make `DEFAULT.json` M-bucketed rather than a single `"any"` entry, with a larger-M tile above roughly M=256, so an uncovered shape degrades gracefully instead of 3–4x.
2. Consider raising the "not found tuned config, will use default config" message above info level. On a large model it scrolls past easily, and it's currently the only signal that a shape is on the slow path.

Method: `triton.testing.do_bench(warmup=25, rep=100)`, bf16 output, all scales set to 1.0 (E8M0 exponent 127), native (unshuffled) weights for group32 and `shuffle_weight(layout=(16,16))` plus fp32 group-128 scales for the ASM arm. Happy to share the microbenchmark or cover additional shapes.

(Measurements gathered with AI assistance; numbers are from real runs on MI355X.)

PR #5759, opened 2026-09-22
[Bugfix][MLA] Size PS prefill partial buffers from the real token budget

### LINKED ISSUE #5729: [MLA] get_ps_metadata_info_v1 should return the partial buffer size, and accept the real token budget
`get_ps_metadata_info_v1` returns the length of `reduce_partial_map` but not the size of the partial `logits` / `attn_lse` buffers that the map indexes into. Callers have to reconstruct that themselves from an undocumented layout rule, and the only number they have to reconstruct it from is a worst case that is far looser than any real workload. Both together made a vLLM allocation grow from ~2 GiB to ~128 GiB after #3606, which is a startup OOM on Kimi K2.5/DeepSeek-V3 with FP8 KV.

Asking for two changes so integrators can use the returned values directly:
1. have get_ps_metadata_info_v1 return the partial row count as a seventh entry so callers do (max_partial_rows, num_head_k, v_head_dim) and never touch the stride when allocating logit/attn_lse.
2. let the bound reflect the real token budget (sum_i qlen_i <= total_qlen, where total_qlen is e.g. max num batched tokens in vllm). We can do this by adding an optional total_qlen argument to get_ps_metadata_info_v1 and if it's provided we can do `qo_tile_cnt = min(qo_tile_cnt, cdiv(total_qlen, qlen_granularity) + batch_size - 1)`

Related: https://github.com/vllm-project/vllm/pull/57923

simondanielsson:
## Motivation

Fixes https://github.com/ROCm/aiter/issues/5729.

The MLA PS metadata scheduler returns a pessimistic bound on the max number of partial entries needed to be allocated. This could lead to OOM. The pessimism stems from the fact that it assumed that all requests used the same (maximal) number of query tokens which is not true in inference engines with a fixed query token budget per step.

This PR fixes it to correctly give a much tighter bound by accounting for fixed token budget (c.f. max num batched tokens in vllm) if passed.

## Technical Details

A temporary fix was required on serving engine side: https://github.com/vllm-project/vllm/pull/57923/changes

It's better if we do this in AITER.

## Test Plan

1. Tests passes locally
2. e2e test with dsv3/kimi k2.5 gsm8k in vllm using aiter built in this branch

## Test Result

Tests passes locally

## Submission Checklist


simondanielsson:
Added regression tests for this

PR #5761, opened 2026-09-22
[HIP] [OPUS] [JIT] Unify gfx950+gfx1250 MXFP4 MQA-logits

shay-li77:
## Motivation

gfx950 (#5332) and gfx1250 (#5656) shipped the same operator as two parallel
implementations: two JIT modules, two schedule builders, two Python surfaces and
two op tests. Every fix had to be landed twice, and two of them already diverged
in exactly the same shape of hole (a compile-time stride behind a rank-only
check). This folds them into one module / schedule / API / test, leaving both
device bodies untouched.

## Technical Details

**One JIT module.** `module_pa_mqa_logits_mxfp4_opus` carries both device bodies;
a version macro selects which compiles (`__gfx950__` MFMA 32x32x64 wave64 vs
`__gfx1250__` WMMA 32x16x128 wave32/TDM), and the fat binary is built with
`--offload-arch` per arch, the idiom `opus_gemm` already uses. The two
`_gfx1250_` C++ files are deleted, their contents moved verbatim.

**One schedule.** gfx1250's two-phase `build_tiles` + `build_sched` is now the
only builder; gfx950's per-row builder is retired. gfx950 runs at
`q_per_block == 1`, where a tile is one row and the tile-union window is that
row's, so the shared builder emits exactly the per-row record its MFMA kernel
already reads.

**The tile cut is skipped at `q_per_block == 1`.** There the cut is the identity
-- tile `t` is row `t` -- so the boundary array is the arange the host already
knows. `build_tiles` exists to serve `q_per_block > 1` and is no longer launched
at 1: **one fewer kernel per forward** on every gfx950 call and on gfx1250's
MTP=1 decode instance, and `plan_buffers` stops allocating a buffer nothing
reads. `build_sched` takes `q_per_block` and reads the mapping through
`tile_row()` with a NULL `cu_tiles`.

**One runtime-arch dispatch.** `pa_mqa_logits_mxfp4_fwd_sched` reads
`get_gpu_arch()` and calls the MFMA or WMMA launcher, then dispatches within it
by `(q_per_block, block_k)`. aiter JITs per target GPU, so the arm that launches
always matches the compiled body.

**One Python surface** `aiter/ops/opus/pa_mqa_logits_mxfp4.py`, arch-parameterized
by `MqaLogitsVariant(name, q_per_block, block_k, cta_resident)`. **One op test**
on both arches (`SUPPORTED_GFX = ["gfx1250", "gfx950"]`), building natural
layouts on gfx1250 and MFMA-permuted ones on gfx950.

**`-mllvm -enable-post-misched=1` is load-bearing for gfx950 and must not be
dropped.** aiter's base JIT flags set it to 0; this module turns it back on,
which is the state the tuned standalone uses. The merged module now applies it
to the gfx1250 body as well (its own module did not) -- measured below.

## Test Plan

- Unified op test on both arches: corner cases on every compiled instance, NaN
  E8M0 scale propagation, and the four perf tables.
- Schedule table compared **byte for byte** against the pre-change builder, since
  the kernel consumes only that table.
- Kernel ISA compared against the standalone reference (`tools/fp4_aiter_isa.sh`
  + `tools/cmp_isa_aiter_vs_standalone.py`), to show the device bodies did not
  move.
- `-mllvm -enable-post-misched=1` A/B on gfx1250, since the merged module applies
  it to a body that never had it.

### gfx1250 (ROCm 10.1, clang 23)

| check | result |
|---|---|
| op test, both instances | 28/28 corner cases, NaN E8M0 6/6, `err = 0` on all 50 perf rows |
| schedule table vs built cut | **byte-identical** over 4 reps x 11 shapes (decode 1-2048 seqs, prefill 16384/25000 rows, with `local_starts` and `row_to_batch`) |
| kernel resources | 269 VGPR / occupancy 3 / no spill / LDS 9728 on both instances, unchanged |
| plan-stage launches | `qlen1_kv64` 2 -> 1 (small) and 3 -> 2 (split); `qlen4_kv64` unchanged |
| plan-stage cost | 8.7-9.1 us saved per forward on MTP=1 decode, 12-20 us on prefill |
| kernel time | 6 runs vs 16 pre-change, geomean 0.9982, inside the measured noise |

**post-misched on gfx1250.** The flag is active (`=1` is appended after the base
`=0`, last-wins) and it does change the ISA -- but it is a pure post-RA
reschedule: resources are identical to the digit, and the reordered code has
*fewer* instructions (VOPD packing, fewer `s_delay_alu`/`v_nop`). 16 paired
alternating runs, binary verified per run: **no shape regresses beyond noise, and
7 prefill shapes are 1.0-2.9% faster with the flag on** (13-16/16 sign
consistency, 0.4-3.6% within-arm spread). Decode shows no direction.

### gfx950

gfx950 (#5332) now runs through the unified module / schedule / API. Every call is at
`q_per_block == 1`, where the tile cut is the identity, so the MFMA body is recompiled, not
rewritten. gfx950 runs the `qlen1_kv64` (`block_k = 64`) instance in production; `qlen1_kv256`
is compiled and correctness-checked but not on the production path. Validated on a gfx950 box
(ROCm 7.2.4), module rebuilt fresh.

**Correctness** — op test, both instances (`qlen1_kv64`, `qlen1_kv256`):

| table | rows | err |
|---|---|---|
| corner (13 cases × 2 instances) | 26 | 0 |
| NaN E8M0 propagation | 6 | 0 |

**Kernel and schedule vs. standalone #5332:**

| check | result |
|---|---|
| kernel ISA, both Table kernels (`block_k` 64 / 256) | byte-identical (full text, 223 VGPR / 0 spill, counts, `s_waitcnt`) — `cta_resident`-independent |
| schedule table, identity cut vs. general built cut at equal `cta_resident` | byte-identical, all fields |
| `build_tiles` at `q_per_block == 1` | skipped (identity); `cu_tiles` no longer allocated |
| host row bound `total_q > q.shape[0]` | raises on gfx950 |

The production `qlen1_kv64` keeps #5332's `SCHED_CTA_TARGET = 1024`, so its schedule table is
byte-identical to #5332 across **every** #5332 scenario, verified on hardware:

| scenario | schedule vs. #5332 |
|---|---|
| prefill fresh / causal / chunked | identical (16384 tiles ⇒ one CTA per tile; grid is `cta_resident`-independent) |
| decode (seqs 1 / 8 / 32 / 128) | identical (`cta_resident` 1024 = #5332's, same CTA split) |

(`cta_resident` is a `build_sched` input — it shapes the schedule, not the kernel ISA. `qlen1_kv256`
carries an untuned `512`, but it is off the gfx950 production path, so it does not affect any
number here.)

**Performance** — same-card A/B on `qlen1_kv64` at #5332's chunked shapes (25000 cols, bs 1/2/4),
`run_perftest` 50/10, min-of-25: unified matches standalone #5332 within noise (≤1%, never slower).
This is the expected result and it generalizes to every scenario above: the ISA is byte-identical
and the schedule table is byte-identical, so the timed kernel is the same machine code on the same
table — the unification only drops one `build_tiles` launch per forward.

## Submission Checklist


valarLip (changes requested):
Reviewed at head of `lxx/dev/pa_mqa_logits_mxfp4_gfx950_unify` against merge-base `25ae2bfa3e`, in isolated worktrees.

The motivation for unifying is well argued, and the surface this PR exposes is clean — I checked that the three defs in `rocm_ops.hpp` match the header declarations and the Python `@compile_ops` stubs positionally, and that no stale reference to the deleted module, the deleted pybind macro, or the removed Python names survives anywhere in the tree.

My concerns are that the unification stops at one *file* rather than one *module*, that it adds host-side cost to a path that runs 61 times per forward, and that several guards only survived the merge in one direction. Taking those in that order.

## Structure

The output is a 2061-line header — 2.5x the repo's 800-line soft ceiling — whose two arch bodies share no code. Measured layout: shared ABI and schedule at L1-655 (~30%), gfx950 traits L657-791, gfx1250 traits L792-1109, the gfx950 wave64 body L1111-1656, the gfx1250 wave32/TDM body L1658-2059. So 68% is arch-specific, sitting in disjoint `#if` islands that each re-declare their own `pin_sgpr` (:1156 vs :1692) and `permlane_head_reduce` (:1164 vs :1706). The file now also carries three spellings of a workgroup barrier: `qshare::cta_sync`, `sched_detail::sync_block`, and opus.hpp's own.

In `kernels.cu`, `check_shapes_mfma` (L37-121) and `check_shapes_wmma` (L190-296) are ~70 near-identical lines differing only in the trait constant *names* for the same quantity, and `launch_sched_mfma` / `launch_sched_wmma` repeat the same eleven kargs assignments and the same launch tail. The `block_tables` width gap below is the drift this duplication has already produced.

Some of what the merge carried over is dead: `opus_logits::sched_slots` and `max_tiles_for` have no C++ caller (Python re-implements both), `ceil_div_i` has no caller, `logits_fp4_qshare_traits_4q` is unused, and four kargs fields (`split_kv`, `num_batches`, `ptr_cu_seq_q`, `ptr_row_to_batch`) are pinned at fixed offsets by a 22-line `offsetof` block without being read.

The naming also inverts: the module-generic names — `opus_mqa_logits_fp4_traits`, `pa_mqa_logits_mxfp4_kernel` — are the gfx950-only ones, while the gfx1250 twins carry the qualifier.

A three-file split (`_opus.h` for ABI and schedule, `_mfma_body.h`, `_wmma_body.h`) plus one templated `check_shapes` and `launch_sched` keeps every benefit of the merge and narrows the JIT rebuild blast radius — as it stands, touching the gfx950 body forces a rebuild of the 400 lines of hand-scheduled gfx1250 TDM code.

## Performance

`const std::string arch = get_gpu_arch();` now runs on **every** `fwd_sched` call. `get_gpu_arch` (aiter_hip_common.h:537) does no caching: a `hipGetDeviceCount`, a `hipGetDevice`, a `hipGetDeviceProperties`, then a `std::string` construct plus `find` plus `substr`. The outer `HipDeviceGuard` also duplicates one the inner launchers already take, and `HipDeviceGuard` (:643-648) has no `prev == target` fast path. Per-layer host cost goes from three HIP device-management calls to six, plus a full `hipDeviceProp_t` fill, which takes a runtime lock.

Neither base TU called `get_gpu_arch` at all — the probe is new with the unification, and both arches pay it. The module's own docstring budgets this op at 61 launches per DeepSeek-V4 forward against ~16 us of host dispatch, so I think this more than cancels the one `build_tiles` launch the PR advertises saving. It's also invisible under cudagraph capture, so a graphed A/B won't surface it. The repo idiom is one directory over — `fmha_fwd_bf16_opus_launch.h:280` does `static const std::string arch_id = get_gpu_arch();` — though here it wants a per-device cache, since the guard keys on `q.device_id`.

The gfx950 grid policy changes twice, with no measurement behind either. `_sched_slots` now rounds the grid up to a multiple of `cta_resident`: base was `max(num_rows, 1024)` with no round-up, HEAD is `ceil(max(n,r)/r)*r`. At `total_q=1025` that launches 2048 CTAs instead of 1025 — 1023 surplus CTAs, each dispatching and issuing a 32-byte `s_load_dwordx8`, 61 times per forward, with `cta_info` allocated for the padded count too. Separately `qlen1_kv256` ships `cta_resident=512`, halving base's `SCHED_CTA_TARGET=1024` split aim, so `settle_and_emit` gives each CTA roughly twice the chunks on the long-window prefill regime that instance exists to serve.

Both values are self-labelled "UNTUNED PLACEHOLDERS" at py:158-161, while the class docstring two screens up (:99-100) says `cta_resident` is "TUNED per instance". And because `decode_shapes` names only `qlen1_kv64`/None and `_resolve_variant` falls back to the arch default, `qlen1_kv256` appears in no perf table on either arch — a large load-balance regression wouldn't show up in any number this PR reports.

One compile-side item: folding the gfx1250 sources into `module_pa_mqa_logits_mxfp4_opus` silently applies `-mllvm -enable-post-misched=1` to the hand-scheduled WMMA body. The deleted `module_pa_mqa_logits_mxfp4_gfx1250_opus` entry carried exactly `['-ffast-math', '-fno-finite-math-only']`, and `-mllvm` flags are per-TU. That body is explicitly hand-scheduled — `__builtin_amdgcn_sched_barrier(0)` around every store (:1981-1983, :2036), the two-accumulator ping-pong (:1831-1836), a TDM pipeline sized by `TENSORCNT_KEEP` (:1072), `pin_sgpr` inline asm. Post-RA rescheduling can reorder the TDM issue/wait pairing or raise pressure past the occupancy the hardcoded `cta_resident=768/3072` were swept for. The PR does report a gfx1250 A/B for the flag, but not gfx1250 VGPR/occupancy, which is what `cta_resident` depends on. The same JSON already gates this exact flag on `get_gfx()=='gfx1250'` around line 354, so a per-arch expression is available.

## Cleanliness

`general_emit`'s new even split does `const int nc = (n + safe - 1) / safe;` then `const int q = n / nc, r = n % nc;` — a device-side divide by zero whenever `n == 0` — and that `nc` shadows the outer `nc` at :416 with a different value. The inline comment asserts ">= 1: a zero-chunk tile emits no slot", but that property isn't local: it holds only because the `upper_bound(s_excl, j) - 1` search at :431-436 resolves a run of equal prefix values to the *last* index, so a zero-`nc` tile can only be selected if it's the block's trailing tile, which `j < block_total` already excludes. Switch that search to `lower_bound`, hoist it, or make the scan inclusive, and the assumption breaks — and integer divide by zero on AMD is an undefined result rather than a fault, so the table silently gets garbage `chunk_start`/`chunk_count`. A `n > 0 ? ... : 1` guard and a rename for the shadowing local would close it.

The cross-arch buffer guard `if buffers.variant not in _variants_for(arch)` rests entirely on frozen-dataclass value equality, and the only field separating gfx950's `qlen1_kv64` from gfx1250's is `cta_resident` — 1024 vs 3072. `MqaLogitsVariant` (:89) has no arch field, and the comment at :158-161 asks someone to measure the gfx950 numbers. The moment that tuning lands on 3072 the two records compare equal, and both this guard and `_as_variant` (:494-500) stop rejecting cross-arch buffers; the only backstop left is `_check_layout`'s ndim sniff, which the module itself documents cannot catch an array reshaped to the right ndim. A pure perf edit would silently disable a correctness guard. An `arch` field, or keying the guard on arch, avoids that.

`tile_kv_tiles_cached` reads `cta_info[t].chunk_count` back through a `const ... * __restrict__` pointer that `emit_fast` wrote through a *different* `__restrict__` pointer in the same kernel. Both helpers inline into `mqa_logits_build_sched` (:545), LLVM emits `!noalias`/`!alias.scope` per inlined call site, and the two restrict-qualified parameters land in disjoint scopes — formally telling the optimizer the load in `settle_safe` cannot see the stores in `emit_fast`. Only the `s_barrier` inside `block_reduce_stats` separates them. Base's `settle_safe` recomputed the count from `local_starts`/`local_ends` and never read the table back. If the load is hoisted or forwarded, `ctas_for()` sums garbage and `general_emit` emits a schedule that either overruns `num_ctas` (tiles dropped, their rows keeping the caller's `-inf`) or under-splits. Dropping `__restrict__` on the read-back overload, or carrying the counts through LDS, removes the hazard.

`pa_mqa_logits_mxfp4_block_table_width` now calls `_device_arch(torch.cuda.current_device() ...)` unconditionally — including when the caller passed explicit `MqaLogitsVariant` objects that need no lookup. Base's docstring said verbatim "Pure arithmetic -- no GPU, no JIT build -- so it is safe to call at startup"; HEAD downgrades that to "Past that arch probe this does no GPU work". A serving stack sizing `block_tables` at config time, on a CPU-only rank, under `CUDA_VISIBLE_DEVICES=''`, or inside a fork-based dataloader worker, now either raises or initializes a CUDA context in a process that was deliberately keeping it lazy. The arch is only needed on the variant-is-None and by-name branches.

`_is_permuted()` (:198) and `_variants()` (:400) are `@functools.cache` on zero-argument functions that probe the *current* device, so the first device touched permanently fixes the scale layout and the variant list for the process.

Two places where the gfx950 arm still drifts from the shared ABI the unification is meant to establish. Its kernel template defaults `SCHED = mqa_logits_sched::Prefill` (:1114 and :1299), an arm whose kargs fields the MFMA launcher never assigns, while gfx1250 defaults to `Table` and static_asserts it (:1716). `kernels.cu:176` passes `Table` explicitly so nothing is broken today, but any new call site omitting the argument instantiates the Prefill arm and reads `ptr_row_to_batch` / `ptr_local_starts` / `num_rows`, all left zero-initialized by `launch_sched_mfma` (:154-170) — a NULL deref, not an error. Same shape of trap for the kargs invariant at header:117 ("ptr_local_ends may not [be NULL], being the store's upper bound"): it's false on the gfx950 arm, so the next gfx950 edit that adopts gfx1250's per-row store mask faults with the header telling the author it can't. Defaulting `SCHED` to `Table` (or dropping the parameter) and either assigning the pointers or scoping the invariant per arm fixes both.

`build_sched` validates `q_per_block >= 1` and `block_k >= 1` but dropped base gfx950's `AITER_CHECK(cta_target >= 1)`, and the Python mirror isn't equivalent: C++ `sched_target`/`settle_and_emit` treat `resident <= 0` as "do not split" and accept it silently, while `_sched_slots` (py:299) computes `-(-n // r) * r`, which raises `ZeroDivisionError` at `r == 0` and returns a *negative* grid at `r < 0` that `_cta_info` then hands to `torch.empty` as a first dimension.

## Correctness

Three device-array length guards did not survive the merge. Base had `AITER_CHECK(local_ends.numel() >= num_rows)` plus the same for `local_starts` and `row_to_batch` (base kernels.cu:190, 217, 225); HEAD has none of the three, only dtype and contiguity. At the identity cut (`q_per_block == 1`, i.e. every gfx950 launch) `num_tiles` *is* the row count, and `emit_fast` reads `local_ends[t]` and `row_to_batch[r0]` for every `t < num_tiles`. So `pa_mqa_logits_mxfp4_plan(cu_seq_q, local_ends=<1024 entries>, total_q=4096)` gives `num_tiles=4096` and reads `local_ends[1024..4095]` past the allocation; a short `row_to_batch` gives an arbitrary `batch_id`, hence the wrong `block_tables` row and garbage page ids. The compensating Python check at pa_mqa_logits_mxfp4.py:762-773 compares `num_tiles` against `q.shape[0]`, never against `local_ends.numel()`, and it runs at launch — after the builder has already executed.

The gfx950 arm has neither of the two bounds its gfx1250 twin gained. `check_shapes_wmma` computes `bt_cols = ceil(max_seq_len/KV_TILE_SIZE) * PAGES_PER_TILE` and raises below it (:249-254); `check_shapes_mfma` has only `AITER_CHECK(block_tables.dim() == 2, ...)` at :51 and then takes `block_tables.size(1)` as an unvalidated row stride at :164, while the kernel indexes `g_bt` at `(chunk_start + tile_idx) * pages_per_tile` (header:1462) with nothing else bounding the column. This PR is what makes that reachable, since it's the first to expose the gfx950 `qlen1_kv256` instance (4 pages per tile) through the shared variant API: a caller sizing `block_tables` by `ceil(max_seq_len/64)` page rounding and running `qlen1_kv256` reads up to three int32 past each row. Three Python docstrings now assert the opposite for both arches — :58 ("The C++ header states the resulting bound", where the paragraph that stated it was deleted with the gfx1250 header), :520 ("the C++ launcher raises below it"), and :481 ("They all take ... the same `block_tables` width"), which gfx950's 64-vs-256 pair contradicts 40 lines later at :522.

Likewise the Table arm's `if(rec.chunk_count <= 0) return;` has no `num_rows` bound where gfx1250 added `|| rec.row_id >= num_rows`, and `kargs.num_rows` is write-only on the MFMA arm (set at kernels.cu:164, never read). `pa_mqa_logits_mxfp4_fwd_sched` is exported through `PA_MQA_LOGITS_MXFP4_PYBIND` and reachable via `_fwd_sched_raw`, so a schedule built for `total_q=4096` launched against `q.shape[0]==1024` gives `row_id` up to 4095 indexing `q_base = p_q + row_id*Q_ROW_BYTES` (4 KB/row) and `out + row_id*stride_out` — reads and writes ~12 MB past both allocations. That bound belongs in the kernel next to the gfx1250 one rather than in one Python wrapper.

Finally, gfx1250's pre-existing `qlen1_kv64` instance quietly stops honouring `cu_seq_q`. Base ran `_gfx1250_compute_tiles` for both gfx1250 variants, and `mqa_logits_build_tiles` writes `cu_tiles[t] = cu_seq_q[batch]` for every `t` past the real tile count, so padding tiles got `r0 == r1` → `tile_union_end() == 0` → `chunk_count = 0` → a surplus slot. HEAD takes `num_tiles = _max_tiles_for(n, batch, v)` = `total_q` and gives tile `t` the real row `t`. A cudagraph caller that pins `q`/`local_ends`/`out` at capacity 256 but has `cu_seq_q=[0,...,100]` this forward: rows 100..255 carry stale non-zero `local_ends` from a previous replay, HEAD schedules them, reads `q[100..255]` and `block_tables[row_to_batch[r]]` (row 255 against a 100-row table), and writes into `out[100..255]` that base left at `-inf`. The new docstring documents the contract change ("the row domain is `total_q`, not `cu_seq_q[batch]`"), but nothing checks it and the PR body doesn't mention the gfx1250 behaviour change.

## Tests and exports

The deleted `op_tests/test_pa_mqa_logits_opus.py` cross-checked gfx950 against an independent `flydsl_prefill`/`flydsl_decode` implementation, and nothing replaces it — while the surviving file's docstring asserts "There is no second implementation to cross-check against on this target", a gfx1250-only claim now stated in a file that runs on gfx950, where `aiter/ops/flydsl/kernels/mqa_logits/pa_mqa_logits_fp4*.py` still exist. More importantly, the HEAD reference and the kernel now consume the same `_scale_to_opus` permutation (:206), so a coordinated MFMA scale-permutation error — the exact silent failure the module docstring says "is caught HERE and nowhere else" — passes.

`check_nan_scale('prefill', 1, 4, [0,1,2,3], kv_row=2)` is labelled "the control that says the NaN is not smeared", but row 3's window `[0,3)` contains column 2, so `want.sum() == 1`, not 0, and the `le == 0` row asserts nothing; its precondition is a bare `assert` that `-O` strips. And `check_row_id_bound` is skipped entirely on gfx950 (:686), so the PR's new gfx950-only host row bound (py:762-773) is untested.

`aiter/ops/opus/__init__.py` drops `pa_mqa_logits_mxfp4_sched` / `_build_sched` / `_sched_slots` / `_sched_buffer_ints` from both imports and `__all__` and adds nothing back, so the op has no package-level or top-level `aiter.*` export at all. In-tree grep is clean and ATOM already uses the submodule path, but an out-of-tree gfx950 caller breaks with an error pointing at no successor, and the retired `cta_target` / `block_k` knobs have no supported replacement.

## Summary

I'd like to see the header split before this lands — one file with 68% arch-specific content in disjoint `#if` islands and two 70-line copies of the same shape check is the structure the merge was supposed to remove, not produce. On performance, the uncached `get_gpu_arch` on a 61-launches-per-forward path likely costs more than the `build_tiles` launch it saves, and the two gfx950 grid changes should carry a measurement, ideally on an instance that some perf table actually exercises. And the four guards that only survived in one direction — the three length checks, the `block_tables` width bound, the `row_id` bound, and gfx1250's `cu_seq_q` contract — are the ones I'd treat as blocking, since each one turns a host-side raise into an unbounded device read.

PR #5773, opened 2026-09-22
[SMI] Add opt-in summary plots and raw-sample timelines

arakowsk-amd:
SMI benchmark results currently require manual processing to compare cases, and aggregate logs discard the timing of individual observations. This adds opt-in automatic summary reports and lossless raw-sample capture with a standalone per-case timeline renderer.

- `AITER_SMI_PLOT=1` captures summary records and renders a report at normal process exit. `AITER_SMI_PLOT_DIR` defaults to `./smi_plots`; each process gets a separate run directory. Existing stdout and `AITER_SMI_OUTPUT_PATH` output are preserved.
- `AITER_SMI_TRACE=1` independently saves every collected sample and its timestamp, plus window boundaries and benchmark metadata. `AITER_SMI_TRACE_DIR` defaults to `./smi_traces`. Raw records are written after each monitoring window, outside the polling loop.
- `aiter/smi_plot.py` renders existing logs or directories into all-test, per-test, and per-case summary views. `aiter/smi_trace_plot.py` explicitly renders raw traces into per-case timelines, comparing variants in separate columns with clocks, power, activity, temperature, and VRAM panels.
- Reports include PNG/SVG plots, an HTML index, CSV, and a manifest. Timeline plots preserve every observation at its actual timestamp without smoothing, resampling, or downsampling; missing and firmware `N/A` readings remain gaps. Aggregate-only logs cannot reconstruct timelines.

Matplotlib is optional and the renderers work without a GPU. Automatic summary rendering runs in a CPU-only subprocess after the benchmark. Detailed timeline rendering is an explicit CLI step. Identical inputs/options produce deterministic report content within the same rendering environment. Documentation covers toggles, output paths, regeneration, and sample interpretation in `docs/smi_plotting.md`.

Validation: 30 combined CPU unit tests pass, plus Black, Ruff, and `git diff --check`. Tested inside `distracted_wu` on b9-2 across six GEMM shapes with both preshuffle variants: 12 ten-second monitoring windows produced 2,403 timestamped samples at a median 50.097 ms interval. All timestamps and original sample dictionaries were retained in the CSV, and six case plots were checked against the raw data. The existing summary path generated its 12-record/eight-plot report, and all 19 container compatibility tests passed. The benchmark source was preserved.

Two shapes emitted elementwise tolerance warnings but passed the benchmark's relative-L2 gate; A-preshuffle and split-K checks passed.


arakowsk-amd:
Added raw SMI capture and per-case timelines in commit 696cb0936, alongside the automatic summary plots already in this PR.

`AITER_SMI_TRACE=1` now preserves every timestamped observation in `samples.jsonl` under a separate run directory. `AITER_SMI_TRACE_DIR` defaults to `./smi_traces`. Capture is independent of `AITER_SMI_PLOT=1`, and raw data is written after each monitoring window. Existing stdout and `AITER_SMI_OUTPUT_PATH` behavior are preserved.

Example enabling both automatic summary plots and raw capture:

```bash
ENABLE_CK=0 \
AITER_SMI_MONITOR=1 \
AITER_SMI_PLOT=1 \
AITER_SMI_PLOT_DIR=/data/gemm_run/summary-plots \
AITER_SMI_TRACE=1 \
AITER_SMI_TRACE_DIR=/data/gemm_run/traces \
AITER_SMI_DURATION=10 \
AITER_SMI_INTERVAL=0.05 \
AITER_SMI_SKIP_FUNCTIONS=run_torch \
python3 op_tests/test_gemm_a8w8_blockscale.py \
  --flydsl --bpreshuffle True --apre True --data-init norm \
  -m 512 -nk 7168,16384
```

Render the detailed timelines explicitly after the run:

```bash
python3 aiter/smi_trace_plot.py /data/gemm_run/traces \
  --output /data/gemm_run/timelines
```

Each case gets PNG/SVG charts comparing variants in separate columns, an HTML index, and CSV/manifest exports. Every sample is plotted at its recorded timestamp without smoothing or downsampling; missing and firmware `N/A` values remain gaps. Existing summary-only logs cannot reconstruct a timeline. These plotting/capture options do not change the benchmark's graph-mode setting.

Validated inside `distracted_wu` on b9-2 with six shapes and both variants: 12 ten-second traces, 2,403 samples, and six case charts. All original samples were preserved; median sampling interval was 50.097 ms. All 19 existing container compatibility tests and 30 combined local CPU tests passed. The benchmark source was unchanged. Full usage is in `docs/smi_plotting.md`.

PR #5775, opened 2026-09-23
fix(custom_all_reduce): send the real IPC offset instead of 0

zhou9402:
## Problem

`_broadcast_ipc` (`aiter/dist/device_communicators/custom_all_reduce.py:602`) pins the IPC offset to `0`:

```python
self._ipc_handle_fn(data_ptr, handle.data_ptr())
return self._gather_ipc_meta((handle, 0))
```

`hipIpcGetMemHandle` always returns a handle for the **whole allocation** containing the pointer, never for the pointer itself, and the importing rank maps that allocation at its own base. A buffer that is a sub-block of a larger allocation therefore has to carry its `ptr - base` offset. Offset `0` is only correct when the pool pointer happens to *be* an allocation base.

Whether it is depends on the state of PyTorch's caching allocator:

- **Fresh allocator** — the IPC pool is the first large allocation and gets its own segment, so `ptr == base` and offset `0` is accidentally right. This is why single-engine runs are fine.
- **Allocator already holding a cached large segment** — a co-resident second engine on the same node, or simply any earlier allocate-and-free — the pool lands as a sub-block, `ptr != base`, and peers read from the base of that segment instead. **No error is raised; the all-reduce just returns data from whatever else lives there.**

The C++ graph path already does this correctly, and even carries the comment — `get_graph_buffer_ipc_meta` in `csrc/include/custom_all_reduce.cuh:3947`:

```cpp
// note: must share the base address of each allocation, or we get wrong address
if (hipPointerGetAttribute(&base_ptr, HIP_POINTER_ATTRIBUTE_RANGE_START_ADDR, ...))
offsets[i] = ((char*)ptr) - ((char*)base_ptr);
```

Only the Python side, for the named pool buffers (`meta` and `input`), was missing it. The C++ importers already apply `offsets[i]` (`custom_all_reduce.cuh:3911`, `:4001`, `:4025`, `:4124`), so no kernel-side change is needed.

## Fix

Resolve the base with `hipMemGetAddressRange` and send the real offset. New helper `_ipc_base_ptr`; `_broadcast_ipc` exports the handle from the base and gathers `data_ptr - base`. Raw `hipMalloc` / `hipExtMallocWithFlags` buffers are their own base, so this is a no-op for them.

## Relation to `AITER_CUSTOM_AR_RAW_INPUT_POOL`

That env var forces a raw `hipMalloc` input pool, whose pointer is always an allocation base — which is exactly why setting it made the problem disappear. It stays as an escape hatch and is still the right thing under `expandable_segments` (VMM pointers genuinely cannot be exported, #4174), but it should no longer be needed for co-resident engines. That matters in practice: a raw pool is invisible to torch's memory accounting, so an engine that sizes its KV cache from torch's numbers over-estimates free memory and can OOM later.

This is a different bug from #4921 (which was `init_dist_env` reading `_pool["input"].tensor` on a raw pool); they were both reachable in the same co-resident configuration, which is why the env var looked like the cure.

## Not a duplicate

Searched open PRs on this repo for `custom_all_reduce ipc offset` and `RAW_INPUT_POOL` — no open PR touches `_broadcast_ipc` or the IPC offset. #4921 is closed and addressed a different failure in the same file.

## Tests

`op_tests/test_ipc_base_ptr.py` (new, single-GPU) pins the offset arithmetic: sub-blocks of one allocation share a base and report distinct non-zero offsets; a raw `hipMalloc` buffer is its own base.

`op_tests/multigpu_tests/test_init_dist_env.py` gains a `dirty` mode that leaves a large segment in the caching allocator before `init_dist_env`, so the pool comes back as a sub-block — the shape a co-resident second engine produces incidentally. Each rank's pool offset is reported, so a run that never landed on a sub-block is distinguishable from one that did and handled it; the existing allreduce-correctness check is what proves the result.

**Validated on gfx950 (MI350X)** — full before/after table in [this comment](https://github.com/ROCm/aiter/pull/5775#issuecomment-5788840837). Summary: three allocator states that put the input pool at a non-zero offset produce **silently wrong** all-reduce results on the unpatched tree and correct results with this patch; states that leave the pool at offset 0 are unaffected either way.

## Note

AI assistance was used to write this change. I have reviewed every changed line.


zhou9402:
## Hardware validation

Measured on gfx950 (MI350X), ROCm 7.2, torch 2.12.1+rocm7.2, TP2, through the real `init_dist_env` path. Each row is a bf16 all-reduce of `(128, 8192)` over 2 ranks, compared exactly against the analytic sum. "unpatched" is the installed aiter as-is; "patched" replaces only `IPCBufferPool._broadcast_ipc` with this PR's version, nothing else.

The allocator is put into a given state before `init_dist_env`, and the offset column is the input pool's own `data_ptr - base`, read back after init.

| allocator state before init | input pool offset | unpatched | patched |
|---|---:|---|---|
| fresh | 0 | correct | correct |
| 512 MiB segment freed, 64 MiB live head | 0 | correct | correct |
| 8 GiB segment freed, 1 GiB live head | 1,082,130,432 | **WRONG** | correct |
| same + small-block fragmentation | 1,216,348,160 | **WRONG** | correct |
| 6 GiB live weights, 4 GiB activation segment freed, 256 MiB live head | 276,824,064 | **WRONG** | correct |


The failures are silent — no exception, no fallback, no log line. The all-reduce simply returns data read from the base of the segment the pool lives in.

### Two notes on reproducing it

Both cost me a wrong first attempt, so they are worth recording:

- **A plain allocate-then-free is not enough.** Splitting a cached block returns its *head*, so the pool still comes back at offset 0. There has to be a **live** allocation in front of it.
- **Scale matters.** At 512 MiB the allocations `init_dist_env` itself makes on the way to the pool consume the remainder of the segment, and the pool gets a fresh segment anyway (the `head64` row above). Multi-GiB reproduces reliably.

That combination is why this is intermittent and why it has gone unnoticed: a first engine on a fresh allocator always lands at offset 0 and is always correct.

Scope note: what is demonstrated here is the dependence on caching-allocator state, using a loaded-engine-shaped allocation pattern as the trigger. I have not run two real co-resident engines end to end.

### Test shape corrected

The `dirty` mode added in this PR originally used the allocate-then-free shape, which measurement showed does not reproduce. It now uses the shape in the last row, with a parent-side guard that skips when there is not enough free memory.

PR #5786, opened 2026-09-23
[JIT] Make CK fmha kernel generation follow GPU_ARCHS

qiangpan2:
CK's FMHA codegen was never told which GPU architecture to build for — it either fell back to `gfx9,gfx950`, or was handed a device name it does not recognize, leaving gfx11/gfx12 with no usable FA kernels. This PR derives CK's `--targets` from `GPU_ARCHS` and passes it at every `01_fmha/generate.py` call site.

## Why

Two independent bugs, both silent until dispatch:

- **`mha_recipes._ck_targets_flag()` passed a raw device name** (`gfx1100`) as `--targets`. CK's factory keys are coarser than device names — `gfx11` covers `gfx1100`–`gfx1103` — so it matched nothing, produced an empty kernel set, and `mha_fwd` failed with `invalid argument for fmha_fwd`. **FA did not work on gfx11/gfx12 at all.**
- **Every other call site omitted `--targets`**, so CK fell back to its built-in `gfx9,gfx950` and `GPU_ARCHS` was ignored.

## What changed

203 lines production, 451 tests. Start at `build_targets.py`, then the gates in `mha.py`:

| File | |
|---|---|
| `aiter/jit/utils/build_targets.py` | +69 — the mapping table (§1) |
| `aiter/ops/mha.py` | +100/−23 — call sites and arch gates (§3, the subtle one) |
| `optCompilerConfig.json` · `mha_recipes.py` · `cpp/mha/compile.py` · `core.py` | remaining `--targets` sites (§2) |
| `op_tests/test_ck_fmha_targets*.py` | +339 — no-GPU tests for the table and every gate |

**§1 — Map `GPU_ARCHS` to a CK factory key.**

```
GPU_ARCHS         CK --targets
─────────────────────────────
gfx950       -->  gfx950
gfx125x      -->  gfx125
gfx120x      -->  gfx12
gfx115x      -->  gfx115
gfx11xx      -->  gfx11
other gfx9*  -->  gfx9
unmapped     -->  gfx9,gfx950   (CK's own default — unchanged)
```

`batch_prefill` needs a gfx9-only variant of this, because CK drops *every* batch_prefill kernel if any `--targets` token is not `gfx9*`.

**§2 — Every `generate.py` invocation now passes it**, across the four files that launch it: `optCompilerConfig.json`, the `cmdGenFunc_*` hooks in `mha.py`, `mha_recipes.py`, and the standalone `op_tests/cpp/mha/compile.py`.

**§3 — fp8 / batch_prefill gates, and which arch they read.** fp8 has no CK factory on gfx11/gfx115, and `batch_prefill` is gfx9-only, so both now raise `NotImplementedError` up front instead of failing deep inside CK dispatch.

The catch: `compile_ops` re-runs `gen_func` on **every** call to build its module-name cache key. So `cmdGenFunc_*` sits on the hot path, not just the build path, and the two callers need different arch sources:

```
PREBUILD (setup.py, no GPU present)
└── cmdGenFunc_mha_fwd(check_build_arch=True)
    └── _check_fp8_factory(get_gfx(), ...)           # build time: honor GPU_ARCHS

live call
└── flash_attn_fp8_pertensor_func(...)
    ├── _check_fp8_factory(get_gfx_runtime(), ...)   # public API: the real card
    └── mha_fwd(...)                                  [compile_ops]
        └── _cmdGenFunc_mha_fwd_lookup(...)           # NEW — runs on every call
            ├── _check_fp8_factory(get_gfx_runtime(), ...)
            └── cmdGenFunc_mha_fwd(check_build_arch=False)
```

Without the split, `GPU_ARCHS=gfx942;gfx1100` makes `get_gfx()` return `gfx1100` (last token wins) and the gate would reject fp8 on a real gfx942 card. `mha_batch_prefill` gets the identical treatment.

## ⚠️ Behavior change

`GPU_ARCHS="gfx942"` wheels no longer contain gfx950 FA kernels — CK's `gfx9,gfx950` default used to slip them in regardless. **Set `GPU_ARCHS` to every arch the wheel must run on.** `docs/installation.rst` updated.

## Verified

- **No GPU needed** — `op_tests/test_ck_fmha_targets{,_wiring}.py`, 43 passed: the mapping table plus every gate, including the `get_gfx()` / `get_gfx_runtime()` split in §3.
- **gfx1201 (RDNA4)** — 4 new cases in `test_mha.py` pass: hdim-80 pad-up fp16/bf16, fp8bf16 hdim64, fp8bf16 GQA (`max_diff=0.05878` against a `0.07` bound — per-tensor quant noise on the GQA broadcast, not a kernel defect).
- **gfx942/gfx950** — the release default `GPU_ARCHS="gfx942;gfx950"` still maps to `gfx9,gfx950`, byte-identical to CK's old default, so the shipped wheel is unaffected; only single-arch builds narrow (see Behavior change). Leaning on this PR's CI for the regression run.
- `ruff check .` (CI-pinned 0.16.0) and `black --check` clean. No CK change, no submodule bump.


zufayu:
@aiter-bot review

qiangpan2:
Responding to each item in order. Two corrections to the PR body have already been pushed; one finding I don't think reproduces.

---

### 1. Ownership and sign-off

> 🔴 `aiter/ops/mha.py` and `aiter/jit/core.py` change which fmha kernels every wheel and JIT build generates (...) yet the PR has zero reviews and author qiangpan2 has no prior commits on either file, whose de-facto owners are slippedJim (23 mha.py commits) and amd-ruitang3 (35 jit/core.py commits) (...) **Reviewer must** @-mention slippedJim and amd-ruitang3 in a PR comment requesting explicit sign-off before merge.

Thanks for the reminder. **@slippedJim**, this changes dispatch in `aiter/ops/mha.py` — `--targets` for CK's fmha codegen is now derived from `GPU_ARCHS` rather than falling through to CK's built-in `gfx9,gfx950`, and fp8 on gfx11/gfx115 plus `mha_batch_prefill` off gfx9 now raise `NotImplementedError` up front. You own that file; please take a look before this merges. The user-visible consequence is called out under **⚠️ Behavior change** in the PR body.

One scope note:

**`aiter/jit/core.py` is not a second blast-radius surface here.** The entire change to that file is five lines of import, so that the `eval()`'d `blob_gen_cmd` strings in `optCompilerConfig.json` can resolve the two helpers:

```python
# json blob_gen_cmd eval() needs these names in module globals.
from build_targets import (  # noqa: F401
    ck_fmha_batch_prefill_gen_targets,
    ck_fmha_targets,
)
```

No logic, no control flow.

---

### 2. RDNA run evidence

> ⚠️ The new RDNA cases in `op_tests/test_mha.py` skip on every arch CI runs (the `_RDNA`/`_GFX12` skipif sets vs the gfx942+gfx950 aiter op runners) and no RDNA CK lane exists, so the mapping (...) has run evidence only from the author's gfx1201 box (4 cases) — gfx1100–1103, gfx1150/1151, gfx1200 and gfx1250 have zero run evidence. **Author must** link the gfx1201 run log (command plus output) and state which other RDNA archs were smoke-tested, or **Reviewer should ask** for an RDNA CI lane before relying on the table.

Correct — no RDNA runner exists, so those cases skip. The coverage is now stated explicitly in the PR body: gfx1201 is the only arch with hardware evidence; gfx1100–1103, gfx1150/1151, gfx1200 and gfx1250 are covered by the mapping-table test only.

One note on the remedy: **there is no CK FA lane on any arch to extend.** `flash_attention_ck` is `if: false  # Disabled until CK tests are ready` (`.github/workflows/flash_attention_integration.yaml:614-615`). That's a repo-level gap rather than something this PR can close — happy to file it separately if useful.

---

### 3. `ci:all` and the label gating

> ⚠️ The diff touches `aiter/ops/mha.py` and `aiter/jit/core.py` — both downstream-consumed paths — but the PR carries only the [JIT] label, so the Atom/SGLang/vLLM jobs all skip; the release default `GPU_ARCHS="gfx942;gfx950"` does map byte-identically to CK's old `gfx9,gfx950` default, which is why this is insurance rather than a blocker. **Reviewer should ask** for one `ci:all` (or at minimum `ci:atom`+`ci:vllm`) run green before merge.

**`ci:all` isn't necessary here — though I'll add it if you'd like.** Reasons:

- `ck_fmha_targets()` reads the full `GPU_ARCHS` list, so the release default `"gfx942;gfx950"` maps to `"gfx9,gfx950"` — byte-identical to CK's old default, as you note.
- The gfx942/gfx950 lanes therefore generate the same kernel set before and after this PR; a green run says nothing about the change itself.
- The configuration that does change is a single-arch build such as `GPU_ARCHS=gfx942`, and no current lane runs one.

---

### 4a. The PR body's call graph

> 📝 The PR body's call graph claims "PREBUILD (setup.py) → cmdGenFunc_mha_fwd(check_build_arch=True)", but setup.py's prebuild (setup.py:296–354) drives codegen through `get_args_of_build` plus `get_mha_varlen_prebuild_variants_by_names` and never calls any `cmdGenFunc_*`, so the `check_build_arch=True`/`get_gfx()` branch in `aiter/ops/mha.py` has no production caller (...) **Author must** correct the diagram.

Correct — PR body updated for the call graph.

---

### 4b. `_cmdGenFunc_mha_fwd_lookup` and keyword `q`

> also `_cmdGenFunc_mha_fwd_lookup(q, *args, **kwargs)` now TypeErrors if `q` arrives as a keyword where the old gen_func hook accepted it. **Author must** (...) confirm no caller passes `q` by keyword.

**This doesn't reproduce — no call shape raises.** Since the failure mode matters, both readings laid out:

**The concern, as I read it:** the wrapper introduces a named first parameter where the old hook had none, so a keyword call such as `mha_fwd(q=..., k=...)` would either

- **(a)** fail to bind `q` at all, or
- **(b)** bind it and then collide on the forward — `cmdGenFunc_mha_fwd(q, *args, **kwargs)` passes `q` positionally, so if `**kwargs` still carried `q` the result would be `TypeError: got multiple values for argument 'q'`.

**Why neither happens:**

- `q` is *positional-or-keyword*, not positional-only — there is no `/` in the signature — so `_lookup(q=tensor, ...)` binds it normally. (a) doesn't occur.
- Binding *consumes* it: once `q` is matched, Python removes it from `**kwargs`. So the positional forward has nothing to collide with. (b) doesn't occur.
- Separately, `check_build_arch` sits after `*`, making it keyword-only, so it cannot be filled positionally even when all 21 parameters are passed positionally.

All three call shapes:

```python
>>> def cmdGenFunc_mha_fwd(q, k, v, dropout_p, *, check_build_arch=True):
...     return (q, k, v, dropout_p, check_build_arch)
>>> def _lookup(q, *args, **kwargs):
...     return cmdGenFunc_mha_fwd(q, *args, **kwargs, check_build_arch=False)
>>> _lookup(1, 2, 3, 0.0)                    # all positional
(1, 2, 3, 0.0, False)
>>> _lookup(q=1, k=2, v=3, dropout_p=0.0)    # q as keyword
(1, 2, 3, 0.0, False)
>>> _lookup(1, 2, v=3, dropout_p=0.0)        # mixed
(1, 2, 3, 0.0, False)
```

Scope: `compile_ops` forwards verbatim — `custom_build_args.update(gen_func(*args, **kwargs))` at `aiter/jit/core.py:1978` — so whatever shape reaches `mha_fwd` reaches the hook unchanged, and the three above are exhaustive. No change made here; happy to be shown a failing case if I've missed one.

---

### Summary

| Item | Outcome |
|---|---|
| 1 — owner sign-off | slippedJim pinged; `jit/core.py` scope clarified (5-line import) |
| 2 — RDNA evidence | Accepted; coverage table added to PR body; no CK FA lane exists to extend |
| 3 — `ci:all` | Label available on request; noted it doesn't cover the changed configuration |
| 4a — call graph | Fixed, PR body updated |
| 4b — keyword `q` | Doesn't reproduce; repro above |

No production code changed in response to this review.

zufayu:
Please remove `op_tests/test_ck_fmha_targets.py` and `op_tests/test_ck_fmha_targets_wiring.py` from `op_tests`. These test JIT target mapping, command construction, and mocked dispatch rather than operator correctness. `op_tests` should remain focused on tests that execute and validate operators. If this coverage is still needed, it should live under a dedicated JIT/unit-test location.

zufayu:
We no longer support CK FMHA on gfx1250. The intended path is ASM/FlyDSL/Triton, using the existing CK-free infrastructure. This PR adds `gfx1250 -> gfx125`, which can generate CK FMHA when `ENABLE_CK` is left at its default. Please remove gfx1250 from the CK FMHA target mapping and make gfx1250 builds skip CK FMHA generation rather than rejecting the platform at runtime.

zufayu:
Since setup PREBUILD does not call `cmdGenFunc_*`, and the production `compile_ops` lookup wrappers always pass `check_build_arch=False`, the `check_build_arch=True` branch currently has no production caller. Please remove this parameter and the unused build-time branch when cleaning up the JIT-only `op_tests`.

zufayu:
Since this is a build-time target-selection change, validating the affected single-architecture configurations does not require GPU hardware. Please add a build/codegen smoke test that exercises the actual prebuild path with `GPU_ARCHS=gfx942` and `GPU_ARCHS=gfx950`, and verifies that a gfx1250 build does not select or generate CK FMHA. Mocking the mapping helpers in `op_tests` does not validate this integration path.

zufayu:
Please preserve the existing CI PREBUILD flow: one `GPU_ARCHS` list and one wheel build. Today it uses `gfx942;gfx950`, and it should remain extensible to `gfx942;gfx950;gfx1250` without introducing a separate gfx1250 flow. CK FMHA target selection should filter out gfx1250, so the combined build still generates only `gfx9,gfx950` CK FMHA kernels, while a gfx1250-only build skips CK FMHA codegen entirely. This should be handled during build/module selection, not with a runtime assertion.

qiangpan2:
Two commits on top of the existing history, no force-push.

---

## `865fea77` — [JIT] Exclude gfx1250 from CK fmha target selection, drop dead check_build_arch param

Production code: `aiter/jit/utils/build_targets.py`, `aiter/ops/mha.py`, `setup.py`, `docs/installation.rst` (+86 −34).

**Dropped `check_build_arch`.**
> "Since setup PREBUILD does not call `cmdGenFunc_*`, and the production `compile_ops` lookup wrappers always pass `check_build_arch=False` ... Please remove this parameter and the unused build-time branch" ([comment](https://github.com/ROCm/aiter/pull/5786#issuecomment-5808981865))

Parameter and its guarded branch removed from all three `cmdGenFunc_*` functions, and the `check_build_arch=False` argument from the three `_*_lookup` wrappers. `grep check_build_arch aiter/ops/mha.py` now returns nothing. The fp8 and batch_prefill gates still run on `get_gfx_runtime()`, unchanged.

**Removed gfx1250 from the CK fmha target mapping.**
> "We no longer support CK FMHA on gfx1250 ... Please remove gfx1250 from the CK FMHA target mapping and make gfx1250 builds skip CK FMHA generation rather than rejecting the platform at runtime." ([comment](https://github.com/ROCm/aiter/pull/5786#issuecomment-5808908448))

`build_targets.py:129` adds `_CK_FMHA_EXCLUDED_ARCHS = frozenset({"gfx1250"})`; `map_gpu_archs_to_ck_fmha_targets` skips it at `:184`. The `("gfx125", "gfx125")` prefix row is dropped with it — gfx1250 was the only gfx125x part in `GFX_MAP`, so that row is unreachable once it's excluded.

**Filtered during build/module selection, in two places.**
> "CK FMHA target selection should filter out gfx1250, so the combined build still generates only `gfx9,gfx950` CK FMHA kernels, while a gfx1250-only build skips CK FMHA codegen entirely. This should be handled during build/module selection, not with a runtime assertion." ([comment](https://github.com/ROCm/aiter/pull/5786#issuecomment-5809015153))

```python
    if not ck_fmha_enabled():                                          # setup.py:260
        exclude_ops.extend(sorted(CK_FMHA_MODULES))
...
        if PREBUILD_KERNELS == 1 and ENABLE_CK and ck_fmha_enabled():  # setup.py:335
```

The second is an additional bypass found while verifying the first: `PREBUILD_KERNELS == 1` injects five hardcoded mha-varlen variants through `get_mha_varlen_prebuild_variants_by_names`, which never consults `get_exclude_ops()` — a gfx1250-only wheel build would otherwise still have generated them.

Verified against the real `blob_gen_cmd`:

```text
gfx942                   -> gfx9
gfx942;gfx950            -> gfx9,gfx950
gfx942;gfx950;gfx1250    -> gfx9,gfx950
gfx1250                  -> gfx9,gfx950   (never reached; modules excluded first)
```

One `GPU_ARCHS` list, one wheel, no separate gfx1250 flow. `_check_fp8_factory` and `_check_batch_prefill_arch` are unchanged — unrelated to gfx1250, and their modules still get built.

**Corrected `docs/installation.rst`**, which claimed CK fmha kernels are generated only for the architectures listed — now carves out the gfx1250 exception.

---

## `fff9e225` — [JIT] Replace mocked CK fmha target tests with real, no-mock coverage

Tests only: deletes `op_tests/test_ck_fmha_targets.py` (83 lines, 5 mock uses) and `op_tests/test_ck_fmha_targets_wiring.py` (256 lines, 26 mock/patch uses); adds `op_tests/jit_tests/test_ck_fmha_prebuild_targets.py` (+235 −339).

**Removed both mocked files from `op_tests`.**
> "Please remove `op_tests/test_ck_fmha_targets.py` and `op_tests/test_ck_fmha_targets_wiring.py` from `op_tests` ... If this coverage is still needed, it should live under a dedicated JIT/unit-test location." ([comment](https://github.com/ROCm/aiter/pull/5786#issuecomment-5808449087))

Deleted rather than relocated — moving them as-is would have carried over the mocking problem. The new directory is a sibling of the existing `flydsl_tests` / `tuning_tests`.

**Replaced them with real, no-mock coverage.**
> "Please add a build/codegen smoke test that exercises the actual prebuild path ... Mocking the mapping helpers in `op_tests` does not validate this integration path." ([comment](https://github.com/ROCm/aiter/pull/5786#issuecomment-5808982174))

9 functions / 25 cases, no GPU, no mocks:

- `test_prebuild_blob_gen_cmd_targets` — subprocess per `GPU_ARCHS` (`get_gfx_list()` is `lru_cache`d), reads the real `blob_gen_cmd` from `core.get_args_of_build`, asserts `--targets`.
- `test_prebuild_blob_gen_cmd_targets_for_batch_prefill` — same for `module_mha_batch_prefill`, which has its own gfx9-only-subset rule.
- `test_setup_get_exclude_ops_*` — loads the real `setup.py` via `importlib` and calls `get_exclude_ops()`.
- `test_ck_fmha_modules_match_config` — re-derives the module set from `optCompilerConfig.json`, so the list can't drift from the config it mirrors.

---

`ruff check .` and `black --check` clean on every touched file; no CK submodule change, no `3rdparty/` change.

PR #5796, opened 2026-09-23
[FlyDSL] AOT-compile the gfx950 FP8 flash-attention forward

gbyu-amd:
## Motivation

`flydsl_flash_attn_fp8_func` (gfx950 DUALWAVE_SWP FP8 forward) chooses its compile-time configuration per call:
- the rescale threshold, from `seqlen_kv`;
- `BLOCK_M`, from the KV tile count and occupancy;
- the split-K factor, from the makespan model.

Each new combination is a JIT compile on the serving path. No FlyDSL AOT kind covered this kernel, so a server always paid these compiles inside live requests. In a Kimi-K3 serving run on MI355X (TP8, MLA chunked prefill), 29–36 variants were compiled per run, each stalling the request that triggered it by roughly 1–2 s of TTFT.

## Why the variant space is finite

The launcher's cache key is fixed by the `_build_fp8` parameters alone. That comes from `traits.cache_tag`, the closure scalars read by `launch_flash_attn_dualwave_swp`. Everything else enters the key by type only:
- runtime `Int32`/`Float32`/`Stream` arguments (batch, sequence lengths, strides, softmax scale, `lse_stride_h`);
- the flattened 1-D tensors, which contribute only their dtype.

So for one head shape, the reachable set is a small product: rescale threshold, `BLOCK_M`, `num_kv_splits`, `batch_interleave_group`, and (`causal`, `return_lse`).

## Changes

- **`aiter/aot/flydsl/fmha_fp8.py`**: new `FMHA_FP8` kind. The job list is enumerated from the heuristics' own module constants (`_FP8_LONG_SEQ`, `_FP8_BLOCK_N`, `_FP8_NARROW_MAX_KV_TILES`, `_FP8_AUTOSPLIT_*`, `_FP8_BATCH_INTERLEAVE_GROUP`), so it follows them if they change. Kernels are compiled through the kernel's existing `_launch.compile` hook, which applies the same compile hints as `_launch`. The placeholder call mirrors the runtime `exe(...)`: the same optional kwargs present or absent, and the same dtypes.
- **`common.py`**: register the kind.
- **`README.md`**: add a table row and a usage section.
- **Opt-in**: nothing is compiled unless `--shape` or `AITER_FLYDSL_AOT_FMHA_FP8="H:Hkv:D:Dv[@layout];..."` is given. `run_aot` / `setup.py` behave exactly as before otherwise.

```bash
python -m aiter.aot.flydsl.fmha_fp8 --shape 12:12:192:128          # Kimi-K3 MLA prefill, TP8
AITER_FLYDSL_AOT_FMHA_FP8="12:12:192:128" python setup.py develop
```

## Test

Kimi-K3 at TP8, `12:12:192:128@varlen_cross`:
- **92 kernels, 92 ok**, about 6 s wall on CPU. No GPU is needed (`COMPILE_ONLY` + `FLYDSL_GPU_ARCH=gfx950`).
- **Key match with serving.** Serving-JIT artifacts were collected from two independent K3 serving runs: 68 directories in total. All 68 directory names and pkl names appear among the 92 AOT outputs.
- **Run-only on GPU.** `FLYDSL_RUNTIME_RUN_ONLY=1` makes a cache miss raise, and the cache held only the AOT output. 40 calls covered 24 distinct (causal, lse, threshold, block_m, splits) configs across B/Sq/Skv buckets (Skv 512–32768). All of them loaded without compiling, and the cache directory did not grow. Outputs and LSE match an fp32 reference on the same fp8 inputs (max abs err 0.056).
- **Serving.** A Kimi-K3 AgentX run was served on MI355X with the 92 artifacts installed (c1, first 30 min of a 3600 s window). No `launch_flash_attn_dualwave_swp` was JIT-compiled: the cache still holds exactly the 92 AOT directories. Paired by request against the same build without AOT, TTFT p90 fell from 2536 to 1909 ms and p50 from 735 to 716 ms.
- **Lint.** `ruff check` (0.16.0) and `black` are clean.

## Caveat for image builders

The FlyDSL cache key hashes function sources and closure scalars, but not the method bodies of helper **classes** such as `op_gemm`, `op_softmax` and `op_lds` in `fmha_gfx950/`. Also, `COMPILE_ONLY` never overwrites an existing pkl. So after changing only those helpers, re-run AOT into an **empty** cache directory rather than on top of an old one; otherwise stale artifacts keep matching. This applies to every FlyDSL AOT kind, not just this one.


coderfeli (dismissed):
LGTM

coderfeli (approved):

PR #5799, opened 2026-09-23
[CustomAllreduce] Re-take the expandable_segments decision at capture time

okorzh-amd:
`CustomAllreduce` refuses its registered (copy-free) CUDA-graph capture path while PyTorch `expandable_segments` is on, because VMM-backed pointers cannot be exported with `hipIpcGetMemHandle` (#4174). It then stages every all-reduce input through a device-to-device copy (`__amd_rocclr_copyBuffer`) into the pre-registered input pool.

That veto is decided **once, in `__init__`**:

```python
if not self._use_vmm and _expandable_segments_enabled():
    ...
    enable_register_for_capturing = False
self.enable_register_for_capturing = enable_register_for_capturing
```

A host that wants the registered path has one clean way to get it: keep expandable segments for weights, KV cache and eager allocations, and turn them off only for the duration of graph capture, so the graph pool's activations are ordinary `hipMalloc` memory and *are* exportable. Today that does nothing. By the time `capture()` runs the veto has already latched, and the copy-in path is taken regardless of what the allocator is doing.

## The fix

- Record what the caller asked for (`_register_for_capturing_requested`) *before* the `expandable_segments` veto.
- In `capture()`, re-evaluate the veto against the **live** allocator state. `_expandable_segments_enabled()` already reads the allocator snapshot, so no new plumbing. If expandable segments are off at that point, use the registered path for this capture.
- Restore the flag in `capture()`'s `finally`, so the decision is scoped to the capture window and never leaks into eager calls.
- The gfx1250 veto is architectural (peers would dereference unregistered pointers on replay) and is deliberately **not** re-evaluated. The VMM transport is unaffected.
- Add a class-level marker, `CustomAllreduce.defers_capture_registration = True`, so hosts can feature-detect this instead of inferring it from a version number. vLLM's side (vllm-project/vllm#57481) probes for it and stays inert without it.

Behaviour with the default configuration is unchanged. If expandable segments are on at `__init__` and still on at capture, the copy-in path is used exactly as before. If they are off at `__init__`, nothing is vetoed in the first place. Disabled communicators (unsupported world size, multi-node group, no `custom_ar` library) are still entered by `graph_capture()`. Both new attributes are set before `__init__` can return early, so `capture()` keeps working there. If `flush_graph_buffers` raises, the flag is still restored.

The re-check only ever *relaxes* the veto, and only for a caller that asked for the registered path. **Precondition for hosts:** every all-reduce input captured in the window must be allocated inside it. A tensor allocated earlier, while expandable segments were on, is still VMM-backed and cannot be IPC-exported. That is the case when the graph pool's activations feed the all-reduces, which is what a graph-capturing host normally does. The precondition is documented on `defers_capture_registration`.

## Impact

Measured with vLLM on 8x MI355X, Kimi-K3 decode (conc 1, 131k context, MTP K=6), rank 0, 4 captured passes. The host turns expandable segments off for capture (vllm-project/vllm#57481). The aiter side was applied as a runtime overlay carrying the same `capture()` logic as this PR:

| | host toggle only | host toggle + this change |
|---|---|---|
| `__amd_rocclr_copyBuffer` launches | 1396 | **236** |
| kernels launched | 12188 | **11028** |
| KDA layer (main stream) | 338.6 us | **263.3 us** |
| MLA layer (main stream) | 591.9 us | **475.8 us** |

The drop is **exactly -1160 copies = 792 two-stage + 368 one-stage all-reduce calls**, i.e. one staging copy per collective, all gone. To confirm a host is getting the registered path, look for `expandable_segments is off at capture time; using the registered custom-allreduce capture path` (logged at INFO on each capture). The `__init__` warning still fires when expandable segments are on at construction. It now says the decision is re-checked at capture instead of saying it forces copy-in.

## Test plan

New `op_tests/multigpu_tests/test_custom_allreduce_capture_registration.py`. Neither existing test covers both axes: `test_init_dist_env.py` sets expandable segments but captures no graph, and `test_custom_allreduce.py` captures a graph but never touches the allocator config. Each rank starts with `expandable_segments:True`, builds the communicator, then captures one all-reduce with and without the host toggle. It asserts:

- the `__init__` veto still fires (`enable_register_for_capturing is False` after construction);
- inside `capture()` the registered path is on if and only if the host turned expandable segments off;
- the decision does not leak out of `capture()`;
- every rank picked the same path. The registered/unregistered choice changes the kernel algorithm and the graph-buffer count, so a split across ranks is a collective mismatch, not a local inefficiency;
- the replayed graph produces the correct sum.

On platforms where the veto cannot happen (gfx1250, the VMM transport, or a build that ignores
`expandable_segments`), the flag assertions are skipped with a warning and only the result is
checked, as `test_init_dist_env.py` does. `-t 3` covers a disabled communicator.

```
for tp in 2 3 4 8; do python3 op_tests/multigpu_tests/test_custom_allreduce_capture_registration.py -t $tp; done
python3 op_tests/multigpu_tests/test_custom_allreduce.py
python3 op_tests/multigpu_tests/test_init_dist_env.py
```

## Test result

8x MI355X (gfx950), `vllm/vllm-openai-rocm:nightly-rocm100-3df4ae15` (torch 2.12.0+rocm10.0.0).
aiter was built from source. The control is upstream `main` at 1ec3cf48e, the base of this PR.

**New test**

| | TP=2 | TP=3 (disabled comm) | TP=4 | TP=8 |
|---|---|---|---|---|
| this PR, `toggle=False` | pass (copy-in) | pass | pass (copy-in) | pass (copy-in) |
| this PR, `toggle=True` | **pass (registered)** | pass | **pass (registered)** | **pass (registered)** |
| `main`, `toggle=True` | FAIL: stays on copy-in | – | FAIL | FAIL |

On `main`, `toggle=True` fails at every TP size with `expected enable_register_for_capturing=True inside capture(), saw False`, so the test catches the latched veto. TP=3 failed with `AttributeError: 'CustomAllreduce' object has no attribute 'enable_register_for_capturing'` on an earlier revision of this PR that set the attributes after `__init__`'s early returns. That is why they are now set first.

**Existing tests**

- `test_init_dist_env.py`: passes, all three modes (default / expandable / raw_override), in 6 of 8 runs across this PR and `main`. Two runs on this PR's tree hung, each time after other multi-GPU tests had run in the same container. Both hangs were in torch's rendezvous (`init_process_group` -> `_create_c10d_store`), which runs before any `CustomAllreduce` is constructed, so they cannot involve this change. The same sequence passed on both trees when repeated, and I could not pin down the cause.
- `test_custom_allreduce.py`: fails **identically on this PR and on `main`** in this environment, with `AttributeError: 'DataFrame' object has no attribute 'host_time_sum'`. The profiler returns no events, and the error is raised in `perftest`'s profiler post-processing (`aiter/test_common.py`), not in an all-reduce check. With expandable segments off, which is that test's default, this change is a no-op anyway: nothing is vetoed, so there is nothing to re-take.


op_tests/multigpu_tests/test_custom_allreduce_capture_registration.py
```diff
@@ -0,0 +1,170 @@
...
+    # mismatch, not a local inefficiency.
+    decisions = {r["during_capture"] for r in rets}
+    assert len(decisions) == 1, f"ranks disagreed on the capture path: {rets}"
+    return {"during_capture": decisions.pop()}
+
+
+if __name__ == "__main__":
+    freeze_support()
+    parser = argparse.ArgumentParser(description="config input of test")
+    parser.add_argument("-t", "--tp_size", type=int, default=2)
```
jamesETsmith: Would it make sense to update the CI to test other TP values (e.g. 4 or 8)?

okorzh-amd: that would depend on capacity current CI maintainers can allocate for it. Likely out of scope for this PR as it is not adding a new test group. 

I've tested 4 and 8 locally on my machine and it works as expected.

okorzh-amd: @Rohan138 do you have any suggestions on this? my expectation PR changes are not rank dependent so test one rank would be enough.

Rohan138: I don't really know AITER CI best practices, maybe @gyohuangxin?

PR #5800, opened 2026-09-23
fix gluon import and benchmark + tune fused_clamp_act_mul for gfx950

nidal567:
## Motivation

On `gfx950`, `fused_clamp_act_mul` regression on Triton 3.8 at large-M prefill shapes up to 7.2%

| M | N (n_half) | Triton 3.4 | Triton 3.8 | delta |
|---|---|---|---|---|
| 16384 | 2048 | 34.02 us | 36.47 us | **+7.18%** |
| 8192 | 4096 | 34.44 us | 36.00 us | **+4.52%** |
| 16384 | 4096 | 62.26 us | 64.88 us | **+4.21%** |

The kernel was not tuned at M, as it was only tuned previously on row width (N). Essentially, `M=1` and `M=16384` at the same N received the same launch config. The triton path launches `grid=(M,)`, meaning one program per row, and `BLOCK_SIZE_N` is the whole row. In this case, `num_warps` is the only knob to be tuned, and past a few thousand rows, the grid alone fills the GPU. The shipped width-keyed value is tuned for small M and is too high for prefill. Decode (M <= 128) is unaffected and measures identically on both compilers.

Also, the benchmark for this kernel was not usable on main. It fails to import `_is_gluon_available()` and it crashes `rocprofv3` runs for shapes due to `rep` being hardcoded. Both issues are fixed here. 

## Technical Details

The kernel now checks the N-specialized `FUSED_CLAMP_ACT_MUL-N={N}.json` for the largest `M_GEQ_<x> <= M` before falling back to the existing `DEFAULT.json`, which is unchanged.

```
aiter/ops/triton/fusions/fused_clamp_act_mul.py                     +32 -8
op_tests/op_benchmarks/triton/bench_fused_clamp_act_mul.py          +10 -2
configs/gfx950/triton/fusions/fused_clamp_act_mul/FUSED_CLAMP_ACT_MUL-N=2048.json  (new)
configs/gfx950/triton/fusions/fused_clamp_act_mul/FUSED_CLAMP_ACT_MUL-N=4096.json  (new)
```

Tuned values (gfx950, from a `num_warps` sweep of 1/2/4/8/16 per shape):

| N | shipped | M >= 4096 | M >= 16384 |
|---|---|---|---|
| 2048 | 4 | **2** | **1** |
| 4096 | 8 | **4** | 4 |
| 512, 8192 | 1, 8 | unchanged (optimal) | |

### Benchmark fixes

- `is_gluon_available()` restored, as the benchmark didn't import properly. The guard now asserts properly.
- `--rep` made configurable, because the default 100 crashes `rocprofv3`. Make it an argument so that one can define the number of reps desired.

## Test Plan

- Correctness tests with pytest
- Black and Ruff checks
- `rocprofv3 --kernel-trace` runs with kernel time (10/80/10 trimmed mean)
- 23 shapes: N in {512, 2048, 4096, 8192} x M in {1, 16, 64, 128, 4096, 8192, 16384}, covering every config bucket plus decode and prefill

## Test Result

**23 shapes: 0 regressions, 5 improvements, 18 neutral.** The 5 improvements are exactly the 5 measured shapes whose `num_warps` changes; every shape whose routing is unchanged is neutral.

| M | N | main | this PR | change |
|---|---|---|---|---|
| 4096 | 2048 | 9.95 us | **9.31 us** | **-6.4%** |
| 16384 | 2048 | 36.46 us | **34.63 us** | **-5.0%** |
| 4096 | 4096 | 19.07 us | **18.44 us** | **-3.3%** |
| 8192 | 4096 | 35.89 us | **34.62 us** | **-3.5%** |
| 16384 | 4096 | 65.60 us | **63.17 us** | **-3.7%** |

**Decode unchanged.** All 13 shapes at M <= 128 are neutral
**Correctness:** `330 passed, 454 skipped, 200 warnings in 16.15s`
**Style:** `black` and `ruff==0.16.0` clean.
**Benchmark:** imports and runs; its default sweep profiles under `rocprofv3` without crashing, and `--rep 100` still reproduces the SIGSEGV on `--shape 1 512` and `--shape 16384 512`.

## Submission Checklist


aiter/ops/triton/fusions/fused_clamp_act_mul.py
```diff
@@ -23,6 +23,16 @@
...
 
+_GLUON_SUPPORTED_ARCHS = ("gfx1250",)
+
+
+def _is_gluon_available():
+    """Check if the gluon backend is supported for current GPU architecture."""
+    try:
+        return any(supported in get_arch() for supported in _GLUON_SUPPORTED_ARCHS)
+    except Exception:  # noqa: BLE001
+        return False
```
brunomazzottiamd: This function is brutally complex! Why not `return get_arch() in _GLUON_SUPPORTED_ARCHS`? I've seen this same piece of code before in another PR, it seems AI agents like to repeat it over and over again... Please, try to simplify this small helper function.

op_tests/op_benchmarks/triton/bench_fused_clamp_act_mul.py
```diff
@@ -222,6 +222,14 @@ def parse_args(args: list[str] | None = None):
         help="weight layout when --weighted 1: 'broadcast' is [M, 1] applied "
         "across the row, 'full' is [M, n_half], one weight per output element.",
     )
+    parser.add_argument(
+        "--rep",
```
brunomazzottiamd: Can you please explain me how did you use `--rep` argument in your tuning of the target op for `gfx950`?

aiter/ops/triton/fusions/fused_clamp_act_mul.py
```diff
@@ -47,6 +58,19 @@ def _get_config(M: int, N: int, block_size_n: int, backend: str) -> dict:
     base = f"{AITER_TRITON_CONFIGS_PATH}/{arch}/{backend}/fusions/fused_clamp_act_mul"
 
     if backend == "triton":
+        specialized = load_config_json(
```
brunomazzottiamd: I think the proposed change in `_get_config` makes sense. However, the function is becoming very repetitive and hard to understand.

It's the same skeleton twice:

1. load `...-N={N}.json`
2. pick largest M-floor bucket
3. else load DEFAULT and pick a different fallback scheme

Only the M-key prefix and the fallback differ.

Propose to shrink `_get_config`:

```python
def _pick_largest_floor(table: dict, prefix: str, value: int) -> dict | None:
    bounds = [int(k[len(prefix):]) for k in table if k.startswith(prefix)]
    hit = max((b for b in bounds if value >= b), default=None)
    return dict(table[f"{prefix}{hit}"]) if hit is not None else None

def _pick_smallest_ceil(table: dict, prefix: str, value: int, fallback_key="any") -> dict:
    for bound in sorted(int(k[len(prefix):]) for k in table if k.startswith(prefix)):
        if value <= bound:
            return dict(table[f"{prefix}{bound}"])
    return dict(table[fallback_key])

def _load_default(base: str, *, arch_fallback: str | None = None) -> dict:
    raw = load_config_json(f"{base}/DEFAULT.json", required=False)
    if raw is None and arch_fallback is not None:
        raw = load_config_json(
            f"{AITER_TRITON_CONFIGS_PATH}/{arch_fallback}/triton/fusions/"
            f"fused_clamp_act_mul/DEFAULT.json",
            required=True,
        )
    elif raw is None:
        raw = load_config_json(f"{base}/DEFAULT.json", required=True)
    return raw
```

Then `_get_config` collapses to something like:

```python
def _get_config(M, N, block_size_n, backend):
    base = f"{AITER_TRITON_CONFIGS_PATH}/{get_arch()}/{backend}/fusions/fused_clamp_act_mul"

    specialized = load_config_json(f"{base}/FUSED_CLAMP_ACT_MUL-N={N}.json", required=False)
    if specialized is not None:
        # Prefer one prefix for both backends going forward: "M_GEQ_"
        m_prefix = "M_GEQ_" if backend == "triton" else "M_LEQ_"  # temporary until gluon JSON migrated
        hit = _pick_largest_floor(specialized, m_prefix, M)
        if hit is not None:
            if backend == "gluon" and hit.get("BLOCK_SIZE_N") is None:
                hit["BLOCK_SIZE_N"] = block_size_n
            return hit

    if backend == "triton":
        return _pick_smallest_ceil(
            _load_default(base, arch_fallback="gfx950"), "N_LEQ_", block_size_n
        )

    config = dict(_load_default(base)["any"])
    if config["BLOCK_SIZE_N"] is None:
        config["BLOCK_SIZE_N"] = block_size_n
    return config
```
Please evaluate this suggestion with your own judgment, I didn't tested it by myself.

brunomazzottiamd (changes requested):
Proposed change makes sense. I'm only worried about code complexity and AI-driven code duplication. Please evaluate my requests with your own judgement.

aiter/ops/triton/configs/gfx950/triton/fusions/fused_clamp_act_mul/FUSED_CLAMP_ACT_MUL-N=2048.json
```diff
@@ -0,0 +1,8 @@
+{
+    "M_GEQ_4096": {
+        "num_warps": 2
+    },
+    "M_GEQ_16384": {
+        "num_warps": 1
+    }
+}
```
azaidy: Can you make the keys M_LEQ and any? If you need to keep M_GEQ fine but what happens for M=4095 and lower? You shouldnt switch to a default file. default file is when N doesnt match. for default config per file you need a key called `any`

aiter/ops/triton/configs/gfx950/triton/fusions/fused_clamp_act_mul/FUSED_CLAMP_ACT_MUL-N=4096.json
```diff
@@ -0,0 +1,5 @@
+{
+    "M_GEQ_4096": {
+        "num_warps": 4
+    }
+}
```
azaidy: Same as the other file?

aiter/ops/triton/fusions/fused_clamp_act_mul.py
azaidy: See how this is M_LEQ. We cant have both GEQ and LEQ keys for the same kernel in triton and gluon. Lets unify if possible?

azaidy: Can you check CLAUDE.md? There is a `resolve_config_dir()` function that you should use instead of manually constructing the path.

azaidy (changes requested):
1. Why tune for only 1 or 2 Ms? Why not more?
2. Can you state the benchmark command you used for the sweep just for reference so that we know if default args were used or more?

